PR #13, opened 2023-12-07
scx_utils::BpfBuilder: properly detect clang version in Ubuntu

arighi:
Apply the same logic of commit 00cd15a ("build: properly detect clang version in Ubuntu") in scx_utils as well.

This allows to build scx_utils properly in Ubuntu.

### DIFF
--- a/rust/scx_utils/src/bpf_builder.rs
+++ b/rust/scx_utils/src/bpf_builder.rs
@@ -217,6 +217,14 @@ pub struct BpfBuilder {
 }
 
 impl BpfBuilder {
+    fn skip_clang_version_prefix(line: &str) -> &str {
+        if let Some(index) = line.find("clang version") {
+            &line[index..]
+        } else {
+            line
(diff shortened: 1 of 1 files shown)


htejun:
@arighi Once things look okay for ubuntu builds, please let me know. I'll cut another release so that the published scx_utils works for ubuntu out of the box.

arighi:
> @arighi Once things look okay for ubuntu builds, please let me know. I'll cut another release so that the published scx_utils works for ubuntu out of the box.

@htejun With this additional fix applied I was able to build everything in scx on Ubuntu 24.04 (and I can use the test schedulers with my own sched-ext enabled Ubuntu kernel - that I'm planning to publish soon in a ppa).

I haven't packaged scx yet, so it still requires to manually apt install a bunch of dependent packages and build everything from source, but I'm planning to create a proper deb also for the scx stuff, one step at a time. :)

PR #19, opened 2023-12-09
build scx on 32-bit architectures

arighi:
A couple of small fixes that allow to build the scx schedulers also on 32-bit architectures (e.g., armhf).


arighi:
Ignore this PR, it actually introduces another bug. Will send another version soon.

PR #26, opened 2023-12-12
scx_central: Break dispatch_to_cpu loop when running out of buffer slots

kkdwivedi:
For the case where many tasks being popped from the central queue cannot be dispatched to the local DSQ of the target CPU, we will keep bouncing them to the fallback DSQ and continue the dispatch_to_cpu loop until we find one which can be dispatch to the local DSQ of the target CPU.

In a contrived case, it might be so that all tasks pin themselves to CPUs != target CPU, and due to their affinity cannot be dispatched to that CPU's local DSQ. If all of them are filling up the central queue, then we will keep looping in the dispatch_to_cpu loop and eventually run out of slots for the dispatch buffer. The nr_mismatched counter will quickly rise and the sched_ext core will notice the error and unload the BPF scheduler.

To remedy this, ensure that we break the dispatch_to_cpu loop when we can no longer perform a dispatch operation. The outer loop in central_dispatch for the central CPU should ensure the loop breaks when we run out of these slots and schedule a self-IPI to the local core, and allow the sched-ext core to consume the dispatch buffer before restarting the dispatch loop again.

A basic way to reproduce this scenario is to do:
taskset -c 0 perf bench sched messaging

The error in the kernel will be:
sched_ext: BPF scheduler "central" errored, disabling
sched_ext: runtime error (dispatch buffer overflow)
bpf_prog_6a473147db3cec67_dispatch_to_cpu+0xc2/0x19a
bpf_prog_c9e51ba75372a829_central_dispatch+0x103/0x1a5

### DIFF
--- a/scheds/kernel-examples/scx_central.bpf.c
+++ b/scheds/kernel-examples/scx_central.bpf.c
@@ -161,6 +161,14 @@ static bool dispatch_to_cpu(s32 cpu)
 			__sync_fetch_and_add(&nr_mismatches, 1);
 			scx_bpf_dispatch(p, FALLBACK_DSQ_ID, SCX_SLICE_INF, 0);
 			bpf_task_release(p);
+			/*
+			 * We might run out of dispatch buffer slots if we continue dispatching
+			 * to the fallback DSQ, without dispatching to the local DSQ of the
+			 * target CPU. In such a case, break the loop now as will fail the
+			 * next dispatch operation.
+			 */
+			if (!scx_bpf_dispatch_nr_slots())
+				break;
 			continue;
 		}
 


scheds/kernel-examples/scx_central.bpf.c
```diff
@@ -142,6 +142,13 @@ static bool dispatch_to_cpu(s32 cpu)
 	s32 pid;
 
 	bpf_repeat(BPF_MAX_LOOPS) {
+		/* We might run out of dispatch buffer slots if we continue dispatching
```
htejun, line 145: Maybe use fully winged comment style to be consistent with other comments?

scheds/kernel-examples/scx_central.bpf.c
```diff
@@ -142,6 +142,13 @@ static bool dispatch_to_cpu(s32 cpu)
...
+		 * to the fallback DSQ, without dispatching to the local DSQ of the
+		 * target CPU. In such a case, break the loop.
+		 */
+		if (!scx_bpf_dispatch_nr_slots())
```
htejun, line 149: Given that we know that there are slots on entry to the function, would it be better to check nr_slots in the FALLBACK_DSQ block right before continue?

htejun (approved):
Generally looks good to me. A couple minor suggestions.

scheds/kernel-examples/scx_central.bpf.c
kkdwivedi: Ack, will do.

scheds/kernel-examples/scx_central.bpf.c
kkdwivedi: Hmm, but I was thinking of a case where say you have 2 slots remaining, you pull out 2 tasks from the central_q which go to the fallback DSQ, and now you pull a 3rd which can go to the target CPU's local DSQ, but it will error out. So while it is true we have slots on entry it may not be true for the SCX_DSQ_LOCAL_ON case for later iterations.

Does that make sense?

kkdwivedi:
I have addressed both comments. I was a bit confused about the second one, please ignore that.

scheds/kernel-examples/scx_central.bpf.c
htejun: Hmm... so, we have either:
```
while (true) {
  if (no slot)
    break;
  find task;
  if (cpu doesn't match) {
    dispatch(FALLBACK);
    continue;
  }
  dispatch(target_cpu);
  break;
}
```

or 

```
while (true) {
  find task;
  if (cpu doesn't match) {
    dispatch(FALLBACK);
    if (more slots)
      continue;
    else
      break;
  }
  dispatch(target_cpu);
  break;
}
```

If we know that there are slots on entry, the behavior should be identical between the two, right? The only difference is that the former would do an extra check which will always indicate more slots.

PR #29, opened 2023-12-13
scx_userland: allocate tasks array based on kernel.pid_max

arighi:
Currently the array of enqueued tasks is statically allocated to a fixed size of USERLAND_MAX_TASKS to avoid potential deadlocks that could be introduced by performing dynamic allocations in the enqueue path.

However, this also adds a limit on the maximum pid that the scheduler can handle, since the pid is used as the index to access the array.

In fact, it is quite easy to trigger the following failure on an average desktop system (making this scheduler pretty much unusable in such scenario):

 $ sudo scx_userland
 ...
 Failed to enqueue task 33258: No such file or directory
 EXIT: BPF scheduler unregistered

Prevent this by using sysctl's kernel.pid_max as the size of the tasks array (and still allocate it all at once during initialization).

The downside of this change is that scx_userland may require additional memory to start and in small systems it could even trigger OOMs. For this reason add an explicit message to the command help, suggesting to reduce kernel.pid_max in case of OOM conditions.

### DIFF
--- a/scheds/kernel-examples/scx_userland.h
+++ b/scheds/kernel-examples/scx_userland.h
@@ -4,8 +4,6 @@
 #ifndef __SCX_USERLAND_COMMON_H
 #define __SCX_USERLAND_COMMON_H
 
-#define USERLAND_MAX_TASKS 8192
-
 /*
  * An instance of a task that has been enqueued by the kernel for consumption
  * by a user space global scheduler thread.
--- a/scheds/kernel-examples/scx_userland.bpf.c
+++ b/scheds/kernel-examples/scx_userland.bpf.c
@@ -23,6 +23,11 @@
 #include <scx/common.bpf.h>
 #include "scx_userland.h"
 
+/*
+ * Maximum amount of tasks enqueued/dispatched between kernel and user-space.
+ */
+#define MAX_ENQUEUED_TASKS 8192
+
 char _license[] SEC("license") = "GPL";
 
 const volatile bool switch_partial;
@@ -49,7 +54,7 @@ static bool usersched_needed;
  */
 struct {
 	__uint(type, BPF_MAP_TYPE_QUEUE);
-	__uint(max_entries, USERLAND_MAX_TASKS);
+	__uint(max_entries, MAX_ENQUEUED_TASKS);
 	__type(value, struct scx_userland_enqueued_task);
 } enqueued SEC(".maps");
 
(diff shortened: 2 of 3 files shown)


scheds/kernel-examples/scx_userland.h
```diff
@@ -4,6 +4,9 @@
 #ifndef __SCX_USERLAND_COMMON_H
 #define __SCX_USERLAND_COMMON_H
 
+/*
+ * Maximum amount of tasks enqueued/dispatched between kernel and user-space.
+ */
 #define USERLAND_MAX_TASKS 8192
```
Decave, line 10: Is this macro still used? If not let's just delete it?

Decave (changes requested):
LGTM, thanks for the fix Andrea. Just left one comment but otherwise it's good to go.

scheds/kernel-examples/scx_userland.h
htejun: It's still used in .bpf.c. There, it's just the max size of the ring buffer and overflow handling seems okay. Maybe it should be renamed for clarity?

Decave: Gah, right, my bad. Yeah, maybe something like `MAX_ENQUEUED_TASKS`? And we can probably just bring it into scx_userland.bpf.c directly. It's also an option to set the max entries value from userspace before the map is loaded, but this seems preferable given the fallback logic as you pointed out.

arighi: It's quite an internal parameter, basically it defines how many items can be pushed at most to the enqueued/dispatched maps, so I don't think there's much value to export this to user-space as a configurable option... but I think we can definitely bring it into .bpf.c, since it's used only there at this point. And I like the name `MAX_ENQUEUED_TASKS`, it's definitely more clear.

Decave: Oh yeah, I meant that you could set the max entries in the map to the value of `/proc/sys/kernel/pid_max`. But yeah I think bringing the macro into the bpf file and statically setting it there is fine / preferable.

arighi: ... and updated.

Setting pid-max entries in the bpf maps would be a bit of a waste IMHO, there's no way a system will have pid-max concurrent tasks running. :)

And if we fail to enqueue a task from in the bpf part the task will just be dispatched immediately. OTOH if we fail to push the task to the dispatched map from the .c we would error and exit, maybe this part can be improved a bit, i.e., re-adding the pid to the vruntime_head list?

Decave (approved):
Looks great, thanks a lot @arighi

scheds/kernel-examples/scx_userland.h
Decave: Agreed, sounds like another nice improvement!

PR #31, opened 2023-12-14
minor refactor of scx_rusty

jordalgo:


davemarchevsky:
Refactoring LGTM, not very familiar w/ `scx_rusty` BPF logic so will let someone else push the button

scheds/rust-user/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -425,10 +425,13 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
-	if (kthreads_local &&
-	    (p->flags & PF_KTHREAD) && p->nr_cpus_allowed == 1) {
+	if (p->nr_cpus_allowed == 1) {
 		cpu = prev_cpu;
-		stat_add(RUSTY_STAT_DIRECT_DISPATCH, 1);
+		if (kthreads_local && (p->flags & PF_KTHREAD)) {
+			stat_add(RUSTY_STAT_DIRECT_DISPATCH, 1);
+		} else {
+			stat_add(RUSTY_STAT_PINNED, 1);
+		}
```
Decave, line 434: I think we can simplify the `if` on line 442 by removing the first `p->nr_cpus_allowed > 1` clause, which should always be true now that we're always doing local dispatching if `p->nr_cpus_allowed == 1`

Decave:
Looks great, thanks for the cleanup. Just left a comment for one more small cleanup we can do because of this refactor.

scheds/rust-user/scx_rusty/src/bpf/main.bpf.c
jordalgo: Good call. Will fix.

Decave (approved):

PR #40, opened 2023-12-18
scx: Add CI action that builds schedulers for PRs

Decave:
When Ubuntu ships with sched_ext, we can also maybe test loading the schedulers (not sure if the runners can run as root though). For now, we should at least have a CI job that lets us verify that the schedulers can _build_. To that end, this patch adds a basic CI action that builds the schedulers.

As is, this is a bit brittle in that we're having to manually download and install a few dependencies. I don't see a better way for now without hosting our own runners with our own containers, but that's a bigger investment. For now, hopefully this will get us _some_ coverage.

### DIFF
--- a/.github/workflows/build-scheds.yml
+++ b/.github/workflows/build-scheds.yml
@@ -1,0 +1,49 @@
+name: build-scheds
+run-name: ${{ github.actor }} PR run
+on: [pull_request]
+jobs:
+  build-schedulers:
+    runs-on: ubuntu-20.04
+    steps:
+
+      ### DOWNLOAD AND INSTALL DEPENDENCIES ###
+
+      # Download dependencies packaged by Ubuntu
+      - run: sudo apt update
+      - run: sudo apt -y install cmake cargo elfutils libelf-dev linux-headers-generic linux-tools-common ninja-build
+
+      # clang 17
+      - run: wget https://apt.llvm.org/llvm.sh
+      - run: chmod +x llvm.sh
+      - run: sudo ./llvm.sh all
+      - run: sudo ln -sf /usr/bin/clang-17 /usr/bin/clang
+      - run: sudo ln -sf /usr/bin/llvm-strip-17 /usr/bin/llvm-strip
+
+      # bpftool
+      - run: git clone --recurse-submodules --branch v7.3.0 https://github.com/libbpf/bpftool.git
+      - run: make -j -C bpftool/src
+      - run: sudo make -j -C bpftool/src install
+
+      # zstd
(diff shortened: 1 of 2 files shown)


Decave:
The CI check failed here even though it passed locally. Will convert to a draft and debug.

htejun (approved):
Please merge whenever it's ready. Out of curiosity, why do we need thread dep in meson?

Decave:
> Please merge whenever it's ready. Out of curiosity, why do we need thread dep in meson?

scx_userland was failing to resolve pthread_create(). Not sure why that's not happening locally, but must be some different dependency.

.github/workflows/build-scheds.yml
```diff
@@ -0,0 +1,49 @@
+name: build-scheds
+run-name: ${{ github.actor }} PR run
+on: [pull_request]
+jobs:
+  build-schedulers:
+    runs-on: ubuntu-20.04
```
arighi, line 6: I'm wondering if we should target `ubuntu-22.04` (or potentially even `ubuntu-latest`) instead, in this way we don't have to download llvm 17 or any other external package, all we need is in the distro.

arighi: Another cool thing that we could do is to use `virtme-ng` (https://github.com/arighi/virtme-ng) to actually run the scx schedulers with any kernel that we want. In this case for example, we could `apt download` the latest kernel from my ppa (with the sched-ext support), unpack the kernel, then use `vng -r /path/to/vmlinuz --user root -- scx_simple` and test the scheduler without having to install a custom kernel.

Decave: > I'm wondering if we should target `ubuntu-22.04` (or potentially even `ubuntu-latest`) instead, in this way we don't have to download llvm 17 or any other external package, all we need is in the distro.

Heh, I'd originally tried to use ubuntu-22.04, but apt-add-repository was hanging for some of the external dependencies. If we can just use everything in the distro itself then that would be great. I'll spin up another PR to try and get it working.

>Another cool thing that we could do is to use virtme-ng (https://github.com/arighi/virtme-ng) to actually run the scx schedulers with any kernel that we want. In this case for example, we could apt download the latest kernel from my ppa (with the sched-ext support), unpack the kernel, then use vng -r /path/to/vmlinuz --user root -- scx_simple and test the scheduler without having to install a custom kernel.

It would be _great_ if we could use virtme-ng. I'd actually like to use that with the sched-ext repo as well, so if we can get it figured out here then I'll also get it setup there.

PR #42, opened 2023-12-20
ci: Run CI job on Ubuntu 22.04

Decave:
Andrea pointed out that we can and should be using Ubuntu 22.04. Unfortunately it still doesn't ship some of the deps we need like clang-17, but it does at least ship virtme-ng, so it's good for us to use this so that we can actually test running the schedulers in a virtme-ng VM when it supports being run in docker.


Decave:
Passing locally but failing on CI. Let me close this so I can test with a push before opening the actual PR

PR #43, opened 2023-12-20
ci: Run CI job on Ubuntu 22.04

Decave:
Andrea pointed out that we can and should be using Ubuntu 22.04. Unfortunately it still doesn't ship some of the deps we need like clang-17, but it does at least ship virtme-ng, so it's good for us to use this so that we can actually test running the schedulers in a virtme-ng VM when it supports being run in docker.

Also, update the job to run on pushes, and not just when a PR is opened

### DIFF
--- a/.github/workflows/build-scheds.yml
+++ b/.github/workflows/build-scheds.yml
@@ -1,19 +1,31 @@
 name: build-scheds
 run-name: ${{ github.actor }} PR run
-on: [pull_request]
+on: [pull_request, push]
 jobs:
   build-schedulers:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
+      ### OTHER REPOS ####
+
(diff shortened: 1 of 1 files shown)


htejun (approved):

arighi (commented):
virtme-ng brings a lot of dependencies (all the kernel build dependencies basically). If this is slowing down the CI too much we may also consider to `git clone https://github.com/arighi/virtme-ng.git` and install only a subset of dependencies (that should be this: `python3-pip qemu-system-x86 qemu-kvm libvirt-clients busybox-static kbd kmod file rsync zstd`), like the bare minimum requirements to run `vng` and test a custom kernel.

But apart than that everything else looks good.

In the meantime I'll work on a fix for virtme-ng to properly support docker (see https://github.com/arighi/virtme-ng/issues/51).

PR #46, opened 2023-12-22
scx_rustland: handle graceful vs non-graceful exit

arighi:
Do not report an exit error message if it's not present. Moreover, distinguish between a graceful exit vs a non-graceful exit.

NOTE: in the future the whole exit handling probably can be moved to a more generic place (scx_utils) to prevent code duplication across schedulers and also to prevent small inconsistencies like this one.

### DIFF
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -275,17 +275,17 @@ impl<'a> Scheduler<'a> {
 
     // Called on exit to get exit code and exit message from the BPF part.
     fn report_bpf_exit_kind(&mut self) -> Result<()> {
+        let cstr = unsafe { CStr::from_ptr(self.skel.bss().exit_msg.as_ptr() as *const _) };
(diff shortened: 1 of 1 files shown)


arighi:
Hm... actually I'm not sure about this change, it looks like other schedulers are doing different checks. For example `scx_layered` is doing `if etype != 64 ` to determine if we need to print a message, `scx_rusty` is doing this check. The C schedulers instead are checking if the message is empty (in `uei_print()`).

Looking at the kernel code it seems that we may want to print the message if `etype >= 64`, that seems to be the limit between graceful and non-graceful exits.

What is the correct API?

arighi:
Pushed another version that makes more sense, or at least it seems more consistent with the kernel's `scx_exit_kind` values.

htejun:
Following what user_exit_info.h is doing is probably the best. Can you please update accordingly? Thanks.

arighi:
I will include this change in a new PR with more improvements for scx_rustland, closing this one for now.

PR #47, opened 2023-12-23
scx_rustland improvements

arighi:
Set of changes for scx_rustland that allow to massively improve the user-space scheduler effectiveness, especially for low latency tasks:
 - provide CPU usage awareness to the user-space scheduler
 - dispatch tasks from the user-space scheduler to the BPF dispatcher in batch, instead of draining the entire task pool all at once
 - distinguish between graceful exit vs non-graceful exit

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/intf.h
+++ b/scheds/rust/scx_rustland/src/bpf/intf.h
@@ -33,6 +33,7 @@ typedef long long s64;
  */
 struct queued_task_ctx {
 	s32 pid;
+	s32 cpu; /* CPU where the task is running */
 	u64 sum_exec_runtime; /* Total cpu time */
 	u64 weight; /* Task static priority */
 };
@@ -49,6 +50,7 @@ struct queued_task_ctx {
  */
 struct dispatched_task_ctx {
 	s32 pid;
+	s32 cpu; /* CPU where the task should be dispatched */
 	u64 payload; /* Task payload */
 };
 
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -35,6 +35,8 @@
 #include <scx/common.bpf.h>
 #include "intf.h"
 
+#define MAX_CPUS 1024
+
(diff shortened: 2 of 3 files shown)


arighi:
Test result: before this change running a `make -j16` in the kernel source dir was tanking my laptop, after these changes I can still watch a video on Youtube while `make -j16` is running.

htejun:
I generally like the direction. This likely is less performant but gives userspace a lot more control over what happens on the CPUs, and optimizing from there if necessary seems like the right approach. Generally looks good to me. A few comments:

- Now that the interaction between the BPF and rust parts are clearer, it might make sense to define the interlocking more concretely. BPF doesn't have full memory model yet but I don't think we need them here anyway. Setting `usersched_needed` should be a release operation and reading it and clearing it should be an acquire. Setting of `usersched_needed` in `rustland_enqueue()` already has full mb in the preceding `nr_enqueues` increment and reading and clearing are followed by multiple spinlocks which provide the needed rmb ordering. Just noting as such should be sufficient here.
- I wonder whether `rustland_stopping()` needs to set `usersched_needed` as the CPU might be running out of things to do; otherwise, there may be unnecessary upto 1s gaps which doesn't look like it would be too difficult to trigger. An alternative and probably more robust implementation would be adding `.update_idle()` method which is called when the CPU is about to go idle, keep track of the number of tasks queued in userspace and trigger usersched iff there are pending tasks when a CPU is about to go to idle. Strictly speaking, this should be the only hand-over mechanism that should be necessary and can avoid e.g. triggering usersched unnecessarily after every enqueue while all CPUs are still busy.
- Note that implementing `.update_idle()` will disable the bulit in idle tracking and `scx_bpf_test_and_clear_cpu_idle()` will stop working. You can keep the builtin idle tracking on by setting `SCX_OPS_KEEP_BUILTIN_IDLE`. However, given that the benefit the code is getting from default idle tracking is rather minimal and doing duplicate idle tracking anyway, maybe the right thing to do is implementing a better "keep running" logic in `.select_cpu()` based on the number of tasks queued on the rust side. ie. if there's nothing to switch to, just keep running?

PR #48, opened 2023-12-27
scx_rustland: clarify and improve BPF / userspace interlocking

arighi:
A set of interlocking improvements for scx_rustland:
 - clarify interlocking between the BPF dispatcher and the user-space scheduler
 - introduce get/set_cpu_owner() primitives as a baseline to implement a better idle tracking in the future
 - move user-space scheduler activation from `.enqueue()` to `.stopping()` (and activate only when needed) to reduce unnecessary user-space activations
 - always bypass user-space scheduling for kthreads

I have also explored the possibility to move the user-space scheduler activation from `.stopping()` to `.update_idle()`, but the scheduler performs quite poorly with this change: it seems that CPUs are not fully used even in presence of CPU intensive workloads, as if the user-space scheduler isn't feeding the dispatcher fast enough. My guess is that activating the user-space scheduler in `.update_idle()` is a bit too late and we get bubbles in the whole scheduling pipeline, compared to do the activation in `.stopping`. But this is something that I'd like to explore more for the next set of improvements for scx_rustland.

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -297,12 +297,12 @@ static bool is_task_cpu_available(struct task_struct *p)
 	struct task_ctx *tctx;
 
 	/*
-	 * Always dispatch per-CPU kthread on the same CPU, bypassing the
-	 * user-space scheduler (in this way we can to prioritize critical
-	 * kernel threads that may potentially slow down the entire system if
-	 * they are blocked for too long).
+	 * Always dispatch kthread on the same CPU, bypassing the user-space
+	 * scheduler (in this way we can to prioritize critical kernel threads
+	 * that may potentially slow down the entire system if they are blocked
+	 * for too long).
 	 */
-	if (is_kthread(p) && p->nr_cpus_allowed == 1)
+	if (is_kthread(p))
 		return true;
 
 	/*
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -49,7 +49,6 @@ char exit_msg[SCX_EXIT_MSG_LEN];
  * Scheduler attributes and statistics.
  */
 u32 usersched_pid; /* User-space scheduler PID */
-static volatile bool usersched_needed; /* Used to wake-up the user-space scheduler */
 const volatile bool switch_partial; /* Switch all tasks or SCHED_EXT tasks */
 const volatile u64 slice_ns = SCX_SLICE_DFL; /* Base time slice duration */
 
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -244,12 +297,12 @@ static bool is_task_cpu_available(struct task_struct *p)
...
-	 * user-space scheduler (in this way we can to prioritize critical
-	 * kernel threads that may potentially slow down the entire system if
-	 * they are blocked for too long).
+	 * Always dispatch kthread on the same CPU, bypassing the user-space
+	 * scheduler (in this way we can to prioritize critical kernel threads
+	 * that may potentially slow down the entire system if they are blocked
+	 * for too long).
 	 */
-	if (is_kthread(p) && p->nr_cpus_allowed == 1)
+	if (is_kthread(p))
```
htejun, line 305: Not an immediate blocker but this might lead to not-so-great behaviors for things like unbound workqueue workers that can consume a lot of CPU cycles (e.g. encryption workers). They actually need scheduling.

htejun (approved):
For the `update_idle` bubbles, I wonder whether what you need is `scx_bpf_kick_cpu()` on self. There's nothing triggering immediate dispatch otherwise. Another thing which might be interesting is that as the BPF and usersched parts get more tightly interlocked, it may make sense to have per-cpu usersched threads. Otherwise, some CPUs may go idle while waiting for usersched thread. If there's a dedicated usersched thread per cpu, every cpu can jump into the usersched thread when it runs out of things to do and then usersched can figure out synchronization (using raw spinlocks) and so on. While conceptually neat in that it really just moves per-cpu scheduling logic to userspace, this might be too much complexity for the purpose of this scheduler, so this is more dumping of the train of thought than anything.

arighi:
> For the `update_idle` bubbles, I wonder whether what you need is `scx_bpf_kick_cpu()` on self. There's nothing triggering immediate dispatch otherwise. Another thing which might be interesting is that as the BPF and usersched parts get more tightly interlocked, it may make sense to have per-cpu usersched threads. Otherwise, some CPUs may go idle while waiting for usersched thread. If there's a dedicated usersched thread per cpu, every cpu can jump into the usersched thread when it runs out of things to do and then usersched can figure out synchronization (using raw spinlocks) and so on. While conceptually neat in that it really just moves per-cpu scheduling logic to userspace, this might be too much complexity for the purpose of this scheduler, so this is more dumping of the train of thought than anything.

I'll do some experiments with `scx_bpf_kick_cpu()`, it should be enough to prevent the bubbles and it shouldn't add too much complexity. The per-cpu usersched is an interesting idea, I'll try to do some experiments as well, but unless it provides consistent performance improvements I'd stick with simplicity. :)

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: > Not an immediate blocker but this might lead to not-so-great behaviors for things like unbound workqueue workers that can consume a lot of CPU cycles (e.g. encryption workers). They actually need scheduling.

Good point! Yes, in that case they definitely need scheduling, I'll revert that in the next set of changes.

PR #50, opened 2023-12-27
Downgrade bindgen to 0.68

jordalgo:
This is so we can package scx_utils into fedora without having to upgrade rust-bindgen
(https://bodhi.fedoraproject.org/updates/FEDORA-2023-18e7f124e1).

To make this happen we need to stop using the `CargoCallbacks::new` constructor which was added in 0.69. Old way seems legit according to the docs:
https://rust-lang.github.io/rust-bindgen/non-system-libraries.html

### DIFF
--- a/rust/scx_utils/Cargo.toml
+++ b/rust/scx_utils/Cargo.toml
@@ -9,7 +9,7 @@ description = "Utilities for sched_ext schedulers"
 
 [dependencies]
 anyhow = "1.0"
-bindgen = "0.69"
+bindgen = ">=0.68, <0.70"
 glob = "0.3"
 lazy_static = "1.4"
 libbpf-cargo = "0.22"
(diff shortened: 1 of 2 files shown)


jordalgo:
Now we get a warning if we build with 0.69:
 ```
 [29/32] Generating rust/scx_utils/scx_utils with a custom command (wrapped by meson to set env)
warning: use of deprecated constant `bindgen::CargoCallbacks`: Use `CargoCallbacks::new()` instead. Please, check the documentation for further information.
   --> src/bpf_builder.rs:509:48
    |
509 |             .parse_callbacks(Box::new(bindgen::CargoCallbacks))
    |                                                ^^^^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default
```

davide125 (approved):

htejun (approved):

Issue #49, opened 2023-12-27
potential deadlock with scx_flatcg


arighi:
scx_flatcg is calling  `bpf_cgroup_from_id()` from `.dispatch()`, that is called with `rq->__lock` held, `bpf_cgroup_from_id()` is basically a kfunc wrapper of `cgroup_get_from_id()` that is calling `kernfs_find_and_get_node_by_id()`, that can acquire `kernfs_idr_lock` (with `rq->_lock` held).

In kernfs we can acquire `kernfs_idr_lock`, be interrupted by a scheduler tick and acquire `rq->__lock`, leading to the following deadlock scenario:
```
        CPU0                    CPU1
        ----                    ----
   lock(kernfs_idr_lock);
                                lock(rq->__lock);
                                lock(kernfs_idr_lock);
   <Interrupt>
    lock(rq->__lock);
```

I don't know exactly how to prevent this without doing a big redesign of scx_flatcg or changing kernfs to use irqsave/irqrestore spinlocks (that seems a pretty strong change... but probably the safest if we want to make `bpf_cgroup_from_id()` available for general usage).

Suggestions?

htejun:
Yeah, the right thing to do is making kernfs_idr_lock irqsafe and that shouldn't make anything noticeably worse. Hmm... given that bpf_cgroup_from_id() is a kfunc which doesn't have any restrictions, it could be that this deadlock is already possible with other BPF progs. If so, it's an easy patch to justify. I'll ask BPF folks whether that's already the case.

arighi:
FWIW, I did a test with this applied and everything looks good: https://github.com/sched-ext/sched_ext/compare/sched_ext...arighi:sched_ext:kernfs-irq-spinlock

I was able to trigger the lockdep splat running scx_flatcg inside virtme-ng:
```
vng -v -- ./build/scheds/c/scx_flatcg
```

BTW I'm also working on a github workflow to test the schedulers automatically using virtme-ng, that should help to catch issues like this in advance, I may be able to send a PR later today or tomorrow.

htejun:
We'll probably have to use raw_spinlock given that it can be nested inside rq lock. BPF folks say it should be possible to trigger the same condition without scx given that the kfunc can be called from any kprobe/tracepoint/fentry attached BPF progs.

htejun:
This issue should be fixed by: https://github.com/sched-ext/sched_ext/commit/1225a9069ff2185007b078b7e447727d229a64fe

PR #51, opened 2023-12-28
test the schedulers in the github workflow using virtme-ng

arighi:
Use virtme-ng to test the schedulers in the github workflow.

Keep in mind that I've only tested this locally (using `gh act`), so in the actual github workflow it may behave differently and it probably requires more testing / planning.

I've opened the PR to start some discussions / suggestions / ideas for now.

Overall, I think it can be really useful, I've already triggered a few bugs with this (see https://github.com/sched-ext/scx/issues/49, https://github.com/sched-ext/sched_ext/pull/101 and https://github.com/sched-ext/sched_ext/issues/74#issuecomment-1870958545).

### DIFF
--- a/.github/workflows/build-scheds.yml
+++ b/.github/workflows/build-scheds.yml
@@ -7,18 +7,20 @@ jobs:
     steps:
       ### OTHER REPOS ####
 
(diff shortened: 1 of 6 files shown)


Decave:
Wow, awesome, thanks Andrea!

PR #53, opened 2023-12-28
Suppress the deprecation warning from bindgen and bump versions

htejun:


jordalgo:
LGTM. Tracking here too: https://github.com/sched-ext/scx/issues/54

PR #56, opened 2023-12-29
scx_rustland: reduce scheduler overhead

arighi:
- always consider the CPU where the scheduler is running as idle
 - fix a bug that was causing the user-space scheduler to spin and improve the logic to activate the user-space scheduler, preventing unnecessary activations

These changes make scx_rustland more reliable and they strongly reduce the scheduler overhead.

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -162,7 +162,7 @@ static void set_cpu_owner(u32 cpu, u32 pid)
 		scx_bpf_error("Failed to look up cpu_map for cpu %u", cpu);
 		return;
 	}
-	*owner= pid;
+	*owner = pid;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -489,7 +505,7 @@ void BPF_STRUCT_OPS(rustland_update_idle, s32 cpu, bool idle)
 	 * Moreover, kick the CPU to make it immediately ready to accept
 	 * dispatched tasks.
 	 */
-	if (__sync_fetch_and_add(&nr_queued, 0)) {
+	if (nr_queued || nr_scheduled) {
```
htejun, line 508: It may be worth explaining how this is interlocked with userland component and a situation where userland component is not invoked even while there are tasks to process can't happen. nr_queued is basically used as a boolean here indicating "something happened since userland looked at it, so better call them at least once". And then while the userland is running it updates its internal state through nr_scheduled. As long as update_idle is called after each userland run, this should be safe, right?

htejun (changes requested):
The commit messages are titled `scx_userland`. Can you please update them?

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: @htejun comment updated adding more details. Let me know if it seems clear enough. Thanks!

htejun (approved):

PR #57, opened 2023-12-30
scx_rustland: improve scheduler cpu selection

arighi:
Implement a more reasonable idle CPU selection logic in the user-space scheduler.

### DIFF
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -216,7 +216,7 @@ struct Scheduler<'a> {
     task_pool: TaskTree,   // tasks ordered by vruntime
     task_map: TaskInfoMap, // map pids to the corresponding task information
     min_vruntime: u64,     // Keep track of the minimum vruntime across all tasks
-    nr_cpus_online: u64,   // Amount of the available CPUs in the system
+    nr_cpus_online: i32,   // Amount of the available CPUs in the system
     struct_ops: Option<libbpf_rs::Link>,
 }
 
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rustland/src/main.rs
```diff
@@ -407,18 +389,27 @@ impl<'a> Scheduler<'a> {
...
                     // Update global minimum vruntime.
                     self.min_vruntime = task.vruntime;
 
+                    // Select a CPU to dispatch the task.
+                    //
+                    // Use the previously used CPU if idle, that is always the best choice (to
+                    // mitigate migration overhead), otherwise pick the next idle CPU available.
+                    if !idle_cpus.contains(&task.cpu) {
+                        task.cpu =  *cpu;
+                    }
```
htejun, line 411: This still can lead to multiple tasks being assigned to the same cpu, right? Something to improve in the future, I guess.

htejun (approved):

scheds/rust/scx_rustland/src/main.rs
arighi: @htejun exactly, we can still assign the same idle cpu to multiple tasks if their `prev_cpu` is still idle, that is not super bad, being too aggressive with migrations also comes with a cost. We could also speculate on the fact that tasks with a close vruntime should be spread across multiple CPUs, so that they can run pretty much at the same time in parallel, rather than piling them up on the same idle CPU (that is also something that I'd like to experiment).

But there is definitely room for improvements in the cpu selection logic and in the userspace scheduler in general.

And I'm happy to see that we are reaching a stage where the BPF component is becoming a more "stable" backend (requiring minimal changes), so that we can start channeling all the optimizations, improvements, tests and experiments in the user-space part. If that's the case it would be a nice success story for this little project IMHO. :)

PR #58, opened 2023-12-31
scx_rustland: prevent dispatching multiple tasks on the same idle cpu

arighi:
When a task is dispatched we always try to pick the previously used CPU (if idle) to minimize the migration overhead. Alternatively, if such CPU is not available, we pick any other idle CPU in the system.

However, we don't update the list of idle CPUs as we dispatch tasks, therefore we may end up sending multiple tasks to the same idle CPU (if their previously used CPU is the same) and we may even skip some idle CPUs completely.

Change this logic to make sure that we never dispatch multiple tasks to the same idle CPU, by updating the list of idle CPUs as we send tasks to the BPF dispatcher.

This also avoids dispatching tasks with a closely matched vruntime to the same CPU, thereby negating the advantages of the vruntime ordering. With this change in place, we ensure that tasks with a similar vruntime are dispatched to different CPUs, leading to significant improvements in latency performance.

### DIFF
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -389,14 +389,14 @@ impl<'a> Scheduler<'a> {
     fn dispatch_tasks(&mut self) {
         let maps = self.skel.maps();
         let dispatched = maps.dispatched();
-        let idle_cpus = self.get_idle_cpus();
+        let mut idle_cpus = self.get_idle_cpus();
 
         // Dispatch only a batch of tasks equal to the amount of idle CPUs in the system.
         //
         // This allows to have more tasks sitting in the task pool, reducing the pressure on the
         // dispatcher queues and giving a chance to higher priority tasks to come in and get
         // dispatched earlier, mitigating potential priority inversion issues.
-        for cpu in &idle_cpus {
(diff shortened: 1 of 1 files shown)


htejun (approved):

scheds/rust/scx_rustland/src/main.rs
```diff
@@ -406,8 +406,14 @@ impl<'a> Scheduler<'a> {
                     //
                     // Use the previously used CPU if idle, that is always the best choice (to
                     // mitigate migration overhead), otherwise pick the next idle CPU available.
-                    if !idle_cpus.contains(&task.cpu) {
-                        task.cpu =  *cpu;
+                    if let Some(id) = idle_cpus.iter().position(|&x| x == task.cpu) {
```
htejun, line 409: Bike-shedding but I find it a little confusing to receive position as id. Maybe something like `idx`, `i` or `pos` is better?

htejun (approved):

scheds/rust/scx_rustland/src/main.rs
arighi: @htejun yeah it is confusing, id sounds more like the value in this context, not the position, pos is a lot more clear.

PR #59, opened 2023-12-31
scx_rustland: lowlatency improvements

arighi:
Low-latency improvements:
 - scx_rustland: always use dispatch_on_cpu() when possible
 - scx_rustland: bypass user-space scheduler for short-lived kthreads 

Reduce user-space overhead:
 - scx_rustland: enable SCX_OPS_ENQ_LAST

UI improvement:
 - scx_rustland: show the CPU where the scheduler is running

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -450,11 +450,9 @@ void BPF_STRUCT_OPS(rustland_dispatch, s32 cpu, struct task_struct *prev)
 		 * task and migrate (if possible); otherwise, dispatch on the
 		 * global DSQ.
 		 */
-		prev_cpu = scx_bpf_task_cpu(p);
-		dbg_msg("usersched: pid=%d prev_cpu=%d cpu=%d payload=%llu",
+		dbg_msg("usersched: pid=%d cpu=%d payload=%llu",
 			task.pid, task.cpu, task.payload);
-		if ((task.cpu != prev_cpu) &&
-		   bpf_cpumask_test_cpu(task.cpu, p->cpus_ptr))
+		if (bpf_cpumask_test_cpu(task.cpu, p->cpus_ptr))
 			dispatch_on_cpu(p, task.cpu, 0);
 		else
 			dispatch_global(p, 0);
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -322,6 +322,12 @@ static bool is_task_cpu_available(struct task_struct *p)
 	if (is_kthread(p) && p->nr_cpus_allowed == 1)
 		return true;
 
+	/*
+	 * No scheduling required if it's the last task running.
+	 */
+        if (enq_flags & SCX_ENQ_LAST)
+		return true;
```
htejun, line 342: I don't quite understand what difference this commit makes given that the default behavior w/o SCX_ENQ_LAST is auto-enq on local dsq. Does this make any behavior difference?

htejun (approved):

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: > I don't quite understand what difference this commit makes given that the default behavior w/o SCX_ENQ_LAST is auto-enq on local dsq. Does this make any behavior difference?

For some reasons with this in place the CPU utilization of scx_rustland drops significantly when the system is idle (like 0.3-0.5%), without this change when the system is idle I can see some cpu usage spikes of up to 5-10%,but I'm not sure why it's happening... My theory was that with SCX_ENQ_LAST in place we could save some unnecessary usersched invocations.

htejun: That's surprising given that the two essentially are doing the same thing. Maybe there's something timing dependent going on or I'm just confused.

arighi: > That's surprising given that the two essentially are doing the same thing. Maybe there's something timing dependent going on or I'm just confused.

I'll run some tests with and without that and collect more info, now I want to understand what's going on exactly :)

arighi: @htejun it's SCX_OPS_ENQ_LAST set in sched_ext_ops.flags that seems to make a difference, the check in `is_task_cpu_available()` does't make any difference and can be dropped apparently.

PR #60, opened 2024-01-01
scx_rustland: prevent starvation and improve responsiveness

arighi:
Improvements and fixes to scx_rustland to prevent some cases of starvation and enhance responsiveness:
 - scx_rustland: evaluate the proper vruntime delta
 - scx_rustland: prevent starvation handling short-lived tasks properly
 - bpf_rustland: do not dispatch the scheduler to the global DSQ
 - scx_rustland: remove SCX_ENQ_LAST check in is_task_cpu_available()
 - scx_rustland: never account more than slice_ns to vruntime
 - scx_rustland: clean up old entries in the task map

Small code readability improvement:
- scx_rustland: small code refactoring
- scx_rustland: rename variable id -> pos for better clarity

With all these changes in place the scheduler can properly handle, on my 8-cores laptop, a parallel kernel build (`make -j32`), a stress test (`stress-ng --cpu 32 --iomix 4 --vm 2 --vm-bytes 128M --fork 4`), a YouTube music video playing in the background, all while reading emails and writing the text of this PR.

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -336,12 +336,6 @@ static bool is_task_cpu_available(struct task_struct *p, u64 enq_flags)
 		return true;
 
 	/*
-	 * No scheduling required if it's the last task running.
-	 */
-        if (enq_flags & SCX_ENQ_LAST)
-		return true;
-
-	/*
 	 * For regular tasks always rely on force_local to determine if we can
 	 * bypass the scheduler.
 	 */
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -252,18 +252,15 @@ static void dispatch_local(struct task_struct *p, u64 enq_flags)
  */
 static void dispatch_on_cpu(struct task_struct *p, s32 cpu, u64 enq_flags)
 {
+	/*
+	 * If it's not possible to dispatch on the selected CPU, re-use the
+	 * previously used one.
+	 */
+	if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr))
+		cpu = scx_bpf_task_cpu(p);
 	dbg_msg("%s: pid=%d cpu=%ld", __func__, p->pid, cpu);
 	scx_bpf_dispatch(p, SCX_DSQ_LOCAL_ON | cpu, slice_ns,
 			 enq_flags | SCX_ENQ_LOCAL);
-}
-
-/*
- * Dispatch a task on the global FIFO.
- */
-static void dispatch_global(struct task_struct *p, u64 enq_flags)
-{
-	dbg_msg("%s: pid=%d", __func__, p->pid);
-	scx_bpf_dispatch(p, SCX_DSQ_GLOBAL, slice_ns, enq_flags);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -252,20 +252,17 @@ static void dispatch_local(struct task_struct *p, u64 enq_flags)
  */
 static void dispatch_on_cpu(struct task_struct *p, s32 cpu, u64 enq_flags)
 {
+	/*
+	 * If it's not possible to dispatch on the selected CPU, re-use the
+	 * previously used one.
+	 */
+	if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr))
+		cpu = scx_bpf_task_cpu(p);
```
htejun, line 260: I think this should work fine but can you please verify that changing CPU affinity for a constantly running task which is bound to one CPU works? ie. Have one thread which busy loops, bind it to one cpu using `taskset`, and then change it to another CPU.

scheds/rust/scx_rustland/src/main.rs
```diff
@@ -351,11 +351,15 @@ impl<'a> Scheduler<'a> {
...
+        max_slice_ns: u64,
     ) {
         // Add cputime delta normalized by weight to the vruntime (if delta > 0).
         if sum_exec_runtime > task_info.sum_exec_runtime {
-            let delta = (sum_exec_runtime - task_info.sum_exec_runtime) / weight;
-            task_info.vruntime += delta;
+            let mut delta = sum_exec_runtime - task_info.sum_exec_runtime;
+            // Never account more than max_slice_ns. This helps to prevent starving a task for too
+            // long in the scheduler task pool.
+            delta = delta.min(max_slice_ns);
```
htejun: I'm curious why this is necessary. Maybe this is needed to mask a different issue such as failing to wind current vtime forward when it should be? Or is this necessary because sometimes the scheduler takes a bit too long to kick them off cpu and their execution pattern becomes too chunky?

htejun (approved):

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: Yep, it seems to work:
```
09:10 arighi@gpd$ yes >/dev/null &
[1] 1797739
6.7.0-3-generic ~
09:10 arighi@gpd$ taskset -pc 2 $!
pid 1797739's current affinity list: 0-7
pid 1797739's new affinity list: 2
6.7.0-3-generic ~
09:10 arighi@gpd$ cut -d' ' -f39 /proc/$!/stat
2
6.7.0-3-generic ~
09:11 arighi@gpd$ taskset -pc 5 $!
pid 1797739's current affinity list: 2
pid 1797739's new affinity list: 5
6.7.0-3-generic ~
09:11 arighi@gpd$ cut -d' ' -f39 /proc/$!/stat
5
```

Of course it changes cpu only if the task is doing something (i.e. a `sleep infinity` won't change cpu), but the same happens with the default scheduler, so we shouldn't change any behavior here.

scheds/rust/scx_rustland/src/main.rs
arighi: > I'm curious why this is necessary. Maybe this is needed to mask a different issue such as failing to wind current vtime forward when it should be? Or is this necessary because sometimes the scheduler takes a bit too long to kick them off cpu and their execution pattern becomes too chunky?

I think I was just trying to fix what was fixed by `2900b20 ("scx_rustland: evaluate the proper vruntime delta")` here, I should have squashed the commits together...

However, now I can see a case where the delta can be bigger than `slice_ns`, but that's a side effect of the fact that we don't clean entries synchronously from the `self.task_info` HashMap when the corresponding tasks exit.

If a pid is reused we would (incorrectly) get the previous `TaskInfo` entry with the `sum_exec_runtime` of the previous task.

Do we have a callback that we can use in `sched_ext_ops` to detect when a task exits in a synchronous way? I was trying to use `.disable`, but it seems to be called only when the scheduler exits, not when the task exits... is this the expected behavior? Am I missing something?

htejun: `.disable` should be called when a task gets freed, which may take place a bit after the task exits.

arighi: > `.disable` should be called when a task gets freed, which may take place a bit after the task exits.

ok, I'll try again with the `.disable` callback to notify the user-space sched when tasks are exiting, it's more elegant than running the periodic garbage collector.

(Maybe we still need to do the garbage collector in case we fail to notify the user-space scheduler, but I don't think it's a big problem for now).

Thanks!

PR #63, opened 2024-01-02
Userland scheduler updates

Decave:
This patch fixes a couple issues in the ~`scx_rustland` and~ `scx_userland` scheduler:

- ~In `scx_rustland`, we keep using builtin idle tracking with `SCX_OPS_KEEP_BUILTIN_IDLE` to avoid racy checks in `ops.select_cpu()`~
- In `scx_userland`, we avoid hangs by using more robust checks to see if there's more work to be done in user space, and sending resched IPIs whenever there is.

### DIFF
--- a/scheds/c/scx_userland.bpf.c
+++ b/scheds/c/scx_userland.bpf.c
@@ -39,13 +39,24 @@ const volatile u32 num_possible_cpus = 64;
 /* Stats that are printed by user space. */
 u64 nr_failed_enqueues, nr_kernel_enqueues, nr_user_enqueues;
 
-struct user_exit_info uei;
+/*
+ * Number of tasks that are queued for scheduling.
+ *
+ * This number is incremented by the BPF component when a task is queued to the
+ * user-space scheduler and it must be decremented by the user-space scheduler
+ * when a task is consumed.
+ */
+volatile u64 nr_queued;
 
 /*
- * Whether the user space scheduler needs to be scheduled due to a task being
- * enqueued in user space.
+ * Number of tasks that are waiting for scheduling.
+ *
+ * This number must be updated by the user-space scheduler to keep track if
+ * there is still some scheduling work to do.
  */
-static bool usersched_needed;
+volatile u64 nr_scheduled;
+
+struct user_exit_info uei;
 
 /*
  * The map containing tasks that are enqueued in user space from the kernel.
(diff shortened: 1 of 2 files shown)


arighi (approved):
Reviewed and tested, LGTM!

The fact that we start to have some duplicate code between the BPF part of scx_rustland and scx_userland makes me think if we should introduce at some point a common "BPF abstraction" to provide the basic APIs to implement the user-space schedulers... something to think about for the future.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -292,7 +292,7 @@ s32 BPF_STRUCT_OPS(rustland_select_cpu, struct task_struct *p, s32 prev_cpu,
 	 * dispatch directly from .enqueue(), bypassing the user-space
 	 * scheduler.
 	 */
-	tctx->force_local = get_cpu_owner(prev_cpu) == 0;
+	tctx->force_local = scx_bpf_test_and_clear_cpu_idle(prev_cpu);
```
htejun, line 295: Isn't this still racy because the rust userland component doesn't participate? `select_cpu` may decide to put a task on `prev_cpu` according to builtin idle tracking but before that becomes visible to userland, it may also decide to assign another task to the same CPU. Getting rid of the race would require implementing atomic op across BPF and userland, no?

Decave: Yeah, it's still racy given that user space can dispatch a task to that core. My thinking is that this makes it slightly less racy by getting rid of the race between context switching, but perhaps we should just keep it the way it is as it's inevitable that we'll need to implement some kind of idle tracking between user space and the kernel anyways?

arighi: > Yeah, it's still racy given that user space can dispatch a task to that core. My thinking is that this makes it slightly less racy by getting rid of the race between context switching, but perhaps we should just keep it the way it is as it's inevitable that we'll need to implement some kind of idle tracking between user space and the kernel anyways?

Thinking more about this, given that both solutions are racy in a way or another, I think I'm more inclined to maintain the current setup and (try to) rely as much as possible on the cpu ownership map. Just because it easily gives the possibility to the user-space scheduler to reserve some CPUs to certain PIDs (like creating potential reservation mechanism to lock certain CPUs to certain PIDs, or implement similar ideas). We don't have to necessarily do that, but having the possibility to do that would be nice, also considering that this scheduler was designed as a foundational framework for implementing more complex scheduling policies in Rust.

htejun: It *should* be possible to make it not racy by using atomic ops from both bpf and rust. Making data types matching might be tricky tho.

Decave: That all makes sense. I dropped the change to scx_rustland.

Decave: > It _should_ be possible to make it not racy by using atomic ops from both bpf and rust. Making data types matching might be tricky tho.

Yeah I think it's possible. Unfortunately we're back to the classic impedance problem of cpumask kptrs in the kernel vs. how they're represented in user space

PR #68, opened 2024-01-04
scx_rustland: refactoring

arighi:
Minor improvement:
 - scx_rustland: make dispatcher more robust
 - scx_rustland: remove unnecessary scx_bpf_dispatch_nr_slots() check
 - scx_rustland: introduce nr_failed_dispatches

Code refactoring:
 - scx_rustland: provide an abstraction layer for the BPF component
 - scx_rustland: update comments and documentation in the BPF part

There is basically no functional change in this pull request, just some code refactoring. The important part is the new abstraction to the BPF component that makes the scheduler code much more readable and understandable.

Maybe in the future we could also consider to move this BPF abstraction to a more generic place, with a proper API documentation, to give the possibility to implement scx user-space schedulers in a more high-level abstracted way.

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -484,9 +484,6 @@ void BPF_STRUCT_OPS(rustland_dispatch, s32 cpu, struct task_struct *prev)
 		struct task_struct *p;
 		struct dispatched_task_ctx task;
 
-		if (!scx_bpf_dispatch_nr_slots())
-			break;
-
 		/*
 		 * Pop first task from the dispatched queue, stop if dispatch
 		 * queue is empty.
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -73,8 +73,11 @@ volatile u64 nr_queued;
  */
 volatile u64 nr_scheduled;
 
-/* Misc statistics */
-volatile u64 nr_user_dispatches, nr_kernel_dispatches, nr_sched_congested;
+/* Dispatch statistics */
+volatile u64 nr_user_dispatches, nr_kernel_dispatches;
+
+/* Failure statistics */
+volatile u64 nr_failed_dispatches, nr_sched_congested;
 
  /* Report additional debugging information */
 const volatile bool debug;
@@ -291,7 +294,18 @@ static void dispatch_task(struct task_struct *p, s32 cpu, u64 enq_flags)
 	if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
 		cpu = scx_bpf_task_cpu(p);
 		if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
-			dbg_msg("%s: pid=%d global", __func__, p->pid);
+			/*
+			 * Both the designated CPU and the previously used CPU
+			 * are unavailable, we need to fallback to the global DSQ.
+			 *
+			 * This is not necessarily a problem, using the global
+			 * DSQ will give a small performance penalty to the
+			 * task (because it needs to wait for the local DSQ to
+			 * be drained). So for now simply report the event as a
+			 * "dispatch failure" and keep going.
+			 */
+			dbg_msg("%s: pid=%d (fail)", __func__, p->pid);
+			__sync_fetch_and_add(&nr_failed_dispatches, 1);
 			scx_bpf_dispatch(p, SCX_DSQ_GLOBAL, slice_ns,
 					 enq_flags);
 			return;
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -276,16 +280,23 @@ static void dispatch_local(struct task_struct *p, u64 enq_flags)
...
-	 * If it's not possible to dispatch on the selected CPU, re-use the
-	 * previously used one.
-	 */
-	if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr))
+	if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
 		cpu = scx_bpf_task_cpu(p);
+		if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
+			dbg_msg("%s: pid=%d global", __func__, p->pid);
+			scx_bpf_dispatch(p, SCX_DSQ_GLOBAL, slice_ns,
+					 enq_flags);
```
Decave, line 296: I wonder if this should eventually be an error condition in the scheduler? Unfortunately we don't yet have an ergonomic API for looking at a task's cpumask in user space so it might be a bit premature.

Also, on the other hand, maybe it would make the scheduler unnecessarily brittle in that a task could exit or something while it's being dispatched from user space. So this fallback is probably fine, but maybe we should add a comment explaining why this could happen?

scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -596,6 +624,12 @@ void BPF_STRUCT_OPS(rustland_disable, struct task_struct *p)
 
 /*
  * Heartbeat scheduler timer callback.
+ *
+ * If the system is completely idle the sched-ext wathcdog may incorrectly
```
Decave, line 628: watchdog

Decave (approved):
Looks great, thanks! I left a couple of nits, but this LGTM. Please feel free to merge once they're addressed.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: > I wonder if this should eventually be an error condition in the scheduler? Unfortunately we don't yet have an ergonomic API for looking at a task's cpumask in user space so it might be a bit premature.

Oh yes! Having an API to access cpumask would be super useful... we could use that in the scheduler to avoid selecting dispatching tasks on CPUs that can't be used.

Right now we could report that as a warning maybe? I don't think we should error the scheduler, it doesn't seem to be a too critical condition, considering that in the worst case scenario we can simply use the global DSQ and everything will work just fine.

Moreover even if we have access to the cpumask in the future we may still do something similar, unless we also provide some mechanisms to lock the cpumask.

> 
> Also, on the other hand, maybe it would make the scheduler unnecessarily brittle in that a task could exit or something while it's being dispatched from user space. So this fallback is probably fine, but maybe we should add a comment explaining why this could happen?

Absolutely, I'll add a comment. Maybe even introduce an extra counter, like `nr_dispatch_failures` or something similar?

Decave: >Right now we could report that as a warning maybe? I don't think we should error the scheduler, it doesn't seem to be a too critical condition, considering that in the worst case scenario we can simply use the global DSQ and everything will work just fine.

Yeah, agreed. And I don't think it necessarily indicates a bad / corrupted scheduler or anything anyways given that everything is async, so this is inherently a bit racy.

>Absolutely, I'll add a comment. Maybe even introduce an extra counter, like nr_dispatch_failures or something similar?

SGTM!

scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -276,16 +283,34 @@ static void dispatch_local(struct task_struct *p, u64 enq_flags)
...
-	/*
-	 * If it's not possible to dispatch on the selected CPU, re-use the
-	 * previously used one.
-	 */
-	if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr))
+	if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
 		cpu = scx_bpf_task_cpu(p);
+		if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
+			/*
+			 * Both the designaged CPU and the previously used CPU
```
Decave, line 298: designated

Decave (approved):
LGTM! Just one small typo

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: ...and fixed! I should really run a syntax checker. And something that checks when I confuse scx_rustland with scx_userland as well. :rofl: 

Thanks!

PR #72, opened 2024-01-06
scx_rustland: enhancements

arighi:
Some fixes and enhancements to make scx_rustland more usable on a desktop / interactive system.

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -410,6 +410,14 @@ static bool is_task_cpu_available(struct task_struct *p, u64 enq_flags)
 		return true;
 
 	/*
+	 * Do not bypass the user-space scheduler if there are pending
+	 * activity, otherwise, we may risk disrupting the scheduler's
+	 * decisions and negatively affecting the overall system performance.
+	 */
+	if (is_usersched_needed())
+		return false;
+
+	/*
 	 * For regular tasks always rely on force_local to determine if we can
 	 * bypass the scheduler.
 	 */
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -282,6 +285,42 @@ static void dispatch_local(struct task_struct *p, u64 enq_flags)
...
+{
+	int i;
+
+	for (i = 0; i < num_possible_cpus; i++) {
+		cpu = (cpu + i) % num_possible_cpus;
+		if (bpf_cpumask_test_cpu(cpu, p->cpus_ptr))
+			return cpu;
+	}
+
+	return -1;
```
Decave, line 302: Something else you could do is call `bpf_cpumask_any_distribute(p->cpus_ptr)` to choose a random CPU from `p->cpus_ptr`? Note that it returns a cpu >= num_cpus if none of the bits are set, so you'd have to do something like:
```
cpu = bpf_cpumask_any_distribute(p->cpus_ptr);
return cpu < num_possible_cpus ? cpu : -1;
```

scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -375,6 +411,14 @@ static bool is_task_cpu_available(struct task_struct *p, u64 enq_flags)
 	if (is_kthread(p) && p->nr_cpus_allowed == 1)
 		return true;
 
+	/*
+	 * Do not bypass the user-space scheduler if there are pending
+	 * activity, otherwise, we may risk to disrupt the scheduler's
```
Decave, line 416: optional suggestion:

s/we may risk to disrupt/we may risk disrupting

Decave (commented):
Overall looks good to me. Left one optional nit, and one suggestion for how we can make `_get_task_cpu_slow()` a lot faster.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: Nice suggestion! Thanks! I'll apply it and run some tests. I think we should still maintain the logic to try the prev_cpu before picking a random one using `bpf_cpumask_any_distribute()`, that should reduce some unnecessary migrations, but I'll test/measure both and check if one shows some clear benefits vs the other.

arighi:
Moreover, I'm not yet fully convinced about the last 2 commits (change the default timeout_ms and slice_ns).

While they can definitely provide some benefits for interactive workloads I think there's also value to keep them as they are, because that can help to better measure future improvements of the scheduler.

Maybe I can add a command line option that would set them if a certain command line option is passed...

arighi:
Alright, I dropped some commits and updated this PR only with the "safe" fixes.

For the more experimental ones I'd like to do more testing (even if the other changes make the scheduler more responsive, they also make it more unpredictable sometimes, so I would like to investigate more on that).

scheds/rust/scx_rustland/src/bpf/main.bpf.c
Decave: Yep, agreed. I was thinking that `bpf_cpumask_any_distribute()` could replace that for loop. But trying to choose prev_cpu first seems like a good idea.

htejun (approved):

PR #79, opened 2024-01-09
[scx_rusty] select_cpu to kick lower pri tasks

jordalgo:
If we get to the end of select_cpu:
if no idle cpu, waking, and no other tasks are in the domain queue then see if any cpu in the domain is running a task with lower priority. Pick the one with the lowest priority, whose runtime is also lower than the task we want to select the cpu for, and kick it.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/intf.h
+++ b/scheds/rust/scx_rusty/src/bpf/intf.h
@@ -17,6 +17,7 @@
 typedef unsigned char u8;
 typedef unsigned int u32;
 typedef unsigned long long u64;
+typedef int s32;
 #endif
 
 #include <scx/ravg.bpf.h>
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -67,6 +67,7 @@ const volatile u32 load_half_life = 1000000000	/* 1s */;
 const volatile bool kthreads_local;
 const volatile bool fifo_sched;
 const volatile bool switch_partial;
+const volatile bool enable_preemption;
 const volatile u32 greedy_threshold;
 const volatile u32 debug;
 
@@ -78,9 +79,10 @@ const volatile u64 slice_ns = SCX_SLICE_DFL;
  */
 struct pcpu_ctx {
 	u32 dom_rr_cur; /* used when scanning other doms */
+	u64 task_vtime; /* vtime of the task currently on the cpu*/
 
 	/* libbpf-rs does not respect the alignment, so pad out the struct explicitly */
-	u8 _padding[CACHELINE_SIZE - sizeof(u32)];
+	u8 _padding[CACHELINE_SIZE - (sizeof(u32) + sizeof(u64))];
 } __attribute__((aligned(CACHELINE_SIZE)));
 
 struct pcpu_ctx pcpu_ctx[MAX_CPUS];
@@ -116,6 +118,16 @@ struct dom_active_pids {
 };
 
 struct dom_active_pids dom_active_pids[MAX_DOMS];
+
+/*
+ * The CPUs associated with a domain
+ */
+struct dom_cpus {
+	s32 cpus[MAX_CPUS];
+	u64 nr_cpus;
+};
+
+struct dom_cpus dom_cpus[MAX_DOMS];
 
 const u64 ravg_1 = 1 << RAVG_FRAC_BITS;
 
@@ -345,6 +357,7 @@ u64 tune_params_gen;
 private(A) struct bpf_cpumask __kptr *all_cpumask;
 private(A) struct bpf_cpumask __kptr *direct_greedy_cpumask;
 private(A) struct bpf_cpumask __kptr *kick_greedy_cpumask;
+private(A) struct bpf_cpumask __kptr *preempted_cpumask;
 
 static inline bool vtime_before(u64 a, u64 b)
 {
@@ -658,6 +671,71 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 	}
 
 	/*
+	 * This checks if there are any CPUs in the task's domain that can be
+	 * preempted i.e. find the highest vtime in the domain that's higher
+	 * (minus a small buffer: slice_ns) than the current task's vtime.
+	 . This is to increase responsiveness for higher priority tasks. The task
+	 * might also be forked/execd so check for wake up only.
+	 */
+	if (enable_preemption && (wake_flags & SCX_WAKE_TTWU)) {
+		u32 dom_id = taskc->dom_id;
+		struct bpf_cpumask *pre_cpumask;
+		struct dom_ctx *domc;
+		struct dom_cpus *domcpus;
+		struct pcpu_ctx *pcpuc;
+		u64 max_vtime;
+		s32 i, cpu_i, found_cpu;
+
+		if (!(domc = bpf_map_lookup_elem(&dom_data, &dom_id))) {
+			scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+			goto enoent;
+		}
+
+		if (!(pre_cpumask = preempted_cpumask)) {
+			scx_bpf_error("No preempted_cpumask found");
+			goto enoent;
+		}
+
+		domcpus = MEMBER_VPTR(dom_cpus, [dom_id]);
+		if (!domcpus)
+		{
+			scx_bpf_error("Failed to lookup dom cpus[%u]", dom_id);
+			goto enoent;
+		}
+
+		found_cpu = -1;
+		max_vtime = p->scx.dsq_vtime;
+
+		bpf_for(i, 0, domcpus->nr_cpus) {
+			if (i >= MAX_CPUS)
+				continue;
+
+			cpu_i = domcpus->cpus[i];
+
+			pcpuc = MEMBER_VPTR(pcpu_ctx, [cpu_i]);
+
+			if (!pcpuc) {
+				scx_bpf_error("No pcpu_ctx");
+				goto enoent;
+			}
+
+			if (bpf_cpumask_test_cpu(cpu_i, p->cpus_ptr)
+			&& !bpf_cpumask_test_cpu(cpu_i, pre_cpumask)
+			&& vtime_before(max_vtime, pcpuc->task_vtime)) {
+				found_cpu = cpu_i;
+				max_vtime = pcpuc->task_vtime;
+			}
+		}
+		if (found_cpu > -1) {
+			cpu = found_cpu;
+			bpf_cpumask_set_cpu(cpu, pre_cpumask);
+			scx_bpf_put_idle_cpumask(idle_smtmask);
+			scx_bpf_dispatch(p, SCX_DSQ_LOCAL, slice_ns, SCX_ENQ_PREEMPT | SCX_ENQ_HEAD);
+			return cpu;
+		}
+	}
+
+	/*
 	 * We're going to queue on the domestic domain's DSQ. @prev_cpu may be
 	 * in a different domain. Returning an out-of-domain CPU can lead to
 	 * stalls as all in-domain CPUs may be idle by the time @p gets
@@ -896,6 +974,12 @@ void BPF_STRUCT_OPS(rusty_running, struct task_struct *p)
 		return;
 	}
 
+	s32 cpu = scx_bpf_task_cpu(p);
+	struct pcpu_ctx *pcpuc = MEMBER_VPTR(pcpu_ctx, [cpu]);
+
+	if (pcpuc)
+		pcpuc->task_vtime = p->scx.dsq_vtime - slice_ns;
+
 	/*
 	 * Global vtime always progresses forward as tasks start executing. The
 	 * test and update can be performed concurrently from multiple CPUs and
@@ -904,14 +988,25 @@ void BPF_STRUCT_OPS(rusty_running, struct task_struct *p)
 	 */
 	if (vtime_before(domc->vtime_now, p->scx.dsq_vtime))
 		domc->vtime_now = p->scx.dsq_vtime;
+
+	if (preempted_cpumask)
+		bpf_cpumask_clear_cpu(cpu, preempted_cpumask);
 }
 
 void BPF_STRUCT_OPS(rusty_stopping, struct task_struct *p, bool runnable)
 {
 	struct task_ctx *taskc;
+	struct pcpu_ctx *pcpuc;
+	s32 cpu;
 
 	if (fifo_sched)
 		return;
+
+	cpu = scx_bpf_task_cpu(p);
+	pcpuc = MEMBER_VPTR(pcpu_ctx, [cpu]);
+
+	if (pcpuc)
+		pcpuc->task_vtime = 0;
 
 	if (!(taskc = lookup_task_ctx(p)))
 		return;
@@ -1156,7 +1251,9 @@ static s32 create_dom(u32 dom_id)
 s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init)
 {
 	struct bpf_cpumask *cpumask;
+	struct dom_cpus *domcpus;
 	s32 i, ret;
+	u32 dom_id;
 
 	cpumask = bpf_cpumask_create();
 	if (!cpumask)
@@ -1179,6 +1276,15 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init)
 	if (cpumask)
 		bpf_cpumask_release(cpumask);
 
+	if (enable_preemption) {
+		cpumask = bpf_cpumask_create();
+		if (!cpumask)
+			return -ENOMEM;
+		cpumask = bpf_kptr_xchg(&preempted_cpumask, cpumask);
+		if (cpumask)
+			bpf_cpumask_release(cpumask);
+	}
+
 	if (!switch_partial)
 		scx_bpf_switch_all();
 
@@ -1188,8 +1294,13 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init)
 			return ret;
 	}
 
-	bpf_for(i, 0, nr_cpus)
+	bpf_for(i, 0, nr_cpus) {
 		pcpu_ctx[i].dom_rr_cur = i;
+		dom_id = cpu_to_dom_id(i);
+		domcpus = MEMBER_VPTR(dom_cpus, [dom_id]);
+		if (domcpus && domcpus->nr_cpus < MAX_CPUS)
+			domcpus->cpus[domcpus->nr_cpus++] = i;
+	}
 
 	return 0;
 }
--- a/scheds/rust/scx_rusty/src/main.rs
+++ b/scheds/rust/scx_rusty/src/main.rs
@@ -137,6 +137,11 @@ struct Opts {
     #[clap(short = 'f', long, action = clap::ArgAction::SetTrue)]
     fifo_sched: bool,
 
+    /// Disable preemption to prioritize lower vtime waking tasks.
+    /// fifo_sched needs to be set to false for preemption to work.
+    #[clap(long, action = clap::ArgAction::SetTrue)]
+    no_preemption: bool,
+
     /// Idle CPUs with utilization lower than this will get remote tasks
     /// directly pushed on them. 0 disables, 100 enables always.
     #[clap(short = 'D', long, default_value = "90.0")]
@@ -1109,6 +1114,7 @@ impl<'a> Scheduler<'a> {
         skel.rodata_mut().kthreads_local = opts.kthreads_local;
         skel.rodata_mut().fifo_sched = opts.fifo_sched;
         skel.rodata_mut().switch_partial = opts.partial;
+        skel.rodata_mut().enable_preemption = !(opts.no_preemption || opts.fifo_sched);
         skel.rodata_mut().greedy_threshold = opts.greedy_threshold;
         skel.rodata_mut().debug = opts.verbose as u32;
 


scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -605,6 +613,49 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 			}
 		}
 	}
+	
+	// the task might also be forked/execd so check for wake up only
+	if (wake_flags & SCX_WAKE_TTWU) {
+		cpu = scx_bpf_pick_idle_cpu((const struct cpumask *)p_cpumask, 0);
```
htejun, line 619: Do we need to this again given that we're here iff the previous attempts have failed?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -605,6 +613,49 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+		if (cpu >= 0) {
+			goto direct;
+		}
+		if (scx_bpf_dsq_nr_queued(taskc->dom_id) == 0) {
```
htejun, line 623: Hmm... what if the task being queued is higher priority than others which are already queued?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -605,6 +613,49 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+			s32 lowestWeight = -1;
+			s32 foundCPU = -1;
+			u32 currentWeight = taskc->weight;
+			u64 currentSumRunTime = p->se.sum_exec_runtime;
```
htejun, line 627: It'd probably be better to follow kernel naming conventions.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -605,6 +613,49 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+			struct cpu_task_data *ctd;
+			// Look for the currently running task with the lowest weight
+			// and with a weight lower than the current task
+			// and with a higher sum_exec_runtime than the current task
```
htejun, line 631: Comment style too.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -605,6 +613,49 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+				ctd = MEMBER_VPTR(cpu_task_data, [cpu]);
+				if (!ctd) {
+					continue;
+				}
+				if (!ctd->has_running_task) {
+					// No running task, let's just use this cpu
+					foundCPU = cpu;
+					break;
+				} else if (ctd->weight < currentWeight &&
+					currentSumRunTime < ctd->sum_exec_runtime &&
```
htejun, line 645: I'm not sure this makes sense. Why are we comparing the tasks' accumulated runtimes? Shouldn't we be comparing vtimes instead?

htejun (changes requested):
Left some comments. How did you test this? Did you see any improvements in terms of interactivity with your changes? Would it make sense to make the behavior optional?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: I still see this code path getting hit in my bpf_trace_printk output

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: That is for my TODO below.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: @Decave Mentioned something about fairness e.g. this is meant to protect a lower priority task always getting kicked off the cpu by a higher priority task. vtime is a combination of accumulated runtimes and weight right? So should I just be using that instead of these separately?

jordalgo:
> How did you test this? 

On my old 4 core (8 cpu) single socket Lenovo Yoga :) I also compiled the kernel in the background: `nice -n 19 make -j 64`. It was laggy with and without this change when using scx_rusty.

> Did you see any improvements in terms of interactivity with your changes? 

Not really, but I need to run a proper benchmark.

> Would it make sense to make the behavior optional?

Sure. That makes sense.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: It's something inherently racy but that doesn't mean we need to keep checking the same condition multiple times. If control reaches this point, it should be safe to consider that we should preempt if we can.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: So, if you track the vtime of the current task, this likely will be solved at the same time.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: Total accumulated execution time doesn't have much to do with fairness unless you're comparing tasks which are started at the same time across their entire lifetimes. vtime is what signifies their current relative priority.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: Not sure I understand how it solves both. Don't I still need to maintain a list of queued tasks for each CPU so I can do the vtime comparison?

htejun: You know that the task which is currently executing had the earliest vtime when it started executing and can also calculate the current vtime of the task from the execution time delta and its weight. Now, this doesn't guraantee that the next task in queue is after the current vtime as we can overrun that within slice. We can probably add a peeking operation to dsq to make it water tight but for now just comparing against the current executing task's vtime should do. Note that we can be more conservative to avoid being wrong - ie. preempt iff the preemptor's vtime is before the preemptee's vtime at the time it started execution.

jordalgo:
Ran a small benchmark on my lenovo. Looks like a small improvement with this change.
```
for i in {1..100}; do sudo ./src/jordalgo/rt-tests/hackbench -s 512 -l 200 -g 15 -f 25 -P | grep "Time:"; done
```

**Results:**
branch: main
avg 5.15162
median 5.1065
p99 6.04676

branch: rusty-update
avg 5.08463
median 5.0975
p99 5.18316

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: So I think the latest updates reflect this suggestion (particularly the last sentence where we look at (store) the vtime of the task when it starts running.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -605,6 +612,41 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
 	}
+	
+	/*
+	 * The task might also be forked/execd so check for wake up only
+	 */
+	if (wake_flags & SCX_WAKE_TTWU) {
+		u64 highest_vtime = 0;
+		s32 found_cpu = -1;
+		u64 current_vtime = p->scx.dsq_vtime;
+		struct cpu_task_data *ctd;
```
Decave, line 623: nit: Could you please add a newline between the variable declarations and the actual code?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -605,6 +612,41 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+				highest_vtime = ctd->dsq_vtime;
+				found_cpu = cpu;
+			}
+		}
+		if (found_cpu >= 0) {
+			cpu = found_cpu;
+			scx_bpf_kick_cpu(found_cpu, SCX_KICK_PREEMPT);
+			goto direct;
+		}
+	}
```
Decave, line 649: It would be nice if we could somehow avoid having to do this O(n) iteration -- this won't work well for even moderately large systems. A couple of thoughts come to mind:

1. Maybe we could just have a max vruntime global in the domc, along with the cpu that has the task executing it? It would be racy so maybe not super ideal, though we could also use a spinlock. it's too bad vruntime isn't a u32 or we could have put that and the cpu into a single u64 and used CAS to update them both atomically, though that's also not without cost.
2. Perhaps we could have a BPF rbtree in the dom that was sorted according to max vruntime of the tasks currently running on a core? That way the lookup would just be O(log(n))

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -860,10 +913,19 @@ void BPF_STRUCT_OPS(rusty_stopping, struct task_struct *p, bool runnable)
...
+		
+	s32 cpu = scx_bpf_task_cpu(p);
+	struct cpu_task_data *ctd;
 
-	if (!(taskc = lookup_task_ctx(p)))
+	ctd = MEMBER_VPTR(cpu_task_data, [cpu]);
+	if (ctd) {
+		ctd->has_running_task = false;
+	}
+
```
Decave, line 926: Extra newline

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -860,10 +913,19 @@ void BPF_STRUCT_OPS(rusty_stopping, struct task_struct *p, bool runnable)
 {
 	struct task_ctx *taskc;
 
-	if (fifo_sched)
+	if (!(taskc = lookup_task_ctx(p)))
 		return;
```
Decave, line 917: Can we `scx_ops_error()` here?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -860,10 +913,19 @@ void BPF_STRUCT_OPS(rusty_stopping, struct task_struct *p, bool runnable)
...
+		
+	s32 cpu = scx_bpf_task_cpu(p);
+	struct cpu_task_data *ctd;
 
-	if (!(taskc = lookup_task_ctx(p)))
+	ctd = MEMBER_VPTR(cpu_task_data, [cpu]);
+	if (ctd) {
+		ctd->has_running_task = false;
+	}
```
Decave, line 925: nit: In kernel coding style, single-line conditionals, loops, etc don't use curly braces

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -860,10 +913,19 @@ void BPF_STRUCT_OPS(rusty_stopping, struct task_struct *p, bool runnable)
...
+	if (!(taskc = lookup_task_ctx(p)))
 		return;
+		
+	s32 cpu = scx_bpf_task_cpu(p);
+	struct cpu_task_data *ctd;
 
-	if (!(taskc = lookup_task_ctx(p)))
+	ctd = MEMBER_VPTR(cpu_task_data, [cpu]);
+	if (ctd) {
+		ctd->has_running_task = false;
```
Decave, line 924: This lookup is a bit racy in that another task with low vruntime could be scheduled on the core shortly after, but I'm not sure if there's any way around that. In general it feels like it's mimicking the behavior of the idle cpumask already. We haven't been able to find an idle core in `rusty_select_cpu()` by the time we get to the preemption logic, so maybe we should just forget this and assume that every core has a task? We'd probably need to implement the peek operation as Tejun suggested.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -605,6 +612,41 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+		u64 current_vtime = p->scx.dsq_vtime;
+		struct cpu_task_data *ctd;
+		/*
+			* Look for the currently running task with the highest vtime
+			* and with a vtime higher than the current task
+			*/
+		bpf_for(cpu, 0, nr_cpus) {
+			if (cpu_to_dom_id(cpu) != taskc->dom_id) {
+				continue;
+			}
```
Decave, line 631: I think we also need to check that the cpu is in the task's `p->cpus_ptr`

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -807,6 +849,17 @@ void BPF_STRUCT_OPS(rusty_running, struct task_struct *p)
...
+		
+	s32 cpu = scx_bpf_task_cpu(p);
+	struct cpu_task_data *ctd;
+
+	ctd = MEMBER_VPTR(cpu_task_data, [cpu]);
+	if (ctd) {
+		ctd->has_running_task = true;
+		ctd->dsq_vtime = p->scx.dsq_vtime;
+	}
+
```
Decave, line 861: nit: extra newline

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: So, if we keep track of the max start vtime of a domain, we can make the decision to preempt or not quickly. Once we decide to preempt, we'll need to scan to either update the new maximum. If we can make the iteration quicker by making it iterate only the CPUs in each domain rather than iterating all CPUs, this should be fine in most cases given that each domain doesn't tend to be too big. I'd like to avoid keeping per-domain rbtree if possible. That likely is more expensive in most cases.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: If we base the preemption decision on the starting vtime - ie. a task can preempt a cpu iff the preempting task's vtime is prior to the preemptee's vtime when it started executing on the cpu, we'd be a bit more conservative with preemption but always correct and wouldn't need peeking, right?

htejun:
@jordalgo I'm unsure whether hackbench captures the purpose of this change (it's good that it's showing gains tho). A scenario which combines saturating low priority workload with higher priority latency sensitive workload would work better. e.g. Do high concurrency, nice -n 19 kernel build and measure how responsive the system is before and after.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
Decave: Yes indeed, that sounds better

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: > If we can make the iteration quicker by making it iterate only the CPUs in each domain rather than iterating all CPUs

Are the CPU ids contiguous within each domain? Or do we have to keep an array of CPU ids for each domain?

> Once we decide to preempt, we'll need to scan to either update the new maximum.

Shouldn't we only wait to do this until the task actually starts running on a CPU?

> Maybe we could just have a max vruntime global in the domc, along with the cpu that has the task executing it? It would be racy so maybe not super ideal, though we could also use a spinlock.

Correct me if I'm wrong but this part of the algo feels like best effort anyway so maybe it's ok if sometimes we don't preempt the maximum vtime task but get close enough and at least ensure that the vtime is less for the preemptor.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: It also seems like other places where we do this check in rusty, we don't log the erorr.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: If there was one coding style I wish we could throw in the garbage and set on fire.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: > > If we can make the iteration quicker by making it iterate only the CPUs in each domain rather than iterating all CPUs
> 
> Are the CPU ids contiguous within each domain? Or do we have to keep an array of CPU ids for each domain?

They aren't, so we'll have to introduce something to help iteration.

> > Once we decide to preempt, we'll need to scan to either update the new maximum.
> 
> Shouldn't we only wait to do this until the task actually starts running on a CPU?

Hmmm... I'm not sure. If we decide to preempt a CPU and on our way to send the preempting task to that CPU, wouldn't it make sense to consider that CPU to be occupied by the new task from then on? Otherwise, we'd be more prone to making multiple preemption decisions that target the same CPU.

> > Maybe we could just have a max vruntime global in the domc, along with the cpu that has the task executing it? It would be racy so maybe not super ideal, though we could also use a spinlock.
> 
> Correct me if I'm wrong but this part of the algo feels like best effort anyway so maybe it's ok if sometimes we don't preempt the maximum vtime task but get close enough and at least ensure that the vtime is less for the preemptor.

Yeah, while not ideal, that may be okay. Note that it can become a cascading problem in principle. If you kick out a task off CPU and that task has a vtime early enough to kick another CPU in the domain, we can be in an awkward situation.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: > We haven't been able to find an idle core in rusty_select_cpu() by the time we get to the preemption logic, so maybe we should just forget this and assume that every core has a task?

Updating the vtime for the CPU (in the `running` callback) will be racy as well right? Maybe you're saying that if we are even checking the vtime of the task that just went off the CPU, it's still probably better than not checking as the now outdated vtime might be a decent indication for the vtimes of the other tasks in the dsq.

jordalgo:
> Do high concurrency, nice -n 19 kernel build and measure how responsive the system is before and after.

I did this but the responsiveness was pretty poor before and after the change so it was hard to see if it was an improvement.

htejun:
> > Do high concurrency, nice -n 19 kernel build and measure how responsive the system is before and after.
> 
> I did this but the responsiveness was pretty poor before and after the change so it was hard to see if it was an improvement.

So, that's most likely indicating that preemption isn't working as intended.

jordalgo:
Ok, I took another crack at this based on all the conversations above and help from @Decave. I haven't yet tested this on my little-lenovo-that-could but wanted to get your thoughts on the overall strategy which is as follows:

- keep a max vtime in the domain data along with the cpu that is running this task with the highest vtime (note: there will be cases where this data will be stale if a task gets switched off and nothing else starts running on it but I think that's ok because that means, as far as we know, there is nothing on that CPU and we should select it).
- in `running` update the max vtime in the domain (as needed) and keep track of the starting vtime for that cpu in the global cpu_data array
- for the preemption in select_cpu:if the max vtime of the domain is greater than the task's current vtime, grab the associated cpu (as the one to kick/preempt) and then find the next highest vtime cpu in that domain and update the max vtime and associated cpu

htejun:
The basic approach looks sane to me. One thing to be awry of is that the code as-is can lead to excessive preemptions as the only requirement for preemption is the waking thread's vtime to be prior to the max vtime in the domain. People usually try to tame that a bit (e.g. apply the min vtime rule that's applied when actually queueing and then check whether the vtime delta from preemptee is big enough and so on). But the current code should be fine as the first step.

jordalgo:
I tested a bit and the preemption does seem a bit aggressive.

> (e.g. apply the min vtime rule that's applied when actually queueing and then check whether the vtime delta from preemptee is big enough and so on)

I'm not entirely sure if I understand - maybe something like this in select_cpu check? 
```
if (vtime_before(current_vtime, domc->vtime_max - slice_ns)) {
```

Or 
```
f (vtime_before(current_vtime, domc->vtime_max) && vtime_before(slice_ns, domc->vtime_now - domc->vtime_max)) {
```

htejun:
> > (e.g. apply the min vtime rule that's applied when actually queueing and then check whether the vtime delta from preemptee is big enough and so on)
> 
> I'm not entirely sure if I understand - maybe something like this in select_cpu check?
> 
> ```
> if (vtime_before(current_vtime, domc->vtime_max - slice_ns)) {
> ```

Yeah, something like this.

> ```
> if (vtime_before(current_vtime, domc->vtime_max) && vtime_before(slice_ns, domc->vtime_now - domc->vtime_max)) {
> ```

I don't understand the second term. What is it trying to do?

Another thing to consider is that when a task gets enqueued we wind-forward their vtime:

  https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_rusty/src/bpf/main.bpf.c#L700

So, the vtime range of the tasks currently enqueued or executing is pretty limited. This may lead to unexpected preemption decisions depending on the preemption criteria. Another thing to think about.

jordalgo:
Tangentially, even though weight is already baked into vtime, I wonder if we shouldn't also take that into account when considering if we should preempt a task e.g. even if this passes `if (vtime_before(current_vtime, domc->vtime_max - slice_ns)) {` maybe we should also compare the weights of the two tasks and only preempt if the preemptee's weight is lower. This might help preemption fighting if there are a lot of high pri tasks fighting to stay on the cpu in the same domain.

htejun:
Yeah, because it involves quite a bit of heuristics, it all comes down to what you wanna achieve. Whatever makes things work better and has good enough rationales should be fine.

jordalgo:
@htejun @Decave - I think this is ready for another look. After some testing I took a different approach that results in less preemption because it seemed, in some of the earlier drafts, that the algo was being a little too aggressive, which resulted in more jitter. 

I realized I could use vtime_now as the max vtime for the domain and instead of iterating over the cpus in a domain to find the "next" max vtime, I just let the system wait to re-set vtime_now with something larger as that seems to happen pretty often, even on my little 8 cpu machine -- this results in a much less complicated and racy implementation. Anyway, let me know what you think.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -89,6 +89,9 @@ struct pcpu_ctx {
 
 struct pcpu_ctx pcpu_ctx[MAX_CPUS];
 
+u64 current_cpu_vtime[MAX_CPUS];
```
htejun, line 92: Maybe add a comment explaining what it's supposed to contain? The `dom_ctx` field too.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -606,6 +609,31 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 		}
 	}
 
+	/*
+	 * The task might also be forked/execd so check for wake up only
```
htejun, line 613: It'd be useful to given an overview of what the block is doing and why.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -606,6 +609,31 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+	 */
+	if ((wake_flags & SCX_WAKE_TTWU) && !fifo_sched) {
```
htejun, line 615: And maybe make the behavior optional so that it's easier to compare.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -606,6 +609,31 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+
+		if (!(domc = bpf_map_lookup_elem(&dom_data, &dom_id))) {
+			scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+			goto enoent;
+		}
+
+		if (vtime_before(p->scx.dsq_vtime, domc->vtime_now - slice_ns)) {
+			cpu_vtime = MEMBER_VPTR(current_cpu_vtime, [domc->vtime_now_cpu]);
+
+			if (cpu_vtime && *cpu_vtime == domc->vtime_now && bpf_cpumask_test_cpu(domc->vtime_now_cpu, p->cpus_ptr)) {
```
htejun, line 628: What does `*cpu_time == domc->vtime_now` do?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -606,6 +609,31 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+				*cpu_vtime = 0;
```
htejun, line 629: And should there be a mechanism to update the hint for the next wakee?

htejun (commented):
This is looking better. I left some comments. Also, how does it actually behave?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: This "attempts" (though it's racy) to make sure we don't try to preempt the same CPU running the same task if we already did so (hence the setting of *cputime to 0 right below).

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: Sorry, don't understand what you mean :(

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: Hmm.. unless it's known this does happen to a noticeable degree, I'd rather not have it. Trying to reduce the size of race window whose size is not known can become confusing while not adding much in practice.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: When the next task wakes up and tries to see what it can preempt, how would that work?

jordalgo: The "next" task that wakes up won't be able to preempt anything until the domain's vtime_now is updated to a new, larger value. This is what makes this algo less aggressive with the preemption.

htejun: Hmm... that sounds like the resulting behavior would be rather erratic with multiple latency sensitive workloads.

jordalgo:
Ok, got another set of updates with some FPS benchmarking using (glxgears). In this iteration, after pre-empting the dom's vtime_max_cpu, select the next highest vtime_max and vtime_max_cpu (also do this if the task who had the highest vtime max stops running e.g. in rusty_stopping).

On my 8 cpu (4 core 2 threads/core) single socket laptop running glxgears (building kernel in the background `make -j 64`). The metrics below are in FPS:

### scx_rusty (trunk)
**Average:** 49.18566667
**Median:** 49.107
**p01:** 41.17767

### scx_rusty (previous iteration - without updating the vtime_max after preemption)
**Average:** 44.95922727
**Median:**: 45.3435
**p01:** 40.14978

### scx_rusty (this version)
**Average:** 54.59343478
**Median:**: 55.07
**p01:** 50.83758

### EEVDF
**Average:** 59.7923913
**Median:** 59.846
**p01:** 59.34174

jordalgo:
@htejun Testing with some logging (again on my laptop) it seems that the preemptor actually ends up on the CPU it kicks/preempts about 70% of the time.

jordalgo:
@Decave I made the changes we discussed e.g.
- protecting a cpu from being preempted if it's already being preempted
- dispatching to the local dsq right from select_cpu

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -904,18 +998,46 @@ void BPF_STRUCT_OPS(rusty_running, struct task_struct *p)
...
 
+	dom_id = taskc->dom_id;
+	domc = bpf_map_lookup_elem(&dom_data, &dom_id);
+	if (!domc) {
+		scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+		return;
+	}
+
+	if (domc->vtime_max == p->scx.dsq_vtime)
+		set_next_dom_vtime_max(domc, dom_id, cpu);
```
Decave, line 1039: I'm not sure I understand this conditional check. we're updating the vtime max for the domain if this task was previously the task with the highest vtime? is this just so that we only do it on a single core?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -462,6 +479,44 @@ static bool task_set_domain(struct task_ctx *taskc, struct task_struct *p,
...
+	domcpus = MEMBER_VPTR(dom_cpus, [dom_id]);
+	if (!domcpus)
+	{
+		scx_bpf_error("Failed to lookup dom cpus[%u]", dom_id);
+		return;
+	}
+
+	bpf_for(i, 0, domcpus->nr_cpus) {
+		if (i >= MAX_CPUS)
+			continue;
```
Decave, line 501: so this lets us avoid having to do a num cpu iteration and instead iterate # cpu in dom times instead?seems reasonable. can we add a comment explaining as much? once we have the cpumask kfunc iterator we can just use that instead and iterate over the domain's cpumask

Decave (commented):
Hi Jordan, this is looking great! Thanks a lot for working on this. Left a few more comments on this pass.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -606,6 +613,36 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 		}
 	}
 
+	/*
+	 * This checks if there are any CPUs in the task's domain that can be
+	 * preempted: e.g. if the task's vtime is less the highest vtime in the
```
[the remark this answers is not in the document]
jordalgo: Ugh, I'm so embarrassed by this mistake as one of my favorite scenes in Get Shorty is the discussion of this difference.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -657,6 +712,39 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+		if (domc->vtime_max != 0
+		&& vtime_before(p->scx.dsq_vtime, domc->vtime_max - slice_ns)
+		&& (preempted_cpumask && !bpf_cpumask_test_cpu(domc->vtime_max_cpu, preempted_cpumask))
+		&& bpf_cpumask_test_cpu(domc->vtime_max_cpu, p->cpus_ptr)) {
+			cpu = domc->vtime_max_cpu;
+			if (preempted_cpumask)
+				bpf_cpumask_set_cpu(domc->vtime_max_cpu, preempted_cpumask);
+
+			set_next_dom_vtime_max(domc, dom_id, cpu);
+			scx_bpf_kick_cpu(cpu, SCX_KICK_PREEMPT);
```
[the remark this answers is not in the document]
jordalgo: Does it need to be ORed with `SCX_ENQ_HEAD` ?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -896,6 +984,12 @@ void BPF_STRUCT_OPS(rusty_running, struct task_struct *p)
 		return;
 	}
 
+	s32 cpu = scx_bpf_task_cpu(p);
+	u64 *cpu_vtime = MEMBER_VPTR(cpu_vtimes, [cpu]);
+
+	if (cpu_vtime)
+		*cpu_vtime = p->scx.dsq_vtime;
```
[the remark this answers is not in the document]
jordalgo: Just curious, is there a benefit to using the pcpu_ctx vs a per cpu bpf map?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
jordalgo: This might be unnecessary, basically if the task is "stopping" then it's no longer the task with the highest vtime for the domain, so we need to pick the next highest so we know what can be preempted in this domain.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -904,18 +998,46 @@ void BPF_STRUCT_OPS(rusty_running, struct task_struct *p)
 	 */
 	if (vtime_before(domc->vtime_now, p->scx.dsq_vtime))
 		domc->vtime_now = p->scx.dsq_vtime;
+
```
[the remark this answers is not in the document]
jordalgo: Actually, there is a comment right above this that mentions that.

scheds/rust/scx_rusty/src/bpf/intf.h
```diff
@@ -17,6 +17,7 @@
 typedef unsigned char u8;
 typedef unsigned int u32;
 typedef unsigned long long u64;
+typedef int s32;
```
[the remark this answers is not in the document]
jordalgo: I don't think so. At least it wouldn't build when I included it in this header.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -67,6 +67,7 @@ const volatile u32 load_half_life = 1000000000	/* 1s */;
 const volatile bool kthreads_local;
 const volatile bool fifo_sched;
 const volatile bool switch_partial;
+const volatile bool no_preemption;
```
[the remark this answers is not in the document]
jordalgo: Yeah I was thinking the same

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -657,6 +712,39 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+	 */
+	if (!no_preemption && !fifo_sched && (wake_flags & SCX_WAKE_TTWU)) {
+		u32 dom_id = taskc->dom_id;
+		struct dom_ctx *domc;
+
+		if (!(domc = bpf_map_lookup_elem(&dom_data, &dom_id))) {
+			scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+			goto enoent;
+		}
+
```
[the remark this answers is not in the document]
jordalgo: I tried that at first but the verifier wasn't happy with me. I don't think it's that smart

scheds/rust/scx_rusty/src/bpf/main.bpf.c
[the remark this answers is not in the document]
Decave: It does not, `SCX_ENQ_PREEMPT` implies `SCX_ENQ_HEAD`. For posterity, can't hurt to include both though.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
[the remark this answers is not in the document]
Decave: If you use a per-cpu bpf map, you can only look up the entry on your current cpu. Using `pcpu_ctx` lets you look up entries on other cores as well

scheds/rust/scx_rusty/src/bpf/main.bpf.c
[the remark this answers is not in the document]
Decave: Hmm, sorry, which comment is that?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
Decave: Ah, I see. So, stopping doesn't necessarily imply no longer runnable. It just means that the core is temporarily not running the task anymore. You'll also want to check `!runnable` here, and I wonder if it might be slightly less confusing to do `domc->vtime_max_cpu == cpu` (not sure it matters though)?

I'm not sure this is worth doing though. It seems potentially undesirable to do a # CPU for loop every time a task stops running on the previously-max-vtime cpu. Consider that this is not going to be an uncommon thing -- it should happen roughly every 1/#CPUs in domain rescheds. A few thoughts:

1. This seems like something that's probably better to start doing only when util >= some threshold, so we don't tax the under-utilized case unnecessarily
2. It doesn't have to be in the first version of this code that lands, but I wonder if we could somehow use an in-BPF rbtree to do this? that way it would be an O(log(# domain CPUs)) operation to insert and remove running tasks, so we wouldn't have to do this O(n) lookup and "rebalance". The only drawback here is we'd be doing tree operations on (almost) every single running / stopping call. hmm, not sure. something to think about

scheds/rust/scx_rusty/src/bpf/main.bpf.c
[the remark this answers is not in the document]
Decave: hmmm, that should work. I've done that in a number of other places. this really doesn't work?

```
if (!preempted_cpumask) {
    scx_bpf_error("No cpumask found");
    goto enoent;
}
```

could you please share the verifier output?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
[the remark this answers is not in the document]
Decave: e.g. means ergo!!

scheds/rust/scx_rusty/src/bpf/main.bpf.c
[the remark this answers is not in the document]
jordalgo: Oh right. I forgot about that limitation.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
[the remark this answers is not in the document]
jordalgo: Starting on line 994:
```
/*
	 * Global vtime always progresses forward as tasks start executing. The
	 * test and update can be performed concurrently from multiple CPUs and
	 * thus racy. Any error should be contained and temporary. Let's just
	 * live with it.
	 */
```

scheds/rust/scx_rusty/src/bpf/main.bpf.c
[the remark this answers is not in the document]
jordalgo: So I believe it's because the use of the preempted_cpumask must be part of the next instruction or else it can't guarantee that it hasn't been modified in another function call e.g. this works:

```
if (!preempted_cpumask) {
  scx_bpf_error("No preempted_cpumask found");
   goto enoent;
}

if (!bpf_cpumask_test_cpu(domc->vtime_max_cpu, preempted_cpumask)
&& bpf_cpumask_test_cpu(domc->vtime_max_cpu, p->cpus_ptr)
...
```

But this does not:
```
```
if (!preempted_cpumask) {
  scx_bpf_error("No preempted_cpumask found");
   goto enoent;
}

if (bpf_cpumask_test_cpu(domc->vtime_max_cpu, p->cpus_ptr)
&& !bpf_cpumask_test_cpu(domc->vtime_max_cpu, preempted_cpumask)
...
```

Here is the end of the verifier log:
```
; && !bpf_cpumask_test_cpu(domc->vtime_max_cpu, preempted_cpumask)
515: (16) if w0 == 0x0 goto pc+40     ; R0=scalar(smin=smin32=0,smax=umax=smax32=umax32=1,var_off=(0x0; 0x1)) refs=1
; && !bpf_cpumask_test_cpu(domc->vtime_max_cpu, preempted_cpumask)
516: (18) r1 = 0xffff88810258e328     ; R1_w=map_value(off=24,ks=4,vs=32,imm=0) refs=1
518: (79) r2 = *(u64 *)(r1 +0)        ; R1_w=map_value(off=24,ks=4,vs=32,imm=0) R2_w=rcu_ptr_or_null_bpf_cpumask(id=29,off=0,imm=0) refs=1
; && !bpf_cpumask_test_cpu(domc->vtime_max_cpu, preempted_cpumask)
519: (61) r1 = *(u32 *)(r9 +16)       ; R1_w=scalar(smin=0,smax=umax=4294967295,var_off=(0x0; 0xffffffff)) R9=map_value(off=0,ks=4,vs=4048,imm=0) refs=1
; && !bpf_cpumask_test_cpu(domc->vtime_max_cpu, preempted_cpumask)
520: (85) call bpf_cpumask_test_cpu#132690
Possibly NULL pointer passed to trusted arg1
processed 401 insns (limit 1000000) max_states_per_insn 2 total_states 42 peak_states 42 mark_read 18
-- END PROG LOAD LOG --
```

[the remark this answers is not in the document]
Comment (Decave) [reply scheds/rust/scx_rusty/src/bpf/main.bpf.c:731]: ah, yes, you have to copy it into a local variable. but that's fine, because you know the global variable isn't actually going to change to another cpumask. it's just a way to get the verifier to not worry about the cpumask being possibly null due to being RCU protected. so something like this should work: https://github.com/sched-ext/scx/blob/main/scheds/c/scx_nest.bpf.c#L436-L441

File: scheds/rust/scx_rusty/src/bpf/main.bpf.c (hunk 24 above)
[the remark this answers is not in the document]
Comment (Decave) [reply scheds/rust/scx_rusty/src/bpf/main.bpf.c:1001]: hmm, I was thinking that we should add another comment explaining that this block was similarly racy as the one above, but after thinking about it more I think you're right that it's fine to just have the existing comment apply to both blocks.

File: scheds/rust/scx_rusty/src/bpf/main.bpf.c (hunk 27 above)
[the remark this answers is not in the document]
Comment (jordalgo) [reply scheds/rust/scx_rusty/src/bpf/main.bpf.c:731]: Ah right right. Sometimes my brain doesn't handle pointer copying correctly 😕

File: scheds/rust/scx_rusty/src/bpf/main.bpf.c (hunk 19 above)
Comment (jordalgo) [reply scheds/rust/scx_rusty/src/bpf/main.bpf.c:1039]: > So, stopping doesn't necessarily imply no longer runnable. It just means that the core is temporarily not running the task anymore.

Just for clarification - if the cpu is not running the task (with the highest vtime) then doesn't that mean, in this context, that we don't want kick this cpu in the preemption path in `select_cpu`?

> This seems like something that's probably better to start doing only when util >= some threshold, so we don't tax the under-utilized case unnecessarily

Sorry, what do you mean by "util" here?

File: scheds/rust/scx_rusty/src/bpf/main.bpf.c [hunk 28]
```diff
@@ -896,6 +933,9 @@ void BPF_STRUCT_OPS(rusty_running, struct task_struct *p)
 		return;
 	}
 
+	s32 cpu = scx_bpf_task_cpu(p);
+	u64 next_vtime_max = p->scx.dsq_vtime - slice_ns;
```
htejun, line 937: I suppose the `- slice_ns` part is to make preemption less aggressive? If so, it might be easier to think about if the adjustment is done when making the preemption decision rather than when remembering the current state.

jordalgo:
This is a slight variation on the strategy I mentioned above to always iterate over the dom cpus. It instead (as @htejun suggested), looks for the highest vtime on those cpus and pre-empts that one. It seemed to perform about the same (in my glxgears FPS test) as picking the first cpu found with a higher vtime and reduces potential bias toward cpus at the beginning of the dom_cpus array.

Additionally, I also ran other benchmarks (hackbench, perf bench) against these various iterations but didn't see significant change against scx_rusty in trunk.

htejun:
Closing for now.

PR #82, opened 2024-01-10
scx_flatcg: Fix fallout from direct dispatch API update

htejun:
flatcg was busy spinning on `try_pick_next_cgroup()` due to tasks w/ custom affinities being put on cgroup dsq's. Fix it and improve misbehavior visibility.


arighi (approved):
LGTM, and thanks for adding the fail and race counters!

PR #85, opened 2024-01-11
scx_rustland: voluntary context switch boost

arighi:
A significant improvement in scheduler responsiveness for low-latency application: in short, the idea is to take into account the amount of voluntary context switches to identify interactive tasks vs CPU-intensive background tasks, and use this criteria to reduce the accounted time slice of interactive tasks (implicitly boosting their priority).

This simple heuristic is much better than the previous approach that was de-prioritizing newly created tasks, as it allows to classify CPU-intensive workloads that are not spawning too many tasks vs the actual interactive low-latency applications. And it also allows avoids to de-prioritize interactive applications that need to fork short-lived tasks, e.g., interactive shell sessions, that receive a big responsiveness improvement with this change.

In conclusion:
 - I can now play Terraria at 60 fps even with a `stress-ng -c 32` running in background
 - I can run shell commands faster while the kernel is building in background `make -j 32` (faster than the default linux scheduler)
 - For my typical personal workload (reading emails, browsing internet, using git, vim, and running shell commands while I recompile kernels in the background) this scheduler now seems to perform nearly as well as the default Linux scheduler, and in some cases even better!

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf.rs
+++ b/scheds/rust/scx_rustland/src/bpf.rs
@@ -127,8 +127,9 @@ const SCHED_EXT: i32 = 7;
 pub struct QueuedTask {
     pub pid: i32,              // pid that uniquely identifies a task
     pub cpu: i32,              // CPU where the task is running (-1 = exiting)
-    pub sum_exec_runtime: u64, // Total cpu time */
-    pub weight: u64,           // Task static priority */
+    pub sum_exec_runtime: u64, // Total cpu time
+    pub nvcsw: u64,            // Voluntary context switches
+    pub weight: u64,           // Task static priority
 }
 
 // Task queued for dispatching to the BPF component (see bpf_intf::dispatched_task_ctx).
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_rustland/src/bpf.rs
```diff
@@ -127,8 +127,9 @@ const SCHED_EXT: i32 = 7;
 pub struct QueuedTask {
     pub pid: i32,              // pid that uniquely identifies a task
     pub cpu: i32,              // CPU where the task is running (-1 = exiting)
-    pub sum_exec_runtime: u64, // Total cpu time */
-    pub weight: u64,           // Task static priority */
+    pub sum_exec_runtime: u64, // Total cpu time
+    pub nvcsw: u64,            // Voluntary context switches
```
htejun, line 131: Another way to communicate this to the userspace scheduler would be sending how the task went off CPU along with the wakeup message and let userland figure out the more detailed statistics, which likely will make it easier to get more detailed understanding (e.g. runtime percentiles and whatnot).

htejun (approved):

scheds/rust/scx_rustland/src/bpf.rs
arighi: Ah that's a good idea. Maybe I can store that info in a BPF_MAP_TYPE_TASK_STORAGE and update some counters in `.running()` and `.stopping()`, then pass this info to the user-space, so it's also everything self-contained and I don't have to rely on any other kernel statistics for that...

PR #87, opened 2024-01-13
scx_userland: use a custom memory allocator to prevent page faults

arighi:
To prevent potential deadlock conditions under heavy loads, any scheduler that delegates scheduling decisions to user-space should avoid triggering page faults.

To address this issue, replace the default Rust allocator with a custom one (RustLandAllocator), designed to operate on a pre-allocated buffer.

This, coupled with the memory locking (via mlockall), prevents page faults from happening during the execution of the user-space scheduler, avoiding the deadlock condition.

This memory allocator is completely transparent to the user-space scheduler code and it is applied automatically when the bpf module is imported.

In the future we may decide to move this allocator to a more generic place (scx_utils crate), so that also other user-space Rust schedulers can use it.

This initial implementation of the RustLandAllocator is very simple: a basic block-based allocator that uses an array to track the status of each memory block (allocated or free).

This allocator can be improved in the future, but right now, despite its simplicity, it shows a reasonable speed and efficiency in meeting memory requests from the user-space scheduler, having to deal mostly with small and uniformly sized allocations.

With this change in place scx_rustland survived more than 10hrs on a heavily stressed system (with stress-ng and kernel builds running in a loop):

 $ ps -o pid,rss,etime,cmd -p `pidof scx_rustland`
     PID   RSS     ELAPSED CMD
   34966 75840    10:00:44 ./build/scheds/rust/scx_rustland/debug/scx_rustland

Without this change it is possible to trigger the sched-ext watchdog timeout in less than 5min, under the same system load conditions.

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/alloc.rs
+++ b/scheds/rust/scx_rustland/src/bpf/alloc.rs
@@ -48,6 +48,14 @@ static MEMORY: RustLandMemory = RustLandMemory {
 pub struct RustLandAllocator;
 
 impl RustLandAllocator {
+    unsafe fn block_to_addr(&self, block: usize) -> *mut u8 {
+        MEMORY.arena.get().cast::<u8>().add(block * BLOCK_SIZE)
+    }
+
+    unsafe fn is_aligned(&self, block: usize, align_size: usize) -> bool {
+        self.block_to_addr(block) as usize & (align_size - 1) == 0
+    }
+
     pub fn lock_memory(&self) {
         unsafe {
             // Call setrlimit to set the locked-in-memory limit to unlimited.
@@ -73,9 +81,6 @@ impl RustLandAllocator {
 unsafe impl GlobalAlloc for RustLandAllocator {
     unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
         let align = layout.align();
-        if align > BLOCK_SIZE {
-            panic!("unsupported alignment: {}", align);
-        }
         let size = layout.size();
 
         // Find the first sequence of free blocks that can accommodate the requested size.
@@ -83,15 +88,18 @@ unsafe impl GlobalAlloc for RustLandAllocator {
         let mut contiguous_blocks = 0;
         let mut start_block = None;
 
-        for (index, &is_allocated) in map_guard.iter().enumerate() {
-            if is_allocated {
-                // Reset consecutive blocks count if an allocated block is encountered.
+        for (block, &is_allocated) in map_guard.iter().enumerate() {
+            // Reset consecutive blocks count if an allocated block is encountered or if the
+            // first block is not aligned to the requested alignment.
+            if is_allocated
+                || (contiguous_blocks == 0 && !self.is_aligned(block, align))
+            {
                 contiguous_blocks = 0;
             } else {
                 contiguous_blocks += 1;
                 if contiguous_blocks * BLOCK_SIZE >= size {
                     // Found a sequence of free blocks that can accommodate the size.
-                    start_block = Some(index + 1 - contiguous_blocks);
+                    start_block = Some(block + 1 - contiguous_blocks);
                     break;
                 }
             }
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_rustland/src/bpf/alloc.rs
```diff
@@ -0,0 +1,134 @@
...
+            }
+        };
+    }
+}
+
+// Override global allocator methods.
+unsafe impl GlobalAlloc for RustLandAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let align = layout.align();
+        if align > BLOCK_SIZE {
```
htejun, line 76: >64 align should be really rare so this should be fine but larger alignments can be supported relatively easily by updating the scan to find the first block to skip to the next alignment position.

scheds/rust/scx_rustland/src/bpf/alloc.rs
```diff
@@ -0,0 +1,134 @@
...
+        // Find the first sequence of free blocks that can accommodate the requested size.
+        let mut map_guard: MutexGuard<[bool; NUM_BLOCKS]> = MEMORY.allocation_map.lock().unwrap();
+        let mut contiguous_blocks = 0;
+        let mut start_block = None;
+
+        for (index, &is_allocated) in map_guard.iter().enumerate() {
+            if is_allocated {
+                // Reset consecutive blocks count if an allocated block is encountered.
+                contiguous_blocks = 0;
+            } else {
```
htejun, line 90: ie. here, the code can just consider the block to be busy if `contiguous_blocks` is zero and the offset isn't aligned.

htejun (approved):
Looks good to me. It may not be relevant to `scx_rustland` right now but in a future where this allocator is used more widely, it may be useful to provide a per-thread override to use the default allocator which can expand the heap.

Having a fixed arena size may become a scalability pain point for schedulers which have per-task memory allocations. Arena size big enough to accommodate tail end use cases which can have a very high number of tasks can be too large for other cases. This can be worked around in a fairly straight-forward way by escaping to the heap-expanding allocator when allocating from `.init_task()`. `.init_task()` isn't in mem reclaim path, can block and be made to operate synchronously w.r.t. userspace. So, if we provide a separate channel for `.init_task()` and do all per-task allocations from there with the "don't allocate from the fixed arena` flag set, this problem can be resolved.

arighi:
> Looks good to me. It may not be relevant to `scx_rustland` right now but in a future where this allocator is used more widely, it may be useful to provide a per-thread override to use the default allocator which can expand the heap.
> 
> Having a fixed arena size may become a scalability pain point for schedulers which have per-task memory allocations. Arena size big enough to accommodate tail end use cases which can have a very high number of tasks can be too large for other cases. This can be worked around in a fairly straight-forward way by escaping to the heap-expanding allocator when allocating from `.init_task()`. `.init_task()` isn't in mem reclaim path, can block and be made to operate synchronously w.r.t. userspace. So, if we provide a separate channel for `.init_task()` and do all per-task allocations from there with the "don't allocate from the fixed arena` flag set, this problem can be resolved.

Hm... but the per-task allocation will be performed by the BPF part. Are you suggesting to allocate some memory from `.init_task()` and share that memory in the address space of the user-space scheduler?

scheds/rust/scx_rustland/src/bpf/alloc.rs
arighi: True. I completely ignored align because right now the scheduler doesn't request any alignment at all, but I can add that considering that it's a fairly easy change.

htejun:
> Hm... but the per-task allocation will be performed by the BPF part. Are you suggesting to allocate some memory from `.init_task()` and share that memory in the address space of the user-space scheduler?

No, I was just thinking about more general case of userspace component needing per-task memory allocation. For rustland, the per-task allocation happens in `drain_queued_tasks()` with `.or_insert_with_key()` call on `task_map`. This can alternatively structured so that BPF `ops.init_task()` calls userspace to trigger insertion so that we don't have dynamic allocation in the scheduling path. This isn't an immediate problem for rustland, but if you imagine a scheduler implementation that wants to service setups where there may be hundreds of thousands of tasks (which does happen in some fringes), setting up the fixed pool upfront can be a bit challenging. Being able to split out the per-task context allocation from the mlockall-backed fixed heap would allow side stepping the issue in most cases.

It isn't anything we need to worry about now. Just something to keep in mind for the future.

PR #90, opened 2024-01-16
scx_rustland: introduce SMT support

arighi:
Introduce a basic support of CPU topology awareness. With this change, the scheduler will prioritize dispatching tasks to idle CPUs with fewer busy SMT siblings, then, it will proceed to CPUs with more busy SMT siblings, in ascending order.

To implement this, introduce a new CoreMapping abstraction, that provides a mapping of the available core IDs in the system along with their corresponding lists of CPU IDs. This, coupled with the get_cpu_pid() method from the BpfScheduler abstraction, allows the user-space scheduler to enforce the policy outlined above and improve performance on SMT systems.

Keep in mind that this improvement is relevent only when the amount of tasks running in the system is less than the amount of CPUs. As soon as the amount of running tasks increases, they will be distributed across all available CPUs and cores, thereby negating the advantages of SMT isolation.

### DIFF
--- a/scheds/rust/scx_rustland/src/topology.rs
+++ b/scheds/rust/scx_rustland/src/topology.rs
@@ -1,0 +1,84 @@
+// Copyright (c) Andrea Righi <andrea.righi@canonical.com>
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use std::collections::HashMap;
+use std::fs;
+
+/// scx_rustland: CPU topology.
+///
+/// CoreMapping provides a map of the CPU topology, with the list of CPU ids in the system grouped
+/// by their corresponding core id.
+///
+/// The CPU / core mapping is stored in a HashMap using the core ID as the key. An example content
+/// of the HashMap can be the following:
+///
+///  core 0: [4, 0]
+///  core 1: [5, 1]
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_rustland/src/topology.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Andrea Righi <andrea.righi@canonical.com>
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use std::collections::HashMap;
+use std::fs;
+
+/// scx_rustland: CPU topology.
```
Decave, line 9: Not necessary for this change, but his could be useful to provide as a rust crate as well. In scx_rusty we collect the host's topology to determine what domains we should create for the scheduler. If we had a single utility for mapping logical CPUs -> physical cores -> LLC packages -> NUMA nodes, I think that would be quite handy.

arighi: @Decave yes, there's a lot of generic stuff that we can reuse in other schedulers (the cpu topology, the allocator, the bpf abstraction). It would be really nice to have them in the scx_utils crate. I'll start look at that and see if we can migrate some of these "generic building blocks" in there.

PR #93, opened 2024-01-17
Set log size to 10M

sirlucjan:
@arighi @htejun @Decave
As mentioned in the previous PR #88  that could help solve the log problem.

### DIFF
--- a/services/scx_central.service
+++ b/services/scx_central.service
@@ -6,7 +6,9 @@ ConditionPathExists=/sys/kernel/debug/sched/ext
 Type=simple
 ExecStart=scx_central
 Restart=always
-StandardOutput=null
+StandardOutput=journal
+StandardError=journal
+LogNamespace=sched-ext
 
 [Install]
 WantedBy=multi-user.target
--- a/services/scx_central.service
+++ b/services/scx_central.service
@@ -6,7 +6,6 @@ ConditionPathExists=/sys/kernel/debug/sched/ext
 Type=simple
 ExecStart=scx_central
 Restart=always
-StandardOutput=journal
 StandardError=journal
 LogNamespace=sched-ext
 
--- a/services/journald@sched-ext.conf
+++ b/services/journald@sched-ext.conf
@@ -1,0 +1,7 @@
+[Journal]
+# Sets the maximum journal size to 10M for each namespace.
+NamespaceMaxUse=10M
+# Sets the maximum log age to 7 days.
+MaxRetentionSec=7day
+# Sets the priority level for messages to be stored to info.
+MaxLevelStore=info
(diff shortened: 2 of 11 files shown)


kode54:
I see these in my journal when running these updated services:

```
Jan 16 18:27:20 mrgency systemd[1]: /usr/lib/systemd/system/scx_rustland.service:11: Unknown key name 'SystemMaxUse' in section 'Service', ignoring.
Jan 16 18:27:20 mrgency systemd[1]: /usr/lib/systemd/system/scx_rustland.service:12: Unknown key name 'RuntimeMaxUse' in section 'Service', ignoring.
Jan 16 18:27:26 mrgency systemd[1]: /usr/lib/systemd/system/scx_rustland.service:11: Unknown key name 'SystemMaxUse' in section 'Service', ignoring.
Jan 16 18:27:26 mrgency systemd[1]: /usr/lib/systemd/system/scx_rustland.service:12: Unknown key name 'RuntimeMaxUse' in section 'Service', ignoring.
```

sirlucjan:
@arighi So it's just as I thought. We will have to add a journald configuration file. We need to think this through somehow

arighi:
hm... I'm wondering, do we need to move these option inside a `[Journal]` section? Something like:
```
[Journal]
SystemMaxUse=5M
RuntimeMaxUse=5M
```

Decave:
@sirlucjan: Maybe Journal Namespaces could be an option? https://www.freedesktop.org/software/systemd/man/latest/systemd-journald.service.html#Journal%20Namespaces

sirlucjan:
@Decave I've to take a look.

sirlucjan:
@kode54 @arighi @htejun @Decave 

It seems that everything works well
Screenshot_20240117_231649

services/scx_central.service
```diff
@@ -6,7 +6,9 @@ ConditionPathExists=/sys/kernel/debug/sched/ext
 Type=simple
 ExecStart=scx_central
 Restart=always
-StandardOutput=null
+StandardOutput=journal
+StandardError=journal
```
htejun, line 10: Aren't these the default?

htejun (approved):

services/scx_central.service
sirlucjan: To be honest, I wasn't sure - I preferred to give just in case

sirlucjan:
@htejun I've removed StandardOutput - because its default value is actually journal. StandardError has a different default value, so I haven't removed it so far.

Decave:
The failure appears to be a double fault that's unrelated to your change: https://pastebin.com/FVJLm1N0.

arighi:
FWIW, tested also on Ubuntu, everything works perfectly, thanks!

PR #94, opened 2024-01-17
scx-rustland: SMT improvements

arighi:
This PR contains an improvement in the SMT logic, it's actually a simplification and, performance-wise, it seem to deliver better results, so it's a win-win IMHO.

As mentioned in the git commit:
```
    Some preliminary results on an AMD Ryzen 7 5800X 8-Cores (SMT enabled):
    running my usual benchmark of measuring the fps of a videogame
    (Counter-Strike 2) during a parallel kernel build-induced system
    overload, shows an improvement of approximately 2x (from 8-10fps to
    15-25fps vs 1-2fps with EEVDF).
```

The other commit is more like a clean-up, apparently now we can drop the `scx_bpf_kick_cpu()` from `update_idle()` without any noticeable performance drop, so let's just get rid of it.

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -572,14 +572,8 @@ void BPF_STRUCT_OPS(rustland_update_idle, s32 cpu, bool idle)
 	 * A CPU is now available, notify the user-space scheduler that tasks
 	 * can be dispatched.
 	 */
-	if (is_usersched_needed()) {
-		/*
-		 * Kick the CPU to make it immediately ready to accept
-		 * dispatched tasks.
-		 */
-		scx_bpf_kick_cpu(cpu, 0);
+	if (is_usersched_needed())
 		set_usersched_needed();
-	}
 }
 
 /*
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -273,30 +273,30 @@ impl<'a> Scheduler<'a> {
             .map(|cpu_id| self.bpf.get_cpu_pid(cpu_id))
             .collect();
 
-        // Create a Vec of tuples containing core id and the number of busy CPU ids for each core.
-        let mut core_status: Vec<(i32, usize)> = cores
+        // Generate the list of idle cores (cores that don't have any task running on their CPUs).
+        let core_idle: Vec<i32> = cores
             .iter()
-            .map(|(&core_id, core_cpus)| {
-                let busy_cpus = core_cpus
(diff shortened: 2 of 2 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -572,14 +572,8 @@ void BPF_STRUCT_OPS(rustland_update_idle, s32 cpu, bool idle)
 	 * A CPU is now available, notify the user-space scheduler that tasks
 	 * can be dispatched.
 	 */
-	if (is_usersched_needed()) {
-		/*
-		 * Kick the CPU to make it immediately ready to accept
-		 * dispatched tasks.
-		 */
-		scx_bpf_kick_cpu(cpu, 0);
```
htejun, line 580: Hmm... I wonder why this is. It *might* be useful to add an option to re-enable this, so that if ppl report stalls, they can try the option as a debug step.

htejun (approved):

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: 'Hm... even with v0.1.5 dropping `scx_bpf_kick_cpu()` doesn't seem to make any difference, any chance that maybe something changed in the kernel? (Going before v0.1.5 is a bit tricky because of the ABI change)

arighi: I'm also wondering if the fact that now I'm always dispatching from `.dispatch()` with `SCX_DSQ_LOCAL_ON | cpu` (using the cpu selected by the user-space scheduler) implicitly kicks the target cpu, therefore the kick in `.update_idle()` is redundant.

htejun: Kernel side hasn't changed. Yeah, `LOCAL_ON|cpu` will wake up the CPU.

arighi: ok, I think everything makes sense now. I was using `SCX_DSQ_GLOBAL` for the scheduler and it wasn't waking up without the explicit kick cpu. Then I changed the scheduler to also use `LOCAL_ON|cpu`, so if that implicitly kicks the cpu we don't need the kick in update_idle anymore.

htejun: Sounds great. No need to add the option then. That sounds like it should be reliable. Thanks.

arighi: I'll update the commit message, so it's more clear why we dropped that.

PR #95, opened 2024-01-18
ci: test the shedulers with the latest sched-ext kernel

arighi:
Instead of downloading a precompiled sched-ext enabled kernel from the Ubuntu ppa, fetch the latest kernel directly from the sched-ext git repository and recompile it on-the-fly using virtme-ng.

This allows to get rid of the Ubuntu ppa dependency, take out from the equation potential Ubuntu-specific patches, and ensures testing all the schedulers with the most up-to-date sched-ext kernel (that should also help to detect potential kernel-related issues in advance).

The downside is that the CI runs will take a bit longer now, because we are recompiling the kernel from scratch. However, the kernel built with virtme-ng is relatively quick to compile and includes all the sched-ext features required for testing.

It's worth noting that this method aligns with the current sched-ext kernel CI, where we test only the in-kernel schedulers (as intended).

This change allows to extend the test coverage, using the same kernel to test also the schedulers included in this repository.

### DIFF
--- a/.github/workflows/build-scheds.yml
+++ b/.github/workflows/build-scheds.yml
@@ -10,17 +10,13 @@ jobs:
       # Hard turn-off interactive mode
       - run: echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-selections
 
-      # Add sched-ext external ppa to get the latest sched-ext kernel
-      - run: sudo add-apt-repository -n -y ppa:arighi/sched-ext-unstable
-      - run: sudo sed -i s/jammy/noble/ /etc/apt/sources.list.d/arighi-ubuntu-sched-ext-unstable-jammy.list
-
       # Refresh packages list
       - run: sudo apt update
 
       ### DOWNLOAD AND INSTALL DEPENDENCIES ###
 
       # Download dependencies packaged by Ubuntu
-      - run: sudo apt -y install coreutils cmake cargo elfutils libelf-dev libunwind-dev libzstd-dev linux-headers-generic linux-tools-common linux-tools-generic ninja-build python3-pip python3-requests qemu-kvm udev iproute2 busybox-static libvirt-clients kbd kmod file rsync zstd
+      - run: sudo apt -y install bison busybox-static cargo cmake coreutils cpio elfutils file flex gcc git iproute2 kbd kmod libcap-dev libelf-dev libunwind-dev libvirt-clients libzstd-dev linux-headers-generic linux-tools-common linux-tools-generic make ninja-build pahole python3-dev python3-pip python3-requests qemu-kvm rsync rustc udev zstd
 
       # clang 17
       # Use a custom llvm.sh script which includes the -y flag for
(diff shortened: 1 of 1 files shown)


.github/workflows/build-scheds.yml
```diff
@@ -51,16 +47,21 @@ jobs:
       # Install virtme-ng
       - run: pip install virtme-ng
 
-      # Download a sched-ext enabled kernel
-      - run: apt download linux-image-unsigned-6.7.0-4-generic linux-modules-6.7.0-4-generic
-      - run: mkdir -p kernel
-      - run: for f in *.deb; do dpkg -x $f kernel; done
+      # Get the latest sched-ext enabled kernel directly from the git repository
+      - run: git clone https://github.com/sched-ext/sched_ext.git linux
```
htejun, line 51: The dev branch is `sched_ext`. One worry is that when we make breaking changes in the kernel, this would immediately break CI for scx repo. Would it be possible to make it try `sched_ext-ci` first and then if that's not present try `sched_ext` so that we have a way to lag a bit when necessary?

htejun (changes requested):

.github/workflows/build-scheds.yml
Decave: Alternatively, we could have a release source tarball for CI that we get and untar here. That would also speed up CI as we wouldn't have to clone the entire repo.

arighi: Then we could just go with `git clone --single-branch -b sched-ext --depth 1`  and it would be like fetching a tarball.

arighi: ...and git ls-remote or similar to check if sched-ext-ci or sched-ext exist.

htejun (approved):

.github/workflows/build-scheds.yml
```diff
@@ -51,16 +47,23 @@ jobs:
...
       - run: pip install virtme-ng
 
-      # Download a sched-ext enabled kernel
-      - run: apt download linux-image-unsigned-6.7.0-4-generic linux-modules-6.7.0-4-generic
-      - run: mkdir -p kernel
-      - run: for f in *.deb; do dpkg -x $f kernel; done
+      # Get the latest sched-ext enabled kernel directly from the git
+      # repository (try the sched_ext-ci branch first, if it doesn't exit use
+      # sched_ext)
+      - run: git clone --single-branch -b sched_ext-ci --depth 1 git@github.com:sched-ext/sched_ext.git || git clone --single-branch -b sched_ext --depth 1 git@github.com:sched-ext/sched_ext.git
```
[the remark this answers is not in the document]
arighi: yeah, sorry for all the pushes and failed builds, it should work now. :)

But I'm glad that we had this conversation, the git clone part is *a lot* faster now, we should do the same in the kernel CI!

Decave (approved):
Fantastic, thanks!!

PR #97, opened 2024-01-19
Set the correct value for sched-ext journald namespace

sirlucjan:
This pull request fixes an erroneous entry fixing the size of the sched-ext log in journald namespace.

However the logs were saved correctly in a separate space, so the mistake in the name of the directive made it impossible to limit its size.

### DIFF
--- a/services/journald@sched-ext.conf
+++ b/services/journald@sched-ext.conf
@@ -1,7 +1,9 @@
 [Journal]
 # Sets the maximum journal size to 10M for each namespace.
-NamespaceMaxUse=10M
+SystemMaxUse=10M
 # Sets the maximum log age to 7 days.
 MaxRetentionSec=7day
 # Sets the priority level for messages to be stored to info.
 MaxLevelStore=info
+# Sets the maximum number of log files that can be stored for each namespace.
+SystemMaxFiles=5


services/journald@sched-ext.conf
```diff
@@ -1,7 +1,9 @@
 [Journal]
 # Sets the maximum journal size to 10M for each namespace.
-NamespaceMaxUse=10M
+SystemMaxUse=10M
 # Sets the maximum log age to 7 days.
 MaxRetentionSec=7day
 # Sets the priority level for messages to be stored to info.
 MaxLevelStore=info
+# Sets the maximum number of log files that can be stored for each namespace.
+SystemMaxFiles=5
```
Decave, line 9: Is this necessary if we set `SystemMaxUse` above?

sirlucjan: I read in the journal documentation that it is sometimes worth adding, but of course I can remove it. This limits the default value from 100 to 5, which seems to make sense when using this exclusively for debugging one namespace.

sirlucjan: @Decave I'm also wondering if it's better to increase the space limit to 25M because scx sends a lot of logs to journal.

arighi: So, IIUC SystemMaxSize determines the max file size to trigger the rotation and SystemMaxFiles determines how many rotations we want to keep before dropping them.

In that case maybe bump up a little bit the max size and for max files 5 is probably good enough.

arighi: ... and I tested this also on Ubuntu, all good.

arighi (approved):

services/journald@sched-ext.conf
```diff
@@ -1,7 +1,9 @@
 [Journal]
 # Sets the maximum journal size to 10M for each namespace.
```
Decave, line 2: Comment needs to be updated

PR #99, opened 2024-01-20
scx_rustland: fixes to improve scheduler stability

arighi:
A set of fixes to improve rustland stability:
  - scx_rustland: increase max amount of enqueued tasks (allow to better sustain task-intensive stress tests)
  - scx_rustland: mitigate unevictable memory page faults (prevent page faults triggered by kcompactd with unevictable memory)
  - scx_rustland: use custom pcpu DSQ instead of SCX_DSQ_LOCAL{_ON} (properly enforce CPU affinity, thanks to @Decave )


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -513,8 +517,22 @@ void BPF_STRUCT_OPS(rustland_dispatch, s32 cpu, struct task_struct *prev)
...
+		 * one. This allows the task to run immediately, even if the
+		 * target CPU is not awake.
+		 */
+		if (task.cpu != cpu)
+			scx_bpf_kick_cpu(task.cpu, 0);
+	}
+
+	/* Consume all tasks enqueued in the current CPU's DSQ */
+	bpf_repeat(MAX_ENQUEUED_TASKS) {
+		if (!scx_bpf_consume(cpu))
```
htejun, line 534: Let's say a task is enqueued in usersched, and then has its affinity updated, and then usersched dispatches the task targeting a CPU which is no longer in the task's cpumask. It wouldn't fail right away but wouldn't the task end up in the wrong dsq and never be consumed?

scheds/rust/scx_rustland/src/bpf/alloc.rs
```diff
@@ -44,6 +46,71 @@ static MEMORY: RustLandMemory = RustLandMemory {
...
+    }
+
+    // Save all the sysctl VM settings in the internal state.
+    fn save(&self) {
+        let compact_unevictable_allowed = "/proc/sys/vm/compact_unevictable_allowed";
+        let value = self.read_procfs(compact_unevictable_allowed);
+        unsafe {
+            VM.compact_unevictable_allowed = value;
+        };
+        self.write_procfs(compact_unevictable_allowed, 0);
```
htejun, line 98: I wonder whether BPF part can detect usersched thread getting blocked on a page fault (or anything which isn't kosher) and then just direct dispatch everything till the usersched gets unblocked. This would require BPF part to keep track of what's queued on the usersched side. Hmm... that's already tracked per-rq in sched_ext core. Maybe we just need to something similar to `scx_bpf_reenq_local()` which re-enqueues everything which is not queued on one of the dsq's.

htejun (approved):

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: > Let's say a task is enqueued in usersched, and then has its affinity updated, and then usersched dispatches the task targeting a CPU which is no longer in the task's cpumask. It wouldn't fail right away but wouldn't the task end up in the wrong dsq and never be consumed?

The usersched doesn't have any knowledge of the cpumask, it assumes that all tasks can run on all CPUs.

It's the BPF dispatcher that validates, in `.dispatch()`, if the target CPU is allowed or not, according to the cpumask. If it's not allowed, it will attempt to dispatch the task to the prev CPU, if that is also not allowed we pick a random CPU from the allowed ones, according to the cpumask.

So, if the affinity changes while the task is enqueued in the usersched, no problem, the BPF part would still bounce the task to a valid CPU.

However, if the affinity changes while the `.dispatch()` is executed (but I'm not sure if that's possible, I assume there's some locking involved while the `.dispatch()` callback is executed), then there could be a race and we may end up consuming the task on a CPU that is not usable by the task.

So the question is, can cpumask change concurrently while `.dispatch()` is executed?

scheds/rust/scx_rustland/src/bpf/alloc.rs
arighi: > I wonder whether BPF part can detect usersched thread getting blocked on a page fault (or anything which isn't kosher) and then just direct dispatch everything till the usersched gets unblocked. This would require BPF part to keep track of what's queued on the usersched side. Hmm... that's already tracked per-rq in sched_ext core. Maybe we just need to something similar to `scx_bpf_reenq_local()` which re-enqueues everything which is not queued on one of the dsq's.

Do you think userfaultfd can help with that? What if the usersched creates a separate userfaultfd thread, when a page fault happens it will be intercepted by the userfaultfd thread that can set a flag in the BPF part, that would temporarily switch it into a "reschedule everything without usersched" mode and then switch back to "regular" mode once the resched is done.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
htejun: `set_cpus_allowed()` is exclusive to scheduling operations through `.dequeue()`. rustland doesn't implement `.dequeue()` but the intended design is that sched_ext core still keeps track of task state so that it can automatically ignore spurious dispatches which take place after dequeue (and followup re-enqueue). Hmm.. I'm not quite sure why this didn't work.

If dispatch and cpumask update are racing, what matters is 1. whether `finish_dispatch()` claims the task before the task gets dequeued by `set_cpus_allowed()` path, and if so 2. whether the task gets put on the target CPU's local dsq before dequeue. If either fails, the dispatch attempt is ignored as they can be treated as the task getting first dispatched and then dequeued. If both succeeded, the task is in local dsq and then dequeued from it.

You're hitting the `scx_ops_error()` call at the end of `dispatch_to_local_dsq()` without this patch, right? Hmm... right, the rq locked here can be a different one from the one the target task is associated with, but `DISPATCHING` is set, which should block `ops_dequeue()`.

I reproduce the problem and understand why it ends up dispatching to the wrong CPU. What should I do to reproduce?

arighi: @htejun it's very easy, try `stress-ng --race-sched 8 --timeout 10`, it should trigger the ops error almost immediately:
```
09:18:51 [WARN] FAIL: SCX_DSQ_LOCAL[_ON] verdict target cpu 3 not allowed for stress-ng-race-[1004999] (err=1024)
```

scheds/rust/scx_rustland/src/bpf/alloc.rs
htejun: I have never used userfaultfd and am not sure, but from glancing the manpage, it *looks* liked it'd need a thread to be runnable and if the scheduler thread is already blocked on fault, I guess that thread would need a special treatment too to deliver the notification to the BPF.

Hmm... if there's a tracepoint or fentry probe we can attach to monitor the relevant page faults, we can filter it by the PID and notify the BPF scheduler right away skipping going through userspace thread for userfaultfd. See `tp_task_rename()` and `scheduler_tick_fentry()` in `scx_layered` for examples. Would this be an option?

arighi: Right, userfault thread needs to run, but we could save the pid and dispatch it immediately in the BPF part from `.enqueue`, allowing to always bypass the user-space scheduler.

But maybe the tracepoint/fentry way is easier, hm... I'll take a look.

arighi:
I've done some small adjustment to the custom percpu DSQ patch, mostly a little code refactoring (and I moved the kick cpu after doing the dispatch, that seems to make everything much more responsive).

I also added a patch that seems to improve a bit the responsiveness of interactive apps under IO pressure and another change that should prevent stalls when the user-space scheduler is preempted by a task from a higher sched_class (implementing .cpu_release as suggested by @htejun). However, I haven't been able to reproduce this problem yet, so before merging it, I'll ask around some people to do some tests.

arighi:
I think I will merge this PR. It is not perfect, but it makes things a lot better.

In particular, I'm not very enthusiastic about reverting the pre-CPU DSQ idea (suggested by @Decave ), but with this change the scheduler is showing performance regressions, so the plan is to go back to a simple and safe way that also doesn't break cpu affinity, in order to make the scheduler usable again and as bug-free as possible.

Then I'll try to figure out the reasons of the regression, so that we can re-add the per-CPU DSQ later, that seems the proper design for this scheduler.

PR #100, opened 2024-01-24
Add README.md for systemd services

sirlucjan:
Brief information on how to use systemd services with a bunch of useful journald commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ $ cargo install scx_rusty
 
 and `scx_rusty` will be built and installed as `~/.cargo/bin/scx_rusty`.
 
-## SystemD services
+## systemd services
 
 See: [services](services/README.md)
 
--- a/README.md
+++ b/README.md
@@ -364,6 +364,9 @@ $ cargo install scx_rusty
 
 and `scx_rusty` will be built and installed as `~/.cargo/bin/scx_rusty`.
 
+## SystemD services
+
+See: [services](services/README.md)
 
 ## Kernel Feature Status
 
--- a/services/README.md
+++ b/services/README.md
@@ -1,6 +1,6 @@
 # A Quick Start Guide
 
-This guide provides instructions on how to start the SCX scheduler and check its logs.
+This guide provides instructions for running the SCX schedulers as a systemd service and checking its logs.
 
 ## Getting Started
 To start the SCX scheduler at boot, you need to run the systemd service as root. Here are the steps:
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -364,6 +364,9 @@ $ cargo install scx_rusty
 
 and `scx_rusty` will be built and installed as `~/.cargo/bin/scx_rusty`.
 
+## SystemD services
```
Decave, line 367: nit: It looks a bit odd to me to say "SystemD" -- can we just stick with "`systemd`"?

services/README.md
```diff
@@ -0,0 +1,54 @@
+# A Quick Start Guide
+
+This guide provides instructions on how to start the SCX scheduler and check its logs.
```
Decave, line 3: I think we should specify that the guide is specifically about how to start the scheduler as a systemd service. What do you think about this?

This guide provides instructions on how to use an SCX scheduler as a systemd service.

services/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+- To check the status of the service, use the following command:
+
+```
+systemctl status scx_SCHEDNAME
+```
+
+## Checking Journald Logs
+
+The SCX schedulers do not log to the main journald. Instead, they save logs in a dedicated journald namespace.
```
Decave, line 34: What do you think about:

The SCX schedulers do not log to the default journald namspace. Instead, they save logs in a dedicated `sched-ext` namespace.

Decave (commented):
Thanks for writing this up! Looks good, just left a few comments

sirlucjan:
@Decave Could you check now?

README.md
sirlucjan: Fixed!

services/README.md
sirlucjan: Fixed (probably!)

services/README.md
sirlucjan: Fixed

services/README.md
```diff
@@ -0,0 +1,54 @@
+# A Quick Start Guide
+
+This guide provides instructions for running the SCX schedulers as a systemd services and checking its logs.
```
Decave, line 3: s/as a systemd services/as a systemd service

Decave (commented):
Looks good! Just one grammar nit

services/README.md
sirlucjan: Fixed

PR #102, opened 2024-01-25
systemd-services: replace ConditionPathExists with ConditionPathIsDirectory

sirlucjan:
Commit https://github.com/sched-ext/sched_ext/pull/122 has caused systemd services to no longer be able to run because the startup conditions cannot be met. A simple change to ConditionPathIsDirectory and pointing to the correct path gets everything working properly again. 

Source: https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html

```
ConditionPathIsDirectory=
ConditionPathIsDirectory= is similar to ConditionPathExists= but verifies that a certain path exists and is a directory.

Added in version 244.
```

### DIFF
--- a/services/scx_central.service
+++ b/services/scx_central.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Start scx_central
-ConditionPathExists=/sys/kernel/debug/sched/ext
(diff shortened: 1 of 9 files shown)


Decave:
Thanks for the fix!

PR #106, opened 2024-01-25
ci: print the latest commit of the checked out sched-ext kernel

arighi:


htejun (approved):

.github/workflows/build-scheds.yml
```diff
@@ -52,6 +52,9 @@ jobs:
       # sched_ext)
       - run: git clone --single-branch -b sched_ext-ci --depth 1 https://github.com/sched-ext/sched_ext.git linux || git clone --single-branch -b sched_ext --depth 1 https://github.com/sched-ext/sched_ext.git linux
 
+      # Print the latest commit of the checked out sched-ext kernel
+      - run: cd linux && git log -1 --pretty=format:"%h %ad %s" --date=short
+
       # Build a minimal kernel (with sched-ext enabled) using virtme-ng
       - run: cd linux && vng -v --build --config .github/workflows/sched-ext.config
```
Decave, line 59: Won't this break now that we're doing `cd linux` twice in a row?

Decave: @arighi / @htejun

arighi: @Decave theoretically no, because each `- run` spawns a separate shell, so when it's done it goes back to the previous path.

PR #110, opened 2024-01-27
scx_rustland: per-CPU DSQs + global shared DSQ

arighi:
Opening this PR mostly as an open discussion.

**Main topic**
Designing a universal BPF layer that allows to implement totally generic schedulers in user-space.

**Specific goal addressed by this PR**
The BPF part should provide an interface that allows the user-space scheduler to select a specific CPU to dispatch a task and it should also allow to not specify any CPU at all (in this case the task can be dispatched in any CPU, like the first one that becomes available).

**Implementation**
The idea is to adapt @Decave's suggestion of introducing per-CPU DSQs, the user-space scheduler sends to the BPF dispatcher a PID with an optional target CPU; the BPF part dispatches the task to the DSQ associated to the target CPU, if specified, otherwise the task is dispatched to a global shared DSQ. Per-CPU DSQs are consumed from the `dispatch()` callback of their corresponding CPU. Shared DSQ is consumed from the `dispatch()` callback of any CPU.

In this way we can provide a totally generic interface that any user-space scheduler can use to implement any type of CPU selection logic.

**Problem**
A task is dispatched to a specific per-CPU DSQ, the `dispatch()` callback is never called on that CPU => starvation.

I haven't been able to find a nice way to enforce the call of the `dispatch()` event on a specific CPU, so that tasks queued there can be consumed. I was assuming that `scx_bpf_kick_cpu()` could be used for this, but it doesn't seem to be the case.

Apparently all this logic works (no starvation) if the target CPU of a task is always the same as the one determined in the `select_cpu` callback. Basically the user-space scheduler can only "acknowledge" `select_cpu`. What I would like to do, instead, is give the ability to the user-space scheduler to potentially override the choice made in `select_cpu`.

So, at the moment, as a workaround, when the user-space scheduler tries to dispatch a task on a CPU that is different than the one selected in `select_cpu`, I simply redirect the task to the shared DSQ (that is consumed from all the CPUs) => no starvation. But I would definitely prefer if we could honor the decision made by the user-space scheduler.

Does all of this make sense? Any suggestion? Thanks!

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -45,6 +45,15 @@ struct user_exit_info uei;
  */
 #define MAX_CPUS 1024
 
+/*
+ * Introduce a custom DSQ shared across all the CPUs, where we can dispatch
+ * tasks that will be executed on the first CPU available.
+ *
+ * Per-CPU DSQs are also provided, to allow the scheduler to run a task on a
+ * specific CPU (see dsq_init()).
+ */
+#define SHARED_DSQ MAX_CPUS
+
 /* !0 for veristat, set during init */
 const volatile s32 num_possible_cpus = 8;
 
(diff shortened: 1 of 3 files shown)


arighi:
@Decave @htejun I think I figured out my starvation issue! It looks like the problem was that I still needed to check the cpumask before dispatching a task to a per-CPU DSQ.

If I dispatch (and consume) a task to an "invalid" CPU I can immediately trigger the starvation issue. So the fix is to check the cpumask before calling `scx_bpf_dispatch()` and if the CPU picked by the user-space scheduler is invalid the BPF layer will re-direct the task to the SHARED_DSQ, kicking the CPU selected by the built-in idle selection logic, so the task is going to run there, using a valid CPU, all good.

In this way the BPF layer is able to provide a generic interface to the user-space scheduler, that is able to pick any arbitrary CPU when dispatching a task, or not specify any CPU at all (`task.cpu = NO_CPU`) and, in this case, the BPF dispatcher will rely on the built-in idle selection logic. If the user-space selects an invalid CPU the BPF layer will fix that transparently, redirecting the task to a valid CPU.

I've updated this PR with all of the above + also some low-latency improvements on top, because I've been testing the scheduler a lot today with all these changes applied and it seems to work pretty well.

All the games that I've tested seem to be responsive, no starvation, no stall, both with and without the background `make -j N` (varying N from 8, 16, 32, 64). I also ran `stress-ng --race-sched N`, as well as `stress-ng -c N`. I even did a video call while building the kernel (`make -j 32`) and the audio/video was really smooth.

In conclusion, everything seems to work as expected now (even better than expected I'd say... at least on my desktop and my laptop).

scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -253,49 +262,112 @@ static bool test_and_clear_usersched_needed(void)
...
+		break;
+	default:
+		/*
+		 * Dispatch a task to a specific per-CPU DSQ if the target CPU
+		 * can be used (according to the cpumask), otherwise redirect
+		 * the task to the shared DSQ.
+		 *
+		 * In the future we may want to provide a way to check the
+		 * cpumask in advance from user-space in a proper synchronized
+		 * way, instead of bouncing the task somewhere else, but for
```
htejun, line 322: So, this should only be a problem when cpus_allowed changes. That synchronization is performed through `.dequeue()` and the problem is that that path can't sleep, so implementing actual handshake with userland is likely not possible. The sequence of events is likely as follows:

1. Task gets enqueued and pushed to userspace and queued there.
2. The task's cpus_allowed is going to be updated, so the task gets dequeued (ignored by rustland) and the cpumask is being updated.
3. Userland sched pushes the task to the kernel for dispatching. Actual dispatch hasn't happened yet.
4. The task's cpus_allowed is updated and the task gets reenqueued.
5. The task gets dispatched targeting now wrong CPU.

So, in scenarios like this, it *should* be safe to just ignore the task. The task is already re-enqueued, so it will get re-dispatched. The problem with the earlier attempt likely was that it was dispatched to the wrong dsq. Then, later when the re-enqueued task gets dispatched, it's ignored as it's already on another dsq.

htejun (approved):
It might be worthwhile to see whether just ignoring the task targeting the wrong dsq works. Other than that, looks good to me.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: Thanks tons for looking at this @htejun !

I've tried to ignore the task completely when the cpumask is not valid, but I still get the `failed to run for 5s` error. To reproduce this I'm always dispatching to CPU#2 from the user-space scheduler and then I run a `stress-ng --race-sched 16` that spawn 16 tasks that are constantly bouncing around using sched_setaffinity() in a busy loop.

However, it looks like we can actually ignore the task if target CPU is not valid *and* also if the CPU assigned to the task (scx_bpf_task_cpu()) is also not valid, see d13ed5c. Does it make sense to you? With this in place everything seems to work and I can't break the scheduler. So I added this change to the PR and updated some comments here and there.

I'm currently stress testing this, but so far so good.

PR #121, opened 2024-02-03
scx_rustland: prevent duplicate PIDs in the task BTreeSet

arighi:
Items in the task BTreeSet are stored by pid and vruntime. Make sure that we never store multiple items with the same PID, so that re-enqueued tasks are not dispatched multiple times.


htejun (approved):
This makes push O(N), right? Maybe change / addition of data structures is needed?

arighi:
Yeah the O(N) is not really nice, especially under massive stress test conditions. I can improve that part, but at least it's correct for now.

PR #122, opened 2024-02-03
common.bpf.h: Add kfunc prototype for scx_bpf_dispatch_cancel()

htejun:
And relocate scx_bpf_dispatch_nr_slots() while at it.


arighi (approved):
I haven't tested it yet (I'll prepare an Ubuntu kernel with https://github.com/sched-ext/sched_ext/pull/135 soon), but in the meantime we can apply this. Thanks!

Issue #123, opened 2024-02-04
scx_userland errored


romanstingler:
```
Feb 04 18:48:37 roman kernel: sched_ext: BPF scheduler "userland" errored, disabling
Feb 04 18:48:37 roman kernel: sched_ext: runnable task stall (watchdog failed to check in for 3.001s)
Feb 04 18:48:37 roman kernel:    scheduler_tick+0x34a/0x3f0
Feb 04 18:48:37 roman kernel:    update_process_times+0x77/0x80
Feb 04 18:48:37 roman kernel:    tick_nohz_highres_handler+0xbf/0x120
Feb 04 18:48:37 roman kernel:    __hrtimer_run_queues+0x104/0x2f0
Feb 04 18:48:37 roman kernel:    hrtimer_interrupt+0xf4/0x390
Feb 04 18:48:37 roman kernel:    __sysvec_apic_timer_interrupt+0x4b/0x160
Feb 04 18:48:37 roman kernel:    sysvec_apic_timer_interrupt+0x36/0x80
Feb 04 18:48:37 roman kernel:    asm_sysvec_apic_timer_interrupt+0x1a/0x20
Feb 04 18:48:37 roman kernel: warning: `ThreadPoolForeg' uses wireless extensions which will stop working for Wi-Fi 7 hardware; use nl80211
```

I watched a YT video and had gnome-system-monitor open searching / filtering for scx
I had a foot term open executing scx_userland and watching the output
and one term with journalctl -f
and one other terminal which I closed and I opened vs code running in wayland
```
Version: 1.86.0-insider
Commit: e244acbb172c428cb219717a07bf55d2737492ca
Date: 2024-01-22T05:35:29.903Z
Electron: 27.2.1
ElectronBuildId: 26149897
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Linux x64 6.8.0-rc2-2-cachyos-rc-lto
```
and I got the entry in the journal that userland errored.

htejun:
@Decave will have more insight but `scx_userland` is very much a proof-of-concept code which may not be able to survive actual usage. If you're interested in testing out userspace scheduler, `scx_rustland` is a very similar concept but developed a lot further, expected to perform reasonably with actual usage and being actively worked on.

Decave:
[...]

PR #124, opened 2024-02-04
scx_userland: Print warning about poor performance

Decave:
Let's make it clear that this scheduler isn't expected to perform well, and instead point people to scx_rustland.

Should clarify the situation for issues such as https://github.com/sched-ext/scx/issues/123

Also increases the timeout for `scx_userland` back to the system default.


htejun (approved):

arighi:
@Decave it looks like the CI isn't happy with this change, because the new "WARNING" is interpreted as a failure. I'm wondering if we should just skip scx_userland from the CI tests or maybe print the message rephrasing "WARNING" to something else, "NOTICE" maybe?

Decave:
Ahh, I see. Ok, let's replace WARNING with NOTICE.

PR #125, opened 2024-02-04
ci: Only do CI runs for pull requests

Decave:
We're basically always runnin two CI jobs: one for a remote push, and another for when a PR is opened. These are essentially measuring the same thing, so let's save CI bandwidth and just do a PR run. This will hopefully make things a bit less noisy as well.


arighi (approved):
Makes sense to me.

PR #127, opened 2024-02-05
scx_userland: Change WARNING to NOTICE

Decave:
Andrea pointed out that printing WARNING like this breaks CI. Let's just replace WARNING with NOTICE.

### DIFF
--- a/scheds/c/scx_userland.c
+++ b/scheds/c/scx_userland.c
@@ -321,7 +321,7 @@ static void print_example_warning(const char *sched)
 	const char *warning_fmt =
 		"**************************************************************************\n"
 		"\n"
-		"WARNING: %s is an example scheduler.\n"
+		"NOTICE: %s is an example scheduler.\n"
 		"\n"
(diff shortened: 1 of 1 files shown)


Decave:
@arighi Looks like we'd also need to remove the "timeout" keyword. Maybe a better solution would be to update the test script to have a set of schedulers that it only verifies can be loaded, and instead just makes sure it doesn't see "BPF program load failed"? Let me know what you think and I can send out a PR that does that

htejun (approved):

arighi:
@Decave yeah the "grep keywords" approach can potentially trigger some false positives. I was also considering to split stderr/stdout (with stderr containing the stderr of the scheduler + the kernel log) an grep the keywords only from the stderr of vng. In this way we should prevent the false positives coming from the scheduler's stdout.

arighi (approved):

arighi:
Something like #128, let's see if that works...

Decave:
Looks like https://github.com/sched-ext/scx/pull/128 did indeed work, so this can be closed. Thanks!

PR #128, opened 2024-02-05
ci: detect errors only from stderr

arighi:
Search for potential errors only in the kernel logs and the scheduler stderr.

In this way we can use "error keywords" in the scheduler's output without triggering false positives in the CI (see for example #127).

NOTE: this works, because virtme-ng, when executed in verbose mode, sends the kernel messages to stderr (together with the command's stderr) and it channels the command's stdout to the stdout of the host.


htejun (approved):

Decave (approved):
Heh, beat me to it! Thanks for the super quick turnaround

PR #129, opened 2024-02-06
Implement solution to infeasible weights problem

Decave:
As described in [this paper], there is an open problem in load balancing
called the "infeasible weights" problem. Essentially, the problem boils
down to the fact that a task with disproportionately high load can be
granted more CPU time than they can actually consume per their duty
cycle.

This patch implements a solution to that problem, wherein we apply the
algorithm described in this paper to adjust all infeasible weights in
the system down to a feasible wight that gives them their full duty
cycle, while allowing the remaining feasible tasks on the system to
share the remaining compute capacity on the machine.

[this paper]: https://drive.google.com/file/d/1fAoWUlmW-HTp6akuATVpMxpUpvWcGSAv/view?usp=drive_link

### DIFF
--- a/scheds/rust/scx_rusty/src/main.rs
+++ b/scheds/rust/scx_rusty/src/main.rs
@@ -815,7 +815,7 @@ impl<'a, 'b, 'c> LoadBalancer<'a, 'b, 'c> {
             loop {
                 let last_pushed = pushed;
 
-                // Pull from the most imbalaned to least.
+                // Pull from the most imbalanced to least.
                 let mut doms_to_pull = BTreeMap::<_, _>::new();
                 std::mem::swap(&mut self.doms_to_pull, &mut doms_to_pull);
                 let mut pull_doms = doms_to_pull.into_iter().rev().collect::<Vec<(_, _)>>();
--- a/scheds/rust/scx_rusty/src/main.rs
+++ b/scheds/rust/scx_rusty/src/main.rs
@@ -6,6 +6,9 @@ mod bpf_skel;
 pub use bpf_skel::*;
 pub mod bpf_intf;
 
+#[macro_use]
+extern crate static_assertions;
+
 use std::cell::Cell;
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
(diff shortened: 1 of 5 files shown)


Decave:
Updated to address bug where we'd sometimes fail to find an infeasible threshold.

scheds/rust/scx_rusty/src/main.rs
```diff
@@ -144,6 +147,12 @@ struct Opts {
     #[clap(short = 'K', long, default_value = "100.0")]
     kick_greedy_under: f64,
 
+    /// When the system's utilization is lower than this, ignore weights and
+    /// use only duty to make load balancing decisions. Advised to be at least
+    /// 100.0 to accommodate proper load distribution.
+    #[clap(short = 'L', long, default_value = "110.0")]
+    lb_dcycle_under: f64,
```
htejun, line 154: Does this need to be a user-visible parameter? When duty cycle sum is lower than what's needed to derive the solution, the algorithm actually is giving the right answer that everyone needs to be adjusted to the same weight. So, we can skip it if we know that there's gonna be no answer but I don't see what control users would want to have over it.

scheds/rust/scx_rusty/src/main.rs
```diff
@@ -563,46 +604,263 @@ impl<'a, 'b, 'c> LoadBalancer<'a, 'b, 'c> {
...
+        const WEIGHT_PER_BUCKET: f64 = bpf_intf::consts_LB_WEIGHT_PER_BUCKET as f64;
+        let (min_weight, _) = self.bucket_range(bucket);
+
+	// Use the mid-point of the bucket when determining weight
+        min_weight + (WEIGHT_PER_BUCKET / 2.0f64)
+    }
+
+    fn apply_infeas_threshold(&mut self,
+                             doms_dcycles_buckets: &[f64],
+                             infeas_thrsh: &f64) {
```
htejun, line 633: What's with taking a single value type as a reference and then cloning it? Is there some advantage for doing that?

htejun (approved):
I haven't followed each step but generally looks good to me. Left a couple minor comments.

scheds/rust/scx_rusty/src/main.rs
Decave: I agree. I added this based on prior conversations we had where I was under the impression that we wanted this to be similar to greedy threshold, etc. I don't think it makes sense for infeasible weights so I'll remove.

scheds/rust/scx_rusty/src/main.rs
Decave: Sorry, this is left over from a revision where I'd fall back to the caller. Thanks for noticing, will update.

PR #131, opened 2024-02-07
scx: Update vmlinux to use SCX_KICK_IDLE

htejun:
SCX_KICK_IDLE should make idle CPU wakeups by nooping out wakeups if the target CPU is known to be not idle. To use the new feature while not breaking the schedulers on older kernels, add CO-RE based compat wrapper.


arighi (approved):
Thanks for doing this! Looks good to me, I'll prep a new kernel and do some tests ASAP.

PR #132, opened 2024-02-08
scx_rustland: fix cpumask stall and prevent stuttery behavior

arighi:
A set of improvements/fixes for rustland.

In particular:

`scx_rustland: use scx_bpf_dispatch_cancel()`

This uses the new `scx_bpf_dispatch_cancel()` API that prevents the cpumask-related stall conditions. I still can't completely ignore the task when the cpumask is not valid anymore, because there are cases where a task is never dispatched to any DSQ. But I can use the cancel operation to redirect the task to the CPU assigned during select_cpu or to the shared DSQ (if the prev cpu is also not available) in a reliable way. With this logic applied I haven't been able to stall the scheduler using `stress-ng --race-sched`, so in any case it is definitely an improved and a reasonable fix IMHO.

And this also allows to enable the following:

`scx_rustland: keep default CPU selection when idle`

This uses a more reasonable CPU assignment logic in the user-space scheduler, that redirects to tasks to the shared DSQ only when the CPU assigned by the BPF part is not idle anymore.

I've also included a fix that prevents stuttery behaviors (I can systematically reproduce this problem playing Team Fortress 2):

`scx_rustland: kick user-space scheduler when a CPU is released`

The idea is to kick the user-space scheduler immediately in the `.stopping()` callback and speculate on the fact that most of the time there's always another task ready to run. This prevents evident lag issues with Team Fortress 2 and also sporadic lags with Counter-Strike 2. The extra overhead added due to potential unnecessary wake-up events seems to be negligible compared to the benefits.

Last but not least, a small change that can be really useful for debugging:

`scx_rustland: dump scheduler statistics before exiting`

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -571,8 +571,15 @@ void BPF_STRUCT_OPS(rustland_stopping, struct task_struct *p, bool runnable)
 	/*
 	 * Mark the CPU as idle by setting the owner to 0.
 	 */
-	if (!is_usersched_task(p))
+	if (!is_usersched_task(p)) {
 		set_cpu_owner(scx_bpf_task_cpu(p), 0);
+		/*
+		 * Kick the user-space scheduler immediately when a task
+		 * releases a CPU and speculate on the fact that most of the
+		 * time there is another task ready to run.
+		 */
+		set_usersched_needed();
+	}
 }
 
 /*
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -88,7 +88,7 @@ volatile u64 nr_queued;
 volatile u64 nr_scheduled;
 
 /* Dispatch statistics */
-volatile u64 nr_user_dispatches, nr_kernel_dispatches;
+volatile u64 nr_user_dispatches, nr_kernel_dispatches, nr_cancel_dispatches;
 
 /* Failure statistics */
 volatile u64 nr_failed_dispatches, nr_sched_congested;
@@ -299,6 +299,9 @@ static void dispatch_task(struct task_struct *p, u64 dsq_id, u64 enq_flags)
 	switch (dsq_id) {
 	case SCX_DSQ_LOCAL:
 	case SHARED_DSQ:
+		scx_bpf_dispatch(p, dsq_id, slice, enq_flags);
+		dbg_msg("dispatch: pid=%d (%s) dsq=%llu",
+			p->pid, p->comm, dsq_id);
 		break;
 	default:
 		/*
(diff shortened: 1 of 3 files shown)


arighi:
...and I think we need to update https://github.com/sched-ext/sched_ext/tree/sched_ext-ci, otherwise the CI will fail because it's still using the old API.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -311,18 +314,26 @@ static void dispatch_task(struct task_struct *p, u64 dsq_id, u64 enq_flags)
...
+
 		cpu = dsq_to_cpu(dsq_id);
 		if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
 			cpu = scx_bpf_task_cpu(p);
 			if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr))
-				return;
-			dsq_id = dsq_to_cpu(cpu);
+				dsq_id = SHARED_DSQ;
+			else
+				dsq_id = cpu_to_dsq(cpu);
```
htejun, line 325: In this path, it probably should always dispatch to `SHARED_DSQ` because there's no guarantee that there's going to be another racing `set_cpumask()` betwen the second `bpf_cpumask_test_cpu()` and the cancel/dispatch. That said, do you still see stalls if you simply do cancel as follows?
```
scx_bpf_dispatch(p, dsq_id, slice, enq_flags);
if (!bpf_cpumask_test_cpu(dsq_to_cpu(dsq_id), p->cpus_ptr))
        scx_bpf_dispatch_cancel();
```

htejun: It'd be also useful to keep a counter for this event.

htejun (approved):

htejun:
Oh, lemme delete the ci branch. We can track the main branch as there's no API breakge.

htejun:
Oops, the branch is protected. Just updated it.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: > In this path, it probably should always dispatch to `SHARED_DSQ` because there's no guarantee that there's going to be another racing `set_cpumask()` betwen the second `bpf_cpumask_test_cpu()` and the cancel/dispatch. That said, do you still see stalls if you simply do cancel as follows?
> ```
> scx_bpf_dispatch(p, dsq_id, slice, enq_flags);
> if (!bpf_cpumask_test_cpu(dsq_to_cpu(dsq_id), p->cpus_ptr))
>         scx_bpf_dispatch_cancel();
> ```

Hm.. that's true, I'm surprised that I haven't triggered this condition yet, even with massive stress-ng race-sched runs. But it definitely seems safer to just bounce the task to the shared DSQ. I'll change this.

About the stall yes, if I just return after cancel I can easily trigger stalls and in the trace I can see that the stalling task is not dispatched to any DSQ.

arighi: > It'd be also useful to keep a counter for this event.

This is a good idea, I'll add it!

htejun: > About the stall yes, if I just return after cancel I can easily trigger stalls and in the trace I can see that the stalling task is not dispatched to any DSQ.

Hmm... that shouldn't be the case because the task gets requeued after cpumask is updated and thus should re-dispatch the task. Once you land this PR, I'll test and see what's going on.

arighi:
> Oops, the branch is protected. Just updated it.

It looks like `scx_bpf_dispatch_cancel()` is still missing in the sched_ext-ci:
```
libbpf: extern (func ksym) 'scx_bpf_dispatch_cancel': not found in kernel or module BTFs
4728
```

htejun:
> It looks like `scx_bpf_dispatch_cancel()` is still missing in the sched_ext-ci:

Sorry, pushed to the wrong tree. Pushed now. Will give you write perm so that you can update it too.

PR #133, opened 2024-02-10
scx_rustland: per-task cpumask generation counter

arighi:
Introduce a per-task generation counter to check the validity of the cpumask at dispatch time.

The logic is the following:

 - the cpumask generation number is incremented every time a task calls .set_cpumask()

 - when a task is enqueued the current generation number is stored in the queued_task_ctx and relayed to the user-space scheduler

 - the user-space scheduler can decide to dispatch the task on the CPU determined by the BPF layer in .select_cpu(), redirect the task to any other specific CPU, or redirect to the first CPU available (using NO_CPU)

 - task is then dispatched back to the BPF code along with its cpumask generation counter

 - at dispatch time the BPF code checks if the generation number is the same and it discards the dispatch attempt if the cpumask is not valid anymore (the task will be automatically re-enqueued by the sched-ext core code, potentially selecting another CPU / cpumask)

 - if the cpumask is valid, but the CPU selected by the user-space scheduler is invalid (according to the cpumask), the task will be transparently bounced by the BPF code to the shared DSQ (in this way the user-space code can be completely abstracted and dispatches that target invalid CPUs can be automatically fixed by the BPF layer)

This solution can prevent stalls due to dispatches targeting invalid CPUs and it can also avoid redundant dispatch events, making the code more efficient and the cpumask interlocking more reliable.

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -88,7 +88,8 @@ volatile u64 nr_queued;
 volatile u64 nr_scheduled;
 
 /* Dispatch statistics */
-volatile u64 nr_user_dispatches, nr_kernel_dispatches, nr_cancel_dispatches;
+volatile u64 nr_user_dispatches, nr_kernel_dispatches,
+	     nr_cancel_dispatches, nr_bounce_dispatches;
 
 /* Failure statistics */
 volatile u64 nr_failed_dispatches, nr_sched_congested;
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -307,30 +346,58 @@ static void dispatch_task(struct task_struct *p, u64 dsq_id, u64 enq_flags)
...
+		 * and a different cpumask.
 		 */
 		scx_bpf_dispatch(p, dsq_id, slice, enq_flags);
 
+		tctx = lookup_task_ctx(p);
+		if (!tctx)
+			return;
+
+		/* Read current cpumask generation counter */
+		curr_cpumask_cnt = __sync_fetch_and_add(&tctx->cpumask_cnt, 0);
```
htejun, line 369: nit: I don't think __sync op is necessary here, `curr_cpumask_cnt = tctx->cpumask` should be sufficient. All the involved operations are already fully synchronized.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -410,6 +479,11 @@ static void get_task_info(struct queued_task_ctx *task,
 		task->cpu = -1;
 		return;
 	}
+
+	tctx = lookup_task_ctx(p);
+	if (!tctx)
+		return;
+	task->cpumask_cnt = __sync_fetch_and_add(&tctx->cpumask_cnt, 0);
```
htejun, line 486: Ditto.

scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -615,6 +689,20 @@ void BPF_STRUCT_OPS(rustland_update_idle, s32 cpu, bool idle)
...
+ */
+void BPF_STRUCT_OPS(rustland_set_cpumask, struct task_struct *p,
+		    const struct cpumask *cpumask)
+{
+	struct task_ctx *tctx;
+
+	tctx = lookup_task_ctx(p);
+	if (!tctx)
+		return;
+	__sync_fetch_and_add(&tctx->cpumask_cnt, 1);
```
htejun, line 703: And this can just be `tctx->cpumask_cnt++`.

htejun (approved):
LGTM. Left a minor comment. I suppose the next step is making the userspace cpu selection code to consider the allowed cpumask?

arighi:
> LGTM. Left a minor comment. I suppose the next step is making the userspace cpu selection code to consider the allowed cpumask?

Updated (you're right, we don't need __sync_fetch_*(), the cpumask counter is per task and it's all synchronized already).

And yes, the next step will be to expose the cpumask to the user-space so that the scheduler can make better decisions (and at that point we could even make the auto-bounce to the shared DSQ logic optional with a flag or similar).

Thanks for the review!

PR #143, opened 2024-02-20
rust: Add topology module to utils crate

Decave:
scx_rusty has logic in the scheduler to inspect the host to automatically build scheduling domains across every L3 cache. This would be generically useful for many different types of schedulers, so let's add it to the scx_utils crate so it can be used by others.

### DIFF
--- a/rust/scx_utils/Cargo.toml
+++ b/rust/scx_utils/Cargo.toml
@@ -9,12 +9,16 @@ description = "Utilities for sched_ext schedulers"
 
 [dependencies]
 anyhow = "1.0"
+bitvec = { version = "1.0", features = ["serde"] }
 # FIXME - We need to allow both 0.68 and 0.69 to accommodate fedora. See the
 # comment in BpfBuilder::bindgen_bpf_intf() for details.
 bindgen = ">=0.68, <0.70"
 glob = "0.3"
+hex = "0.4.3"
 lazy_static = "1.4"
 libbpf-cargo = "0.22"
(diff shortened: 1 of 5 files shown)


Decave:
This could probably be a done a bit cleaner in terms of encapsulation, etc, but should hopefully be good enough for now. We can further clean up once we have more callers.

htejun (approved):

arighi (approved):
Thanks for this, I'm definitely going to reuse it in scx_rustland!

BTW, I think it would be really helpful to add at the top a simple description of how to use this crate, maybe with a simple example, like how do I initialize a structure that allows me to determine the list of siblings of a specific CPU? Or how can I use the crate to print the CPU topology of my system?

arighi:
Also, nice that you triggered a bug with rustland :smile: 
```
thread 'main' panicked at src/bpf.rs:162:43:
misaligned pointer dereference: address must be a multiple of 0x8 but is 0x56516a51e004
```

Good thing we have CI (looking at the bug).

Decave:
@arighi I added quite a bit of documentation to the crate. I'm going to merge this for now, but please feel free to take a look and provide any additional suggestions, improvements, etc. A couple of things that I want to highlight:

1. I view this very much as in an intermediate state. It should be usable and correct by schedulers, but I think we need to make it a lot more generic. For example, by not allowing the user to create a Topology object from cpumasks, and to instead just always create a fully descriptive representation of the host topology.

2. It doesn't yet have support for identifying NUMA nodes.

PR #145, opened 2024-02-20
scx_rustland: additional comand line options 

arighi:
Add two command line options to adjust rustland behavior without recompiling the code. These are the usual changes that I apply manually in the code to test the "generic layer" of the scheduler.

Having them available as command line options can help for testing and debugging, in particular for users that may need to evaluate how the scheduler works under specific scenarios.

New options:
 - `--full-user`: disable all the in-kernel optimizations and sends all the scheduling events and actions to the user-space scheduler
 - `--builtin-idle`: apply the built-in idle selection logic in the user-space scheduler (scheduler will dispatch tasks in the CPU selected in the `.select_cpu()` callback, using the built-in idle selection logic); also change the default behavior to dispatch tasks on the first CPU available (that helps to improve system responsiveness - e.g., gaming)

### DIFF
--- a/scheds/rust/scx_rustland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rustland/src/bpf/main.bpf.c
@@ -97,6 +97,14 @@ volatile u64 nr_failed_dispatches, nr_sched_congested;
  /* Report additional debugging information */
 const volatile bool debug;
 
+ /*
+  * Enable/disable full user-space mode.
+  *
+  * In full user-space mode all events and actions will be sent to user-space,
+  * basically disabling any optimization to bypass the user-space scheduler.
+  */
+const volatile bool full_user;
+
 /* Allow to use bpf_printk() only when @debug is set */
 #define dbg_msg(_fmt, ...) do {						\
 	if (debug)							\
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_rustland/src/bpf/main.bpf.c
```diff
@@ -511,7 +519,7 @@ void BPF_STRUCT_OPS(rustland_enqueue, struct task_struct *p, u64 enq_flags)
 	 * potentially slow down the entire system if they are blocked for too
 	 * long (i.e., ksoftirqd/N, rcuop/N, etc.).
 	 */
-	if (is_kthread(p) && p->nr_cpus_allowed == 1) {
+	if (!full_user && is_kthread(p) && p->nr_cpus_allowed == 1) {
```
htejun, line 522: Out of curiosity, how stable is `full_user` mode? This opens up  other non-memory-allocation dependencies and the surface for long tail deadlock events may get a lot larger, I think. I suppose it really depends on how much the userland scheduler interacts with the kernel.

htejun (approved):

scheds/rust/scx_rustland/src/bpf/main.bpf.c
arighi: I wasn't sure to include this specific condition in `full-user` mode (usually I don't, but without this it wouldn't be "full user"...).

If we don't include this condition `full-user` mode seems to be pretty stable, at least for the tests that I've done. System is less responsive when it's mostly idle, of course, because tasks are not dispatched directly, but that is expected. And when the system is busy even the regular mode becomes "full-user" (no idle cpus => no direct dispatch => full-user), so no difference in this case.

Thinking more about this, we should probably avoid sending per-cpu kthread to the user-space scheduler and consider `full-user` just as a way to disable direct dispatch from BPF. In this way everything would be stable and good enough for testing this particular behavior.

arighi: Alright, updated `full-user` to exclude single-CPU kthread and make it more stable / reliable. I like it better.

PR #154, opened 2024-02-23
Topology refactor

Decave:
The topology.rs crate is insufficiently generic, and reflects
implementation details of scx_rusty more than it provides generic use
cases for modeling a host's topology. This adds a new topology crate
that replaces topology.rs, and also updates `scx_rusty` to use the new
create.


htejun (approved):

htejun:
> > Error: Failed to open or read file "/sys/devices/system/node/node0/cpu0/cpufreq/scaling_min_freq"
> 
> Do we need to compile with `CONFIG_CPU_FREQ`?

I guess, but maybe make that part optional?

htejun:
Looks great to me. Down the line it may be useful to have an ability to mock an arbitrary topology so that scheduler developers can test behaviors without actually having to have actual machines.

Decave:
> Looks great to me. Down the line it may be useful to have an ability to mock an arbitrary topology so that scheduler developers can test behaviors without actually having to have actual machines.

Definitely. It shouldn't be terribly difficult to do with a builder pattern. As you said, we can add it down the line when needed.

PR #155, opened 2024-02-23
scx_rustland: cpu topology refactoring

Decave:
The new topology crate allows us to replace the custom rustland topology
logic with the logic in the topology crate itself. Let's update rustland accordingly.

Note that this also requires us to add lookups from the root topology object to
all of the Cpu and Core objects in the system. No harm, no foul!


rust/scx_utils/src/topology.rs
```diff
@@ -193,14 +195,47 @@ impl Topology {
         let span = cpus_online()?;
         let nodes = create_numa_nodes(&span)?;
 
-        Ok(Topology { nodes, nr_cpus, span })
+        // For convenient and efficient lookup from the root topology object,
+        // create two BTreeMaps to the full set of Core and Cpu objects on the
+        // system. We clone the objects that are located further down in the
+        // hierarchy rather than dealing with references, as the entire
+        // Topology is read-only anyways. If we ever make the Topology dynamic
+        // with e.g. CPU hotplug, we'll have to change this.
```
htejun, line 203: I think one approach we can take here is just restarting the scheduler after an hotplug event. I think some embedded applications use hoptlug for power management but outside of that it's a really really cold path, so it's unlikely to be worthwhile to support it in a fully dynamic fashion. We can e.g. provide a scx_utils sched trait where individual schedulers implement load / init callbacks and then scx_utils code can just unload and reload the scheduler whenever topology changes so that all schedulers get hotplug function without much effort.

htejun (approved):

rust/scx_utils/src/topology.rs
Decave: Ahhh, great point, that does sound like a much better solution. Let me remove the mention of hotplug from the comment.

Decave:
I'll let @arighi take a look before I land.

arighi (approved):
Looks great, thanks!

PR #159, opened 2024-02-24
Add new infeasible.rs crate

Decave:
We want to avoid every scheduler implementation from having to implement
load balancing, but we also want to enable sufficient flexibility where
not every program has to have the same partition of scheduling domains,
etc. To enable this, a new load_balance crate is added which
encapsulates all of the logic for being given duty cycle and weight, and
performing the necessary math to adjust for infeasibility.

In addition, update `scx_rusty` to use the new crate.

### DIFF
--- a/rust/scx_utils/src/lib.rs
+++ b/rust/scx_utils/src/lib.rs
@@ -53,3 +53,7 @@ pub use topology::Node;
 
 mod cpumask;
 pub use cpumask::Cpumask;
+
+mod infeasible;
+pub use infeasible::LoadAggregator;
+pub use infeasible::LoadLedger;
--- a/rust/scx_utils/src/infeasible.rs
+++ b/rust/scx_utils/src/infeasible.rs
@@ -1,0 +1,459 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+//! # SCX Load Calculator
+//!
+//! A crate providing abstractions for calculating load between scheduling
+//! domains.
+//!
+//! Load Balancing Primer
+//! ---------------------
+//!
+//! Let's begin by doing a brief overview of load balancing. In order to
+//! understand load balancing, we'll introduce the relevant terms, and explain
+//! what load balancing is from there:
+//!
+//! *Weight*
+//!
+//! A positive integer value representing a task's priority in the system. The
+//! meaning of weight is ultimately up to the caller to determine and apply, but
+//! conceptually speaking weight is a linear scaling factor of a task's
+//! perceived load (defined below) in the system.
+//!
+//! *Duty Cycle*
+//!
+//! A task's duty cycle is the proportion of time that it could have used a CPU
+//! in some time window. For example, say that a task does nothing but spin in a
+//! tight loop for 10ms, and then sleep for 10ms. Then in a 20ms time window,
+//! its duty cycle would be 0.5.
+//!
+//! Note that this is not the same thing as the proportion of time it's
+//! _runnable_. For example, if you had two such tasks sharing a CPU, one of
+//! them may wait for 10ms to use the core while the other one runs, and then it
+//! would run for its 10ms duty cycle. It was runnable for 20ms, but could only
+//! actually use the CPU for 10ms, so its duty cycle was 10ms / 20ms = 0.5.
+//!
+//! *Load*
+//!
+//! A scheduling entity's load l_x is simply the product of its weight and duty
+//! cycle:
+//!
+//! l_x = w_x * d_x
+//!
+//! *Infeasible Weights*
+//!
+//! At a conceptual level, the goal of a load balancer is of course to balance
+//! load across the system. If a scheduling entity has load l_x, and the total
+//! sum of all entities' loads on the system is L, then entity x should receive
+//! l_x / L proportion of the available CPU capacity on the system.
+//!
+//! This formulation works fine in many cases, but can break down when an
+//! entity's proportion of load (and thus allotted CPU capacity) exceeds the
+//! amount of CPU it can consume.
+//!
+//! For example, say you were on a system with 2 sets of 2-core groups (for a
+//! total of 4 cores on the system), eight tasks with the following duty cycle
+//! and weight:
+//!
+//! ID          Weight          Duty Cycle              Load
+//!
+//! 0           1               1.0                     1
+//! 1           1               1.0                     1
+//! 2           1               1.0                     1
+//! 3           1               1.0                     1
+//! 4           1               1.0                     1
+//! 5           1               1.0                     1
+//! 6           1               1.0                     1
+//! 7           10000           1.0                     10000
+//!
+//!
+//! The load sum L of the system is 1 * 7 + 1 * 10000 = 10007. This means that
+//! tasks 0-6 have a load proportion of 1 / 10007 ~= 0.0001, and task 7 has a
+//! load proportion of 0.9993. In other words, tasks 0-6 are entitled to
+//! 0.0001 * 4 = 0.0004 CPUs worth of capacity, and task 7 is entitled to
+//! 0.9993 * 4 = 3.9972 CPUs worth of capacity.
+//!
+//! Task 7 can only consume at most 1.0 CPU due to its duty cycle being 1.0, so
+//! its weight is "infeasible" as the amount of CPU capacity it would be
+//! afforded exceeds what it can actually use.
+//!
+//! What we instead want is to find an adjusted weight w' that we can assign to
+//! Task 7 such that it gets its full duty cycle of 1.0, but allows the
+//! remaining 3 cores worth of compute to be equally spread amongst the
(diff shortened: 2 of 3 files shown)


rust/scx_utils/src/lib.rs
```diff
@@ -53,3 +53,6 @@ pub use topology::Node;
 
 mod cpumask;
 pub use cpumask::Cpumask;
+
+mod load_balance;
+pub use load_balance::LoadBalancer;
```
htejun, line 58: I  wonder whether the name is too broad. It's likely that scheduler implementations want to use the term load balancer for themselves and it can get confusing for utility create to occupy that name. Would it make sense to name it more narrowly? This is really just to solve the infeasible weight problem, right? Would it make sense to name it accordingly?

rust/scx_utils/src/load_balance.rs
```diff
@@ -0,0 +1,318 @@
...
+//! domain; adjusted for any infeasible weights.
+//!
+//! Note that the abstractions here are meant to be generic across schedulers.
+//! LoadAggregator assumes nothing about the scheduling domains being passed to
+//! it. For example, they may span layers defined in a scheduler, domains
+//! specified by the user via cpumask strings, or domains that span L3 caches.
+//! The crate simply aggregates load according to domain with IDs specified by
+//! the caller, computes global load sums, and adjusts any infeasible weights
+//! such that all infeasible scheduling entities still get their full duty
+//! cycle.
```
htejun, line 26: The last sentence may be misleading for people who aren't already familiar with the problem. I think it'd be better to either give a bit longer explanation (ie. is infeasible too high or low? is weight being adjusted up or down?) or just direct to the document.

rust/scx_utils/src/load_balance.rs
```diff
@@ -0,0 +1,318 @@
...
+//!
+//! Assume we're on a 16-core (32-CPU) host with two core complexes:
+//!
+//!```
+//!     let mut aggregator = LoadBalancer::aggregator(32, false);
```
htejun, line 31: Is this a common pattern? Wouldn't `LoadAggregator::new(32, false)` be more usual? Is it to lower the number of structs that users have to import?

rust/scx_utils/src/load_balance.rs
```diff
@@ -0,0 +1,318 @@
...
+const MIN_WEIGHT: usize = 1;
+const MAX_WEIGHT: usize = 10000;
+
+#[derive(Debug)]
+pub struct LoadBalancer {
+    dom_load_sums: Vec<f64>,
+    dom_dcycle_sums: Vec<f64>,
+    global_dcycle_sum: f64,
+    global_load_sum: f64,
+    w_prime: f64,
```
htejun, line 64: Would something like `effective_max_weight` be more intuitive?

rust/scx_utils/src/load_balance.rs
```diff
@@ -0,0 +1,318 @@
...
+            max_weight: 0,
+            global_dcycle_sum: 0.0f64,
+            global_load_sum: 0.0f64,
+            w_prime: 10000.0f64,
+            dcycle_only,
+        }
+    }
+
+    /// Return the global, host-wide sum of duty cycles.
+    pub fn global_dcycle_sum(&self) -> f64 {
```
htejun, line 84: Dumb question. What's the benefit of these accessors? Given that the embedding struct contains only data, what's the difference from just allowing users access the members?

rust/scx_utils/src/load_balance.rs
```diff
@@ -0,0 +1,318 @@
...
+            dom_load_sums,
+            dom_dcycle_sums,
+            global_dcycle_sum: self.global_dcycle_sum,
+            global_load_sum: self.global_load_sum,
+            w_prime: self.w_prime,
+        }
+    }
+
+    /// Record an instance of some domain's load (by specifying its weight and
+    /// dcycle). Returns an error if duty cycle is specified once for a given
```
htejun, line 164: "more than once"

rust/scx_utils/src/load_balance.rs
```diff
@@ -0,0 +1,318 @@
...
+#[derive(Debug)]
+pub struct LoadAggregator {
+    doms: BTreeMap<usize, Domain>,
+    global_loads: BTreeMap<usize, f64>,
+    nr_cpus: usize,
+    max_weight: usize,
+    global_dcycle_sum: f64,
+    global_load_sum: f64,
+    w_prime: f64,
+    dcycle_only: bool,
```
htejun, line 135: I wonder whether it'd be easier to use if the same struct contained the unadjusted and adjusted values or does it get unwieldy with prefixes? But, right now, if I want to compare the values before and after adjustments, I'd be comparing `load_aggregator.X` vs. `load_balancer.X` which isn't very intuitive. Wouldn't it be clearer if it's more explicit? e.g. maybe sth like `load_aggregator.in.X` vs `load_aggregator.eff.X`.

rust/scx_utils/src/load_balance.rs
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        if !self.global_loads.contains_key(&weight) {
+            self.global_loads.insert(weight, 0.0f64);
+        }
+        let mut weight_dcycle = self.global_loads.get(&weight).unwrap().clone();
+        weight_dcycle += dcycle;
+        self.global_loads.insert(weight, weight_dcycle);
+
+        if approx_eq(dcycle, 0.0f64) {
```
htejun, line 191: Why do we need this? Would this problematic in configurations which have a lot of weight slots (e.g. compound cgroup * task weights)?

rust/scx_utils/src/load_balance.rs
```diff
@@ -0,0 +1,318 @@
...
+//! document:
+//!
+//! https://drive.google.com/file/d/1fAoWUlmW-HTp6akuATVpMxpUpvWcGSAv
+
+use anyhow::bail;
+use anyhow::Result;
+use std::collections::BTreeMap;
+
+const MIN_WEIGHT: usize = 1;
+const MAX_WEIGHT: usize = 10000;
```
htejun, line 56: Do we need to limit the max weight?

htejun (approved):

rust/scx_utils/src/lib.rs
Decave: Yeah, let's rename it to `infeasible_calc.rs` or `infeasible_weights.rs`. I'm leaning towards `load_calc.rs` just because it also sums loads that can be used for under-utilized systems, etc. But yeah LoadBalance is too broad (as evidenced by the fact that I had to change the object name in rusty).

rust/scx_utils/src/load_balance.rs
Decave: My thinking was more so that `LoadAggregator` is only really useful in the context of creating a `LoadBalancer`. In my experience it's fairly normal to do this for builder patterns, but I'm fine with changing it if that's your preference.

rust/scx_utils/src/load_balance.rs
Decave: This way the struct can't be mutated. It felt like a bit of a tighter API to do it this way and force it to be read-only.

rust/scx_utils/src/load_balance.rs
Decave: When I was implementing infeasible weights in rusty I saw a lot of cases where we'd have a very, very, very low nonzero number that functionally had no effect on load, but still made us do extra work, and even worse, possibly have to deal with issues from finite precision like having a very small negative number when computing `lambda_x`.

rust/scx_utils/src/load_balance.rs
Decave: Hmm, I suppose not. I guess there's no reason for us to stop callers from implementing their own weighting scheme beyond just e.g. `cpu.weight`. Will adjust.

arighi:
> The CI failure is an `scx_qmap` timeout:
> 
> ```
> EXIT: runnable task stall (kworker/2:1[53] failed to run for 6.016s)
> [   18.812322] ACPI: PM: Preparing to enter system sleep state S5
> [   18.813521] kvm: exiting hardware virtualization
> [   18.813805] reboot: Power down
> FAIL: scx_qmap
> ```
> 
> `scx_qmap` is another pure example scheduler, so I think we can similarly put it in the no-timeout bucket.

I agree, and scx_userland as well. See #160 .

Decave:
@htejun pushed a number of changes, please feel free to take another look. The comments that I haven't directly addressed are still left unresolved above, but everything else should have been addressed so let me know if you see otherwise.

rust/scx_utils/src/load_balance.rs
htejun: Regardless of the optimization, the calculations should be safe. I'm not sure it'd be safe to depend on the threshold to protect against that.

htejun:
I still find the name confusing given that all it does is solving infeasible weights (or is the plan to expand on it?) and it feels like leaving generic names like LoadCalc to actual scheduler implementations would be better, but this is definitely a nitpick. Please feel free to merge.

htejun (approved):

PR #163, opened 2024-02-28
Add libbpf as a submodule

jordalgo:
This is to potentinally reduce issues with folks using different versions of libbpf at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ repo](https://mesonbuild.com/Quick-guide.html#installation-from-source) and call
   standard utilities including `awk`.
 - `clang`: >=16 required, >=17 recommended
 - `libbpf`: >=1.2.2 required, >=1.3 recommended (`RESIZE_ARRAY` support is
-  new in 1.3)
+  new in 1.3). It's preferred to link statically against the source from the libbpf git submodule.
 - Rust toolchain: >=1.72
 - `libelf`, `libz`, `libzstd` if linking against staic `libbpf.a`
 - `bpftool` (usually available in `linux-tools-common`)
@@ -210,9 +210,21 @@ repo](https://mesonbuild.com/Quick-guide.html#installation-from-source) and call
 commands in the root of the tree builds and installs all schedulers under
 `~/bin`.
 
+#### Static linking against libbpf submodule (preferred)
+
+ ```
+$ cd $SCX
+$ git submodule init
+$ make -C libbpf/src
+$ meson setup build --prefix ~
+$ meson compile -C build
+$ meson install -C build
+```
+
+#### Dynamic linking against libbpf
 ```
 $ cd $SCX
-$ meson setup build --prefix ~
+$ meson setup build --prefix ~ -D libbpf_a=disabled
 $ meson compile -C build
 $ meson install -C build
 ```
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libbpf"]
+	path = libbpf
+	url = https://github.com/libbpf/libbpf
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -210,6 +210,18 @@ repo](https://mesonbuild.com/Quick-guide.html#installation-from-source) and call
 commands in the root of the tree builds and installs all schedulers under
 `~/bin`.
 
+#### Static linking against libbpf submodule (preferred)
+
+ ```
+$ cd $SCX
+$ git submodule init
+$ make -C libbpf/src
+$ meson setup build --prefix ~ -D libbpf_a=$SCX/libbpf/src/libbpf.a
```
Decave, line 219: Should we make this the default behavior, and instead allow people to override `libbpf_a` to something else?

libbpf
Decave, line 1: Should we instead have the submodule track the latest release, which would be v1.3.0?

Decave (commented):
Nice, thanks! This will hopefully make things way less annoying for people. LGTM, just left a couple comments before stamping.

README.md
jordalgo: That makes sense. Will update

libbpf
jordalgo: I don't feel strongly but that makes sense

ptr1337:
Could be this directly added into meson/scx add provide a option to use statically or dynamically linked one?

README.md
Decave: Agreed. Ideally the build script can minimize the pain of working with submodules by hiding the details when possible.

jordalgo:
@ptr1337 Probably my ignorance of meson, but I don't quite understand your comment. Isn't the libbpf_a user option deciding if libbpf is statically linked or not?

ptr1337:
> @ptr1337 Probably my ignorance of meson, but I don't quite understand your comment. Isn't the libbpf_a user option deciding if libbpf is statically linked or not?

Seems like after the force push this option is included :) Seems good.

Decave:
> @ptr1337 Probably my ignorance of meson, but I don't quite understand your comment. Isn't the libbpf_a user option deciding if libbpf is statically linked or not?

I think what he's requesting is:
- Make static linking the default behavior
    - By default, statically link against the submodule libbpf
    - Else, allow user to specify another library to statically link against
 - Else, allow user to dynamically link against libbpf.so

jordalgo:
> > @ptr1337 Probably my ignorance of meson, but I don't quite understand your comment. Isn't the libbpf_a user option deciding if libbpf is statically linked or not?
> 
> I think what he's requesting is:
> 
> * Make static linking the default behavior
>   
>   * By default, statically link against the submodule libbpf
>   * Else, allow user to specify another library to statically link against. Honestly, not sure we even need this?
> * Else, allow user to dynamically link against libbpf.so

Cool. I think I did this in my latest changes - please correct me if I missed something.

The last thing I'll do is have meson do the `git submodule update --init --recursive` as @arighi suggested.

jordalgo:
Actually, can we land this as is and I'll add the `git submodule update --init --recursive` in a follow up PR (mostly so I can test a fresh clone).

jordalgo:
@Decave @htejun - Ok to merge?

README.md
```diff
@@ -210,13 +210,25 @@ repo](https://mesonbuild.com/Quick-guide.html#installation-from-source) and call
...
+$ make -C libbpf/src
 $ meson setup build --prefix ~
 $ meson compile -C build
 $ meson install -C build
 ```
 
+#### Dynamic linking against libbpf
+```
+$ cd $SCX
+$ meson setup build --prefix ~-D libbpf_a=disabled
```
Decave, line 227: Missing a space between ~ and -D

Decave (approved):
LGTM, if you wouldn't mind please fixing the one typo, I can merge

README.md
jordalgo: Fixed

PR #165, opened 2024-02-29
build: limit the maximum amount of parallel cargo build

arighi:
Each cargo build is already parallelized, spreading multiple rustc across all the available CPUs by default.

Allowing to run multiple instances of cargo at the same time doesn't provide any benefit and it can only increase the risk of triggering OOM conditions or overloading the build system.

Therefore, limit the amount of parallel cargo build instances to 1.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,6 +130,11 @@ endif
 cargo_env = environment()
 cargo_env.set('BPF_CLANG', bpf_clang.full_path())
 
+#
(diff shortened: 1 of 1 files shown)


Decave (approved):

htejun:
Wait, I'm not sure this actually does what you say. This doesn't limit the number of cargo instances from meson side. This limits the number of rustc instances that each cargo instance kicks off, right? This will make cargo builds really slow.

PR #167, opened 2024-02-29
Revert "Merge pull request #165 from sched-ext/reduce-rust-build-load"

htejun:
This reverts commit a7b39f24e2fa17297cdbef9a95f60bc54c6ddada, reversing changes made to cf7404fb03afc9a4c839825fb6692f31fac08cde.

The PR doesn't do what the description says. It instead limits the number of rustc instances to 1 for each cargo build making rust builds extremely slow. Let's revert and try again.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,11 +130,6 @@ endif
 cargo_env = environment()
 cargo_env.set('BPF_CLANG', bpf_clang.full_path())
 
-#
-# Limit the maximum amount of parallel builds for the Rust schedulers.
-#
-cargo_env.set('CARGO_BUILD_JOBS', '1')
-
 foreach flag: bpf_base_cflags
   cargo_env.append('BPF_BASE_CFLAGS', flag, separator: ' ')
 endforeach


arighi:
Hm... yeah that's definitely not what I wanted to do, will do more tests, sorry for the noise.

htejun:
No worries. Hopefully, there's a way to do this with meson.

htejun:
We're also having trouble integrating libbpf build. I'm half-way convinced that meson isn't a great option for this repo and might switch to something which is more expressive and less restrictive. If you have some ideas on which build system would be better, please share.

htejun:
So, possibly a better way of serializing rust builds is introducing serial dependencies across all of them so that e.g. scx_layered depends on scx_rustland which depends on scx_rusty and so on, so that the build system is forced to serialize them.

arighi:
I wish there was a better way than adding direct dependencies like that, but if that works I guess it can be a reasonable workaround.

If there was a way to set `--njobs=1` by default, unless specified otherwise, everything would work perfectly. I can't believe meson doesn't have something like that.

Anyway, I'll investigate a bit.

htejun:
`--njbos=1` is okay but not perfect in that it'd serialize compilation of all c scheds. They tend to be pretty small and we don't have many of them, so that's mostly okay, I guess.

PR #170, opened 2024-03-04
meson-scripts/build_libbpf: Accommodate meson setting CC to "ccache $COMPILER"

htejun:
Otherwise, we end up passing CC=ccache to libbpf's Makefile which triggers an error as ccache invoked on its own can't act as a stand-in for the compiler.


arighi (approved):
Tested and LGTM, thanks!

PR #172, opened 2024-03-05
Always build libbpf as a PIE

jordalgo:
This is to fix an error sometimes seen when compiling with gcc,
whereby the position independent sched ext rust libraries
don't play nicely with the libbpf library if it's not also built
as position independent.

This also explicitly sets the rustc relocation-mode to "pic",
which is the default (just so this doesn't accidentally change
out from under us).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -80,7 +80,7 @@ if should_build_libbpf
   run_command('git', 'submodule', 'foreach', '--recursive', 'git', 'clean', '-xfd', check: true)
 
   # setup the build target
-  executable('cc_cflags_probe', 'meson-scripts/cc_cflags_probe.c', install: false)
+  executable('cc_cflags_probe', 'meson-scripts/cc_cflags_probe.c', install: false, pie: true)
 
   make_jobs = 1
   if nproc.found()
@@ -199,7 +199,7 @@ if libbpf_a != ''
   endforeach
 
   cargo_env.append('RUSTFLAGS',
-                   '-C link-args=-lelf -C link-args=-lz -C link-args=-lzstd -L '
+                   '-C relocation-model=pic -C link-args=-lelf -C link-args=-lz -C link-args=-lzstd -L '
                    + fs.parent(libbpf_a))
 
   #


meson-scripts/build_libbpf
```diff
@@ -7,18 +7,27 @@ out=${out#\"}
...
     cc="$cc ${args[idx]}"
 fi
 
-declare -a cflags=()
+# This fixes an issue that was happening in some cases when building with gcc:
+# relocation R_X86_64_32 against `.rodata' can not be used when making a PIE
+# object; recompile with -fPIE
+if [ "$cc" = "cc" ]; then
+    cflags+="-fPIE "
+fi
```
Decave, line 24: I'm not understanding what the meaning is of `-fPIE` for a static library. My guess is that this implies `-fPIC` or something? Anyways, I think if we're going to do this we should just always build with `-fPIE`. Clearly it's something that rustc requires.

meson-scripts/build_libbpf
```diff
@@ -7,18 +7,27 @@ out=${out#\"}
...
+
+cflags+="-O2 "
 
 for arg in ${args[@]:(idx+1)}; do
     case $arg in
-	-I*|-M*|-o|-c) ;;
+	-I*|-M*|-O*|-o|-c) ;;
 	-*) cflags+="$arg ";;
     esac
 done
```
Decave, line 33: By the way, not sure it matters but I believe we should only need to build `libbpf.a` target

meson-scripts/build_libbpf
```diff
@@ -7,18 +7,27 @@ out=${out#\"}
...
 if [ "$cc" = "ccache" ]; then
     idx=$((idx+1))
     cc="$cc ${args[idx]}"
 fi
 
-declare -a cflags=()
+# This fixes an issue that was happening in some cases when building with gcc:
+# relocation R_X86_64_32 against `.rodata' can not be used when making a PIE
+# object; recompile with -fPIE
+if [ "$cc" = "cc" ]; then
```
htejun, line 22: I don't think this is a reliable way to detect gcc. meson has `compiler.has_argument()` test, so maybe test whether `-fPIE` is supported there and add that to build flags?

meson-scripts/build_libbpf
htejun: Reads to me it's position independent but can't be linked to create dynamic libs. No idea why it's needed.

meson-scripts/build_libbpf
htejun: Is there a reason for excluding -O? Why don't we want to be synced with how other parts of scx is built?

meson-scripts/build_libbpf
Decave: Yeah, I've just never seen `-fPIE` used to create a lib. I thought it was always `-fPIC` given that it's not creating an executable.

meson-scripts/build_libbpf
Decave: Or let's just always build with `-fPIE` to minimize the build complexity / variance?

jordalgo:
Limited the scope of these changes. Luckily meson has a "pie" option for compiling executables so we'll automatically get the right flag for the selected compiler (tested with both clang and gcc).

jordalgo:
> Is there a reason for excluding -O? Why don't we want to be synced with how other parts of scx is built?

@Decave Mentioned that we should always be compiling libbpf with -O2 -- I guess the question is - why don't we do this by default for our other exes/libs?

htejun (approved):

htejun:
> > Is there a reason for excluding -O? Why don't we want to be synced with how other parts of scx is built?
> 
> @Decave Mentioned that we should always be compiling libbpf with -O2 -- I guess the question is - why don't we do this by default for our other exes/libs?

Because optimized builds are slow, debugging can be painful and should by controlled by setting buildtype appropriately. I don't know whether we can make meson default to release builds, which would be fine, but it'd be annoying if the buildtype is set to debug and everything else is built accordingly but just libbpf isn't.

Decave:
> > > Is there a reason for excluding -O? Why don't we want to be synced with how other parts of scx is built?
> > 
> > 
> > @Decave Mentioned that we should always be compiling libbpf with -O2 -- I guess the question is - why don't we do this by default for our other exes/libs?
> 
> Because optimized builds are slow, debugging can be painful and should by controlled by setting buildtype appropriately. I don't know whether we can make meson default to release builds, which would be fine, but it'd be annoying if the buildtype is set to debug and everything else is built accordingly but just libbpf isn't.

Well, the counter-argument to that is that libbpf is just a dependency the same as any other whose build type we don't control. In my mind, having libbpf as a submodule was more so meant to be a convenience to avoid versioning issues than a way to control how we compile libbpf. Given how closely it interfaces with the scheduler though.

PR #178, opened 2024-03-08
rusty: Implement NUMA-aware load balancing

Byte-Lab:
Right now, scx_rusty has no notion of domains spanning NUMA nodes, and makes no
distinction when making load balancing decisions, or work stealing. This can
cause problems on multi-NUMA machines, as load balancing and work stealing
across NUMA nodes has significantly different cost from across L3 cache
boundaries.

In order to better support multi-NUMA machines, this commit adds another layer
to the rusty load balancer, which balances across NUMA nodes using a different
cost function from balancing across domains. Load balancing now takes place
over the span of two passes:

1. In the first pass, we fix imbalances across NUMA nodes by moving tasks
   between domains across those NUMA node boundaries. We require a load
   imbalance of at least 17% in order to move load at this stage. The ratio of
   load imbalance we attempt to adjust (50%) and the maximum amount of load
   we're allowed to push out of a domain (50%) is still the same as when
   balancing between domains inside a NUMA node, but this is easy to tune with
   the current setup.

2. Once we've balanced across NUMA nodes, we iterate over all nodes and balance
   between the domains within each NUMA node. The cost function here is the
   same as what it has been thus far: we require at least a 5% imbalance in
   order to trigger load balancing.

There are a few additional changes / improvements to load balancing in this
commit:

1. NUMA nodes and domains are now ordered according to their load by using
   SortedVec objects. We were previously using BTreeMap keyed by load, but this
   was suboptimal due to the fact that it doesn't allow duplicate entries.

2. We're no longer exporting load balancing statistics as a vector of data such
   as load sums, averages, and imbalances. This is instead all encapsulated in
   the load balancing hierarchy we setup in lb.load_balance(). These statistics
   are not yet exported, but they will be in a subsequent commit.

One of the issues with this commit is that it does introduce some
almost-identical logic that somehow begs to be deduplicated. For example, when
we balance between NUMA nodes, the logic for iterating over push nodes and
pushing to pull nodes is very similar to the logic of iterating over push
domains and pull domains when balancing within a node. It may be that this can
be improved.

The following are some benchmarks run on an Intel Xeon Gold 6138 (2 x 40 core
processor):

kcompile
--------

On Commit a27648c74210 ("afs: Fix setting of mtime when creating a
file/dir/symlink"):

1. make allyesconfig
2. make -j $(nproc) built-in.a
3. make -j clean
4. goto 2

Runtime
-------

         o-----------o-----------o----------o
         | scx_rusty |     CFS   |   Delta  |
---------o-----------o-----------o----------o
Mean     | 562.688s  | 566.085s  | -.6%     |
---------o-----------o-----------o----------o
Variance | 0.54387   | 0.72431   | -24.9%   |
---------o-----------o-----------o----------o

         o-----------o-----------o----------o
         | rusty NUMA| rusty ORIG|   Delta  |
---------o-----------o-----------o----------o
Mean     | 562.688s  | 563.209s  | -.092%   |
---------o-----------o-----------o----------o
Variance | 0.54387   | 0.42038   | 29.38%   |
---------o-----------o-----------o----------o

scx_rusty with NUMA awareness clearly beats CFS, but only barely beats
scx_rusty without it. This isn't necessarily super surprising given that
this is kcompile, which has very poor front-end CPU locality. Further
experimentation with toggling the cost function for performing
migrations may improve this further.

CPU util
--------

         o-----------o-----------o----------o
         | scx_rusty |     CFS   |   Delta  |
---------o-----------o-----------o----------o
Mean     | 7654.25%  | 7551.67%  | 1.11%    |
---------o-----------o-----------o----------o
Variance | 165.35714 | 158.3333  | 4.436%   |
---------o-----------o-----------o----------o

         o-----------o-----------o----------o
         | rusty NUMA| rusty ORIG|   Delta  |
---------o-----------o-----------o----------o
Mean     | 7654.25%  | 7641.57%  | 0.1659%  |
---------o-----------o-----------o----------o
Variance | 165.35714 | 1230.619  | -86.5%   |
---------o-----------o-----------o----------o

As expected, CPU util is quite a bit higher with scx_rusty than it is
with CFS. Further experiments that could be interesting are always
enabling direct-greedy stealing between domains within a NUMA node, and
then comparing rusty NUMA and rusty ORIG. rusty NUMA prevents stealing
between NUMA nodes, so this would show whether the locality introduced
by NUMA awareness appropriately offsets the loss of work conservation.

Major PFs
---------

         o-----------o-----------o----------o
         | scx_rusty |     CFS   |   Delta  |
---------o-----------o-----------o----------o
Mean     | 5332      | 3950      | 36.566%  |
---------o-----------o-----------o----------o
Variance | 6975.5    | 5986.333  | 16.5237% |
---------o-----------o-----------o----------o

         o-----------o-----------o----------o
         | rusty NUMA| rusty ORIG|   Delta  |
---------o-----------o-----------o----------o
Mean     | 5332      | 5336.5    | -.084%   |
---------o-----------o-----------o----------o
Variance | 6975.5    | 955.5     | 630.03%  |
---------o-----------o-----------o----------o

Also as expected, major page faults are far highe higher with scx_rusty
than with CFS. This is expected even with NUMA awareness, given that
scx_rusty is still less sticky than CFS.

Further experiments that could be interesting are tuning the threshold
for which we perform x NUMA migrations to try and keep this value even
lower. The rate of major page faults between rusty NUMA and rusty ORIG
were very close, though rusty NUMA was a bit lower.


scheds/rust/scx_rusty/src/load_balance.rs
```diff
@@ -0,0 +1,1082 @@
...
+
+        let push_imbal = push_node.load.imbal();
+        let pull_imbal = pull_node.load.imbal();
+        let xfer = push_node.xfer_between(&pull_node);
+
+        let mut delta = 0.0f64;
+        let mut push_doms = std::mem::take(&mut push_node.push_domains).into_vec();
+        for push_dom in push_doms.iter_mut().rev() {
+            let mut pull_doms = std::mem::take(&mut pull_node.pull_domains).into_vec();
+            for pull_dom in pull_doms.iter_mut() {
```
htejun, line 920: Does this mean that the amount that can be balanced across nodes is limited by the existing internal imbalance in the participating nodes? If so, this may be okay in practice as there usually would be on-going perturbance in the system but there can be pathological cases - e.g. the same number of full duty-cycle high weight threads per CPU on the source or target node. Given that the imbalance is between whole nodes, the right thing to do probably is to transfer from the highest load domain to the lowest one?

htejun (approved):

scheds/rust/scx_rusty/src/load_balance.rs
Byte-Lab: >Does this mean that the amount that can be balanced across nodes is limited by the existing internal imbalance in the participating nodes?

Sorry, I'm not sure I understand the question. Imbalance at the node level is calculated according to the per-node average. The per-domain average doesn't come into play here beyond us only being able to transfer tasks if they help equalize the imbalance, just like with cross-domain.

>Given that the imbalance is between whole nodes, the right thing to do probably is to transfer from the highest load domain to the lowest one?

That's the idea of iterating over the ordered push domains and pull domains. Or are you suggesting we just always iterate over highest load domain -> lowest load domain instead of doing two passes where we do NUMA <-> NUMA, followed by intra-NUMA (and applying a different cost for going cross-NUMA vs. just cross-domain)?

The issue with doing that is the imbalance could be solved by migrating between domains in the same node. We don't want to push to domains whose node-wide load is not imbalanced with respect to the global node load average.

Byte-Lab: Spoke offline, I see what you're saying. We could have for example two nodes with load imbalance between them, but each having domains with exactly equal load. It's not clear whether we'd correctly partition the domains into push or pull lists in such a scenario.

Let me rejigger this to just have a node own a flat list of domains, rather than a tuple of pushers, pullers, and balancers.

Byte-Lab:
@htejun / @arighi Addressed your feedback items. I also added another flag to enable tuning a cross-NUMA greedy task stealing threshold.

arighi (approved):

scheds/rust/scx_rusty/src/load_balance.rs
```diff
@@ -0,0 +1,1045 @@
...
+    }
+
+    fn transfer_between_nodes(&mut self,
+                              push_node: &mut NumaNode,
+                              pull_node: &mut NumaNode) -> Result<f64> {
+        debug!("Inter node {} -> {} started",
+               push_node.id, pull_node.id);
+
+        let push_imbal = push_node.load.imbal();
+        let pull_imbal = pull_node.load.imbal();
```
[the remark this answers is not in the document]
Byte-Lab: Hmm, but the node's imbalance is relative to the load average across all NUMA nodes. See https://github.com/sched-ext/scx/blob/3e0a17fc9abec9ba3c45cdb6fce95b98387f84cb/scheds/rust/scx_rusty/src/load_balance.rs#L580. The balance or imbalance of internal domains shouldn't affect anything here, including when we actually migrate the task, as the imbalance we're comparing against when deciding whether migrating a task will improve things or not is the imbalance between the two nodes, not the imbalance between the domains pushing and pulling the node.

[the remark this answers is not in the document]
htejun: I was misreading the code. The imbal calculation is against numa nodes, not domains. This looks correct to me. Sorry about the noise.

Byte-Lab:
Ok, cleaned the code up a bit more and added some comments to explain what we're doing. I'm going to merge this as is so we can iterate on future changes.

PR #179, opened 2024-03-08
Fetch and build bpftool by default

jordalgo:
This pairs with the new default behavior to fetch and build libbpf and is mostly being used so we can use the latest bpftool and libbpf.


htejun (approved):

jordalgo:
@Byte-Lab - Updated based on your feedback - now we always fetch libbpf for bpftool via submodule update

htejun (approved):

PR #185, opened 2024-03-12
scx_layered: Implement layer properties `exclusive` and `min_exec_us`

htejun:
- `exclusive`: A task in an exclusive layer occupies the whole core when it executes.
- `min_exec_us`: Minimum execution time a task in a given layer is charged. This can be used to penalize tasks which keeps waking up and going immediately back to sleep without doing much.


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -399,7 +417,7 @@ s32 BPF_STRUCT_OPS(layered_select_cpu, struct task_struct *p, s32 prev_cpu, u64
 
 	/* not much to do if bound to a single CPU */
 	if (p->nr_cpus_allowed == 1) {
-		if (!bpf_cpumask_test_cpu(cpu, layer_cpumask))
+		if (!bpf_cpumask_test_cpu(prev_cpu, layer_cpumask))
```
dschatzberg, line 420: I'm kind of surprised verifier didn't choke on this previously - isn't this fixing a call with an uninitialized variable?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -476,16 +494,28 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
...
+		     sib_cctx->current_preempt)) {
+			lstat_inc(LSTAT_EXCL_COLLISION, layer, cctx);
+			continue;
+		}
 
 		scx_bpf_kick_cpu(cpu, SCX_KICK_PREEMPT);
+		if (sib_cctx && !sib_cctx->maybe_idle) {
+			lstat_inc(LSTAT_EXCL_PREEMPT, layer, cctx);
+			scx_bpf_kick_cpu(sib, SCX_KICK_PREEMPT);
+		}
```
dschatzberg, line 518: This whole logic is a bit hard to parse. Some comments could help:

1) `sib_cctx` is only set if `layer->exclusive` so this second condition is really the same as `layer->exclusive && ...`
2) `!sib_cctx->maybe_idle` is still a bit hard for me to follow. I guess the idea here is to kick if there's any chance the sibling could be running something.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -494,14 +524,27 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
...
 	int idx;
 
+	if (!(cctx = lookup_cpu_ctx(-1)))
+		return;
+
+	if (sib >= 0 && (sib_cctx = lookup_cpu_ctx(sib)) &&
+	    sib_cctx->current_exclusive) {
+		gstat_inc(GSTAT_EXCL_IDLE, cctx);
+		return;
+	}
```
dschatzberg, line 546: Isn't this all racy? Does it just not matter or is there some synchronization I'm missing

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -675,31 +718,61 @@ void BPF_STRUCT_OPS(layered_running, struct task_struct *p)
...
+
+	cctx->layer_cycles[lidx] += used;
 	cctx->current_preempt = false;
+	cctx->prev_exclusive = cctx->current_exclusive;
+	cctx->current_exclusive = false;
 
 	/* scale the execution time by the inverse of the weight and charge */
 	p->scx.dsq_vtime += used * 100 / p->scx.weight;
+	cctx->maybe_idle = true;
```
dschatzberg, line 775: Why is this maybe_idle? Just because it's racy to access from other cores?

dschatzberg (approved):
This all looks reasonable to me, just some of the bpf-side logic is a little tricky to parse, see questions inline

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: Yes, it is fixing an uninitialized use bug, and yeah, no idea why BPF didn't bark on it. That said, clang was generating warnings on it. We just didn't see them because we only show clang messages when compilation fails. I don't know how but we really should fix that.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: 1. Yeah, so testing for `sib_cctx` is, ignoring the tests which cannot fail but have to be there for verifier, the same as testing `exclusive && if the CPU has a sibling`.
2. If this task is exclusive, we want to make sure the sibling CPU is idle. so if the other CPU is not idle, I kick it w/ PREEMPT so that it expires the current task and enters the dispatch path, where it will see that I'm running an exclusive task and return without dispatching anything forcing it into idle.

Will add comments.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: It is very racy. The mechanism doesn't provide any hard guarantee and just provides best-effort optimization, so should hopefully be okay? It does change the scheduling behavior substantially, so it looks like the races aren't hurting too much. I'll add a comment.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: Yeah, it's not just racy but also inaccurate because it gets asserted between `stopping()` of the prev task and `running()` of the next task. It's a good enough but not great approximation.

scheds/rust/scx_layered/src/bpf/main.bpf.c
Byte-Lab: Could we make warnings fatal?

htejun: I'm generally not a big fan of making warnings fatal because compilers sometimes aren't great and it makes build flaky for silly reasons like compiler generating spurious warnings in some releases but maybe clang is better in this regard? So, yeah, we can do that but being able to report warnings as warnings would be better.

Byte-Lab: I've never had that experience myself (at least, when using -Wall; -Weverything is a different story), but I expect that's because I haven't been using bleeding-edge compiler releases like we are for this project. That said, IMO it will end up saving us more time / issues in the long term to treat warnings as failures. Consider that even if a warning is spurious that we'll probably want to address it so that the build output is clean. Otherwise, they might just start to look like noise, no?

htejun: gcc more than once introduced warnings which triggered obviously spuriously which then got cleaned up in later point releases and `-Werror` can push issues which can just be nuisances to actual downstream breakages. It's not a big issue, just not ideal. So, if there's no good way to make the build process dump clang warnings, `-Werror` is better than the current situation.

Issue #147, opened 2024-02-21
Unable to start scx_rusty after topology commit


kode54:
I am unable to start scx_rusty after https://github.com/sched-ext/scx/commit/8aba090d4f18d760f8a7a3adf1da4b5652f48de8 because it thinks I have 16 cores, 32 threads, when I actually have 8 cores, 16 threads:

```
> sudo scx_rusty
09:21:27 [INFO] CPUs: online/possible = 16/32
thread 'main' panicked at /usr/src/debug/scx-scheds-git/scx/rust/scx_utils/src/topology.rs:218:21:
Failed to open or read core_id file "/sys/devices/system/cpu/cpu16/topology/core_id"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ptr1337:
> I am unable to start scx_rusty after [8aba090](https://github.com/sched-ext/scx/commit/8aba090d4f18d760f8a7a3adf1da4b5652f48de8) because it thinks I have 16 cores, 32 threads, when I actually have 8 cores, 16 threads:
> 
> ```
> > sudo scx_rusty
> 09:21:27 [INFO] CPUs: online/possible = 16/32
> thread 'main' panicked at /usr/src/debug/scx-scheds-git/scx/rust/scx_utils/src/topology.rs:218:21:
> Failed to open or read core_id file "/sys/devices/system/cpu/cpu16/topology/core_id"
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

Old issue, this is an issue in the linux kernel but the topology patch probably breaks the old fix for it.

See here:
https://github.com/sched-ext/sched_ext/issues/69#issuecomment-1796172180

Workaround is to boot with nr_cpus=16

arighi:
Hm.. interesting, I'm wondering if this affects also functions like `scx_bpf_pick_idle_cpu()`, `scx_bpf_pick_any_cpu()`, and similar... like could they return a cpu id that doesn't really exist?

ptr1337:
> I'll address this soon. We shouldn't have regressed rusty from the refactor. I needed to take care of something internally for Meta, but I should have the bandwidth to fix up Topology more completely soon.

I think it would be generally also important to have a fixup for the general issue instead of workaround that issue with sched-ext/scx-rusty.
The AMD Devs should really look into fixing that issue, since it could have also a bunch of other interactions with internal kernel configurations

Byte-Lab:
[...]

PR #186, opened 2024-03-13
rusty: Account for disabled but offline CPUs

Byte-Lab:
As described in https://bugzilla.kernel.org/show_bug.cgi?id=218109, https://github.com/sched-ext/scx/issues/147 and
https://github.com/sched-ext/sched_ext/issues/69, AMD chips can sometimes report fully disabled CPUs as offline, which causes us to count them when looking at /sys/devices/system/cpu/possible.

While it may cause the scheduler to make suboptimal scheduling decisions (by e.g. mis-calculating average domain loads), we should still not crash the scheduler when we try to load it. Let's instead just emit a warning, and carry on.

Should address https://github.com/sched-ext/scx/issues/147

### DIFF
--- a/scheds/rust/scx_rusty/src/main.rs
+++ b/scheds/rust/scx_rusty/src/main.rs
@@ -35,6 +35,7 @@ use libbpf_rs::skel::OpenSkel as _;
 use libbpf_rs::skel::Skel as _;
 use libbpf_rs::skel::SkelBuilder as _;
 use log::info;
+use log::warn;
 use scx_utils::init_libbpf_logging;
 use scx_utils::uei_exited;
 use scx_utils::uei_report;
@@ -262,10 +263,25 @@ impl<'a> Scheduler<'a> {
         skel.rodata_mut().nr_doms = domains.nr_doms() as u32;
         skel.rodata_mut().nr_cpus = top.nr_cpus() as u32;
 
+        let mut warned = false;
         for cpu in 0..top.nr_cpus() {
-            let dom_id = domains.cpu_dom_id(&cpu).unwrap();
-            skel.rodata_mut().cpu_dom_id_map[cpu] =
-                dom_id.clone().try_into().expect("Domain ID could not fit into 32 bits");
+            if let Some(dom_id) = domains.cpu_dom_id(&cpu) {
+                skel.rodata_mut().cpu_dom_id_map[cpu] =
+                    dom_id.clone().try_into().expect("Domain ID could not fit into 32 bits");
+            } else {
+                // As described in
+                // https://bugzilla.kernel.org/show_bug.cgi?id=218109, some
+                // chips may incorrectly report disabled CPUs as offline. It
+                // doesn't appear that this is going to be corrected anytime
+                // soon, so just put the CPUs in domain 0 so we can get past
+                // startup.
+                if !warned {
+                    warn!("Found a CPU without a domain, putting in domain 0. May experience suboptimal scheduling.");
+                    warn!("Please see https://bugzilla.kernel.org/show_bug.cgi?id=218109 for more details");
+                    warned = true;
+                }
+                skel.rodata_mut().cpu_dom_id_map[cpu] = 0;
+            }
         }
 
         for numa in 0..domains.nr_nodes() {


scheds/rust/scx_rusty/src/main.rs
```diff
@@ -262,10 +263,25 @@ impl<'a> Scheduler<'a> {
...
+            } else {
+                // As described in
+                // https://bugzilla.kernel.org/show_bug.cgi?id=218109, some
+                // chips may incorrectly report disabled CPUs as offline. It
+                // doesn't appear that this is going to be corrected anytime
+                // soon, so just put the CPUs in domain 0 so we can get past
+                // startup.
+                if !warned {
+                    warn!("Found a CPU without a domain, putting in domain 0. May experience suboptimal scheduling.");
+                    warn!("Please see https://bugzilla.kernel.org/show_bug.cgi?id=218109 for more details");
```
htejun, line 280: Given how widespread this is, I'm not sure warning & possibly causing suboptimal behavior can be a solution. If our plan to handle hotplugs is by reloading scheduler, maybe the topology code can always just report online cpus and we can ignore the discrepancies between possible and online CPUs in the schedulers?

arighi:
I'm wondering if we should address this issue in a more generic way. For instance, `scx_bpf_select_cpu_dfl()` can also return offline CPUs, that may lead to sub-optimal scheduling decisions (maybe we should add `cpu_online(cpu)` checks to the idle selection logic?).

Moreover, in many places we're just relying on `libbpf_num_possible_cpus()`, that also includes offline cpus. I think it'd be nicer to rely on a list/mask of online cpus instead. Opinions?

htejun:
Oh, `scx_bpf_select_cpu_dfl()` returning !online CPU is an oversight. Let me fix that.

htejun:
So, https://github.com/sched-ext/sched_ext/pull/161 should improve the situation a bit but `scx_bpf_select_cpu_dfl()` could return a possible but offline CPU at most once after the scheduler is loaded, so this shouldn't have meaningful impact one way or the other.

For the most part, it'd be better to make cpu hotplug as transparent as possible to schedulers but I'm afraid they'd still need to understand possible and online CPUs. The use of `libbpf_num_possible_cpus()` is likely incorrect because it returns the weight of the possible cpumask rather than the maximum possible index and we're assuming that they match.

It's possible to have a system which has holes in possible CPU map. e.g. A system can have possible CPUs 0, 1, 4, 5 and online 0, 4. In this case, `libbpf_num_possible_cpus()` will return 4 but if the scheduler operates under the assumption that the maximum CPU ID is 3, it will break. In kernel the distinction is between `num_possible_cpus()` and `nr_cpu_ids`.

So, I think what we need to do is sth like:

1. Provide `nr_cpu_ids` to userland code and replace `libbpf_num_possible_cpus()` usages with it.
2. Teach the topology code to distinguish possible and online CPUs and allow schedulers to just consider the online CPUs if they want.
3. Update schedulers so that they ignore possible cpumask and can correctly operate with sparse online cpumask.
4. Build framework to update the topo info and reload the scheduler after a CPU hotplug event.

arighi:
>     1. Provide `nr_cpu_ids` to userland code and replace `libbpf_num_possible_cpus()` usages with it.

The topology crate (Rust) is already able to figure out the online cpus (fetching the info from `/sys/devices/system/cpu/online`), maybe we should move this (or part of this) logic to a C library (libscx)?

In this way the topology crate and the C schedulers can use the library and we could implement this in just one place, avoiding C/Rust code duplication, without exposing additional info to userland.

Byte-Lab:
>@arighi agreed we don't want to duplicate logic, but perhaps we could just add C projections and leave the logic in rust?

As another point, this is probably not the last piece of logic we'll want to share between C and rust. For example, we may also want to share the infeasible weights calculation logic. So it might make sense to invest the time now in figuring out how to add C projections wrapping the rust stuff.

arighi:
@Byte-Lab right, I think we're heading toward this concept of shared code between C and Rust, so in perspective I really think we should invest some time on such shared library, code, or similar.

But for now I think it's fine to address the specific issues. Later today I'll work on the rustland part to address this issue.

htejun:
I like the C projection idea so that we can keep all the actual implementations in rust. Given that all our practical schedulers have rust userland, adding it isn't a pressing issue, but yeah, it'd be great if C scheds can use the stuff in scx_utils.

Byte-Lab:
Going to rework this to remove the reliance on nr_possible for everything other than stats collection (which uses percpu maps).

PR #187, opened 2024-03-13
scx_layered: Make config json assume default vaules for unspecified fields

htejun:
This makes writing configs and allows introducing new fields without breaking existing configs.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -302,20 +302,29 @@ enum LayerMatch {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 enum LayerKind {
     Confined {
-        cpus_range: Option<(usize, usize)>,
         util_range: (f64, f64),
+	#[serde(default)]
+        cpus_range: Option<(usize, usize)>,
+	#[serde(default)]
         min_exec_us: u64,
     },
     Grouped {
-        cpus_range: Option<(usize, usize)>,
         util_range: (f64, f64),
+	#[serde(default)]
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/main.rs
```diff
@@ -302,20 +302,29 @@ enum LayerMatch {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 enum LayerKind {
     Confined {
-        cpus_range: Option<(usize, usize)>,
         util_range: (f64, f64),
```
Byte-Lab, line 305: I guess there's no way to specify for f64 tuples? If not, would it work to split them into separate fields?

htejun: It's more that there's no good default value to use given that that's the main parameter determining the sizing of the layer.

Byte-Lab: It's not a strictly necessary parameter though, is it? You could use `cpus_range` and leave `util_range` entirely out of the equation.

htejun: Hmm... so the only case where we don't need util_range is if `cpus_range` is specifying the exact number of CPUs - ie. it's specifying the same min and max. I know that web is using that currently but that's unlikely to be a common use case. The design is more that `util_range` dynamically determines the layer sizes and with optional `cpus_range` confinement as safeguard.

Another way to look at it is that for all other values the default values are obvious - no explicit cpu count limits and no extra features turned on, but what default values would be appropriate for `util_range`?

Byte-Lab (approved):

PR #188, opened 2024-03-14
topology: support single CPU systems

arighi:
We are failing to parse /sys/devices/system/cpu/online in systems with just one CPU, for example:

 $ vng -r --cpus 1 -- scx_rusty
 Error: Failed to parse online cpus 0

Correctly handle strings containing only a single CPU during parsing.

Fixes: c5a3b83b ("topology: Add new topology crate")

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -275,8 +275,13 @@ fn cpus_online() -> Result<Cpumask> {
         let (min, max) = match sscanf!(group.trim(), "{usize}-{usize}") {
             Ok((x, y)) => (x, y),
             Err(_) => {
-                bail!("Failed to parse online cpus {}", group.trim());
-            }
+                match sscanf!(group.trim(), "{usize}") {
+                    Ok(x) => (x, x),
+                    Err(_) => {
+                        bail!("Failed to parse online cpus {}", group.trim());
+                    }
+                }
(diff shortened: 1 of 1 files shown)


Byte-Lab (approved):
Thanks for the fix!

rust/scx_utils/src/topology.rs
```diff
@@ -275,8 +275,13 @@ fn cpus_online() -> Result<Cpumask> {
...
-            }
+                match sscanf!(group.trim(), "{usize}") {
+                    Ok(x) => (x, x),
+                    Err(_) => {
+                        bail!("Failed to parse online cpus {}", group.trim());
+                    }
+                }
+            },
         };
         for i in min..max {
```
Byte-Lab, line 286: I think we'll also need https://github.com/sched-ext/scx/pull/191/commits/583696f9402f0ae66834b8164c274dc1a65d77f4 right?

Byte-Lab: Just pushed so should be able to just rebase

arighi: > I think we'll also need [583696f](https://github.com/sched-ext/scx/commit/583696f9402f0ae66834b8164c274dc1a65d77f4) right?

Indeed, thanks for the extra fix! :)

PR #189, opened 2024-03-14
scx_rustland: mitigate sub-optimal performance with offline CPUs

arighi:
Most of the schedulers assume that the amount of possible CPUs in the system represents the actual number of CPUs available.

This is not always true: some CPUs may be offline or certain CPU models (AMD CPUs for example) may include unavailable CPUs in this number.

This can lead to sub-optimal performance or even errors in the scheduler (see for example [1][2]).

Ideally, we need to attack this issue in a more generic way, such as having a proper API provided by a C library, that can be used by all schedulers and the topology Rust module (scx_utils crate).

But for now, let's try to mitigate most of the common sub-optimal cases separately inside each scheduler.

For rustland we can apply some mitigations both in select_cpu() (for the BPF part) and in the user-space part:

 - the former is fixed in the sched-ext kernel by commit 94dc0c01b957 ("scx: Use cpu_online_mask when resetting idle masks"). However, adding an extra check `cpu < num_possible_cpus` in select_cpu(), allows to properly support AMD CPUs, even with kernels that don't have the cpu_online_mask fix yet (this doesn't always guarantee the validity of cpu, but it should be enough to mitigate the majority of the potential sub-optimal cases, without introducing any significant overhead)

 - the latter can be fixed relying on topology.span(), instead of topology.nr_cpus(), to count the amount of available CPUs in the system.

[1] https://github.com/sched-ext/sched_ext/issues/69
[2] https://github.com/sched-ext/scx/issues/147

Link: https://github.com/sched-ext/sched_ext/commit/94dc0c01b95709316fa590c862ca6ed221e2d6c4

### DIFF
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -264,14 +264,15 @@ impl<'a> Scheduler<'a> {
         let init_page_faults: u64 = 0;
 
         // Low-level BPF connector.
+        let nr_online_cpus = topo.span().weight() + 1;
         let bpf = BpfScheduler::init(
             opts.slice_us,
-            topo.nr_cpus() as i32,
+            nr_online_cpus as i32,
             opts.partial,
             opts.full_user,
             opts.debug,
         )?;
-        info!("{} scheduler attached", SCHEDULER_NAME);
+        info!("{} scheduler attached - {} online CPUs", SCHEDULER_NAME, nr_online_cpus);
 
         // Return scheduler object.
         Ok(Self {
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rustland/src/main.rs
```diff
@@ -264,14 +264,15 @@ impl<'a> Scheduler<'a> {
         let init_page_faults: u64 = 0;
 
         // Low-level BPF connector.
+        let nr_online_cpus = topo.span().weight() + 1;
```
Byte-Lab, line 267: FYI, there's a small bug in topo where it computes span, that I'm about to send a fix for. Once that's addressed, you shouldn't need to do + 1 here.  I'll merge this and update your code here as part of that fix.

arighi: > FYI, there's a small bug in topo where it computes span, that I'm about to send a fix for. Once that's addressed, you shouldn't need to do + 1 here. I'll merge this and update your code here as part of that fix.

I actually seemed like a bug, but I wasn't sure, thanks for clarifying it!

Byte-Lab: https://github.com/sched-ext/scx/pull/191 should fix it

PR #192, opened 2024-03-16
scx_lavd: add LAVD (Latency-criticality Aware Virtual Deadline) scheduler

multics69:
Provide a deadline-based scheduler focused on latency-critical tasks inspired by gaming workloads. scx_lavd quantifies how much a task is latency-critical and leverages the task's latency-criticality information in making various scheduling decisions (e.g., task's execution deadline, time slice, etc). It consists of rust and BPF parts. The rust part simply loads the BPF scheduler and conducts the chores (e.g., printing statistics, etc). The BPF part makes all the scheduling decisions.


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	 * task should be scheduled. It is determined by two factors: how
+	 * urgent it is - vdeadline_delta_ns - and when it becomes eligible if
+	 * overscheduled - eligible_time_ns.
+	 */
+	calc_virtual_dealine_delta(p, taskc, enq_flags);
+	calc_eligible_delta(p, taskc);
+}
+
+static bool put_local_rq(struct task_struct *p, struct task_ctx *taskc,
+			 s32 cpu_id, u64 enq_flags)
```
arighi, line 1253: Do we really need to pass cpu_id in `put_local_rq()`? Being a local dispatch it shouldn't require a target cpu right?

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    fn running(&mut self) -> bool {
+        RUNNING.load(Ordering::Relaxed) && !uei_exited!(&self.skel.bss().uei)
+    }
+
+    fn run(&mut self) -> Result<()> {
+        while self.running() {
+            let interval_ms = self.prep_introspec();
+            std::thread::sleep(Duration::from_millis(interval_ms));
+            self.rb_mgr.poll(Duration::from_millis(100)).unwrap();
```
arighi, line 245: Is there any reason to have this `std::thread::sleep()` here? Can we just do:
```
        self.rb_mgr.poll(Duration::from_millis(interval_ms)).unwrap();
```

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+            self.cleanup_introspec();
```
arighi, line 246: Maybe it would be nice to print something in this loop, statistics or even a simple "I'm alive". :smile:. Considering that all the other schedulers are printing something in the main loop, someone may (incorrectly) assume that the scheduler is not functional.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+ * The scheduler treats the over-scheduled (or ineligible) tasks to enforce the
+ * fair use of CPU time. It defers choosing over-scheduled tasks to reduce the
+ * frequency of task execution. The deferring time- ineligible duration- is
+ * proportional to how much time is over-spent and added to the task's
+ * deadline. Additionally, suppose a task is compute-intensive and not
+ * latency-critical tasks. In that case, so its runtime per schedule is long
+ * enough without voluntarily yielding CPU time, the scheduler reduces the time
+ * slice, too. Note that reducing the time slice of a latency-critical task for
+ * fairness is not very effective because the scheduling overhead might be
+ * detrimental.
```
arighi, line 102: This last part about fairness is not very clear to me. Let me try to rephrase to make sure I understand:

Additionally, if a task is compute-intensive and not latency-critical, the scheduler automatically reduces its time slice, since its runtime per schedule is sufficiently long enough without voluntarily yielding the CPU.

Is this the same concept?

arighi (commented):
Overall it looks good to me. I've done some quick tests and I haven't seen any obvious issue. I left some inline comments.

Moreover, you may want to add a Signed-off-by line to your commits (all the other commits have a SoB line) and there are also some little coding style issues here and there, but nothing relevant.

For the Rust part you may want to run `rustfmt scheds/rust/scx_lavd/src/main.rs` to automatically fix all the formatting issues.

htejun (approved):
I scanned through the code. Overall, looks great to me. Thanks for the contribution and please feel free to land whenever you are ready. We can iterate in the tree.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	struct task_struct *waker;
+	struct task_ctx *taskc;
+	u64 now, interval;
+
+	/*
+	 * When a task @p is wakened up, the wake frequency of its waker task
+	 * is updated. The @current task is a waker and @p is a waiter, which
+	 * is being wakened up now.
+	 */
+	if (!(enq_flags & SCX_ENQ_WAKEUP))
```
htejun, line 1413: Maybe `is_wakeup_ef()` is more consistent? I wonder how helpful the `is_wakeup*()` helpers are tho given that they each just test a single flag and used in one or two places.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	greedy_threshold = lat_prio_to_greedy_thresholds[lat_prio];
+
+	gr_ft = (greedy_ratio * 1000) / greedy_threshold;
+	if (gr_ft < 1000)
+		gr_ft = 1000;
+	else
+		gr_ft *= LAVD_SLICE_GREEDY_FT;
+
+	return gr_ft;
+	}
```
htejun, line 799: indentation

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	 *    \/
+	 * [STOPPING] --> [ENQ] --> [RUNNING]
+	 *    /\                        |
+	 *    |                         |
+	 *    +-------------------------+
+	 */
+	const static int valid_tgt_stat[] = {
+		LAVD_TASK_STAT_ENQ,
+		LAVD_TASK_STAT_RUNNING,
+		LAVD_TASK_STAT_STOPPING,
```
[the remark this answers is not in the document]
htejun: Also, what were the transition cases which didn't follow the above transition diagram? I'm having a bit of hard time thinking up a case where it would deviate from the above.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+				 struct cpu_ctx *cpuc)
+{
+	u64 now, run_time_ns;
+
+	/*
+	 * If the task was runnable before governed by this scheduler, ignore
+	 * this run.
+	 */
+	now = bpf_ktime_get_ns();
+	if (now < taskc->last_start_clk)
```
htejun, line 1221: Here too, I don't think this can actually happen.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	 * If there is a relevant introspection command with @p, process it.
+	 */
+	try_proc_introspec_cmd(p, taskc, LAVD_CPU_ID_HERE);
+}
+
+static bool slice_fully_consumed(struct cpu_ctx *cpuc, struct task_ctx *taskc)
+{
+	u64 run_time_ns = 0;
+
+	if (taskc->last_stop_clk > taskc->last_start_clk)
```
htejun, line 1470: Here too, I'm really interested when this wouldn't hold. scx's state transition order is strictly `runnable -> [running -> stopping]* -> stopping`. I *think* there shouldn't be any case where the above wouldn't hold.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	if (!cpuc) {
+		scx_bpf_error("cpu_ctx_stor lookup failed");
+		return;
+	}
+
+	memset(cpuc, 0, sizeof(*cpuc));
+
+	__sync_fetch_and_sub(&nr_cpus_onln, 1);
+	update_sys_cpu_load();
+}
```
htejun, line 1586: Out of curiosity, does steamdeck use CPU on/offline for power management?

htejun (approved):
Okay, I actually went through the code this time. I really like the simplicity of it. The heuristics are easy to understand and conceptually concise. Please feel free to land anytime. Some overall thoughts:

- It looks like it would be relatively straight-forward to encapsulate the LAVD scheduling algorithm into a library so that it can be used by other schedulers to manage their in-domain scheduling.
- The mixture of raw nice values, 0-40 priority values and weights are a bit cumbersome. I suppose the intention was to ease table-mapping across different terms but that's not that difficult to do with well-defined linear weight values. The use of static_prio also makes it difficult to use it for flattened cgroup hierarchical scheduling as the flattened weight values need more than 40 levels of distinction. I think it should be possible to unify the values used to standardized weights without meaningfully affecting actual operation.
- There is no preemption. Was it that it wasn't necessary given the slices are fairly short and interactive tasks are prioritized heavily already? Or would it be interesting to explore that in the future?

SoulHarsh007:
:wave: I am not sure if this is the right place to report this, but I find the following kernel oops when using scx_lavd: https://paste.soulharsh007.dev/p/43a4476.log

Kernel Version: 6.8.1-2-cachyos

multics69:
Thank you for the report. If possible, can you provide some information when/how the bug happened? For example, shutdowning a laptop while listening a audio, etc. 

> 👋 I am not sure if this is the right place to report this, but I find the following kernel oops when using scx_lavd: https://paste.soulharsh007.dev/p/43a4476.log
> 
> Kernel Version: 6.8.1-2-cachyos

Byte-Lab:
Thanks for the report @SoulHarsh007. This crash would be due to a bug in the core framework, so I'm going to go ahead and merge this as it's an issue we'll have to fix over in https://github.com/sched-ext/sched_ext. I'll create an issue over there (tagging you in it) so we can track accordingly.

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+        setrlimit(Resource::MEMLOCK, soft_limit, rlimit::INFINITY).unwrap();
+
+        // Open the BPF prog first for verification.
+        let mut skel_builder = BpfSkelBuilder::default();
+        skel_builder.obj_builder.debug(opts.verbose > 0);
+        let mut skel = skel_builder.open()
+                                   .context("Failed to open BPF program")?;
+
+        // Initialize skel according to @opts.
+	let nr_cpus_onln = num_cpus::get() as u64;
```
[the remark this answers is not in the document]
multics69: Thanks for the suggestion. I will fix it.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: That's a good catch. Hehe, that's a side effect of multiple iterations. I will fix it.

scheds/rust/scx_lavd/src/main.rs
multics69: That's because it is assumed self.prep_introspec() is called every interval_ms micro-seconds. So self.prep_introspec() resets the introspection command every call.

Byte-Lab:
@multics69 hope it's OK that I merged and to address the feedback in a subsequent PR -- some folks were asking about playing with this and I felt it might be easier to just merge it and iterate in tree given that it had been open for a few days.

scheds/rust/scx_lavd/src/main.rs
multics69: That's a good suggestion. For now, since scx_lavd does not collect global statistics, I will change to print one scheduling decision every second.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Your version is better. I will change the last paragraph as follows:

 * The scheduler treats the over-scheduled (or ineligible) tasks to enforce the
 * fair use of CPU time. It defers choosing over-scheduled tasks to reduce the
 * frequency of task execution. The deferring time- ineligible duration- is
 * proportional to how much time is over-spent and added to the task's
 * deadline. Additionally, if a task is compute-intensive and not
 * latency-critical, the scheduler automatically reduces its time slice, since
 * its runtime per schedule is sufficiently long enough without voluntarily
 * yielding the CPU. Note that reducing the time slice of a latency-critical
 * task for fairness is not very effective because the scheduling overhead
 * might be detrimental.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	 * frequently ineligible for low (tail) latency.
+	 */
+	greedy_threshold = lat_prio_to_greedy_thresholds[lat_prio];
+
+	return taskc->greedy_ratio <= greedy_threshold;
+}
+
+static bool is_wakeup_wf(u64 wake_flags)
+{
+	return wake_flags & (SCX_WAKE_TTWU | SCX_WAKE_SYNC);
```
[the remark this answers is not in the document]
multics69: Thanks for letting me know. I will fix it.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Will fix it. Thanks!

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: No, but the current SteamDeck firmware reports that 2x more CPUs are available so I used the number of online CPUs.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	 * When a task @p is wakened up, the wake frequency of its waker task
+	 * is updated. The @current task is a waker and @p is a waiter, which
+	 * is being wakened up now.
+	 */
+	if (!(enq_flags & SCX_ENQ_WAKEUP))
+		return;
+
+	waker = bpf_get_current_task_btf();
+	taskc = bpf_task_storage_get(&task_ctx_stor, waker, 0, 0);
+	if (!taskc) {
```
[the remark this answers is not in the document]
multics69: I will also refactor get_cpu_ctx[_id]().

multics69:
Sure, no problem. Thanks for taking care of that bug!

scheds/rust/scx_lavd/src/bpf/main.bpf.c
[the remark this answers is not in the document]
multics69: That's a good suggestion. That is much easier to read. I will change it as you suggested.

Two types of invalid state transitions have been observed so far, and there are reasons for that. The two are as follows:

- STOPPING -> RUNNING: This can happen when the task is forked (STOPPING) and then scheduled (RUNNING) without enqueuing.
- ENQ -> ENQ: The underlying scx core can issue enqueues more than once for a task.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: I will fix it as you suggested.

is_wakeup_*() helpers are not very useful. They are mostly for code readability. I wish the code could be read without much knowledge of the kernel flags. As it becomes too cumbersome, I will do refactoring once again.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: You are right. I will remove the unnecessary condition check.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: You are right. These two unnecessary condition checks were added when the state transition checking was implemented. I removed the condition check.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	/*
+	 * Aggregate nice priority and latency boost priority, then convert the
+	 * final priority to weight.
+	 */
+	nice_prio = get_nice_prio(p);
+	lat_boost_prio = boost_lat(p, taskc, is_wakeup);
+	prio = sum_prios_for_lat(p, nice_prio, lat_boost_prio);
+
+	weight = sched_prio_to_latency_weight[prio];
+	return weight;
```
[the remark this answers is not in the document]
multics69: You are right. I will fix it.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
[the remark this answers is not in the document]
htejun: BTW, whether a task is under SCX or not doesn't determine whether `ops.init_task()` is called on them. Even tasks which are currently under `RT` and stay that way until it exits or the scx scheduler unloads will have `ops.init/exit_task()` invoked on them. Here, it's just that swappers are really special (e.g. they share the same PID across different CPUs).

scheds/rust/scx_lavd/src/bpf/main.bpf.c
htejun: It might be a good idea to verify first by triggering `scx_bpf_error()` just in case, at least until we're sure.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	 */
+	cpu_id = scx_bpf_select_cpu_dfl(p, prev_cpu, wake_flags, &found_idle);
+
+	if (found_idle && is_wakeup_wf(wake_flags)) {
+		struct task_ctx *taskc;
+
+		taskc = bpf_task_storage_get(&task_ctx_stor, p, 0, 0);
+		if (!taskc) {
+			scx_bpf_error("task_ctx_stor lookup failed");
+			return prev_cpu;
```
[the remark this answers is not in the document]
multics69: Thanks a lot! It could be a super nasty bug! I will fix it as suggested.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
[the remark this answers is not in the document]
htejun: I thought about invoking `ops.init/exit_task()` on `swapper`s but I think that's likely to cause more subtle problems than help anything. There's one per cpu and they all share the invalid PID 0, so schedulers which use PID to index stuff can get really confused and that can be a lot more difficult to notice than the bugs caused by not invoking `ops.init()` on them. I'll add documentation noting that `ops.init/exit()` are never invoked on the `swapper`s.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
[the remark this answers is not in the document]
htejun: ENQ -> ENQ transitions are because `ops.dequeue()` is not implemented. What's actually happening is ENQ -> DEQ -> ENQ which happens when task attributes are updated (e.g. nice level, allowed cpus and so on). The DEQ cycling is always transient, so ignoring this is completely fine and will result in the correct behavior for lavd.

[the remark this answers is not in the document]
htejun: The STOPPING -> RUNNING transitions are because there are several cases where scx core skips enqueue including:
- bypass mode is turned on (this is turned on during both init and exit. it's also used across suspend/resume operations.
- `SCX_OPS_ENQ_EXITING` is not set and an exiting task was woken up.
- The associated CPU is not fully online.

These events should be rare so the impact is limited but the current code is probably not behaving as intended right now as it'd end up ignoring STOPPING -> RUNNING transition completely when what's actually happening is more akin to STOPPING -> ENQ -> RUNNING where ENQ step is implicit.

I wonder whether  a better way to do this is using `ops.runnable()` and `ops.quiescent()` to add and subtract from system load numbers. They are guaranteed to follow `runnable ->  [ running -> stopping -> ]* quiescent` transitions and using them would make both ENQ -> ENQ and STOPPING -> RUNNING go away  and thus simplify the resulting code.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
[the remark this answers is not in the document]
multics69: I agree that putting the swapper under the scx would cause more problems than good. PID 0 could be used for invalid PID in many scx schedulers.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+ * among concurrent tasks. For example, in the case of a normal priority task
+ * with priority 19 (or nice 0), its corresponding value is 10,000,000 nsec (or
+ * 10 msec). This guarantees that any tasks enqueued in 10 msec after the task
+ * is enqueued will not compete with the task. The array is adapted from the
+ * MuQSS scheduler. We choose a different distribution for
+ * sched_prio_to_latency_weight on purpose instead of inversing
+ * sched_prio_to_slice_weight. That is because sched_prio_to_slice_weight is
+ * too steep to use for latency. We normalized the values so that the normal
+ * priority (nice 0) has a deadline of 10 msec. The virtual deadline ranges
+ * from 1.5 msec to 60.9 msec.
```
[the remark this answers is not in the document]
multics69: I will update the comments. Thanks!

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1749 @@
...
+	return enq_flags & SCX_ENQ_WAKEUP;
+}
+
+static u64 calc_eligible_delta(struct task_struct *p, struct task_ctx *taskc)
+{
+	/*
+	 * We calculate how long a task should be ineligible for execution. To
+	 * this end, the scheduler stretches the ineligible duration of a task
+	 * so it can control the frequency of the task's running to let the
+	 * task pay its debut. Reducing the time slice of a task would be
```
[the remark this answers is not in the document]
multics69: Will fix it.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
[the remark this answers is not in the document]
multics69: Thanks for the detailed explanation! I updated the comments based on your explanation. 

For the STOPPING->RUNNING transition, actually I tried this time but it does not help improving the accuracy of task statistics. I thnink that's because they are rare and special cases.

Using ops.runnable() and ops.quiescent() is not the information that scx_lavd wants to collect. The goal is to collect task's actual CPU time not how long the task has been runnable. So I think accumulating time between ops.running() and ops.stopping() should be what we want.

shelterx:
Not sure if this is the correct place since it's merged, but I got this:
```
sched_ext: BPF scheduler "lavd" errored, disabling
sched_ext: runnable task stall (kworker/15:2[20480] failed to run for 5.206s)
   scx_watchdog_workfn+0x12d/0x1c0
   process_scheduled_works+0x1e4/0x420
   worker_thread+0x2ff/0x620
   kthread+0x1b3/0x250
   ret_from_fork+0x34/0x40
   ret_from_fork_asm+0x11/0x20
hrtimer: interrupt took 13489 ns
```

Also I had huge audio lag/issues in game when this scheduler was running.

multics69:
Thank you for reporting the issue. Would you mind providing a bit more detail which might help me to reproduce the problem? For example, CPU model, the played game, other programs ran, etc.

> Not sure if this is the correct place since it's merged, but I got this:
> 
> ```
> sched_ext: BPF scheduler "lavd" errored, disabling
> sched_ext: runnable task stall (kworker/15:2[20480] failed to run for 5.206s)
>    scx_watchdog_workfn+0x12d/0x1c0
>    process_scheduled_works+0x1e4/0x420
>    worker_thread+0x2ff/0x620
>    kthread+0x1b3/0x250
>    ret_from_fork+0x34/0x40
>    ret_from_fork_asm+0x11/0x20
> hrtimer: interrupt took 13489 ns
> ```
> 
> Also I had huge audio lag/issues in game when this scheduler was running.

shelterx:
@multics69 
Some more info, CPU is Intel 9900KS, nothing else was running.
Another crash with more output, the game is AC Valhalla.
```
07:17:48 [INFO] scx_lavd scheduler is initialized
07:17:48 [INFO] scx_lavd scheduler starts running.

DEBUG DUMP

kworker/u32:0[2494] triggered exit kind 1026:
  runnable task stall (audio_client_ma[2660] failed to run for 6.928s)

Backtrace:
  process_scheduled_works+0x238/0x420
  worker_thread+0x37f/0x670
  kthread+0x2dd/0x320
  ret_from_fork+0x34/0x40
  ret_from_fork_asm+0x11/0x20

Runqueue states
---------------

CPU 0   : nr_run=1 flags=0x0 cpu_rel=0 ops_qseq=369019 pnt_seq=1198790
          curr=ACValhalla.exe[2590] class=ext_sched_class

 *R ACValhalla.exe[2590] -2ms
      cpus=ffff


CPU 1   : nr_run=1 flags=0x0 cpu_rel=0 ops_qseq=328046 pnt_seq=840722
          curr=kworker/u32:4[181] class=ext_sched_class

 *R kworker/u32:4[181] +0ms
      cpus=ffff


CPU 2   : nr_run=1 flags=0x0 cpu_rel=0 ops_qseq=356138 pnt_seq=952088
          curr=ACValhalla.exe[2586] class=ext_sched_class

 *S ACValhalla.exe[2586] +0ms
      cpus=ffff


CPU 4   : nr_run=1 flags=0x0 cpu_rel=0 ops_qseq=337482 pnt_seq=899483
          curr=swapper/4[0] class=idle_sched_class

  R ACValhalla.exe[2583] +0ms
      sticky/holding_cpu=-1/-1 dsq_id=0x8000000000000002
      cpus=ffff

    __se_sys_futex_waitv+0x2f7/0x3b0
    do_syscall_64+0x8e/0x170
    entry_SYSCALL_64_after_hwframe+0x46/0x4e

CPU 5   : nr_run=1 flags=0x0 cpu_rel=0 ops_qseq=218990 pnt_seq=518331
          curr=swapper/5[0] class=idle_sched_class

  R ACValhalla.exe[2585] +0ms
      sticky/holding_cpu=-1/-1 dsq_id=0x8000000000000002
      cpus=ffff

    __se_sys_futex_waitv+0x2f7/0x3b0
    do_syscall_64+0x8e/0x170
    entry_SYSCALL_64_after_hwframe+0x46/0x4e

CPU 9   : nr_run=1 flags=0x0 cpu_rel=0 ops_qseq=459582 pnt_seq=1259272
          curr=kworker/u32:0[2494] class=ext_sched_class

 *R kworker/u32:0[2494] +0ms
      cpus=ffff

    irq_work_run+0x65/0x190
    __sysvec_irq_work+0x1c/0xc0
    sysvec_irq_work+0x69/0x80
    asm_sysvec_irq_work+0x1a/0x20
    scx_watchdog_workfn+0x18c/0x2a0
    process_scheduled_works+0x238/0x420
    worker_thread+0x37f/0x670
    kthread+0x2dd/0x320
    ret_from_fork+0x34/0x40
    ret_from_fork_asm+0x11/0x20

CPU 10  : nr_run=3 flags=0x0 cpu_rel=0 ops_qseq=430700 pnt_seq=1089487
          curr=vkd3d-swapchain[2683] class=ext_sched_class

 *R vkd3d-swapchain[2683] -3ms
      cpus=ffff


  R audio_client_ma[2660] -6928ms
      sticky/holding_cpu=-1/-1 dsq_id=0x8000000000000002
      cpus=ffff

    __x64_sys_poll+0xfe/0x1a0
    do_syscall_64+0x8e/0x170
    entry_SYSCALL_64_after_hwframe+0x46/0x4e

  R kworker/10:3[602] -6740ms
      sticky/holding_cpu=-1/-1 dsq_id=0x0
      cpus=0400

    kthread+0x2dd/0x320
    ret_from_fork+0x34/0x40
    ret_from_fork_asm+0x11/0x20

CPU 12  : nr_run=1 flags=0x0 cpu_rel=0 ops_qseq=228768 pnt_seq=557851
          curr=ACValhalla.exe[2584] class=ext_sched_class

 *R ACValhalla.exe[2584] +0ms
      cpus=ffff


Error: EXIT: runnable task stall (audio_client_ma[2660] failed to run for 6.928s)

```

Issue #195, opened 2024-03-20
[Bug Report] Not able to start scx programs with Kernel `6.8.1-cachyos-lto`


yqlbu:
## Summary

As the title suggests. It was working 3 days ago with Kernel `6.8.0-cachyos-lto` before I sync up the upstream changes.

## Addtional Info

Hardware:

```console
⯁ flake git:(master) ✗ ❯❯❯ neofetch
          ▗▄▄▄       ▗▄▄▄▄    ▄▄▄▖            kev@nixos-x1-carbon
          ▜███▙       ▜███▙  ▟███▛            -------------------
           ▜███▙       ▜███▙▟███▛             OS: NixOS 24.05.20240316.c75037b (Uakari) x86_64
            ▜███▙       ▜██████▛              Host: LENOVO 21KC0000CD
     ▟█████████████████▙ ▜████▛     ▟▙        Kernel: 6.8.1-cachyos
    ▟███████████████████▙ ▜███▙    ▟██▙       Uptime: 4 mins
           ▄▄▄▄▖           ▜███▙  ▟███▛       Packages: 1184 (nix-system), 1367 (nix-user), 6 (flatpak)
          ▟███▛             ▜██▛ ▟███▛        Shell: fish 3.7.0
         ▟███▛               ▜▛ ▟███▛         Resolution: 2880x1800
▟███████████▛                  ▟██████████▙   DE: Hyprland (Wayland)
▜██████████▛                  ▟███████████▛   WM: .Hyprland-wrapp
      ▟███▛ ▟▙               ▟███▛            Theme: Flat-Remix-GTK-Grey-Darkest [GTK2/3]
     ▟███▛ ▟██▙             ▟███▛             Icons: Papirus-Dark [GTK2/3]
    ▟███▛  ▜███▙           ▝▀▀▀▀              Terminal: alacritty
    ▜██▛    ▜███▙ ▜██████████████████▛        CPU: Intel Ultra 7 155H (22) @ 1.400GHz
     ▜▛     ▟████▙ ▜████████████████▛         GPU: Intel Arc Graphics]
           ▟██████▙       ▜███▙               Memory: 2921MiB / 31778MiB
          ▟███▛▜███▙       ▜███▙
         ▟███▛  ▜███▙       ▜███▙
         ▝▀▀▀    ▀▀▀▀▘       ▀▀▀▘
```

CPU: `Intel(R) Core(TM) Ultra 7 155H`

[...]

PR #196, opened 2024-03-21
topology: Fall back to cache 0

Byte-Lab:
As described in https://github.com/sched-ext/scx/issues/195, apparently some chips don't export information about their cache topology. There's not much we can do if we don't have that information, so let's just assume a unified cache per node if that happens.

Andrea suggested this patch -- I'm applying exactly what he proposed, with a slightly modified comment.


arighi (approved):
That was basically tested already in #195, the change is the same as the one that I suggested with the addition of a nice comment.

Therefore, LGTM.

Thanks!

htejun:
We probably should further investigate why the cache ID is missing. Maybe we can ask for `lstopo` output?

PR #197, opened 2024-03-25
scx_lavd: improve virtual deadline calculation 

multics69:
This PR improves the virtual deadline calculation. It consists of four commits including

1. improving the accuracy of runtime per schedule calculation
2. tuning LAVD_LC_RUNTIME_MAX as a side effect of (1)
3. replacing the previous latency weight array from MuQss scheduler to the inverse of our nice weight array
4. improving latency criticality to priority mapping

These changes help especially for over-loaded cases.


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1259,10 +1259,16 @@ static void update_stat_for_stop(struct task_struct *p, struct task_ctx *taskc,
...
+	 * Update task's run_time. If a task got slice-boosted -- in other
+	 * words, its time slices have been fully consumed multiple times,
+	 * stretch the measured runtime according to the slice_boost_prio.
+	 * The stretched runtime more accurately reflects the actual runtime
+	 * per schedule as if a large enough time slice was given in the first
+	 * place.
 	 */
 	run_time_ns = now - taskc->last_start_clk;
-	taskc->run_time_ns = calc_avg(taskc->run_time_ns, run_time_ns);
+	run_time_boosted_ns = run_time_ns * (1 + taskc->slice_boost_prio);
```
htejun, line 1342: I'm a bit confused by this change. `slice_boost_prio` is used to boost `p->scx.slice` for long-running tasks. That boosting occurs in `lavd_running()` which is also where `last_start_clk` is recorded. So, in this function which is called from `lavd_stopping()`, `run_time_ns` should reflect the fully boosted runtime and if it gets boosted again by `slice_boost_prio` for accounting leading to double boosting. What am I missing?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -673,6 +678,27 @@ static void update_sys_cpu_load(void)
 		load_actual += cpuc->load_actual;
 		load_ideal += cpuc->load_ideal;
 
+		/*
+		 * Accumulate task's latency criticlity information.
+		 *
+		 * Updating cpuc->* is racy. However, it could degrade the
+		 * accuracy a little bit in very rare cases, so we do embrace
+		 * embrace for performance on purpose.
```
htejun, line 686: Double `embrace`. I'm having a bit of difficulty understanding what the comment is trying to say. Is it saying that while updating cpu->* is racy, the resulting impact on accuracy should be small and very rare and thus should be fine?

htejun (approved):
Left a couple comments but generally looks fine to me. Please feel free to merge as you see fit.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Yes. Thanks for the suggestion. I will revise the sentence.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Oh, you are right. The `slice_boost_prio` should be updated before calculating `run_time_boosted_ns`. I will change the code to call `adjust_slice_boost()` inside `update_stat_for_stop()`.

htejun: I'm still confused by the fix. `slice_boost_prio`'s function is extending the slice duration for the task, right? And at this point, the task ran with the extended slice since `lavd_running()` and is now stopping. So, `now - taskc->last_start_clk` should already record the extended slice that the task has just consumed. I'm confused by it needs to be multiplied by `(1 + taskc->slice_boost_prio)` again here when the slice had already been adjusted when the task started running.

multics69: Thank you for asking the question again. It makes my thought process more apparent.

I wanted to handle this case -- 1) a task consumed its assigned time slice entirely multiple times, and 2) it was scheduled out only due to time slice expiration (i.e., no voluntary yield). In this case, if a long time slice was given in the first place, the measured runtime should be longer than the current calculation, and multiplying `slice_boost_prio` is one way to compensate for such under-measurement.

For example, support that Task A has fully consumed time slide in three times -- 2 msec, 2 msec, and 2 msec -- without yielding a CPU. The measured runtime will be 2 msec. However, if the time slice of 6 msec were given in the first place, it will consume the 6ms time slide entirely. By multiplying `slice_boost_prio` to the measured runtime, we can somehow mitigate such an under-measurement problem.

What do you think?

By the way, I think the code needs to reduce (or reset) `slice_boost_prio` when a task yields a CPU for condition (2).

multics69: It would probably be better to accumulate consecutive full-time slice consumptions without yield rather than approximating it using `slice_boost_prio`.

htejun: Ah, I see what you mean, so the code is trying to use `slice_boost_prio` to tell how many consecutive times the task has consumed its slice because it remained RUNNABLE through multiple slices.

I wonder whether this was necessary because previously `update_stat_for_stop()` wasn't being called for cases where the task switches between running and stopping multiple times before going back to sleep. So, before #199, those would be `LAVD_TASK_STAT_RUNNING` -> `STOPPING` -> `RUNNING` -> `STOPPING` transitions. The code would have ignored the `STOPPING` -> `RUNNING` transition and thus subsequently `STOPPING` -> `STOPPING` transitions - ie. the accounting would only happen for the first slice that was fully consumed.

After #199 tho, this should no longer be the case. When the task expires due to slice exhaustion and goes back on a CPU with new slice, both will be correctly accounted for no matter how many times that occurs. ie. The code should already be doing the correct thing of accumulating the actually consumed slice multiple times. Each time the slice is fully consumed, the slice would get boosted and when each slice expires, the extended slice will be accounted correctly on expiration.

multics69: That makes sense. After #199, it is not necessary. I will clean up the code and submit another PR.

PR #199, opened 2024-03-26
scx_lavd: Clean up task state transition tracking

htejun:
scx_lavd was implementing custom task state transition tracking to avoid updating statistics on invalid transitions. However, sched_ext core already implements clearly defined task state transitions. This PR updates scx_lavd so that it uses sched_ext's task state transitions. Note that the commits are structured so that `transit_task_stat()` is verified to always concur with sched_ext state transitions before the custom mechanism is removed.

While at it, this PR also removes the unnecessary `extern crate` declarations.


multics69:
The changes are beautiful and works well on my side.

multics69 (commented):
The changes are beautiful and work well on my side.

PR #204, opened 2024-03-30
scx_lavd: directly accumulate task's runtimes for consecutive runnings

multics69:
Drop slice_boost_prio's multiplying runtime of task's consecutive runnings. Instead of such approximation, let's directly measure the accumulated run time of consecutive runnings in a runnable<-->quiescent transition. The calculation of run_freq has been updated according to this change. In addition, the upper bound of ineligible duration (LAVD_ELIGIBLE_TIME_MAX) increases to better reflect the actual distribution of tasks' ineligible duration. Finally, the rest commits improves the code for better readability.


htejun (approved):

multics69:
Thank you @htejun for the revirew!

PR #211, opened 2024-04-03
openrc init support

nycko123:
Additional meson option to build with openrc init.

Tested on Artix.


sirlucjan (approved):
LGTM.

sirlucjan:
Thanks. This will definitely satisfy users NOT using systemd. And it works on exactly the same principle - as it detects systemd, it will install systemd services. If it detects openrc, it will also install the appropriate files.

nycko123:
In this form it's up to the packager to decide, which init should be used in a distribution. There is no auto detection implemented.

PR #213, opened 2024-04-04
ci: make test build more robust

arighi:
Always use custom linux headers to prevent potential build failures and also use `--jobs=1` during the build to prevent overloading the system during the CI run.

### DIFF
--- a/.github/workflows/build-scheds.yml
+++ b/.github/workflows/build-scheds.yml
@@ -16,7 +16,7 @@ jobs:
       ### DOWNLOAD AND INSTALL DEPENDENCIES ###
 
       # Download dependencies packaged by Ubuntu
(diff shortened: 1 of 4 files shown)


.github/workflows/build-scheds.yml
```diff
@@ -48,11 +48,14 @@ jobs:
...
+      # Generate kernel headers
+      - run: cd linux && make headers
+
       ### END DEPENDENCIES ###
 
       # The actual build:
-      - run: meson setup build -Dkernel=$(pwd)/linux
-      - run: meson compile -C build
+      - run: meson setup build -Dkernel=$(pwd)/linux -Dkernel_headers=./linux/usr/include
+      - run: meson compile -C build --jobs=1
```
Byte-Lab, line 58: Just out of curiosity, why do we want to avoid overloading the system during a CI run given that all it's doing is building + testing?

arighi: At the moment this is not critical, but in the future, if we add more Rust components, they'll be all built in parallel by meson in a N^2 way (where N is the amount of CPUs), so there's the risk that at some point we may hit OOM conditions.

The issue is that meson will try to parallelize the sub-projects spawning N tasks, each task will run cargo that is not aware of meson, so it'll also try to spawn N tasks, causing a massive NxN parallel build workload... serializing each cargo run would still build the Rust components in parallel, but limited to N tasks.

Byte-Lab (approved):

PR #217, opened 2024-04-04
scx_rustland_core: separate crate source code from assets

arighi:
scx_rustland_core needs to ship both a binary part and a source code part, which will be used to build schedulers based on it.

To effectively publish the scx_rustland_core crate on crates.io we need to properly separate the source code assets from the crate's main source code.

To achieve this, move the assets into a separate directory and declare them inside a [lib] section in Cargo.toml.

This allows to publish the crate on crates.io, providing also a clear separation between source code and assets.


htejun (approved):

htejun:
I should update scx_utils to do the same thing instead of the manual tar/untar from build.rs.

PR #218, opened 2024-04-04
Gracefully handle hotplug in scx_rusty

Byte-Lab:
The scx_rusty scheduler does not support hotplug, and expects a static
host topology throughout its runtime. Though the kernel does have
support for detecting hotplug events, we currently don't detect this in
the kernel, nor surface it to user space when it happens. Now that we
have scx_bpf_exit(), we can gracefully exit the kernel in the event of a
hotplug, and communicate to user space that it should restart the
scheduler.
    
This patch adds that support to scx_rusty. Note that this assumes that
we're running on a recent enough kernel that has scx_bpf_exit(). If it
doesn't, then we instead just error out of the kernel scheduler and exit
the application.

### DIFF
--- a/rust/scx_utils/src/user_exit_info.rs
+++ b/rust/scx_utils/src/user_exit_info.rs
@@ -22,6 +22,7 @@ pub enum ScxExitKind {
     None = bindings::scx_exit_kind_SCX_EXIT_NONE as isize,
     Done = bindings::scx_exit_kind_SCX_EXIT_DONE as isize,
     Unreg = bindings::scx_exit_kind_SCX_EXIT_UNREG as isize,
+    UnregBPF = bindings::scx_exit_kind_SCX_EXIT_UNREG_BPF as isize,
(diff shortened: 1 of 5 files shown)


htejun (approved):

arighi:
It'd be nice if there was a generic way to apply this change automatically to all the schedulers, considering that most (all?) of them are not handling CPU hotplugging nicely at the moment, maybe a flag `SCX_OPS_EXIT_ON_CPU_HOTPLUG` or similar...

Anyway, this looks good to me for now, I'll do the same in scx_rustland_core.

Byte-Lab:
Oh yeah, that's a good idea, definitely seems like this would be something generally useful. @htejun wdyt?

htejun:
Yeah, it'd be nice to make this generic in scx_utils. Traits are tough to do with skels unfortunately, so we might have to do even more macros, which is a bit sad but not the end of the world.

PR #224, opened 2024-04-08
scx_lavd: support preemption (in some scenarios)

multics69:
This PR implements preemption in scx_lavd. More specially, a task on scx_lavd can be kicked before its time slice is exhausted. The PR consists of two parts:

1) When a task is enqueued to the global DSQ, the scheduler checks if it is worth investing a preemption effort, and if so, it finds and kicks a victim CPU to run a more urgent task.

2) It monitors the system's current load. When overloaded, it lengthens the virtual deadline. It helps the system run smoothly when heavily overloaded. 

Note that the current preemption logic very judiciously performs preemption since kicking a CPU is expensive.


htejun (approved):

htejun:
Approved but the commit history looks a bit off. Might be a good idea to clean up before landing?

PR #236, opened 2024-04-23
scheds-rust: build rust schedulers in sequence

### LINKED ISSUE #202: meson compile should obey jobs flag
I tried to compile the schedulers on my system but my machine was quickly rendered unusable. No matter what I do it uses all cores and it looks like it's hard coded to make_jobs = nproc. Im not sure what all is compiling with the make_jobs flag but that was the only thing I saw with a super quick glance through.

takase1121:
Potentially fixes #202.

This is honestly a temporary fix until a proper fix comes by.

### DIFF
--- a/scheds/rust/meson.build
+++ b/scheds/rust/meson.build
@@ -1,5 +1,23 @@
+# the previous scheduler in the compile sequence
+sched = []
+
+# to add a new scheduler, assign the output of your custom_target to sched
+# and add sched as a dependency to your custom_target. For example:
+#
+# sched = custom_target('scx_mysched',
+#                       ...
+#                       depends: [mydep, sched],
+#                       build_always_stale: true)
 subdir('scx_layered')
 subdir('scx_rusty')
 subdir('scx_rustland')
 subdir('scx_rlfifo')
 subdir('scx_lavd')
+
+# the target to compile all rust schedulers
+custom_target('rust_scheds',
+              input: 'meson.build',
+              output: '@PLAINNAME@.__PHONY__',
+              command: ['touch', '@PLAINNAME@.__PHONY__'],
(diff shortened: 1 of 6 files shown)


scheds/rust/scx_lavd/meson.build
```diff
@@ -1,7 +1,8 @@
-custom_target('scx_lavd',
+sched = custom_target('scx_lavd',
               output: '@PLAINNAME@.__PHONY__',
               input: 'Cargo.toml',
               command: [cargo, 'build', '--manifest-path=@INPUT@', '--target-dir=@OUTDIR@',
                         cargo_build_args],
               env: cargo_env,
-              build_by_default: true)
+              depends: [sched],
```
htejun, line 7: Is the dependency being different from other schedulers intentional?

scheds/rust/meson.build
```diff
@@ -1,5 +1,23 @@
+# the previous scheduler in the compile sequence
+sched = []
+
+# to add a new scheduler, assign the output of your custom_target to sched
+# and add sched as a dependency to your custom_target. For example:
```
htejun, line 5: Would you mind adding a bit of explanation on why we're doing this?

scheds/rust/scx_lavd/meson.build
htejun: Ah, right. It wasn't there from before. I don't think that's correct but a separate issue.

scheds/rust/meson.build
takase1121: Fixed. Let me know if the explanation is not detailed enough.

htejun:
Thank you so much. This has been an issue for a while and I'm really glad that the situation can be improved.

PR #238, opened 2024-04-23
scx_rustland: reduce overhead by caching host topology

arighi:
Looking at `perf top` it seems that the scheduler can spend a significant amount of time in topology::Cache::span(), especially when the system is running a significant amount of tasks:

   5.46%  scx_rustland    [.] scx_utils::topology::Cache::span

Considering that scx_rustland doesn't support CPU hotplugging yet (it requires a full restart to properly handle CPU hotplug events), we can completely avoid this overhead by caching the cores/CPUs mapping at the beginning, when the scheduler starts, instead of constantly re-evaluating the CPU topology information.

This allows to reduce the scheduler overhead by ~10% CPU utilization under heavy load conditions (from ~68% -> ~60%, according to `top`).

### DIFF
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -10,6 +10,7 @@ mod bpf;
 use bpf::*;
 
 use scx_utils::Topology;
+use scx_utils::TopologyMap;
 
 use std::thread;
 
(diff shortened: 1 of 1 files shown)


htejun (approved):

Byte-Lab (approved):
LGTM, but we should probably also make the Cpumask implementation more efficient. I'll work on that

Byte-Lab:
In case you want to give it a try, this might help: https://github.com/sched-ext/scx/pull/241

arighi:
> In case you want to give it a try, this might help: #241

Updated this PR on top of #241, also introducing a new TopologyMap object, that can be used to cache Topology information (instead of implementing a custom local caching in `scx_rustland`).

arighi:
(meh... pushed the wrong branch, should be good now)

PR #239, opened 2024-04-24
Add CPU frequency related helpers and extend scx_layered

hodgesds:
This change adds `scx_bpf_cpuperf_cap`, `scx_bpf_cpuperf_cur` and `scx_bpf_cpuperf_set` definitions that were recently introduced into [`sched_ext`](https://github.com/sched-ext/sched_ext/pull/180). It adds a `perf` field to `scx_layered` to allow for controlling performance per layer.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -722,6 +722,7 @@ void BPF_STRUCT_OPS(layered_running, struct task_struct *p)
 	struct cpu_ctx *cctx;
 	struct task_ctx *tctx;
 	struct layer *layer;
+	u32 cpu;
 
 	if (!(cctx = lookup_cpu_ctx(-1)) || !(tctx = lookup_task_ctx(p)) ||
 	    !(layer = lookup_layer(tctx->layer)))
@@ -734,12 +735,14 @@ void BPF_STRUCT_OPS(layered_running, struct task_struct *p)
 	cctx->current_exclusive = layer->exclusive;
 	tctx->started_running_at = bpf_ktime_get_ns();
 
+	cpu = scx_bpf_task_cpu(p);
+
 	/*
 	 * If this CPU is transitioning from running an exclusive task to a
 	 * non-exclusive one, the sibling CPU has likely been idle. Wake it up.
 	 */
 	if (cctx->prev_exclusive && !cctx->current_exclusive) {
-		s32 sib = sibling_cpu(scx_bpf_task_cpu(p));
+		s32 sib = sibling_cpu(cpu);
 		struct cpu_ctx *sib_cctx;
 
 		/*
@@ -753,6 +756,9 @@ void BPF_STRUCT_OPS(layered_running, struct task_struct *p)
 			scx_bpf_kick_cpu(sib, 0);
 		}
 	}
+
+	if (layer->perf > 0)
+		scx_bpf_cpuperf_set(cpu, layer->perf);
 
 	cctx->maybe_idle = false;
 }
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -312,6 +312,8 @@ enum LayerKind {
         cpus_range: Option<(usize, usize)>,
         #[serde(default)]
         min_exec_us: u64,
+        #[serde(default)]
+        perf: u64,
     },
     Grouped {
         util_range: (f64, f64),
@@ -323,6 +325,8 @@ enum LayerKind {
         preempt: bool,
         #[serde(default)]
         exclusive: bool,
+        #[serde(default)]
+        perf: u64,
     },
     Open {
         #[serde(default)]
@@ -331,6 +335,8 @@ enum LayerKind {
         preempt: bool,
         #[serde(default)]
         exclusive: bool,
+        #[serde(default)]
+        perf: u64,
     },
 }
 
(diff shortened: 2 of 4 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -560,30 +560,42 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 
 	/* consume preempting layers first */
 	bpf_for(idx, 0, nr_layers)
-		if (layers[idx].preempt && scx_bpf_consume(idx))
```
hodgesds, line 563: I'm not sure if `layered_dispatch` is the right place to put this, so please let me know if there is a more suitable place. Probably could have cleaned up some of the redundant checks as well.

hodgesds:
Test setup:
```
$ cat /proc/cpuinfo | grep 'model name' | head -n 1
model name      : Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz
$ cat /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor | uniq -c 
     80 schedutil
```
Config:
```
[
  {
    "name": "batch",
    "comment": "tasks under system.slice or tasks with nice value > 0",
    "matches": [
      [
        {
          "CgroupPrefix": "system.slice/"
        }
      ],
      [
        {
          "NiceAbove": 0
        }
      ]
    ],
    "kind": {
      "Confined": {
        "util_range": [
          0.01,
          1.0
        ],
        "cpus_range": [
          0,
          80
        ],
        "min_exec_us": 1000,
        "perf": 128
      }
    }
  },
  {
    "name": "normal",
    "comment": "the rest",
    "matches": [
      []
    ],
    "kind": {
      "Open": {
        "min_exec_us": 100,
        "preempt": true,
        "exclusive": true,
        "perf": 128
      }
    }
  }
]
```
The image shows it was able to clock down the frequency effectively.
image

scheds/rust/scx_layered/src/main.rs
```diff
@@ -312,6 +312,8 @@ enum LayerKind {
         cpus_range: Option<(usize, usize)>,
         #[serde(default)]
         min_exec_us: u64,
+        #[serde(default)]
+        perf: u64,
```
Byte-Lab, line 316: So this is certainly one option for how to expose this knob to configurations. Another option would be to expose this as something like QoS, and have the scheduler implementation abstract how that's eventually applied with `scx_bpf_cpuperf_set()`. A few things to consider:

1. Say we have a layer that has very relaxed runtime requirements, and one with very sensitive latency requirements. If both layers are occupying CPUs on the system, it may be desirable to run the low-pri layer at a low frequency if the goal is to allow the high-pri layer running on other cores to run hotter. On the other hand, if the high-pri layer isn't running at that moment, we'd ideally like to run the lower-pri layer at a higher frequency given that it won't affect the high-pri layer. If we directly passed perf through the config like this, it would cause a low-pri layer to _always_ run with low perf even if there's nothing else on the system for it to contend with. Maybe that's OK (or possibly even desirable given how predictable the behavior would be) for a configuration-driven scheduler which really isn't meant to be a general purpose scheduling solution, but it also does feel a bit suboptimal.

2. I might be getting a bit ahead of myself here, but if we ever want to have layered run on asym chips like ARM big.LITTLE, just specifying perf on its own wouldn't be sufficient. Maybe having QoS would be, and the scheduler could internally use `scx_bpf_cpuperf_set()`, `scx_bpf_cpuperf_cap()`, and `scx_bpf_cpuperf_cur()` to find the right CPU and frequency for it? Not sure, maybe we'd need new configs.

3. I expect it would be difficult for a user to know what value to choose for perf. It seems like what a user really wants to say is, "this layer is super important and should be boosted". Maybe it would even be as simple as keeping a cpumask that has bits set for every CPU running a high-pri task. If the weight of that cpumask is nonzero, then we'd know that we (may) need to throttle any non-high-pri tasks.

What do you think?

scheds/rust/scx_layered/src/bpf/main.bpf.c
Byte-Lab: I think we'll want to put this in `layered_running()`. That callback is invoked just before a task starts running on a CPU, so we can look at what layer it's in and apply the frequency tuning accordingly.

As an aside, `layered_dispatch()` isn't the only place where we can choose where a task will run. `layered_select_cpu()`, which is called on the task wakeup (and task fork) path, decide which CPU a task should be migrated to before being enqueued. We can "directly dispatch" to local DSQs from there, which is conceptually the same thing as calling `scx_bpf_consume()`. See `scx_bpf_dispatch(p, SCX_DSQ_LOCAL, ...)`. In the former case, we're putting a task directly on the local DSQ where the core kernel will pull the task from to run, and in the latter case, we're taking a task from another DSQ and putting it on the local DSQ. The end result is the same -- we're specifying a task that should run next on that CPU. By instead just looking up a task's layer from `layered_running()`, we're guaranteed to hit all the cases.

If we want to speed up the context switch path, we could also reset the frequency in `layered_stopping()`, but that could have the downside of accidentally throttling a CPU that's running a high-pri layer task.

scheds/rust/scx_layered/src/main.rs
htejun: I think it'd be better to expose the raw number. Providing an interface which has higher level semantics requires a lot more knowledge than is currently available in the stack. We just don't have enough context to determine what a semantical QoS level should map to on a given hardware (e.g. the desktop processors I'm testing have only three or so perf levels and they aren't evenly spaced and I have no idea what the perf / power ratio is at each point).

Also, given how scx_layered is for hand crafted boutique setups, abstraction here can be more of a hurdle than being useful.

htejun (approved):

Byte-Lab:
Should we add logic to account for running on kernels that don't have `scx_bpf_cpuperf_set()`?

PR #240, opened 2024-04-24
layered: Add ability to match by process name

Byte-Lab:
If a library creates threads, those threads will often have the same
name. If two different processes of different priority both use a
library, it may be that we want the library's threads in each process to
be put into different layers.

To support this, let's add the ability to filter not only by task name,
but also by process name via the task thread group leader's comm.

Tested by creating two executables named "foo" and "bar", which both
spawn a bunch of tasks named "exp_worker" that spin until being
interrupted. With this config: https://pastebin.com/QSGwnkDF, the tasks
were correctly matched to the expected layers.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -598,6 +598,12 @@ static bool match_one(struct layer_match *match, struct task_struct *p, const ch
 		memcpy(comm, p->comm, MAX_COMM);
 		return match_prefix(match->comm_prefix, comm, MAX_COMM);
 	}
+	case MATCH_PCOMM_PREFIX: {
+		char pcomm[MAX_COMM];
+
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -598,6 +598,12 @@ static bool match_one(struct layer_match *match, struct task_struct *p, const ch
 		memcpy(comm, p->comm, MAX_COMM);
 		return match_prefix(match->comm_prefix, comm, MAX_COMM);
 	}
+	case MATCH_PNAME_PREFIX: {
```
htejun, line 601: Would `PCOMM` be a name which is more consistent with `COMM`?

htejun (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
Byte-Lab: Agreed, done

PR #258, opened 2024-05-01
Update README with instructions for building against local libbpf

hodgesds:
Add extra info to the README for linking against local `libbpf.a`. It may not be the preferred way of building, but still could be useful in certain situations such as testing changes against a local libbpf version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,17 @@ $ meson install -C build
 
 Notes: `meson setup` will also clone both libbpf and bpftool repos and `meson compile` will build them both.
 
+### Static linking against system libbpf
+Note, depending on your system configuration `libbpf_a` and `libbpf_h` may be
+in different directories. The system libbpf version needs to match the minimum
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -141,6 +141,16 @@ $ meson install -C build
 
 Notes: `meson setup` will also clone both libbpf and bpftool repos and `meson compile` will build them both.
 
+### Static linking against system libbpf
+Note, depending on your system configuration `libbpf.a` may be in a different
+directory.
+ ```
+$ cd $SCX
+$ meson setup build --prefix ~ -D libbpf_a=/usr/lib64/libbpf.a
```
htejun, line 149: Can you also add the include override? Depending on the version difference matching custom .a with system headers might not work too well.

hodgesds: Good catch! Did some more digging and #259 is probably a better solution compared to using the system libbpf.

PR #261, opened 2024-05-03
Make scx_rusty interactive

Byte-Lab:
## Overview

This is the first iteration that attempts to make `scx_rusty` more accommodating to interactive workloads. The patch set does the following:

- Makes `scx_rusty` a deadline scheduler, rather than a simple vtime scheduler. The deadline is calculated according to a number of factors:
    - First and foremost, the deadline is determined by a task's average runtime. Unlike with EEVDF, which sets a task's deadline based on its slice length, `scx_rusty` tracks a task's average runtime, and scales its deadline inversely according to its weight.
    - `scx_rusty` also tracks the frequency with which a task is blocked, and the frequency with which a task wakes other tasks. In the former case, the task is likely to be a consumer, and in the latter case, a producer (or both if the frequency is high for both). We calculate a `lat_prio` value for tasks when setting a deadline, which is inversely scaled by a task's block and waker frequency, and positively scaled by a task's average runtime.

While we could almost certainly go a lot further with calculating tasks' lat_prio values, this already performs quite well. More on this below.

- Updates `scx_rusty` to have a dynamic slice length. A longer slice length is used for under-utilized hosts, and a much shorter slice length is used for over-utilized hosts. These under/over util slice length values can be set when the scheduler is loaded, but are static thereafter. The scheduler will track utilization, and will adjust to use one or the other slice length depending on that util.

This is another example where we could almost certainly go further. For example, we could track slice length as a per-task construct, and e.g. throttle a task's slice when we determine that its average runtime is too high. For now, this also gives us very good results.

## Results

`scx_rusty` seems to perform remarkably well on common interactive workloads. I'll describe some of the benchmarks I ran below.

All benchmarks were run on a Ryzen 9 7950X. Each benchmark (unless stated otherwise) was run concurrently with an active Spotify session, as well as severe CPU contention via `$ stress-ng -c $((4 * $(nproc)))`:

### Terraria
Running terraria with `scx_rusty` under the above conditions results in roughly a 60-70% FPS improvement compared to EEVDF (on v6.8). See  the following video for a demonstration: https://drive.google.com/file/d/1fyHt9BYGha6apl7HAkibwpy52UTi8-AQ/view?usp=sharing

### Civilization 6
Running the standard, CPU-bound AI benchmark on Civilization 6 resulted in roughly a 2.5x improvement with `scx_rusty` over EEVDF:

#### EEVDF:
civ_6_eevdf

#### `scx_rusty`:
civ_6_rusty

If run without overcommitting the host, both schedulers appeared to perform equally.

### kcompile
Finally, I also tested doing a kcompile while the system is severely overutilized:

```
$ make CC=clang -j allyesconfig
$ /usr/bin/time make CC=clang -j $(nproc) # <-- testing this
```

#### EEVDF:
```
35:37.42elapsed
```

#### `scx_rusty`:
```
35:19.05elapsed
```

Only a sample size of one so not statistically significant, but at least indicative that this doesn't cause a regression.

# Future work

Here are some ideas for how this patch set could be expanded upon in the future:
1. Being more intentional and mathematical with how we're calculating lat_prio from waker/block frequencies and avg_runtime. The values chosen were fairly arbitrary and anecdotal. We should probably be more intentional and mathematically sound with how we're applying these.
2. Dynamically setting per-task slice, and be more fine-grained than just two global values
3. Make `scx_rusty` preempting to even further help interactive tasks

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -144,6 +144,22 @@ struct {
 	__uint(map_flags, 0);
 } task_data SEC(".maps");
 
+static struct dom_ctx *try_lookup_dom_ctx(u32 dom_id)
+{
+	return bpf_map_lookup_elem(&dom_data, &dom_id);
+}
+
+static struct dom_ctx *lookup_dom_ctx(u32 dom_id)
+{
+	struct dom_ctx *domc;
+
+	domc = try_lookup_dom_ctx(dom_id);
+	if (!domc)
+		scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+
+	return domc;
+}
+
 static struct task_ctx *lookup_task_ctx(struct task_struct *p)
 {
 	struct task_ctx *taskc;
@@ -217,11 +233,8 @@ static void dom_dcycle_adj(u32 dom_id, u32 weight, u64 now, bool runnable)
 	s64 adj = runnable ? 1 : -1;
 	u32 bucket_idx = 0;
 
-	domc = bpf_map_lookup_elem(&dom_data, &dom_id);
-	if (!domc) {
-		scx_bpf_error("Failed to lookup dom_ctx");
+	if (!(domc = lookup_dom_ctx(dom_id)))
 		return;
-	}
 
 	bucket = lookup_dom_bucket(domc, weight, &bucket_idx);
 	lockw = lookup_dom_lock(dom_id, weight);
@@ -258,14 +271,12 @@ static void dom_load_xfer_task(struct task_struct *p, struct task_ctx *taskc,
 	struct ravg_data task_dcyc_rd;
 	u64 from_dcycle[2], to_dcycle[2], task_dcycle;
 
-	from_domc = bpf_map_lookup_elem(&dom_data, &from_dom_id);
+	from_domc = lookup_dom_ctx(from_dom_id);
 	from_lockw = lookup_dom_lock(from_dom_id, weight);
-	to_domc = bpf_map_lookup_elem(&dom_data, &to_dom_id);
+	to_domc = lookup_dom_ctx(to_dom_id);
 	to_lockw = lookup_dom_lock(to_dom_id, weight);
-	if (!from_domc || !from_lockw || !to_domc || !to_lockw) {
-		scx_bpf_error("dom_ctx / lock lookup failed");
+	if (!from_domc || !from_lockw || !to_domc || !to_lockw)
 		return;
-	}
 
 	from_bucket = lookup_dom_bucket(from_domc, weight, &idx);
 	to_bucket = lookup_dom_bucket(to_domc, weight, &idx);
@@ -413,10 +424,8 @@ static void refresh_tune_params(void)
 		if (is_offline_cpu(cpu))
 			continue;
 
-		if (!(domc = bpf_map_lookup_elem(&dom_data, &dom_id))) {
-			scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+		if (!(domc = lookup_dom_ctx(dom_id)))
 			return;
-		}
 
 		if (tune_input.direct_greedy_cpumask[cpu / 64] & (1LLU << (cpu % 64))) {
 			if (direct_greedy_cpumask)
@@ -454,18 +463,16 @@ static bool task_set_domain(struct task_ctx *taskc, struct task_struct *p,
 		return false;
 	}
 
-	old_domc = bpf_map_lookup_elem(&dom_data, &old_dom_id);
-	if (!old_domc) {
-		scx_bpf_error("Failed to lookup old dom%u", old_dom_id);
+	old_domc = lookup_dom_ctx(old_dom_id);
+	if (!old_domc)
 		return false;
-	}
 
 	if (init_dsq_vtime)
 		vtime_delta = 0;
 	else
 		vtime_delta = p->scx.dsq_vtime - old_domc->vtime_now;
 
-	new_domc = bpf_map_lookup_elem(&dom_data, &new_dom_id);
+	new_domc = try_lookup_dom_ctx(new_dom_id);
 	if (!new_domc) {
 		if (new_dom_id == NO_DOM_FOUND) {
 			taskc->offline = true;
@@ -550,11 +557,9 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 			const struct cpumask *idle_cpumask;
 			bool has_idle;
 
-			domc = bpf_map_lookup_elem(&dom_data, &taskc->dom_id);
-			if (!domc) {
-				scx_bpf_error("Failed to find dom%u", taskc->dom_id);
+			domc = lookup_dom_ctx(taskc->dom_id);
+			if (!domc)
 				goto enoent;
-			}
 			d_cpumask = domc->cpumask;
 			if (!d_cpumask) {
 				scx_bpf_error("Failed to acquire dom%u cpumask kptr",
@@ -657,10 +662,8 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 		struct dom_ctx *domc;
 		struct bpf_cpumask *tmp_direct_greedy, *node_mask;
 
-		if (!(domc = bpf_map_lookup_elem(&dom_data, &dom_id))) {
-			scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+		if (!(domc = lookup_dom_ctx(dom_id)))
 			goto enoent;
-		}
 
 		tmp_direct_greedy = direct_greedy_cpumask;
 		if (!tmp_direct_greedy) {
@@ -829,11 +832,9 @@ dom_queue:
 		u32 dom_id = taskc->dom_id;
 		struct dom_ctx *domc;
 
-		domc = bpf_map_lookup_elem(&dom_data, &dom_id);
-		if (!domc) {
-			scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+		domc = lookup_dom_ctx(dom_id);
+		if (!domc)
 			return;
-		}
 
 		/*
 		 * Limit the amount of budget that an idling task can accumulate
@@ -1025,11 +1026,9 @@ void BPF_STRUCT_OPS(rusty_running, struct task_struct *p)
 	if (fifo_sched)
 		return;
 
-	domc = bpf_map_lookup_elem(&dom_data, &dom_id);
-	if (!domc) {
-		scx_bpf_error("Failed to lookup dom[%u]", dom_id);
+	domc = lookup_dom_ctx(dom_id);
+	if (!domc)
 		return;
-	}
 
 	/*
 	 * Global vtime always progresses forward as tasks start executing. The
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -317,15 +356,46 @@ static void dom_load_xfer_task(struct task_struct *p, struct task_ctx *taskc,
...
+	struct lock_wrapper* lockw;
+	u64 min_vruntime;
+
+	lockw = lookup_dom_vtime_lock(domc->id);
+	if (!lockw) {
+		return 0;
+	}
+
+	bpf_spin_lock(&lockw->lock);
+	min_vruntime = domc->min_vruntime;
```
htejun, line 379: The locking around is unlikely to make any difference when reading a single value.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -440,32 +505,225 @@ static void refresh_tune_params(void)
 	}
 }
 
+/*
+ * log2 helper functions taken from scx_lavd
+ */
+static inline __attribute__((always_inline)) u32 bpf_log2(u32 v)
```
htejun, line 511: We should collect these into a math header later.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -440,32 +505,225 @@ static void refresh_tune_params(void)
...
+};
+
+static u64 sched_prio_to_latency_weight(u64 prio)
+{
+	if (prio >= DL_MAX_LAT_PRIO) {
+		scx_bpf_error("Invalid prio index");
+		return 0;
+	}
+
+	return sched_prio_to_weight[DL_MAX_LAT_PRIO - prio - 1];
```
htejun, line 574: So, this can just return `p->scx.weight` if the rest of the code assumes [1, 10000] scale.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -440,32 +505,225 @@ static void refresh_tune_params(void)
...
+	 * of those tasks. Thus, by giving a latency boost to the producer /
+	 * consumer tasks, we optimize for the case of scheduling tasks that
+	 * are on the critical path for serial workchains, and have a
+	 * disproporionate impact on the latency of a workload.
+	 *
+	 * We multiply the frequencies of wait_freq and waker_freq somewhat
+	 * arbitrarily, based on observed performance for audio and gaming
+	 * interactive workloads.
+	 */
+	waker_freq = max_clamp(taskc->waker_freq, DL_FREQ_FT_MAX);
```
htejun, line 650: Maybe just `max()`?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -440,32 +505,225 @@ static void refresh_tune_params(void)
...
+	avg_run_raw = scale_fair(avg_run_raw, p->scx.weight);
+	avg_run = bpf_log2l(avg_run_raw + 1);
+
+	if (avg_run < lat_prio) {
+		lat_prio -= avg_run;
+	} else {
+		lat_prio = 0;
+	}
+
+	lat_scale = sched_prio_to_latency_weight(lat_prio);
```
htejun, line 684: Here, I think we're going exponential -> logarithmic -> exponential again. This isn't urgent at all but it'd be nice to remove the prio to weight table. We already have task's priority in weight in `p->scx.weight` and the boosting calculation doesn't necessarily need the table. We can either handle the exponential factors directly or use more generic exponent calculation.

htejun (approved):
This overall looks fantastic. Thanks for the excellent work.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -440,32 +505,225 @@ static void refresh_tune_params(void)
...
+	lat_scale = max_clamp(lat_scale, LB_MAX_WEIGHT);
+
+	return scale_fair(taskc->avg_runtime, lat_scale);
```
multics69, line 687: It will be nice to describe the idea behind this logic. As far as I understand, essentially the task's deadline is determined by its average runtime scaled down by weighted latency priority, which is again determined by two frequencies and runtime.

arighi:
>     * Updates `scx_rusty` to have a dynamic slice length. A longer slice length is used for under-utilized hosts, and a much shorter slice length is used for over-utilized hosts. These under/over util slice length values can be set when the scheduler is loaded, but are static thereafter. The scheduler will track utilization, and will adjust to use one or the other slice length depending on that util.
> 
> 
> This is another example where we could almost certainly go further. For example, we could track slice length as a per-task construct, and e.g. throttle a task's slice when we determine that its average runtime is too high. For now, this also gives us very good results.

Given that `scx_rusty` has the concept of domains, it'd be interesting to see if there's any benefit using a per-domain slice, calculated as a function of each individual domain's load.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -440,32 +505,225 @@ static void refresh_tune_params(void)
...
+{
+	u64 dom_vruntime, min_vruntime;
+	struct dom_ctx *domc;
+
+	if (!(domc = lookup_dom_ctx(taskc->dom_id)))
+		return;
+
+	dom_vruntime = dom_min_vruntime(domc);
+	min_vruntime = dom_vruntime - slice_ns;
+	if (vtime_before(p->scx.dsq_vtime, min_vruntime)) {
```
multics69, line 700: Hmm... this if statement is not quite intuitive to me. Could you please share the idea behind this if statement?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1080,6 +1377,11 @@ void BPF_STRUCT_OPS(rusty_set_weight, struct task_struct *p, u32 weight)
 			   taskc->weight, weight);
 
 	taskc->weight = weight;
+
+	if (fifo_sched)
+		return;
+
+	clamp_task_vtime(p, taskc, 0);
```
multics69, line 1384: I wonder if there is a case that a task is set_weight()-ed after it is enqueued.
If there are such cases, don't we need to re-order the task in the DSQ?
Otherwise, we don't need to calculate the deadline here?

multics69 (approved):
Overall, it looks good to me. Although the details are different, I am glad that the high-level idea of LAVD was adopted. Also, I like the idea of considering the task's nice weight when calculating latency priority. I left comments asking for further clarification of high-level ideas behind some logic.

ptr1337:
Tested on 7950X3D and provides a massive improvement in interactivity. Benchmarks are also looking good compared to EEVDF.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
Byte-Lab: Agreed, I'll do that in a follow on patch set just to avoid more churn on this one.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
Byte-Lab: So, the reason I did this is that _technically_ what we were doing before is undefined behavior according to the C standard:

>The execution of a program contains a data race if it contains two conflicting actions in
>different threads, at least one of which is not atomic, and neither happens before the
>other. Any such data race results in undefined behavior.

In practical terms this is of course unlikely to be a problem as long as the compiler doesn't do anything annoying like issue torn writes or reads. Given that we've gone this far without synchronizing, I can just simplify and get rid of the extra lock. Maybe in a follow-on change I can just add `SCX_READ_ONCE()` and `SCX_WRITE_ONCE()` macros to match the LKMM?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
Byte-Lab: Tejun and I spoke offline -- so the general idea of doing exponential -> logarithmic -> exponential as opposed to just using the frequencies was that it seemed more extensible to map latency priority to some integer value in a constrained range, in case we want to add any more factors down the road. I think the main thing that's a bit weird about this approach is that niceness doesn't necessarily makes sense here. It's not _unusable_, but it does constrain what we can do later with e.g. added cgroup support. So I'm going to land this in its current form, but in a follow up I'm planning to remove the niceness table indirection and come up with some other scaling function for latency priority.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -440,32 +505,225 @@ static void refresh_tune_params(void)
...
+	 * disproporionate impact on the latency of a workload.
+	 *
+	 * We multiply the frequencies of wait_freq and waker_freq somewhat
+	 * arbitrarily, based on observed performance for audio and gaming
+	 * interactive workloads.
+	 */
+	waker_freq = max_clamp(taskc->waker_freq, DL_FREQ_FT_MAX);
+	blocked_freq = max_clamp(taskc->blocked_freq, DL_FREQ_FT_MAX);
+	freq_factor = blocked_freq * waker_freq * waker_freq;
+	freq_factor = scale_fair(freq_factor, p->scx.weight);
```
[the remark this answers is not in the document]
Byte-Lab: Definitely, I'll add it. Your comment also made me realize that I meant to scale `freq_factor` _positively_ by weight rather than negatively (which we do want to do for `avg_runtime`. So I'll not only add a comment, but also test changing this to scale positively with weight to see what the effect is. Thanks!

scheds/rust/scx_rusty/src/bpf/main.bpf.c
Byte-Lab: The idea here is that we're limiting the amount of vtime budget that a task can accumulate to be at most one slice. If we didn't clamp this here, then a task could sleep for e.g. 1 day, and then come back and have essentially infinite CPU capacity until it caught up hours later. I'll add a comment explaining as much.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
Byte-Lab: I've added a comment that hopefully captures what's going on here a bit more clearly. But yes, you're exactly right. Essentially what we're doing is:

1. Computing some "lat_prio" value as a function of `blocked_freq`, `waker_freq`, and `avg_runtime`; all scaled by a task's weight. `blocked_freq` and `waker_freq` scale `lat_prio` positively, and `avg_runtime` scales `lat_prio` negatively.
2. With `lat_prio`, we compute the weight value that will ultimately be used to calculate our deadline value (relative to a task's current vtime). Let's call this weight `wi_lat`
3. We finally compute the task's deadline (i.e. CPU request length r_i, in EEVDF terms) by computing the inverse scaling of its average runtime according to `wi_lat`:

```
r_i = taskc->avg_runtime * 100 / wi_lat
```

scheds/rust/scx_rusty/src/bpf/main.bpf.c
Byte-Lab: Hmm, yes, I think you're right that we don't need to set the deadline here. A task should never have `ops.set_weight()` called while it's enqueued in a DSQ. The core scheduler will first issue a dequeue before calling `p->sched_class->set_weight()`. So I think we can just rely on the enqueue path to set the min vtime and deadline. Thanks for catching this!

scheds/rust/scx_rusty/src/bpf/main.bpf.c
multics69: That's a good idea. One thing you might need to consider is that the first exponential distribution is much more skewed than the second exponential distribution. As far as such actual data distribution characteristics are considered, it should be good and flexible for cgroup support.

PR #263, opened 2024-05-03
scx_lavd: support CPU frequency scaling

multics69:
This PR supports CPU frequency scaling in scx_lavd in a minimal form. 

We keep track of 1) utilization of each CPU and 2) _performance criticality_ of each task to know the required CPU performance (e.g., frequency) demand. The performance criticality of a task denotes how critical it is to CPU performance (frequency). Like the notion of latency criticality, we use three factors: the task's average runtime, wake-up frequency, and waken-up frequency. A task's runtime is longer, and its two frequencies are higher; the task is more performance-critical because it would be a bottleneck in the middle of the task chain.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/intf.h
+++ b/scheds/rust/scx_lavd/src/bpf/intf.h
@@ -21,8 +21,11 @@ typedef signed long s64;
 
 typedef int pid_t;
 
+#define U64_MAX ((u64)~0ULL)
+
 enum {
 	TASK_COMM_LEN = 16,
+	SCX_SLICE_INF = U64_MAX,
 };
 
 #define __kptr
@@ -49,13 +52,13 @@ enum consts {
 	NSEC_PER_USEC			= 1000ULL,
 	NSEC_PER_MSEC			= (1000ULL * NSEC_PER_USEC),
 	LAVD_TIME_ONE_SEC		= (1000ULL * NSEC_PER_MSEC),
-	LAVD_TIME_INFINITY_NS		= 0xFFFFFFFFFFFFFFFFULL,
+	LAVD_TIME_INFINITY_NS		= SCX_SLICE_INF,
 	LAVD_MAX_CAS_RETRY		= 8,
 
 	LAVD_TARGETED_LATENCY_NS	= (15 * NSEC_PER_MSEC),
 	LAVD_SLICE_MIN_NS		= (300 * NSEC_PER_USEC),/* min time slice */
 	LAVD_SLICE_MAX_NS		= (3 * NSEC_PER_MSEC),	/* max time slice */
-	LAVD_SLICE_UNDECIDED		= LAVD_TIME_INFINITY_NS,
+	LAVD_SLICE_UNDECIDED		= SCX_SLICE_INF,
 	LAVD_SLICE_GREEDY_FT		= 3,
 	LAVD_LOAD_FACTOR_ADJ		= 6,
 	LAVD_LOAD_FACTOR_MAX		= (10 * 1000),
@@ -77,6 +80,8 @@ enum consts {
 	LAVD_CPU_UTIL_INTERVAL_NS	= (100 * NSEC_PER_MSEC), /* 100 msec */
 	LAVD_CPU_ID_HERE		= ((u16)-2),
 	LAVD_CPU_ID_NONE		= ((u16)-1),
+
+	LAVD_FREQ_CPU_UTIL_THRES	= 950, /* 95.0% */
 
 	LAVD_PREEMPT_KICK_LAT_PRIO	= 18,
 	LAVD_PREEMPT_KICK_MARGIN	= (LAVD_SLICE_MIN_NS >> 1),
@@ -102,12 +107,16 @@ struct sys_cpu_util {
 
 	volatile s64	inc1k_low;	/* increment from low LC to priority mapping */
 	volatile s64	inc1k_high;	/* increment from high LC to priority mapping */
+
+	volatile u64	avg_perf_cri;	/* average performance criticality */
 };
 
 /*
  * Per-CPU context
  */
 struct cpu_ctx {
+	volatile u64	util;		/* average of the CPU utilization */
+
 	/* 
 	 * Information used to keep track of CPU utilization
 	 */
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_lavd/src/bpf/intf.h
```diff
@@ -21,8 +21,11 @@ typedef signed long s64;
 
 typedef int pid_t;
 
+#define U64_MAX ((u64)~0ULL)
+
 enum {
 	TASK_COMM_LEN = 16,
+	SCX_SLICE_INF = U64_MAX,
```
htejun, line 28: Hmm... this should be available in `vmlinux.h` but I guess there's no easy way to expose that to the userland part. I'll think about how this can be solved in a better way. Not an issue with your PR in any way.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1044,7 +1074,7 @@ static int boost_lat(struct task_struct *p, struct task_ctx *taskc,
 		     struct cpu_ctx *cpuc, bool is_wakeup)
 {
 	u64 run_time_ft = 0, wait_freq_ft = 0, wake_freq_ft = 0;
-	u64 lat_cri_raw = 0;
+	u64 lat_cri_raw, perf_cri_raw;
```
htejun, line 1077: Doesn't look like `perf_cri_raw` is used anywhere?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1761,6 +1816,44 @@ void BPF_STRUCT_OPS(lavd_dispatch, s32 cpu, struct task_struct *prev)
...
+	u64 max_load, cpu_load;
+	u32 cpuperf_target;
+
+	/*
+	 * If a task with an average performance criticality of the current CPU
+	 * consumes 100% of CPU time, we set the CPU performance to the
+	 * maximum.
+	 */
+	max_load = cutil_cur->avg_perf_cri * 1000 /* max cpu util */;
+	cpu_load = taskc->perf_cri * cpuc->util;
```
htejun, line 1831: If a CPU has a high average perf criticality, wouldn't that mean that we want the CPU to be running fast even if the current task doesn't happen to have high criticality? e.g. In a single CPU system, if there's an active task which would qualify for highest performance level, wouldn't that warrant keeping the clock high?

htejun (approved):
Generally looks good to me. Given that the task to CPU mapping is completely arbitrary now, the per-CPU determination of perf target seems like it may lead to arbitrary artifacts but maybe the future plan is to have better control over how tasks are assigned to CPUs depending on their criticality and other factors?

scheds/rust/scx_lavd/src/bpf/intf.h
multics69: Exactly, there is no easy way to expose some kernel definitions to the userland. It might be better a header into two files -- one for common between userland and kernel and another for kernel only.

htejun: Yeah, this has been bothering me for a while. I'll try to see if there's a nice solution that can be shared by everyone.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Ooops. Fixed.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Ah, it compares against the system-wide average performance criticality (`cutil_cur->avg_perf_cri`).
The key idea is as follows:

- We determine the clock frequency of a CPU using two factors: 1) the current CPU utilization (cpuc->util) and 2) the current task's performance criticality (taskc->perf_cri) compared to the system-wide average performance criticality (cutil_cur->avg_perf_cri).

- When a current CPU utilization is 100%, and the current task's performance criticality is the same as the system-wide average criticality, we set the target CPU frequency to the maximum.

- In other words, even if CPU utilization is not so high, the target CPU frequency could be high when the task's performance criticality is high enough (i.e., boosting CPU frequency). On the other hand, the target CPU frequency could be low even if CPU utilization is high when a non-performance-critical task is running (i.e., deboosting CPU frequency).

multics69:
This is just the first iteration for frequency scaling. As you correctly pointed out, the current arbitrary task for CPU mapping makes frequency management less effective. I _feel_ that clustering similar performance criticality tasks into a CPU would be more effective (e.g., triggering hardware turbo boost by under-clocking a few cores) than the conventional cache-affinity-based task placement. I am still trying to figure out how to reconcile performance-criticality-based placement vs. cache-affinity-based placement.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
htejun: For simplicity, let's assume a single CPU system. Let's say there are two tasks. One with low criticality and the other high. The low one is executing and the high one is right behind. The current logic would put the CPU in a lower clock state, right? Or is it that as the high perf criticality one is likely to have priority over the low one anyway, this mostly doesn't matter?

multics69: Yes, your understanding is correct. If it happens, the clock frequency will go up in the next tick, so I assumed that doesn't matter much. One potential solution would be calling scx_bpf_cpuperf_set() at ops.running() too. My slight worry about such an approach is the cost of scx_bpf_cpuperf_set()—actually, the cost of scaling the driver's hook, especially when task's runtime is short (say 100 usec). If the cost of scx_bpf_cpuperf_set() is cheap enough, we can call scx_bpf_cpuperf_set() in two places: at ops.running() for immediate reaction and at ops.tick() for reflecting changed system load. What do you think?

htejun: I don't have any practical experience with cpufreq, so my opinion doesn't really matter. I was just trying to read the code and imagine how it would work. In case they might be helpful, the followings are what I learned from reading kernel code and listening to someone who's experimenting with this internally:

- The schedutil and cpufreq code does have its own rate-limiting mechanism.
- But still, trying to switch the freq too often does have noticeable overhead.

The reason why I kept asking about the same question is because the in-kernel implementation seems to be 1. use per-cpu aggregate value to determine target freq 2. avoid ramping down freq if util is close to 100%. ie. it looked like one of the main goals of the in-kernel implementation is avoiding the total amount of work done compared to max-freq condition. I was just wondering whether there would be conditions where lavd's implementation would lose total amount of work performed. So, just a theoretical curiosity than anything else.

multics69: Then I think it would be better to keep the code as it is. I also considered two things (from the book). Updating the target frequency at ops.tick() is a kind of rate-limiting with a bounded delay. I will come up with more sophisticate approach after reading the kernel code more to adopt a similar (or better) strategy.

PR #275, opened 2024-05-10
topology: Check if NUMA sysfs directory not present

Byte-Lab:
We've gotten bug reports of people trying to use scx_rusty, but having it exit immediately due to not having a domain. This can happen if running in an environment where for some reason, there are no NUMA nodes in the sysfs hierarchy. For example, this was observed when someone was running scx_rusty on wsl2.

We could assume by default that there's a single NUMA node, but it'd require us to rewrite some of the parsing logic. Let's just explicitly handle the case for now.

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -441,5 +441,10 @@ fn create_numa_nodes(online_mask: &Cpumask) -> Result<Vec<Node>> {
 
         nodes.push(node);
     }
+
+    if nodes.is_empty() {
+        bail!("No node found, does sysfs directory /sys/devices/system/node exist?");
+    }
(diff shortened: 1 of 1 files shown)


ptr1337 (approved):
This seems to work fine.

User in the discord has tested this in WSL, and get following output:
```
sudo scx_rusty
Error: No node found, does sysfs directory /sys/devices/system/node exist?
```

htejun (approved):

Byte-Lab:
Sorry for the churn, but I'm actually going to close this. We've had multiple users reach out who haven't compiled with `CONFIG_NUMA`, and we might as well just do the small amount of work to support them. I'll send another PR shortly.

PR #276, opened 2024-05-10
scx_utils: report an explicit error when another scheduler is running

arighi:
If another scheduler is already running, the Rust schedulers based on scx_utils are reporting an error like the following, that can be a bit difficult to understand:

 Error: Failed to attach struct ops

 Caused by:
     bpf call "libbpf_rs::map::Map::attach_struct_ops::{{closure}}" returned NULL

Change the scx_ops_attach macro to check if another sched_ext scheduler is running and in that case report a more explicit error.

With this applied:

 $ sudo scx_rustland
 Error: another sched_ext scheduler is already running

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -10,6 +10,7 @@ use std::ffi::CStr;
 use std::ffi::CString;
 use std::mem::size_of;
 use std::slice::from_raw_parts;
+use std::io;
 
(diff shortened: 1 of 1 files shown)


sirlucjan (approved):
Great job @arighi

ptr1337:
LGTM, tested locally.

ptr1337 (approved):

PR #281, opened 2024-05-12
scx_rustland: properly support offline CPUs

### LINKED ISSUE #280: [scx_rustland] Runtime error (invalid DSQ ID), likely related to `nosmt` kernel cmdline parameter being set
Hi,
`scx_rustland` https://github.com/sched-ext/scx/commit/5de8ff5bd8d228aba4660b5b01f3831f961f9923 crashes when some of smt cores are disabled.
Tested with and without `nosmt` kernel cmdline parameter on the same kernel https://github.com/sched-ext/sched_ext/commit/5a00279e4cf4f6da92b3e021727bc145153e8547.
No problem with other scx_* schedulers [except scx_pair].

dmesg:
```
[18868.377213] sched_ext: BPF scheduler "rustland" errored, disabling
[18868.377219] sched_ext: runtime error (invalid DSQ ID 0x000000000000000a)
[18868.377221]    scx_bpf_consume+0xbd/0x110
[18868.377227]    bpf_prog_615c6e385c3ea84d_rustland_dispatch+0x184/0x1df
```
log:
```
sudo ./scx_rustland -d
21:46:34 [INFO] RustLand scheduler attached - 8 online CPUs
21:46:34 [INFO] vruntime=0
21:46:34 [INFO]   tasks=0
21:46:34 [INFO]   nr_user_dispatches=0 nr_kernel_dispatches=11
21:46:34 [INFO]   nr_cancel_dispatches=0 nr_bounce_dispatches=0
21:46:34 [INFO]   nr_waiting=1437 [nr_queued=1437 + nr_scheduled=0]
21:46:34 [INFO]   nr_failed_dispatches=0 nr_sched_congested=0 nr_page_faults=0 [OK]
21:46:34 [INFO] slice boost = 100
21:46:34 [INFO] Running tasks:
21:46:34 [INFO]   core  0 cpu  0 pid=0
21:46:34 [INFO]   core  1 cpu  2 pid=0
21:46:34 [INFO]   core  2 cpu  4 pid=0
21:46:34 [INFO]   core  3 cpu  6 pid=0
21:46:34 [INFO]   core  4 cpu  8 pid=[self]
21:46:34 [INFO]   core  5 cpu 10 pid=0
21:46:34 [INFO]   core  6 cpu 12 pid=0
21:46:34 [INFO]   core  7 cpu 14 pid=0

DEBUG DUMP

swapper/10[0] triggered exit kind 1024:
  runtime error (invalid DSQ ID 0x000000000000000a)

Backtrace:
  scx_bpf_consume+0xbd/0x110
  bpf_prog_615c6e385c3ea84d_rustland_dispatch+0x184/0x1df

CPU states
----------

CPU 6   : nr_run=2 flags=0x0 cpu_rel=0 ops_qseq=13924727 pnt_seq=23974268
          curr=scx_rustland[110404] class=ext_sched_class

 *R scx_rustland[110404] +0ms
      sticky/holding_cpu=-1/-1 dsq_id=(n/a) dsq_vtime=0
      cpus=5555

  R systemd-udevd[538] +0ms
      scx_state/flags=3/0x1 dsq_flags=0x0 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x8000000000000002 dsq_vtime=17237280334746
      cpus=ffff

    __x64_sys_epoll_wait+0x73/0x110
    do_syscall_64+0x85/0x1a0
    entry_SYSCALL_64_after_hwframe+0x6c/0x74

21:46:34 [INFO] Unregister RustLand scheduler
Error: EXIT: runtime error (invalid DSQ ID 0x000000000000000a)
```

arighi:
During the initialization phase the scheduler needs to be aware of all the available CPUs in the system (also those that are offline), in order to create a proper per-CPU DSQ for all of them.

Otherwise, if some cores are offline, we may get errors like the following:

  swapper/7[0] triggered exit kind 1024:
    runtime error (invalid DSQ ID 0x0000000000000007)

  Backtrace:
    scx_bpf_consume+0xaa/0xd0
    bpf_prog_42ff1b9d1ac5b184_rustland_dispatch+0x12b/0x187

Change the code to configure the BpfScheduler object with the total amount of CPUs available in the system and prevent such failure.

This fixes #280.

### DIFF
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -290,16 +290,16 @@ impl<'a> Scheduler<'a> {
         let init_page_faults: u64 = 0;
 
         // Low-level BPF connector.
-        let nr_online_cpus = topo_map.nr_cpus_online();
+        let nr_cpus = topo_map.nr_cpus_possible();
         let bpf = BpfScheduler::init(
             opts.slice_us,
-            nr_online_cpus as i32,
+            nr_cpus as i32,
             opts.partial,
 	    opts.exit_dump_len,
             opts.full_user,
             opts.debug,
         )?;
-        info!("{} scheduler attached - {} online CPUs", SCHEDULER_NAME, nr_online_cpus);
+        info!("{} scheduler attached - {} CPUs", SCHEDULER_NAME, nr_cpus);
 
         // Return scheduler object.
         Ok(Self {


scheds/rust/scx_rustland/src/main.rs
```diff
@@ -290,16 +290,16 @@ impl<'a> Scheduler<'a> {
         let init_page_faults: u64 = 0;
 
         // Low-level BPF connector.
-        let nr_online_cpus = topo_map.nr_cpus_online();
+        let nr_cpus = topo_map.nr_cpus_possible();
```
htejun, line 293: Note that possible cpumask can have holes and thus the number of possible CPUs may deviate from the last CPU ID + 1. In the kernel, the difference is between `num_possible_cpus` and `nr_cpu_ids`. They are usually the same but not always.

@Byte-Lab Skimming the code, it looks like while the topo code would report the correct `span` and `nr_cpus_possible`, reporting the two combo is likely to mislead users as it's easy for people to assume that the possible cpu IDs are consecutive. It'd probably be better to report `nr_cpu_ids` and document the fact that it may deviate from the number of possible CPUs.

arighi: Good point. We definitely need `nr_cpu_ids` in this case. I'm also wondering if I should just use `MAX_CPUS` to create all the DSQs, without worrying about potential holes or offline CPUs, in this way we should even be able to handle CPU hotplugging events.

htejun: `nr_cpu_ids` should do the job. CPU hotplugging can ever only happen among the possible CPUs, or you can walk the possible cpumask and create a DSQ for each possible CPU too.

PR #286, opened 2024-05-15
scx_rustland: introduce low power mode

arighi:
Introduce a "low power" mode in `scx_rustland` (provided by `scx_rustland_core`) that can be used to reduce power consumption with a minimal impact on the performance of latency-critical tasks.

The idea is to prevent CPUs from immediately waking up when they enter the idle state, if there are other CPUs active. Having other CPUs active still allows to dispatch latency-critical tasks, so system is still responsiveness, while the background CPU-intensive tasks receive an additional throttling.

This has the consequence of reducing the overall system throughput by intentionally creating bubbles in the scheduling and exploiting the additional time spent by the CPUs in idle state to save power.

The following test case summarizes the benefit of the "low power" mode:
- play a video game (Terraria) while recompiling the kernel
- measure game performance (fps) and core power consumption (W)
- compare the result of normal mode vs low-power mode
    
Result:
```
                      Game performance | Power consumption |
         ------------+-----------------+-------------------+
         normal mode |          60 fps |               6W  |
      low-power mode |          60 fps |               3W  |
```

Considering that the CPU intensive tasks are massively throttled the "low power" mode should be used only in emergency conditions (for example when a mobile system is close to run out of power), or simply to extend the battery life of mobile/portable devices.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf.rs
+++ b/rust/scx_rustland_core/assets/bpf.rs
@@ -293,6 +293,12 @@ impl<'cb> BpfScheduler<'cb> {
         }
     }
 
+    // Counter of currently running tasks.
+    #[allow(dead_code)]
+    pub fn nr_running_mut(&mut self) -> &mut u64 {
+        &mut self.skel.bss_mut().nr_running
+    }
+
     // Counter of queued tasks.
     #[allow(dead_code)]
     pub fn nr_queued_mut(&mut self) -> &mut u64 {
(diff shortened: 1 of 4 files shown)


htejun (approved):

multics69:
This is brilliant!

arighi:
Alright, I've been testing this with different combination of options and workloads and it seems pretty stable, therefore let's just merge it, we'll do further improvements in tree.

PR #290, opened 2024-05-16
Avoid redundant substraction in rsigmoid_u64

vax-r:
## Summary
Originally the implementation of function rsigmoid_u64 will perform substraction even when the value of "v" equals to the value of "max" , in which the result is certainly zero.

We can avoid this redundant substration by changing the condition from ">" to ">=" since we know when the value of "v" and "max" are equal we can return 0 without any substract operation.


Byte-Lab (approved):
LGTM, will let Changwoo stamp + merge

multics69 (approved):
Nice catch! Thanks a lot!

PR #293, opened 2024-05-18
Add missing skel.attach() calls

htejun:
C SCX_OPS_ATTACH() and rust scx_ops_attach() macros were not calling .attach() and were only attaching the struct_ops. This meant that all non-struct_ops BPF programs contained in the skels were never attached which breaks e.g. scx_layered.

Let's fix it by adding .attach() invocation the the attach macros.


anakryiko (approved):
lgtm, but "were never loaded " -> "were never attached"

Byte-Lab (approved):

PR #294, opened 2024-05-18
Fix warnings

Byte-Lab:
When compiling with warnings enabled, I noticed some bugs in scx_rusty, and stray warnings in scx_layered. To make it easier to catch such issues in the future, we:

1. Extend the build system to apply -Werror if the built-in meson --werror `meson setup` flag is specified (option is `true` by default)
2. Fix the issues in `scx_layered`, `scx_rusty`, `scx_rustland`, `scx_rlfifo`, and `scx_lavd`

The issues in `scx_rusty` are correctness issues, not just benign build issues. They seem to marginally improve interactivity, but it seems somewhat clear that rusty is really in need of preemption support to cut down on rq delays. The other fixes are all benign, but we make them anyways as otherwise the build will break for anyone using the default options.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,7 @@
 project('sched_ext schedulers', 'c',
         version: '0.1.9',
-        license: 'GPL-2.0')
+        license: 'GPL-2.0',
+        default_options: 'werror=true',)
 
 if meson.version().version_compare('<1.2')
   error('meson >= 1.2 required')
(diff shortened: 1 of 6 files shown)


htejun (approved):
We haven't don't anything to address this so let's just turn on werror by default at least until we can surface warnings automatically.

Byte-Lab:
> We haven't don't anything to address this so let's just turn on werror by default at least until we can surface warnings automatically.

Done, so also included updates to `scx_rustland` and `scx_lavd` so the build doesn't break.

multics69 (approved):
Looks good to me. Thanks!

ptr1337:
Tested Rusty and LAVD locally. All works fine.
LGTM

PR #298, opened 2024-05-20
scx_simple: re-add __COMPAT_scx_bpf_switch_all()

arighi:
Although newer kernels default to switching-all, some users might still be using the scheduler with older kernels.

Therefore, ensure all tasks are moved to the SCHED_EXT class by calling __COMPAT_scx_bpf_switch_all() during init, so that scx_simple can still operate on these older kernels as well.

Fixes: cf66e58 ("Sync from kernel (670bdab6073)")

### DIFF
--- a/scheds/c/scx_simple.bpf.c
+++ b/scheds/c/scx_simple.bpf.c
@@ -129,6 +129,7 @@ void BPF_STRUCT_OPS(simple_enable, struct task_struct *p)
 
 s32 BPF_STRUCT_OPS_SLEEPABLE(simple_init)
 {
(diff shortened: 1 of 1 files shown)


vax-r:
Mind if I ask what's the meaning and usage of switching-all ?
Doesn't the scheduler be chagned to `scx_simple` when `scx_simple` 's bpf program loaded into the kernel ?

arighi:
@vax-r it is possible to apply a scheduler only to the tasks that explicitly migrate to the `SCHED_EXT` class via `sched_setscheduler()`, or to all tasks (that is the default behavior now).

htejun (approved):

PR #302, opened 2024-05-20
Build: Make -Werror non-default

Byte-Lab:
It seems that libbpf and bpftool have some places where -Werror triggers when compiled with gcc. Let's make it non-default. Users can always enable it locally if they want to by invoking `meson setup` with the --werror flag.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,6 @@
 project('sched_ext schedulers', 'c',
         version: '0.1.9',
-        license: 'GPL-2.0',
-        default_options: 'werror=true',)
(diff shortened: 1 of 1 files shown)


arighi (approved):

sirlucjan (approved):
I am most in favor. Let's make werror as optional + add appropriate flags ( #297 and #301 ) - static libbpf is unfortunately problematic here and causes a lot of warnings, which with werror causes huge compilation problems.

multics69 (commented):
LGTM

PR #303, opened 2024-05-20
simple: Add comment explaining use of SHARED_DSQ

Byte-Lab:
scx_simple is a basic scheduler that does either basic vtime, or global FIFO, scheduling. At first glance, it may be confusing why we create a separate DSQ rather than just using SCX_DSQ_GLOBAL. Let's add a comment explaining the reason for this, so that users that are going over scx_simple as an example scheduler don't get confused.


arighi (approved):
LGTM

PR #305, opened 2024-05-21
scx_rustland: introduce fifo mode

arighi:
This is a v2 of the FIFO mode feature, more tested and tuned, based on additional experimental results (specifically to determine the optimal USERSCHED_TIMER_NS and the conditions to automatically enter and exit to/from FIFO mode).

The idea is to give an option to automatically transition to a FIFO scheduler when the system is underutilized and switch to the user-space scheduler only when the system is over commissioned.

This allows to maximize performance during regular system use, for example gaming without additional stress tests running, while also ensuring responsiveness if a CPU-intensive workload is suddenly started.

FIFO mode can lead to less predictable performance (due to the potential transitions between the scheduling policies), therefore it is provided as an optional feature that can be disabled when performance predictability is crucial, such as in real-time audio applications or during live streaming.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -172,6 +172,11 @@ struct task_ctx {
 	 * current task's cpumask.
 	 */
 	u64 cpumask_cnt;
+
+	/*
+	 * Dispatch immediately to the local DSQ.
+	 */
+	bool dispatch_local;
 };
 
 /* Map that contains task-local storage. */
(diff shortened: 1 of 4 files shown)


rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -460,41 +452,12 @@ s32 BPF_STRUCT_OPS(rustland_select_cpu, struct task_struct *p, s32 prev_cpu,
...
-	 * If we are coming from a wait state, try to find the best CPU relying
-	 * on the built-in idle selection logic (eventually migrating the
-	 * task).
+	 * Try to find the best CPU relying on the built-in idle selection
+	 * logic, eventually migrating the task and dispatching directly from
+	 * here if a CPU is available, unless full_user mode is enabled.
 	 */
 	cpu = scx_bpf_select_cpu_dfl(p, prev_cpu, wake_flags, &is_idle);
-	if (is_idle) {
+	if (is_idle && !full_user) {
```
htejun, line 480: So, this would mark the picked CPU as not idle and then if `!full_user` ignore it, which will strand the cpu for a while. It probably would make sense to test `full_user` before calling `scx_bpf_select_cpu_dfl()`.

htejun (approved):

rust/scx_rustland_core/assets/bpf/main.bpf.c
arighi: @htejun the idea here was to use the built-in logic to pick an idle CPU, but not directly dispatch it if `full_user` is specified, so that the task can be forced to go to the user-space scheduler.

Probably with `full_user` enabled it just makes more sense to simply return `prev_cpu` and let the user-space scheduler decide the CPU to use, according to its own idle tracking logic. At the end `full_user` is provided mostly for debugging purposes, so I'm not really worried about performance here.

arighi: Re-thinking more about this, I like a lot more to simply ignore the built-in idle selection logic in full-user mode, also from a design perspective `--full-user` means that *all* the scheduling decisions are delegated to the user-space scheduler, idle selection logic included.

Therefore I pushed a change on top to completely ignore the built-in idle selection logic when running in full-user mode and make this option incompatible with `--builtin-idle`.

htejun (approved):

arighi:
@tejun sorry... I just realized that I've pushed too much stuff in this PR... but it's still something that I've been tested a lot anyway, so it's not totally bad. :) If you think it's better I can revert the additional commits and create a separate PR.

The extra changes seem to mitigate the audio cracking issues that I was getting when the system is massively overloaded.

Just for the records the extra changes are the following:
 - dispatch interactive tasks on the first CPU available and non-interactive tasks on the CPU selected by the idle selection logic
 - assign the same time slice both to interactive and non-interactive tasks
 - implement a second-chance migration in `select_cpu()`: if a task has dispatched directly do not migrate it immediately but try to keep it on prev_cpu for another round

arighi:
... actually let me do things properly, I'll revert this one and will send a new one with the right commits.

PR #307, opened 2024-05-22
scx_rustland: improve audio workload and performance predictability

arighi:
A set of changes that seems to mitigate the audio cracking issues that I was experiencing when the system is massively over-commissioned.

Here's a summary of these changes:
 - dispatch interactive tasks on the first CPU available and non-interactive tasks on the CPU selected by the idle selection logic
 - assign the same time slice both to interactive and non-interactive tasks
 - implement a second-chance migration in `select_cpu()`: after a task has been directly dispatched prevent to immediately migrate it on a different CPU, but force it to stay on `prev_cpu` for another round

Tested both on an Intel Core i7 and AMD Ryzen 7 5800X 8-Core with multiple games / audio workloads. On both systems these changes seem to improve the responsiveness both in terms of fps and audio quality.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -172,6 +172,14 @@ struct task_ctx {
 	 * current task's cpumask.
 	 */
 	u64 cpumask_cnt;
+
+	/*
+	 * This flag tracks when a task is directly dispatched on a CPU. In
+	 * this case, we want to continue dispatching it on the same CPU for an
+	 * additional round to take advantage of cache locality before
+	 * considering migrating it to a different CPU.
+	 */
+	bool allow_migration;
 };
 
 /* Map that contains task-local storage. */
@@ -468,6 +476,7 @@ static void dispatch_user_scheduler(void)
 s32 BPF_STRUCT_OPS(rustland_select_cpu, struct task_struct *p, s32 prev_cpu,
 		   u64 wake_flags)
 {
+	struct task_ctx *tctx;
 	bool is_idle = false;
 	s32 cpu;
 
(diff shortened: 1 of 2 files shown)


rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -479,18 +485,41 @@ s32 BPF_STRUCT_OPS(rustland_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+	tctx = lookup_task_ctx(p);
+	if (!tctx)
+		return prev_cpu;
+
+	/*
+	 * If the previously used CPU is still available, keep using it to take
+	 * advantage of the cached working set.
+	 */
+	if (scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
+		tctx->dispatch_local = true;
```
htejun, line 497: BTW, you can now directly call `scx_bpf_dispatch()` from `select_cpu()`, so you don' t have to set `dispatch_local` to tell `enqueue()`.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -479,18 +485,41 @@ s32 BPF_STRUCT_OPS(rustland_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+		return prev_cpu;
+	}
+
+	/*
+	 * If the task was directly dispatched give it a second chance to
+	 * remain on the current CPU, instead of immediately migrating it.
+	 */
+	if (tctx->dispatch_local) {
```
htejun, line 505: And then maybe this can be renamed to clarify that it's tracking what happened the last time? Not a big difference but might make it a bit easier to follow.

Overall, yeah, other schedulers too are a lot more eager to move tasks across CPU boundaries than the kernel's. This has some benefits in terms of work conservation but at the expense of L1/2 locality. This probably is an area other schedulers can improve on too - e.g. consider overall saturation level of the system, how long the prev CPU has been busy since the waking task ran on it the last time (as a proxy for measuring cache decay), and how soon the CPU is expected to be freed up.

htejun (approved):

rust/scx_rustland_core/assets/bpf/main.bpf.c
arighi: Yeah, I can definitely keep the flag to track the direct dispatches, and then dispatch directly from `select_cpu()`... I guess my brain is just too used to the "old" dispatch_local way :)

rust/scx_rustland_core/assets/bpf/main.bpf.c
arighi: Good point about renaming the variable to clarify that it's basically used to keep track of the previous direct dispatch. Will change it and clarify also the comment.

About being less aggressive with migrations, it'd be nice also to add some topology awareness to the equation, like try to stick not just on `prev_cpu`, but on a subset of CPUs that share the L2 cache with `prev_cpu`, for example. Maybe that's something that we could even add to the in-kernel idle selection logic to improve all schedulers in general?

htejun: Yeah, teaching some topology awareness to `select_cpu_dfl()` is something on the todo list.

arighi:
Alright, renamed the flag `dispatch_local` -> `allow_migration`, changing the logic and updating the comment to make its purpose more clear, and moved all the direct dispatches to `select_cpu()`.

htejun (approved):

PR #309, opened 2024-05-23
rusty: Rework deadline as a signed sum

Byte-Lab:
Currently, a task's deadline is computed as its vtime + a scaled function of its average runtime (with its deadline being scaled down if it's more interactive). This makes sense intuitively, as we do want an interactive task to have an earlier deadline, but it also has some flaws.

For one thing, we're currently ignoring duty cycle when determining a task's deadline. This has a few implications. Firstly, because we reward tasks with higher waker and blocked frequencies due to considering them to be part of a work chain, we implicitly penalize tasks that rarely ever use the CPU because those frequencies are low. While those tasks are likely not part of a work chain, they also should get an interactivity boost just by pure virtue of not using the CPU very often. This should in theory be addressed by vruntime, but because we cap the amount of vtime that a task can accumulate to one slice, it may not be adequately reflected after a task runs for the first time.

Another problem is that we're minimizing a task's deadline if it's interactive, but we're also not really penalizing a task that's a super CPU hog by increasing its deadline. We sort of do a bit by applying a higher niceness which gives it a higher deadline for a lower weight, but its somewhat minimal considering that we're using niceness, and that the best an interactive task can do is minimize its deadline to near zero relative to its vtime.

What we really want to do is "negatively" scale an interactive task's deadline with the same magnitude as we "positively" scale a CPU-hogging task's deadline. To do this, we make two major changes to how we compute deadline:

1. Instead of using niceness, we now instead use our own straightforward scaling factor. This was chosen arbitrarily (by experimentation with some games) to be a scaling by 1000, but we can and should improve this in the future.

2. We now create a _signed_ linear latency priority factor as a sum of the three following inputs:
   - Work-chain factor (log_2 of product of blocked freq and waker freq)
   - Inverse duty cycle factor (log_2 of the inverse of a task's duty cycle -- higher duty cycle means lower factor)
   - Average runtime factor (Higher avg runtime means higher average runtime factor)

We then compute the latency priority as:

	lat_prio := Average runtime factor - (work-chain factor + duty cycle factor)

This gives us a signed value that can be negative. With this, we can compute a non-negative weight value by calculating a weight from the absolute value of lat_prio, and use this to scale slice_ns. If lat_prio is negative we calculate a task's deadline as its vtime MINUS its scaled slice_ns, and if it's positive, it's the task's vtime PLUS scaled slice_ns.

This ends up working well because you get a higher weight both for highly interactive tasks, and highly CPU-hogging / non-interactive tasks, which lets you scale a task's deadline "more negatively" for interactive tasks, and "more positively" for the CPU hogs.

With this change, we get a significant improvement in FPS. On a 7950X, if I run the following workload:

	$ stress-ng -c $((8 * $(nproc)))

1. I get 60 FPS when playing Stellaris (while time is progressing at max speed), whereas EEVDF gets 6-7 FPS.

2. I get ~15-40 FPS while playing Civ6, whereas EEVDF seems to get < 1 FPS. The Civ6 benchmark doesn't even start after over 4 minutes in the initial frame with EEVDF, but gets us 13s / turn with rusty.

3. It seems that EEVDF has improved with Terraria in v6.9. It was able to maintain ~30-55 FPS, as opposed to the ~5-10FPS we've seen in the past. rusty is still able to maintain a solid 60-62FPS consistently with no problem, however.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/intf.h
+++ b/scheds/rust/scx_rusty/src/bpf/intf.h
@@ -17,6 +17,7 @@
 typedef unsigned char u8;
 typedef unsigned int u32;
 typedef unsigned long long u64;
+typedef long long s64;
 #endif
 
 #include <scx/ravg.bpf.h>
@@ -43,12 +44,29 @@ enum consts {
 	NSEC_PER_SEC            = NSEC_PER_USEC * USEC_PER_SEC,
 
 	/* Constants used for determining a task's deadline */
-	DL_RUNTIME_SCALE	= 2, /* roughly scales average runtime to */
-				     /* same order of magnitude as waker  */
-				     /* and blocked frequencies */
-	DL_MAX_LATENCY_NS	= (50 * NSEC_PER_MSEC),
+	DL_RUNTIME_FACTOR	= 100,
 	DL_FREQ_FT_MAX		= 100000,
 	DL_MAX_LAT_PRIO		= 39,
+	/*
+	 * Duty cycle is the proportion of time in some interval during which a
+	 * task could have used CPU. In other words, for some time interval
+	 * [t_0, t_1], a task's duty cycle is the value between [0, 1]
+	 * corresponding to how much time in the interval it _could_ have used.
+	 * For example, if it could have run the entire time its duty cycle
+	 * would be 1, and if it could only have run for half of the interval
+	 * its duty cycle would be .5. Note that a task that's runnable for the
+	 * first half of the interval, and running for the second half at which
+	 * point it blocks, would still have a duty cycle of .5.
+	 *
+	 * We track duty cycle in BPF using the running average helpers. The
+	 * value below is approximately the value that is observed in the
+	 * kernel when a task has duty cycle ~= 1. In user space we can do
+	 * floating point arithmetic in e.g. the rust implementation of
+	 * ravg_read(), but here in the kernel we're stuck with fixed-point
+	 * integer arithmetic and must therefore use values like this when
+	 * determining a task's latency priority when calculating its deadline.
+	 */
+	DL_FULL_DCYCLE		= 650000,
 
 	/*
 	 * When userspace load balancer is trying to determine the tasks to push
@@ -115,6 +133,12 @@ struct task_ctx {
 	/* frequency with which a task wakes other tasks (producer) */
 	u64 waker_freq;
 	u64 last_woke_at;
+
+	/*
+	 * vruntime tracked by the scheduler. Separate from p->scx.dsq_vtime as
+	 * this is set by the scheduler.
+	 */
+	u64 vruntime;
 
 	/* The task is a workqueue worker thread */
 	bool is_kworker;
--- a/scheds/rust/scx_rusty/src/bpf/intf.h
+++ b/scheds/rust/scx_rusty/src/bpf/intf.h
@@ -134,6 +134,8 @@ struct task_ctx {
 	u64 waker_freq;
 	u64 last_woke_at;
 
+	s64 lat_prio;
+
 	/*
 	 * vruntime tracked by the scheduler. Separate from p->scx.dsq_vtime as
 	 * this is set by the scheduler.
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -705,8 +705,52 @@ static s64 task_compute_dl(struct task_struct *p, struct task_ctx *taskc,
 	 * interactivity scores and a higher average runtime.
 	 */
 	lat_prio = avg_run_linear - inv_dcycle_linear - work_chain_linear;
+
+	/*
+	 * If the task has been previously been assigned a lower (meaning
+	 * heavier) latency priority than what we're computing now, allow it to
+	 * temporarily inherit that lower latency priority.
+	 *
+	 * This is necessary to avoid the case of a CPU-hogging task being
+	 * penalized for latency, when in fact it's on the critical path for
+	 * running an interactive workload.
+	 *
+	 * For example, consider a workload where there's a game that uses a
+	 * CPU-hogging worker thread to crunch through logic that's required to
+	 * progress the game and properly render a frame (e.g. think Factorio,
+	 * Satisfactory, etc). The task will be CPU-intensive and will run with
+	 * high duty cycle, and will also have fairly low wake/block
+	 * frequencies due to it running for high duty cycle, but it's still
+	 * just as critical for avoiding jitter as a render thread.
+	 *
+	 * To accommodate such workloads, we allow a task to inherit the
+	 * latency priority of another task if it either wakes, or was woken,
+	 * by the other task.
+	 *
+	 * Finally, note that this also implements something of a decay
+	 * function for tasks that were previously interactive, but are no
+	 * longer. For example, if a task was assigned a latency priority of
+	 * -10, and then a latency priority of 5, it will run with the -10
+	 *  priority for this scheduling cycle, but will have to run at 5 in
+	 *  the next cycle if it's still high.
+	 *
+	 * As a future improvement, we may want to look into having some kind
+	 * of running average tracking of latency priority that decays over a
+	 * time period rather than hard-setting it on the wakeup path, and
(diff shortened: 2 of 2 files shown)


Byte-Lab:
Multiple pushes was from adjusting the `sched_prio_to_latency_weight()` multiplier. 1000 seems to work best for now, but definitely something to further improve on.

scheds/rust/scx_rusty/src/bpf/intf.h
```diff
@@ -43,12 +44,10 @@ enum consts {
...
 
 	/* Constants used for determining a task's deadline */
-	DL_RUNTIME_SCALE	= 2, /* roughly scales average runtime to */
-				     /* same order of magnitude as waker  */
-				     /* and blocked frequencies */
-	DL_MAX_LATENCY_NS	= (50 * NSEC_PER_MSEC),
+	DL_RUNTIME_FACTOR	= 100,
 	DL_FREQ_FT_MAX		= 100000,
 	DL_MAX_LAT_PRIO		= 39,
+	DL_FULL_DCYCLE		= 650000,
```
arighi, line 50: How did you determine the optimal value fo `DL_FULL_DCYCLE`? I'm wondering if this should be a function of the time slice... like, what happens if I start rusty using a different slice_ns?

arighi (approved):
Nice work, I see a massive improvement in responsiveness when the system is overloaded!

I left a couple of comments, mostly about clarifying the code, but definitely approved!

scheds/rust/scx_rusty/src/bpf/intf.h
multics69: It would be a good idea to document how to determine DL_FULL_DCYCLE (or idea behind the DL_FULL_DCYCLE), which, I guess, can be determined relatively to the maximum time slice.

multics69 (approved):
Thank you for the fantastic work! Reflecting duty cycles maches well with my observation in EEVDF and LAVD. While LAVD does not directly consider the duty cycle, after adding eligibility, I found that LAVD works better under highly loaded system. Since CPU hoggers are mostly ineligible, eligibility might have similar effect to duty cycle. However, the duty cycle is more direct measurement so I like it!

multics69:
I ran the changes in my usual test case. It seems to have a regression.
Here are debug dumps.

```
DEBUG DUMP                                                                                                                                                                                                           
                                                                                                                                                                                                                     
kworker/u64:0[43915] triggered exit kind 1026:                                                                                                                                                                       
  runnable task stall (cc1[73847] failed to run for 11.390s)                                                                                                                                                         
                                                                                                                                                                                                                     
Backtrace:                                                                                                                                                                                                           
  scx_watchdog_workfn+0x146/0x1d0                                                                                                                                                                                    
  process_one_work+0x193/0x3c0
  worker_thread+0x3a9/0x4f0
  kthread+0xd2/0x100                                 
  ret_from_fork+0x34/0x50                            
  ret_from_fork_asm+0x1a/0x30

Runqueue states                                      
---------------                                      

CPU 0   : nr_run=7 flags=0x0 cpu_rel=0 ops_qseq=150024 pnt_seq=208897
          curr=dav1d-worker[44211] class=ext_sched_class

 *R dav1d-worker[44211] -4ms
      cpus=ffff                                      


  R cc1[73847] -11390ms                              
      scx_state/flags=3/0x1 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x0
      cpus=ffff                                      

  R cc1[73793] -2910ms                               
      scx_state/flags=3/0x1 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x0
      cpus=ffff                                      

  R cc1[74207] -2904ms                               
      scx_state/flags=3/0x1 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x0
      cpus=ffff                                      

```

```
DEBUG DUMP                                                                                                                                                                                                           
                                                                                                                                                                                                                     
kworker/u64:5[13216] triggered exit kind 1026:                                                                                                                                                                       
  runnable task stall (cc1[40964] failed to run for 11.790s)                                                                                                                                                         
                                                                                                                                                                                                                     
Backtrace:                                                                                                                                                                                                           
  scx_watchdog_workfn+0x146/0x1d0                                                                                                                                                                                    
  process_one_work+0x193/0x3c0                                                                                                                                                                                       
  worker_thread+0x3a9/0x4f0                                                                                                                                                                                          
  kthread+0xd2/0x100                                                                                                                                                                                                 
  ret_from_fork+0x34/0x50                                                                                                                                                                                            
  ret_from_fork_asm+0x1a/0x30                                                                                                                                                                                        
                                                                                                                                                                                                                     
Runqueue states                                                                                                                                                                                                      
---------------                                                                                                                                                                                                      
                                                                                                                                                                                                                     
CPU 0   : nr_run=16 flags=0x0 cpu_rel=0 ops_qseq=36886 pnt_seq=71515                                                                                                                                                 
          curr=cc1[43395] class=ext_sched_class                                                                                                                                                                      
                                                                                                                                                                                                                     
 *R cc1[43395] -4ms                                                                                                                                                                                                  
      cpus=ffff                                      


  R cc1[40964] -11790ms                              
      scx_state/flags=3/0x1 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x0
      cpus=ffff                                      

  R cc1[41146] -11124ms                              
      scx_state/flags=3/0x1 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x0
      cpus=ffff                                      

  R cc1[41695] -7540ms                               
      scx_state/flags=3/0x1 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x0
      cpus=ffff                                      

```

multics69:
What I did to trigger the above is as follow:

1. Run scx_rusty
2. Play a YouTube streaming on a browser (in my case, I played [this](https://www.youtube.com/watch?v=Mc7axL9dXpU) on my Edge). 
3. Run `glxgears`
4. Kernel compile with `make -j$((8 * $(nproc)))`

htejun (approved):

scheds/rust/scx_rusty/src/bpf/intf.h
Byte-Lab: So, I actually determined it from looking at the output of a task with duty cycle approximately = 1. The rust implementation gives a proper float, but we're stuck with using fixed-point integer arithmetic in the kernel. I'll add a comment explaining this.

Byte-Lab: So, it's not really related to time slice. It tracks the period of time during which a task is runnable, as a proxy for it running / its duty cycle in some time interval. We accumulate these in rusty_quiescent() and rusty_runnable(). I'll add a comment with a bit more explanation.

Byte-Lab:
Pushed code that addresses @arighi and @multics69's feedback. Going to take a look at the regression that Changwoo pointed out.

vax-r:
Incredible works you guys are doing , I'm just here to learn what you changed and the reasons.
Must to say I learn a lot from this PR , it's amazing .

ptr1337:
Actually, this PR seems to introduce some regression for some users in a specific workload.

The user did following:
- Kernel Compilation running in background with all cores
- Rocket League with Metrics

Following version has been used for testing:
- scx-scheds 0.19-5 (CachyOS repository, based on latest commit of this repository right now)
- scx-scheds-git (latest commit, including this PR)

Here the video can be found. There is a quite huge interactivity/FPS drop visible in Rocket Leauge:
https://streamable.com/0jz2zu

Byte-Lab:
Alright, this seems to beat the old rusty on RocketLeague now as well. Interestingly, EEVDF on v6.10 beats both versions of rusty on RocketLeague, but still gets trounced on Terraria and Civ6. On Civ6 especially the game becomes unplayable with EEVDF under heavy background load, but is still fairly responsive with rusty.

This isn't perfect, but I'm going to merge it for now and we can continue to iterate in tree.

Byte-Lab:
Getting reports that this causes some regressions in cachy, so lemme revert.

Issue #308, opened 2024-05-22
scx_rustland_core: Trying to allocate too large for scx_dsp_ctx from percpu allocator


RinHizakura:
I was trying to play with the `scx_rustland` scheduler on the `sched_ext` kernel, but an error happened to me.

```
$ sudo ./scx_rustland
Error: Failed to attach struct_ops BPF programs

Caused by:
    bpf call "libbpf_rs::map::Map::attach_struct_ops::{{closure}}" returned NULL
```

Also from `dmesg` we have some information for the issue.

```
[ 2605.866281] ------------[ cut here ]------------
[ 2605.866320] illegal size (32792) or align (8) for percpu allocation
[ 2605.866348] WARNING: CPU: 9 PID: 592 at mm/percpu.c:1754 pcpu_alloc+0x82b/0x920
[ 2605.866379] Modules linked in:
[ 2605.866396] CPU: 9 PID: 592 Comm: scx_rustland Tainted: G        W          6.9.0-rc7-g0042b59beb1b #7
[ 2605.866406] Hardware name: QEMU Standard PC (i440FX + PIIX, 1996), BIOS 1.15.0-1 04/01/2014
[ 2605.866414] Sched_ext: rustland (prepping)
[ 2605.866419] RIP: 0010:pcpu_alloc+0x82b/0x920
[ 2605.866430] Code: c6 d3 96 5f 8e e9 19 ff ff ff f7 c3 00 20 00 00 0f 85 6d ff ff ff 90 48 c7 c7 be 65 52 8e 48 89 ee 4c 89 ea e8 76 0c dc ff 90 <0f> 0b 90 90 e9 50 ff ff ff e8 c7 f1 da 00 ff 0d 71 33 08 02 0f 85
[ 2605.866439] RSP: 0018:ffffb27d0020bc30 EFLAGS: 00000246
[ 2605.866446] RAX: 96301220be868d00 RBX: 0000000000000cc0 RCX: ffffffff8ee55000
[ 2605.866451] RDX: 0000000000000002 RSI: 00000000ffffdfff RDI: ffffffff8ee85290
[ 2605.866455] RBP: 0000000000008018 R08: 0000000000001fff R09: ffffffff8ee55290
[ 2605.866460] R10: 0000000000005ffd R11: 0000000000000004 R12: 0000000000008000
[ 2605.866464] R13: 0000000000000008 R14: 0000000000000000 R15: 000000000000801b
[...]

PR #310, opened 2024-05-23
Reduce MAX_ENQUEUED_TASKS to fit percpu allocator

RinHizakura:
The setting of `ops->dispatch_max_batch` leads to a too large allocated size for percpu allocator, and it will be unhappy if we want a size larger than `PCPU_MIN_UNIT_SIZE`. Reduce `MAX_ENQUEUED_TASKS` for fix.

The details for the issue can be found at #308.


arighi (approved):
Looks good to me. I also stress-tested it with a `stress-ng -c 4096` and the scheduler (and the system) survived without any problem. Thanks!

PR #312, opened 2024-05-23
scx_layered: Improve affn_viol handling and implement dump method

htejun:
- Any tasks with custom affinity which are enqueued outside of wakeup path were considered to be `affn_viol` and queued on the global DSQ. This is unnecessary. The per-layer DSQs and preemption path can handle tasks with custom affinities just fine for tasks in open or grouped layers. Update enqueue to better handle tasks with custom affinities.
- `affn_viol` tasks were queued onto `SCX_DSQ_GLOBAL` which has precedence over all user and thus per-layer DSQs. It doesn't make sense to boost tasks which are in violation. Let's create user fallback DSQs and use them to handle affinity violations.
- Add dump method to aid debugging.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -234,7 +234,8 @@ struct task_ctx {
 	struct bpf_cpumask __kptr *layered_cpumask;
 
 	bool			all_cpus_allowed;
-	u64			started_running_at;
+	u64			runnable_at;
+	u64			running_at;
 };
 
 struct {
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -563,6 +589,8 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 		if (layers[idx].preempt && scx_bpf_consume(idx))
 			return;
 
+	scx_bpf_consume(HI_FALLBACK_DSQ);
```
dschatzberg, line 592: What's the rationale for not checking the return value here and returning? Elsewhere you are only ever consuming a single task which makes this all a strict priority, but here you're falling through.

dschatzberg (approved):
Mostly looks reasonable to me, see question inline regarding consume()

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: Yeah, that's wrong. Will fix.

PR #322, opened 2024-05-27
scx_lavd: Adding READ_ONCE()/WRITE_ONCE() macros

vax-r:
In order to prevent compiler from merging or refetching load/store operations or unwanted reordering, we take the implemetation of READ_ONCE()/WRITE_ONCE() from kernel sources and add them into `scheds/include/scx/common.bpf.h`.

Use WRITE_ONCE() in function flip_sys_cpu_util() to ensure the compiler doesn't perform unnecessary optimization so the modification of bit flipping can be visible to readers as fast as possible.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -151,7 +151,7 @@ char _license[] SEC("license") = "GPL";
 volatile u64			nr_cpus_onln;
 
 static struct sys_cpu_util	__sys_cpu_util[2];
-static volatile int		__sys_cpu_util_idx;
+static int		__sys_cpu_util_idx;
 
 const volatile bool		no_freq_scaling;
 const volatile u8		verbose;
@@ -506,21 +506,21 @@ static struct cpu_ctx *get_cpu_ctx_id(s32 cpu_id)
 
 static struct sys_cpu_util *get_sys_cpu_util_cur(void)
 {
-	if (__sys_cpu_util_idx == 0)
+	if (READ_ONCE(__sys_cpu_util_idx) == 0)
 		return &__sys_cpu_util[0];
 	return &__sys_cpu_util[1];
 }
 
 static struct sys_cpu_util *get_sys_cpu_util_next(void)
 {
-	if (__sys_cpu_util_idx == 0)
+	if (READ_ONCE(__sys_cpu_util_idx) == 0)
 		return &__sys_cpu_util[1];
 	return &__sys_cpu_util[0];
 }
 
 static void flip_sys_cpu_util(void)
 {
-	__sys_cpu_util_idx ^= 0x1;
+	WRITE_ONCE(__sys_cpu_util_idx, __sys_cpu_util_idx ^ 0x1);
 }
 
 static __attribute__((always_inline))
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -278,6 +278,61 @@ static inline u32 bpf_log2l(u64 v)
 #define unlikely(x) __builtin_expect(!!(x), 0)
 #define __maybe_unused __attribute__((__unused__))
 
+/*
+ * The folloing functions are taken from kernel sources
+ * under /include/asm-generic/rwonce.h in order to prevent
+ * compiler from refetching reads or writes, also forbids
+ * compiler from reordering successive reads or writes.
+ */
+
+typedef __u8  __attribute__((__may_alias__))  __u8_alias_t;
+typedef __u16 __attribute__((__may_alias__)) __u16_alias_t;
+typedef __u32 __attribute__((__may_alias__)) __u32_alias_t;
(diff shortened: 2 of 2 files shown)


htejun (approved):

htejun (changes requested):
This looks great in general. A few points:

- Let's add a comment explaining that this was lifted from the kernel tree.
- The use of `WRITE_ONCE()` doesn't have too much to do with the "speed" of the update but more correctness of it. Without `WRITE_ONCE()`, the compiler for example is allowed to generate split updates or skip them altogether.
- It's often necessary and a best practice to pair `WRITE_ONCE()` usages with `READ_ONCE()`. If the variable is being updated in a way that the compiler may not be expecting, the compiler will most likely be able to make incorrect assumptions when reading it too. So, please also update the reading of the variable.

vax-r:
> This looks great in general. A few points:
> 
> * Let's add a comment explaining that this was lifted from the kernel tree.
> * The use of `WRITE_ONCE()` doesn't have too much to do with the "speed" of the update but more correctness of it. Without `WRITE_ONCE()`, the compiler for example is allowed to generate split updates or skip them altogether.
> * It's often necessary and a best practice to pair `WRITE_ONCE()` usages with `READ_ONCE()`. If the variable is being updated in a way that the compiler may not be expecting, the compiler will most likely be able to make incorrect assumptions when reading it too. So, please also update the reading of the variable.

@htejun - Thanks for the suggestions ! I've made the corresponding changes , please take a look if you're available.

Btw, I want to ask what's the actual difference between the GNU built-in functions "__syn_* and the `READ_ONCE()/WRITE_ONCE()` macros ? as I read the content in [memory-barrier.txt](https://www.kernel.org/doc/Documentation/memory-barriers.txt) , I thought `READ_ONCE()/WRITE_ONCE()` can make gaurantees on execution order, also the "happen-before" relationship between different statements . Can't it be consider as a memory barrier with relaxed memory order ?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -506,21 +506,21 @@ static struct cpu_ctx *get_cpu_ctx_id(s32 cpu_id)
...
-	if (__sys_cpu_util_idx == 0)
+	if (READ_ONCE(__sys_cpu_util_idx) == 0)
 		return &__sys_cpu_util[1];
 	return &__sys_cpu_util[0];
 }
 
 static void flip_sys_cpu_util(void)
 {
-	__sys_cpu_util_idx ^= 0x1;
+	WRITE_ONCE(__sys_cpu_util_idx, __sys_cpu_util_idx ^ 0x1);
```
multics69, line 523: READ_ONCE() should be used for the XOR op.

multics69 (changes requested):
READ_ONCE()/WRITE_ONCE() is a macro for `volatile` qualifier. It enforces compiler generate a memory load/store (not register load/store). However, it does not generate hardware atomic instruction, which is generated by the built-in atomics.

Once every access of the variable is replaced by READ/WRITE_ONCE(), you can drop the volatile qualifier in the variable definition.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
vax-r: You mean we need to add `READ_ONCE(__sys_cpu_util_idx);` to ensure the load operation of the value of `__sys_cpu_util_idx` is sequenced-before the XOR operation on it right ?

multics69: Yup

htejun: So, if you wanna use `READ_ONCE()`, it should look like `WRITE_ONCE(__sys_cpu_util_idx, READ_ONCE(__sys_cpu_util_idx ^ 0x1))`. Adding `READ_ONCE()` as a separate statement doesn't really achieve anything.

However, because the writer is already guaranteed to see the latest value (it's the sole writer and no optimization can break update visibility from oneself), so I don't think `READ_ONCE()` is necessary here. These volatile updates express the relationships between unsynchronized the writer and its readers. Even when there are multiple writers, they're either usually synchronized among themselves or don't really care what other writers are doing.

vax-r: @htejun 
I made the requested changes, thanks for the advise !
By the way, I've removed the `volatile` qualifier , so I think `READ_ONCE()` is still necessary here ?

htejun:
Oh yeah, `READ_ONCE()` should only be on the variable itself but really no need to have it in the write path. There are no ordering concerns here.

vax-r:
> Oh yeah, `READ_ONCE()` should only be on the variable itself but really no need to have it in the write path. There are no ordering concerns here.

I see, I've made the requested changes. Please have a look while you're available, thanks !

multics69 (approved):

htejun (approved):

htejun (approved):

PR #329, opened 2024-06-03
scx: Unify loading and running boilerplate across rust schedulers

htejun:
Make restart handling with user_exit_info simpler and consistently use the load and report macros consistently across the rust schedulers. This makes all schedulers automatically handle auto restarts from CPU hotplug events. Note that this is necessary even for scx_lavd which has CPU hotplug operations as CPU hotplug operations which took place between skel open and scheduler init can still trigger restart.


htejun:
@arighi The commit ends up changing unrelated formatting in `rlfifo`. If this doesn't agree with your `cargo fmt`, let's come up with a way to sync that.

multics69 (approved):
LGTM. Thanks!

Byte-Lab (approved):

arighi (approved):
LGTM

PR #338, opened 2024-06-07
scx_lavd: core compaction for low power consumption

multics69:
When system-wide CPU utilization is low, it is very likely all the CPUs are running with very low utilization. That means all CPUs run with low clock frequency thanks to dynamic frequency scaling and very frequently go in and out from/to C-state. That results in low performance (i.e., low clock frequency) and high power consumption (i.e., frequent P-/C-state transition).
    
The idea of *core compaction* is using less number of CPUs when   system-wide CPU utilization is low. The chosen cores (called "active   cores") will run in higher utilization and higher clock frequency, and the rest of the cores (called "idle cores") will be in a C-state for a much longer duration. Thus, the core compaction can achieve higher  performance with lower power consumption.
    
One potential problem of core compaction is latency spikes when all the active cores are overloaded. A few techniques are incorporated to solve this problem.
    
1) Limit the active CPU core's utilization below a certain limit (say 50%).
    
2) Do not use the core compaction when the system-wide utilization is
 moderate (say 50%).
 
3) Do not enforce the core compaction for kernel and pinned user-space tasks since they are manually optimized for performance.
    
In my experiments, under a wide range of system-wide CPU utilization
(5%—80%), the core compaction reduces 7-20% power consumption without sacrificing average and 99p tail latency.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/intf.h
+++ b/scheds/rust/scx_lavd/src/bpf/intf.h
@@ -77,21 +77,29 @@ enum consts {
 	LAVD_ELIGIBLE_TIME_MAX		= (LAVD_SLICE_MIN_NS >> 8),
 
 	LAVD_CPU_UTIL_MAX		= 1000, /* 100.0% */
-	LAVD_CPU_UTIL_INTERVAL_NS	= (100 * NSEC_PER_MSEC), /* 100 msec */
+	LAVD_CPU_UTIL_INTERVAL_NS	= (25 * NSEC_PER_MSEC),
 	LAVD_CPU_ID_HERE		= ((u32)-2),
 	LAVD_CPU_ID_NONE		= ((u32)-1),
+	LAVD_CPU_ID_MAX			= 512,
 
 	LAVD_PREEMPT_KICK_LAT_PRIO	= 15,
 	LAVD_PREEMPT_KICK_MARGIN	= (LAVD_SLICE_MIN_NS >> 3),
 	LAVD_PREEMPT_TICK_MARGIN	= (LAVD_SLICE_MIN_NS >> 8),
 
+	LAVE_TC_PER_CORE_MAX_CTUIL	= 500, /* maximum per-core CPU utilization */
+	LAVD_TC_NR_ACTIVE_MIN		= 1, /* num of mininum active cores */
+	LAVD_TC_NR_OVRFLW		= 1, /* num of overflow cores */
+	LAVD_TC_CPU_PIN_INTERVAL	= (100 * NSEC_PER_MSEC),
+	LAVD_TC_CPU_PIN_INTERVAL_DIV	= (LAVD_TC_CPU_PIN_INTERVAL /
+					   LAVD_CPU_UTIL_INTERVAL_NS),
+
 	LAVD_GLOBAL_DSQ			= 0,
 };
 
 /*
- * System-wide CPU utilization
+ * System-wide stats
  */
-struct sys_cpu_util {
+struct sys_stat {
 	volatile u64	last_update_clk;
 	volatile u64	util;		/* average of the CPU utilization */
 	volatile u64	load_factor;	/* system load in % (1000 = 100%) for running all runnables within a LAVD_TARGETED_LATENCY_NS */
@@ -108,17 +116,19 @@ struct sys_cpu_util {
 	volatile s64	inc1k_high;	/* increment from high LC to priority mapping */
 
 	volatile u64	avg_perf_cri;	/* average performance criticality */
+
+	volatile u64	nr_violation;	/* number of utilization violation */
+	volatile int	nr_active;	/* number of active cores */
 };
 
 /*
  * Per-CPU context
  */
 struct cpu_ctx {
-	volatile u64	util;		/* average of the CPU utilization */
-
 	/* 
 	 * Information used to keep track of CPU utilization
 	 */
+	volatile u64	util;		/* average of the CPU utilization */
 	volatile u64	idle_total;	/* total idle time so far */
 	volatile u64	idle_start_clk;	/* when the CPU becomes idle */
 
@@ -150,6 +160,14 @@ struct cpu_ctx {
 	volatile u16	lat_prio;	/* latency priority */
 	volatile u8	is_online;	/* is this CPU online? */
 	s32		cpu_id;		/* cpu id */
+
+	/*
+	 * Fields for core compaction
+	 *
+	 * NOTE: The followings MUST be placed at the end of this struct.
+	 */
+	struct bpf_cpumask __kptr *tmp_a_mask;	/* temporary cpu mask */
+	struct bpf_cpumask __kptr *tmp_o_mask;	/* temporary cpu mask */
 };
 
 struct task_ctx {
@@ -199,6 +217,7 @@ struct task_ctx_x {
 	u64	sys_load_factor; /* system load factor in [0..100..] */
 	u64	avg_lat_cri;	/* average latency criticality */
 	u64	avg_perf_cri;	/* average performance criticality */
+	u32	nr_active;	/* number of active cores */
 };
 
 
--- a/scheds/rust/scx_lavd/src/bpf/intf.h
+++ b/scheds/rust/scx_lavd/src/bpf/intf.h
@@ -86,7 +86,7 @@ enum consts {
 	LAVD_PREEMPT_KICK_MARGIN	= (LAVD_SLICE_MIN_NS >> 3),
 	LAVD_PREEMPT_TICK_MARGIN	= (LAVD_SLICE_MIN_NS >> 8),
 
-	LAVE_TC_PER_CORE_MAX_CTUIL	= 500, /* maximum per-core CPU utilization */
+	LAVD_TC_PER_CORE_MAX_CTUIL	= 500, /* maximum per-core CPU utilization */
 	LAVD_TC_NR_ACTIVE_MIN		= 1, /* num of mininum active cores */
 	LAVD_TC_NR_OVRFLW		= 1, /* num of overflow cores */
 	LAVD_TC_CPU_PIN_INTERVAL	= (100 * NSEC_PER_MSEC),
--- a/scheds/rust/scx_lavd/src/bpf/intf.h
+++ b/scheds/rust/scx_lavd/src/bpf/intf.h
@@ -49,6 +49,7 @@ extern void bpf_iter_task_destroy(struct bpf_iter_task *it) __weak __ksym;
  */
 enum consts {
 	CLOCK_BOOTTIME			= 7,
+	CACHELINE_SIZE			= 64,
 	NSEC_PER_USEC			= 1000ULL,
 	NSEC_PER_MSEC			= (1000ULL * NSEC_PER_USEC),
 	LAVD_TIME_ONE_SEC		= (1000ULL * NSEC_PER_MSEC),
@@ -168,7 +169,7 @@ struct cpu_ctx {
 	 */
 	struct bpf_cpumask __kptr *tmp_a_mask;	/* temporary cpu mask */
 	struct bpf_cpumask __kptr *tmp_o_mask;	/* temporary cpu mask */
-};
+} __attribute__((aligned(CACHELINE_SIZE)));
 
 struct task_ctx {
 	/*
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -1822,6 +1822,35 @@ static void put_local_rq_no_fail(struct task_struct *p, struct task_ctx *taskc,
 	scx_bpf_dispatch(p, SCX_DSQ_LOCAL, LAVD_SLICE_UNDECIDED, enq_flags);
 }
 
+static s32 pick_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags,
+		      bool *is_idle)
+{
+	struct bpf_cpumask *p_mask = p->cpus_ptr;
+	s32 cpu_id;
+
+	/*
+	 * Pick a fully idle core.
+	 */
+	cpu_id = scx_bpf_pick_idle_cpu(p_mask, SCX_PICK_IDLE_CORE);
+	if (cpu_id >= 0)
+		goto bingo;
+
+	/*
+	 * If there is no fully idle core, pick any idle core.
+	 */
+	cpu_id = scx_bpf_pick_idle_cpu(p_mask, 0);
+	if (cpu_id >= 0)
+		goto bingo;
+
+	/*
+	 * If there is no idle core, stay on the previous core.
+	 */
+	cpu_id = prev_cpu;
+
+bingo:
+	return cpu_id;
+}
+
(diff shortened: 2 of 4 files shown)


scheds/rust/scx_lavd/src/bpf/intf.h
```diff
@@ -77,21 +77,29 @@ enum consts {
...
+	LAVD_CPU_UTIL_INTERVAL_NS	= (25 * NSEC_PER_MSEC),
 	LAVD_CPU_ID_HERE		= ((u32)-2),
 	LAVD_CPU_ID_NONE		= ((u32)-1),
+	LAVD_CPU_ID_MAX			= 512,
 
 	LAVD_PREEMPT_KICK_LAT_PRIO	= 15,
 	LAVD_PREEMPT_KICK_MARGIN	= (LAVD_SLICE_MIN_NS >> 3),
 	LAVD_PREEMPT_TICK_MARGIN	= (LAVD_SLICE_MIN_NS >> 8),
 
+	LAVE_TC_PER_CORE_MAX_CTUIL	= 500, /* maximum per-core CPU utilization */
```
Byte-Lab, line 89: s/LAVE/LAVD?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -794,52 +816,193 @@ static void update_sys_cpu_load(void)
...
+	nr_active = calc_nr_active_cpus(stat_cur);
+	nr_cpus = nr_active + LAVD_TC_NR_OVRFLW;
+	bpf_for(i, 0, nr_cpus_onln) {
+		if (i >= LAVD_CPU_ID_MAX)
+			break;
+
+		/*
+		 * Skip offline cpu
+		 */
+		// cpu = get_cpu_id_of_order(i);
```
Byte-Lab, line 938: nit: Consider removing this stray comment

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -794,52 +816,193 @@ static void update_sys_cpu_load(void)
...
+		cpu = cpu_order[i];
+		cpuc = get_cpu_ctx_id(cpu);
+		if (!cpuc || !cpuc->is_online) {
+			bpf_cpumask_test_and_clear_cpu(cpu, active);
+			bpf_cpumask_test_and_clear_cpu(cpu, ovrflw);
```
Byte-Lab, line 943: Is there a reason you're using the RMW version of these? If not then I think you can just use `bpf_cpumask_clear_cpu()` and `bpf_cpumask_set_cpu()`.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1919,30 +2179,117 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
...
-	 * Now, a CPU has no task to run, so it gets a task from the global run
-	 * queue for execution.
+	 * If this CPU is not either in active or overflow CPUs, it tries to
+	 * find and run a task pinned to run on this CPU.
 	 */
-	scx_bpf_consume(LAVD_GLOBAL_DSQ);
+	__COMPAT_DSQ_FOR_EACH(p, LAVD_GLOBAL_DSQ, 0) {
+		/*
+		 * Prioritize kernel tasks because most kernel tasks are pinned
+		 * to a paritular CPU and latency-critical (e.g., ksoftirqd,
```
Byte-Lab, line 2220: s/paritular/particular

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1919,30 +2179,117 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
...
+		 * kworker, etc).
+		 */
+		if (is_kernel_task(p)) {
+			scx_bpf_consume(LAVD_GLOBAL_DSQ);
```
Byte-Lab, line 2224: Do we want to consume the task here? I think you'd have to do something like:

```
if (__COMPAT_scx_bpf_consume_task(BPF_FOR_EACH_ITER, p))
        break;
```

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1919,30 +2179,117 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
...
+		 * CPU. The scheduled task might be runnable on the active
+		 * cores. We will optimize this path after introducing per-core
+		 * DSQ.
+		 */
+		scx_bpf_consume(LAVD_GLOBAL_DSQ);
+
+		/*
+		 * This is the first time a particular pinned user-space task
+		 * is run on this CPU at this interval. From now on, this CPU
+		 * will be part of the activee CPU so can be used to run the
```
Byte-Lab, line 2261: s/activee/active

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1919,30 +2179,117 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
...
+
+		/*
+		 * If a task can run on active or overflow CPUs, it just does
+		 * nothing to go idle. 
+		 */
+		if (bpf_cpumask_intersects(cast_mask(active), p->cpus_ptr))
+			goto release_break;
+		if (bpf_cpumask_intersects(cast_mask(ovrflw), p->cpus_ptr))
+			goto release_break;
+
```
Byte-Lab, line 2245: I expect I'm misunderstanding the code, but don't we want to `continue` here? I thought the idea was to see if there's any task that _has_ to run on this CPU, and if so, don't we want to continue iterating until we find it?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1919,30 +2179,117 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
...
+		scx_bpf_consume(LAVD_GLOBAL_DSQ);
+
+		/*
+		 * This is the first time a particular pinned user-space task
+		 * is run on this CPU at this interval. From now on, this CPU
+		 * will be part of the activee CPU so can be used to run the
+		 * pinned task and the other tasks.
+		 */
+		bpf_cpumask_test_and_set_cpu(cpu, active);
+		scx_bpf_kick_cpu(cpu, __COMPAT_SCX_KICK_IDLE);
```
Byte-Lab, line 2265: Should we check the return value of `scx_bpf_consume()` to see if we were able to successfully consume a task before promoting this core to active?

Byte-Lab (commented):
This looks great Changwoo. I like that it's simple but works well. I left a few comments based on my first reading of the code.

scheds/rust/scx_lavd/src/bpf/intf.h
multics69: Fixed.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Fixed.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Fixed.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Fixed.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: active CPU mask is need to be atomically updated since it can be updated in two places concurrently -- do_core_compaction() and ops.dispatch(). In my checking the code, bpf_cpumask_clear_cpu() and bpf_cpumask_set_cpu() do not have LOCK prefix so they do not guarantee the atomicity. So I used the CAS version of set/clear_cpu(). Does it make sense?

For the overflow mask, I can change it to use non-atomic version since there is no concurrent writes.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: I wasn't sure if there is a case `cpu` is not the current CPU. If so, we don't need a kick. Will fix it.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: The idea is that if there is a pinned user-space process, we will promote the CPU to the active set anyway. So we will just consume any task and move on. Does it make sense?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Since that particular kernel task could be consumed by the other CPU, we just consume any task and move on. We could iterate over the loop to find out the exact kernel task but that seems overkill here, and this seems a good enough approximation. What do you think?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Fixed.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: I changed the update of overflow cpumask to non-atomic version.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
Byte-Lab: Ah yes, it does make sense. Thanks for explaining.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
Byte-Lab: >I wasn't sure if there is a case cpu is not the current CPU. If so, we don't need a kick. Will fix it.

Ah, sorry for the confusion but I just rechecked and apparently we can invoke ops.dispatch() for remote cores for core sched.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
Byte-Lab: I see. This seems like a reasonable approximation for now. If there's an issue we can always optimize.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -2321,69 +2677,150 @@ s32 BPF_STRUCT_OPS(lavd_init_task, struct task_struct *p,
...
+	if (err)
+		return err;
+
+	/*
+	 * Allocate cpumask for task compaction.
+	 *  - active CPUs: a group of CPUs will be used for now.
+	 *  - overflow CPUs: a pair of hyper-twin which will be used when there
+	 *    is no idle active CPUs.
+	 */
+	init_cpumasks();
```
Byte-Lab, line 2822: Should we check the return value here as well?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -2321,69 +2677,150 @@ s32 BPF_STRUCT_OPS(lavd_init_task, struct task_struct *p,
...
-	struct bpf_timer *timer;
-	u64 now;
-	u32 key = 0;
-	int cpu;
-	int err;
+	struct bpf_cpumask *cpumask;
+	cpumask = bpf_cpumask_create();
+	if (!cpumask)
+		return -ENOMEM;
+	bpf_cpumask_clear(cpumask);
```
Byte-Lab, line 2695: Just FYI: This isn't technically necessary. `bpf_cpumask_create()` will memset the cpumask to 0 before returning it.

Byte-Lab (approved):
Looks great, thanks for answering my questions. Great to see util start to be applied to tracking active cpu sets.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Fixed. Thanks!

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Oh, you are right. Fixed.

scheds/rust/scx_lavd/src/bpf/intf.h
```diff
@@ -150,6 +160,14 @@ struct cpu_ctx {
...
 	s32		cpu_id;		/* cpu id */
+
+	/*
+	 * Fields for core compaction
+	 *
+	 * NOTE: The followings MUST be placed at the end of this struct.
+	 */
+	struct bpf_cpumask __kptr *tmp_a_mask;	/* temporary cpu mask */
+	struct bpf_cpumask __kptr *tmp_o_mask;	/* temporary cpu mask */
 };
```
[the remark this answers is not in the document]
multics69: Definitely. I fixed it. Thanks!

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -794,52 +816,201 @@ static void update_sys_cpu_load(void)
...
+			/*
+			 * active_cpumask needs to be atomically updated since
+			 * there are concurrent writes -- one from
+			 * do_core_compaction() and another from
+			 * ops.dispatch(). On the other hand, ovrflw_cpumask
+			 * does not have such concurrent writes so that it can
+			 * be safely updated with a non-atomic (no LOCK prefix)
+			 * bitwise instruction.
+			 */
+			bpf_cpumask_test_and_clear_cpu(cpu, active);
```
htejun, line 950: `bpf_cpumask_clear_cpu()` uses `clear_bit()` which is guaranteed to be atomic and has the lock prefix on x86.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1919,30 +2187,118 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 	put_global_rq(p, taskc, cpuc, enq_flags);
 }
 
+static bool is_kernel_task(struct task_struct *p)
+{
+	return p->mm == NULL;
```
htejun, line 2192: Testing `PF_KTHREAD` is probably more idiomatic.

htejun (approved):

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: I double-check the kernel code. You are right. Thanks!

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Fixed. Thanks!

PR #340, opened 2024-06-07
scx_layered: Improve yield, preemption and other behaviors

htejun:
- An yielder is now charged the full slice by default.
- The amount of slice which is given up on each yield can now by modulated per-layer using the new `yield_ignore` parameter.
- Preemption is now first tried on the previous CPU the preempting task was on which should result in better locality behavior.
- Confined layers can now preempt.
- A new layer property `preempt_first` is added which prefers to preempt the CPU the task was previously on over idle CPUs.
- Long running tasks now can keep running on the same CPU within `max_exec_us`.
- More behavior improvements and statistics.


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -442,33 +461,146 @@ s32 BPF_STRUCT_OPS(layered_select_cpu, struct task_struct *p, s32 prev_cpu, u64
...
+
+	/*
+	 * If exclusive, we want to make sure the sibling CPU, if there's
+	 * one, is idle. However, if the sibling CPU is already running a
+	 * preempt task, we shouldn't kick it out.
+	 */
+	if (layer->exclusive && (sib = sibling_cpu(cand)) >= 0 &&
+	    (!(sib_cctx = lookup_cpu_ctx(sib)) || sib_cctx->current_preempt)) {
+		lstat_inc(LSTAT_EXCL_COLLISION, layer, cctx);
+		return false;
```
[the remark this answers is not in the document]
htejun: It's really just testing the sib's `current_preempt` and the second usage usages both `sib` and `sib_cctx` down the line. I can't think of a good way to factor that.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -384,56 +385,74 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
+	/*
+	 * If @p prefers to preempt @prev_cpu than finding an idle CPU and
+	 * @prev_cpu is preemptible, tell the enqueue path to try to preempt
+	 * @prev_cpu. The enqueue path will also retry to find an idle CPU if
+	 * the preemption attempt fails.
+	 */
+	if (from_selcpu && should_try_preempt_first(prev_cpu, layer, layered_cpumask)) {
+		cctx->try_preempt_first = true;
+		cpu = -1;
+		goto out_put;
```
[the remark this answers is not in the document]
htejun: Indeed. Yeah, it does allow it for some reason. Lemme fix that.

Byte-Lab (approved):
Looks great!

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -384,56 +385,72 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
 
-s32 BPF_STRUCT_OPS(layered_select_cpu, struct task_struct *p, s32 prev_cpu, u64 wake_flags)
+static __always_inline
+bool should_try_preempt_first(s32 cand, struct layer *layer,
+			      const struct cpumask *layered_cpumask)
+{
+	struct cpu_ctx *cand_cctx, *sib_cctx;
+	s32 sib;
+
+	if (!layer->preempt || !layer->preempt_first)
```
arighi, line 395: Maybe I'm missing something here, but shouldn't be `&&` instead of `||`?

htejun: Hmmm... so, it's saying "if the layer is not preempting or preempt_first is not enabled, tell the caller to not try to preempt first". ie. Both need to be enabled to try preempt_first, no?

Byte-Lab: @htejun maybe to simplify we could force `preempt_first` to be `false` if the layer isn't specified with `"preempt": true`? Or force `preempt: true` to be true if `preempt_first` is set.

arighi: > Hmmm... so, it's saying "if the layer is not preempting or preempt_first is not enabled, tell the caller to not try to preempt first". ie. Both need to be enabled to try preempt_first, no?

Oh I see, now it's clear. Not a blocker obviously, but I like @Byte-Lab suggestion,  I think it makes the code easier to understand for those who aren't familiar with the whole logic. Thanks!

PR #341, opened 2024-06-09
scx_rusty: Elimate data races possibility for domain min_vruntime

vax-r:
## Summary
READ_ONCE()/WRITE_ONCE() macros are added in commit 0932fde, we should be able to utilize the macros to get around the possibility of data races for domc->min_vruntime.

_Personally I think `dom_min_vruntime` function is redundant, it simply does the load operation of `domc->min_vruntime`, nothing special. It doesn't provide better readability either, the variable `domc->min_vruntime` is easy and trivial enough to understand itself._

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -377,21 +377,7 @@ static void dom_dcycle_xfer_task(struct task_struct *p, struct task_ctx *taskc,
 
 static u64 dom_min_vruntime(struct dom_ctx *domc)
 {
-	/*
-	 * Technically, this is undefined behavior according to the C standard,
-	 * section 5.1.2.4:
-	 *
-	 * The execution of a program contains a data race if it contains two
-	 * conflicting actions in different threads, at least one of which is
-	 * not atomic, and neither happens before the other. Any such data race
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -391,7 +391,7 @@ static u64 dom_min_vruntime(struct dom_ctx *domc)
 	 *
 	 * XXX: Once those macros are added, update this access.
 	 */
-	return domc->min_vruntime;
+	return READ_ONCE(domc->min_vruntime);
```
Byte-Lab, line 394: LGTM, thanks. Could you please remove the comment above as well? I agree that the function doesn't help readability, but it's nice to be able to prevent callers from having to all do `READ_ONCE()` everywhere

Byte-Lab (commented):
Thanks for adding this! Looks great, just left one request to remove a now unnecessary comment.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
vax-r: I see, I've removed the unused comment.

Byte-Lab (approved):

PR #345, opened 2024-06-11
scx_rustland: prevent starvation

arighi:
A set of changes that helps prevent potential starvation conditions.

This also seems to improve system responsiveness (~4-5% fps during the typical "gaming while building the kernel" benchmark) and it also helps to prevent potential audio cracking issues.

This patch has been produced thanks to the extensive iteration tests performed by SoulHarsh007 <harsh.peshwani@outlook.com> from the CachyOS community. Big kudos to him for all his valuable feedback.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -456,7 +456,7 @@ static void dispatch_user_scheduler(void)
 	 * Dispatch the scheduler on the first CPU available, likely the
 	 * current one.
 	 */
-	dispatch_task(p, SHARED_DSQ, 0, 0, 0);
+	dispatch_task(p, SHARED_DSQ, 0, 0, SCX_ENQ_PREEMPT);
 	bpf_task_release(p);
 }
 
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -367,7 +367,6 @@ dispatch_task(struct task_struct *p, u64 dsq_id,
 	s32 cpu;
 
 	switch (dsq_id) {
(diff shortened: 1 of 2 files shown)


rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -432,8 +431,8 @@ dispatch_task(struct task_struct *p, u64 dsq_id,
 		dbg_msg("dispatch: pid=%d (%s) dsq=%llu enq_flags=%llx slice=%llu",
 			p->pid, p->comm, dsq_id, enq_flags, slice);
 
-		/* Wake up the target CPU (only if idle) */
-		scx_bpf_kick_cpu(cpu, __COMPAT_SCX_KICK_IDLE);
+		/* Wake up the target CPU */
+		scx_bpf_kick_cpu(cpu, 0);
```
htejun, line 435: We should root cause why this makes a difference as this is indicating a bug either in SCX core or rustland.

htejun (approved):

rust/scx_rustland_core/assets/bpf/main.bpf.c
arighi: I'll investigate more, I also want to understand better what's going on, so let's wait a bit before merging this.

arighi:
I'm going to merge all these commits except the last one, since they provide clear benefits in terms of responsiveness and they help to prevent audio cracking.

About the stall issue reported by SoulHarsh007, I'd like to investigate more and better understand the nature of the problem, so I'm going to drop the last commit (the one that always kicks the CPUs without using `__COMPAT_SCX_KICK_IDLE`). Even if it seems to prevent the stall, I'm worried that we would just hide the real problem, hence I'm excluding it for now.

This issue is still tracked in #353, so we can merge and close this PR and I'll create a new PR once we have a proper fix.

PR #358, opened 2024-06-14
README: Add information about restoring default values

sirlucjan:
It is worth adding information on how to restore the default values.


arighi (commented):
I think you can just unset the variable name, you don't have to specify the content. Apart than that LGTM, thanks!

sirlucjan:
> I think you can just unset the variable name, you don't have to specify the content. Apart than that LGTM, thanks!

Indeed!

arighi (approved):

PR #363, opened 2024-06-16
Strip compat support

htejun:
In preparation of upstreaming, set the currently released v6.9 kernels as the minimum requirement and strip the compat features. The only remaining one is for dsq iters. This needs updates from BPF side and will be handled later.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf.rs
+++ b/rust/scx_rustland_core/assets/bpf.rs
@@ -246,7 +246,6 @@ impl<'cb> BpfScheduler<'cb> {
 
         skel.bss_mut().usersched_pid = std::process::id();
         skel.rodata_mut().slice_ns = slice_us * 1000;
-        skel.rodata_mut().switch_partial = partial;
         skel.rodata_mut().debug = debug;
         skel.rodata_mut().full_user = full_user;
         skel.rodata_mut().low_power = low_power;
--- a/scheds/rust/scx_rusty/src/main.rs
+++ b/scheds/rust/scx_rusty/src/main.rs
@@ -314,7 +314,6 @@ impl<'a> Scheduler<'a> {
         skel.rodata_mut().load_half_life = (opts.load_half_life * 1000000000.0) as u32;
         skel.rodata_mut().kthreads_local = opts.kthreads_local;
         skel.rodata_mut().fifo_sched = opts.fifo_sched;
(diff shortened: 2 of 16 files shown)


rust/scx_rustland_core/assets/bpf.rs
```diff
@@ -246,7 +246,6 @@ impl<'cb> BpfScheduler<'cb> {
 
         skel.bss_mut().usersched_pid = std::process::id();
         skel.rodata_mut().slice_ns = slice_us * 1000;
-        skel.rodata_mut().switch_partial = partial;
```
arighi, line 249: We should also drop `partial` from `BpfScheduler`, update scx_rustland and scx_rlfifo, and drop the option `--partial` from scx_rustland), but it can be done later in a separate commit (this is just as a reminder).

scheds/rust/scx_rusty/src/main.rs
```diff
@@ -314,7 +314,6 @@ impl<'a> Scheduler<'a> {
         skel.rodata_mut().load_half_life = (opts.load_half_life * 1000000000.0) as u32;
         skel.rodata_mut().kthreads_local = opts.kthreads_local;
         skel.rodata_mut().fifo_sched = opts.fifo_sched;
-        skel.rodata_mut().switch_partial = opts.partial;
```
arighi, line 317: Ditto for scx_rusty about `--partial`.

arighi (approved):
LGTM, left few minor comments that can be also addressed later.

Moreover, we should also update the `sched_ext-ci` branch, or CI will fail.

rust/scx_rustland_core/assets/bpf.rs
htejun: `--partial` still sets `SCX_OPS_SWITCH_PARTIAL` in `sched_ext_ops.flags`, so it should keep working the same.

scheds/rust/scx_rusty/src/main.rs
htejun: Ditto.

multics69 (approved):
multics69@gmail.com

scheds/rust/scx_rusty/src/main.rs
arighi: Ah, I was missing this part, that's right. So everything looks good, thanks!

PR #369, opened 2024-06-18
scx_lavd: properly check for idle CPUs in pick_cpu()

arighi:
It seems that we are not updating `is_idle` when we find an idle CPU with pick_cpu(), causing unnecessary rescheduling events when select_cpu() is called.

To resolve this, ensure that the is_idle state is correctly set. Additionally, always ensure that the task is dispatched to the local DSQ immediately upon finding (and reserving) an idle CPU.


multics69 (approved):
Huge thanks @arighi ! The bug was creeped in while implementing core compaction.

PR #371, opened 2024-06-18
Add build-id to build process

Byte-Lab:
We want schedulers to be able to print, log, etc the build ID of the
repository.
    
These patches update scx_utils accordingly to use vergen to generate the output of
`git describe`, and then updates scx_rusty to print that build ID when it's loaded.
In the future, we could further update scx_utils to include all of the various features
available in [0]. A subsequent patch will update scx_rusty to print this build ID value.

[0]: https://docs.rs/vergen/8.3.1/vergen/index.html

### DIFF
--- a/rust/scx_utils/build.rs
+++ b/rust/scx_utils/build.rs
@@ -3,8 +3,14 @@
 // This software may be used and distributed according to the terms of the
 // GNU General Public License version 2.
 
+use vergen::EmitBuilder;
 include!("src/builder.rs");
 
 fn main() {
-    Builder::new().build()
+    Builder::new().build();
+    EmitBuilder::builder()
+        .all_git()
+        .cargo_target_triple()
+        .emit()
+        .unwrap();
 }
--- a/rust/scx_utils/Cargo.toml
+++ b/rust/scx_utils/Cargo.toml
@@ -31,4 +31,5 @@ metrics-util = "0.17.0"
 [build-dependencies]
 bindgen = ">=0.68, <0.70"
 tar = "0.4"
+vergen = { version = "8.0.0", features = ["cargo", "git", "gitcl"] }
 walkdir = "2.4"
--- a/rust/scx_utils/src/build_id.rs
+++ b/rust/scx_utils/src/build_id.rs
@@ -1,0 +1,61 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+//
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use std::fmt::Write;
+
+lazy_static::lazy_static! {
+    static ref GIT_VERSION: String = {
+        let mut ver = String::new();
+        match option_env!("VERGEN_GIT_SHA") {
+            Some(v) if v != "VERGEN_IDEMPOTENT_OUTPUT" => {
+                ver += "g";
+                ver += v;
+                if let Some("true") = option_env!("VERGEN_GIT_DIRTY") {
+                    ver += "-dirty";
+                }
+            }
+            _ => {}
+        }
+        ver
+    };
+    static ref BUILD_TAG: String = {
+        let mut tag = env!("VERGEN_CARGO_TARGET_TRIPLE").to_string();
+        if cfg!(debug_assertions) {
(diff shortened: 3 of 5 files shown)


meson-scripts/gen_build.in.h
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/*
+ * NOTE: THIS FILE IS AUTO-GENERATED BY THE BUILD SYSTEM AND IS NOT TRACKED BY
+ * SOURCE CONTROL. DO NOT EDIT THIS FILE -- IT WILL BE OVERWRITTEN BY
+ * SUBSEQUENT BUILDS.
+ */
```
Byte-Lab, line 6: Reviewers: This is kind of confusing because _this_ file (`gen_build.in.h`) of course is actually tracked by source control, but the `build.h` file it generates by replacing `@VCS_TAG` is not tracked. We could get around this by echoing this comment to the `build.h` file from `gen_bindings_h`, but it seemed better to have less logic in bash given that this file is basically never going to change. I'm fine with either way.

htejun:
If we limit the scope to rust schedulers, would this become simpler with https://crates.io/crates/vergen?

Byte-Lab:
> If we limit the scope to rust schedulers, would this become simpler with https://crates.io/crates/vergen?

Hmm, so I guess the idea is that we create this [`EmitBuilder()`](https://docs.rs/vergen/8.3.1/vergen/struct.EmitBuilder.html#) and configure it to emit the `git` features, and then we can get the git sha from other files in the crate using environment variables like `env!("VERGEN_GIT_DESCRIBE")`? Certainly an option if we don't want to support C for the build IDs (assuming I'm correctly understanding how this works), and definitely a bit less code and weird build complexity.

OTOH, it's not _too_ much logic, and we do still have C schedulers so there is arguably some value in supporting them (though admittedly not much given that they're basically unused at this point).

htejun:
Oh, and as for C builds. I'm not sure. It'd be nice to have but I don't think it matters practically and it's kinda nice to keep them simple for syncing with the kernel tree too.

Byte-Lab:
@htejun applied your suggestion. This is way simpler, thanks for suggesting it.

rust/scx_utils/build.rs
```diff
@@ -3,8 +3,13 @@
...
 // GNU General Public License version 2.
 
+use vergen::EmitBuilder;
 include!("src/builder.rs");
 
 fn main() {
-    Builder::new().build()
+    Builder::new().build();
+    EmitBuilder::builder()
+        .all_git()
```
arighi, line 12: One issue that I see here is when we build this from a source package we won't have `.git` and apparently in this case vergen would emit something like `VERGEN_IDEMPOTENT_OUTPUT`. It'd be nice if we could fallback to the version define in Cargo.toml in this case...

arighi (approved):
Left a comment about packaging, but overall it looks good, we can fix the packaging issue later.

rust/scx_utils/build.rs
htejun: `.cargo_target_triple()` can be useful too.

rust/scx_utils/src/compat.rs
```diff
@@ -17,6 +17,8 @@ lazy_static::lazy_static! {
     read_enum("scx_ops_flags", "SCX_OPS_SWITCH_PARTIAL").unwrap_or(0);
 }
 
+pub const SCX_BUILD_ID: &'static str = env!("VERGEN_GIT_DESCRIBE");
```
htejun, line 20: Note that `VERGEN_GIT_DESCRIBE` can be confusing as it will show the scx repo version, not the scheduler version in `Cargo.toml`. They're not the same numbers. `resctl-demo` needs almost exactly the same thing and maybe we can just copy what it's doing? https://github.com/facebookexperimental/resctl-demo/blob/main/rd-util/src/lib.rs#L106

Byte-Lab:
@htejun / @arighi does this look more reasonable?

htejun (approved):

arighi (approved):
Tested both recompiling it from the git source dir and inside a package (without git). All good.

PR #372, opened 2024-06-19
scx_utils: Utilize Entry API for BTreeMap insertion

vax-r:
## Summary
Take advantages of BTreeMap's Entry API working with or_insert() to do the conditional insertion. Insert only when the entry doesn't exist. Doing so can reduce the amount of code and provide better readability and perform in-place manipulation.
## Reference
* [BTreeMap example](https://doc.rust-lang.org/nightly/std/collections/struct.BTreeMap.html#examples)
* [BTreeMap Entry API](https://doc.rust-lang.org/nightly/std/collections/struct.BTreeMap.html#method.entry)

### DIFF
--- a/rust/scx_utils/src/infeasible.rs
+++ b/rust/scx_utils/src/infeasible.rs
@@ -313,25 +313,18 @@ impl LoadAggregator {
             bail!("weight {} is less than minimum weight {}", weight, MIN_WEIGHT);
         }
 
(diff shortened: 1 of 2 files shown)


Byte-Lab (approved):
TIL about `insert_or`. Thanks a lot for the patch, this is definitely much more readable and clean.

PR #373, opened 2024-06-19
scx_lavd: Reuse can_task1_kick_task2

vax-r:
## Summary
Use the function `can_task1_kick_task2()` to replace places which also checking the `comp_preemption_info` between two cpus for better consistency.


multics69 (approved):
Looks good. Thanks!

PR #374, opened 2024-06-19
scx_rustland_core: include buddy-alloc and refactor allocator code

arighi:
The dependency of the buddy-alloc crate [1] seems to cause some troubles with packaging, mostly because the selftests for the crate are failing when it's compiled in release mode.

For example:

 $ cargo test --release -- --nocapture
 thread 'tests::fast_alloc::test_basic_malloc' panicked at src/tests/fast_alloc.rs:25:13:
 assertion `left == right` failed
   left: 0
  right: 42

Some of these failures with BuddyAlloc can be fixed by using a memory arena buffer aligned to page size.

However, some test failures with FastAlloc persist that cannot be resolved merely by aligning the pre-allocated memory arena to the page size, as mentioned in [2].

The concern is that this may potentially lead to actual memory bugs.

Therefore, it seems safer to refactor the custom allocator code to simply use BuddyAlloc, dropping FastAlloc completely.

To achieve this, the entire BuddyAlloc code has been directly included in scx_rustland_core, referencing the original project and its MIT licensing information (with the entire code still distributed under the GPLv2 license).

Then the code has been slightly modified to remove FastAlloc and the external dependency on the buddy-alloc crate has been dropped.

From a performance perspective this change doesn't seem to introduce any measurable regression.

[1] https://github.com/jjyr/buddy-alloc
[2] https://github.com/jjyr/buddy-alloc/issues/16


htejun (approved):

jordalgo (approved):
My knowledge of Rust stinks but LGTM :)

PR #377, opened 2024-06-20
rusty: Integrate stats with the metrics framework

jfernandez:
We need a layer of indirection between the stats collection and their output destinations. Currently, stats are only printed to stdout. Our goal is to integrate with various telemetry systems such as Prometheus, StatsD, and custom metric backends like those used by Meta and Netflix. Importantly, adding a new backend should not require changes to the existing stats code.

This patch introduces the `metrics` [1] crate, which provides a framework for defining metrics and publishing them to different backends.

The initial implementation includes the `dispatched_tasks_count` metric, tagged with `type`. This metric increments every time a task is dispatched, emitting the raw count instead of a percentage. A monotonic counter is the most suitable metric type for this use case, as percentages can be calculated at query time if needed. Existing logged metrics continue to print percentages and remain unchanged.

A new flag, `--enable-prometheus`, has been added. When enabled, it starts a Prometheus endpoint on port 9000 (default is false). This endpoint allows metrics to be charted in Prometheus or Grafana dashboards.

Example of charting 1s rate of dispatched tasks by type:
image

Future changes will migrate additional stats to this framework and add support for other backends.

[1] https://metrics.rs/

### DIFF
--- a/scheds/rust/scx_rusty/src/main.rs
+++ b/scheds/rust/scx_rusty/src/main.rs
@@ -35,6 +35,9 @@ use libbpf_rs::skel::OpenSkel;
 use libbpf_rs::skel::Skel;
 use libbpf_rs::skel::SkelBuilder;
 use log::info;
+use metrics::counter;
+use metrics::Counter;
+use metrics_exporter_prometheus::PrometheusBuilder;
 use scx_utils::compat;
 use scx_utils::init_libbpf_logging;
 use scx_utils::scx_ops_attach;
@@ -189,6 +192,10 @@ struct Opts {
     /// times to increase verbosity.
     #[clap(short = 'v', long, action = clap::ArgAction::Count)]
     verbose: u8,
+
+    /// Enable the Prometheus endpoint for metrics on port 9000.
+    #[clap(long, action = clap::ArgAction::SetTrue)]
+    enable_prometheus: bool,
 }
 
 fn read_total_cpu(reader: &procfs::ProcReader) -> Result<procfs::CpuStat> {
@@ -204,6 +211,46 @@ pub fn sub_or_zero(curr: &u64, prev: &u64) -> u64 {
         res
     } else {
         0
+    }
+}
+
+struct Metrics {
+    wsync: Counter,
+    wsync_prev_idle: Counter,
+    prev_idle: Counter,
+    greedy_idle: Counter,
+    pinned: Counter,
+    direct_dispatch: Counter,
+    direct_greedy: Counter,
+    direct_greedy_far: Counter,
+    dsq: Counter,
+    greedy_local: Counter,
+    greedy_xnuma: Counter,
+    kick_greedy: Counter,
+    repatriate: Counter,
+    dl_clamped: Counter,
+    dl_preset: Counter,
+}
+
+impl Metrics {
+    fn new() -> Self {
+        Self {
+            wsync: counter!("dispatched_tasks_total", "type" => "wsync"),
+            wsync_prev_idle: counter!("dispatched_tasks_total", "type" => "wsync_prev_idle"),
+            prev_idle: counter!("dispatched_tasks_total", "type" => "prev_idle"),
+            greedy_idle: counter!("dispatched_tasks_total", "type" => "greedy_idle"),
+            pinned: counter!("dispatched_tasks_total", "type" => "pinned"),
+            direct_dispatch: counter!("dispatched_tasks_total", "type" => "direct_dispatch"),
+            direct_greedy: counter!("dispatched_tasks_total", "type" => "direct_greedy"),
+            direct_greedy_far: counter!("dispatched_tasks_total", "type" => "direct_greedy_far"),
+            dsq: counter!("dispatched_tasks_total", "type" => "dsq"),
+            greedy_local: counter!("dispatched_tasks_total", "type" => "greedy_local"),
+            greedy_xnuma: counter!("dispatched_tasks_total", "type" => "greedy_xnuma"),
+            kick_greedy: counter!("kick_greedy_total"),
+            repatriate: counter!("repatriate_total"),
+            dl_clamped: counter!("dl_clamped_total"),
+            dl_preset: counter!("dl_preset_total"),
+        }
     }
 }
 
@@ -228,6 +275,8 @@ struct Scheduler<'a> {
     nr_lb_data_errors: u64,
 
     tuner: Tuner,
+
+    metrics: Metrics,
 }
 
 impl<'a> Scheduler<'a> {
@@ -353,6 +402,8 @@ impl<'a> Scheduler<'a> {
                 opts.slice_us_underutil * 1000,
                 opts.slice_us_overutil * 1000,
             )?,
+
+            metrics: Metrics::new(),
         })
     }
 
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rusty/src/main.rs
```diff
@@ -446,17 +452,41 @@ impl<'a> Scheduler<'a> {
...
+        counter!("dispatched_tasks_count", "type" => "wsync").increment(wsync);
+        counter!("dispatched_tasks_count", "type" => "prev_idle").increment(prev_idle);
+        counter!("dispatched_tasks_count", "type" => "greedy_idle").increment(greedy_idle);
+        counter!("dispatched_tasks_count", "type" => "pinned").increment(pinned);
+        counter!("dispatched_tasks_count", "type" => "direct_dispatch").increment(direct_dispatch);
+        counter!("dispatched_tasks_count", "type" => "direct_greedy").increment(direct_greedy);
+        counter!("dispatched_tasks_count", "type" => "direct_greedy_far").increment(direct_greedy_far);
+        counter!("dispatched_tasks_count", "type" => "dsq").increment(dsq);
+        counter!("dispatched_tasks_count", "type" => "greedy_local").increment(greedy_local);
+        counter!("dispatched_tasks_count", "type" => "greedy_xnuma").increment(greedy_xnuma);
```
dschatzberg, line 489: This registers a new counter on each `report()`. I would think we want to register the counters ahead of time and only set their absolute value here, no?

scheds/rust/scx_rusty/src/main.rs
```diff
@@ -470,45 +500,50 @@ impl<'a> Scheduler<'a> {
...
-            stat_pct(bpf_intf::stat_idx_RUSTY_STAT_DSQ_DISPATCH),
-            stat_pct(bpf_intf::stat_idx_RUSTY_STAT_GREEDY_LOCAL),
-            stat_pct(bpf_intf::stat_idx_RUSTY_STAT_GREEDY_XNUMA),
+            stat_pct(dsq),
+            stat_pct(greedy_local),
+            stat_pct(greedy_xnuma),
         );
 
+        let kick_greedy = stat(bpf_intf::stat_idx_RUSTY_STAT_KICK_GREEDY);
+        let repatriate = stat(bpf_intf::stat_idx_RUSTY_STAT_REPATRIATE);
```
dschatzberg, line 534: You're not exposing these to the metrics backend?

scheds/rust/scx_rusty/src/main.rs
```diff
@@ -470,45 +500,50 @@ impl<'a> Scheduler<'a> {
...
+        let dl_preset = stat(bpf_intf::stat_idx_RUSTY_STAT_DL_PRESET);
         info!(
             "dl_clamped={:5.2} dl_preset={:5.2}",
-            stat_pct(bpf_intf::stat_idx_RUSTY_STAT_DL_CLAMP),
-            stat_pct(bpf_intf::stat_idx_RUSTY_STAT_DL_PRESET),
+            stat_pct(dl_clamped),
+            stat_pct(dl_preset),
         );
 
         info!("slice_length={}us", self.tuner.slice_ns / 1000);
```
dschatzberg, line 549: I wonder if it's possible to make all this logging of stats into a metrics backend itself. That would reduce some of the boilerplate needed to add more stats.

dschatzberg (commented):
This looks very similar to what I added to scx_layered. I'm mostly happy with the approach, but there's still a fair bit of boilerplate. One idea I kicked around was making a procedural macro to define stats from a Struct which would generate a lot of this boilerplate.

scheds/rust/scx_rusty/src/main.rs
jfernandez: I can add them in this PR as well

scheds/rust/scx_rusty/src/main.rs
jfernandez: Yes, absolutely. My plan for the next PR was to make a terminal log exporter and have that own this logic.

scheds/rust/scx_rusty/src/main.rs
jfernandez: Let me look into that and see if we'd benefit from instantiating them once and reusing them. I was under the impression that the library authors want this pattern and have optimized the implementation to avoid the perceived overhead of registering new counters on each call.

htejun:
This looks better than OM and no objection from me. Some things to consider for the future:
- Compact and declarative stat definition would be lovely.
- Down the line, if all rust scheds follow the same recipe, it'd be great.
- It'd be nicer if we can just pack stats into a struct and hand it to a stat backend which can then present it however it wants to present (one of those could be stdout formatter that the scheduler implementation provides). It's so much easier to deal with such structs for e.g. calculating deltas, fractions, serializing into something else and so on.

scheds/rust/scx_rusty/src/main.rs
jfernandez: @dschatzberg I created a `Metrics` struct to hold the metrics and made it a member of the Scheduler struct. We now only register the metrics once.

scheds/rust/scx_rusty/src/main.rs
jfernandez: Added

jfernandez:
@Byte-Lab @htejun agreed and aligned on the long-term vision. My plan is to use `scx_rusty` to learn the common observability patterns and iterate to generalize this solution for all schedulers. I'll start working next on removing boilerplate.

@dschatzberg I applied your feedback and I also created a Struct to hold the metrics. If you are not opposed, I'd like to first create a terminal logging backend so that we can clean up the `report` fn, and then circle back to removing boilerplate as I mentioned above.

dschatzberg:
@jfernandez Yeah, that's totally fine by me. The idea I had regarding a macro was to make it more like the `clap` crate works with CLI options, where decorators on each field in the Metrics struct would allow a macro to generate the `new()` impl and even the fetch + increment bits.

BTW, do you need .absolute() and not .increment() now that the counters are pre-registered?

jfernandez:
> BTW, do you need .absolute() and not .increment() now that the counters are pre-registered?

@dschatzberg no, `.increment()` is still the correct function to call. Internally, counters should be a monotonic incrementing value. A `gauge!` is what we'd use if we want to instrument an absolute value. In V1 of this PR, calling `counter!` was correctly only registering the metric once and subsequent calls would return the cached metric. But initializing once and storing in the struct is a bit more efficient and sets us up for removing boiler plate down the line.

How these monotonic counters are exported is up to the target backend. For Prometheus, since it's a poll-based system, the counter increases the value in memory until there is a scrape event, then it gets reset back to 0. Then the Prometheus backend simply adds to the value and you need to chart it with `rate(dispatched_tasks_count[1m])` to see the rate per second.

For something like Netflix's `spectatord`, state is handled by the daemon, and we'd emit the metric on every `increment!` call without needing to store any state.

jfernandez:
One more change, I got the convention wrong. It's `_total`, not count. Updated.

Byte-Lab:
> > The idea I had regarding a macro was to make it more like the clap crate works with CLI options, where decorators on each field in the Metrics struct would allow a macro to generate the new() impl and even the fetch + increment bits.
> 
> Ah yeah, I'm familiar with this pattern and I like it as well. I will explore this when I focus on removing boiler plate.
> 
> > BTW, do you need .absolute() and not .increment() now that the counters are pre-registered?
> 
> @dschatzberg no, `.increment()` is still the correct function to call. Internally, counters should be a monotonic incrementing value. A `gauge!` is what we'd use if we want to instrument an absolute value. In V1 of this PR, calling `counter!` was correctly only registering the metric once and subsequent calls would return the cached metric. But initializing once and storing in the struct is a bit more efficient and sets us up for removing boiler plate down the line.

@jfernandez I'm a bit confused here. These `counter!`s are monotonic, but they're incremented and never reset from the BPF side. So if we incremented the values then we'd be incrementing using absolute values. Can we not use the [`absolute()`](https://docs.rs/metrics/latest/metrics/struct.Counter.html#method.absolute) function? In my mind a gauge is meant for something that's non-monotonic.

> How these monotonic counters are exported is up to the target backend. For Prometheus, since it's a poll-based system, the counter increases the value in memory until there is a scrape event, then it gets reset back to 0. Then the Prometheus backend simply adds to the value and you need to chart it with `rate(dispatched_tasks_count[1m])` to see the rate per second.

Hmm, I'm still not quite following how we won't need state to track this (meaning, we record what the value was last time, and then increment based on that). Even if we emit the metric on every call to `increment!()`, we're still emitting an absolute value.

> For something like Netflix's `spectatord`, state is handled by the daemon, and we'd emit the metric on every `increment!` call without needing to store any state.

jfernandez:
@Byte-Lab ah! I assumed that the bpf stats were being reset for each loop. If they are always incrementing, then I need to rework this. Let me get back to you on this.

Byte-Lab:
> @Byte-Lab ah! I assumed that the bpf stats were being reset for each loop. If they are always incrementing, then I need to rework this. Let me get back to you on this.

Sorry @jfernandez, as we discussed on slack you weretotally correct. The values are reset on each read here: https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_rusty/src/main.rs#L433-L435. What you have now LG.

Byte-Lab (approved):

PR #379, opened 2024-06-21
Add topo.nr_cpu_ids() to Topology crate

Byte-Lab:
In some cases, a host may have an odd topology where there are gaps in
CPU IDs (including between possible CPUs). A common pattern in
schedulers is to perform allocations for every possible CPU ID, such as
creating a per-cpu DSQ. In order to avoid confusing schedulers, let's
track the maximum CPU ID on a system so that we can return the number of
CPU IDs on the system which is inclusive of gaps.

This closes the loop on the discussion in https://github.com/sched-ext/scx/pull/281#pullrequestreview-2051347033.


htejun (approved):

arighi (approved):
LGTM, thanks for doing this!

PR #382, opened 2024-06-23
scx_rusty: Refactor ridx assignment in populate_tasks_by_load

vax-r:
## Summary
Origin assignment of the variable `ridx` is equivalent to comparing between "ridx" and "wids - MAX_PIDS". It's confusing using `if` statement and compare "wids - ridx" with "MAX_PIDS".

Using u64 max library helper function to perform the comparison and provide better readability.

### DIFF
--- a/scheds/rust/scx_rusty/src/load_balance.rs
+++ b/scheds/rust/scx_rusty/src/load_balance.rs
@@ -686,9 +686,7 @@ impl<'a, 'b> LoadBalancer<'a, 'b> {
         let mut pids = vec![];
 
         let (mut ridx, widx) = (active_pids.read_idx, active_pids.write_idx);
-        if widx - ridx > MAX_PIDS {
-            ridx = widx - MAX_PIDS;
-        }
+        ridx = ridx.max(widx - MAX_PIDS);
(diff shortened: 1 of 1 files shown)


Byte-Lab (approved):
Yeah that was super confusing, thanks for the cleanup

Byte-Lab:
Unfortunately we have to revert this -- the reason we had it this way is to avoid underflow. Otherwise we can get:

```
[void@maniforge scx]$ rusty
21:32:36 [INFO] Running scx_rusty (build ID: 0.8.1-g41d60aef04d2a7fc0da738bf50391ea5f232ddd3 x86_64-unknown-linux-gnu/debug)
21:32:36 [INFO] ---
21:32:36 [INFO] NUMA[00] mask= 0b11111111111111111111111111111111
21:32:36 [INFO]   DOM[00] mask= 0b00000000111111110000000011111111
21:32:36 [INFO]   DOM[01] mask= 0b11111111000000001111111100000000
21:32:36 [INFO] Rusty scheduler started!
thread 'main' panicked at src/load_balance.rs:689:25:
```

PR #384, opened 2024-06-23
scx_utils: Add log_recorder module for metrics-rs

jfernandez:
This change adds a new module to the scx_utils crate that provides a log recorder for metrics-rs. The log recorder will log all metrics to the console at a configurable interval in an easy to read format. Each metric type will be displayed in a separate section. Indentation will be used to show the hierarchy of the metrics. This results in a more verbose output, but it is easier to read and understand.

scx_rusty was updated to use the log recorder and all explicit metric logging was removed.

Counters will show the total count and the rate of change per second. Counters with an additional label, like `type` in
`dispatched_tasks_total` in rusty, will show the count, rate, and percentage of the total count.

```
Counters:
  dispatched_tasks_total: 65559 [1344.8/s]
    prev_idle: 44963 (68.6%) [966.5/s]
    wsync_prev_idle: 15696 (23.9%) [317.3/s]
    direct_dispatch: 2833 (4.3%) [35.3/s]
    dsq: 1804 (2.8%) [21.3/s]
    wsync: 262 (0.4%) [4.3/s]
    direct_greedy: 1 (0.0%) [0.0/s]
    pinned: 0 (0.0%) [0.0/s]
    greedy_idle: 0 (0.0%) [0.0/s]
    greedy_xnuma: 0 (0.0%) [0.0/s]
    direct_greedy_far: 0 (0.0%) [0.0/s]
    greedy_local: 0 (0.0%) [0.0/s]
  dl_clamped_total: 1290 [20.3/s]
  dl_preset_total: 514 [1.0/s]
  kick_greedy_total: 6 [0.3/s]
  lb_data_errors_total: 0 [0.0/s]
  load_balance_total: 0 [0.0/s]
  repatriate_total: 0 [0.0/s]
  task_errors_total: 0 [0.0/s]
```

Gauges will show the last set value:

```
Gauges:
  slice_length_us: 20000.00
```

Histograms will show the average, min, and max. The histogram will be reset after each log interval to avoid memory leaks, since the data structure that holds the samples is unbounded.

```
Histograms:
  cpu_busy_pct: avg=1.66 min=1.16 max=2.16
  load_avg node=0: avg=0.31 min=0.23 max=0.39
  load_avg node=0 dom=0: avg=0.31 min=0.23 max=0.39
  processing_duration_us: avg=297.50 min=296.00 max=299.00
```
[Screencast from 2024-06-23 16-52-37.webm](https://github.com/sched-ext/scx/assets/3172/6282a5a5-808e-4656-94f7-a21b28288b8e)

### DIFF
--- a/rust/scx_utils/src/log_recorder.rs
+++ b/rust/scx_utils/src/log_recorder.rs
@@ -1,0 +1,284 @@
+// Copyright (c) Netflix, Inc.
+// Author: Jose Fernandez
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use std::collections::HashMap;
+use std::sync::atomic::Ordering::Relaxed;
+use std::sync::Arc;
+use std::thread;
+use std::time::Duration;
+use std::time::Instant;
+
+use anyhow::Result;
+use log::info;
+use metrics::Counter;
+use metrics::Gauge;
+use metrics::Histogram;
+use metrics::Key;
+use metrics::KeyName;
+use metrics::Metadata;
+use metrics::Recorder;
+use metrics::SharedString;
+use metrics::Unit;
+use metrics_util::registry::AtomicStorage;
+use metrics_util::registry::Registry;
+
+/// A builder for creating a new instance of `LogRecorder` and installing it as
+/// the global recorder.
+///
+/// Example:
+///
+/// ```rust
+/// LogRecorderBuilder::new()
+///     .with_reporting_interval(Duration::from_secs(3))
+///     .install()?;
+/// ```
+pub struct LogRecorderBuilder {
+    reporting_interval: Duration,
+}
+
+impl LogRecorderBuilder {
+    pub fn new() -> LogRecorderBuilder {
+        Self {
+            reporting_interval: Duration::from_secs(3),
+        }
+    }
+
+    /// Sets the interval at which the recorder will log the metrics.
+    pub fn with_reporting_interval(mut self, interval: Duration) -> Self {
+        self.reporting_interval = interval;
+        self
+    }
+
+    /// Installs the log recorder as the global recorder.
+    pub fn install(self) -> Result<()> {
+        let recorder = LogRecorder {
+            registry: Arc::new(Registry::<Key, AtomicStorage>::atomic()),
+        };
+        recorder.start(self.reporting_interval);
+        metrics::set_global_recorder(recorder)?;
+        Ok(())
+    }
+}
+
+/// A metrics recorder that logs metrics to the terminal.
+///
+/// `LogRecorder` implements the `Recorder` trait from the metrics-rs framework.
+/// It maintains an in-memory registry of metrics and uses a background thread
+/// to report all metrics at regular intervals.
+///
+/// Use the `LogRecorderBuilder` to create a new instance of `LogRecorder` and
+/// install it as the global recorder.
+struct LogRecorder {
+    registry: Arc<Registry<Key, AtomicStorage>>,
+}
+
+impl Recorder for LogRecorder {
+    fn describe_counter(&self, _: KeyName, _: Option<Unit>, _: SharedString) {
+        unimplemented!()
+    }
+
+    fn describe_gauge(&self, _: KeyName, _: Option<Unit>, _: SharedString) {
+        unimplemented!()
+    }
+
+    fn describe_histogram(&self, _: KeyName, _: Option<Unit>, _: SharedString) {
+        unimplemented!()
+    }
+
+    fn register_counter(&self, key: &Key, _: &Metadata<'_>) -> Counter {
+        self.registry
+            .get_or_create_counter(key, |c| c.clone().into())
+    }
+
+    fn register_gauge(&self, key: &Key, _: &Metadata<'_>) -> Gauge {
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_rusty/src/main.rs
```diff
@@ -623,12 +583,11 @@ impl<'a> Scheduler<'a> {
         );
 
         lb.load_balance()?;
+        self.metrics.processing_duration.record(started_at.elapsed().as_micros() as f64);
```
jfernandez, line 586: I moved capturing this metric at close as possible to the function we are instrumenting. This way we don't need to pass the duration around. Likewise for `cpu_busy` a few lines above.

rust/scx_utils/src/log_recorder.rs
```diff
@@ -0,0 +1,275 @@
...
+
+                        if min == 0.0 || *element < min {
+                            min = element.clone();
+                        }
+
+                        if *element >= max {
+                            max = element.clone();
+                        }
+                    }
+                });
```
tobz, line 261: Not sure if you saw it, but there's `AtomicBucket::clear_with`, which is generally preferred to `data`/`data_with` + `clear` when you're trying to atomically clear the bucket and observe all the values you cleared.

jfernandez: I wasn't aware of that fn, thanks. Updated to use it.

scheds/rust/scx_rusty/src/main.rs
```diff
@@ -543,68 +549,21 @@ impl<'a> Scheduler<'a> {
...
-        info!("slice_length={}us", self.tuner.slice_ns / 1000);
-        info!("direct_greedy_cpumask={}", self.tuner.direct_greedy_mask);
-        info!("  kick_greedy_cpumask={}", self.tuner.kick_greedy_mask);
+        self.metrics.task_errors.increment(stat(bpf_intf::stat_idx_RUSTY_STAT_TASK_GET_ERR));
+        self.metrics.lb_data_errors.increment(self.nr_lb_data_errors);
+        self.metrics.load_balance.increment(stat(bpf_intf::stat_idx_RUSTY_STAT_LOAD_BALANCE));
+        
+        self.metrics.slice_length.set(self.tuner.slice_ns as f64 / 1000.0);
 
+        // We need to dynamically create the  metrics for each node and domain
```
Byte-Lab, line 558: extra space between 'the' and 'metrics'

scheds/rust/scx_rusty/src/main.rs
```diff
@@ -543,68 +549,21 @@ impl<'a> Scheduler<'a> {
...
+        // because we don't know how many there are at compile time. Metrics 
+        // will be cached and reused so this is not a performance issue.
         for node in lb_stats.iter() {
-            info!("{}", node);
+            histogram!("load_avg", "node" => node.id.to_string())
+                .record(node.load.load_avg() as f64);
             for dom in node.domains.iter() {
-                info!("{}", dom);
+                histogram!("load_avg", "node" => node.id.to_string(), "dom" => dom.id.to_string())
+                    .record(dom.load.load_avg() as f64);
```
Byte-Lab, line 566: Let's not do this in this PR to keep the footprint small (we're already passing `lb_stats` to this function so I think what you have here is correct), but should we maybe move this into `load_balance.rs` instead? I think it makes sense for `main.rs` to print any of the statistics exported from BPF, but it might also be nice to let the load balancing subsystem track its own stats (and in general to allow schedulers to distribute how stats are collected).

Byte-Lab (approved):
Awesome work, this looks great. Just left two very small nits and a thought for a subsequent PR that has no bearing on this one. Once you push those small nit fixes I'll merge.

rust/scx_utils/src/log_recorder.rs
```diff
@@ -0,0 +1,276 @@
...
+    fn register_histogram(&self, key: &Key, _: &Metadata<'_>) -> Histogram {
+        self.registry
+            .get_or_create_histogram(key, |h: &Arc<metrics_util::AtomicBucket<f64>>| {
+                h.clone().into()
+            })
+    }
+}
+
+impl LogRecorder {
+    // Starts starts a background thread that logs the metrics at an interval
```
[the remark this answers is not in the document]
jfernandez: Fixed

scheds/rust/scx_rusty/src/main.rs
jfernandez: Fixed

scheds/rust/scx_rusty/src/main.rs
jfernandez: Yeah, this is an artifact of having to centralize metric printing in the `report` fn. I started moving some stuff out and this will be next. I wasn't sure how you'd feel about it but sounds like we are aligned :)

jfernandez:
The failure in CI appears to be unrelated to this change:
```
23:56:35 [INFO] Histograms:
23:56:35 [INFO]   cpu_busy_pct: avg=NaN min=0.00 max=0.00
23:56:35 [INFO]   processing_duration_us: avg=NaN min=0.00 max=0.00
23:56:35 [INFO] ---
thread 'main' panicked at src/load_balance.rs:689:25:
```
I will fix the NaN in the print output though.

PR #388, opened 2024-06-25
scheds: introduce scx_bpfland

arighi:
Overview
========

This scheduler is derived from scx_rustland, but it is fully implemented in BFP with minimal user-space Rust part to process command line options, collect metrics and logs out scheduling statistics.

Unlike scx_rustland, all scheduling decisions are made by the BPF component.

Motivation
==========

The primary goal of this scheduler is to act as a performance baseline for comparison with scx_rustland, allowing for a better assessment of the overhead caused by kernel/user-space interactions.

It can also be used to deploy prototypes initially tested in the scx_rustland scheduler. In fact, this scheduler is expected to outperform scx_rustland, due to the elimitation of the kernel/user-space overhead.

Scheduling policy
=================

scx_bpfland is a vruntime-based sched_ext scheduler that prioritizes interactive workloads. Its scheduling policy closely mirrors scx_rustland, but it has been re-implemented in BPF with some small adjustments.

Tasks are categorized as either interactive or regular based on their average rate of voluntary context switches per second: tasks that exceed a specific voluntary context switch threshold are classified as interactive.

Interactive tasks are prioritized in a higher-priority DSQ, while regular tasks are placed in a lower-priority DSQ. Within each queue, tasks are sorted based on their weighted runtime, using the built-in scx vtime ordering capabilities (scx_bpf_dispatch_vtime()).

Moreover, each task gets a time slice budget. When a task is dispatched, it receives a time slice equivalent to the remaining unused portion of its previously allocated time slice (with a minimum threshold applied).

This gives latency-sensitive workloads more chances to exceed their time slice when needed to perform short bursts of CPU activity without being interrupted (i.e., real-time audio encoding / decoding workloads).

Results
=======

According to the initial test results, using the same benchmark "playing a videogame while recompiling the kernel", this scheduler seems to provide a +5% improvement in the frames-per-second (fps) compared to scx_rustland, with video games such as Cyberpunk 2077, Counter-Strike 2 and Baldur's Gate 3.

Initial test results indicate that this scheduler offers around a +5% improvement in frames-per-second (fps) compared to scx_rustland when using the benchmark "playing a video game while recompiling the kernel".

This improvement was observed in games such as Cyberpunk 2077, Counter-Strike 2, and Baldur's Gate 3.


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,490 @@
...
+			weight *= weight_boost;
+	}
+
+	/*
+	 * Evaluate weighted time slice and assign the task a time slice within
+	 * the range [1, slice_ns]. Assigning a minimum slice of 1 ensures
+	 * strictly monotonic progress of the vruntime.
+	 */
+	slice = MAX(slice * 100 / weight, 1);
+	slice = MIN(slice, slice_ns);
```
Byte-Lab, line 397: Hmm, so for vtime we theoretically _want_ some tasks to accumulate more than a slice worth of vtime if they have a low weight. Say they ran for `slice_ns` and had a weight of 1. That should mean that they should get 1/100 as much CPU as a task that with the default weight of 100. If we did `slice = MIN(slice, slice_ns)`, then they'll always accumulate the same as a task with weight 100.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,490 @@
...
+	tctx = lookup_task_ctx(p);
+	if (!tctx)
+		return;
+
+	/*
+	 * Classify interactive tasks based on the average amount of their
+	 * voluntary context switches and automatically boost their priority
+	 * (that will affect their vruntime).
+	 */
+	if (nvcsw_thresh && weight_boost) {
```
Byte-Lab, line 384: Should we do this check after we've updated `tctx->avg_nvcsw`?

Byte-Lab (commented):
Hey Andrea, overall this looks great. Just left a few small comments / questions.

sirlucjan:
First conclusions: it works quite efficiently under heavy load, but the application takes noticeably longer to start than with rusty/rustland. In addition, when stopping and resuming video on YT, the sound crackles slightly but the system itself runs completely smooth. No freezes or hangs. After all, for a first attempt, it is better than good. Thanks @arighi

sirlucjan:
After the latest update, the sound problems disappeared.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,490 @@
...
+}
+
+void BPF_STRUCT_OPS(bpfland_enqueue, struct task_struct *p, u64 enq_flags)
+{
+	/*
+	 * Always dispatch per-CPU kthreads immediately.
+	 *
+	 * This allows to prioritize critical kernel threads that may
+	 * potentially slow down the entire system if they are blocked for too
+	 * long (i.e., ksoftirqd/N, rcuop/N, etc.).
```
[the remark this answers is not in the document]
arighi: Correct, I'll clarify this in the comment. I agree that it shouldn't be a problem considering that the target for this scheduler is mostly desktop environments, but it's good to have it written down at least.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,490 @@
...
+	/*
+	 * Classify interactive tasks based on the average amount of their
+	 * voluntary context switches and automatically boost their priority
+	 * (that will affect their vruntime).
+	 */
+	if (nvcsw_thresh && weight_boost) {
+		if (tctx->avg_nvcsw >= nvcsw_thresh)
+			tctx->is_interactive = true;
+		if (tctx->is_interactive)
+			weight *= weight_boost;
```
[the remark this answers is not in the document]
arighi: Thanks for noticing it, you're absolutely right! That was basically copied from rustland, in the first implementation of bpfland I was dispatching all the tasks to the same `SHARED_DSQ` and I wasn't very happy about performance, so I tried to add a separate `PRIO_DSQ` and it was definitely better. In fact, I'm planning to apply the same change to rustland as well (I already have a patch in my local repo). With that in place the whole "priority boost" logic becomes useless. Therefore, we can definitely drop the priority boosting, I'll update this in the next push.

Side note: it's interesting to notice that `scx_rustland` has been used to prototype `scx_bpfland` and vice versa (for the `PRIO_DSQ`), so it was actually useful to re-implement the same scheduler in a different way, this could be yet another example to prove the usefulness of having easy-to-test pluggable schedulers.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Right.. the idea here was to put an upper limit for the slice, to make sure it never becomes too big, but `slice_ns` is probably too small, because it would "smooth" too much the different tasks' priorities. Maybe I should introduce a `slice_ns_max` or limit the slice to `slice_ns` and then apply the weight? (I'll do some tests with both)

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Interesting... in this way we would get a more accurate classification of the task as soon as the metrics are updated. I like this, will do some tests with it!

Good point, thanks for your valuable review @Byte-Lab !

scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -0,0 +1,265 @@
...
+
+struct Metrics {
+    nr_running: Gauge,
+    nr_kthread_dispatches: Gauge,
+    nr_direct_dispatches: Gauge,
+    nr_prio_dispatches: Gauge,
+    nr_shared_dispatches: Gauge,
+}
+
+impl Metrics {
```
jfernandez, line 105: Nice to see that you are using the metrics framework!

arighi: @jfernandez yep, thanks for adding it! I'm planning to use the metrics in scx_rustland as well.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,490 @@
...
+	 * the range [1, slice_ns]. Assigning a minimum slice of 1 ensures
+	 * strictly monotonic progress of the vruntime.
+	 */
+	slice = MAX(slice * 100 / weight, 1);
+	slice = MIN(slice, slice_ns);
+
+	/*
+	 * Update task's vruntime.
+	 */
+	p->scx.dsq_vtime += slice;
```
multics69, line 402: It seems that `p->scx.dsq_vtime` is monotonically increasing. If it is true, does it mean a long-running task could be penalized over a freshly forked task? Am I correct?

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,490 @@
...
+
+	/*
+	 * Wake-up the target CPU to make sure that the task is consumed as
+	 * soon as possible.
+	 *
+	 *  We know the CPU is idle because direct dispatching only occurs to
+	 *  idle CPUs from select_cpu(), so using SCX_KICK_IDLE is unnecessary
+	 *  in this case.
+	 */
+	scx_bpf_kick_cpu(cpu, 0);
```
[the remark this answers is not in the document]
arighi: it's not just a performance optimization, it's actually required: we know that the CPU is idle and the task is dispatched to a DSQ that can be consumed only from the corresponding CPU (note that I'm not using `SCX_DSQ_LOCAL` here). If we don't kick the CPU the task will be stalled indefinitely in the DSQ.

But I will definitely add a comment to clarify this aspect.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,490 @@
...
+	 * so they should naturally release the CPU quickly and give a chance
+	 * to the regular tasks to run.
+	 */
+	if (scx_bpf_consume(PRIO_DSQ))
+		return;
+
+	/*
+	 * Lastly, consume regular tasks from the shared DSQ.
+	 */
+	scx_bpf_consume(SHARED_DSQ);
```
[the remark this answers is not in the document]
arighi: That is a really good idea. I'll add a comment for now and in the future I'll add a tunable for that, thanks!

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: It should be the opposite, tasks are executed in order from the smallest vtime to the largest vtime. So we are actually penalizing the newly forked tasks. In fact one issue with this scheduler is that it's not very responsive at forking new tasks, like opening a new terminal session when the system is overloaded...

arighi: Meh I think I got the vruntime logic backward... then we can definitely improve the scheduler, lol! @multics69 thank you so so much for your review! Will push a fix for this.

arighi:
Alright, I think I've addressed pretty much all the comments from @Byte-Lab and @multics69 (thanks tons for your review!). I'll wait for a feedback from the CachyOS community to make sure there are no major regressions, but so far so good on my side.

arighi:
`scx_bpfland` seems to be in pretty good shape at this point.

Given the extensive testing and positive feedback from the CachyOS community (no stall, regression, bug, etc.), I believe we can merge this and continue refining the scheduler in-tree. Additionally, I would like to start porting some of the changes from `scx_bpfland` back to `scx_rustland`, so that they can be both used to prototype ideas one for the other.

In terms of performance, `scx_bpfland` significantly outperforms `scx_rustland`, especially  in terms of throughput (as expected) and is nearly on par with `scx_rusty` for certain workloads. Its small memory footprint can make it an ideal choice for systems with limited memory capacity, such as embedded devices or small cloud instances.

In conclusion, I believe `scx_bpfland` can be a valuable addition to our pool of sched_ext schedulers. Therefore, I'm merging this.

Let me know if you have any objection, concern, comment, thanks!

PR #394, opened 2024-06-27
scx_utils: clarify error about missing CONFIG_DEBUG_INFO_BTF

arighi:
If CONFIG_DEBUG_INFO_BTF is not enabled in the kernel, the C schedulers report the following error via libbpf, clearly indicating the missing kernel config:

 libbpf: kernel BTF is missing at '/sys/kernel/btf/vmlinux', was CONFIG_DEBUG_INFO_BTF enabled?

In contrast, the Rust schedulers report a less clear error:

 thread 'main' panicked at /home/arighi/src/scx/rust/scx_utils/src/compat.rs:23:9:
 btf__load_vmlinux_btf() returned NULL

Make sure to report a similar error, so that users have a better clue about the missing kernel config. After this change the error looks like the following:

 thread 'main' panicked at /home/arighi/src/scx/rust/scx_utils/src/compat.rs:23:9:
 btf__load_vmlinux_btf() returned NULL, was CONFIG_DEBUG_INFO_BTF enabled?

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -20,7 +20,7 @@ lazy_static::lazy_static! {
 fn load_vmlinux_btf() -> &'static mut btf {
     let btf = unsafe { btf__load_vmlinux_btf() };
     if btf.is_null() {
-        panic!("btf__load_vmlinux_btf() returned NULL");
+        panic!("btf__load_vmlinux_btf() returned NULL, was CONFIG_DEBUG_INFO_BTF enabled?")
     }
     unsafe { &mut *btf }
 }


sirlucjan (approved):
Good job!

arighi:
BTW @htejun is there any reason why we don't just depend on `DEBUG_INFO_BTF` on the kernel side? Like:
```
diff --git a/kernel/Kconfig.preempt b/kernel/Kconfig.preempt
index f035c87d02f1..f3d140c3acc1 100644
--- a/kernel/Kconfig.preempt
+++ b/kernel/Kconfig.preempt
@@ -135,7 +135,7 @@ config SCHED_CORE
 
 config SCHED_CLASS_EXT
 	bool "Extensible Scheduling Class"
-	depends on BPF_SYSCALL && BPF_JIT
+	depends on BPF_SYSCALL && BPF_JIT && DEBUG_INFO_BTF
 	help
 	  This option enables a new scheduler class sched_ext (SCX), which
 	  allows scheduling policies to be implemented as BPF programs to
```

htejun (approved):

PR #397, opened 2024-06-28
sched_utils: Add log recorder format customization

jfernandez:
This change adds the ability to customize the log recorder format for each metric type. There is a default format that is used if no custom `MetricFormatter` is provided. This is the same format that was used before this change.

The `MetricFormatter` should be implemented by the user to customize the format of the log recorder. The `LogRecorderBuilder` now takes a `MetricFormatter` as an optional parameter.

Following changes will allow additional customization of the log recorder format, such as how many metrics are logged per line.

### DIFF
--- a/rust/scx_utils/src/log_recorder.rs
+++ b/rust/scx_utils/src/log_recorder.rs
@@ -37,12 +37,14 @@ use metrics_util::registry::Registry;
 /// ```
 pub struct LogRecorderBuilder {
     reporting_interval: Duration,
+    metric_formatter: Box<dyn MetricFormatter + Send + Sync>,
 }
 
 impl LogRecorderBuilder {
     pub fn new() -> LogRecorderBuilder {
         Self {
             reporting_interval: Duration::from_secs(3),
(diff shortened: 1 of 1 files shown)


jfernandez:
One more note: I don't expect this LogRecorder to work for every scheduler, even after we enable more custom formatting. I think it's fine to continue to print explicitly to meet personal preferences. The primary goal of introducing metrics-rs is to enable exporting of metrics to backends. The LogRecorder is a nice to have if the format works for you, or you are bootstrapping a new scheduler. It also a good example that others can follow to create customer exporters.

arighi (approved):
This looks useful to me and I'd like to start using it in my schedulers, thanks @jfernandez !

jfernandez:
Thanks @arighi.

I'm sharing context from the Slack discussion on this topic: @htejun suggested an additional hook that gets all counters in a structure so that you can customize how to group them visually. I think this is a good idea and will implement it in the following PR.

htejun (approved):

htejun (approved):

PR #400, opened 2024-06-28
meson: restore previous libbpf version and update bpftool

arighi:
The upstrem bpftool git repo (https://github.com/libbpf/bpftool.git) is periodically force pushed and the specific commit that we needed is not available anymore.

Instead of failing we are actually fetching the latest bpftool (HEAD) that introduced some breakage initially fixed by commit e59c48a6 ("Update libbpf commit hash").

However, updating libbpf seems to introduce a run-time problem and all the schedulers are failing to start:

 libbpf: failed to find skeleton map ''
 libbpf: failed to populate skeleton maps for 'bpf_bpf': -3

So, revert libbpf to the previous version and update the commit for bpftool to use a version that still allows to generate a compatible BPF skel.

Fixes: e59c48a6 ("Update libbpf commit hash")

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -124,7 +124,7 @@ if should_build_libbpf
   endforeach
 
   message('Fetching libbpf repo')
-  libbpf_commit = 'c1a6c770c46c6e78ad6755bf596c23a4e6f6b216'
+  libbpf_commit = '42065ea6627ff6e1ab4c65e51042a70fbf30ff7c'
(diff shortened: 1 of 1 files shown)


ptr1337:
Fixes the segfaults and co.

Tested-by: Peter Jung <ptr1337@cachyos.org>

arighi:
Merging this so that we can continue to build and test the scx schedulers. We will investigate later how to move to the new libbpf properly if it's needed.

PR #401, opened 2024-06-28
meson: run cargo build in release mode when using plain buildtype

vnepogodin:
On archlinux its suggested to use plain buildtype in order to not ignore system provided flags, for example for packaging


arighi (approved):
I was wondering if we should also include `minsize` here, but oddly enough it's supposed to have debug enabled (https://mesonbuild.com/Builtin-options.html). I guess it's just some meson weirdness... anyway, LGTM, thanks!

PR #405, opened 2024-06-28
Run CI-tests with latest kernel

otteryc:
Since sched_ext has been integrated into the kernel source, the reference to the GitHub sched-ext/sched_ext repository is no longer up-to-date. This commit updates the CI configuration to run tests with the latest kernel.


htejun:
This will eventually be what we want to do but right now the upstream tree doesn't have all the features that are in the CI branch and will cause some schedulers to fail, so let's wait a bit till they reach parity.

htejun (approved):

htejun:
This got fixed by @arighi's recent PR to fix CI (needed more changes). Closing. Thanks.

PR #410, opened 2024-07-03
scx_bpfland: enhance performance consistency and predictability

arighi:
A set of changes to make performance more consistent and prevent spikes:
 - scx_bpfland: adjust default time slice to 5ms (default tuning for latency)
 - scx_bpfland: ensure task time slice never exceeds the slice_ns limit (latency improvement)
 - scx_bpfland: prevent starvation of regular tasks (prevent interactive tasks from monopolizing the CPUs)
 - scx_bpfland: unify dispatching kthreads with direct CPU dispatches (dispatch kthreads using per-CPU DSQs instead of `SCX_DSQ_LOCAL`)
 - scx_bpfland: drop built-in idle CPU selection logic (drop unused option / cleanup)

Thanks to the CachyOS community that provided an extensive amount of test results and feedback about these changes.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -51,11 +51,6 @@ const volatile u64 slice_ns_min = 500000;
 const volatile u64 slice_ns_lag;
 
 /*
- * Enable built-in idle selection logic.
- */
-const volatile bool builtin_idle;
-
-/*
  * Threshold of voluntary context switches used to classify a task as
  * interactive.
  */
@@ -164,6 +159,30 @@ static int calloc_cpumask(struct bpf_cpumask **p_cpumask)
 		bpf_cpumask_release(cpumask);
 
 	return 0;
+}
+
+/*
+ * Set the state of a CPU in a cpumask.
+ */
+static bool set_cpu_state(struct bpf_cpumask *cpumask, s32 cpu, bool state)
+{
+	struct bpf_cpumask *mask;
+	int ret = false;
+
+	bpf_rcu_read_lock();
+
+	mask = cpumask;
+	if (!mask)
+		goto out_rcu;
+	if (state)
+		ret = bpf_cpumask_test_and_set_cpu(cpu, mask);
+	else
+		ret = bpf_cpumask_test_and_clear_cpu(cpu, mask);
+
+out_rcu:
+	bpf_rcu_read_unlock();
+
+	return ret;
 }
 
 /*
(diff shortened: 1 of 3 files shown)


sirlucjan:
I've been testing it for a few days. Everything works very well.

ptr1337:
Also tested by me. Performance seems very good as well as reliability and interactivity.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -166,6 +161,30 @@ static int calloc_cpumask(struct bpf_cpumask **p_cpumask)
...
 
+/*
+ * Set the state of a CPU in a cpumask.
+ */
+static bool set_cpu_state(struct bpf_cpumask *cpumask, s32 cpu, bool state)
+{
+	struct bpf_cpumask *mask;
+	int ret = false;
+
+	bpf_rcu_read_lock();
```
htejun, line 178: This isn't necessarily from this PR but what is RCU read lock protecting?

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -56,6 +56,13 @@ const volatile u64 slice_ns_lag;
  */
 const volatile u64 nvcsw_thresh = 10;
 
+/*
+ * Schedule up to the starvation_thresh number of priority tasks before being
+ * required to schedule a regular task. This ensures that regular tasks are not
+ * starved by interactive tasks.
+ */
+const volatile u64 starvation_thresh = 100;
```
htejun, line 64: Down the line, it *may* lead to more consistent behavior if the mechanism is time based rather than the number of events. Another thing to consider is that if there are tasks which are affine to some subset of CPUs, they might still be able to starve, so maybe a mechanism which can transfer timing out batch tasks to per-cpu DSQs will probably be more fool-proof. Just ideas for possible future improvements.

htejun (approved):

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Hm.. I had the impression that the verifier was complaining without the RCU protection, but it's not the case (and I also don't see anything to protect here). Probably it was needed to protect something else that was moved after some refactoring. Anyway, thanks for noticing it, I'll drop the RCU read lock here!

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Yes, a time-based approach would definitely be more effective. I'll work on that.

I also like the idea of transferring tasks to per-CPU DSQs. That's something I coud use also for the CPU hotplug support, to transfer tasks to another DSQ from the `.cpu_offline()` callback, but I haven't found a way to pop tasks from a non-local DSQ and move them to another DSQ, except via `scx_bpf_consume()`, but this doesn't give me the task, it just "internally" transfers it to the local DSQ and of course I can't use consume in `cpu_offline()`...

arighi:
Removed the unnecessary RCU read lock protection + implemented a better time-based starvation prevention. Tested with the usual gaming + building kernel, everything looks good so far (better than before apparently).

arighi:
Merging this as it seems pretty stable and well-tested.

PR #417, opened 2024-07-11
scx_lavd: improve virtual deadline and current clock handling

multics69:
The key to this commit is the change from a physical clock to a virtual, logical clock in calculating deadlines. This commit guarantees that there is no such task whose virtual deadline is smaller than the virtual current clock. This means any enqueuing task can compete with any other already enqueued tasks. This allows a latency-critical task to be immediately scheduled if needed.

Besides this major change, there are code clean-ups, including:
- Drop lat_prio_to_greedy_thresholds table
- Induce sched_prio_to_latency_weight table from slice weight

Plus, there is a minior tuning to avoid fork bomb.


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -2951,6 +2966,10 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(lavd_init)
 	if (err)
 		return err;
 
+	/*
+	 * Initilize the current logical clock.
+	 */
+	WRITE_ONCE(cur_logical_clk, now);
```
htejun, line 2972: I wonder whether it'd be less confusing to start `cur_logical_clk` from some set value (e.g. 0) rather than wallclock time.

multics69: Thanks @htejun for the suggestion. I will update it and made some minior changes then merge it.

PR #421, opened 2024-07-12
scx_lavd: improve time slice and waker freq calculation

multics69:
This PR contains two improvements of scx+lavd:

1) Change the time slice calculation based on the task's ratio in total weights to the total runnable count.  The weight-based calculation deprioritizes nicer tasks too much, but the runnable count-based scheme does not.

2) Improve the accuracy of waker frequency. In the prior approach, any `current` task making another task runnable was considered a waker. However, that is not always true. So we consider the `current` task is a waker only when two tasks are related (e.g., parent-child, etc.)

With these two changes, the scheduler becomes more resilient under heavy load.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -2428,6 +2428,12 @@ freq_out:
 		try_decrease_cpuperf_target(cpuc_run);
 }
 
+static bool are_related_tasks(struct task_struct *p, struct task_struct *q)
+{
+	return p->parent == q || q->parent == p ||
+	       p->tgid == q->tgid || p->parent == q->parent;
+}
+
 void BPF_STRUCT_OPS(lavd_runnable, struct task_struct *p, u64 enq_flags)
 {
 	struct cpu_ctx *cpuc;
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -2455,7 +2461,13 @@ void BPF_STRUCT_OPS(lavd_runnable, struct task_struct *p, u64 enq_flags)
 	if (!is_wakeup_ef(enq_flags))
 		return;
 
+	/*
+	 * Filter out unrelated tasks.
+	 */
 	waker = bpf_get_current_task_btf();
+	if (!are_related_tasks(p, waker))
```
htejun, line 2468: You can reliably detect whether `@p` is being waken up by the `current` task by testing `SCX_ENQ_WAKEUP` in `@enq_flags`.

htejun (approved):

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Hmm... it is strange. I already check SCX_ENQ_WAKEUP in is_wakeup_ef(). However, I still see irrelevant pairs of {`@p`, `current`} are with the SCX_ENQ_WAKEUP flag. Is there any possibility that the kernel config might affect? I am using the CachyOS kernel.

htejun: AFAICS, `SCX_ENQ_WAKEUP` can only be set when `ops.enqueue()` is being called from `try_to_wakeup()`. Maybe there is a bug in the direct dispatch code and we're leaking flags across enqueues but I don't immediately see how that could happen.

multics69: Hmm... I will merge this PR to the main for now. Then will further investigate why we are getting `SCX_ENQ_WAKEUP` unexpectedly.

PR #429, opened 2024-07-14
scx_bpfland: small improvements

arighi:
Small improvements for the CPU hotplug support, time slice evaluation and in the logic to classify interactive/regular tasks. No major improvement in this PR, mostly small fixes/adjustments to enhance the quality of the code.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -5,21 +5,6 @@
 #include <scx/common.bpf.h>
 #include "intf.h"
 
-/*
- * Maximum amount of CPUs supported by the scheduler.
- */
-#define MAX_CPUS	1024
-
-/*
- * DSQ used to dispatch regular tasks.
- */
-#define SHARED_DSQ	MAX_CPUS
-
-/*
- * Priority DSQ used to dispatch interactive tasks.
- */
-#define PRIO_DSQ	(MAX_CPUS + 1)
-
 char _license[] SEC("license") = "GPL";
 
 /* Allow to use bpf_printk() only when @debug is set */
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -333,6 +356,18 @@ static int dispatch_direct_cpu(struct task_struct *p, s32 cpu, u64 enq_flags)
...
 	scx_bpf_dispatch_vtime(p, dsq_id, slice, vtime, enq_flags);
 
+	/*
+	 * If the CPU has gone offline notify that the task needs to be
+	 * consumed from another CPU.
+	 */
+	offline = offline_cpumask;
+	if (!offline)
+		return 0;
+	if (bpf_cpumask_test_cpu(cpu, cast_mask(offline))) {
```
htejun, line 363: Hmm... this is a bit tricky. I think this can be a lot simpler if we had a way to wait a RCU grace period.

htejun (approved):

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -324,14 +319,16 @@ static inline u64 task_slice(struct task_struct *p)
...
 
 /*
- * Return the DSQ ID associated to a CPU, or SHARED_DSQ if the CPU is not
+ * Return the DSQ ID associated to a CPU, or shared_dsq_id if the CPU is not
  * valid.
  */
 static u64 cpu_to_dsq(s32 cpu)
 {
-	if (cpu < 0 || cpu >= MAX_CPUS) {
+	u64 cpu_max = scx_bpf_nr_cpu_ids();
```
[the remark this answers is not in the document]
arighi: Yes, good point, I'll change it to nr_cpu_ids, so it's more consistent with the rest of the code.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: It is quite ugly, also offline_cpumask is allocated once at init and it goes away only when the scheduler exits (it's never re-allocated), this is just to make the verifier happy...

htejun (approved):

PR #436, opened 2024-07-16
rust Updated libbpf-rs & libbpf-cargo to 0.24

danielocfb:
Use updated libbpf-rs & libbpf-cargo. Generated skeletons now contain directly accessible map and program objects, no longer necessitating the use of accessor methods. As a result, the risk for mutability conflicts is reduced greatly.


danielocfb:
Not going to un-draft this until the next `libbpf-rs` release is cut (which could take a while). We could land it as-is conceptually, but I suspect you guys don't want to depend on a random `libbpf-rs` snapshot, as that would block any attempt of publishing your crates.

danielocfb:
> Not going to un-draft this until the next `libbpf-rs` release is cut

`0.24.0` is out, so I think this change can be reviewed now.

arighi (approved):

danielocfb:
Please wait with merge a little longer. Still investigating a potential issue.

danielocfb:
> Please wait with merge a little longer. Still investigating a potential issue.

We should be good now.

htejun:
FYI, @danielocfb for some reason, your commit's commit date is earlier than its parent and `tig` has trouble rendering it correctly.

PR #438, opened 2024-07-17
Refactor select_cpu + enqueue for proper synchronization and handle of !wakeup

dschatzberg:
Two (significant) related commits here:

1) refactor enqueue() so it works on !wakeup (where tasks bypass select_cpu)
2) Fix the synchronization to actually ensure enqueue() has been invoked with some RCU-like pattern.


htejun:
I think it's lacking necessary memory barriers even on x86 as rmb isn't implied but this looks like a great starting point.

htejun (approved):

htejun:
Is this PR still in progress?

dschatzberg:
I was planning on repurposing it just for the RCU bits since there's some
context. Hopefully next week I'll update it

PR #439, opened 2024-07-18
scx_bpfland: introduce dynamic nvcsw threshold

arighi:
Instead of using a static value to classify tasks based on their average amount of voluntary context switches, try to periodically evaluate an optimal threshold, based on a global average of voluntary context switches among of all the running tasks.

Tasks with an average amount of voluntary context switches greater than the global average will be classified as interactive.

The global average is evaluated as an exponentially weighted moving average (EWMA), as:

  avg(t) = avg(t - 1) * 0.75 - task_avg(t) * 0.25

This approach is more efficient than iterating through all tasks and it helps to prevent rapid fluctuations that may be caused by bursts of voluntary context switch events.

The dynamic nvcsw threshold enables a more precise adjustment of the classification criteria to swiftly respond to global system changes: tasks can be quickly classified as interactive, but if the system experiences too many interactive events, the criteria for maintaining interactive status become stricter. This creates a natural selection process where only the most deserving tasks remain interactive.

Additionally, introduce the new option `--nvcsw-max-thresh N`, which allows to extend or restrict the fluctuation range of the global average threshold for voluntary context switches.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -59,10 +59,18 @@ const volatile u64 slice_ns_lag;
 const volatile bool local_kthreads;
 
 /*
- * Threshold of voluntary context switches used to classify a task as
- * interactive.
+ * Maximum threshold of voluntary context switches.
+ *
+ * This limits the range of nvcsw_avg_thresh (see below).
  */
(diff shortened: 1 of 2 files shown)


sirlucjan:
Feel free to add

Tested-by: Piotr Gorski <piotrgorski@cachyos.org>

htejun (approved):

PR #441, opened 2024-07-19
scx_layered: Add separate module for metrics

hodgesds:
Refactor the main module for scx_layered to move metrics into a separate module. This change does no functional differences, only code structure. This will make it a little easier to navigate the logic in the main scheduler code.

### DIFF
--- a/scheds/rust/scx_layered/src/metrics.rs
+++ b/scheds/rust/scx_layered/src/metrics.rs
@@ -1,0 +1,197 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+use std::sync::atomic::AtomicI64;
+use std::sync::atomic::AtomicU64;
+use prometheus_client::metrics::family::Family;
+use prometheus_client::metrics::gauge::Gauge;
+use prometheus_client::registry::Registry;
+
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/metrics.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub l_kick: Family<Vec<(String, String)>, Gauge<f64, AtomicU64>>,
+    pub l_yield: Family<Vec<(String, String)>, Gauge<f64, AtomicU64>>,
+    pub l_yield_ignore: Family<Vec<(String, String)>, Gauge<i64, AtomicI64>>,
+    pub l_migration: Family<Vec<(String, String)>, Gauge<f64, AtomicU64>>,
+    pub l_cur_nr_cpus: Family<Vec<(String, String)>, Gauge<i64, AtomicI64>>,
+    pub l_min_nr_cpus: Family<Vec<(String, String)>, Gauge<i64, AtomicI64>>,
+    pub l_max_nr_cpus: Family<Vec<(String, String)>, Gauge<i64, AtomicI64>>,
+}
+
+impl OpenMetricsStats {
```
hodgesds, line 57: I almost wish this had a `report` method that was most of the logic from the `Sched::report` method, but that's too much refactoring.

htejun (approved):

PR #442, opened 2024-07-19
layered: Don't dispatch to LO_FALLBACK_DSQ

Byte-Lab:
Non-kthreads with custom affinities in non-open layers are dispatched into a LO_FALLBACK_DSQ, with the idea being that they're penalized for their custom affinities. When a host is fully utilized, these tasks can end up being starved due to LO_FALLBACK_DSQ being consumed only when there are no other layers to consume from. In internal workloads at Meta, we've observed that this can happen in practice.

Longer term, we can probably address this by implementing layer weights and applying that to fallback DSQs to avoid starvation. For now, let's just dispatch them to the same fallback DSQ to avoid this starvation issue.


htejun:
Oh, let's keep it and just replace the enqueue on `LO` with `HI` with comment explaining what's wrong. We want to come back to this later and it'd be better to have documentation on what we wanted but didn't work.

Byte-Lab:
Forgot to add a comment, will do that and repush shortly

htejun (approved):

PR #443, opened 2024-07-20
scx_lavd: overhaul the virtual deadline algorithm

multics69:
This PR contains the major overhaul of the virtual deadline algorithm and relevant code cleanup. 

- Instead of stretching the time-space upon enqueue, we now advance the current virtual clock reverse proportionally to the system load. Under overload, the clock goes slower, so latency-critical tasks have more chance to cut in the timeline.

- Instead of estimating service time (vruntime in CFS) from runtime and run frequency, we now directly measure the service time for eligibility enforcement.

- We drop the runtime factor in calculating latency criticality since it is already considered in calculating the task's virtual deadline. Instead, we additionally consider the task's starvation factor (i.e., how much a task starved from the average service time) in calculating the latency criticality. By incorporating the starvation factor, we can systematically avoid the watchdog time-out error from the scx framework.

- Instead of inheriting the parent's properties for a forked task, a forked task will be treated as a greedy task until the scheduler knows its true properties. This helps to avoid stalling under a fork bomb.

- After the overhaul, we cleaned up many unnecessary codes and optimizations. Notably, we dropped the sched_prio_to_slice_weight[] table and directly used p->scx.weight.

- Note that we first tried to maintain ineligible runnable tasks separately. However, we later removed this because it became unnecessary after the overhaul.


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1956,26 +1983,49 @@ static void put_global_rq(struct task_struct *p, struct task_ctx *taskc,
...
 
 	/*
-	 * If the current task has something to yield, try preempt it.
+	 * If the task is ineligible, keep it to the ineligible run queue.
 	 */
-	p_run = bpf_get_current_task_btf();
-	taskc_run = try_get_task_ctx(p_run);
-	if (taskc_run && p_run->scx.slice != 0)
-		try_yield_current_cpu(p_run, cpuc, taskc_run);
+	t = bpf_obj_new(typeof(*t));
```
htejun: I think it should be possible to pre-allocate this node on `task_init()` and keep it on taskc so that enqueue path doesn't have to do dynamic allocations but at the same time `bpf_obj_new()` might be cheap enough for this to not matter.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1472,6 +1462,28 @@ static void update_stat_for_runnable(struct task_struct *p,
...
+	/*
+	 * The clock should not go backward, so do nothing.
+	 */
+	if (taskc->vdeadline_log_clk <= cur_logical_clk) {
+		return;
+	}
+
+	/*
+	 * Advance the clock up to the task's deadline. When overloaded,
+	 * advnace the clock slower so other can jump in the run queue.
```
htejun, line 1231: typo: advnace

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1206,10 +1206,23 @@ static int map_lat_cri_to_lat_prio(u32 lat_cri)
...
+{
+	struct sys_stat *stat_cur = get_sys_stat_cur();
+	u64 ratio;
+
+	/*
+	 * Prioritize tasks whose service time is smaller than average.
+	 */
+	ratio = (LAVD_LC_STARVATION_FT * stat_cur->avg_svc_time) /
+		taskc->svc_time;
+	return ratio + 1;
```
htejun, line 1065: Shouldn't starvation avoidance be part of the virtual timeline management rather than implemented through boosting interactivity? Is this an artifact of eligible and ineligible timelines being managed separately?

htejun: Oh that gets removed later. I'm curious why deadline in itself isn't sufficient for starvation avoidance.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
htejun: nvm, this gets removed later.

htejun (approved):

multics69:
Thanks @htejun for the review. I will merge it to the main.

PR #445, opened 2024-07-23
scx_rusty: Simplify LoadBalancer::populate_tasks_by_load()

danielocfb:
Simplify LoadBalancer::populate_tasks_by_load() by cutting out the heap allocation bits, by moving mutable accesses in front of immutable ones. Because multiple immutable accesses (between bss and rodata) do not conflict, we don't need the intermediate PID storage.


danielocfb:
@Byte-Lab asked privately about this part and whether it can be simplified and the indirection removed. I think it can, but I can't really test it. Take it or leave it. Once https://github.com/sched-ext/scx/pull/436 lands conflict of this kind should mostly be a thing of the past anyway.

Byte-Lab (approved):

PR #449, opened 2024-07-25
scx_layered: Add a hi fallback dsq per llc

hodgesds:
Add a hi fallback dsq per llc to help avoid starvation issues. The proper solution of having weights per layer still needs to be done, but this may be a short term solution.


hodgesds:
I'm still seeing starvation issues with this, so maybe I'll just try a better solution instead.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -862,8 +865,10 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 		}
 	}
 
-	if (scx_bpf_consume(HI_FALLBACK_DSQ))
-		return;
+	bpf_for(idx, 0, nr_llcs) {
+		if (scx_bpf_consume(HI_FALLBACK_DSQ + idx))
```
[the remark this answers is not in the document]
hodgesds: Yeah, I need to fix that up in general with the per llc dsqs.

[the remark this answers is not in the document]
htejun: Yeah, it also should be done in stages - local LLC first, then LLCs in the node, and then (maybe depending on some conditions) LLCs in other nodes.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -862,9 +890,25 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
...
+		if (llc_node_id(llc_id) != node_id)
+			continue;
+
+		next_dsq_id = llc_hi_fallback_dsq_id(llc_id);
+		if (next_dsq_id == dsq_id)
+			continue;
+
+		if (scx_bpf_consume(next_dsq_id))
+			return;
+	}
```
htejun, line 910: Can't we end up in a situation where a task which is affine to one node is queued on the fallback dsq of another node?

hodgesds: Hmmm yeah... so is consuming from the llc local dsq sufficient then? Perhaps I need to rethink the load balancing a bit more.

htejun: Hmm... isn't it in the other direction? We can end up in a situation where a task is queued on a fallback DSQ which will never be consumed by a CPU that can run that task.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -36,7 +36,7 @@ enum consts {
 	USAGE_HALF_LIFE		= 100000000,	/* 100ms */
 
 	HI_FALLBACK_DSQ		= MAX_LAYERS * MAX_DOMS,
```
htejun, line 38: Maybe rename to sth like `HI_FALLBACK_DSQ_BASE` to indicate that it's not just single DSQ?

htejun (approved):

PR #453, opened 2024-07-27
scx_bpfland: adjust time slice based on waiting tasks and online CPUs

arighi:
Scale the time slice based on the average number of waiting tasks normalized to the amount of online CPUs.

This allows to be not too aggressive at reducing the time slice when some tasks are waiting, thus providing a better throughput with more CPU-intensive benchmarks (such as apache / nginx from the Phoronix test suite), while still maintaining a good level of responsiveness with pure latency-sensitive benchmarks (like hackbench).

Moreover, refresh the amount of waiting tasks only after a task has been consumed from the global DSQs; this allows to save some calls to scx_bpf_dsq_nr_queued() while still maintaining a realistic measure about the amount of tasks waiting to be scheduled.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -309,31 +309,18 @@ static inline u64 task_vtime(struct task_struct *p)
 }
 
 /*
- * Return the amount of tasks waiting to be dispatched.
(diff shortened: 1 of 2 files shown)


arighi:
This change seems to introduce some regressions with the "gaming while recompiling the kernel" benchmark. It needs more testing, closing for now.

PR #457, opened 2024-07-28
scx_lavd: support two-level scheduling for heavy-loaded cases (like bpfland)

multics69:
This PR introduces two-level scheduling similar to scx_bpfland. The two-level scheduling consists of two DSQs: 1) latency-critical run queue and 2) regular run queue. The scheduler prioritizes scheduling tasks on the latency-critical queue but makes its best effort to schedule tasks on the regular queue. The scheduler could be more resilient under heavy load by segregating regular, non-latency-critical tasks from latency-critical tasks. Besides this major change, the PR contains minor improvements in calculating latency criticality metrics.

For the record, this is the rest results by Veilink (CachyOS community). 
https://flightlesssomething.duckdns.org/benchmark/109

In the future, this two-level scheduler will be extended to support AMP architecture.


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1749,9 +1754,11 @@ static void put_global_rq(struct task_struct *p, struct task_ctx *taskc,
 	}
 
 	/*
-	 * Enqueue the task to the eligible DSQ based on its virtual deadline.
+	 * Enqueue the task to one of the DSQs based on its virtual deadline.
 	 */
-	scx_bpf_dispatch_vtime(p, LAVD_GLOBAL_DSQ, LAVD_SLICE_UNDECIDED,
+	if (no_2_level_scheduling || is_lat_cri_task(taskc))
```
arighi, line 1759: I was a bit confused by this, but basically with `no_2_level_scheduling` all tasks are considered "latency critical" and they're all dispatched to the `LAVD_LATENCY_CRITICAL_DSQ`. Maybe it could be useful to add a small comment, so it can help the reader to immediately understand this logic.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -2019,7 +2074,32 @@ void BPF_STRUCT_OPS(lavd_dispatch, s32 cpu, struct task_struct *prev)
 		bpf_task_release(p);
 		break;
 	}
+
+	if (no_2_level_scheduling || ret)
+		goto unlock_out;
   
+	bpf_for_each(scx_dsq, p, LAVD_REGULAR_DSQ, 0) {
+		if (is_kernel_task(p)) {
```
arighi, line 2082: This is more a question about sched_ext in general, since we have the same behavior pretty much with all the schedulers, but I'll just ask here.

What happens if the task can't run on the CPU where it's consumed (e.g., if `cpu` is not allowed according to  `p->cpus_ptr)`? Is it just re-enqueued? If that's the case I was wondering if it'd be more efficient to check here if cpu is allowed before consuming the task, otherwise we could move to the next task and kick `scx_bpf_task_cpu(p)`. WDYT?

arighi (approved):
@multics69 this looks great! I already see clear benefits doing some quick tests with an heavy background load. Happy that bpfland inspired this!

I left a couple of comments, but in general I think we can merge this as it is.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1084,40 +1065,48 @@ static void boost_lat(struct task_struct *p, struct task_ctx *taskc,
 	 */
 	wait_freq_ft = calc_freq_factor(taskc->wait_freq);
 	wake_freq_ft = calc_freq_factor(taskc->wake_freq);
-	starvation_ft  = calc_starvation_factor(taskc);
 
 	/*
 	 * Wake frequency and wait frequency represent how much a task is used
 	 * for a producer and a consumer, respectively. If both are high, the
-	 * task is in the middle of a task chain.
+	 * task is in the middle of a task chain. We prioritize a producer.
```
arighi, line 1072: This is quite interesting. I had the impression that we may want to prioritize the consumer, not the producer. I see an analogy with reads/writes, you may want to prioritize readers, that can process their data and pass their result to the next step of the pipeline. By prioritizing writers you may end up filling up all the "buffer" and the less-prioritized readers would block the flow of the pipeline. Do you have some data / info that proves the opposite of this?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Thank you, @arighi, for the suggestion. I will update it as suggested.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: As far as I understand, `scx_bpf_consume` iterates the specified DSQ and consumes the first consumable task, respecting task's cpumask. If there is no such task, it will return false.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: First of all, I didn't conduct very serious study on this. Emperically, I found that many of the tasks that I want to prioritize have higher frequency of waking up the other tasks (i.e., producer). At least, this change helps reducing audio cracking in a certain scenarios.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
arighi: You're right, this check is implemented in `task_can_run_on_remote_rq()`. Hm... I'm wondering how much overhead tasks that can only run on certain CPUs (or even worse only on 1 CPU) can introduce  when they are dispatched to a global DSQ, Would it be more efficient to dispatch them to a per-CPU DSQ? This would allow to save some cycles when the global DSQ is consumed from all the CPUs, but I'm not sure if it's worth it, considering that we may also need to avoid prioritizing these tasks too much... need to do some tests with this.

htejun: It would obviously depend on the frequency but skipping over tasks in DSQ isn't particularly expensive and there usually aren't *that* many tasks with modified cpumasks. I doubt the overhead would be noticeable especially on smaller machines.

htejun (approved):
On a glance, looks fine to me but what's the fairness story? Is it that latency critical tasks aren't supposed to consume too much CPU anyway or more that we don't really care about fairness?

multics69:
Thanks, @htejun, for the comment. You are right. This design is more like a two-level of scheduler class within LAVD. So, it does not guarantee *fair* progress in non-latency-critical tasks. After merging it, I will investigate further how to enforce fairness.

PR #458, opened 2024-07-30
scx_lavd: set correct size for cpu_ctx_stor

arighi:
The max_entries parameter in BPF_MAP_TYPE_PERCPU_ARRAY defines the number of values per CPU and for cpu_ctx_stor we only need one item: the CPU context.

Set max_entries to 1 to avoid allocating unnecessary memory and slightly reduce the memory footprint.


multics69 (approved):
Thanks @arighi ! Looks good to me.

PR #459, opened 2024-07-30
scx_layered: Add per cpu layer iterator offset

hodgesds:
Add a per cpu iterator offset to round robin when iterating on layers. This is to make selection from different layers more fair so that the first layer is not always selected first.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -60,6 +60,37 @@ static inline s32 sibling_cpu(s32 cpu)
 		return -1;
 }
 
+static u32 cpu_ctx_layer_idx_inc(struct cpu_ctx *cctx)
+{
+	if (cctx->layer_idx >= nr_layers || cctx->layer_idx > MAX_LAYERS) {
+		cctx->layer_idx = 0;
+	} else {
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -886,10 +925,15 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 
 	/* consume !preempting open layers */
 	bpf_for(idx, 0, nr_layers) {
+		layer_idx = iter_layer_cpu_ctx(cctx->layer_idx, idx);
+		if (layer_idx > MAX_LAYERS) {
```
hodgesds, line 929: I tried to drop these checks but was hitting verifier errors.

htejun: Does shifting the test into `iter_layer_cpu_ctx()` before returning work?

htejun (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: Yeah, let me try to clean that up.

PR #464, opened 2024-08-02
scx_lavd: improve the calculation of ineligibility duration

multics69:
The key to this PR is to improve the calculation of the ineligibility duration. And, the 2-level scheduling was dropped because the lavd scheduler works well under heavy load with the ineligibility duration optimization. More specifically, the PR includes the following commits:

- drop 2-level-scheduling
- improve the equation for calculating ineligibility duration
- consider task's wakeup and vruntime (starvation) more aggressively
- print the overall status of a scheduled task
- classify CPUs into BIG and little ones based on their average capacity


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1905,7 +1911,7 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 
 static bool is_kernel_task(struct task_struct *p)
 {
-	return p->flags & PF_KTHREAD;
+	return !!(p->flags & PF_KTHREAD);
```
htejun, line 1899: Doesn't really matter but `!!` doesn't make any difference here. The implicit bool casting for the return value does the same thing.

htejun (approved):

PR #467, opened 2024-08-04
scx_lavd: Make FlatTopology::new() a bit prettier

htejun:
- Use .enumerate() consistently while building the cpu_fids vector.

- Use .then_with() to chain .cmp() when sorting cpu_fids.

Both reduce visual clutter.


multics69 (approved):
This looks much better! Thanks @htejun !

PR #468, opened 2024-08-04
scx_rustland refactoring

arighi:
First set of changes to re-align `scx_rustland` with the recent changes done in `scx_bpfland`, mostly re-using the same idle selection logic and the CPU hotplugging support.

The other major change involves removing the interactive task classification from the BPF component (in `scx_rustland_core`) and re-implementing it fully in user space in `scx_rustland`. This allows to eliminate the overhead of sending the `nvcsw` counter for each task to the user-space part, so all the potential schedulers based on `scx_rustland_core` that don't require this logic can benefit from this.

These changes are potentially introducing performance regressions in `scx_rustland`, but they can be fixed/improved later, these changes are also in preparation to define a better and more generic user-space API, so at the moment `scx_rustland` will prioritize the design of the user-space framework more than the performance (for performance we have `scx_bpfland`).

Last minor change: updated my copyright info in `scx_rustland_core`, `scx_rustland` and `scx_rlfifo`, dropping my old email that doesn't work anymore.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf.rs
+++ b/rust/scx_rustland_core/assets/bpf.rs
@@ -57,9 +57,6 @@ pub const RL_PREEMPT_CPU: u64 = bpf_intf::RL_PREEMPT_CPU as u64;
 /// objects) and dispatch tasks (in the form of DispatchedTask objects), using respectively the
 /// methods dequeue_task() and dispatch_task().
 ///
-/// The CPU ownership map can be accessed using the method get_cpu_pid(), this also allows to keep
-/// track of the idle and busy CPUs, with the corresponding PIDs associated to them.
-///
 /// BPF counters and statistics can be accessed using the methods nr_*_mut(), in particular
 /// nr_queued_mut() and nr_scheduled_mut() can be updated to notify the BPF component if the
 /// user-space scheduler has some pending work to do or not.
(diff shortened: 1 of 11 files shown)


rust/scx_rustland_core/assets/bpf.rs
```diff
@@ -70,7 +68,6 @@ pub struct QueuedTask {
     pub pid: i32,              // pid that uniquely identifies a task
     pub cpu: i32,              // CPU where the task is running (-1 = exiting)
     pub sum_exec_runtime: u64, // Total cpu time
-    pub nvcsw: u64,            // Voluntary context switches
```
htejun, line 73: Does this reduction actually make a difference? Depending on the frequency of the classification, reading it from proc can be a lot more expensive.

htejun (approved):

rust/scx_rustland_core/assets/bpf.rs
arighi: In terms of overhead it's negligible, but the point is that it's a very specific metric required just by a single particular scheduler, potentially we may have other schedulers based on `scx_rustland_core` that don't need this metric at all.

Adding more and more metrics could lead to a bloated `QueuedTask` and it'd like to keep it as small as possible. Especially for the metrics that are already available in user-space. Sure, it's probably slower reading it from procfs, but it's only affecting the task classifier that doesn't need to be extremely fast, so worst case it'll be a bit slower at classifying interactive tasks vs regular tasks.

PR #475, opened 2024-08-08
scx_layered: Migrate to metric-rs framework

jfernandez:
This change migrates scx_layered to the metric-rs framework. The transition is mostly seamless, with one notable exception: when the --open-metrics-format flag is enabled, Prometheus metrics will now be available at localhost:9000 instead of being printed to stdout. The flag name remains unchanged for backward compatibility but will be renamed in a future update to better reflect its new behavior.

```
17:03:24 [INFO] Enabling Prometheus endpoint: http://localhost:9000
17:03:24 [INFO] CPUs: online/possible=16/16 nr_cores=8 17:03:24 [INFO] configuring node 0, LLCs 1
17:03:24 [INFO] configuring llc 0 for node 0
17:03:24 [INFO] Layered Scheduler Attached
```

The log formatting has been preserved exactly as before. A new custom log_recorder has been added to maintain the same log format. Additionally, a new synchronous reporting approach has been implemented. The scx_utils log_recorder runs in a separate background thread, while the scx_layered log_recorder is injected into the Scheduler struct, and the `report()` function is called synchronously. This ensures that logs are flushed at the same interval as the scheduler's reporting loop and makes it easier to share the layer structs needed for some log lines.

There are no changes to the metrics themselves, but future updates will modify some of them to be counters and timers instead of gauges.

I also intend to iterate on removing boilerplate in metric capturing and log reporting. Currently, we define the metrics for capturing but don't have a way to reuse that definition when printing metrics, leading to hardcoded names. A warning has been added for attempts to access non-existent metrics to detect drift. This issue will be addressed in a future update.

[Screencast from 2024-08-08 11-02-20.webm](https://github.com/user-attachments/assets/8d7d3ced-2183-4cc0-8664-3a8d338e17e0)

### DIFF
--- a/scheds/rust/scx_layered/src/log_recorder.rs
+++ b/scheds/rust/scx_layered/src/log_recorder.rs
@@ -1,0 +1,294 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+// Copyright (c) Netflix, Inc.
+// Author: Jose Fernandez
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use std::sync::atomic::Ordering::Relaxed;
+use std::sync::Arc;
+
+use bitvec::vec::BitVec;
+use log::info;
+use log::warn;
+use metrics::Counter;
+use metrics::Gauge;
+use metrics::Histogram;
+use metrics::Key;
+use metrics::KeyName;
+use metrics::Label;
+use metrics::Metadata;
+use metrics::Recorder;
+use metrics::SharedString;
+use metrics::Unit;
+use metrics_util::registry::AtomicStorage;
+use metrics_util::registry::Registry;
+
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_layered/src/log_recorder.rs
```diff
@@ -0,0 +1,294 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+// Copyright (c) Netflix, Inc.
+// Author: Jose Fernandez
+
```
jfernandez, line 4: A significant portion of this file was taken from log reporting logic in main.rs

Please advice whether this header correctly captures this or if you'd like me to do any modifications.

jfernandez:
Closing this in favor of https://github.com/sched-ext/scx/pull/496

PR #479, opened 2024-08-09
version-tool: Suppress spurious dependency version mismatch warning

htejun:
When updating in-tree crate versions, do_rust_ver() was returning the new version string which is then put into rust_deps triggering version mismatch warning in do_rust_deps().

There's no reason for do_meson_ver() and do_rust_ver() to return the new version. Always return the old version.


hodgesds (approved):
LGTM

PR #491, opened 2024-08-12
scx_bpfland: primary domain

arighi:
This adds the concept of "primary scheduling domain" to bpfland, that can be used to prioritize scheduling on a subset of cores. This is mostly focused at supporting hybrid architectures (systems with P-cores / E-cores).

The user can define the primary scheduling domain via the new option `--primary-domain CPUMASK`; if a primary domain is define the scheduler will try to dispatch tasks on the cores specified by `CPUMASK`, unless they become saturated, at which points tasks will be allowed to overflow on the other available cores.

This allows to define multiple performance profiles in hybrid systems, i.e., use a primary domain of e-cores only for better power consumption or a primary domain of p-cores only for better performance.

Some test results on an Dell Precision 5480 with 13th Gen Intel(R) Core(TM) i7-13800H hybrid cores:

```
                      +-----+-----+------+-----+----------+
                      | min | max | avg  |       |        |
                      | fps | fps | fps  | stdev | power  |
    +-----------------+-----+-----+------+-------+--------+
    | EEVDF           | 28  | 34  | 31.0 |  1.73 |  3.5W  |
    | bpfland-p-cores | 33  | 34  | 33.5 |  0.29 |  3.5W  |
    | bpfland-e-cores | 25  | 26  | 25.5 |  0.29 |  2.2W  |
    +-----------------+-----+-----+------+-------+--------+
```

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: GPL-2.0 */
 /*
- * Copyright (c) 2024 Andrea Righi <righi.andrea@gmail.com>
+ * Copyright (c) 2024 Andrea Righi <andrea.righi@linux.dev>
  */
 #include <scx/common.bpf.h>
 #include "intf.h"
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -107,6 +107,12 @@ volatile u64 nr_running, nr_waiting, nr_interactive, nr_online_cpus;
 UEI_DEFINE(uei);
 
 /*
+ * Mask of CPUs that the scheduler can use, until the system becomes saturated,
+ * at which point tasks may overflow to other available CPUs.
+ */
+private(BPFLAND) struct bpf_cpumask __kptr *allowed_cpumask;
+
+/*
  * Mask of offline CPUs, used to properly support CPU hotplugging.
  */
 private(BPFLAND) struct bpf_cpumask __kptr *offline_cpumask;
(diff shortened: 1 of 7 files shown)


arighi:
Thanks for the review @multics69 , yeah the enable_cpu() stuff is pretty cool, something that I learned from @Byte-Lab :)

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -914,6 +959,53 @@ s32 get_nr_online_cpus(void)
 	return cpus;
 }
 
+static int init_allowed_cpus(void)
+{
+	struct bpf_cpumask *mask;
+	int err = 0;
+
+	bpf_rcu_read_lock();
```
htejun, line 967: Hmm... can you please explain the locking a bit? I don't understand what rcu read locks are protecting.

htejun (approved):

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Thanks @htejun, the rcu read lock is definitely not needed, the goal was just to make the verifier happy.

I've pushed a new change removing the rcu locking both from the `allowed_cpumask` initialization and, more important, from `pick_idle_cpu()`. We still need it to call `bpf_cpumask_set_cpu()`, but that is just affecting the initialization, so basically there's not extra locking involved at runtime, which is nice.

scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -46,6 +48,69 @@ use scx_utils::UserExitInfo;
...
+            return false;
+        }
+        let idx = cpu / 64;
+        let bit = cpu % 64;
+        self.mask
+            .get(idx)
+            .map_or(false, |&val| val & (1 << bit) != 0)
+    }
+
+    pub fn from_str(hex_str: &str) -> Result<Self, std::num::ParseIntError> {
```
hodgesds, line 76: in the future it might be nice to accept a `--cpu-list` similar to how [`taskset`](https://man7.org/linux/man-pages/man1/taskset.1.html) does it.

arighi: @hodgesds good idea. I'll add that as a future change, it can be done all in Rust, so it should be trivial, thanks for the suggestion!

PR #496, opened 2024-08-14
scx_stat: Initial commit

htejun:
scx_stat is a statistics reporting library with the following goals:

- Minimal boilerplate code. Statistics are defined as a regular rust struct which can contain integers, floats, strings, structs, and array or BTreeMap of them. The struct and each field can be annotated using the "stat" attribute. The "Stat" derive macro generates the necessary metadata.

- On-demand reporting. ScxStatServer implements a UNIX domain socket server which communicates using a simple json protocol. Stat structs can be added to the server with closures to output them. There can be any number of readers. ScxStatServer can be queried for the metadata of all stat structs that it may report so that a generic tool can be written to pipe the reported statistics to other frameworks (e.g. openmetrics).

- The interface protocol is pretty simple and it isn't difficult to interact directly in json. ScxStatClient is provided to further simplify client implementation.

- See rust/scx_stat/examples/{server,client}.rs for usage.


rust/scx_stat/src/stat.rs
```diff
@@ -0,0 +1,262 @@
...
+                    return Ok(dict);
+                }
+            }
+        }
+        Ok(Self::Datum(kind))
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum ScxStatAttr {
```
hodgesds, line 162: what are your thoughts on having a set of additional tags/labels (ex [prometheus](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels)). I guess the server/client implementations could add any additional tags downstream.

hodgesds (approved):
LGTM

jfernandez:
This framework should work for exporting scx metrics to our backend. Once this lands, I can write a client for our metrics backend and provide more feedback from that experience.

PR #505, opened 2024-08-16
scx_stats: Add support for no-value user attributes and a bunch of ot…

htejun:
…her changes

- Allow no-value user attributes which are automatically assigned "true" when specified.

- Make "top" attribute string "true" instead of bool true for consistency. Testing for existence is always enough for value-less attributes.

- Don't drop leading "_" from user attribute names when storing in dicts. Dropping makes things more confusing.

- Add "_om_skip" to scx_layered fields which don't jive well with OM. scxstats_to_openmetrics.py is updated accordignly and no longer generates warnings on those fields.

- Examples and README updated accordingly.

### DIFF
--- a/rust/scx_stats/README.md
+++ b/rust/scx_stats/README.md
@@ -64,7 +64,23 @@ defined:
 In addition, arbitrary user attributes which start with "_" can be added to
 both structs and fields. They are collected into the "user" dict of the
 containing struct or field. When the value of such user attribute is not
-specified, the string "true" is assigned by default. See the examples.
+specified, the string "true" is assigned by default. For example,
+[scripts/scxstats_to_openmetrics.py](scripts/scxstats_to_openmetrics.py)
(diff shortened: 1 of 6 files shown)


rust/scx_stats/README.md
```diff
@@ -53,8 +48,23 @@ struct ClusterStats {
...
+*struct-only attributes*
+
+- top: Marks the top-level statistics struct which is reported by default.
+  Used by generic tools to find the starting point when processing the
+  metadata.
+
+In addition, arbitrary user attributes which start with "_" can be added to
+both structs and fields. They are collected into the "user" dict of the
+containing struct or field. When the value of such user attribute is not
+specified, the string "true" is assigned by default. See the examples.
```
hodgesds, line 67: might be worth adding something about `_om_skip`?

hodgesds (approved):

PR #508, opened 2024-08-19
scx_lavd: fix a potential watchdog timeout error at multi-NUMA/CCX platforms

multics69:
Before this fix, the scheduler enqueues a task to one of enqueuing CPU's DSQ. That could cause a watchdog timeout error. For example, suppose that a task pinned to NUMA domain 1 is enqueued to a DSQ on NUMA domain 0. The task on the DSQ in a wrong NUMA domain will never be exeuted until a CPU on the correct NUMA domain (NUMA domain 1 in this example) steals the task by any chance.

The fix always enqueues a task to a correct DSQ by referring task's thread information (`task->threadinfo.cpu`).

Besides the major change, it also includes minor code cleanup around ops.enqueue().

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -1285,6 +1285,181 @@ static void update_stat_for_quiescent(struct task_struct *p,
 	cpuc->load_run_time_ns -= clamp_time_slice_ns(taskc->run_time_ns);
 }
 
+static bool could_run_on_prev(struct task_struct *p, s32 prev_cpu,
+			      struct bpf_cpumask *a_cpumask,
+			      struct bpf_cpumask *o_cpumask)
+{
+	bool ret;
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -362,6 +362,14 @@ static struct task_ctx *get_task_ctx(struct task_struct *p)
 	return taskc;
 }
 
+static s32 get_task_cpu_id(struct task_struct *p)
```
htejun, line 365: You should be able to use `scx_bpf_task_cpu()` instead.

htejun (approved):

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Ah, got it. I will change it.

PR #509, opened 2024-08-19
scx_bpfland: topology awareness

arighi:
Introduce some concepts of topology awareness to scx_bpfland:
 - L2 / L3 cache awareness
 - cpu frequency awreness (via primary scheduling domain)

These changes enable better utilization of hybrid architectures (P-cores / E-cores) and generally improve performance by keeping tasks running on the same cores and caches, thereby enhancing the reuse of their working set.

### DIFF
--- a/scheds/rust/scx_bpfland/src/main.rs
+++ b/scheds/rust/scx_bpfland/src/main.rs
@@ -109,9 +109,107 @@ impl CpuMask {
     }
 }
 
+fn get_primary_cpus(powersave: bool) -> std::io::Result<Vec<usize>> {
+    let cpu_base_path = "/sys/devices/system/cpu/";
+    let mut cpu_freqs = Vec::new();
+
+    // Iterate over each CPU directory and collect CPU ID and its max frequency.
+    for entry in std::fs::read_dir(cpu_base_path)? {
+        let entry = match entry {
+            Ok(e) => e,
+            Err(_) => continue,  // Skip if there's an error
+        };
+
+        let path = entry.path();
+        if path.is_dir() && path.file_name().unwrap().to_str().unwrap_or("").starts_with("cpu") {
+            if let Some(cpu_id_str) = path.file_name().unwrap().to_str().unwrap_or("").strip_prefix("cpu") {
+                if let Ok(cpu_id) = cpu_id_str.parse::<usize>() {
+                    let max_freq_path = path.join("cpufreq/cpuinfo_max_freq");
+                    if max_freq_path.exists() {
+                        if let Ok(max_freq) = std::fs::read_to_string(&max_freq_path) {
+                            if let Ok(freq) = max_freq.trim().parse::<u64>() {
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -329,18 +353,155 @@ impl<'a> Scheduler<'a> {
...
+        if !path.exists() {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::NotFound,
+                "file not found",
+            ));
+        }
+        std::fs::read_to_string(path).map(|content| content.trim().to_string())
+    }
+
+    fn read_cpu_ids(sysfs_path: &str) -> Result<Vec<usize>, std::io::Error> {
```
htejun, line 477: Can't it use `scx_utils::topology`? Is something missing there?

scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -109,9 +109,107 @@ impl CpuMask {
...
+        let entry = match entry {
+            Ok(e) => e,
+            Err(_) => continue,  // Skip if there's an error
+        };
+
+        let path = entry.path();
+        if path.is_dir() && path.file_name().unwrap().to_str().unwrap_or("").starts_with("cpu") {
+            if let Some(cpu_id_str) = path.file_name().unwrap().to_str().unwrap_or("").strip_prefix("cpu") {
+                if let Ok(cpu_id) = cpu_id_str.parse::<usize>() {
+                    let max_freq_path = path.join("cpufreq/cpuinfo_max_freq");
```
htejun, line 127: Ditto, I *think* this is already available in the topology.

htejun (approved):

scheds/rust/scx_bpfland/src/main.rs
arighi: yep, I'm planning to use `scx_utils::topology` here, I just need to repeat some stress tests with cpu hotplugging, because I was able to trigger a sysfs read failure (or similar) with it. I'm going to reproduce, fix, and then rewrite this code to use topology.

arighi:
@htejun just added a couple of commits on top to use `scx_utils::Topology`, instead of re-implementing the same methods in bpfland. I repeated my CPU hotplugging stress tests and I couldn't trigger any error (probably the issue was in my old code, not in Topology).

@hodgesds FYI, I added two new methods to struct Cpu: `l2_id()` and `l3_id()`, to get respectively the L2 and L3 cache ID associated to the CPU. It shouldn't break any of your stuff, but maybe take a quick look and see if it makes sense. Thanks!

htejun (approved):

ptr1337 (approved):
Tested on 9950. Works fine, no regressions found.

sirlucjan (approved):
No regressions during heavy load

hodgesds (approved):

PR #510, opened 2024-08-19
scx_layered: Use topology for core selection

hodgesds:
Currently the core selection logic in scx_layered uses the first available core in the bitmask. This is suboptimal when the scheduler is configured with specific NUMA/LLC restrictions and with multiple layers. This refactors the core selection to use a preferred order of cores based on the layer index.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -896,12 +896,23 @@ impl CpuPool {
         }
     }
 
-    fn alloc_cpus<'a>(&'a mut self, cpus: &BitVec) -> Option<&'a BitVec> {
-        let cores = self.cpus_to_cores(cpus).ok()?;
-        for core in cores.iter_ones() {
-            self.available_cores.set(core, false);
-            self.update_fallback_cpu();
-            return Some(&self.core_cpus[core]);
+    fn alloc_cpus<'a>(&'a mut self, layer: &Layer) -> Option<&'a BitVec> {
+        let available_cpus = self.available_cpus_in_mask(&layer.allowed_cpus);
+        let available_cores = self.cpus_to_cores(&available_cpus).ok()?;
+
+        for alloc_core in layer.core_alloc_order() {
+            match available_cores.get(*alloc_core) {
+                Some(bit) => {
+                    if *bit {
+                        self.available_cores.set(*alloc_core, false);
+                        self.update_fallback_cpu();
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/main.rs
```diff
@@ -1030,16 +1048,43 @@ impl Layer {
             }
         }
 
+        let mut nodes = topo.nodes().iter().collect::<Vec<_>>().clone();
+        let num_nodes = nodes.len();
+        nodes.rotate_left(if idx <= num_nodes { idx } else { idx % num_nodes });
```
hodgesds, line 1053: This is the main logic, it makes use of [`rotate_left`](https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.rotate_left), which annoying will panic if the rotate size is too large. To get around that it mods by the number of nodes, llcs, or cores if the number of layers is larger than any of those values.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -875,12 +875,23 @@ impl CpuPool {
...
-        for core in cores.iter_ones() {
-            self.available_cores.set(core, false);
-            self.update_fallback_cpu();
-            return Some(&self.core_cpus[core]);
+    fn alloc_cpus<'a>(&'a mut self, layer: &Layer) -> Option<&'a BitVec> {
+        let available_cpus = self.available_cpus_in_mask(&layer.allowed_cpus);
+        let available_cores = self.cpus_to_cores(&available_cpus).ok()?;
+
+        for alloc_core in layer.core_alloc_order() {
+            match available_cores.get(*alloc_core) {
```
hodgesds, line 883: this goes from the core id to the bitmask bit for checking if the core is allowed to be used

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1030,16 +1048,57 @@ impl Layer {
             }
         }
 
+        let mut nodes = topo.nodes().iter().collect::<Vec<_>>().clone();
+        let num_nodes = nodes.len();
+        let is_left = idx % 2 == 0;
+        if is_left {
+            nodes.rotate_left(if idx <= num_nodes { idx } else { idx % num_nodes });
```
htejun, line 1055: Maybe sth like `let rot_by |idx, len| if idx <= len { idx } else { idx % len };` would help? Hmm... isn't this the same as `idx % num_nodes`?

htejun (approved):

PR #513, opened 2024-08-19
ci: Fix veristat pull request workflow

hodgesds:
See the [failure](https://github.com/sched-ext/scx/actions/runs/10389671253) which needs to have an action defined.


meson-scripts/veristat
```diff
@@ -21,7 +21,7 @@ if [ -n "${SCHED}" ]; then
   if [ -n "${KERNEL}" ]; then
     timeout --preserve-status ${GUEST_TIMEOUT} \
       vng --force-9p -v -r ${KERNEL} -- \
-        veristat ${BPF_PATH}
+        ./veristat ${BPF_PATH}
```
hodgesds, line 24: This is to use the downloaded `veristat` inside `vng`. I think this will work?

PR #514, opened 2024-08-19
scx_stats, scx_layered: Implement independent stats client sessions

htejun:
Currently, scx_layered generates stats at a fixed interval and every stats client reads out the content of the latest update. This means that a client whose poll interval is longer than the scheduler's would see sampled stats (e.g. the average and nr_cpus ranges wouldn't represent the whole period) and a client with shorter poll interval may see the same outputs repeated.

This PR implements `ScxStatsServer::add_stats_ops()` which enables the user to specify `open()` and `close()` closures. The `open()` closure creates a `read()` closure per client session which can carry mutable states. In addition, it provides multiplexed communication channels between the clients and the server user which can be useful when implementing stateful stats generation.

`scx_layered` is updated to make use of the new feature. After this PR, every client always gets accurate stats, including the time averages and period nr_layer_cpus min/max tracking, independent of each other.

### DIFF
--- a/rust/scx_stats/src/server.rs
+++ b/rust/scx_stats/src/server.rs
@@ -2,6 +2,7 @@ use crate::{Meta, ScxStatsMeta};
 use anyhow::{anyhow, Context, Result};
 use log::warn;
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
 use std::collections::BTreeMap;
 use std::io::{BufRead, BufReader, Write};
 use std::os::unix::net::{UnixListener, UnixStream};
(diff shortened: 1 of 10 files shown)


rust/scx_stats/src/server.rs
```diff
@@ -138,45 +292,163 @@ impl ScxStatsServerInner {
...
+            }
+            match stream {
+                Ok(stream) => {
+                    let data = self.data.clone();
+                    let exit = self.exit.clone();
+
+                    let (req_pair, res_pair) = ChannelPair::<Req, Res>::bidi();
+                    match add_req.send(res_pair) {
+                        Ok(()) => debug!("sent new channel to proxy"),
+                        Err(e) => warn!("ScxStatsServer::proxy() failed ({})", &e),
```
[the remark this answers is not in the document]
htejun: Yeah, some statistics can be useful. I'll put that on the todo list.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1448,42 +1462,38 @@ impl<'a, 'b> Scheduler<'a, 'b> {
 
         // Other stuff.
         let proc_reader = procfs::ProcReader::new();
-        let sys_stats = Arc::new(Mutex::new(SysStats::default()));
 
-        let mut sched = Self {
-            struct_ops: None,
+        // XXX If we try to refresh the cpumasks here before attaching, we
```
hodgesds, line 1466: good call on this, I think I've seen this while testing some of the topology awareness

scheds/rust/scx_layered/src/stats.rs
```diff
@@ -429,32 +426,101 @@ impl SysStats {
...
+                    retry_cnt += 1;
+                    sleep(Duration::from_secs(1));
+                    continue;
+                }
+                _ => Err(e)?,
+            },
+        };
+        retry_cnt = 0;
+
+        while !shutdown.load(Ordering::Relaxed) {
```
hodgesds, line 506: is this inner one needed?

scheds/rust/scx_layered/src/stats.rs
```diff
@@ -429,32 +426,101 @@ impl SysStats {
...
-        if sst.at != last_at {
+        let mut client = match ScxStatsClient::new().connect() {
+            Ok(v) => v,
+            Err(e) => match e.downcast_ref::<std::io::Error>() {
+                Some(ioe) if ioe.kind() == std::io::ErrorKind::ConnectionRefused => {
+                    if retry_cnt == 1 {
+                        info!("Stats server not avaliable, retrying...");
+                    }
+                    retry_cnt += 1;
+                    sleep(Duration::from_secs(1));
```
hodgesds, line 498: maybe give up after a max number of retries?

scheds/rust/scx_layered/src/stats.rs
htejun: I think so. Otherwise, the inner loop will never exit on ctrl-c.

scheds/rust/scx_layered/src/stats.rs
htejun: I don't know. After how many? I think the mental model is like following a log file. Wouldn't it make sense to just let it sit there until told to exit?

hodgesds (approved):
seems legit. I think the unbounded channel is a little scary from a resource leaking perspective, but imo is easier to handle/debug something that could block.

htejun:
> seems legit. I think the unbounded channel is a little scary from a resource leaking perspective, but imo is easier to handle/debug something that could block.

We should be able to switch to bounded channels with 0 size without noticeable behavior changes. There's only supposed to be one message in flight in each connection pair at any given time and all the operations are fully synchronous (one send followed by one receive or vice-versa). Yeah, using bounded is probably better here. Will do that later.

PR #515, opened 2024-08-19
ci: fix veristat for PRs

hodgesds:
Make sure veristat is available for CI for PRs.


hodgesds:
I think this should pass now after the merge runs and generates the veristat diff files

PR #516, opened 2024-08-19
scx_stats/scripts/scxstats_to_openmetrics: Retry connection

htejun:
It now retries until told to exit. This is a bit easier to use and matches `scx_layered --monitor`.


hodgesds (approved):
LGTM

PR #518, opened 2024-08-19
scx_layered: Add `--run-example` and enable CI testing

htejun:


hodgesds (approved):
LGTM

hodgesds:
I was thinking of adding a layer with the UID of the current user as a follow up. Now that args can be passed we can probably do some permutations of different scheduler configs as well

PR #522, opened 2024-08-20
scx_bpfland: use scx_utils::Cpumask

arighi:
Rely on `scx_utils::Cpumask` instead of re-implementing a custom `CpuMask` struct.

This is just some code refactoring, no functional change.

### DIFF
--- a/rust/scx_utils/src/cpumask.rs
+++ b/rust/scx_utils/src/cpumask.rs
@@ -224,6 +224,53 @@ impl fmt::Display for Cpumask {
     }
 }
 
+impl Cpumask {
+    fn fmt_with(&self, f: &mut fmt::Formatter<'_>, case: char) -> fmt::Result {
+        let mut masks: Vec<u32> = self
+            .as_raw_slice()
+            .iter()
+            .map(|x| [*x as u32, (x >> 32) as u32])
+            .flatten()
+            .collect();
+
+        // Throw out possible stray from u64 -> u32.
+        masks.truncate((self.nr_cpus + 31) / 32);
+
+        // Print the highest 32bit. Trim digits beyond nr_cpus.
+        let width = match (self.nr_cpus + 3) / 4 % 8 {
+            0 => 8,
+            v => v,
+        };
+        match case {
+            'x' => write!(f, "{:0width$x}", masks.pop().unwrap(), width = width)?,
+            'X' => write!(f, "{:0width$X}", masks.pop().unwrap(), width = width)?,
+            _ => unreachable!(),
+        }
+
+        // The rest in descending order.
+        for submask in masks.iter().rev() {
+            match case {
+                'x' => write!(f, " {:08x}", submask)?,
+                'X' => write!(f, " {:08X}", submask)?,
+                _ => unreachable!(),
+            }
+        }
+        Ok(())
+    }
+}
+
+impl fmt::LowerHex for Cpumask {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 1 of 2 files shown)


htejun (approved):

rust/scx_utils/src/cpumask.rs
```diff
@@ -224,6 +224,40 @@ impl fmt::Display for Cpumask {
     }
 }
 
+impl fmt::LowerHex for Cpumask {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let slice = self.as_raw_slice();
+
+        // Treat the last chunk separately, because it may not require the 16 hex digits padding.
+        if let Some(last) = slice.last() {
+            write!(f, "{:x}", last)?;
```
htejun, line 233: One problem with this is that if the leading hexes are zero, they'll be omitted even for possible CPUs leading to unaligned outputs. How about something like the following?
```
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let mut masks: Vec<u32> = self
            .as_raw_slice()
            .iter()                                
            .map(|x| [*x as u32, (x >> 32) as u32]) 
            .flatten()                             
            .collect();                            

        // Throw out possible stray from u64 -> u32.
        masks.truncate((*NR_CPU_IDS + 31) / 32);

        // Print the highest 32bit. Trim digits beyond NR_CPU_IDS.
        let width = match (*NR_CPU_IDS + 3) / 4 % 8 {
            0 => 8,                                
            v => v,                                
        };                                         
        write!(f, "{:0width$x}", masks.pop().unwrap(), width = width)?;

        // The rest in descending order.
        for submask in masks.iter().rev() {
            write!(f, " {:08x}", submask)?;
        }                                          
        Ok(())                                     
    }                                              
```
This also cuts it on 32bit boundaries which is close to how kernel prints out cpumasks and easier to read.

htejun (approved):

rust/scx_utils/src/cpumask.rs
arighi: > One problem with this is that if the leading hexes are zero, they'll be omitted even for possible CPUs leading to unaligned outputs. How about something like the following?

Oh I see, even if it's equivalent, it'd be nicer to print all the bits for all the possible CPUs.

> 
> ```
>     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
>         let mut masks: Vec<u32> = self
>             .as_raw_slice()
>             .iter()                                
>             .map(|x| [*x as u32, (x >> 32) as u32]) 
>             .flatten()                             
>             .collect();                            
> 
>         // Throw out possible stray from u64 -> u32.
>         masks.truncate((*NR_CPU_IDS + 31) / 32);
> 
>         // Print the highest 32bit. Trim digits beyond NR_CPU_IDS.
>         let width = match (*NR_CPU_IDS + 3) / 4 % 8 {
>             0 => 8,                                
>             v => v,                                
>         };                                         
>         write!(f, "{:0width$x}", masks.pop().unwrap(), width = width)?;
> 
>         // The rest in descending order.
>         for submask in masks.iter().rev() {
>             write!(f, " {:08x}", submask)?;
>         }                                          
>         Ok(())                                     
>     }                                              
> ```
> 
> This also cuts it on 32bit boundaries which is close to how kernel prints out cpumasks and easier to read.

I like this, also with the extra space between each 32-bit chunk to make the cpumask more readable. I'll apply this change and do some tests, thanks!

arighi:
Added @htejun's fmt() and refactored the code a bit to avoid code duplication.

htejun (approved):

Issue #521, opened 2024-08-20
Huge log files


Kawanaao:
Hi all!

I would like to suggest introducing the LOG_LEVEL variable or simply LOG=false into the services (for openrc, I don’t know if this is relevant for systemd), the reason is that scx produces too much information as a service, 140MiB of logs are collected in one day
logrotate, unfortunately, will be redundant due to the need to see errors as well

sirlucjan:
In systemd, we limited the size of the logo to 50M [1]. 
Maybe @nycko123  will know what to do with openrc. 

1. https://github.com/sched-ext/scx/blob/main/services/systemd/journald%40sched-ext.conf#L3

Kawanaao:
> In systemd, we limited the size of the logo to 50M [1]. Maybe @nycko123 will know what to do with openrc.
> 
> 1. https://github.com/sched-ext/scx/blob/main/services/systemd/journald%40sched-ext.conf#L3

In openrc this is done via logrotate, but in any case the problem is more likely that there are too many entries in the logs, I would like to have fewer of them or disable them, since they overwrite all the most important things

sirlucjan:
rusty, lavd and bpfland have scx_stats support (in -git version). Could you check?

ptr1337:
There are no log files anymore, this gets handled via scx_stats and of the schedulers --monitor function. I think we can close that issue.

PR #523, opened 2024-08-20
openrc: Add logrotate support for openrc systems

Kawanaao:
Optional logrotate support for openrc based systems that have logrotate

### DIFF
--- a/services/openrc/meson.build
+++ b/services/openrc/meson.build
@@ -4,6 +4,9 @@ install_data('scx.initrd', install_dir: '/etc/init.d', rename: 'scx')
 # Install the 'scx' file to the '/etc/default' directory
 install_data('../scx', install_dir: '/etc/default')
 
+# Install the 'scx.logrotate' file to /etc/logrotate.d/scx
+install_data('scx.logrotate', install_dir: '/etc/logrotate.d', rename: 'scx')
+
 # Symlinking /etc/default/scx to /etc/conf.d/scx
 install_symlink(
       'scx',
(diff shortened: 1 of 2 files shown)


Kawanaao:
Fully tested on my system, the configuration files perform the task, the only question is the size of the log files, if according to #521 the log size within the normal range will be lower than 50 MB

Kawanaao:
Fully tested on my system, the configuration files perform the task, the only question is the size of the log files, if according to #521 the log size within the normal range will be lower than 50 MB

```
reading config file /etc/logrotate.d/scx
acquired lock on state file /var/lib/misc/logrotate.status
Reading state from file: /var/lib/misc/logrotate.status
Allocating hash table for state file, size 64 entries
Creating new state
Creating new state

Handling 1 logs

rotating pattern: /var/log/scx/*.log 1048576 bytes empty log files are not rotated, only log files >= 4194304 bytes are rotated, no old logs will be kept
considering log /var/log/scx/scx.log
  Now: 2024-08-20 19:36
  Last rotated at 2024-08-20 19:35
  log needs rotating
rotating log /var/log/scx/scx.log, log->rotateCount is 0
dateext suffix '-20240820'
glob pattern '-[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]'
renaming /var/log/scx/scx.log.1 to /var/log/scx/scx.log.2 (rotatecount 1, logstart 1, i 1), 
old log /var/log/scx/scx.log.1 does not exist
log /var/log/scx/scx.log.2 doesn't exist -- won't try to dispose of it
skip copying /var/log/scx/scx.log to /var/log/scx/scx.log.1
truncating /var/log/scx/scx.log
```

sirlucjan (approved):

htejun (approved):

PR #530, opened 2024-08-21
scx_utils::topology: Use lazy_static instead of LazyLock

htejun:
LazyLock is stable but has become so only very recently and can trigger build errors on not-too-old stable rustc's which are still in wide use. Let's use lazy_static instead for now.


arighi (approved):
LGTM and it fixes the build error mentioned in #527

PR #534, opened 2024-08-21
scx_layered: Drop SCX_OPS_ENQ_LAST

htejun:
The meaning of SCX_OPS_ENQ_LAST will change with future kernel updates and enqueueing on local DSQ will no longer be sufficient to avoid stalls. No reason to do it anyway. Just drop it.


hodgesds (approved):
LGTM

PR #535, opened 2024-08-22
scx_bpfland: better time slice control

arighi:
Enable SCX_OPS_ENQ_LAST, so that sched_ext core won't automatically re-schedule the task on the same CPU, implicitly assigning a time slice of SCX_SLICE_DFL, if no other task wants to run on that CPU.

Instead, explicitly replenish the task's time slice from ops.dispatch() if the task still wants to run and no other task is selected.

Moreover, instead of determining the task time slice in ops.enqueue(), refresh the time slice immediately before the task is started on its assigned CPU in ops.running().

This allows to use a more precise time slice, adjusted based on the actual amount of tasks that are currently waiting to be scheduled.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -413,7 +413,6 @@ static u64 cpu_to_dsq(s32 cpu)
 static int dispatch_direct_cpu(struct task_struct *p, s32 cpu, u64 enq_flags)
 {
 	struct bpf_cpumask *offline;
-	u64 slice = task_slice(p);
 	u64 vtime = task_vtime(p);
 	u64 dsq_id = cpu_to_dsq(cpu);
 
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -1282,5 +1296,6 @@ SCX_OPS_DEFINE(bpfland_ops,
 	       .init_task		= (void *)bpfland_init_task,
 	       .init			= (void *)bpfland_init,
 	       .exit			= (void *)bpfland_exit,
+	       .flags			= SCX_OPS_ENQ_LAST,
```
htejun, line 1299: Does setting this flag make any behavior difference? If the expired slice is always replenished in `.dispatch()` the auto re-enqueue is actually the behavior that would match new kernel behavior (as tasks that keep running won't go through `enqueue` at all).

htejun (approved):

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Actually I'm not noticing any behavior change with / without the flag. IIUC the auto re-enqueue if triggered if I don't consume any other task in `ops.dispatch()` and the prev->scx.slice isn't replenished. But if I replenish the time slice I won't trigger the auto re-enqueue anyway, so basically I don't have to set the `SCX_OPS_ENQ_LAST` flag.

htejun: Yeah, the only difference would be whether you get `.enqueue` call if you forget to set the slice. Given that that's not supposed to happen and the kernel will be changed to warn on that, leaving it off probably makes more sense.

arighi: ack, flag dropped, commit updated, will do more testing, but it seems to work as expected so far.

PR #539, opened 2024-08-23
scx_rusty: Convert to scx_stats

htejun:
- Convert scx_rusty to use scx_stats.
- Some updates to scx_stats in the process.
- scx_layered --monitor behavior updated so that both scx_rusty and scx_layered can behave consistently.


hodgesds (approved):
lgtm, I think `--stats` is far more clear in what it does.

PR #543, opened 2024-08-23
scx_bpfland, scx_lavd: Improve help info a bit

anh0516:
I wanted to do some work improving the end-user documentation, as I noticed some random typos and other things that were worded a little unclearly or just not explained at all.

I made some minor grammatical changes to scx_bpfland, and I added explanations to scx_lavd that were poorly explained or missing entirely; for example, that the power mode options control the defaults for the four scheduling controls wasn't explained anywhere, and I only learned that by paging through the Rust code, or using the correct terminology "SMT sibling" rather than "logical core."

In a perfect world, the documentation for every scheduler should be formatted exactly the same, and the information useful to end users shouldn't just be embedded in the source code. While an experienced sysadmin might not mind, it is a barrier of entry for, say, a gamer who is interested in tuning scx to their needs and has installed prebuilt schedulers from their distro's repos (I think only CachyOS packages them right now, but that will likely change in the future.) This is just a small start.

I'll probably do some more work when I have the time.

### DIFF
--- a/scheds/rust/scx_bpfland/src/main.rs
+++ b/scheds/rust/scx_bpfland/src/main.rs
@@ -165,8 +165,8 @@ fn parse_cpumask(cpu_str: &str) -> Result<Cpumask, anyhow::Error> {
 
 /// scx_bpfland: a vruntime-based sched_ext scheduler that prioritizes interactive workloads.
 ///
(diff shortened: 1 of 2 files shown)


arighi (approved):
Nice cleanups, thanks for this

htejun (approved):

PR #549, opened 2024-08-24
scx_utils: Add retryable errors

hodgesds:
Automatically (try to) create the socket dir if it doesn't exist. Sometimes the server will fail with:
```
$ sudo ./target/release/scx_layered --run-example --stats 5
thread '<unnamed>' panicked at src/main.rs:1960:74:
called `Result::unwrap()` on an `Err` value: No such file or directory (os error 2)
21:52:54 [INFO] CPUs: online/possible=16/16 nr_cores=8
21:52:54 [INFO] configuring node 0, LLCs 2
21:52:54 [INFO] configuring llc 0 for node 0
21:52:54 [INFO] configuring llc 1 for node 0
21:52:55 [INFO] Layered Scheduler Attached. Run `scx_layered --monitor` for metrics.
^CEXIT: unregistered from user space
````

### DIFF
--- a/rust/scx_utils/src/misc.rs
+++ b/rust/scx_utils/src/misc.rs
@@ -15,11 +15,17 @@ where
     T: for<'a> Deserialize<'a>,
 {
     let mut retry_cnt: u32 = 0;
+
+    const RETRYABLE_ERRORS: [std::io::ErrorKind; 2] = [
+        std::io::ErrorKind::NotFound,
+        std::io::ErrorKind::ConnectionRefused,
+    ];
+
     while !should_exit() {
         let mut client = match StatsClient::new().connect() {
             Ok(v) => v,
             Err(e) => match e.downcast_ref::<std::io::Error>() {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by htejun** [comment]
So, in `launch()`, there's this:
```rust
        if let Some(dir) = path.parent() {
            std::fs::create_dir_all(dir).with_context(|| format!("creating {:?}", dir))?;
        }
```
Shouldn't this be enough? It's doing `create_dir_all()` on the parent. If you can reproduce the problem, can you find out why that's not working?

hodgesds:
I think I see the problem in `scx_layered`, during startup the [monitor](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/main.rs#L1974) is called before the `launch()` is called during the scheduler init.

hodgesds:
Updated by adding more retryable errors. In the case where the server hadn't started the path wouldn't be found so the stats server connection would never happen.

PR #553, opened 2024-08-25
scx_lavd: Switch introspection to use scx_stats

htejun:
This allows introspection to be done on-demand while the scheduler is running so that we can reduce log noise while maintaining debuggability. It should be straightforward to add other introspection criteria and different types of stats too.


multics69 (approved):
Thanks @htejun ! LGTM

PR #555, opened 2024-08-25
scx_stats: Shorten exported names and add prelude module

htejun:
Let's make it a bit easier to use:

- Shorten exported names by changing the prefix from ScxStats to Stats. This should be distinctive enough and more inline with how most libraries name their exports.

- Importing the right set of traits can be tricky. Introduce prelude module so that importing is a bit less painful.


hodgesds (approved):
Makes sense

PR #556, opened 2024-08-25
scx_bpfland: Use scx_stats

htejun:
Make scx_bpfland use scx_stats and make the reported values interval deltas.


htejun:
This PR is stacked on top of htejun/scx_stats. Please take a look at diff against that branch instead of main.

arighi (approved):
I was planning to start looking at scx_stats for bpfland, but you already did everything, awesome! :)

Tested and LGTM, thanks!

Issue #546, opened 2024-08-24
Create Cargo workspace for Rust schedulers


vimproved:
It would be helpful for packaging purposes if there was a top-level Cargo workspace that contained all of the Rust schedulers as members, since it would allow compiling with only one invocation of Cargo instead of many individual ones. Additionally, it would enable all schedulers to share a Cargo.lock, which would have multiple benefits.

htejun:
Thanks for the suggestion. This has been an annoying problem for quite a while and it's so much better with cargo workspace.

PR #559, opened 2024-08-25
build: Use workspace to group rust sub-projects

htejun:
meson build script was building each rust sub-project under rust/ and scheds/rust/ separately. This means that each rust project is built independently which leads to a couple problems - 1. There are a lot of shared dependencies but they have to be built over and over again for each proejct. 2. Concurrency management becomes sad - we either have to unleash multiple cargo builds at the same time possibly thrashing the system or build one by one.

We've been trying to solve this from meson side in vain. Thankfully, in issue #546, @vimproved suggested using cargo workspace which makes the sub-projects share the same target directory and built together by the same cargo instance while still allowing each project to behave independently for development and publishing purposes.

Make the following changes:

- Create two cargo workspaces - one under rust/, the other under scheds/rust/. Each contains all rust projects underneath it.

- Don't let meson descend into rust/. These are libraries used by the rust schedulers. No need to build them from meson. Cargo will build them as needed.

- Change the rust_scheds build target to invoke `cargo build` in scheds/rust/ and let cargo do its thing.

- Remove per-scheduler meson.build files and instead generate custom_targets in scheds/rust/meson.build which invokes `cargo build -p $SCHED`.

- This changes rust binary directory. Update README and meson-scripts/install_rust_user_scheds accordingly.

- Remove per-scheduler Cargo.lock as scheds/rust/Cargo.lock is shared by all schedulers now.

- Unify .gitignore handling.

The followings are build times on Ryzen 3975W:

Before:
  ________________________________________________________
  Executed in  165.93 secs    fish           external
     usr time   40.55 mins    2.71 millis   40.55 mins
     sys time    3.34 mins   36.40 millis    3.34 mins

After:
  ________________________________________________________
  Executed in   36.04 secs    fish           external
     usr time  336.42 secs    0.00 millis  336.42 secs
     sys time   36.65 secs   43.95 millis   36.61 secs

Wallclock time is reduced 5x and CPU time 7x.


arighi (approved):
This is really nice! Finally we figured out a sane solution for the massive parallel builds.

Tested & LGTM

multics69 (approved):
Thank you @htejun ! It is really nice.

sirlucjan (approved):

ptr1337 (approved):
Tested with the "scx-scheds-git" PKGBUILD.
Test has been done on the 9950X
With this patch:
``` 
==> Finished making: scx-scheds-git 1.0.3.r92.ga4fb550-1 (So 25 Aug 2024 17:31:16 CEST)

________________________________________________________
Executed in   18.88 secs    fish           external
   usr time  125.54 secs    0.00 micros  125.54 secs
   sys time   13.28 secs  305.00 micros   13.28 secs
``` 


without:
```
==> Finished making: scx-scheds-git 1.0.3.r92.ga4fb550-1 (So 25 Aug 2024 17:34:20 CEST)

________________________________________________________
Executed in  170.71 secs    fish           external
   usr time   49.48 mins  201.00 micros   49.48 mins
   sys time    1.64 mins   84.00 micros    1.64 mins

```

Issue #560, opened 2024-08-25
scx_bpfland: Support ignoring CPUFreq EPP for systems that don't support it


anh0516:
On CPUs that are too old to support pstate EPP, or on kernels built without CPUFreq support, scx_bpfland repeatedly tries to read the energy_performance_preference, but it doesn't exist.

Here's an strace snippet:
```
openat(AT_FDCWD, "/sys/devices/system/cpu/cpufreq/policy0/energy_performance_preference", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
futex(0x60d31239cb40, FUTEX_WAIT_BITSET_PRIVATE, 4294967295, {tv_sec=757, tv_nsec=435783311}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
openat(AT_FDCWD, "/sys/devices/system/cpu/cpufreq/policy0/energy_performance_preference", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
futex(0x60d31239cb40, FUTEX_WAIT_BITSET_PRIVATE, 4294967295, {tv_sec=758, tv_nsec=436022196}, FUTEX_BITSET_MATCH_ANY) = -1 ETIMEDOUT (Connection timed out)
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
openat(AT_FDCWD, "/sys/devices/system/cpu/cpufreq/policy0/energy_performance_preference", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
sched_yield()                           = 0
[...]

PR #561, opened 2024-08-25
scx_bpfland: prevent reading energy profile if not available

arighi:
Avoid to periodically read the current performance profile from /sys/devices/system/cpu/cpufreq/policy0/energy_performance_preference if it's not available (i.e., with older CPUs or kernels without cpufreq).

This fixes issue #560.


hodgesds:
LGTM

PR #565, opened 2024-08-26
scx_loader: Add scheduler loader via system DBUS interface

vnepogodin:
This pull request introduces a utility with a DBUS interface to initiate schedulers with any flags potentially supported by a specific scheduler. It accomplishes this by starting schedulers as child processes.

The scx_loader offers the following features:
1. The `StartScheduler` method accepts the scx_name of the scheduler (e.g., "scx_rusty" or "scx_bpfland") and the scheduler mode (profile) as an enumerated value (raw input expects an unsigned integer).
2.  The `StartSchedulerWithArgs` method accepts the scx_name of the scheduler and the scheduler CLI arguments.
3. The `StopScheduler` method can be used to terminate a running scheduler.
4. The `CurrentScheduler` property indicates the currently active scheduler; if none is running, it returns "unknown." (Note: This property may be modified in the future to return an error instead.)
5. The `SchedulerMode` property provides information about the currently set scheduler mode.
6. The `SupportedSchedulers` property provides the list of currently supported schedulers.


htejun:
What are the benefits of turning each scheduler into library instead of just having a dbus attached launcher? The cost is that this bifurcates how schedulers are run depending on how it's run which does make things a bit more obscure. So, my preference would be layering on top.

vnepogodin:
> What are the benefits of turning each scheduler into library instead of just having a dbus attached launcher? The cost is that this bifurcates how schedulers are run depending on how it's run which does make things a bit more obscure. So, my preference would be layering on top.

Generally loading schedulers as native code(crates), is more reliable and self-contained, allows for better integration with schedulers if needed. Although I will rework `scx_loader` to manage child processes instead of loading/unloading schedulers.

In future some schedulers can adjusted separately to run as crates instead of child proc if needed

scheds/rust/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
 members = ["scx_lavd", "scx_bpfland", "scx_rustland", "scx_rlfifo",
-           "scx_rusty",
+           "scx_rusty", "scx_loader",
```
htejun, line 3: Let's put this under `rust` instead of `scheds/rust`

scheds/rust/scx_loader/meson.build
```diff
@@ -0,0 +1,9 @@
+sched = custom_target('scx_loader',
+              output: '@PLAINNAME@.__PHONY__',
+              input: 'Cargo.toml',
+              command: [cargo, 'build', '--manifest-path=@INPUT@', '--target-dir=@OUTDIR@',
+                        cargo_build_args],
+              env: cargo_env,
+              depends: sched_deps,
+              build_by_default: false,
+              build_always_stale: true)
```
htejun, line 9: If you pull in the latest, `scheds/rust/meson.build` has an example of how to hook this up from the parent directory. I think we just need to add the same template to `rust/meson.build`, specify `scx_loader` as the only target (as others are libraries) and make the top level `meson.build` to descend into the `rust` directory.

htejun:
Also, would it make sense to make the name a bit more specific - e.g. `scx_dbus_loader`?

scheds/rust/Cargo.toml
vnepogodin: sure

scheds/rust/scx_loader/meson.build
vnepogodin: I'm tried to add that. seems to work)

Although installation of dbus spec is a bit messy :)

meson-scripts/install_rust_user_scheds
```diff
@@ -15,13 +15,41 @@ for manifest in "$MESON_SOURCE_ROOT"/scheds/rust/*/Cargo.toml; do
...
+    fi
+
+    # XXX - Can we detect the current buildtype and install the correct one?
+    if [ ${#bins[@]} -gt 1 ]; then
+        echo "Found multiple binaries for $name under $target_dir, installing the latest which may not be what you want" 1>&2
+    fi
+
+    # Install bus spec for scx_loader
+    if [ "$name" = "scx_loader" ]; then
+        install -D -m 0644 "${source_dir}/org.scx.Loader.conf" "${DESTDIR}/usr/share/dbus-1/system.d/org.scx.Loader.conf"
```
htejun, line 52: I wonder whether the right thing to do is invoking `cargo install` with the right parameters. e.g. if someone does `cargo install scx_loader`, it should also install the dbus-1 file too, right?

rust/meson.build
```diff
@@ -0,0 +1,25 @@
...
+
+# the target to compile all rust binaries
+custom_target('rust_binaries',
+              output: '@PLAINNAME@.__PHONY__',
+              input: 'Cargo.toml',
+              command: cargo_cmd,
+              env: cargo_env,
+              depends: sched_deps,
+              build_by_default: true,
+              build_always_stale: true)
```
htejun, line 12: I would make it explicitly build only `rust_binaries`. No need to build libraries separately by default.

meson-scripts/install_rust_user_scheds
vnepogodin: yeah it's required. hmm I don't think its possible to install custom files with `cargo install`

vnepogodin: Looks like not possible yet to install custom files with `cargo install`, but possible with tools like `cargo xtask`

https://github.com/rust-lang/cargo/issues/2729

htejun: I see. Let's leave as-is for the time being.

rust/meson.build
vnepogodin: do you mean to set `build_always_stale` to false on `rust_binaries`?

htejun: There need to be two targets:
1. The default target which gets triggered by default.
2. Package-specific target so that users can do `meson compile -C build $TARGET`.
We don't want to build libraries by default for 1) but want to provide (we weren't before) individual targets for 2).

So, the default target should do `cargo build -p scx_loader`. It will probably look something like:
```
rust_libs = [ ... ]
rust_bins = [ ... ]
rust_all = rust_libs + rust_bins

// `-p bin` for each in rust_bin to cargo args
custom_target('rust_binaries', ...)

foreach package in rust_all
    custom_target(package, ...)
```

Please feel free to leave it as is. I can update afterwards too.

vnepogodin: Oh. I'll leave it as is then )

vnepogodin:
@htejun I'm currently thinking if we should provide additional systemd service for scx_loader to be able to autostart it by dbus, even if the service was enabled/started by user dbus will search by name and start it with propagation of called dbus method/property

htejun (approved):

PR #567, opened 2024-08-26
scx_bpfland: convert scheduler crate into lib+bin crate

vnepogodin:
This PR converts scheduler scx_bpfland crate into lib+bin crate, to be able to use(load/unload) scheduler by other crates


vnepogodin:
> r coordinate this chang

agree

arighi:
I'm going to close this one for now. I think the plan is to have scx_loader running the scheduler binaries first, then at some point in the future we may decide to resume thiis lib/bin approach.  Feel free to re-open it if needed.

PR #568, opened 2024-08-26
scx_rustland_core: small core design improvements

arighi:
Small design improvements, mostly focused at enhancing `scx_rustland_core` usability.

API changes:
 - temporarily dropped `RL_PREEMPT_CPU` (preemption will be re-added later with a better design)
 - allow to specify a vtime in `DispatchedTask` (the vtime will be passed to `scx_bpf_dispatched_vtime()`)

Misc Improvements:
 - `scx_rlfifo`: refactoring to make the code more suitable to be used as a generic template
 - include the FIFO code in the `README.md` of `scx_rustland_core` (in this way the README can be used standalone as a guide on how to implement schedulers using `scx_rustland_core`)

PR #569, opened 2024-08-26
scx_layered: Clean up in-code documentation; add commas for consistency

anh0516:
Minor work on scx_layered. There wasn't really that much to do.

I also added the same comma to scx_lavd, scx_rusty, and scx_mitosis as I added to scx_bpfland previously and scx_layered in this PR for consistency.

### DIFF
--- a/scheds/rust/scx_lavd/src/main.rs
+++ b/scheds/rust/scx_lavd/src/main.rs
@@ -113,7 +113,7 @@ struct Opts {
     #[clap(long)]
     monitor_sched_samples: Option<u64>,
 
-    /// Enable verbose output including libbpf details. Specify multiple
+    /// Enable verbose output, including libbpf details. Specify multiple
     /// times to increase verbosity.
     #[clap(short = 'v', long, action = clap::ArgAction::Count)]
     verbose: u8,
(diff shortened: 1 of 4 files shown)


hodgesds (approved):
Nice, thanks for the PR!

anh0516:
An unrelated scx_rustland test failure? That's not good. The log looks fine though, cleanly starting, using, stopping the scheduler and then powering off the VM. What's going on there?

hodgesds:
> An unrelated scx_rustland test failure? That's not good. The log looks fine though, cleanly starting, using, stopping the scheduler and then powering off the VM. What's going on there?

I saw that as well, looks unrelated. Maybe you could try rebasing? I see a recent (#568) scx_rustland PR also failed. cc @arighi

anh0516:
From #568 

> CI is failing for unrelated reasons (maybe systems on github are a bit busy?). Merging for now, instead of restarting the failed CI jobs.

So let's just wait and see if it fixes itself, I guess.

arighi:
> > An unrelated scx_rustland test failure? That's not good. The log looks fine though, cleanly starting, using, stopping the scheduler and then powering off the VM. What's going on there?
> 
> I saw that as well, looks unrelated. Maybe you could try rebasing? I see a recent (#568) scx_rustland PR also failed. cc @arighi

scx_rustland here failed due to an oom, we should probably create larger instances, by default virtme-ng is using 1G, but it might be a bit short if we run multiple stress-ng's (especially with the user-space schedulers). Maybe we can increase this to 2G?

hodgesds:
Merging for now since this is doc updates, will take a look at the CI jobs.

PR #570, opened 2024-08-26
scx_utils: Add Big/Little core logic to Topology

hodgesds:
This borrows some of the logic in scx_lavd for figuring out if a core is a P/E core. If this makes sense we can add helper methods directly on the topology to return P/E cores so that each scheduler doesn't have to reinvent the same logic.

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -74,7 +74,7 @@ use anyhow::Result;
 use glob::glob;
 use sscanf::sscanf;
 use std::collections::BTreeMap;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::slice::Iter;
 
 lazy_static::lazy_static! {
@@ -95,6 +95,12 @@ lazy_static::lazy_static! {
     pub static ref NR_CPUS_POSSIBLE: usize = libbpf_rs::num_possible_cpus().unwrap();
 }
 
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub enum CoreType {
+    Big,
+    Little,
+}
+
 #[derive(Debug, Clone)]
 pub struct Cpu {
     id: usize,
@@ -105,6 +111,7 @@ pub struct Cpu {
     l2_id: usize,
     l3_id: usize,
     llc_id: usize,
+    pub core_type: CoreType,
 }
 
 impl Cpu {
@@ -157,6 +164,7 @@ pub struct Core {
     id: usize,
     cpus: BTreeMap<usize, Cpu>,
     span: Cpumask,
+    pub core_type: CoreType,
 }
 
 impl Core {
@@ -409,7 +417,12 @@ fn cpus_online() -> Result<Cpumask> {
     Ok(mask)
 }
 
-fn create_insert_cpu(cpu_id: usize, node: &mut Node, online_mask: &Cpumask) -> Result<()> {
+fn create_insert_cpu(
+    cpu_id: usize,
+    node: &mut Node,
+    online_mask: &Cpumask,
+    avg_cpu_freq: Option<usize>,
+) -> Result<()> {
     // CPU is offline. The Topology hierarchy is read-only, and assumes
     // that hotplug will cause the scheduler to restart. Thus, we can
     // just skip this CPU altogether.
@@ -449,10 +462,24 @@ fn create_insert_cpu(cpu_id: usize, node: &mut Node, online_mask: &Cpumask) -> R
         span: Cpumask::new()?,
     });
 
+    let core_type = match avg_cpu_freq {
+        Some(avg_freq) => {
+            if max_freq >= avg_freq {
+                CoreType::Big
+            } else if max_freq < avg_freq {
+                CoreType::Little
+            } else {
+                CoreType::Big
+            }
+        }
+        None => CoreType::Big,
+    };
+
     let core = cache.cores.entry(core_id).or_insert(Core {
         id: core_id,
         cpus: BTreeMap::new(),
         span: Cpumask::new()?,
+        core_type: core_type.clone(),
     });
 
     core.cpus.insert(
@@ -466,6 +493,7 @@ fn create_insert_cpu(cpu_id: usize, node: &mut Node, online_mask: &Cpumask) -> R
             l2_id: l2_id,
             l3_id: l3_id,
             llc_id: llc_id,
+            core_type: core_type.clone(),
         },
     );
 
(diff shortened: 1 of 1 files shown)


rust/scx_utils/src/topology.rs
```diff
@@ -95,6 +95,13 @@ lazy_static::lazy_static! {
     pub static ref NR_CPUS_POSSIBLE: usize = libbpf_rs::num_possible_cpus().unwrap();
 }
 
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub enum CoreType {
+    PCore,
+    ECore,
+    Unknown,
```
hodgesds, line 102: Not a huge fan of the naming, so if someone can think of a better naming convention.

hodgesds:
I'll test this later when I have access to a hybrid machine, for now all the uniform architecture logic works correctly (hacked some logs into layered):
```
19:49:58 [INFO] core 0: pcore false ecore false
19:49:58 [INFO] core 1: pcore false ecore false
19:49:58 [INFO] core 2: pcore false ecore false
19:49:58 [INFO] core 3: pcore false ecore false
19:49:58 [INFO] core 4: pcore false ecore false
19:49:58 [INFO] core 5: pcore false ecore false
```

rust/scx_utils/src/topology.rs
htejun: Would `Big` and `Little` be better? Those are the OG terms and a bit wider spread, I think.

rust/scx_utils/src/topology.rs
```diff
@@ -150,13 +158,26 @@ impl Cpu {
     pub fn llc_id(&self) -> usize {
         self.llc_id
     }
+
+    /// Returns true if the max frequency of the cpu is higher than the average frequency of all
+    /// cores.
+    pub fn is_pcore(&self) -> bool {
+        self.core_type == CoreType::PCore
+    }
```
htejun, line 166: Can we just `pub` these fields rather than keeping adding trivial accessors? Not sure what these add.

rust/scx_utils/src/topology.rs
```diff
@@ -449,10 +487,24 @@ fn create_insert_cpu(cpu_id: usize, node: &mut Node, online_mask: &Cpumask) -> R
...
     });
 
+    let core_type = match avg_cpu_freq {
+        Some(avg_freq) => {
+            if max_freq > avg_freq {
+                CoreType::PCore
+            } else if max_freq > 0 && max_freq < avg_freq {
+                CoreType::ECore
+            } else {
+                CoreType::Unknown
```
htejun, line 497: Maybe just default to `Big`?

rust/scx_utils/src/topology.rs
hodgesds: Yeah, calling `is_ecore` just seemed wrong 😄

rust/scx_utils/src/topology.rs
```diff
@@ -449,10 +487,24 @@ fn create_insert_cpu(cpu_id: usize, node: &mut Node, online_mask: &Cpumask) -> R
         span: Cpumask::new()?,
     });
 
+    let core_type = match avg_cpu_freq {
+        Some(avg_freq) => {
+            if max_freq > avg_freq {
```
htejun, line 492: The kernel's capacity numbers are nice because it gives the relative max performance. `driver/base/arch_topology.c` exposes CPU capacity but it's not implemented for x86, unfortunately. `scx_bpf_cpuperf_cap()` exposes this but that's not accessible without running BPF progs. Given that `scx_bpf_cpuperf_cap()` can be called from any BPF progs, I wonder whether it'd be possible to make topology crate access those numbers? What do you think?

hodgesds: I was thinking of having a version available on for userspace via the `Topology` and a separate bpf helpers in `scheds/include/scx/topology.bpf.h` or something like that. I had also tried `scx_bpf_cpuperf_cap` in the past and as you mention it doesn't work on x86.

htejun: Ah, right. This is broken on x86. Once this is fixed, what I'm saying is that it might be useful to make `Topology` crate to query `scx_bpf_cpuperf_cap()` using its own embedded BPF program and then populate cap values accordingly in the struct so that userspace can use them.

rust/scx_utils/src/topology.rs
htejun: Another nitpick. Why not `CoreType::Big` and `CoreType::Little`? Do we ever import specific enum values?

htejun (approved):

hodgesds:
double checked the parsing logic and made a few fixes, should be working now:
```
22:28:14 [INFO] id: 0 type: Big
22:28:14 [INFO] id: 1 type: Big
22:28:14 [INFO] id: 2 type: Big
22:28:14 [INFO] id: 3 type: Big
22:28:14 [INFO] id: 4 type: Little
22:28:14 [INFO] id: 5 type: Little
22:28:14 [INFO] id: 6 type: Little
22:28:14 [INFO] id: 7 type: Little
22:28:14 [INFO] id: 8 type: Little
22:28:14 [INFO] id: 9 type: Little
22:28:14 [INFO] id: 10 type: Little
22:28:14 [INFO] id: 11 type: Little
```

rust/scx_utils/src/topology.rs
multics69: I also tried `scx_bpf_cpuperf_cap()`, but it returns the same 1024 for all CPU types.

rust/scx_utils/src/topology.rs
```diff
@@ -449,10 +463,24 @@ fn create_insert_cpu(cpu_id: usize, node: &mut Node, online_mask: &Cpumask) -> R
...
+                CoreType::Big
+            } else if max_freq > 0 && max_freq < avg_freq {
+                CoreType::Little
+            } else {
+                CoreType::Big
+            }
```
multics69, line 474: I think it could be a bit clearer the remove the last else and make the first if condition to `<=`. So the code should look like:

```
if max_freq >= avg_freq {
   // ...
}
} else if max_freq > 0 && max_freq < avg_freq {
  // ...
}
```

rust/scx_utils/src/topology.rs
```diff
@@ -449,10 +463,24 @@ fn create_insert_cpu(cpu_id: usize, node: &mut Node, online_mask: &Cpumask) -> R
         span: Cpumask::new()?,
     });
 
+    let core_type = match avg_cpu_freq {
+        Some(avg_freq) => {
+            if max_freq > avg_freq {
+                CoreType::Big
+            } else if max_freq > 0 && max_freq < avg_freq {
```
multics69, line 470: Do we need to check `max_freq > 0`? I think it should be always true.

rust/scx_utils/src/topology.rs
multics69: It would be nice to have a `Turbo` type. When there are two or more types of CPU with different max freq, the max freq CPUs are turbo bootable CPUs. This is true at least Intel Raptor Lake and Qualcomm Snapdragon platforms.

multics69 (approved):

rust/scx_utils/src/topology.rs
```diff
@@ -449,10 +463,24 @@ fn create_insert_cpu(cpu_id: usize, node: &mut Node, online_mask: &Cpumask) -> R
...
+                CoreType::Little
+            } else {
+                CoreType::Big
+            }
+        }
+        None => CoreType::Unknown,
```
htejun, line 476: Do we need unknown? Wouldn't the scheduler treat it as `Big` anyway?

htejun: e.g. maybe just printing out a warning when parsing fails and hiding the complexity from the schedulers would be better?

hodgesds: Yeah, should always fallback to big. Doesn't make sense to add extra complexity to handle that case.

PR #571, opened 2024-08-27
scx_lavd: make a loop easier to correctly verify

multics69:
With an ill combination of old kernel and old LLVM, the BPF verifier incorrectly detects an infinite loop. After changing the loop with a constant end, the old verifier can pass the code.


hodgesds (approved):
LGTM

PR #572, opened 2024-08-27
scx_bpfland: fix turbo boost domain nullifying primary domain limits

arighi:
When creating the turbo boost scheduling domain, we might use a full CPU mask (selecting all possible CPUs) to indicate "do not prioritize turbo boost CPUs" or when all CPUs have the same maximum frequency.

This approach works when the primary domain also contains all the CPUs, as the complete overlap allows the CPU selection logic to ignore the turbo boost domain and start picking CPUs directly from the primary domain.

However, if the primary domain doesn't include all CPUs, the two domains won't fully overlap, which can lead to the turbo boost domain incorrectly including all CPUs, thereby negating the restrictions set by the primary scheduling domain.

To resolve this, an empty CPU mask should be used for the turbo boost domain when turbo boost CPUs aren't prioritized. If the turbo boost domain is empty, it should be entirely bypassed, and the selection should proceed directly to the primary domain.

Reported-by: Changwoo Min <changwoo@igalia.com>

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -592,7 +592,9 @@ retry:
 	 * Try to dispatch on the turbo boosted CPUs first. If we can't find
 	 * any idle CPU, re-try again with the primary scheduling domain.
 	 */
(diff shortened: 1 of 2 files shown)


multics69 (approved):
The code looks good to me. Also, I tested on my Raptor Lake NUC and it works well.

PR #573, opened 2024-08-27
scx_stats: Drop sched-ext namespace

sirlucjan:
Since scx_stats has been implemented, namespace is already redundant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,79 @@ $ cargo install scx_rusty
 
 and `scx_rusty` will be built and installed as `~/.cargo/bin/scx_rusty`.
 
+## Checking scx_stats
+
+With the implementation of scx_stats, schedulers no longer display statistics by default. To display the statistics from the currently running scheduler, a manual user action is required.
+Below are examples of how to do this.
+
+- To check the scheduler statistics, use the
+
+```
+scx_SCHEDNAME --monitor $INTERVAL
+```
+
+for example 0.5 - this will print the output every half a second
+
+```
+scx_bpfland --monitor 0.5
+```
+Some schedulers may implement different or multiple monitoring options. Refer to --help of each scheduler for details.
+Most schedulers also accept ` --stats $INTERVAL` to print the statistics directly from the scheduling instance.
+
+#### Examples
+
+- scx_bpfland
+
+```
+❯ scx_bpfland --monitor 5
+[scx_bpfland] tasks -> run:  3/4  int: 2  wait: 3    | nvcsw: 3    | dispatch -> dir: 0     prio: 73    shr: 9
+[scx_bpfland] tasks -> run:  4/4  int: 2  wait: 2    | nvcsw: 3    | dispatch -> dir: 1     prio: 3498  shr: 1385
+[scx_bpfland] tasks -> run:  4/4  int: 2  wait: 2    | nvcsw: 3    | dispatch -> dir: 1     prio: 2492  shr: 1311
+[scx_bpfland] tasks -> run:  4/4  int: 2  wait: 3    | nvcsw: 3    | dispatch -> dir: 2     prio: 3270  shr: 1748
+```
+
+- scx_rusty
+
+```
+❯ scx_rusty --monitor 5
+###### Thu, 29 Aug 2024 14:42:37 +0200, load balance @  -265.1ms ######
+cpu=   0.00 load=    0.17 mig=0 task_err=0 lb_data_err=0 time_used= 0.0ms
+tot=     15 sync_prev_idle= 0.00 wsync= 0.00
+prev_idle= 0.00 greedy_idle= 0.00 pin= 0.00
+dir= 0.00 dir_greedy= 0.00 dir_greedy_far= 0.00
+dsq=100.00 greedy_local= 0.00 greedy_xnuma= 0.00
+kick_greedy= 0.00 rep= 0.00
+dl_clamp=33.33 dl_preset=93.33
+slice=20000us
+direct_greedy_cpus=f
+  kick_greedy_cpus=f
+  NODE[00] load=  0.17 imbal=  +0.00 delta=  +0.00
+   DOM[00] load=  0.17 imbal=  +0.00 delta=  +0.00
+```
+
+- scx_lavd
+
+```
+❯ scx_lavd --monitor-sched-samples 1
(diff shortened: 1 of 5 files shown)


services/README.md
```diff
@@ -62,26 +62,29 @@ systemctl unset-environment SCX_FLAGS_OVERRIDE
...
 ```
 
-- To verify the amount of space taken up by the logs, use:
+## Checking scx_stats
+
+- To check the scheduler statistics, use the
 
 ```
-journalctl --namespace=sched-ext --disk-usage
+scx_SCHEDNAME --monitor X
```
arighi, line 85: Maybe change to `--monitor SEC` here to make it more clear that the argument defines the interval (in seconds) to print the output.

services/README.md
```diff
@@ -62,26 +62,29 @@ systemctl unset-environment SCX_FLAGS_OVERRIDE
...
 ```
 
+X - for example 1 - this will print the output every second
+
+
```
arighi, line 90: And maybe add a real example, like:
```
scx_bpfland --monitor 0.5
```

To make it even more clear and to show that it also supports numbers with decimals.

arighi (approved):
Left a couple of comments (feel free to ignore), but overall looks good!

sirlucjan:
@arighi I don't like to ignore wise advice. Is it better now?

services/README.md
multics69: I agree adding real examples would be nicer here.
For example, in case of LAVD, its monitoring option is a bit different from the others. 

    /// Run in monitoring mode. Show the specified number of scheduling
    /// samples every second.
    #[clap(long)]
    monitor_sched_samples: Option<u64>,

multics69 (approved):

sirlucjan:
@arighi @multics69 

Maybe we should move the log information to a separate README or write information about it in the main one?

arighi:
Hm... maybe the monitor info should go in the main README, since they're affecting everyone and not only those that are running the schedulers via systemd.

sirlucjan:
Let me know if the corrected examples are  good and I will send the revisions to README.

README.md
```diff
@@ -344,6 +344,87 @@ $ cargo install scx_rusty
...
+|    428 |  111375 | scx_lavd          | RHBGN |    2 |   -1 |        8703000 |  3000000 |   34812 |       12 |      70 |       20 |       0 |         0 |   1702362 |         0 |         0 |       22 |       84 |      822 |       53 |      1 |
+|    429 |  111375 | scx_lavd          | RHBGP |    2 |    1 |         145411 |  3000000 |    2472 |       51 |      71 |       20 |       0 |      2976 |     95521 |     26907 |     10348 |       71 |       85 |      393 |       26 |      1 |
+|    430 |  111375 | scx_lavd          | RHBGN |    3 |   -1 |          83886 |  3000000 |    1482 |       53 |      72 |       20 |       0 |      5681 |     51129 |     16819 |     21793 |       72 |       85 |     1024 |       67 |      1 |
+|    431 |  105466 | scx_lavd          | RHBEN |    2 |   -1 |              0 |  3000000 |       1 |       71 |      72 |       20 |       0 |      2037 |     38775 |      1404 |      1969 |       60 |       86 |      919 |       50 |      1 |
+```
+
+- scx_rustland
+
+```
+❯ run0 scx_rustland --monitor 5
```
ptr1337, line 419: Why using here run0?

README.md
```diff
@@ -344,6 +344,87 @@ $ cargo install scx_rusty
...
+- To check the scheduler statistics, use the
+
+```
+scx_SCHEDNAME --monitor SEC
+```
+
+for example 0.5 - this will print the output every half a second
+
+```
+scx_bpfland --monitor 0.5
```
ptr1337, line 358: Personally would remove this part with the "for example 0.5". This just makes the README bigger with not much value

README.md
```diff
@@ -344,6 +344,87 @@ $ cargo install scx_rusty
...
+dir= 0.00 dir_greedy= 0.00 dir_greedy_far= 0.00
+dsq=100.00 greedy_local= 0.00 greedy_xnuma= 0.00
+kick_greedy= 0.00 rep= 0.00
+dl_clamp=33.33 dl_preset=93.33
+slice=20000us
+direct_greedy_cpus=f
+  kick_greedy_cpus=f
+  NODE[00] load=  0.17 imbal=  +0.00 delta=  +0.00
+   DOM[00] load=  0.17 imbal=  +0.00 delta=  +0.00
+###### Thu, 29 Aug 2024 14:42:42 +0200, load balance @ -1268.9ms ######
```
ptr1337, line 390: Would also reduce here the logs, so that the README is not too much filled with logs.

ptr1337 (changes requested):
Do we really need to explain from every supported scx scheduler how to use monitor?

I think you could write, which are the **supported** schedulers and bring **one** example. instead of showing of every scheduler a example.

README.md
```diff
@@ -344,6 +344,87 @@ $ cargo install scx_rusty
 
 and `scx_rusty` will be built and installed as `~/.cargo/bin/scx_rusty`.
 
+## Checking scx_stats
+
+- To check the scheduler statistics, use the
```
htejun, line 349: How about something like:

Most schedulers implement `--monitor $INTERVAL` which fetches and shows the detailed statistics from the running scheduler. Note that this does not launch the scheduler. The monitoring instance only shows the statistics. Some schedulers may implement different or multiple monitoring options. Refer to `--help` of each scheduler for details. Most schedulers also accept `--stats $INTERVAL` to print the statistics directly from the scheduling instance.

README.md
sirlucjan: Removed

README.md
sirlucjan: Fixed

README.md
sirlucjan: I think such an example can stay to show exactly what it is about, but if more people think the same, I will remove it.

README.md
sirlucjan: Updated, could you take a look now @htejun

sirlucjan:
> Do we really need to explain from every supported scx scheduler how to use monitor?
> 
> I think you could write, which are the **supported** schedulers and bring **one** example. instead of showing of every scheduler a example.

@multics69 asked about examples.

README.md
htejun: Maybe use "statistics" consistently? Schedulers do print out logs like info and warning messages. It just doesn't print out detailed statistics by default.

arighi:
It looks like we have enough approvals and with scx_rustland also converted to use scx_stats I think we can safely apply this change.

PR #574, opened 2024-08-27
scx_bpfland: rely on Topology to classify CPU types

arighi:
Integrate in Topology the concept of "Turbo" CPUs (to represent the turbo-boosted cores), refactor scx_bpfland to use simply rely on Topology to classify the CPUs in Turbo, Big and Little.

@hodgesds let me know if you see any problem with the new core type "Turbo" or if you have a better idea to represent it in struct Cpu (this is more of a RFC).

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -74,7 +74,7 @@ use anyhow::Result;
 use glob::glob;
 use sscanf::sscanf;
 use std::collections::BTreeMap;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::slice::Iter;
 
 lazy_static::lazy_static! {
@@ -97,7 +97,7 @@ lazy_static::lazy_static! {
 
 #[derive(Debug, Clone, Eq, Hash, PartialEq)]
 pub enum CoreType {
-    Big,
+    Big { turbo: bool },
     Little,
 }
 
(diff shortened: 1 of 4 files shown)


rust/scx_utils/src/topology.rs
```diff
@@ -97,6 +97,7 @@ lazy_static::lazy_static! {
 
 #[derive(Debug, Clone, Eq, Hash, PartialEq)]
 pub enum CoreType {
+    Turbo,
     Big,
```
htejun, line 101: Would it make sense to make `Turbo` a variant of `Big`? ie. so that it's `Big { turbo: bool }`? That way, the hierarchy is encoded in the type system and people can use `match` more naturally rather than using `is_*()` helpers.

arighi: Hm... that seems really nice, will check that!

hodgesds:
I think this should work well, don't have access to a turbo type Intel CPU machine 😄

arighi:
Added @htejun's  idea (turbo as a property of Big) and also added the special strings `all` and `none` to inizialize a cpumask using `Cpumask::from_str()`. I need this in scx_bpfland to make the code cleaner, but they could be useful for other cases as well.

htejun (approved):

hodgesds (approved):
This is much nicer, thanks for cleaning it up!

PR #576, opened 2024-08-28
scx_lavd: Add power mode clarification to --no-prefer-turbo-core

anh0516:
`--no-prefer-turbo-core` was added after I added the clarification that scx_lavd's tunables are set by the specified power mode. Add the same clarification to it.


multics69 (approved):
Thanks!

PR #582, opened 2024-08-29
scx_layered: Add layer growth config

hodgesds:
Add a per layer config for different implementations of layer growth algorithms. Convert the existing default logic into a default layer growth algorithm and add a linear implementation.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -94,6 +94,7 @@ lazy_static::lazy_static! {
 			preempt_first: false,
 			exclusive: false,
                         slice_us: 20000,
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/main.rs
```diff
@@ -442,6 +445,17 @@ enum LayerMatch {
     TGIDEquals(u32),
 }
 
+#[derive(Clone, Debug, Parser, Serialize, Deserialize)]
+#[clap(rename_all = "snake_case")]
+enum LayerGrowthAlgo {
```
hodgesds, line 450: I dislike all the naming here, so feel free to suggest something better.

htejun (approved):

PR #583, opened 2024-08-29
scx_bpfland: always rely on prev_cpu with single-CPU tasks

arighi:
When selecting an idle for tasks that can only run on a single CPU, always check if the previously used CPU is sill usable and online, instead of trying to figure out the single allowed CPU looking at the task's cpumask.

Apparently, per-CPU kthreads can run on a CPU that is not reported in their allowed cpumask (or the cpumask is not properly updated or coeherent with the actual task state).

This could lead to stalls, because we may end up dispatching the kthread to a per-CPU DSQ that is not compatible with its allowed cpumask.

Example:

kworker/u32:2[173797] triggered exit kind 1026:
  runnable task stall (kworker/2:1[70] failed to run for 7.552s)
...
  R kworker/2:1[70] -7552ms
      sticky/holding_cpu=-1/-1 dsq_id=0x8 dsq_vtime=234483011369
      cpus=04

In this case kworker/2 can only run on CPU #2 (cpus=0x4), but it's dispatched to dsq_id=0x8, that can only be consumed by CPU 8 => stall.

To prevent this, always check if prev_cpu is usable, online and idle, with single-CPU tasks. Otherwise, do not dispatch the task directly to a per-CPU DSQ and bounce it to a shared DSQ instead (either priority or regular, depending on its interactive state).

Moreover, introduce an explicit error check in dispatch_direct_cpu() to improve detection of similar issues in the future, and drop lookup_task_ctx() in favor of try_lookup_task_ctx(), since we can now safely handle all the cases where the task context is not found.


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -554,6 +527,24 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
...
+	 * For tasks that can run only on a single CPU, we can simply verify if
+	 * their only allowed CPU is still usable, online and idle.
+	 *
+	 * Moreover, if local_kthreads is enabled, always dispatch per-CPU
+	 * kthreads directly to their target CPU, independently on its idle
+	 * state.
+	 */
+	if (p->nr_cpus_allowed == 1) {
+		if (bpf_cpumask_test_cpu(prev_cpu, p->cpus_ptr) &&
+		    bpf_cpumask_test_cpu(prev_cpu, online_cpumask) &&
```
multics69, line 540: I wonder if an explicit checking against online_cpumask is indeed necessary. I guess anyway such case will be handled by  scx_bpf_test_and_clear_cpu_idle(prev_cpu). What do you think?

multics69 (approved):
Overall LGTM

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: I don't think `scx_bpf_test_and_clear_cpu_idle()` is aware of the CPUs that may go offline here, the idle_masks are initialized by `reset_idle_masks()`, that is called only from `scx_ops_enable()`, so IIUC idle masks are not updated on hotplug events.

@htejun should we call `reset_idle_masks()` from `handle_hotplug()`? That would temporarily invalidate the current idle state, but it should converge soon, and in this way offline CPUs would be automatically considered as busy without having to check the online cpumask.

htejun: It's safe to dispatch to offline CPUs. sched_ext core will just punt them to `SCX_DSQ_GLOBAL` if the target CPU is offline, so the BPF scheduler shouldn't need to worry about that - the next time the task comes around, its `prev_cpu` won't be the offline CPU.

htejun:
I *think* this might be masking a bug elsewhere. While `cpus_allowed` is being updated, a running gets enqueued on the "wrong" CPU before migration is actually executed but that gets dequeued right after and then migration is executed so that the task is actually on the allowed CPU. This shouldn't lead to a situation where a task is stuck on a DSQ for a CPU which it's not allowed on. How reproducible is the problem?

htejun (approved):

htejun:
Please feel free to merge but it'd probably be a good idea to root cause it better.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: With `SCX_DSQ_LOCAL` / `SCX_DSQ_LOCAL_ON` yes, but `dispatch_direct_cpu()` here is actually dispatching to the custom per-CPU DSQ, that can be only consumed by the corresponding CPU. If I dispatch the task and the CPU goes offline, the task won't be consumed.

htejun: Ah, okay, bpfland implements its own hotplug ops and doesn't restart. Down the line, a better way to deal with this probably is draining the per-cpu DSQ from `ops.cpu_offline()`. The online mask is cleared after the scheduler on that CPU is shut down, so there is still gonna be gaps.

arighi:
> I _think_ this might be masking a bug elsewhere. While `cpus_allowed` is being updated, a running gets enqueued on the "wrong" CPU before migration is actually executed but that gets dequeued right after and then migration is executed so that the task is actually on the allowed CPU. This shouldn't lead to a situation where a task is stuck on a DSQ for a CPU which it's not allowed on. How reproducible is the problem?

I can reproduce this pretty fast with `stress-ng --race-sched 256`.

If you want to see it happening quickly, add this:
https://github.com/sched-ext/scx/commit/42064c4cfe946f7582ba06c225b7ef0b4d6adf59

Then run the stress-ng above and this happens immediately:
```
stress-ng-race-[2277861] triggered exit kind 1025:
  scx_bpf_error (prev_cpu 1 invalid for 2277867 stress-ng-race-)
```

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Yep, that was my original idea, but I need an API to trigger a re-enqueue of all the tasks in a custom DSQ (I mentioned this during office hours), because I can't use `scx_bpf_consume()` directly from `ops.cpu_offline()`.

Anyway, even if there's a gap and a task is dispatched to a per-cpu DSQ associated to an offline CPU it can still be consumed from other CPUs via `consume_offline_cpus()`. Maybe I can just get rid of these online checks here and just let the tasks be consumed by the other CPUs.

arighi:
Changed the logic a little bit to simply skip the idle CPU detection when `prev_cpu` is not in `p->cpus_ptr`, since it's pointless to find an optimal idle CPU for a task while it's explicitly changing its affinity.

This allows to prevent the stall issue and it also seems a better approach in general.

PR #587, opened 2024-08-29
scx_layered: Add override context for layer hinting

hodgesds:
Add thread hinting to scx_layered by a user mappable map that allows for overriding layers for a pid. There's no locking or safety ☠️ yet.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -33,6 +33,7 @@ enum consts {
 	MAX_COMM		= 16,
 	MAX_LAYER_MATCH_ORS	= 32,
 	MAX_LAYERS		= 16,
+	MAX_LAYER_OVERRIDES	= 1024,
 	USAGE_HALF_LIFE		= 100000000,	/* 100ms */
 
 	HI_FALLBACK_DSQ		= MAX_LAYERS * MAX_DOMS,
@@ -162,4 +163,8 @@ struct layer {
 	unsigned int		perf;
 };
 
+struct user_ctx {
+	u64			task_overrides[MAX_LAYER_OVERRIDES];
+	u32			nr_overrides;
+};
 #endif /* __INTF_H */
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -34,6 +34,7 @@ enum consts {
 	MAX_LAYER_MATCH_ORS	= 32,
 	MAX_LAYERS		= 16,
 	MAX_LAYER_OVERRIDES	= 1024,
+	MAX_OVERRIDE_YIELDS	= 64,
 	USAGE_HALF_LIFE		= 100000000,	/* 100ms */
 
 	HI_FALLBACK_DSQ		= MAX_LAYERS * MAX_DOMS,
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -162,4 +163,8 @@ struct layer {
 	unsigned int		perf;
 };
 
+struct user_ctx {
+	u64			task_overrides[MAX_LAYER_OVERRIDES];
+	u32			nr_overrides;
```
htejun, line 168: Can this have a generation number so that it doesn't have to be walked every time?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -314,6 +323,7 @@ struct task_ctx {
 	struct bpf_cpumask __kptr *layered_cpumask;
 
 	bool			all_cpus_allowed;
+	bool			is_override;
```
htejun, line 326: and then this can be the copy of the generation number where it's only effective if it matches the current gen.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -162,4 +163,8 @@ struct layer {
 	unsigned int		perf;
 };
 
+struct user_ctx {
```
htejun, line 166: Given that we have uid matches, `user` may be a bit confusing? How about just `override_ctx` or `overrides`.

hodgesds:
Updated with a some better names and some poor synchronization. There's now a way to yield tasks back to the scheduler which should be interesting. There's certainly some vruntime error introduced with user overrides though.

hodgesds:
This will likely need to support pid namespaces, which will be... fun

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -418,6 +430,96 @@ int BPF_PROG(tp_task_rename, struct task_struct *p, const char *buf)
...
+			 */
+			tctx->layer = layer_idx;
+			tctx->override_gen = octx->gen;
+			__sync_fetch_and_add(&layer->nr_tasks, 1);
+			bpf_task_release(p);
+		}
+
+		// To yield control back to the BPF scheduler userspace is
+		// allowed to yield a task. Reset all the yielded tasks.
+		bpf_for(i, 0, octx->nr_yields) {
```
htejun, line 501: When override gen changes, wouldn't that automatically revoke overrides of all tasks that aren't anymore in the current generation? ie. `maybe_refresh_layer()` can test whether its gen is behind the global gen and if so, revert back to the original layer and then clear the gen number to indicate that the situation has been dealt with.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -418,6 +430,96 @@ int BPF_PROG(tp_task_rename, struct task_struct *p, const char *buf)
...
+			/*
+			 * XXX: Need to adjust the vtime delta with the
+			 * previous layer similar to maybe_refresh_layer.
+			 * However, from this context the value is read only so
+			 * it may be incorrect.
+			 *
+			 * p->scx.dsq_vtime = layer->vtime_now;
+			 *
+			 */
+			tctx->layer = layer_idx;
```
htejun, line 493: So, the way to do this, I think, is just updating the generation numbers here and what the target index should be and let `maybe_refresh_layer()` do the actual update. That way, it'll update in a synchronized manner and vtime will be taken care of too.

hodgesds: Yeah, I think that makes sense... keep the updates in one place.

htejun:
Is this still in progress?

hodgesds:
I'll close for now, I think I ran into a few issues that would be painful:

- dealing with cgroups and pid namespace translations when bpf maps are mapped to user space processes.

PR #591, opened 2024-08-30
scx_lavd: introduce "autopilot" mode and misc. optimization & bug fix

multics69:
The PR contains the following changes:

- Introduce "autopilot" mode, which dynamically switch the scheduler's power mode between powrsave, balanced, and performance according to system-wide CPU utilization. Under low load (<5%), the scheduler runs  with powersave mode. Under moderate load (5% < x < 30%), it runs with balanced mode. Under high load (>30%), it runs with performance mode.
- Rename the "--auto" option to "--autopower" to avoid the confusion with the "autopilot" mode.
- Print the power mode change in the autopower mode
- Add a fastpath in ops.select_cpu() for a sharply pinned task
- Improve the accuracy of cpu utilization calculation

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/intf.h
+++ b/scheds/rust/scx_lavd/src/bpf/intf.h
@@ -89,6 +89,10 @@ enum consts {
 	LAVD_CC_CPU_PIN_INTERVAL_DIV	= (LAVD_CC_CPU_PIN_INTERVAL /
 					   LAVD_SYS_STAT_INTERVAL_NS),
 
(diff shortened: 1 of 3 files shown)


multics69:
Thanks for the comment @1Naim. That's a good point. Maybe.  Since the balanced mode (i.e., core compaction) behaves like the performance mode as the CPU utilization reaches 50%, the upper limit should be below 50%.

Once the PR lands, I will test some more configurations -- e..g, 10% -  40% -- on several different hardware settings.

PR #597, opened 2024-09-02
scx_lavd: misc updates (verifier, README, monitor option name, and micro-optimization)

multics69:
This PR contains the following misc updates:

- check a core type of the previous cpu at pick_idle_cpu()
- rename "--monitor_sched_samples" to "--monitor" for consistency
- update README
- make the old verifier happy (bpf_cpumask_set_cpu)

### DIFF
--- a/scheds/rust/scx_lavd/README.md
+++ b/scheds/rust/scx_lavd/README.md
@@ -11,8 +11,9 @@ leveraging the task's latency-criticality information in making various
 scheduling decisions (e.g., task's deadline, time slice, etc.). As the name
 implies, LAVD is based on the foundation of deadline scheduling. This scheduler
 consists of the BPF part and the rust part. The BPF part makes all the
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_lavd/src/main.rs
```diff
@@ -125,7 +125,7 @@ struct Opts {
     /// Run in monitoring mode. Show the specified number of scheduling
     /// samples every second.
     #[clap(long)]
-    monitor_sched_samples: Option<u64>,
+    monitor: Option<u64>,
```
htejun, line 128: What lavd prints out is a bit different from stats printed out by other schedulers and the parameter's meaning is also different (number of samples instead of interval). Maybe it'd be useful to add more traditional statistics to lavd too which can then use the `--stats` and `--monitor` option rather than renaming sampling to take up the name.

htejun (approved):

PR #600, opened 2024-09-03
scx_bpfland: improve cpufreq awareness

arighi:
Provide a basic cpufreq control, based on the selected EPP and scheduler's performance profile.

Improve turbo boost CPU selection logic.

### DIFF
--- a/scheds/rust/scx_bpfland/src/main.rs
+++ b/scheds/rust/scx_bpfland/src/main.rs
@@ -58,9 +58,10 @@ const SCHEDULER_NAME: &'static str = "scx_bpfland";
 
 #[derive(PartialEq)]
 enum Powermode {
-    Powersave,
-    Performance,
     Turbo,
+    Performance,
+    Powersave,
+    Any,
 }
 
 fn get_primary_cpus(mode: Powermode) -> std::io::Result<Vec<usize>> {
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -356,15 +348,27 @@ impl<'a> Scheduler<'a> {
...
-        let res =
-            File::open("/sys/devices/system/cpu/cpufreq/policy0/energy_performance_preference")
-                .and_then(|mut file| {
-                    let mut contents = String::new();
-                    file.read_to_string(&mut contents)?;
-                    Ok(contents.trim().to_string())
-                });
-
-        res.unwrap_or_else(|_| "none".to_string())
+        let energy_pref_path =
```
hodgesds, line 351: At some point maybe we can pull some of the cpu frequency governor into the topology helpers.

arighi: Yes, that's a good idea. I'll send a separate PR for that.

BTW, unrelated but similar topic, it'd be nice to include also `amd_pstate_highest_perf` to the turbo cores logic.

htejun:
There's no "auto" mode when SCX scheduler is loaded and schedutil governor is used (there's no signal that the kernel side can use or even know). If you don't change anything, it'll just remain at the last setting.

arighi:
Ah.. I was missing that, then I always need to set a value, I can't just ignore it (unless I actually want to keep the previous value). Ok, thanks @htejun I'll change the logic to take that aspect into account.

htejun (approved):

PR #605, opened 2024-09-04
scx_rustland_core: avoid critical failures due by missing task context

arighi:
Prevent triggering a critical error when a local context for a task can't be found. Instead, handle the error gracefully to enhance the robustness of the schedulers based on scx_rustland_core.


htejun (approved):
It *might* be useful to distinguish when lookup failures can happen and when not, as if they do when they aren't supposed to, they tend to indicate BPF or scheduler bug. Maybe doing that doesn't make sense in rustland_core tho.

arighi:
Hm... probably the best is to print a message in debugfs when it happens, at least in this way `scx_rustland_core`-based schedulers have the option to enable debugging and check if/when it's happening.

PR #615, opened 2024-09-05
scx_bpfland: enable "auto" mode by default

arighi:
Rename "turbo domain" to "preferred domain", that conceptually is more generic and introduce the new option `--preferred-domain CPUMASK`, which allows users to define the preferred domain, specifying a cpumask as a hex number. By default ("auto") the scheduler will always try to detect and use the fastest CPUs in the system.

Moreover, adjust the cpufreq logic to use "auto" both with the "balance_power" and "balance_performance" EPP profiles.

Then, enable "auto" mode by default: the scheduler will try to automatically determine the optimal primary domain, preferred domain and cpufreq level, based on the selected scheduler and energy profiles.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -140,9 +140,9 @@ UEI_DEFINE(uei);
 private(BPFLAND) struct bpf_cpumask __kptr *primary_cpumask;
 
 /*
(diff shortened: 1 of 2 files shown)


sirlucjan:
I have been using the “auto” flag for several weeks and have not found any regressions.

arighi:
Added @sirlucjan tested-by line, thanks!

htejun (approved):

PR #623, opened 2024-09-06
fix/enable rust tests, make build faster

likewhatevs:
This PR fixes rust tests and configures CI to run them on commit. 

It does this by configuring 3 CI jobs.

One CI job will, hourly, build and cache a new sched_ext kernel if there is a new commit on for-next. This periodic job only runs on main. 

The other two CI jobs run meson tests and cargo tests. These use that cached kernel.

All branches can that cached kernel from main. not-main branches will create their own caches if one for main does not exist for for-next's current head, but these are less useful because they are not shared like main's is.

When the kernel is cached, job turnaround time (test time) looks to be ~12 minutes: https://github.com/likewhatevs/scx/actions/runs/10736459682

Not-cached job time looks like 20-25 minutes.

I'm not sure if I should replace the current configs with this config (I think it's a superset, or can be made one easily enough) or add another (as this PR does atm). 

Lmk what you think, thanks!

### DIFF
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -1,0 +1,138 @@
+name: caching-build
+on:
+  # only runs on main, hourly cache update used by all branches
+    schedule:
+      - cron: "0 * * * *"
+    push:
+  
+jobs:
+  build-kernel:
+    runs-on: ubuntu-22.04
+    steps:
+      # redundancy to exit fast
+      - run: echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-selections
+      - run: sudo apt update
+      - run: sudo apt install -y git --no-install-recommends
+      # get latest head commit of sched_ext for-next
+      - run: echo "SCHED_EXT_KERNEL_COMMIT=$(git ls-remote https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git heads/for-next | awk '{print $1}')" >> $GITHUB_ENV
+
+      - uses: actions/checkout@v4
+      
+      # use cached kernel if available, create after job if not
+      - name: Cache Kernel
+        id: cache-kernel
+        uses: actions/cache@v4
+        with:
+          path: |
+            linux
+          key: ${{ env.SCHED_EXT_KERNEL_COMMIT }}
+
+      - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}          
+        uses: ./.github/actions/install-deps-action
+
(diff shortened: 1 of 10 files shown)


.github/workflows/caching-build.yml
```diff
@@ -0,0 +1,138 @@
...
+        run: exit -1
+      
+      - uses: Swatinem/rust-cache@v2
+        with: 
+          shared-key: ${{ env.SCHED_EXT_KERNEL_COMMIT }}
+          workspaces: rust
+          prefix-key: "1"
+      - run: cargo build  --manifest-path rust/Cargo.toml
+      - run: cargo test --manifest-path rust/Cargo.toml --no-run
+      - run: vng -m10G --force-9p -r linux/arch/x86/boot/bzImage --net user --exec "cargo test --manifest-path rust/Cargo.toml"
```
arighi, line 137: May I suggest to change the vng command as following:
```
vng -v -m10G -c8 --force-9p -r linux/arch/x86/boot/bzImage  --net user -- cargo test --manifest-path rust/Cargo.toml
```
 - `-v` allows to catch potential OOM conditions (even if 10G should be enough to run the tests)
 - `-c8` always using 8 CPUs allows to run tests in a consistent environment, independently on the CPUs on the host (I'm thinking at the Cpumask tests)
 - `--exec` is the old-style deprecated syntax, better to move to the new syntax

rust/scx_utils/src/cpumask.rs
```diff
@@ -17,36 +17,42 @@
 //! hexadecimal string:
 //!
 //!```
-//!     let all_zeroes = cpumask::new();
-//!     let str = "0xff00ff00";
-//!     let from_str_mask = cpumask::from_string(str);
+//!     use scx_utils::Cpumask;
+//!     let all_zeroes = Cpumask::new();
+//!     let str = String::from("0xff00ff00");
```
arighi, line 22: I think this one is going to fail unless we have all these CPUs inside the vng instance (hence my comment about using just 8 CPUs). So, maybe change this code chunk with:
```
use scx_utils::Cpumask;
let all_zeroes = Cpumask::new();
let str = String::from("0xf0");
```

arighi (approved):
This looks great, thanks for doing this! I left a couple of comments, but I overall I really like this and I think it's a good improvement for our CI.

rust/scx_utils/src/cpumask.rs
likewhatevs: Ahh that makes sense, thank you!

htejun (approved):

arighi (approved):

Issue #625, opened 2024-09-06
scx_utils: bug in cpumask from hex str logic?


likewhatevs:
when this line of cpumask doctest:
https://github.com/sched-ext/scx/blob/9301d7d78a8b118a78fc0ed0101b5a7e7bf2bbcf/rust/scx_utils/src/cpumask.rs#L22

is ran in vng with 8 cpus (-c8):
https://github.com/sched-ext/scx/blob/eaa636dfcca35e23d84ac0af846bce32f99b03a9/.github/workflows/caching-build.yml#L137

the following failure reliably occurs:
https://github.com/likewhatevs/scx/actions/runs/10746355875/job/29807130862#step:10:910

and changing the cpumask string from `0xf0` to `0x0f` as such:
```
//!     let from_str_mask = Cpumask::from_str(&String::from("0x0f"));
```

fixes the issue:
https://github.com/likewhatevs/scx/actions/runs/10746447550/job/29807366108

I will open a PR "fixing" builds by changing  `0xf0` to `0x0f` in that doctest, but, unless I am misunderstanding something, `0xf0` erroring out like that while the number of cpus is set to 8 is a bug.

arighi:
@likewhatevs bah... my bad, I told you the wrong vng option, the right one to set the num of CPU is `-p`, not `-c`. 
 Pretty funny that I'm the one not knowing about the vng options. :sweat_smile:

PR #627 should fix it without having to change the test.

likewhatevs:
ahh lol, just rechecked help, yep that's it, thanks!

PR #626, opened 2024-09-06
scx_utils: fix failing doctest

likewhatevs:
GH-625 describes how this fixes a failing doctest (i.e. avoid triggering bug).

Fix this so CI signal can be acceptable until this bug (or the test, possibly) is properly fixed.


arighi:
Fixed by #627, we don't have to change the test, so closing this one. Thanks for pointing this out!

PR #627, opened 2024-09-07
ci: fix vng command to set the right amount of CPUs

arighi:
The right options to set the amount of CPUs in vng is `-p`, not `-c`. Change the command to use the long format `--cpu` and `--memory` for better clarity.

This fixes issue #625.


arighi:
Merging this right away as it clearly seems to fix the CI failure with the cpumask unit test.

PR #628, opened 2024-09-07
release: Add cargo-publish.py

htejun:
Let's make cargo publishing a bit less painful.


arighi (approved):
That's really useful, thanks!

Issue #610, opened 2024-09-05
scx_rusty: Don't use PID to identify tasks between userspace and BPF


htejun:
scx_rusty implements most of load balancing logic in userspace and uses PID to identify tasks between userspace and BPF. However, a task's PID can change when a non-leader thread exec(2)s as it swaps its PID with the group leader before killing all other threads and exec(2)ing. This PID change can lead to various malfunctions in scx_rusty including occasional crashes. #437 is likely one of the symptoms too.

It'd be useful to create a repro case which can demonstrate the problem reliably and fix it by changing the key to something actually persistent and unique - e.g. the task pointer.

PR #630, opened 2024-09-07
scx_rusty: handle gracefully tasks without context in dom_xfer_task()

arighi:
Make dom_xfer_task() static to satisfy the verifier regarding the use of the task_ctx pointer.

Then pass the task context directly as a pointer to dom_xfer_task() to eliminates the need for task and task context lookups from the PID, which can fail if the task context lookup from the PID is unsuccessful (due to racy behavior).

This seems to fix failures with scx_rusty in our CI.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -389,31 +389,18 @@ static u64 dom_min_vruntime(struct dom_ctx *domc)
 	return READ_ONCE(domc->min_vruntime);
 }
 
-int dom_xfer_task(pid_t pid, u32 new_dom_id, u64 now)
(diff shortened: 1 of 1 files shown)


arighi:
Hm.. nevermind, it's still failing, will take a look later, I may have done something silly.

htejun:
The unreliable behavior is probably from #610.

arighi:
Right, the "pid" logic introduces a bunch of potential issues and the fix is not so trivial. I opened this, because our CI seems to fail reliably because of this, but I wasn't able to reproduce it locally (for some reasons it seems to happen only inside the github action...). I'll investigate more to see if we can find a better reproducer (probably after LPC), let's put this on hold for now.

htejun:
Closing for now. Please feel free to reopen if necessary.

PR #631, opened 2024-09-07
have stress tests retry 3 times

likewhatevs:
I saw the stress test being flaky (failing sometimes and passing others when I was touching unrelated things) the other day.

Not sure if we want to allow this to pass if it passes 1 of X times (w/ current X being 3).

Opening this PR in the event this is desirable.

Here is the relevant part of a CI job using this config:

https://github.com/likewhatevs/scx/actions/runs/10754556307/job/29825295907#step:14:1715

### DIFF
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -101,7 +101,11 @@ jobs:
       # Test schedulers
       - run: meson compile -C build test_sched
       # Stress schedulers
-      - run: meson compile -C build stress_tests
+      - uses: cytopia/shell-command-retry-action@v0.1.2
+        name: stress test
+        with:
+          retries: 3
+          command: meson compile -C build stress_tests
       - run: meson compile -C build veristat
 
   rust-test:


htejun (approved):

arighi (commented):
Hm... this could hide potential issues IIUC... I'd rather do the opposite: retry 3 times and pass if all 3 are ok. In general we don't have a lot of flaky tests, usually when we have failures it means that there's actually something that needs to be fixed.

likewhatevs:
> Hm... this could hide potential issues IIUC... I'd rather do the opposite: retry 3 times and pass if all 3 are ok. In general we don't have a lot of flaky tests, usually when we have failures it means that there's actually something that needs to be fixed.

> Hm... this could hide potential issues IIUC...

Yes

> In general we don't have a lot of flaky tests, usually when we have failures it means that there's actually something that needs to be fixed.

In that case yeah, this isn't a thing we'd want to do. I noticed flakiness in the stress tests overall when I wrote this but didn't know if that was status quo or indicative of a idk now problem (after seeing the split tests from my other pr, I think it's indicative of an issue in rusty and that this shouldn't be merged).

arighi:
Right, scx_rusty at the moment is showing an issue that needs to be fixed (looking at it...). And if we can't figure out a fix quickly and it gets too spammy we can temporarily exclude it from the tests.

Issue #462, opened 2024-07-31
Update the CI workflows to run on only schedulers relevant for the the change


hodgesds:
## Why
This will make everyone's lives a little better working on sched_ext by having faster CI signals.

## What
Update the [existing github workflow](https://github.com/sched-ext/scx/blob/main/.github/workflows/build-scheds.yml) to aware of changes that it is running on. For example a change to `scheds/rust/scx_bpfland` should not trigger any of the schedulers written in C (`scheds/c`) nor any other rust schedulers (ex: `scheds/rust/scx_lavd`). Changes to common/build code should still test all schedulers (ex: `meson-scripts`).

arighi:
Sounds like a really nice thing to me, I was actually planning to look at this at some point. I should be able to find some time this week.

likewhatevs:
I think GH-632 meets idk the intent of this well.

It splits out test jobs etc. but always runs them.

GH actions don't (to the best of my knowledge) have a "not ran state" that is able to accept non-predefined data.

That's not to say one couldn't hack it in (nor would that be particularly tricky), but like, that paired with the inability to compose steps really well (actions afaik can only run shell cmds) would mean that conditionally running jobs like this would probably need both:
a) some kind of artifact or cache passing between steps to communicate what should be ran.
b) having less helpful pass/fail states (i.e. pass == not ran).
c) an uncomfortable amount of boilerplate (which I'd probably add some script to write, so that's less of an issue).

(b) is the best reason to just run all at once (and they're fast for now).

Imo, how mitosis is handled covers how to deal w/ jobs that continually fail (turn them off by removing them from the list in meson and the lists in the ci config), so maybe that PR meets the need of this issue.

PR #632, opened 2024-09-08
split integration test execution

likewhatevs:
split build and test jobs to reduce ci turnaround time
and make it clear what is failing when something fails.

also add virtiofsd to deps to make test compilation faster
(most test time is compliation) and remove all force 9ps.

also have rust tests on schedulers be ran.

time w/ no caches is 23 minutes, time with caches is 7 minutes:

https://github.com/likewhatevs/scx/actions/runs/10756353007

i think this should address GH-462, it goes as far as the trade off's involved make sense imo

### DIFF
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -25,10 +25,21 @@ jobs:
         with:
           path: |
             linux
-          key: ${{ env.SCHED_EXT_KERNEL_COMMIT }}
+          key: kernel-build-${{ env.SCHED_EXT_KERNEL_COMMIT }}
 
       - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
         uses: ./.github/actions/install-deps-action
+
+      # cache virtiofsd (goes away w/ 24.04)
+      - name: Cache virtiofsd
+        id: cache-virtiofsd
+        uses: actions/cache@v4
+        with:
+          path: |
+            /usr/lib/virtiofsd
+          key: virtiofsd-binary
+      - if: ${{ steps.cache-virtiofsd.outputs.cache-hit != 'true' && steps.cache-kernel.outputs.cache-hit != 'true' }}
+        run: cargo install virtiofsd && sudo cp -a ~/.cargo/bin/virtiofsd /usr/lib/
 
       # cache bzImage alone for rust tests (disk space limit workaround)
       - name: Cache bzImage
@@ -37,7 +48,7 @@ jobs:
         with:
           path: |
             linux/arch/x86/boot/bzImage
-          key: ${{ env.SCHED_EXT_KERNEL_COMMIT }}
+          key: kernel-bzImage-${{ env.SCHED_EXT_KERNEL_COMMIT }}
 
       - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
         name: Clone Kernel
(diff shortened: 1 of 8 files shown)


.github/workflows/caching-build.yml
```diff
@@ -25,19 +25,30 @@ jobs:
...
+      # cache virtiofsd (goes away w/ 24.04)
+      - name: Cache virtiofsd
+        id: cache-virtiofsd
+        uses: actions/cache@v4
+        with:
+          path: |
+            /usr/lib/virtiofsd
+          key: virtiofsd-binary
+      - if: ${{ steps.cache-virtiofsd.outputs.cache-hit != 'true' && steps.cache-kernel.outputs.cache-hit != 'true' }}
+        run: cargo install virtiofsd && sudo cp -a ~/.cargo/bin/virtiofsd /usr/lib/
```
arighi, line 42: Ahh, really nice! I was trying to figure out a way to add virtiofsd, it was easier than expected! :)

.github/workflows/caching-build.yml
```diff
@@ -129,9 +170,55 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
         with:
-          shared-key: ${{ env.SCHED_EXT_KERNEL_COMMIT }}
           workspaces: rust
+          key: ${{ matrix.component }}
+          prefix-key: "1"
+      - run: cargo build  --manifest-path rust/${{ matrix.component }}/Cargo.toml
+      - run: cargo test --manifest-path rust/${{ matrix.component }}/Cargo.toml --no-run
+      - run: vng -v --memory 10G --cpu 8 --force-9p -r linux/arch/x86/boot/bzImage  --net user -- cargo test --manifest-path rust/${{ matrix.component }}/Cargo.toml
```
arighi, line 178: I think we should drop `--force-9p` also from here.

.github/workflows/caching-build.yml
```diff
@@ -129,9 +170,55 @@ jobs:
...
+        with:
+          workspaces: scheds/rust
+          key: ${{ matrix.scheduler }}
           prefix-key: "1"
-      - run: cargo build  --manifest-path rust/Cargo.toml
-      - run: cargo test --manifest-path rust/Cargo.toml --no-run
-      - run: vng -v --memory 10G --cpu 8 --force-9p -r linux/arch/x86/boot/bzImage  --net user -- cargo test --manifest-path rust/Cargo.toml
+      - run: cargo build  --manifest-path scheds/rust/${{ matrix.scheduler }}/Cargo.toml
+      - run: cargo test --manifest-path scheds/rust/${{ matrix.scheduler }}/Cargo.toml --no-run
+      - run: vng -v --memory 10G --cpu 8 --force-9p -r linux/arch/x86/boot/bzImage  --net user -- cargo test --manifest-path scheds/rust/${{ matrix.scheduler }}/Cargo.toml
```
arighi, line 223: Ditto, now that we have virtiofsd we don't need `--force-9p`.

arighi (approved):
Left a few comments, but overall I really like this, especially the virtiofsd part! Thanks for working on this.

I'm not entirely familiar with all the specifics of the matrix strategy on GitHub, but it seems really cool. I think we should go ahead and merge this to give it a try.

.github/workflows/caching-build.yml
likewhatevs: lol ofc I missed the instance of that in the file I was editing the most, thx

likewhatevs: ahh forgot rg skips hidden directories and `.github` is one, thx.

arighi (approved):

PR #633, opened 2024-09-08
enable docs generation and upload

likewhatevs:
this pr configures documentation generation and publishing to https://sched_ext.github.io/scx once it is merged and enabled.

this would look like this page: https://likewhatevs.github.io/scx/

commits to master would trigger the generation of new docs.

the config needed to enable this beyond that in this PR is the following:
Safari 2024-09-08 04 22 34

the following page details how to have comments be such that pretty docs are rendered: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 libbpf/
 *.gitignore
 PKGBUILD
+target
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -222,3 +222,41 @@ jobs:
       - run: cargo test --manifest-path scheds/rust/${{ matrix.scheduler }}/Cargo.toml --no-run
       - run: vng -v --memory 10G --cpu 8 -r linux/arch/x86/boot/bzImage  --net user -- cargo test --manifest-path scheds/rust/${{ matrix.scheduler }}/Cargo.toml
 
+  pages:
+    runs-on: ubuntu-22.04
+    if: github.ref == 'refs/heads/main'
+    permissions:
+      contents: write  # To push a branch
+      pages: write  # To push to a GitHub Pages site
+      id-token: write # To update the deployment status
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/install-deps-action
+      - name: Build Book and Linux Docs
+        run: |
+          curl https://sh.rustup.rs -sSf | sh -s -- -y
+          rustup install nightly
+          RUSTDOCFLAGS="--enable-index-page -Zunstable-options" ~/.cargo/bin/cargo +nightly doc --workspace --no-deps --bins --lib --examples --document-private-items --all-features
+          sudo apt update
+          sudo apt install build-essential graphviz sphinx-doc python3-sphinx-rtd-theme texlive-latex-recommended python3-yaml -y
+          cargo install htmlq
+          git clone --single-branch -b for-next --depth 1 https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git linux
+          cd linux
+          make DOCS_THEME=sphinx_rtd_theme SPHINXDIRS=scheduler htmldocs
+          make DOCS_THEME=sphinx_rtd_theme SPHINXDIRS=bpf htmldocs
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 libbpf/
 *.gitignore
 PKGBUILD
+target
+Cargo.lock
```
hodgesds, line 5: Do we want these for reproducible builds?

arighi:
Looks good overall, can you rebase so that it only include the change about the doc without the split build commit? Thanks!

.gitignore
arighi: Yeah, I think we actually want to keep track of `Cargo.lock` changes in the commits.

likewhatevs: ahh ok, will add that in, I was trying to avoid unintended consequences of there being a top level Cargo.lock (but maybe there are none).

likewhatevs:
rebased, I also updated the args to generate docs because I noticed the docs for scx lib packages were missing.

likewhatevs:
undrafting pr -- needed to use an incognito window to view updated docs w/ libs in them: https://likewhatevs.github.io/scx/ 
CleanShot 2024-09-08 at 11 52 46@2x

hodgesds (approved):
LGTM

Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[workspace]
```
htejun, line 1: It isn't great that we end up with overlapping workspace definitions and overlapping sets of `Cargo.lock` files. I suppose it wasn't possible to make it work with the existing workspaces under `rust/` and `scheds/rust/`? Hmm... there are no strong reasons to use separate workspaces for `rust/` and `scheds/rust/`, I don't think. If docs needs a single unified workspace, can you please make the build also use the top level workspace?

htejun (changes requested):
This is great but I think we should avoid defining overlapping cargo workspaces whether that's by making docs work with the two existing workspaces or making build use the top level workspace.

Also, not directly related to this but, in terms of documentation, what'd be really useful is generating documentation from kernel source files. Kernel does have its own document generation but my experience hasn't been that great and we need to generate documentation for a very small subset, so it'd be great if we can do targeted documentation generation and link from here.

likewhatevs:
> Also, not directly related to this but, in terms of documentation, what'd be really useful is generating documentation from kernel source files. Kernel does have its own document generation but my experience hasn't been that great and we need to generate documentation for a very small subset, so it'd be great if we can do targeted documentation generation and link from here.

good idea, I'll look into that, thanks!

Cargo.toml
likewhatevs: > I don't think. If docs needs a single unified workspace, can you please make the build also use the top level workspace?

That works, and I think it also pushes things in a good direction overall (i.e. closer to the build being driven by cargo, and in a way that will "just work" w/ more folk's IDE's and rust tooling overall).

htejun:
#635 is more comprehensive conversion to single workspace. This drops the nested `meson.build` files too and updates install script accordingly. If that looks okay, can you please rebase on top?

likewhatevs:
> #635 is more comprehensive conversion to single workspace. This drops the nested `meson.build` files too and updates install script accordingly. If that looks okay, can you please rebase on top?

that looks great, much simpler, will do!

likewhatevs:
https://likewhatevs.github.io/scx/ links to the latest generated docs. There are links to kernel documentation for scheduler and bpf there. Not sure if those are the kernel docs we want, but given the names of things to add I can add them pretty easily.

htejun:
> https://likewhatevs.github.io/scx/ links to the latest generated docs. There are links to kernel documentation for scheduler and bpf there. Not sure if those are the kernel docs we want, but given the names of things to add I can add them pretty easily.

That's fantastic. Thank you. I think we just need https://likewhatevs.github.io/scx/scheduler/sched-ext.html; however, it should include docs generated from function comments in `kernel/sched/ext.c` and some header files and a reference to `man bpf-helpers`. There were some notations to include some function comments in the generated docs,. I'll look into that tomorrow.

htejun (approved):

PR #634, opened 2024-09-08
scx_bpfland: Documentation consistency fix

anh0516:


arighi:
Looks good. I see the PR contains some merges from main, probably they're dropped if we merge, but just to make sure, can you rebase and include only the last commit? Thanks!

anh0516:
I reverted my branch so it only contains the one commit. Is there a way for me to pull in changes since then without creating extraneous commits? I can't find one. Easy solution is to just wait until this is merged.

arighi (approved):
Looks good now, thanks.

PR #635, opened 2024-09-09
build: Use a single top-level rust workspace

htejun:
Rust build was using two separate workspaces - rust/ and scheds/rust. There's no reason to separate them and it makes doc generation tricky. Use single top level workspace so that we can drive all rust building from cargo.


likewhatevs:
I think `rust/meson.build` needs to be deleted also btw (these changes break it and it isn't being called).

arighi (approved):

PR #636, opened 2024-09-09
bpfland: loop optimization and logical redundancy elimination

firelzrd:
Trying to reduce some of the logical redundancy and code amount, while maintaining readability and simpleness.
cpumasks bound to respective cache levels {p, l2, l3}, dsq_ids for each of the global DSQs {preferred, shared} are now arrays, and can be iterated using loop syntax.

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -13,19 +13,18 @@
 #include <asm-generic/errno.h>
 #include "user_exit_info.h"
 
-#define PF_WQ_WORKER			0x00000020	/* I'm a workqueue worker */
-#define PF_KTHREAD			0x00200000	/* I am a kernel thread */
-#define PF_EXITING			0x00000004
-#define CLOCK_MONOTONIC			1
+#define PF_WQ_WORKER 0x00000020 /* I'm a workqueue worker */
+#define PF_KTHREAD   0x00200000 /* I am a kernel thread */
+#define PF_EXITING   0x00000004
+#define CLOCK_MONOTONIC 1
 
 /*
  * Earlier versions of clang/pahole lost upper 32bits in 64bit enums which can
  * lead to really confusing misbehaviors. Let's trigger a build failure.
  */
-static inline void ___vmlinux_h_sanity_check___(void)
-{
+static inline void ___vmlinux_h_sanity_check___(void) {
 	_Static_assert(SCX_DSQ_FLAG_BUILTIN,
-		       "bpftool generated vmlinux.h is missing high bits for 64bit enums, upgrade clang and pahole");
+		"bpftool generated vmlinux.h is missing high bits for 64bit enums, upgrade clang and pahole");
 }
 
 s32 scx_bpf_create_dsq(u64 dsq_id, s32 node) __ksym;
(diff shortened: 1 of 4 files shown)


scheds/include/scx/common.bpf.h
```diff
@@ -389,13 +388,14 @@ static inline u32 log2_u32(u32 v)
...
+	if (hi)
+		return log2_u32(hi) + 32 + 1;
+	else
+		return log2_u32(v) + 1;
 }
 
 #include "compat.bpf.h"
 
 #endif	/* __SCX_COMMON_BPF_H */
+
```
arighi, line 401: Can you include only the changes to `scheds/rust/scx_bpfland`?

scheds/include/scx/common.h
```diff
@@ -60,13 +59,13 @@ typedef int64_t s64;
...
+#define RESIZE_ARRAY(__skel, elfsec, arr, n) \
+	do { \
+		size_t __sz; \
+		bpf_map__set_value_size((__skel)->maps.elfsec##_##arr, \
+			sizeof((__skel)->elfsec##_##arr->arr[0]) * (n)); \
+		(__skel)->elfsec##_##arr = \
+			bpf_map__initial_value((__skel)->maps.elfsec##_##arr, &__sz); \
 	} while (0)
 
 #include "user_exit_info.h"
```
arighi, line 71: Ditto. Better to restrict the changes only to the bpfland code.

scheds/include/scx/compat.bpf.h
```diff
@@ -7,22 +7,22 @@
...
-		__VA_ARGS__,							\
-	};
+#define SCX_OPS_DEFINE(__name, ...) \
+    SEC(".struct_ops.link") \
+    struct sched_ext_ops __name = { \
+        __VA_ARGS__, \
+    };
 
-#endif	/* __SCX_COMPAT_BPF_H */
+#endif /* __SCX_COMPAT_BPF_H */
```
arighi, line 28: Ditto.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -8,23 +8,29 @@
...
+/* Global DSQ level iterator */
+enum {qidx_prio = 0, qidx_shared = 1};
+#define for_each_dsq_level(level) \
+	for(int level = 0; level < 2; level++)
+/* Cache layer iterator */
+enum {cidx_l3 = 0, cidx_l2 = 1, cidx_p = 2};
+int cla_disp_name(int index) {return 3 - index;}
+int codr_p_l2_l3 = 0xC6, codr_l2_l3_p = 0xE1, codr_l2_l3 = 0x31;
+#define for_each_cache_layer(layer, order) \
+	for (int layer = order; (layer & 3) != 3; layer >>= 2)
```
arighi, line 28: I'm having a bit of a hard time to understand the logic here, especially the codr_p_* values... can you add a comment to explain the logic here, to make the code easier to read?

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -696,101 +671,54 @@ static s32 pick_idle_cpu(struct task_struct *p,
...
-		}
-
-		/*
-		 * Search for any other full-idle core in the primary domain.
-		 */
-		cpu = bpf_cpumask_any_and_distribute(cast_mask(p_mask), idle_smtmask);
-		if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
-		    scx_bpf_test_and_clear_cpu_idle(cpu))
-			goto out_put_cpumask;
+#define TRY_IDLE_MASK(layer, idle_mask_type) do { \
```
arighi, line 685: Adding `#define` inside the code makes it a little bit hard to read, maybe we can move these to the top, outside of the function definition.

arighi (changes requested):
I left a few comments, mostly about coding style. In particular, it'd be great if you could include only the changes to the scx_bpfland code.

Some of the changes don't match the rest of the coding style or make the code a little bit more difficult to read. Maybe we can try with smaller changes? Like the cache level iterator or the prio/shared DSQ iterator both look like a good refactoring idea.

Thanks!

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
firelzrd: Yes, sure.
I'll include the comments.
(Currently I'm having a hard time shooting the bug that's preventing the scx from starting, probably due my ignorance in BPF virtual machine's specification.)

firelzrd:
Comments were added.
Thank you for the suggestion.
I am still tackling the program won't-start issue.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
firelzrd: The code itself is in the best spot (as it's supposed to be used there, and never more), but you are absolutely right that the macro looks nasty.
pre-processor directives can be indented, so maybe it can be indented like the other code around to reside better.
If there's any better idea, please let me know, so I can get rid of it.

scheds/include/scx/common.bpf.h
firelzrd: Yes. I rolled back most of the cosmetic stuff within and outside of main.bpf.c.
Thank you for the suggestion.

firelzrd:
I see that the eBPF verifier for some reason turns a pointer into a scalar value while copy operations from an array to another.
And operations like `bpf_cpumask_and` and others starts to fail, since the pointer is no longer a pointer, even if it's stored in `struct bpf_cpumask*`.
That's what I found out so far. And this makes me crazy.
Let me think about how to sort this out.

firelzrd:
It seems that currently there's no good way to make it consider as a pointer instead of a scalar when using an array of pointer.
So I'm closing this PR for now.
Thanks for your time!

Issue #639, opened 2024-09-10
CI: move to ubuntu-24.04


arighi:
Ubuntu 24.04 is now available for all GitHub Actions, we should move to this new image, get rid of the custom clang and use packages from the distro.

I've done some experiments in https://github.com/arighi/scx/tree/ci-ubuntu-24.04, but I'm having some troubles with `.github/actions/install-deps-action/action.yml`. I would like to keep installing virtme-ng via pip for now (because the vng from 24.04 is buggy and it takes a long time to update the package in Ubuntu - working on it), but I haven't figured out how to specify `runs-on: ubuntu-24.04` in the `install-deps` action.

Maybe @likewhatevs could help us a bit here...

likewhatevs:
ahh so runs-on applies to the entire job (and has to be set on the job).

actions are (err, in my view) pretty much just a fancy way to write/reuse a shell script across commands. they need to be called from a job section or a workflow yaml to be used.

will give it a try and open a PR if things go smoothly

PR #641, opened 2024-09-10
migrate ci vm to 24.04

likewhatevs:
migrate ci vm to ubuntu 24.04

things checked:
* ci works with all caches deleted: https://github.com/likewhatevs/scx/actions/runs/10788376061
* ci works with caches present: https://github.com/likewhatevs/scx/actions/runs/10788172911 
* generated documentation still works: https://likewhatevs.github.io/scx/

Closes: #639

### DIFF
--- a/.github/actions/install-deps-action/action.yml
+++ b/.github/actions/install-deps-action/action.yml
@@ -16,30 +16,11 @@ runs:
     ### DOWNLOAD AND INSTALL DEPENDENCIES ###
 
     # Download dependencies packaged by Ubuntu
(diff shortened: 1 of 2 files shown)


.github/actions/install-deps-action/action.yml
```diff
@@ -16,30 +16,11 @@ runs:
...
-    - run: sudo ln -sf /usr/bin/llvm-strip-17 /usr/bin/llvm-strip
-      shell: bash
-    # meson
-    - run: pip install meson
-      shell: bash
-
-    # Install virtme-ng
-    - run: pip install virtme-ng
+    # meson and virtme-ng
+    - run: pip3 install meson virtme-ng --break-system-packages
```
arighi, line 23: Should we apt install meson? Is there any reason to have it installed via pip?

I'd like to do this for virtme-ng as well, but at the moment the version from pip is much more stable, once it's fixed in Ubuntu we can also move virtme-ng to the list of packages installed via apt install.

arighi (approved):
Left a little comment about meson, but LGTM, thanks!

.github/actions/install-deps-action/action.yml
likewhatevs: > Should we apt install meson? Is there any reason to have it installed via pip?

None I'm aware of provided the build passes and works. 24.04 has 1.3.2 and pip has 1.5.1.

PR #644, opened 2024-09-10
scx_layered: Pass layer spec for core growth algo

hodgesds:
Pass in the layer spec when determining the layer core growth algo. This should make it easier to implement layer growth algos that are spec specific.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -1090,6 +1090,7 @@ impl CpuPool {
 }
 
 fn layer_core_order(
+    spec: &LayerSpec,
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/main.rs
```diff
@@ -1593,15 +1595,15 @@ impl<'a, 'b> Scheduler<'a, 'b> {
         skel.maps.rodata_data.nr_llcs = 0;
 
         for node in topo.nodes() {
-            info!(
+            debug!(
```
hodgesds, line 1598: drive by log cleanup

likewhatevs: i am a fan of the boy scout rule

PR #662, opened 2024-09-19
scx_rustland_core: prevent CI failures

arighi:
A couple of changes that should finally prevent all the CI failures.

With this applied, the schedulers based on scx_rustland_core will always dispatch kthreads directly from the BPF component. This should be a safer option for now, allowing to prevent potential stalls also under heavy stress test conditions.

In the future, we will provide a better API to enforce stricter selection criteria for tasks permitted to bypass user-space scheduling.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf.rs
+++ b/rust/scx_rustland_core/assets/bpf.rs
@@ -518,7 +518,7 @@ impl<'cb> BpfScheduler<'cb> {
             LIBBPF_STOP => {
                 // A valid task is received, convert data to a proper task struct.
                 let task = unsafe { EnqueuedMessage::from_bytes(&BUF.0).to_queued_task() };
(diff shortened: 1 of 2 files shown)


arighi:
Merging to see if we can reduce the amount of failures with CI.

likewhatevs:
I think we want those failures.

I haven't looked a to at how stress-ng is being called in the tests, but like, stress-ng is stalling which doesn't make the tests fail like kernel threads stalling does, but the tests should still fail (they're passing with stalls atm):

https://github.com/sched-ext/scx/actions/runs/10936241301/job/30359538896#step:16:363

arighi:
meh ok... it's still failing then, I really don't understand what's going on with the stress-ng fork stressor, it really breaks rustland badly... I will investigate more.

PR #663, opened 2024-09-19
scx_bpfland: Remove the usage of cast_mask in bpfland_enqueue

vax-r:
## Summary
The usage of `cast_mask()` within `bpfland_enqueue` aims to cast the type of "p->cpus_ptr" from "struct bpf_cpumask *" to "const struct cpumask *". However, the type of "p->cpus_ptr" is already "const cpumask_t *" aka "const struct cpumask *", so no conversion is needed.

Moreover, passing a value of type "struct cpumask *" into "struct bpf_cpumask *" leads to compiling error.


arighi (changes requested):
Looks good to me, but can you drop the changes to Cargo.lock, so that we can only have the cast_mask change applied with this PR? Thanks!

vax-r:
@arighi -
No problem ! I've made the requested change, please take a look while you're available.

arighi (approved):

PR #664, opened 2024-09-19
scx_layered: Refactor match_layer() and implement helper function to access cpumask within bpf_cpumask

vax-r:
## Summary
There're two main changes in this PR, describing in sequence in the following content.

First, refactor `match_layer()` to prevent the compilation error caused by the uninitialization of the variable `nr_match_ors`.

Secondly, implement a helper function to access cpumask within `struct bpf_cpumask`.

Because the fact that `struct bpf_cpumask` isn't the same as `struct cpumask` aka `cpumask_t`, conversions between the two types can't be done directly in a safe way.

The actual cpumask within a `struct bpf_cpumask` is embedded within it as a structure member. To access the actual cpumask, a helper function `bpf_cpumask_access()` is implemented and used to wrap `nodec->cpumask` and `cachec->cpumask` before they're passed into `bpf_cpumask_test_cpu()`.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1066,11 +1066,14 @@ int match_layer(u32 layer_id, pid_t pid, const char *cgrp_path)
 	if (!p)
 		return -EINVAL;
 
-	if (layer_id >= nr_layers || nr_match_ors > MAX_LAYER_MATCH_ORS)
+	if (layer_id >= nr_layers)
 		goto err;
 
 	layer = &layers[layer_id];
 	nr_match_ors = layer->nr_match_ors;
+
+	if (nr_match_ors > MAX_LAYER_MATCH_ORS)
+		goto err;
 
 	bpf_for(or_idx, 0, nr_match_ors) {
 		struct layer_match_ands *ands;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -254,6 +254,14 @@ static struct cpumask *lookup_layer_cpumask(int idx)
 	}
 }
 
+/*
+ * Access a cpumask in read-only mode (typically to check bits).
+ */
+static const struct cpumask *cast_mask(struct bpf_cpumask *mask)
+{
+	return (const struct cpumask *)mask;
+}
+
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -254,6 +254,14 @@ static struct cpumask *lookup_layer_cpumask(int idx)
 	}
 }
 
+/* Helper function to access the actual cpumask which is embedded
+ * in struct bpf_cpumask.
+ */
+static struct cpumask *bpf_cpumask_access(struct bpf_cpumask *mask)
+{
+	return &mask->cpumask;
```
hodgesds, line 262: Might want to take a look at the [same ``function](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/bpf/progs/cpumask_common.h#n63) that is used for tests. The function can safely cast between the types ([docs](https://www.kernel.org/doc/html/v6.5/bpf/cpumasks.html#struct-bpf-cpumask)), so no need to reference the inner mask. It's probably worth making it return a `const struct cpumask *` as that is what many of the [kfuncs](https://www.kernel.org/doc/html/v6.5/bpf/cpumasks.html#operations-between-cpumasks) require anyways.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1066,12 +1074,15 @@ int match_layer(u32 layer_id, pid_t pid, const char *cgrp_path)
...
 		return -EINVAL;
 
-	if (layer_id >= nr_layers || nr_match_ors > MAX_LAYER_MATCH_ORS)
+	if (layer_id >= nr_layers)
 		goto err;
 
 	layer = &layers[layer_id];
 	nr_match_ors = layer->nr_match_ors;
 
+	if (nr_match_ors > MAX_LAYER_MATCH_ORS)
```
hodgesds, line 1083: Nice catch!

hodgesds (changes requested):

scheds/rust/scx_layered/src/bpf/main.bpf.c
vax-r: Oh thanks for the reminder, I've made the requested changes. Thanks very much for the detailed and patient reply, it helps me gain more knowledge !

hodgesds (approved):

PR #666, opened 2024-09-19
scx_layered: Add topology aware preemption

### LINKED ISSUE #659: scx_layered: Add topology aware preemption
The current logic for [preemption](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/main.bpf.c#L783) in `scx_layered` is to iterate over all CPUs with approximate round robin approach. It might better to first try to preempt within the local LLC and then the local NUMA node before considering other CPUs. This pattern could be useful elsewhere for making topology aware iteration. This may require additional data structures to efficiently implement, see the [`cpu_llc_id_map`](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/main.bpf.c#L21) as a reference.

hodgesds:
Add topology aware preemption, which changes the order of preemption to first attempt to preempt CPUs in the local LLC, local NUMA node and finally across all NUMA nodes. This could probably be made a little more efficient in the use of the bpf cpumasks, but should work for now. The old logic is still available behind the `-t` (disable topology) flag. Resolves #659


rust/Cargo.lock
JakeHillion, line 1: Is there a reason we weren't committing this before? Imo it should be committed, just surprising to see it now.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -88,19 +90,21 @@ struct cpu_ctx {
 	u64			lstats[MAX_LAYERS][NR_LSTATS];
 	u64			ran_current_for;
 	u32			layer_idx;
-	u32			node_idx;
 	u32			cache_idx;
+	u32			node_idx;
```
JakeHillion, line 94: Did these swap for a reason?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -780,16 +782,121 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
 			return;
 	}
 
-	bpf_for(idx, 0, nr_possible_cpus) {
-		s32 cand = (preempt_cursor + idx) % nr_possible_cpus;
+	if (!disable_topology) {
+		if (!(cachec = lookup_cache_ctx(cctx->cache_idx)) ||
+		    !(nodec = lookup_node_ctx(cctx->node_idx)))
+			return;
```
JakeHillion, line 788: nit: would it be clearer if we split these into two early returns?
```suggestion
                if (!(cachec = lookup_cache_ctx(cctx->cache_idx)))
                        return;
		if (!(nodec = lookup_node_ctx(cctx->node_idx)))
			return;
```

JakeHillion (approved):
This looks good to me. I wonder if it's worth changing the stuff in the new `if (!disable_topology)` into two separate functions? Might make things easier to follow.

rust/Cargo.lock
hodgesds: I'm not sure.... I think @likewhatevs recently removed the lock files from .gitignore so we could have reproducible builds. So maybe it hasn't been committed yet?

scheds/rust/scx_layered/src/bpf/intf.h
hodgesds: was thinking going from the least specific to the most specific for ordering, but either way is fine

rust/Cargo.lock
likewhatevs: I did remove the git ignore, but it is kinda weird one would be there at all tbh. Like a top-level Cargo.lock I'd get (workspace level), or a project level Cargo.lock I'd also get (project level).

In the middle seems off to me.

hodgesds: I think I know what happened, I think I created this branch before the lock files and before some of the refactoring of the builds. Then after rebasing the old lock file was still around and I added it on accident, so should be safe to delete.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -788,16 +790,112 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
...
+		}
+
+		bpf_cpumask_copy(topo_cpus, (const struct cpumask*)cachec->cpumask);
+
+		/*
+		 * First try preempting in the local LLC
+		 */
+		bpf_for(idx, 0, cachec->nr_cpus) {
+			s32 preempt_cpu = bpf_cpumask_any_distribute((const struct cpumask*)topo_cpus);
```
htejun: Does this reliably try out the CPUs when looped upon?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -788,16 +790,112 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
...
+			bpf_cpumask_release(attempted);
+			bpf_cpumask_release(topo_cpus);
+			return;
+		}
+		bpf_cpumask_copy(topo_cpus, (const struct cpumask*)nodec->cpumask);
+		bpf_cpumask_xor(topo_cpus,
+				(const struct cpumask*)attempted,
+				(const struct cpumask*)topo_cpus);
+
+		bpf_for(idx, 0, nodec->nr_cpus) {
```
htejun, line 844: It'd be cool if this can be factored into a helper function.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -788,16 +790,112 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
...
+			}
+			bpf_cpumask_clear_cpu(preempt_cpu, topo_cpus);
+			bpf_cpumask_set_cpu(preempt_cpu, attempted);
+			if (bpf_cpumask_empty((const struct cpumask*)topo_cpus))
+				break;
+		}
+
+		/*
+		 * Finally try across nodes
+		 */
```
htejun, line 860: It may not even be helpful to preempt across node boundaries. Might be worthwhile to make this configurable and see whether this helps or not.

rust/Cargo.lock
htejun: Yeah, now, there should be only one lock file at the top level.

htejun (approved):

hodgesds:
Refactored preemption int a separate function and cleaned up a lot of the repetitive casting with the helper that was added in #664. The verifier has a hard time with tracking `bpf_cpumask`s, especially when trying to use `goto`s, so there is a little copy/pasta.

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: Pulled out the preemption into a separate function, but yeah we should have some different iterator abstractions across topologies.

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: Added in a trace and it seems to work.

htejun (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: @htejun I check this again and the kfunc is calling [`cpumask_first`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/cpumask.h#n253). Probably need to fix this to make it more fair.

PR #668, opened 2024-09-20
enable ide's etc. to work on the bpf.c files

likewhatevs:
this makes it so that clangd and ide tools which use clangd can work on the bpf.c code.

nothing should actually be changed outside of that ide/editor environment, all the changes are ifdef'ed on LSP which is set in the added .clangd file.

image

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,2 @@
+CompileFlags:
+  Add: [-DLSP, -Wno-missing-declarations, -Wno-typedef-redefinition, -Wno-ignored-attributes, -xc]
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -27,8 +27,15 @@
  * This software may be used and distributed according to the terms of the
  * GNU General Public License version 2.
  */
+#ifdef LSP
+#define __bpf__
+#include "../../../../scheds/include/scx/common.bpf.h"
+#include "intf.h"
+#else
 #include <scx/common.bpf.h>
 #include "intf.h"
+#endif
+
 
 char _license[] SEC("license") = "GPL";
 
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -30,12 +30,11 @@
 #ifdef LSP
 #define __bpf__
 #include "../../../../scheds/include/scx/common.bpf.h"
-#include "intf.h"
 #else
 #include <scx/common.bpf.h>
-#include "intf.h"
 #endif
 
+#include "intf.h"
 
 char _license[] SEC("license") = "GPL";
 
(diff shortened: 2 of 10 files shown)


likewhatevs:
I also limited edits to those needed to make this work for mitosis, rusty and layered; but this would work with all of them (tested locally).

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -27,8 +27,15 @@
  * This software may be used and distributed according to the terms of the
  * GNU General Public License version 2.
  */
+#ifdef LSP
+#define __bpf__
+#include "../../../../scheds/include/scx/common.bpf.h"
+#include "intf.h"
```
hodgesds, line 33: nit: include `intf.h` after the ifdef

.clangd
```diff
@@ -0,0 +1,2 @@
+CompileFlags:
+  Add: [-DLSP, -Wno-missing-declarations, -Wno-typedef-redefinition, -Wno-ignored-attributes, -xc]
```
hodgesds, line 2: maybe add the clang format as well ([kernel docs](https://www.kernel.org/doc/html/latest//dev-tools/clang-format.html))?

rust/scx_rustland_core/assets/bpf/main.bpf.c
likewhatevs: ahh yeah that makes sense lol

.clangd
likewhatevs: added kernel config

rust/scx_stats/examples/stats_defs.rs.h
```diff
@@ -3,23 +3,24 @@
...
-    pub name: String,
-    #[stat(desc = "an event counter")]
-    pub events: u64,
-    #[stat(desc = "a gauge number")]
-    pub pressure: f64,
+	pub name : String,
+#[stat(desc = "an event counter")]
+		   pub events : u64,
+#[stat(desc = "a gauge number")]
+				pub pressure : f64,
```
htejun, line 13: Oh no, this looks pretty broken.

likewhatevs:
will squash when merge -- learned a new thing today (so much nicer than reset soft and push -f repeatedly lol).

htejun:
Hey, so, I went through some of it and I'm not sure this is a good idea. For rust, enforcing linting makes sense because everyone does that and each of us is likely to use the same tool to format while working on the code anyway. For C, it's just not that way and I don't think the forced uniformity is worth all that much. The language is not that complex. The statements are expected to be shorter and simpler and there are things going on that the compiler can't make sense of.

htejun (changes requested):
Please see the other comment.

likewhatevs:
Reverted this to what it initially was -- make ide's/lsp work nice on bpf code. An added benefit of this is that we should also get hover docs for comments in functions in common.bpf.h, etc.

Will squash merge once approved instead of soft reset and push -f.

htejun (approved):

PR #676, opened 2024-09-23
scx_loader: Add initial automatic scheduler switching via --monitor-no-dbus

MitchellAugustin:
Adds an option --monitor-no-dbus in scx_loader that will monitor CPU
utilization and start scx_lavd when any CPU exceeds 90% for more than 5
seconds. scx_lavd will be terminated if all CPUs are below 90% for
more than 30 seconds. When this flag is specified, scx_loader's
dbus functionality is not utilized.

I plan to add the following additional functionality based on and improving upon this in the future, assuming this is something the sched_ext team is on board with:
* Additional heuristics to determine when to switch to other schedulers
* Improvements to this commit's heuristic, informed by findings from my ongoing benchmarking effort
* Integration of these monitoring heuristics into scx_loader's dbus interface as an "automatic" profile (main reason I'm proposing this as an addition to scx_loader and not as its own module)
* Integrating monitoring functionality into systemd unit


MitchellAugustin:
Please let me know if you'd prefer this to be its own module rather than being part of scx_loader. Thanks!

rust/scx_loader/src/main.rs
```diff
@@ -139,11 +149,73 @@ impl ScxLoader {
...
+// Monitors CPU utilization and enables scx_lavd when utilization of any CPUs is > 90%
+async fn monitor_cpu_util() -> Result<()> {
+    let mut system = System::new_all();
+    let mut running_sched: Option<Child> = None;
+    let mut cpu_above_threshold_since: Option<Instant> = None;
+    let mut cpu_below_threshold_since: Option<Instant> = None;
+    
+    let high_utilization_threshold = 90.0;
+    let low_utilization_threshold_duration = Duration::from_secs(30);
+    let high_utilization_trigger_duration = Duration::from_secs(5);
```
arighi, line 161: I'd be nice to have these as tunables in a config file (maybe `/etc/default/scx` or similar), but it's ok to have them hard-coded for now, so that we can test and see if we like the "automatic" mode.

rust/scx_loader/src/main.rs
```diff
@@ -63,6 +66,13 @@ struct ScxLoader {
     channel: UnboundedSender<ScxMessage>,
 }
 
+#[derive(Parser, Debug)]
+#[clap(author, version, about, long_about = None)]
+struct Args {
+    #[clap(long, short, action)]
+    monitor_no_dbus: bool,
```
arighi, line 73: Maybe we should just call this option `--auto`?

arighi (approved):
I think it's good to have a first implementation of the "auto" mode, just to have something to play with.

We haven't decided any precise plans for the future about scx_loader, except for the following:
 - it will be the main interface with systemd (at some point systemd will start scx_loader, instead of starting the schedulers directly)
 - it will act as the main "control panel" to switch schedulers (via dbus)

We haven't defined yet any precise plan for the "automatic" mode, but we can definitely add some predefined logic like this one for now.

Therefore, looks good to me. Thanks!

ptr1337:
Im a bit confused about the scheduler choosing. LAVD is for example a "Gaming" Scheduler, but most games will never reach 90% CPU Usage on any Core. 

Besides pretty nice addition, thank you!

rust/scx_loader/src/main.rs
```diff
@@ -139,11 +149,73 @@ impl ScxLoader {
...
+        
+        if any_cpu_above_threshold {
+            if cpu_above_threshold_since.is_none() {
+                cpu_above_threshold_since = Some(Instant::now());
+            }
+            
+            if cpu_above_threshold_since.unwrap().elapsed() > high_utilization_trigger_duration {
+                if running_sched.is_none() {
+                    println!("CPU Utilization exceeded 90% for 5 seconds, starting scx_lavd");
+                    running_sched = Some(std::process::Command::new("scx_lavd").spawn().expect("Failed to start scx_lavd"));
```
vnepogodin, line 176: 1. Process spawning from tokio should used here instead. To not mix API and have it consistent
2. I think Enum to string conv should be used here instead of hardcoding

vnepogodin (commented):
All in all switching should be done on the top level using loader API, which should be implemented soon

vnepogodin:
Also would make sense to have the "monitor" functionality on dbus stream subscribe

rust/scx_loader/src/main.rs
```diff
@@ -139,11 +149,73 @@ impl ScxLoader {
...
+        } else {
+            cpu_above_threshold_since = None;
+            
+            if cpu_below_threshold_since.is_none() {
+                cpu_below_threshold_since = Some(Instant::now());
+            }
+            
+            if cpu_below_threshold_since.unwrap().elapsed() > low_utilization_threshold_duration {
+                if let Some(mut running_sched_loc) = running_sched.take() {
+                    println!("CPU utilization dropped below 90% for more than 30 seconds, exiting latency-aware scheduler");
```
vnepogodin, line 190: would be better to have log::info!(for example) instead of println

MitchellAugustin:
> Im a bit confused about the scheduler choosing. LAVD is for example a "Gaming" Scheduler, but most games will never reach 90% CPU Usage on any Core.
> 
> Also, we might make the Scheduler and Utilization configurable.
> 
> Besides pretty nice addition, thank you!

@ptr1337 My thinking in that wasn't really based on the idea that the game itself would reach 90% CPU utilization alone, but rather the fact that LAVD shines when a game is running alongside other processes that are reaching high utilization levels. While it's unlikely that users would be doing things like playing games while compiling the kernel in the background regularly, I could absolutely see lower-end hardware reaching 90% utilization on some CPUs with only a few things running (ex: the game being played, Discord, Chrome, possibly a video playing on a second monitor, etc). In those scenarios, LAVD would be valuable to have running.

MitchellAugustin:
> We haven't decided any precise plans for the future about scx_loader, except for the following:
>
> * it will be the main interface with systemd (at some point systemd will start scx_loader, instead of starting the schedulers directly)
> * it will act as the main "control panel" to switch schedulers (via dbus)

@arighi glad to hear it, this is what I was imagining as well :)

MitchellAugustin:
@vnepogodin @arighi I addressed all of the major concerns - LMK if that looks good.

I agree that this should be tunable and integrated with the dbus stream, but I think those might be good as separate issues. Is there any existing work on bringing up an /etc/default/scx as Andrea mentioned (just to ensure I'm not duplicating work)? If not, I'll add that to my backlog for a future issue.

vnepogodin:
> @vnepogodin @arighi @ptr1337 thanks for the feedback - I addressed all of the major concerns - LMK if that looks good.
> 
> I agree that this should be tunable and integrated with the dbus stream, but I think those might be good as separate issues. Is there any existing work on bringing up an /etc/default/scx as Andrea mentioned (just to ensure I'm not duplicating work)? If not, I'll add that to my backlog for a future issue.

not yet, but had plans to make config for the loader to configure modes for scx schedulers. I think it would make sense to put there instead of `/etc/default/scx`

rust/scx_loader/src/main.rs
```diff
@@ -139,11 +148,73 @@ impl ScxLoader {
...
+        
+        if any_cpu_above_threshold {
+            if cpu_above_threshold_since.is_none() {
+                cpu_above_threshold_since = Some(Instant::now());
+            }
+            
+            if cpu_above_threshold_since.unwrap().elapsed() > high_utilization_trigger_duration {
+                if running_sched.is_none() {
+                    log::info!("CPU Utilization exceeded 90% for 5 seconds, starting scx_lavd");
+                    running_sched = Some(Command::new(get_name_from_scx(&SupportedSched::Lavd)).spawn().expect("Failed to start scx_lavd"));
```
htejun, line 175: While workload dependent scheduler selection may make sense in some situations, I'm skeptical about switching schedulers based on system statistics. These are pretty disruptive operations and it'd probably make much more sense to put efforts towards making schedulers more well rounded.

htejun:
Overall, I'm not sure whether this is a direction we should be taking. Automatic selection of scheduler may make some sense but I'm not sure we want to be switching schedulers dynamically based on system utilization. The selection criteria should be something more persistent. We don't want to have schedulers that are good only when system utilization is >90%, for example.

htejun:
Maybe it can be interesting as an experimentation, but unless others feel strongly, I'd like to steer away from this particular direction.

MitchellAugustin:
@htejun Thanks for the review. Here's my more detailed view on why I like this approach, at least as an _option_ for users:

> While workload dependent scheduler selection may make sense in some situations, I'm skeptical about switching schedulers based on system statistics.

This is understandable. However, I don't see _this_ system statistic as the final/only heuristic for scx_loader's auto mode. I'm currently doing some benchmarking to figure out which workloads perform better with each scx scheduler and the default scheduler, which I hope will help me develop a more robust strategy for inferring the workload from system stats. (thus, my end goal is "workload-dependent scheduler selection," but in a way that is transparent to the end user.)

I think there are likely many different types of workloads that would benefit from _fundamentally_ different scheduling philosophies, but also that many end users running those workloads don't want to necessarily be doing scheduler research themselves. Part of what SCX as a paradigm unlocks in my view is the opportunity to abstract that away from the user. 
As an example, _today_, games running with scx_lavd perform much better than with the default scheduler when CPU utilization elsewhere is high (in my experience), but I also imagine that lavd's scheduling philosophy is not something users would necessarily want to use on _all_ workloads. 

> it'd probably make much more sense to put efforts towards making schedulers more well rounded.

My personal opinion is that there would be value in exploring both paths simultaneously, as it would hopefully help us roll out some of these performance improvements to end users a bit faster since the area of effect would be lower when the experimental schedulers aren't running all the time.

Anyway, let me know what you think. Thanks! :)

htejun:
Workload dependent loading does make sense although it's likely that the schedulers will become more versatile as they develop. e.g. While `scx_lavd` has primarily focusing on gaming, most of that translates well to many interactive use cases and probably even to latency sensitive servers. That said, there definitely are cases that can benefit from a very different scheduling approach (e.g. some servers do better with dumb FIFO scheduling) and it can be nice to provide auto selection mechanism. I'm skeptical system wide utilization is going to be that. As I wrote before, it likely would have to be based on features which are significantly more persistent than that.

Anyways, as an experimentation, I don't see a reason to not do it but let's mark it as such and contain it within the loader.

htejun (approved):

MitchellAugustin:
Noted, thanks everyone!

ptr1337 (approved):

sirlucjan (approved):

PR #678, opened 2024-09-24
scx_rustland_core: fix mm stall

arighi:
Prevent stall bypassing user-space scheduling for tasks that are handling page faults.


hodgesds:
I was trying to see if there was a way to get the faulting state of the task struct, but this seems like a good approach. Do you think tasks being signaled would need a similar handling?

arighi:
@hodgesds good point, I'll run the signal stress-ng stressor.

I was also wondering if there was a more generic approach rather than adding all these kprobes, but I haven't found anything in task_struct that could potentially "hint" when a task is in such "critical" paths....

htejun (approved):

PR #679, opened 2024-09-24
scx_common_bpf: Append cast_mask()

vax-r:
## Summary
Remove cast_mask() function distributed throughout different schedulers and add it in common.bpf.h so every scheduler can reference it once they need to.


hodgesds (approved):
Thanks for cleaning this up!

PR #680, opened 2024-09-24
add 'continue on error' to stress tests in ci jobs

likewhatevs:
I think I see PRs being harder to write because all parts of a CI job are cancelled when one fails.

I think I am also starting to see that we have enough largely disjoint moving pieces that there will often be one that is failing stress tests at any time.

Make CI run all stress tests always to address this.


likewhatevs:
Before:

image

After:

image

PR #683, opened 2024-09-24
scx_layered: Make layered idle CPU selection topology aware

hodgesds:
Make scx_layered idle CPU selection topology aware. It first tries to select an idle CPU from the local LLC before attempting to select an idle CPU from the local node.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -615,6 +615,7 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
 		}
 		bpf_cpumask_copy(pref_idle_cpumask, idle_cpumask);
 		bpf_cpumask_and(pref_idle_cpumask, cache_cpumask, pref_idle_cpumask);
+		bpf_cpumask_and(pref_idle_cpumask, layer_cpumask, pref_idle_cpumask);
 		trace("pick_idle: llc idle_cpumask=%p", pref_idle_cpumask);
 		if ((cpu = pick_idle_cpu_from(cache_cpumask, prev_cpu, pref_idle_cpumask)) >= 0)
 			goto out_put;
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -584,29 +593,66 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
+		 * Try a CPU in the current LLC
+		 */
+		idle_cpumask = scx_bpf_get_idle_cpumask();
+		if (!pref_idle_cpumask || !idle_cpumask) {
+			cpu = -1;
+			goto out_put;
+		}
+		bpf_cpumask_copy(pref_idle_cpumask, idle_cpumask);
+		bpf_cpumask_and(pref_idle_cpumask, cache_cpumask, pref_idle_cpumask);
+		bpf_cpumask_and(pref_idle_cpumask, layer_cpumask, pref_idle_cpumask);
```
htejun, line 618: This is likely okay for now but cpumasks can become rather big and these cpumask manipulations can become expensive. Down the line, it probably would make sense to maintain these masks persistently in each layer.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1030,7 +1076,7 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
 
 		/*
 		 * If @p is in an open layer, keep running if there's any idle
-		 * CPU. If confined, keep running iff the layer has idle CPUs.
+		 * CPU. If confined, keep running if the layer has idle CPUs.
```
htejun, line 1079: I think `iff` is intentional, meaning `if and only if`.

htejun (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: That makes sense and then we'd only have to update the masks when the layers change size.

PR #686, opened 2024-09-25
scx_utils: Add gpu-topology crate feature

### LINKED ISSUE #589: scx_utils: Add a crate feature for opt-in GPU topology support
Add a crate feature to the `scx_utils` crate for GPU topology support. Not all systems need GPU topology support and enabling GPU topology support will bring in vendor specific dependencies. See the `scx_utils` [`Cargo.toml`](https://github.com/sched-ext/scx/blob/main/rust/scx_utils/Cargo.toml) and the docs for [crate features](https://doc.rust-lang.org/cargo/reference/features.html) for details.

This requires only knowledge of Rust/Cargo so it should be pretty straight forward.

frelon:
The gpu-topology feature can be enabled to include GPUs when generating a topology-map. Disabling the feature will remove the nvml-wrapper dependency as well as GPU-specific code in topology.rs.

Most of the code was moved to a new module in rust/scx_utils/src/gpu.rs but some of it was kept in topology.rs and hidden behind #[cfg(feature = "gpu-topology")].

Fixes #589

### DIFF
--- a/rust/scx_utils/Cargo.toml
+++ b/rust/scx_utils/Cargo.toml
@@ -19,7 +19,7 @@ lazy_static = "1.4"
 libbpf-cargo = "0.24.1"
 libbpf-rs = "0.24.1"
 log = "0.4.17"
-nvml-wrapper = "0.10.0"
+nvml-wrapper = { version = "0.10.0", optional = true }
 paste = "1.0"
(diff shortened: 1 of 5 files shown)


hodgesds (approved):
Thanks for working on this, LGTM!

PR #689, opened 2024-09-25
scx_utils: Fix linter errors

frelon:
Fix linter errors using `cargo fmt`

Signed-off-by: Fredrik Lönnegren <fredrik.lonnegren@suse.com>

### DIFF
--- a/rust/scx_utils/src/gpu.rs
+++ b/rust/scx_utils/src/gpu.rs
@@ -1,4 +1,4 @@
-#![cfg(feature="gpu-topology")]
+#![cfg(feature = "gpu-topology")]
 
 use crate::misc::read_file_usize;
 use nvml_wrapper::bitmasks::InitFlags;
@@ -77,5 +77,3 @@ pub fn create_gpus() -> BTreeMap<usize, Vec<Gpu>> {
 
     gpus
 }
-
-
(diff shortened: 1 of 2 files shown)


htejun:
I wonder whether this is just a different version of `cargo` being used. @JakeHillion, aren't we auto-formatting everything now?

hodgesds:
I think there is some noise in the linter test from some changes in the loader, see the test [failures](https://github.com/sched-ext/scx/actions/runs/11040959192/job/30670157058?pr=689). It might be from using [nightly](https://github.com/sched-ext/scx/blob/main/.github/workflows/caching-build.yml#L25) for `cargo fmt`.

likewhatevs:
The command is `cargo +nightly-2024-09-10 fmt` (presuming cargo was installed via rustup or rustup is installed), I have another PR which runs that for the whole repo (it needs that atm to make ci signal green).

frelon:
I see, looks like this is already merged in #691, thanks!

JakeHillion:
Thanks folks for bringing this check back in line. @likewhatevs is right, you need a specific version. When I enabled this I found that we needed nightly cargo/rustfmt for this to work, the ordinary version complained a lot, hence the pinning to a specific version of nightly in an effort for stability. The downside is everyone needs to have the identical version, alternatively you can copy the patch provided in the CI failure and apply it with `cat | patch -p1`.

Now this check has been okay for a while I'll make it a requirement on PRs, slightly more work for contributors but it'll keep it clean in future.

htejun:
@JakeHillion can you document how to configure cargo to make it format the same way? Can that be a part of the repo itself maybe?

PR #692, opened 2024-09-25
Sync from kernel and re-enable scx_flatcg and scx_pair

htejun:
Receive pending updates from kernel which includes cgroup support and `scx_bpf_dispatch[_vtime]_from_dsq()` and re-enable `scx_flatcg` and `scx_pair`.


hodgesds (approved):
LGTM

likewhatevs (approved):
fan of the comments w/ version in compat

PR #693, opened 2024-09-25
scx_loader: Add systemd service and on-DBUS launch

vnepogodin:
Provide additional systemd service for `scx_loader` to be able to autostart it by DBUS, even if the service was enabled/started by user, DBUS will search by name and start it with propagation of called dbus method/property

ref: "Example 7. DBus services" https://www.freedesktop.org/software/systemd/man/256/systemd.service.html

### DIFF
--- a/rust/scx_loader/README.md
+++ b/rust/scx_loader/README.md
@@ -46,6 +46,42 @@
 
 **Note:** Replace the example scheduler names and arguments with the actual ones you want to use.
 
+## DBUS and Systemd Service
+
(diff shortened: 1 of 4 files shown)


ptr1337 (approved):
Please add into the systemd services README info about loader and its separate service to avoid confusion

vnepogodin:
> Please add into the systemd services README info about loader and its separate service to avoid confusion

I think will add to scx_loader README, that systemd service activation/running is handled by DBUS automatically. Also that starting scx_loader.service manually is not required

sirlucjan (approved):

ptr1337 (approved):

PR #695, opened 2024-09-26
Setup "debugging" and misc cleanup

likewhatevs:
This PR reworks how CI is set up a bit to make debugging a bit simpler.

* Fix a couple of misc errors in build scripts.
* Tweak scripts/kconfigs to make bpftrace work.
* Update how CI caching works to make builds faster (6 minute turnaround time)
* Update CI config to generate per-scheduler debug archives w/ guest dmesg/scheduler stdout, guest stdout, bpftrace script output, veristat output.

* Update build scripts to accept the following: 
*** VNG RW -- write to host filesystem (better caching, logging).
** For stress tests in particular (via ini config): 
*** QEMU Opts -- to facilitate reproducing bugs (i.e. high core count). ** bpftrace scripts -- specify bpftrace scripts to run during stress tests.

* All files ending in `*.ci.log`  are stored in archives accessible on the job's run page.
image

Here is an example of how jobs look w/ this (and the downloadable logs): https://github.com/likewhatevs/scx/actions/runs/11053261025

* note -- passing qemu_opts to vng doesn't work yet, I'm not sure how to make that command work w/ subprocess yet (might be an issue in vng, need to look). 

* also note -- the logs saved look nice if opened w/ a tool (like bat) that deals with escape codes etc.
image

I kinda hope we can build a collection of bpftrace scripts we can run against all schedulers to idk make debugging simpler.


hodgesds (approved):
LGTM

PR #697, opened 2024-09-26
do not cache fast jobs dependencies

likewhatevs:
do not cache fast jobs dependencies

doing this will make slow jobs caches get evicted (and the fast jobs are not the long tail anyway).


hodgesds (approved):
LGTM

PR #702, opened 2024-09-27
scx_lavd: more accurately determine the performance criticality threshold

multics69:
We used the average performance criticality of tasks as a threshold to
determine the proper core type (big or little). However, if the big
core's compute capacity is not half of the total compute capacity, such
an average-based determination becomes suboptimal. If fewer tasks are
classified as performance-critical tasks and requested to run on big
cores, the big cores would be wasted by stealing arbitrary
non-performance-critical tasks. That could result in performance
instability.

Hence, determine the threshold more accurately by considering (active)
big cores' compute capacity and the (approximated) distribution of
performance criticality of tasks.


multics69:
Thank you @hodgesds for review the code.
I also found that lavd is failing. I will take a look and send a separate PR when fixed.

PR #703, opened 2024-09-28
scx_stats: Implement macro #stat_doc to autogen doc from stat desc

minosfuture:
The doc of scx_layered `Opt` is out of sync.

Implement attribute macro #stat_doc to generate doc from the `desc` property, and apply #stat_doc to `LayerStats` and `SysStats` in scx_layered.

Also fix typo and replace `#` with "count of" to avoid messing with markdown syntax in the generated rust doc.

Signed-off-by : Ming Yang <minos.future@gmail.com>

### DIFF
--- a/rust/scx_stats/scx_stats_derive/src/lib.rs
+++ b/rust/scx_stats/scx_stats_derive/src/lib.rs
@@ -3,6 +3,7 @@ use scx_stats::{StatsData, StatsKind, StatsMetaAux};
 use std::sync::atomic::{AtomicU64, Ordering};
 use syn::parse_macro_input;
 use syn::spanned::Spanned;
+use syn::{Attribute, Data, DeriveInput, Fields, Lit};
 
 static ASSERT_IDX: AtomicU64 = AtomicU64::new(0);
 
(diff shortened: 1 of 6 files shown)


minosfuture:
Adding screenshot of generated rust doc:

image

image

hodgesds:
Looks like a slight linter error on the [workflow](https://github.com/sched-ext/scx/actions/runs/11081823266/job/30799685936?pr=703) for `cargo fmt`, it's probably a difference in the cargo version. Otherwise, this looks great!

rust/scx_stats/scx_stats_derive/src/lib.rs
```diff
@@ -46,3 +47,99 @@ pub fn stat(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+                        Ok(())
+                    })
+                    .unwrap_or_else(|err| {
+                        panic!("Failed to parse the stat attribute: {}", err);
+                    });
+                }
+                new_attrs.push(attr);
+            }
+
+            // If a description string was found, add a #[doc = "..."] attribute
```
htejun, line 104: Can we just rename desc field to doc?

scheds/rust/scx_layered/src/stats.rs
```diff
@@ -55,9 +57,9 @@ pub struct LayerStats {
     pub load: f64,
     #[stat(desc = "fraction of total load")]
     pub load_frac: f64,
-    #[stat(desc = "# tasks")]
+    #[stat(desc = "count of tasks")]
```
htejun, line 60: Are these changes beacause doc doesn't allow hash sign?

minosfuture: it's because the hash sign in the front indicates heading in markdown.

rust/scx_stats/scx_stats_derive/src/lib.rs
minosfuture: changing `#[stat(desc = "...")]` into `#[stat(doc = "...")]` won't make it into the generated documentation because doc attribute is actually `#[doc = "..."]`.
Also, `desc` in `stat` macro here is put into the `StatsMeta`, for example, here is the expanded rust code:
image

with the change in this PR, the original struct is extended with document like the following, where `/// ...` lines are added by the `stat_doc` macro.
image

htejun (approved):

rust/scx_stats/scx_stats_derive/src/lib.rs
htejun: Hmmm... I wonder whether it should be synced the other way then - ie. let scx_stats macro to take the `///` comment and turn that into description field. What do you think?

minosfuture: I think it should be the current way. Documents can be out of rules, while a `desc` property feels constrained to be one short sentence. In the current way, it's also composable as one can add more doc in comment. So more than sync, it's more an extension to the doc from the desc property. Screenshots for a demo are attached: 
image

image

htejun: Understood. Thanks for the explanation and the PR!

PR #704, opened 2024-09-28
scx_lavd: fix two potential bugs

multics69:
This PR contains two potential bugs, one of which causes a stress test failure in CI.
- scx_lavd: do not perform the victim selection for an invalid cpu
- scx_lavd: fix incorrect neighbor_bit initialization


arighi (approved):
Looks good. Thanks for fixing this.

Issue #611, opened 2024-09-05
scx_rusty: Make load balancing decision executed reliably


htejun:
Right now, after the userland loadbalancer makes migration decisions (move task X to domain N), the decision is recorded in the `lb_data` map which is a map from `pid_t` to destination domain number. Then, on the enqueue path, `rusty_enqueue()` checks whether the task has a matching entry in `lb_data` and if so executes the requested migration. This means that the application of LB decisions isn't reliable - it depends on the task being migrated running in the following period. Otherwise, the decision is ignored.

While this works okay in practice as the LB just keep retrying until the domains are balanced, this makes the behavior less predictable. It'd be great to make the load balancing decisions executed reliably. Maybe `test_run` can be used to execute migrations immediately - see `set_power_profile()` in `scx_lavd` for an example.

vax-r:
Is it still available ? If so I would love to help

vax-r:
The draft PR is ready but I've encountered a problem which might need some help. The machine I'm running on now is using AMD Ryzen 7 5700X3D 8-Core Processor and under rusty, it has only 1 NUMA node and 1 domain, so the load balancing step would be stopped before it can actually do anything. That way I can't test whether my change has made some improvement or not, maybe someone would be so kind to test the PR for me ?

If that would do I'll send a draft PR first and see what's the testing result, otherwise I'll try to think of other ways to test it.

htejun:
You can use `-C` option to define arbitrary LB domains, which should be sufficient for testing.

vax-r:
> You can use `-C` option to define arbitrary LB domains, which should be sufficient for testing.

Thank you ! I can test it now and found some problems, I'll figure it out and send a PR later.

vax-r:
I've open a PR , please take a look while you're available, thanks !

PR #705, opened 2024-09-28
scx_rusty: Enable recording of active task_struct

vax-r:
## Summary
This is for the pre-work of #611 .

Store currently active tasks with their pointer "struct task_struct *" within "struct dom_active_pids". In order to achieve operation such as lookup "struct task_struct *" by task's pid in the future implementation.

We'll need to be able to reference `task_struct` with `pid` so we can use `test_run()` from userspace Load Balancer with task's `pid` to set its domain.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -154,6 +154,7 @@ struct dom_active_pids {
 	u64 read_idx;
 	u64 write_idx;
 	s32 pids[MAX_DOM_ACTIVE_PIDS];
+	struct task_struct *tasks[MAX_DOM_ACTIVE_PIDS];
 };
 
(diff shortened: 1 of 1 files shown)


likewhatevs:
I think you need to migrate to just using task struct pointers instead of pid (and get the verifier to be happy with that) to avoid pid missing errors across the bpf and rust code. I have a WIP PR that ought to do that (at the "make the verifier happy part" atm) which might be good in a few days.

vax-r:
> I think you need to migrate to just using task struct pointers instead of pid (and get the verifier to be happy with that) to avoid pid missing errors across the bpf and rust code. I have a WIP PR that ought to do that (at the "make the verifier happy part" atm) which might be good in a few days.

Yes I know the PR you are working on, it should allow me to use just `u64` to refer to the corresponding `task_struct` . I should make this as a draft and migrate when your PR is merged. Thank you!

PR #706, opened 2024-09-29
Revert "scx_rustland_core: prevent deadlock with per-CPU DSQs and CPU…

arighi:
… affinity"

It seems that with the latest kernel the per-CPU DSQ stall while executing sched_setaffinity() doesn't happen anymore.

Therefore, get rid of the temporary workaround introduced by commit 86db45f ("scx_rustland_core: prevent deadlock with per-CPU DSQs and CPU affinity") and restore the old behavior, which offers more fair scheduling policy.


hodgesds (approved):
LGTM

PR #710, opened 2024-09-30
have rusty use a counter of task ptr instead of pid

likewhatevs:
Have rusty track a counter of unique task ptr (mapped to task ptr) instead of pid.

This seems to fix the task missing failures and might work (i.e. the counter is bumped atomically and lookup works both ways (fake pid -> task struct, task struct to fake pid).


likewhatevs:
going back to the other approach due to issues w/ s32 rollover (i.e. dealing with that wrt/ systems w/ long uptime etc.).

PR #712, opened 2024-10-01
Add #stat_doc attribute macro to Stats structs

minosfuture:
`#stat_doc` extends the document from stat desc property.

Add this attribute macro to the remaining Stats structs.

This follows up #703.


hodgesds (approved):
Thank you!

Issue #684, opened 2024-09-25
scx_rusty: Ends immediately after start when kernel compilation is running.


YUBY64:
Rusty ends immediately after start when kernel compilation is running. If I start Rusty first then kernel compilation then Rusty works well. Other schedulers are unaffected.

```❯ sudo scx_rusty
13:25:47 [INFO] Running scx_rusty (build ID: 1.0.4-gaea431c x86_64-unknown-linux-gnu)
13:25:47 [INFO] NODE[00] mask= ffff
13:25:47 [INFO]  DOM[00] mask= ffff
13:25:48 [INFO] Rusty scheduler started! Run `scx_rusty --monitor` for metrics.

DEBUG DUMP
================================================================================

scx_rusty[105233] triggered exit kind 1025:
  scx_bpf_error (Failed to lookup task 105197)

Backtrace:
  scx_bpf_error_bstr+0xf9/0x1a0
  bpf_prog_3d1f5ab55b43878c_dom_xfer_task+0x61/0x13b0
  bpf_prog_bf2f195073ed7dd1_task_set_domain+0x153/0x1c9
  bpf_prog_f9ebe23c1ccd299e_task_pick_and_set_domain+0x4b/0x9f
  bpf_prog_a0805f20ddc55e97_rusty_set_cpumask+0x98/0xc2

CPU states
----------

CPU 0   : nr_run=1 flags=0x1 cpu_rel=0 ops_qseq=420843 pnt_seq=226345
          curr=scx_rusty[105233] class=ext_sched_class

 *R scx_rusty[105233] +0ms
      scx_state/flags=3/0x5 dsq_flags=0x0 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=(n/a) dsq_vtime=0
      cpus=ffff

    scx_dump_state+0x71a/0x8e0
    scx_ops_error_irq_workfn+0x40/0x50
    irq_work_run_list+0x50/0x90
    irq_work_run+0x18/0x50
    __sysvec_irq_work+0x1c/0xb0
    sysvec_irq_work+0x64/0x80
    asm_sysvec_irq_work+0x1a/0x20
    scx_ops_enable.isra.0+0xaf0/0xfd0
    bpf_struct_ops_link_create+0x12c/0x180
    __sys_bpf+0x1b5d/0x2bb0
    __x64_sys_bpf+0x25/0x30
    do_syscall_64+0x82/0x190
[...]

PR #713, opened 2024-10-01
Replace PID with Task Pointer in Rusty

likewhatevs:
~I wasnt't able to reopen https://github.com/sched-ext/scx/pull/672 because I force pushed (just learned about that limit).~

~This is a continuation of that PR.~

~I closed that PR and opened https://github.com/sched-ext/scx/pull/710 because, while the first PR's approach passed stress tests and seemed to work, it kind of didn't make sense that I could deref the casted pointer to task_struct.~

~I updated that initial PR to explicltly deref that casted pointer to confirm this works/the verifier allows it and it seems to work so, this works?~

see latest comment

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -149,21 +149,21 @@ struct {
 	__uint(map_flags, 0);
 } dom_dcycle_locks SEC(".maps");
 
-struct dom_active_pids {
+struct dom_active_tptrs {
 	u64 gen;
 	u64 read_idx;
 	u64 write_idx;
-	s32 pids[MAX_DOM_ACTIVE_PIDS];
+	u64 tptrs[MAX_DOM_ACTIVE_TPTRS];
 };
 
-struct dom_active_pids dom_active_pids[MAX_DOMS];
+struct dom_active_tptrs dom_active_tptrs[MAX_DOMS];
 
 const u64 ravg_1 = 1 << RAVG_FRAC_BITS;
 
-/* Map pid -> task_ctx */
+/* Map tptr -> task_ctx */
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__type(key, pid_t);
+	__type(key, u64);
 	__type(value, struct task_ctx);
 	__uint(max_entries, 1000000);
 	__uint(map_flags, 0);
@@ -185,20 +185,37 @@ static struct dom_ctx *lookup_dom_ctx(u32 dom_id)
 	return domc;
 }
 
+static u64 t_to_tptr(struct task_struct *p)
+{
+	u64 tptr;
+	int err;
+
+	err = bpf_probe_read_kernel(&tptr, sizeof(tptr), &p);
+
+	if (err){
+		scx_bpf_error("Failed to cast task_struct addr to tptr");
+		return 0;
+	}
+	return tptr;
+}
+
 static struct task_ctx *try_lookup_task_ctx(struct task_struct *p)
 {
-	s32 pid = p->pid;
+	u64 tptr;
+	tptr = t_to_tptr(p);
 
-	return bpf_map_lookup_elem(&task_data, &pid);
+	return bpf_map_lookup_elem(&task_data, &tptr);
 }
 
 static struct task_ctx *lookup_task_ctx(struct task_struct *p)
 {
 	struct task_ctx *taskc;
+	u64 tptr;
+	tptr = t_to_tptr(p);
 
 	taskc = try_lookup_task_ctx(p);
 	if (!taskc)
-		scx_bpf_error("task_ctx lookup failed for pid %d", p->pid);
+		scx_bpf_error("task_ctx lookup failed for tptr %llu", tptr);
 
 	return taskc;
 }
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -417,7 +417,9 @@ static int dom_xfer_task(u64 task_id, u32 new_dom_id, u64 now)
 	if (!get_task_from_id(task_id, p)) {
 		scx_bpf_error("Failed to lookup task_struct");
 		return 0;
-	}
+	} 
+	
+	s32 a = p->pid;
```
htejun, line 422: Compiler will most likely eliminate this.

vax-r:
Wouldn't this design throw away the origin safety provided by `bpf_task_from_pid()` ? It was able to perform reference count on the `task_struct` .
Using a map to store is awesome but it seems that we do not have mechanism to cleanup the unused <k, v> pair in the map, as the time going, the map size can grow really big, even it's acceptable, we would still like to avoid redundant memory usage , maybe having that implemented will be nice?

vax-r:
Btw I want to ask what's the actual cause of `Failed to lookup task %d` ?

likewhatevs:
I *think* this is the best we (or well, at least I) can do atm. 

So like, I tried a few ways of tracking the association between the initial pid of a task struct and the pointer of the task struct. 

My most recent attempts errored at null-checking the task context read using that association:
```
; if (!taskc) goto free_task; @ main.bpf.c:456
372: (15) if r3 == 0x0 goto pc+27
R3 !read_ok
processed 11065 insns (limit 1000000) max_states_per_insn 8 total_states 740 peak_states 298 mark_read 18
-- END PROG LOAD LOG --
```

Unless I'm reading something wrong or misunderstand something, I think this is the best we can do ATM (i.e. until it's possible to null check the retrieved variable).

I think this might be alright though because like, tests *do* pass and logs on those tests do not have stalls (which, I think, is the main reason to be concerned we are tracking everything).

vax-r:
> I think this might be alright though because like, tests _do_ pass and logs on those tests do not have stalls (which, I think, is the main reason to be concerned we are tracking everything).

It seems to me that what the commit does is only refactor the origin code , rather than actually changing any logic or mechanism when dealing with `pid` and `task_struct`, how can it pass this time ? It seems confusing to me.

hodgesds:
> This is correct. What I am wondering is if the current behavior (without that exception being thrown) is problematic. I would think I would see a stall if it were if I understand how things work correctly now, but I'm not super certain of that.

I think the idea is that if correct scheduling decisions can't be made by the the scheduler it's better to exit (the scheduler) rather than making bad scheduling decisions.  There's still some bugs in rusty, so in some ways it maybe is better to fail fast in more places. In this case of looking up the task context I think it's probably ok to bail the scheduler, because if the task context is lost/incorrect then making correct scheduling decisions is probably not handled correctly elsewhere.

hodgesds:
> This is the part where like, IDK. So like, if we lost the task, it wouldn't be scheduled, and we would see a stall?

I think the problem with `stress-ng` tests is that with the `--fork` tests the tasks come and go really fast, I'm not sure if there's an easy way to trigger the behavior (maybe compiling the kernel like in #684) and see what happens with this change. Maybe do a kernel build with something crazy like ````make -j $(nproc * 2)```.

likewhatevs:
> I think the problem with stress-ng tests is that with the --fork tests the tasks come and go really fast, I'm not sure if there's an easy way to trigger the behavior (maybe compiling the kernel like in https://github.com/sched-ext/scx/issues/684) and see what happens with this change. Maybe do a kernel build with something crazy like ````make -j $(nproc * 2)```.

gonna run it on laptop for a day and do the terraria/compile test/see if I can find bad behavior, good idea, thx!

likewhatevs:
Alright so I think I finally got this (and learned a ton along the way).

This PR now has Rusty use Task pointer casted to u64 to track tasks instead of pid.

A couple things I learned:

In bpf, you can't cast "trusted pointers". I don't fully get it, but t_to_tptr makes the cast work w/ the verifier.

Making functions static makes some problems go away, but can also super quickly drive things over the instruction limit.

*More interestingly*, I tried only grabbing pid via probe_read (rather than a pointer to task_struct), and to then grab task_struct via `bpf_task_from_pid` (because maybe the bpf locking could be useful/I was curious). 

What I found is that is *super* racy (i.e. adding a counter and a while loop made stress test last longer while before failing with a missing pid error), so I kind of think we need to not use `bpf_task_from_pid` anywhere.

hodgesds (approved):
I think this will work!

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -397,31 +414,43 @@ static u64 dom_min_vruntime(struct dom_ctx *domc)
...
+    struct task_struct *p;
+	int err_task;
+	struct task_struct *task;
+
+	task = (struct task_struct *)tptr;
+
+	err_task = bpf_probe_read_kernel(&p, sizeof(struct task_struct *), &task);
+
+	if (err_task) scx_bpf_error("Failed to retrieve task_struct for tptr %llu", tptr);
+	if(p) return p;
```
hodgesds, line 427: nit: consistent spacing

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -397,31 +414,43 @@ static u64 dom_min_vruntime(struct dom_ctx *domc)
 	return READ_ONCE(domc->min_vruntime);
 }
 
-int dom_xfer_task(pid_t pid, u32 new_dom_id, u64 now)
+static struct task_struct *tptr_to_task(u64 tptr){
```
hodgesds, line 417: nit:
```
static struct task_struct *tptr_to_task(u64 tptr)
{
        struct task_struct *p, *task;
	int err_task;
```

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -788,9 +819,10 @@ static bool task_set_domain(struct task_ctx *taskc, struct task_struct *p,
 	if (bpf_cpumask_intersects((const struct cpumask *)d_cpumask,
 				   p->cpus_ptr)) {
 		u64 now = bpf_ktime_get_ns();
-
-		if (!init_dsq_vtime)
-			dom_xfer_task(p->pid, new_dom_id, now);
+		
+		if (!init_dsq_vtime){
```
hodgesds, line 823: nit:
```
if (!init_dsq_vtime) {
```

PR #719, opened 2024-10-02
scx_rusty: Implementation of rusty_cpu_online

vax-r:
## Summary
Implement rusty_cpu_online so we can correctly initialize the per-cpu context for every cpu. Originally we only initialize per-cpu context when the scheduler is starting up, if some cpus was offline at start-up time but become online later in the runtime, we won't be able to promise the per-cpu context is correct for each cpu.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -1839,6 +1839,12 @@ static s32 initialize_cpu(s32 cpu)
 	return -ENOENT;
 }
 
+void BPF_STRUCT_OPS(rusty_cpu_online, s32 cpu)
+{
+	if(initialize_cpu(cpu))
+		scx_bpf_error("Failed to initialize cpu");
+}
+
 s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init)
 {
 	s32 i, ret;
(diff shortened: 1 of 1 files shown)


htejun:
If a scheduler doesn't implement cpu_on/offline, the scheduler will be auto restaretd on hotplug and thus reinitialize everything, which is a rather heavy-handed way to handle hotplug but should be fine for use cases which don't use hotplug extensively (AFAIK, this is prevalent only in some embedded usecases).

Also, how is the change tested? If rusty is going to implement hot CPU onlining without restarting, it needs to refresh topology information so that the newly online CPU can be put into the correct domain and that information is not available before the CPU is actually online.

vax-r:
> Also, how is the change tested? If rusty is going to implement hot CPU onlining without restarting, it needs to refresh topology information so that the newly online CPU can be put into the correct domain and that information is not available before the CPU is actually online.

I've only ran some stress test and test suite but didn't actually make a CPU offline/online, you're right I should figure out how to perform it and the topology part should be modified as well.



> If a scheduler doesn't implement cpu_on/offline, the scheduler will be auto restaretd on hotplug and thus reinitialize everything, which is a rather heavy-handed way to handle hotplug but should be fine for use cases which don't use hotplug extensively (AFAIK, this is prevalent only in some embedded usecases).

So I assume we do not need this for now? or it will be a nice-to-have feature ?

htejun:
As-is, this blocks the existing auto restart mechanism while not actually implementing working hotplug mechanism and the whole change is not tested in any meaningful way. I appreciate the effort but please reconsider how you contribute. I'm sure that will also be helpful in improving how you approach technical problems.

PR #726, opened 2024-10-03
Run integration tests against bpf-next kernel every 6 hrs.

likewhatevs:
Run integration tests against bpf-next kernel every 6 hrs.

This seems like a reasonable way to do this (i.e. don't add noise to folks developing/testing things, but also let us know if there are issues w/ our code and bpf-next).

Will add a comment to a successful job run on my fork once it runs.


hodgesds (approved):
LGTM

likewhatevs:
https://github.com/likewhatevs/scx/actions/runs/11161649351/job/31024642405 -- ran fine but used the wrong kernel due to caching, fixed that.

PR #729, opened 2024-10-03
scx_pair: Implement cgrp_q_busy as bpf hash map

vax-r:
## Summary
Implementation for transfering array "cgrp_q_idx_busy" into a bpf hash map "cgrp_q_busy" for faster access and consistent datastructure using bpf map over the whole scheduler lookup process.


htejun:
I don't think this is a useful change.

PR #736, opened 2024-10-04
scx_lavd: split main.bpf.c into multiple files

multics69:
As the main.bpf.c file grows, it gets hard to maintain. So, split it into multiple logical files. There is no functional change.


hodgesds (approved):
One step closer to being able to reuse some of the logic with other schedulers :+1:

multics69:
Thanks @hodgesds !

PR #740, opened 2024-10-05
scx_layered: Use idle smt mask for idle selection

hodgesds:
In the non topology aware code the idle smt mask is used for finding idle cpus. Update topology aware idle selection to also use the idle smt mask. In certain benchmarks this can improve performance.

Tested on `12th Gen Intel(R) Core(TM) i3-1220P` with P/E cores and `stress-ng` configured with `BigLittle` growth algorithm on big cores.

With idle smt mask:
```
stress-ng -c 2 -t 10 -M
stress-ng: info:  [17956] setting to a 10 secs run per stressor
stress-ng: info:  [17956] dispatching hogs: 2 cpu
stress-ng: metrc: [17956] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s CPU used per       RSS Max
stress-ng: metrc: [17956]                           (secs)    (secs)    (secs)   (real time) (usr+sys time) instance (%)          (KB)
stress-ng: metrc: [17956] cpu               40708     10.00     20.00      0.00      4071.28        2035.65       100.00          7264
stress-ng: info:  [17956] skipped: 0
stress-ng: info:  [17956] passed: 2: cpu (2)
stress-ng: info:  [17956] failed: 0
stress-ng: info:  [17956] metrics untrustworthy: 0
stress-ng: info:  [17956] successful run completed in 10.00 secs
```
Idle mask:
```
stress-ng -c 2 -t 10 -M
stress-ng: info:  [18308] setting to a 10 secs run per stressor
stress-ng: info:  [18308] dispatching hogs: 2 cpu
stress-ng: metrc: [18308] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s CPU used per       RSS Max
stress-ng: metrc: [18308]                           (secs)    (secs)    (secs)   (real time) (usr+sys time) instance (%)          (KB)
stress-ng: metrc: [18308] cpu               38429     10.00     20.00      0.00      3842.14        1921.07       100.00          7360
stress-ng: info:  [18308] skipped: 0
stress-ng: info:  [18308] passed: 2: cpu (2)
stress-ng: info:  [18308] failed: 0
stress-ng: info:  [18308] metrics untrustworthy: 0
stress-ng: info:  [18308] successful run completed in 10.00 secs
```

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -184,6 +184,7 @@ struct layer {
 	bool			preempt;
 	bool			preempt_first;
 	bool			exclusive;
+	bool			idle_smt;
 	int			growth_algo;
 
 	u64			vtime_now;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -650,7 +650,7 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
 					      idle_cpumask)) >= 0)
 			goto out_put;
 	} else {
-		idle_cpumask = scx_bpf_get_idle_cpumask();
+		idle_cpumask = scx_bpf_get_idle_smtmask();
 		if (!pref_idle_cpumask || !idle_cpumask) {
 			cpu = -1;
 			goto out_put;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -650,7 +650,12 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
 					      idle_cpumask)) >= 0)
 			goto out_put;
 	} else {
-		idle_cpumask = scx_bpf_get_idle_smtmask();
+		if (layer->idle_smt) {
+			idle_cpumask = scx_bpf_get_idle_smtmask();
+		} else {
+			idle_cpumask = scx_bpf_get_idle_cpumask();
+		}
+
 		if (!pref_idle_cpumask || !idle_cpumask) {
 			cpu = -1;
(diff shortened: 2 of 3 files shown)


likewhatevs (approved):
That's pretty cool for a tiny change. I'd run a longer test though (maybe a minute or two)?

hodgesds:
> That's pretty cool for a tiny change. I'd run a longer test though (maybe a minute or two)?

Yeah, the results really depend on whether or not the big cores are used. I'm not sure if it's better to pick and idle little core vs a hyperthread big core though. Maybe we should run a bunch of benchmarks. With the current setting it will be more sticky to big cores. Or we could just be lazy and make it a config that the user has to determine.

JakeHillion:
> > That's pretty cool for a tiny change. I'd run a longer test though (maybe a minute or two)?
> 
> Yeah, the results really depend on whether or not the big cores are used. I'm not sure if it's better to pick and idle little core vs a hyperthread big core though. Maybe we should run a bunch of benchmarks. With the current setting it will be more sticky to big cores. Or we could just be lazy and make it a config that the user has to determine.

It depends a lot on the workload right? My guess would be that there are very few tasks which prefer the smaller cache/tiny ALU of a Little core versus a SMT thread. I think defaulting to Big is reasonable and having a config option wouldn't hurt, especially considering this is layered.

hodgesds:
> It depends a lot on the workload right? My guess would be that there are very few tasks which prefer the smaller cache/tiny ALU of a Little core versus a SMT thread.

Yeah, and the other interesting thing is it depends on the number of big cores in the system as well. For my little router there's only two big cores, using the idle smt mask really kills performance when both big cores have something running on them (depending on config, preemption etc). So maybe it is best to have this as a config, maybe it could be per layer even.

hodgesds:
Adding it to the layer config works really nice. On the low priority little core layers it spreads out tasks rather nicely and on the important work running on pcores it keeps cores compacted nicely.

PR #741, opened 2024-10-05
scx_bpfland: rework lowlatency mode

arighi:
In lowlatency mode (option --lowlatency) tasks are ordered using a deadline that is evaluated as the vruntime minus a certain "bonus", determined in function of the max time slice and the average amount of voluntary context switches, to amplify the priority boost of the tasks that are voluntarily releasing the CPU (which are typically interactive).

However, this method can be extremely unfair in some cases: tasks with short bursts of voluntary context switches may receive a huge priority boost, making the rest of the system almost unresponsive (see massive hackbench stress tests for example).

To prevent this rework the task's deadline logic to use the vruntime and a "deadline component" that is a function of the average used time slice, scaled using a dynamic task priority (evaluated as the static task priority and the its average amount of voluntary context switches).

This logic seems to prevent excessive prioritization of tasks performing short intensive bursts of voluntary context switches.

It also makes lowlatency mode in scx_bpfland (somehow) more similar to the deadline logic used by scx_rusty.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -236,6 +236,21 @@ struct task_ctx {
 	u64 avg_nvcsw;
 
 	/*
+	 * Task's average used time slice.
+	 */
+	u64 avg_runtime;
+
+	/*
+	 * Last task's execution time.
+	 */
+	u64 last_running;
+
+	/*
+	 * Task's deadline.
+	 */
+	u64 deadline;
+
+	/*
 	 * Set to true if the task is classified as interactive.
 	 */
 	bool is_interactive;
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -345,35 +371,59 @@ static u64 calc_avg_clamp(u64 old_val, u64 new_val, u64 low, u64 high)
...
+	if (!lowlatency)
+		return 0;
+
+	/*
+	 * If the task has not ran during the previous slice_ns period, use its
+	 * vruntime as deadline to give it a priority boost. This allows to
+	 * speed up tasks that are mostly sleeping and they suddenly need to
+	 * react fast.
+	 */
+	if (vtime_before(tctx->last_running + slice_ns, bpf_ktime_get_ns()))
```
multics69, line 406: I guess such a task is mostly synchronously wakened up. If so, does this give double prioritization in addition to SCX_WAKE_SYNC handling?

multics69 (approved):
Looks great! Welcome to join the deadline club!

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Possibly... in this case SCX_WAKE_SYNC will promote the task to interactive, so it'll be dispatched to the priority DSQ and in addition to that the deadline will be set equal to its vruntime, so it will likely be executed before the other tasks in the prio DSQ.

This may result in excessive prioritization, but it seems to help to maintain system responsiveness when running massive hackbench / perf bench stress tests (that are also classified as interactive tasks and dispatched to the priority DSQ).

PR #743, opened 2024-10-06
scx_layered: Add big cpumask

hodgesds:
Add big cpumask to scx_layered and prefer selecting big idle cores when using the `BigLittle` growth algo. Future work can be preferring selecting little idle cores for the `LittleBig` growth algo.


JakeHillion (approved):
lgtm!

PR #746, opened 2024-10-07
scx_layered: lighten/reduce nested loops in layered dispatch

likewhatevs:
I *think* (like, *maybe* think) this improves the responsiveness of layered by fixing some off-by-1 error.

Interesting thing about the results below:
18.72/15.60=1.2 (main fork user sec / branch fork user sec)
146579/131561=1.114152370 (main bogo ops / branch bogo ops)
9/7.66 = 1.17493473  (main cpu used % per instance / branch cpu used % per instance)

I think that, in combination with the rest of the results, is saying:
This change reduces the absolute number of instructions executed for this test, but it decreases how long it takes (in terms of time) for those instructions to be executed and how much CPU% those insns cost.

```
main

STRESS OUTPUT
stress-ng: info:  [295] setting to a 14 secs run per stressor
stress-ng: info:  [295] dispatching hogs: 32 cpu, 32 fork
stress-ng: info:  [295] note: /proc/sys/kernel/sched_autogroup_enabled is 1 and this can impact scheduling throughput for processes not attached to a tty. Setting this to 0 may improve performance metrics
stress-ng: metrc: [295] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s CPU used per       RSS Max
stress-ng: metrc: [295]                           (secs)    (secs)    (secs)   (real time) (usr+sys time) instance (%)          (KB)
stress-ng: metrc: [295] cpu              632537     14.00    279.98      0.62     45175.58        2254.20        62.63          6400
stress-ng: metrc: [295] fork             146579     14.00     18.72     21.60     10469.61        3635.57         9.00          2628
stress-ng: info:  [295] skipped: 0
stress-ng: info:  [295] passed: 63: cpu (31) fork (32)
stress-ng: info:  [295] failed: 0
stress-ng: info:  [295] metrics untrustworthy: 0
stress-ng: info:  [295] successful run completed in 14.02 secs
STRESS OUTPUT DONE
[   26.559795] ACPI: PM: Preparing to enter system sleep state S5
[   26.559945] kvm: exiting hardware virtualization
[   26.560172] reboot: Power down
[ble: elapsed 59.949s (CPU 554.7%)] ./local-e2e.sh scx_layered


branch

STRESS OUTPUT
stress-ng: info:  [296] setting to a 14 secs run per stressor
stress-ng: info:  [296] dispatching hogs: 32 cpu, 32 fork
[...]

PR #749, opened 2024-10-07
layered: split dispatch into no_topo version

JakeHillion:
Refactor layered_dispatch into two functions: layered_dispatch_no_topo and
layered_dispatch. layered_dispatch will delegate to layered_dispatch_no_topo in
the disable_topology case.

Although this code doesn't run when loaded by BPF due to the global constant
bool blocking it, it makes the functions really hard to parse as a human. As
they diverge more and more it makes sense to split them into separate
manageable functions.

This is basically a mechanical change. I duplicated the existing function,
replaced all `disable_topology` with true in `no_topo` and false in the
existing function, then removed all branches which can't be hit.

Test plan:
- Runs on my dev box (6.9.0 fbkernel) with `scx_layered --run-example -n`.
- As above with `-t`.
- CI.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1214,7 +1214,8 @@ no:
 	return false;
 }
 
-void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
+static __noinline
+void layered_dispatch_no_topo(s32 cpu, struct task_struct *prev)
 {
 	s32 sib = sibling_cpu(cpu);
 	struct cpu_ctx *cctx, *sib_cctx;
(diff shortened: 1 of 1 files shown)


likewhatevs:
I *think* this is easier to implement if you rebase on #746 btw (also less verifier problems if you're running into those w/ the large boolean expressions in nested loops that pulls out).

JakeHillion:
> I _think_ this is easier to implement if you rebase on #746 btw (also less verifier problems if you're running into those w/ the large boolean expressions in nested loops that pulls out).

Sorry, bit of noise here as this is now a different PR than when it began (out of practice!). All rebased on current `main` and looks fine now.

JakeHillion:
I think this adds some warnings due to unused variables but for ease of verification I'd like to merge it as is, this way it's a very mechanical change without making any adjustments to the code.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1214,8 +1214,89 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+		    scx_bpf_consume(layer_idx))
+			return;
+	}
+
+	scx_bpf_consume(LO_FALLBACK_DSQ);
+}
+
 void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 {
+  if (disable_topology)
```
hodgesds, line 1297: just make sure this is formatted properly, otherwise LGTM

hodgesds (approved):

Issue #725, opened 2024-10-02
scx_rusty: Fix crashes on CPU hotplug


htejun:
It's easy to trigger crashes with `scx_rusty` when racing it against CPU hotplug operations. For examples, the following two backtraces are reproduced by repeatedly running the following two commands while running `scx_rusty --stats 5`:

```
root@rbw ~# for i in (seq 8 23); echo 0 > /sys/devices/system/cpu/cpu$i/online; sleep .1; end
root@rbw ~# for i in (seq 8 23); echo 1 > /sys/devices/system/cpu/cpu$i/online; sleep .1; end
```

1. monitor failing most likely due to stat reading failing on CPU which is newly offline:
```
thread '<unnamed>' panicked at scheds/rust/scx_rusty/src/main.rs:663:74:                                                                                                                                                                                                                                                                                                                       
called `Result::unwrap()` on an `Err` value: EOF while parsing a value at line 1 column 0                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                               
Stack backtrace:                                                                                                                                                                                                                                                                                                                                                                               
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from                                                                                                                                                                                                                                                                                                                     
   1: scx_stats::client::StatsClient::request                                                                                                                                                                                                                                                                                                                                                  
   2: scx_utils::misc::monitor_stats                                                                                                                                                                                                                                                                                                                                                           
   3: std::sys_common::backtrace::__rust_begin_short_backtrace                                                                                                                                                                                                                                                                                                                                 
   4: core::ops::function::FnOnce::call_once{{vtable.shim}}                                                                                                                                                                                                                                                                                                                                    
   5: std::sys::pal::unix::thread::Thread::new::thread_start                                                                                                                                                                                                                                                                                                                                   
   6: <unknown>                                                                                                                                                                                                                                                                                                                                                                                
   7: <unknown>                                                                                                                                                                                                                                                                                                                                                                                
stack backtrace:                                                                                                                                                                                                                                                                                                                                                                               
   0: rust_begin_unwind                                                                                                                                                                                                                                                                                                                                                                        
   1: core::panicking::panic_fmt                                                                                                                                                                                                                                                                                                                                                               
   2: core::result::unwrap_failed                                                                                                                                                                                                                                                                                                                                                              
```

2. invalid domain lookup:
```
[...]

PR #751, opened 2024-10-07
scx_rusty: Fix BPF crash during CPU hotplug

ryantimwilson:
This PR fixes the crash 2 in https://github.com/sched-ext/scx/issues/725

When hotplugging CPUs in rapid succession, scx_rusty would crash with:
```
scx_bpf_error (Failed to lookup dom[4294967295]
```

The root cause is if the scheduler is restarted fast enough, a task on a previously hotplugged CPU may not have moved off that CPU yet. Thus, the CPU -> domain map would contain an invalid domain (u32::max) and we would fail to lookup the domain correctly in rusty_select_cpu for prev_cpu.

To fix this, if the CPU is offline, we do not try to allocate to the same NUMA node (assuming hotplug is a rare operation) beyond domestic domain. Instead we use greedy allocation - first idle, then busy - then any CPU.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -423,9 +423,9 @@ static struct task_struct *tptr_to_task(u64 tptr)
 
 	err_task = bpf_probe_read_kernel(&p, sizeof(struct task_struct *), &task);
 
-	if (err_task) 
+	if (err_task)
 		scx_bpf_error("Failed to retrieve task_struct for tptr %llu", tptr);
-	if (p) 
+	if (p)
 		return p;
 }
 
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1004,7 +1004,19 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 		struct dom_ctx *domc;
 		struct bpf_cpumask *tmp_direct_greedy, *node_mask;
 
-		if (!(domc = lookup_dom_ctx(dom_id)))
+		/*
+		* CPU may be offline e.g. CPU was removed via hotplugging and scheduler
```
hodgesds, line 1008: nit: most of the other commends keep alignment of `*` chars

hodgesds (approved):
LGTM. I think once we get hotplug working with all the schedulers we can add that to the CI tests

scheds/rust/scx_rusty/src/bpf/main.bpf.c
ryantimwilson: ugh, can't believe i missed this

hodgesds (approved):

PR #754, opened 2024-10-08
scx_layered: Add doc comment to CpuPool

minosfuture:
Add doc comment to `CpuPool` as a quick reference for each member. Most importantly, differentiate "cpu" and "core", as logical core and physical core, respectively.


hodgesds (approved):
Awesome!

Issue #728, opened 2024-10-03
scx_bpfland: race with p->cpus_ptr


arighi:
@sirlucjan found an interesting bug with bpfland, but it can potentially affect any scheduler that uses per-CPU DSQs.

In short, under certain rare conditions, a task can be dispatched to a per-CPU DSQ,  but the task's cpumask doesn't allow to run the task on that particular CPU, hence there's a stall, because the task will never be consumed from the DSQ.

Example:
````
paź 02 22:24:03 cachyos bash[612]:   R ksoftirqd/2[34] -7430ms
paź 02 22:24:03 cachyos bash[612]:       scx_state/flags=3/0x9 dsq_flags=0x1 ops_state/qseq=0/0
paź 02 22:24:03 cachyos bash[612]:       sticky/holding_cpu=-1/-1 dsq_id=0x4 dsq_vtime=51313284698144
paź 02 22:24:03 cachyos bash[612]:       cpus=4
paź 02 22:24:03 cachyos bash[612]:     kthread+0x24c/0x290
paź 02 22:24:03 cachyos bash[612]:     ret_from_fork+0x37/0x50
paź 02 22:24:03 cachyos bash[612]:     ret_from_fork_asm+0x1a/0x30
paź 02 22:24:03 cachyos bash[612]:   R kworker/2:2[302578] -1800ms
paź 02 22:24:03 cachyos bash[612]:       scx_state/flags=3/0x9 dsq_flags=0x1 ops_state/qseq=0/0
paź 02 22:24:03 cachyos bash[612]:       sticky/holding_cpu=-1/-1 dsq_id=0x4 dsq_vtime=51322424785271
paź 02 22:24:03 cachyos bash[612]:       cpus=4
paź 02 22:24:03 cachyos bash[612]:     kthread+0x24c/0x290
paź 02 22:24:03 cachyos bash[612]:     ret_from_fork+0x37/0x50
paź 02 22:24:03 cachyos bash[612]:     ret_from_fork_asm+0x1a/0x30
````

In this case `ksoftirqd/2` was dispatched to the DSQ 4, but, according to its cpumask (`cpus=4`) it can only run on CPU 2. Same with `kworker/2` (and it's interesting to notice that in this trace even 2 kthreads have been affected by this issue).

[...]

PR #755, opened 2024-10-08
scx_bpfland: prevent per-CPU DSQ stall with per-CPU kthreads

arighi:
Since per-CPU kthreads may show an inconsistent prev_cpu and/or cpumask, dispatch them directly to local DSQ and allow to preempt the current running task.

This allows to prevent per-CPU kthread stalls and it also helps to prioritize them, as are usually important for system performance and responsiveness.

Moreover, change the behavior of --local-kthreads to prioritize all kthreads when this option is used.

This addresses issue #728.

NOTE: ideally we may want to fix this in the kernel by making sure to always expose a consistent prev_cpu and cpumask also for kthreads, but at the moment this change allows to prevent some annoying stalls and performance-wise it doesn't seem to introduce any regression. In fact, the usual gaming/fps benchmarks show even a slight improvement in responsiveness with this change applied.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -46,15 +46,12 @@ const volatile u64 slice_ns_min = 500ULL * NSEC_PER_USEC;
 const volatile s64 slice_ns_lag;
 
 /*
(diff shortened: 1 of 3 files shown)


sirlucjan (approved):
Ive tested it, no more stalls.

htejun (approved):

PR #756, opened 2024-10-08
scx_layered: Add stress-ng example layer

hodgesds:
Add a stress-ng example layer, which will be used for CI testing with stress-ng.


JakeHillion (approved):
Seems good to me! I guess we can starve people if they're using `run-example` and running `stress-ng` but I think that's fine.

likewhatevs (approved):

PR #759, opened 2024-10-08
layered: attempt to work steal from own llc before others

JakeHillion:


hodgesds (approved):
LGTM

hodgesds (approved):

JakeHillion:
I planned to set the new rotating helper function to:

```c
static __always_inline
u32 rotate_llc_id(u32 base_llc_id, u32 rotation)
{
	u32 out = base_llc_id + rotation;
	if (out >= nr_llcs)
		out -= nr_llcs;
	if (out >= nr_llcs)
		out = 0;
	return out;
}
```

With the hope it would execute quicker in the fast path. Verifies fine, but `bpf_trace_printk` shows that this isn't working as intended. I added a print before the return and the out variable is completely unaffected by the subtraction and set. The number is `>= nr_llcs`, which makes no sense, as it has been set on the previous line to 0 if that is the case.

This is potentially a bug with clang's inlining or the eBPF execution environment. Sighs.

Oh well, let's use the mod for now. It was premature optimisation anyway.

hodgesds (approved):

PR #761, opened 2024-10-08
scx_layered: fix stalls on ci

likewhatevs:
This fixes stalls on CI.

It makes checks around dispatch be more explicit (by initializing them to -1).

It also checks `cpu_hi_fallback_dsq_id` before checking for matching and open layers.

It also restores some up cleaned-up logic around preemption that that err, prevents stalls.

I tested it without each of these changes so these are the minimum to fix things.


likewhatevs:
testing automerge with this to see if like, it waits to merge like it ought to have earlier...

Issue #609, opened 2024-09-05
scx_rusty: Fix load miscalculation when machine is saturated


htejun:
rusty implements load balancer based on load sums where a task's load is defined as its `weight * duty cycle`. Nice level 0 maps to the default weight 100 and float 1.0, so a nice-0 thread which is constantly runnable has the load of 1.0. The following is from `scx_rusty --stats 1` on a Ryzen 3900x (four CCXs with 3 cores / 6 threads on each CCX, totaling 24 CPUs) with `stress -c 23` as the workload:
```
###### Wed, 4 Sep 2024 23:07:57 -0400, load balance @  -651.7ms ######
cpu=  95.74 load=   23.43 mig=0 task_err=0 lb_data_err=0 time_used= 2.0ms
tot=    344 sync_prev_idle= 0.00 wsync= 0.00
prev_idle= 0.00 greedy_idle= 0.00 pin= 0.00
dir= 0.29 dir_greedy= 0.00 dir_greedy_far= 0.00
dsq=72.38 greedy_local=27.33 greedy_xnuma= 0.00
kick_greedy= 2.91 rep=22.67
dl_clamp=69.48 dl_preset=30.23
slice=20000us
direct_greedy_cpus=038038
  kick_greedy_cpus=ffffff
  NODE[00] load= 23.43 imbal=  +0.00 delta=  +0.00
   DOM[00] load=  5.98 imbal=  +0.12 delta=  +0.00
   DOM[01] load=  5.99 imbal=  +0.13 delta=  +0.00
   DOM[02] load=  5.96 imbal=  +0.10 delta=  +0.00
   DOM[03] load=  5.50 imbal=  -0.36 delta=  +0.00
```
As `stress -c 23` creates 23 full duty cycle threads of the default weight, the load sum should be around 23 and it checks out. However, the following is the output with `stress -c 25` on the same setup:
```
###### Wed, 4 Sep 2024 23:09:16 -0400, load balance @  -738.0ms ######
cpu= 100.00 load= 3461.06 mig=0 task_err=0 lb_data_err=0 time_used= 2.1ms
tot=   1362 sync_prev_idle= 0.00 wsync= 0.00
prev_idle= 0.00 greedy_idle= 0.00 pin= 0.00
dir= 0.00 dir_greedy= 0.00 dir_greedy_far= 0.00
dsq=75.62 greedy_local=24.38 greedy_xnuma= 0.00
kick_greedy= 0.00 rep=23.42
[...]

PR #763, opened 2024-10-08
[rusty] Fix load stats when host is under-utilized

ryantimwilson:
This PR fixes https://github.com/sched-ext/scx/issues/609

When the machine is under-utilized when using scx_rusty (e.g. avg util < 99.9%), weights are ignored in the load balancer since all weights will be infeasible and therefore we should assume each task has equal weight. Thus, the default weight was assumed to be 1.

However, when the machine become >= 99.9% utilized, we end up using the real weights and correcting for infeasible weights. The default weight in the kernel is 100, so what ends up happening is we see load scale up by 100x which is a confusing experience for the user.

The fix is to set all weights = 100 in the load balancer if the machine is under-utilized so we don't see a major surprising jump in load metrics.

We end up using weight = 100 as opposed to 1 which might be more intuitive since scx_layered also shows with 100 multiplicative factor (e.g. 100 threads at full util = 10000 load). So I prefer being consistent between schedulers. Note scx_layered already handles this correctly by exporting raw load and load adjusted for infeasible weights in all cases.

### DIFF
--- a/rust/scx_utils/src/misc.rs
+++ b/rust/scx_utils/src/misc.rs
@@ -90,3 +90,25 @@ pub fn read_file_usize(path: &Path) -> Result<usize> {
         }
     }
 }
+
+/* Load is reported as weight * duty cycle
+ *
+ * In the Linux kernel, EEDVF uses default weight = 1 s.t.
+ * load for a nice-0 thread runnable for time slice = 1
+ *
+ * To conform with cgroup weights convention, sched-ext uses
+ * the convention of default weight = 100 with the formula
+ * 100 * nice ^ 1.5. This means load for a nice-0 thread
+ * runnable for time slice = 100.
+ *
+ * To ensure we report load metrics consistently with the Linux
+ * kernel, we divide load by 100.0 prior to reporting metrics.
+ * This is also more intuitive for users since 1 CPU roughly
+ * means 1 unit of load.
+ *
+ * We only do this prior to reporting as its easier to work with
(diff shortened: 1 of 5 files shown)


htejun (approved):

ryantimwilson:
Talked to @htejun offline, gonna change reported to be scaled / 100 to be consistent with kernel reporting

ryantimwilson:
@htejun I updated the PR to report load metrics / 100.0 for both scx-rusty and scx-layered (the two schedulers that report load metrics)

Note I only updated the reporting of metrics. Load metrics outside of the stats.rs modules still use weight * 100.0. I'm assuming we may want this because BPF doesn't support floating point and scx_layered updates load for layers in BPF. Let me know if you had something else in mind

rust/scx_utils/src/misc.rs
```diff
@@ -90,3 +90,25 @@ pub fn read_file_usize(path: &Path) -> Result<usize> {
...
+ *
+ * To ensure we report load metrics consistently with the Linux
+ * kernel, we divide load by 100.0 prior to reporting metrics.
+ * This is also more intuitive for users since 1 CPU roughly
+ * means 1 unit of load.
+ *
+ * We only do this prior to reporting as its easier to work with
+ * weight as integers in BPF / userspace than floating point.
+ */
+pub fn normalize_load_metric(metric: f64) -> f64 {
```
hodgesds, line 112: Thanks for adding this!

PR #764, opened 2024-10-09
scx_layered: fix ci

likewhatevs:
this PR fixes CI.

the old stress test config wasn't testing much (it was testing something, but what, idk).

the new config tests things well.

the shortest path to fixing CI is reverting the loop reorganizing I merged this morning *and* setting vng configuration to the same settings used for `test_sched` (w/o this, w/ the new config, stress OOMs on high core machines).


likewhatevs:
cc @hodgesds @JakeHillion not sure of this, but I *suspect* the other refactoring might like, run into the same issue (the test we were using to see if things worked wasn't doing anything).

PR #769, opened 2024-10-09
scx_layered: Cleanup debug messages

hodgesds:
Cleanup debug messages to use a common prefix when the scheduler is initialized.


likewhatevs (approved):
LGTM

PR #770, opened 2024-10-10
scx_layered: setup matrix job to run key paths of layered through verifier/stress

likewhatevs:
scx_layered: setup matrix job to run key paths of layered through verifier/stress

this will increase delays wrt/ ci jobs (i.e. they'll still be fast, but they're probably going to be throttled a bit).

that being said, it these cover the fixes in #768, so idk lets do it?

also updates local stress runs a bit (fixes an issue w/ header pass option and enables arg passthrough).

seems to work fine + is better than what we have now wrt/ keeping main not broken.

image

image

### DIFF
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -164,7 +164,7 @@ jobs:
         if: always()
       - run: mkdir -p ./log_save/
         if: always()
-      # no symlink following here (to avoid cycles)
+      # no symlink following here (to avoid cycle`s)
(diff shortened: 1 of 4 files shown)


likewhatevs:
gotta remember to make these all required after adding also (avoid merge when rdy doing the wrong thing)

hodgesds (approved):
LGTM! I think this will be useful for some of the other schedulers as well that have some different verification paths based on configuration.

PR #772, opened 2024-10-10
scx_bpfland: fix cpumask initialization error

arighi:
In the WAKE_SYNC path lf L3 cache awareness is disabled (--disable-l3) we may hit the following error:

  Error: EXIT: scx_bpf_error (CPU L3 cpumask not initialized)

Fix this by setting the L3 cpumask to the whole primary domain if L3 cache awareness is disabled.


1Naim:
Works as intended now.

```
```

arighi:
Merging since it's a trivial bug fix and it has been verified by the bug reporter.

PR #775, opened 2024-10-10
scx_layered: Cleanup topology preempt path

hodgesds:
Cleanup some some ugly code in the idle/preempt path to remove unneeded checks.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -700,13 +700,13 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
 		return -1;
 	}
 
-	pref_idle_cpumask = bpf_cpumask_create();
-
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -700,14 +700,14 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
 
-	pref_idle_cpumask = bpf_cpumask_create();
-
 	if (layer->idle_smt) {
 		idle_cpumask = scx_bpf_get_idle_smtmask();
 	} else {
 		idle_cpumask = scx_bpf_get_idle_cpumask();
 	}
 
+	pref_idle_cpumask = bpf_cpumask_create();
```
hodgesds, line 709: Moving this here because on older kernels the verifier loses the bpf_cpumask tracking easily.

Issue #738, opened 2024-10-04
scx_layered: Implement a reverse layer weight DSQ algorithm


hodgesds:
As part of introducing layer weights in #735 a new interface is added for [implementing different algorithms](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/main.bpf.c#L119) for how DSQs should be iterated across layers. One algorithm that has not been implemented is a [reverse weight iteration](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/main.bpf.c#L131). The goal is to implement an algorithm to iterate across the layers in order of highest weight to lowest weight.

PR #777, opened 2024-10-10
[layered] Implement reverse weight DSQ algorithm

ryantimwilson:
This PR addresses https://github.com/sched-ext/scx/issues/738 by implementing a reverse weight iteration algorithm in scx-layered. This means we iterate through DSQs from layers with the highest weight to lowest weight. The goal is to avoid stalls that happen from lower weighted layers consuming all available resources. Although the opposite problem could happen here, so this is just a temporary algorithm until we have more robust load balancing across layers.


hodgesds (approved):
LGTM, just in time for a refactor 😄

PR #778, opened 2024-10-10
[layered] rename util.bpf.c -> util.bpf.inc for clarity

JakeHillion:
'util.bpf.c' isn't a translation unit in its own right, and is instead expected
to be included in exactly one position in 'main.bpf.c'. Rename it '.bpf.inc' so
this is clear to the end user/build systems.


hodgesds:
I think if you want to break this up you would have a `util.bpf.c` and a `util.bpf.h` and then include the `util.bpf.h` in the `main.bpf.c`.

likewhatevs:
I would strongly prefer we make this a `.h` file to solve that issue wrt/ other build systems (.h works w/ all of them and dev tooling etc.).

wrt/ why:

bpf.c or bpf.h -- is a c file (albeit w/o proper ide config etc., yet)
WezTerm 2024-10-10 22 28 06

bpf.inc -- is a txt file.
WezTerm 2024-10-10 22 27 57

.inc is gonna be treated as a text file w/o special configs/overrides in editors etc. and (presumably) additional fighting w/ clangd once that is done.

The direction I'd hope this file would go in (err, that I should do sometime tbh, lol) is to add a `#ifdef LSP` block to add the necessary imports to enable clangd etc. to work on it (and maybe making it .h to make use in other build systems etc. simpler), but I haven't added that yet because I haven't needed to edit it.

likewhatevs:
For example, I would love to see things go in this direction (because it would make development a lot easier)
CleanShot 2024-10-10 at 22 40 05@2x

JakeHillion:
> I think if you want to break this up you would have a `util.bpf.c` and a `util.bpf.h` and then include the `util.bpf.h` in the `main.bpf.c`.

Yeah, headers/separately compiled TUs is the appropriate solution here. It involves (I believe) changing the `libbpf` loading code to load multiple compiled objects, potentially in the correct order, and linking them, which seems like a lot more work than changing the extension. It's the technically correct way to go but the overhead is kind of high, and I'm not sure we get a big win from it over including text content. I do want to split into more files though.

> I would strongly prefer we make this a `.h` file to solve that issue wrt/ other build systems (.h works w/ all of them and dev tooling etc.).

I'm pretty hard against a `.h` file here - it's not a `.h` file, as including it in multiple TUs would break (multiple definitions of the non-static `match_prefix`, and probably the maps too). To me this is quite clearly a broken header, as if you end up with more files and multiple includes of the header things break.

I'm less sure about the IDE breakage, mine still renders it as C with no language server but I don't know how common that is. I don't think it's valid for an IDE to treat this as a `.bpf.c` file as is though - it's missing headers like `#include <bpf/bpf_helpers.h>` for the `bpf_for` it uses (because it's not really a `.bpf.c` file) so I'm surprised anything comes out of running a language server on it...

--

I still think this is the best short term option, unless we do whatever `libbpf` work is required to ship multiple objects which doesn't seem worth it at this point. I've seen `.inc` files elsewhere for this "I'm including a blob of text that will be parsed as C in this one specific place", which is better than including another `.bpf.c` file that looks like it should compile independently.

likewhatevs:
> I think if you want to break this up you would have a util.bpf.c and a util.bpf.h and then include the util.bpf.h in the main.bpf.c.

This lets this file work for everyone using an ide/fancy editor w/o having to know how to make clangd work also, so I like this as much as .h or no change fwiw.

likewhatevs:
> I'm pretty hard against a .h file here - it's not a .h file, as including it in multiple TUs would break (multiple definitions of the non-static match_prefix, and probably the maps too). To me this is quite clearly a broken header, as if you end up with more files and multiple includes of the header things break.

Ahh btw, making this a `.h` works and doesn't break other build systems etc. (i.e. I just reference this file with a .h extension and symlink .h to .c).

likewhatevs:
Added an example of what the change to .inc costs here: https://github.com/sched-ext/scx/pull/782 , that also makes ide's etc. work well on that file if we choose to go w/ that.

WezTerm 2024-10-11 08 06 51

I'm certain lsp is better than clarity of naming wrt/ ppl editing these files, but that being said, so long as we don't start doing this to files in common (which would impact all scheduler impls), I don't have that strong of an opinion (i.e. one of the goals of scx is ease of use/development, if any particular scheduler is harder to work on, that's fine, if all are, that's an issue).

htejun:
Would something like `.inc.c` work? My preference would be making it either `.h` or `.c`. I don't think not being equal to usual `.c` or `.h` is an actual problem. BPF is already way weirder in more serious ways.

likewhatevs:
> Would something like .inc.c work? My preference would be making it either .h or .c.

Wrt/ tooling and go-to-definition etc., yes.

JakeHillion:
> Would something like `.inc.c` work? My preference would be making it either `.h` or `.c`. I don't think not being equal to usual `.c` or `.h` is an actual problem. BPF is already way weirder in more serious ways.

Yeah it could do, though I'll have to update the tooling again. For now the import can change it to a `.inc` instead of a `.h` which won't cause collisions with upstream if we do add a `util.bpf.h` that's an actual header, and that can be carried internally for as long as needs be.

Will close this for now, and hopefully follow up in the future with the multi-file loading support/proper headers.

PR #780, opened 2024-10-11
scx_bpfland: drop per-cpu DSQs

arighi:
Get rid of the per-CPU DSQ logic, since it seems to add more issues than benefit.

With this set of changes bpfland seems to be more stable and it also shows better performance in terms of responsiveness (according to my benchmarks and feedback from the CachyOS community).

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -1059,29 +1059,12 @@ void BPF_STRUCT_OPS(bpfland_dispatch, s32 cpu, struct task_struct *prev)
 		return;
 
 	/*
-	 * If the current task expired its time slice, its CPU is still a
-	 * full-idle SMT core and no other task wants to run, simply replenish
-	 * its time slice and let it run for another round on the same CPU.
-	 *
-	 * Note that bpfland_stopping() won't be called if we replenish the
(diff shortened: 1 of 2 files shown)


1Naim:
```
```

sirlucjan:
```
Tested-by: Piotr Gorski <piotrgorski@cachyos.org>
```

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -581,11 +581,11 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags)
...
 	/*
 	 * If the task isn't allowed to use its previously used CPU it means
-	 * that it's changing affinity. In this case just pick a random CPU in
-	 * its new allowed CPU domain.
+	 * that it's changing affinity. In this case try to pick any random
+	 * idle CPU in its new allowed CPU domain.
 	 */
 	if (!bpf_cpumask_test_cpu(prev_cpu, p->cpus_ptr))
-		prev_cpu = bpf_cpumask_any_distribute(p->cpus_ptr);
+		return scx_bpf_pick_idle_cpu(p->cpus_ptr, 0);
```
multics69, line 470: I wonder if the error return of scx_bpf_pick_idle_cpu() needs to be considered (e.g., -EBUSY?).

multics69 (approved):

PR #782, opened 2024-10-11
layered -- make lsp work nice on util include file

likewhatevs:
This is what I was talking about on https://github.com/sched-ext/scx/pull/778

WezTerm 2024-10-11 08 05 49

It makes it nice to edit that file, and most folks editing this file get this benefit because of https://github.com/likewhatevs/scx/blob/main/.clangd


hodgesds (approved):
LGTM

PR #787, opened 2024-10-11
scx_layered: Cleanup non topology path

hodgesds:
More cleanup in the non topology path to remove copy/pasta declarations.


likewhatevs (approved):
LGTM

PR #790, opened 2024-10-12
scx_lavd: do not inspect scx_lavd process itself

multics69:
Print the task status of scx_lavd is not useful, so filter it out.


arighi (approved):
Maybe save the pid and check for that? Otherwise we may skip tasks that are named scx_lavd* (intentionally or not). But overall makes sense.

multics69:
I also considered checking pid but I thought that approach does not have benefit. Since there are maximum two (or more) scx_lavd processes, that is more expensive than comparing 8 bytes.

Issue #734, opened 2024-10-04
Fix build issue that `struct user_pt_regs` is not found on aarch64


minosfuture:
A compile issue that it is unable to find the definition of `struct user_pt_regs` when compiling `scx_rlfifo` and `scx_rustland` (they use BPF_KPROBE  and thus requires `user_pt_regs`).

`struct user_pt_regs` indeed is not included in `scheds/include/vmlinux/vmlinux.h`.

Regenerating `vmlinux.h` from the linux upstream `88264981f20` can solve the problem.

This is likely an issue only on aarch64. Generating vmlinux.h during compilation, as bpftool repo does, may be a solution.

Log:
```bash
:~/scx$ cargo build                                                                                                                                                           
warning: /home/debian/scx/rust/scx_rustland_core/Cargo.toml: unused manifest key: lib.include                                                                                              
   Compiling scx_rustland_core v2.1.2 (/home/debian/scx/rust/scx_rustland_core)                                                                                                            
   Compiling scx_stats_derive v1.0.5 (/home/debian/scx/rust/scx_stats/scx_stats_derive)                                                                                                    
   Compiling scx_bpfland v1.0.5 (/home/debian/scx/scheds/rust/scx_bpfland)                   
   Compiling scx_lavd v1.0.5 (/home/debian/scx/scheds/rust/scx_lavd)                                                                                                                       
   Compiling scx_rusty v1.0.5 (/home/debian/scx/scheds/rust/scx_rusty)                                                                                                                     
   Compiling scx_loader v1.0.5 (/home/debian/scx/rust/scx_loader)                                                                                                                          
   Compiling scx_rustland v1.0.5 (/home/debian/scx/scheds/rust/scx_rustland)                                                                                                               
   Compiling scx_rlfifo v1.0.5 (/home/debian/scx/scheds/rust/scx_rlfifo)                     
error: failed to run custom build command for `scx_rlfifo v1.0.5 (/home/debian/scx/scheds/rust/scx_rlfifo)`                                                                                
                                                                                                                                                                                           
Caused by:                                                                                                                                                                                 
  process didn't exit successfully: `/home/debian/scx/target/debug/build/scx_rlfifo-1aec066430c5fe0f/build-script-build` (exit status: 101)
  --- stdout                                                                                 
[...]

PR #793, opened 2024-10-13
Use per-arch vmlinux.h

minosfuture:
Address #734 

* Use per-arch vmlinux.h
	* vmlinux.h is not compatible across archs.
	* Handle this compatibility issue by
		* Add arch info into vmlinux.h real file name
		* Link vmlinux.h to the target-arch real file at build time
		* Use target-arch real file for scx_utils bindgen.
	* Also refactored clang related logic into a new clang_info mod, which is shared by bpf_builder.rs and builder.rs.
* Add vmlinux.h for multiple arch
	* Following the change of using per-arch vmlinux.h. Add it for the remaining archs.
* scx_mitosis: Fix static assertion of scx_bpf_task_cgroup failing __weak check
	* it failed the static assertion in macro bpf_ksym_exists.


likewhatevs:
I don't totally follow why those build failures are occurring, but like, I know all need to pass.

minosfuture:
vmlinux.h for non-aarch64 ones are incorrect. Marking this PR WiP for now.

rust/scx_utils/src/builder.rs
```diff
@@ -33,8 +37,37 @@ impl Builder {
...
+                return Ok(entry.path().to_string_lossy().into_owned());
+            }
+        }
+        Err(anyhow!("vmlinux.h for arch {} is not found", kernel_target))
+    }
+
     fn gen_bindings(&self) {
         let out_dir = env::var("OUT_DIR").unwrap();
+        let clang = ClangInfo::new().unwrap();
+        let vmlinux_dir = Path::new(&BPF_H)
```
likewhatevs, line 62: I think I like this (and all signals green). 

@hodgesds what do you think (roughly having build.rs create the appropriate symlink for each platform). 

It might make some editor stuff work less well but like, worst case anyone bothered by that could just symlink their platform's `.h` to the right one in their src dir and this solves a real problem sooooo idk?

arighi: > If we go w/ this, I think I'd try this out: https://github.com/arighi/virtme-ng/releases/tag/v1.30 . In theory, it shouldn't be that tricky expanding out test suite to cover arm.

I've uploaded the new virtme-ng version on pypi, so I think we're already using v1.30 in our CI.

minosfuture:
gentle ping for review. @htejun @hodgesds

likewhatevs:
@minosfuture tied up w/ layered issues atm. I have a pi running 6.12-rc3 I mean to test this on though.

rust/scx_utils/Cargo.toml
```diff
@@ -37,6 +37,12 @@ bindgen = ">=0.68, <0.70"
 tar = "0.4"
 vergen = { version = "8.0.0", features = ["cargo", "git", "gitcl"] }
 walkdir = "2.4"
+anyhow = "1.0.65"
+glob = "0.3"
+libbpf-cargo = "0.24.1"
+sscanf = "0.4"
+lazy_static = "1.4"
+version-compare = "0.1"
```
htejun, line 45: Can we keep these in alphabetic order?

htejun (approved):

rust/scx_utils/Cargo.toml
htejun: We can clean up later. Will merge as-is. Thanks a lot for working on this!

htejun:
@minosfuture I reverted the first two commits to fix build for C scheds. @likewhatevs, we probalby should include building C scheds in CI.

I think an easier approach here would be putting per-arch `vmlinux.h` files under their own directory - ie. name them `scheds/include/$ARCH/vmlinux-$KVER-$SHA1.h` which is statically symlinked to `scheds/include/$ARCH/vmlinux.h` and then make the build scripts add the per-arch directory as an `-I` option. That way we don't have to create something stateful during build and the only thing build scripts (both cargo and meson) do is just adding the right per-arch include directory.

minosfuture:
@htejun I'll give it a try. It does sound simpler that no extra step of creating the symlink is needed during both meson and cargo build.

Sorry about the regression. I'll open another PR for this.

PR #796, opened 2024-10-14
scripts: Convert sched ftrace helper scripts to python

arighi:
Merge the sched_switch ftrace helper scripts into a single python script that prints the result to stdout.

In this way it's possible to generate a perfetto-compatible trace running:

 $ sudo ./scripts/sched_ftrace.py > sched.ftrace


multics69 (approved):

hodgesds (approved):
Good idea!

PR #802, opened 2024-10-15
scx_rusty: Return null pointer when task retrievement failed

vax-r:
## Summary
Return a NULL pointer when task "p" retrievement failed inside "tptr_to_task()" to avoid compilation error.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -263,7 +263,7 @@ static u64 calc_freq_factor(u64 freq, u64 weight_ft)
 }
 
 static u64 calc_weight_factor(struct task_struct *p, struct task_ctx *taskc,
-			      struct cpu_ctx *cpuc_cur, u64 enq_flags)
+			      u64 enq_flags)
 {
 	u64 weight_boost = 1;
 	u64 weight_ft;
@@ -283,10 +283,10 @@ static u64 calc_weight_factor(struct task_struct *p, struct task_ctx *taskc,
 		weight_boost += LAVD_LC_WEIGHT_BOOST;
 
 	/*
-	 * Reset task's lock and futex boost count
-	 * for a lock holder to be boosted only once.
+	 * A pinned-task tends to be latency-critical.
 	 */
-	reset_lock_futex_boost(taskc, cpuc_cur);
+	if (p->nr_cpus_allowed == 1)
+		weight_boost += LAVD_LC_WEIGHT_BOOST;
 
 	/*
 	 * Prioritize a lock holder for faster system-wide forward progress.
@@ -301,7 +301,7 @@ static u64 calc_weight_factor(struct task_struct *p, struct task_ctx *taskc,
 }
 
 static u64 calc_lat_cri(struct task_struct *p, struct task_ctx *taskc,
-			struct cpu_ctx *cpuc_cur, u64 enq_flags)
+			u64 enq_flags)
 {
 	u64 weight_ft, wait_freq_ft, wake_freq_ft, runtime_ft;
 	u64 lat_cri;
@@ -310,7 +310,7 @@ static u64 calc_lat_cri(struct task_struct *p, struct task_ctx *taskc,
 	 * Adjust task's weight based on the scheduling context, such as
 	 * if it is a kernel task, lock holder, etc.
 	 */
-	weight_ft = calc_weight_factor(p, taskc, cpuc_cur, enq_flags);
+	weight_ft = calc_weight_factor(p, taskc, enq_flags);
 
 	/*
 	 * A task is more latency-critical as its wait or wake frequencies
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -425,8 +425,7 @@ static struct task_struct *tptr_to_task(u64 tptr)
 
 	if (err_task)
 		scx_bpf_error("Failed to retrieve task_struct for tptr %llu", tptr);
-	if (p)
-		return p;
+	return p ? p : NULL;
```
hodgesds, line 428: Silly question, why are either of these needed? If `bpf_profile_read` fails it should hit the `scx_bpf_error` (which maybe should return null) and then it should be fine to return `p` always.

vax-r: Yeah you're right, I think `return p` directly is enough, failures should be catch by line 426.
I'll make an update and try to test the number of branches we can save from this.

vax-r:
@hodgesds -
I remove the checking of `p` for `tptr_to_task`, after that I did a profiling for the system's branch-misses , and the result are showned as the following ( while running kernel compilation as workload )
```
// before the change
$ sudo bpftrace -e 'hardware:branch-misses { @[comm] = count(); } interval:s:60 { exit(); }'
...
@[scx_rusty]: 30
...
// after the change
$ sudo bpftrace -e 'hardware:branch-misses { @[comm] = count(); } interval:s:60 { exit(); }'
...
@[scx_rusty]: 27
...
```
Tbh there doesn't seems to have much difference in terms of branch-misses.
I am wondering the reason , shouldn't the code change reduce the number of branches and thus reduce the chance to have branch-misses? or in fact the chance of this code path isn't frequent enough to see the difference ?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -437,11 +436,6 @@ int dom_xfer_task(u64 tptr, u32 new_dom_id, u64 now)
 
 	p = tptr_to_task(tptr);
 
-	if (!p) {
```
hodgesds, line 440: I don't think removing this makes sense. The only place `dom_xfer_task‎` is called is in `task_set_domain`, which already does the lookup of the task. So the `struct task_struct *p` should probably just be passed as a parameter to this function along with the `struct task_ctx`. cc @likewhatevs

hodgesds:
> Tbh there doesn't seems to have much difference in terms of branch-misses.
> I am wondering the reason , shouldn't the code change reduce the number of branches and thus reduce the chance to have branch-misses? or in fact the chance of this code path isn't frequent enough to see the difference ?

I think your bpftrace script is profiling/reporting the userspace part of rusty and not the bpf code, which would be unrelated to this change.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
likewhatevs: > I don't think removing this makes sense. The only place dom_xfer_task‎ is called is in task_set_domain, which already does the lookup of the task. So the struct task_struct *p should probably just be passed as a parameter to this function along with the struct task_ctx.

If that passes the verifier, I think that makes the most sense. That being said, p has always been an external lookup (previously `bpf_task_from_pid`, now `bpf_probe_read_kernel`), and IIRC/IIUC that's key to getting this bit of code passing the verifier (at least it was when I last edited it).

WRT/ this change as it is now, I think it works and works fine *but* it's kinda relying on `lookup_task_ctx` to throw an error to do the right thing (which it does now), but I'm not super sure that reliance is a great thing (i.e. better to explicitly check and throw for when that changes IMO).

I guess I'm not super opposed to this change as is if like there's a reason (like idk making things run on aarch64 or something) for it, but idk of any such reason atm. 

@vax-r , what compilation error is this fixing (i.e. CI passes without this)?

vax-r: @likewhatevs -
sorry for the late reply, I'm not sure why some compiling errors and warnings keep popping up while CI doesn't show those warnings. For the part I'm trying to fix , the compilation message is `src/bpf/main.bpf.c:430:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]`, after I execute `meson compile -C build` from scx root directory.

likewhatevs: I recommend going with the approach @hodgesds recommended or a way without the downsides I mentioned above (because these issues aren't on CI yet).

vax-r: > I recommend going with the approach @hodgesds recommended or a way without the downsides I mentioned above (because these issues aren't on CI yet).

No problem , I understand you after some further study and I restore the part of the checking within `dom_xfer_task()` , task a look while you're available, thanks.

hodgesds (approved):

hodgesds:
I think you need to rebase

PR #806, opened 2024-10-16
scx_layered: emit more specific errors when bpf task storage map is missing entries

likewhatevs:
do not throw errors when bpf task storage map is missing entries

the idea behind using a task storage map is it handles removing keys when processes go away. 

when things disappear from this map this is not an error/we should continue without error (because the task is gone).

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -393,12 +393,12 @@ static struct task_ctx *lookup_task_ctx_may_fail(struct task_struct *p)
 	return bpf_task_storage_get(&task_ctxs, p, 0, 0);
 }
 
-static struct task_ctx *lookup_task_ctx(struct task_struct *p)
+static struct task_ctx *lookup_task_ctx(struct task_struct *p, const char caller[])
 {
 	struct task_ctx *tctx = lookup_task_ctx_may_fail(p);
 
 	if (!tctx)
-		scx_bpf_error("task_ctx lookup failed");
+		scx_bpf_error("task_ctx lookup failed: %s", caller);
 
 	return tctx;
 }
@@ -453,7 +453,7 @@ int BPF_PROG(tp_cgroup_attach_task, struct cgroup *cgrp, const char *cgrp_path,
 			break;
 		}
 
-		if ((tctx = lookup_task_ctx(next)))
+		if ((tctx = lookup_task_ctx(next, __func__)))
 			tctx->refresh_layer = true;
 	}
 
@@ -728,7 +728,7 @@ s32 BPF_STRUCT_OPS(layered_select_cpu, struct task_struct *p, s32 prev_cpu, u64 
 	struct layer *layer;
 	s32 cpu;
 
-	if (!(cctx = lookup_cpu_ctx(-1)) || !(tctx = lookup_task_ctx(p)))
+	if (!(cctx = lookup_cpu_ctx(-1)) || !(tctx = lookup_task_ctx(p, __func__)))
 		return prev_cpu;
 
 	/*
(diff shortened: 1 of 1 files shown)


likewhatevs:
I think this reasoning is sound but not 100% on it fwiw.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -393,7 +393,7 @@ static struct task_ctx *lookup_task_ctx_may_fail(struct task_struct *p)
 	return bpf_task_storage_get(&task_ctxs, p, 0, 0);
 }
 
-static struct task_ctx *lookup_task_ctx(struct task_struct *p)
+static struct task_ctx *lookup_task_ctx_kay_fail(struct task_struct *p)
```
hodgesds, line 396: Who's kay and why did they fail? 😆

likewhatevs: lol typo in the sed when I got tired of editing. 

funnily enough, the verifier was happy with and without the typo's.

likewhatevs:
finally got the typos and removed (now) dead code (I don't think that lookup failing should ever be an error given the semantics of TASK_MAP).

htejun:
In the sched paths, this lookup shouldn't fail and we may be masking underlying bugs if we ignore lookup failures. We may decide to paper over these failures for practical purposes (I would still put it behind an option) but we should understand exactly why the failure happened.

likewhatevs:
IDK if this is reasonable, but I kind of think it is what we should do (emit a specific error every time we call lookup instead of a less specific one).

Instead of hiding errors, provide more specific ones so that we can more easily figure out why this is occurring.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -395,12 +395,12 @@ static struct task_ctx *lookup_task_ctx_may_fail(struct task_struct *p)
...
-	struct task_ctx *tctx = lookup_task_ctx_may_fail(p);
+    struct task_ctx *tctx = lookup_task_ctx_may_fail(p);
 
-	if (!tctx)
-		scx_bpf_error("task_ctx lookup failed");
+    if (!tctx)
+        scx_bpf_error("task_ctx lookup failed");
 
-	return tctx;
+    return tctx;
```
htejun, line 403: I think currently the C files are using tabs for indentation. Probably best to stay consistent.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -466,8 +470,12 @@ SEC("tp_btf/task_rename")
 int BPF_PROG(tp_task_rename, struct task_struct *p, const char *buf)
 {
 	struct task_ctx *tctx;
+	
+	tctx = lookup_task_ctx_may_fail(p);
+	if(!tctx)
+			scx_bpf_error("task_ctx not found in tp_task_rename");
```
htejun, line 476: Shouldn't this be already visible in the backtrace which is captured by scx_bpf_error()?

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: ahh you know I keep tripping up over that, will fix but will also see if I can add an config so editors etc. do the right thing. knowing tab not space helps, thx.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: Kind of.  What we get is this:

```
sched_ext: scx_bpf_error (task_ctx lookup failed) 
```

*but*, we can use that to get a full stacktrace easily enough.

If we think this isn't going to be a recurring issue, maybe we don't make this change and just use the stacktraces to fix this issue and having logs like that is unnecessary.

That being said, task map seems like the kind of thing that will make this a recurring issue (so maybe make this easier to debug in the future). IDK tbh.

htejun: The full error message should contain the backtrace. If it's not there, we're just not capturing it right.

likewhatevs: We are definitely not capturing it right. The thinking here is to add a stop-gap until then (because that's probably a good way's off (i.e. a refactor of how we are doing logging and metrics exporting away)) *if* we think this is going to be a recurring issue.

I think this will be a recurring issue but it's hard to tell. I updated this w/ a cleaner way to get logs like that in the event that logging this better is like the right balance, but it could be that leaving the logs as they are now is the right balance, idk.

PR #807, opened 2024-10-16
scx_layered: Add decayed vtime for starvation prevention

hodgesds:
Add a decayed vtime that is used to calculate a weighted decayed vtime based on layer weights. The weighted decayed vtime is then used to determine if a layer is being starved. When a layer is determined to be started it will prioritized for DSQ consumption. Thanks to @JakeHillion for helping with the implementation.

Tested with/without topology awareness by running `nproc * 2` of `stress-ng` for 35 seconds with no scheduler ejections. On the `main` branch the scheduler will eject if running `stress-ng -c $(nproc)`.
```
$ sudo stress-ng -c $(($(nproc) * 2)) -t 35
stress-ng: info:  [1165012] setting to a 35 secs run per stressor
stress-ng: info:  [1165012] dispatching hogs: 160 cpu
stress-ng: info:  [1165012] skipped: 0
stress-ng: info:  [1165012] passed: 160: cpu (160)
stress-ng: info:  [1165012] failed: 0
stress-ng: info:  [1165012] metrics untrustworthy: 0
stress-ng: info:  [1165012] successful run completed in 35.09 secs
$ stress-ng -c $(($(nproc) * 2)) -t 35
stress-ng: info:  [1052981] setting to a 35 secs run per stressor
stress-ng: info:  [1052981] dispatching hogs: 160 cpu
stress-ng: info:  [1052981] skipped: 0
stress-ng: info:  [1052981] passed: 160: cpu (160)
stress-ng: info:  [1052981] failed: 0
stress-ng: info:  [1052981] metrics untrustworthy: 0
stress-ng: info:  [1052981] successful run completed in 35.27 secs
$ sudo stress-ng -c $(($(nproc) * 2)) -t 35
stress-ng: info:  [1061554] setting to a 35 secs run per stressor
stress-ng: info:  [1061554] dispatching hogs: 160 cpu
stress-ng: info:  [1061554] skipped: 0
stress-ng: info:  [1061554] passed: 160: cpu (160)
stress-ng: info:  [1061554] failed: 0
stress-ng: info:  [1061554] metrics untrustworthy: 0
stress-ng: info:  [1061554] successful run completed in 37.44 secs
```

Performance numbers:
`main`:
```
$ sudo stress-ng -c 10 -t 20 -M
stress-ng: info:  [1244923] setting to a 20 secs run per stressor
stress-ng: info:  [1244923] dispatching hogs: 10 cpu
stress-ng: metrc: [1244923] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s CPU used per       RSS Max
stress-ng: metrc: [1244923]                           (secs)    (secs)    (secs)   (real time) (usr+sys time) instance (%)          (KB)
stress-ng: metrc: [1244923] cpu              259307     20.00    199.49      0.04     12963.66        1299.58        99.75          5120
stress-ng: info:  [1244923] skipped: 0
stress-ng: info:  [1244923] passed: 10: cpu (10)
stress-ng: info:  [1244923] failed: 0
stress-ng: info:  [1244923] metrics untrustworthy: 0
stress-ng: info:  [1244923] successful run completed in 20.01 secs
```
branch:
```
$ sudo stress-ng -c 10 -t 20 -M
stress-ng: info:  [1251999] setting to a 20 secs run per stressor
stress-ng: info:  [1251999] dispatching hogs: 10 cpu
stress-ng: metrc: [1251999] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s CPU used per       RSS Max
stress-ng: metrc: [1251999]                           (secs)    (secs)    (secs)   (real time) (usr+sys time) instance (%)          (KB)
stress-ng: metrc: [1251999] cpu              256058     20.00    199.37      0.05     12801.52        1283.99        99.70          5120
stress-ng: info:  [1251999] skipped: 0
stress-ng: info:  [1251999] passed: 10: cpu (10)
stress-ng: info:  [1251999] failed: 0
stress-ng: info:  [1251999] metrics untrustworthy: 0
stress-ng: info:  [1251999] successful run completed in 20.01 secs
```
The variance of a run is a few hundred ops/sec so there is little/no noticeable performance impact.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -181,6 +181,7 @@ struct layer {
 	int			growth_algo;
 
 	u64			vtime_now;
+	u64			vtime_decay;
 	u64			nr_tasks;
 
 	u64			load;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -99,6 +99,8 @@ u32 rotate_llc_id(u32 base_llc_id, u32 rotation)
 // return the dsq id for the layer based on the LLC id.
 static __noinline u64 layer_dsq_id(u32 layer_id, u32 llc_id)
 {
+	if (disable_topology)
+		return layer_id;
 	return (layer_id * nr_llcs) + llc_id;
 }
 
@@ -236,6 +238,17 @@ static void lstat_add(enum layer_stat_idx idx, struct layer *layer,
 		scx_bpf_error("invalid layer or stat idxs: %d, %d", idx, layer->idx);
 }
 
+static void lstat_set(enum layer_stat_idx idx, struct layer *layer,
+		      struct cpu_ctx *cctx, s64 val)
+{
+	u64 *vptr;
+
+	if ((vptr = MEMBER_VPTR(*cctx, .lstats[layer->idx][idx])))
+		(*vptr) = val;
+	else
+		scx_bpf_error("invalid layer or stat idxs: %d, %d", idx, layer->idx);
+}
+
 static void lstat_inc(enum layer_stat_idx idx, struct layer *layer,
 		      struct cpu_ctx *cctx)
 {
@@ -356,14 +369,26 @@ static void refresh_cpumasks(int idx)
 	trace("LAYER[%d] now has %d cpus, seq=%llu", idx, layer->nr_cpus, layer->cpus_seq);
 }
 
+static void decay_layer_vtime(int layer_idx)
+{
+	struct layer *layer;
+
+	layer = MEMBER_VPTR(layers, [layer_idx]);
+	// TODO: Make the decay value configurable
+	layer->vtime_decay = (layer->vtime_decay / 8) * 7;
+	trace("LAYER[%d] vtime_decay: %d", layer_idx, layer->vtime_decay);
+}
+
 SEC("fentry")
 int BPF_PROG(sched_tick_fentry)
 {
 	int idx;
 
 	if (bpf_get_smp_processor_id() == 0)
-		bpf_for(idx, 0, nr_layers)
+		bpf_for(idx, 0, nr_layers) {
 			refresh_cpumasks(idx);
+			decay_layer_vtime(idx);
+		}
 	return 0;
 }
 
@@ -1200,6 +1225,56 @@ no:
 	return false;
 }
 
+/*
+ * Helper function to consume layer DSQs when a system is under starvation.
+ */
+static __always_inline bool consume_starved(s32 cpu)
+{
+	struct layer *layer;
+	u32 idx;
+	u64 decay_vtime_sum = 0;
+	u64 layer_weight_sum = 0;
+	u64 decay_proportion_vtime = 0;
+	u64 vtime_weighted_decay = 0;
+
+	bpf_for(idx, 0, nr_layers) {
+		layer = MEMBER_VPTR(layers, [idx]);
+		decay_vtime_sum += layer->vtime_decay;
+		layer_weight_sum += layer->weight;
+	}
+
+	// Use a per cpu layer to consume from. With enough CPUs this
+	// should consume starved layers efficiently.
+	// TODO: Make this work for nr_layers > nr_possible_cpus
+	idx = ((u32)cpu & 0xfff) % nr_layers;
+	if (idx > nr_layers)
+		return false;
+
+	layer = MEMBER_VPTR(layers, [idx]);
+	if (!layer)
+		return false;
+
+	if (layer_weight_sum > decay_vtime_sum)
+		return false;
+
+	/*
+	 * If the weighted decayed vtime of the layer is less than the
+	 * proportional weighted vtime then the layer is potentially starved
+	 * and should be consumed from.
+	 */
+	vtime_weighted_decay = decay_vtime_sum / layer_weight_sum;
+	decay_proportion_vtime = vtime_weighted_decay * layer->weight;
+
+	if (layer->vtime_decay <= decay_proportion_vtime) {
+		u32 llc_id = cpu_to_llc_id(cpu);
+		u64 dsq_id = layer_dsq_id(layer->idx, llc_id);
+		if ( scx_bpf_consume(dsq_id))
+			return true;
+	}
+
+	return false;
+}
+
 static __noinline
 void layered_dispatch_no_topo(s32 cpu, struct task_struct *prev)
 {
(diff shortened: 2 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -356,14 +369,26 @@ static void refresh_cpumasks(int idx)
 	trace("LAYER[%d] now has %d cpus, seq=%llu", idx, layer->nr_cpus, layer->cpus_seq);
 }
 
+static void decay_layer_vtime(int layer_idx)
+{
+	struct layer *layer;
+
+	layer = MEMBER_VPTR(layers, [layer_idx]);
+	// TODO: Make the decay value configurable
```
likewhatevs, line 377: IIUC this configurable is going to be how you let less-important tasks sit for say 20 seconds when saturated, right?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1238,6 +1313,13 @@ void layered_dispatch_no_topo(s32 cpu, struct task_struct *prev)
 		return;
 	}
 
+	dsq_id = cpu_hi_fallback_dsq_id(cpu);
```
htejun, line 1316: Why consume hi fallback before preempting layers? Or rather, shouldn't fallback layers also have a weight and mechanism to avoid starvation?

scheds/rust/scx_layered/src/bpf/main.bpf.c
JakeHillion: So there are two parts to this. This decay value is effectively how much history you're considering when deciding whether a task is underserved, this will likely be a whole scheduler argument rather than layer specific, I don't see much benefit to considering different amounts of history for each layer.

The other configurable that we didn't add is in `consume_starved`, where we compare directly to `decay_proportion_time`.  It should be possible to add a fudge factor there per layer that decides how aggressively each layer is scheduled when underserved, but I'm not sure what physical quantity it represents. Probably some sort of variation on "time to wait before being served".

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -236,6 +238,17 @@ static void lstat_add(enum layer_stat_idx idx, struct layer *layer,
...
+		      struct cpu_ctx *cctx, s64 val)
+{
+	u64 *vptr;
+
+	if ((vptr = MEMBER_VPTR(*cctx, .lstats[layer->idx][idx])))
+		(*vptr) = val;
+	else
+		scx_bpf_error("invalid layer or stat idxs: %d, %d", idx, layer->idx);
+}
+
```
JakeHillion, line 251: Is this used? I can't see it in this diff.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -181,6 +181,7 @@ struct layer {
 	int			growth_algo;
 
 	u64			vtime_now;
+	u64			vtime_decay;
```
htejun, line 184: This might be okay for now but `vtime_now` and `vtime_decay` are likely to make this cacheline very hot which can become a scalability problem for layers e.g. spanning multiple nodes or otherwise a large number of CPUs.

Ideally, everything that can be split per-DSQ should be per-DSQ. There's no reason to keep layer-wide `vtime_now`. This can be per-DSQ as long as delta translation is applied whenever tasks move from one DSQ to another.

As for `vtime_decay`, it's a bit trickier as it is something which needs to track global usage. However, this can also be kept per-DSQ and summed up periodically.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1200,6 +1225,56 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+	u32 idx;
+	u64 decay_vtime_sum = 0;
+	u64 layer_weight_sum = 0;
+	u64 decay_proportion_vtime = 0;
+	u64 vtime_weighted_decay = 0;
+
+	bpf_for(idx, 0, nr_layers) {
+		layer = MEMBER_VPTR(layers, [idx]);
+		decay_vtime_sum += layer->vtime_decay;
+		layer_weight_sum += layer->weight;
```
JakeHillion, line 1243: We did mention extracting this to the Rust, but it's not a big deal to do it here. Adding for reference.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1200,6 +1225,56 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
 	return false;
 }
 
+/*
+ * Helper function to consume layer DSQs when a system is under starvation.
+ */
+static __always_inline bool consume_starved(s32 cpu)
```
htejun, line 1231: I wonder whether it'd be simpler if we rather skip overconsuming layers rather than adding an extra step to consume starving layers.

Going back to the idea of having per-cpu candidate layer list, for each CPU, there should only be a handful of layers that a CPU should be running. If usage is tracked per-DSQ, which DSQ to consume can be decided in an a lot more granular fashion.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1826,6 +1908,8 @@ void BPF_STRUCT_OPS(layered_stopping, struct task_struct *p, bool runnable)
 	if (cctx->yielding && used < layer_slice_ns)
 		used = layer_slice_ns;
 	p->scx.dsq_vtime += used * 100 / p->scx.weight;
+	layer->vtime_decay += used * 100 / p->scx.weight;
```
htejun, line 1911: I'm not sure this makes sense. Imagine two layers with weight 100 and 200. They have simliar workload compositions but all tasks in the latter layer is niced to have 1/2 per-task weights. This would cancel out the layer weight difference which likely isn't the intention of the configuration.

vtimes and weights usually don't translate well outside of its own timeline. Here, it should be accumulating unadjusted `used` as per-task weights are what direct only how tasks compete within their own layers. As such, `vtime_decay` is probably not the best name. Maybe sth like `usage_avg`?

Also, this is a rather crude way of implementing layer weights. Being a starvation avoidance mechanism, this direction may work out okay but it's likely that there are finer-grained ways to implement this without incurring extra overhead.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1200,6 +1225,56 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+
+	if (layer_weight_sum > decay_vtime_sum)
+		return false;
+
+	/*
+	 * If the weighted decayed vtime of the layer is less than the
+	 * proportional weighted vtime then the layer is potentially starved
+	 * and should be consumed from.
+	 */
+	vtime_weighted_decay = decay_vtime_sum / layer_weight_sum;
```
htejun, line 1265: I don't think this calculation is correct. Imagine three layers with weight 1, 2 and 100 with the third layer completely idle and the first two layers competing against each other. The expected result is 1:2 distribution but with the above it will get 101:102 distribution.

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: I think one issue is that the vtime still gets charged to the layer when a task is put onto the fallback dsq. I guess those vtimes need to be tracked as well.

PR #811, opened 2024-10-16
scx_layered: Add DSQ context

hodgesds:
Add context for DSQs for tracking vtime and avg vtime.

some output from `trace`:
```
          <idle>-0       [000] d.h.1 4698978.290296: bpf_trace_printk: DSQ[0] usage_avg: 0 vtime_now; 12142054 ravg: 64313
          <idle>-0       [000] d.h.1 4698978.290297: bpf_trace_printk: DSQ[1] usage_avg: 0 vtime_now; 0 ravg: 0
          <idle>-0       [000] d.h.1 4698978.290298: bpf_trace_printk: DSQ[2] usage_avg: 1411949 vtime_now; 8950678140 ravg: 3462294305
          <idle>-0       [000] d.h.1 4698978.290299: bpf_trace_printk: DSQ[3] usage_avg: 0 vtime_now; 23401062580 ravg: 15538729024
          <idle>-0       [000] d.h.1 4698978.290300: bpf_trace_printk: DSQ[4] usage_avg: 0 vtime_now; 0 ravg: 0
          <idle>-0       [000] d.h.1 4698978.290300: bpf_trace_printk: DSQ[5] usage_avg: 0 vtime_now; 0 ravg: 0
          <idle>-0       [000] d.h.1 4698978.290301: bpf_trace_printk: DSQ[6] usage_avg: 0 vtime_now; 9345761902 ravg: 2277506371
          <idle>-0       [000] d.h.1 4698978.290302: bpf_trace_printk: DSQ[7] usage_avg: 0 vtime_now; 9345730072 ravg: 2179806149
          <idle>-0       [000] d.h.1 4698978.290303: bpf_trace_printk: DSQ[1024] usage_avg: 896 vtime_now; 899095518 ravg: 171967308
          <idle>-0       [000] d.h.1 4698978.290304: bpf_trace_printk: DSQ[1025] usage_avg: 0 vtime_now; 747609880 ravg: 83412288
          <idle>-0       [000] d.h.1 4698978.291296: bpf_trace_printk: DSQ[0] usage_avg: 0 vtime_now; 12142054 ravg: 64313
          <idle>-0       [000] d.h.1 4698978.291297: bpf_trace_printk: DSQ[1] usage_avg: 0 vtime_now; 0 ravg: 0
          <idle>-0       [000] d.h.1 4698978.291298: bpf_trace_printk: DSQ[2] usage_avg: 1235451 vtime_now; 8950678140 ravg: 3462294305
          <idle>-0       [000] d.h.1 4698978.291299: bpf_trace_printk: DSQ[3] usage_avg: 0 vtime_now; 23401062580 ravg: 15538729024
          <idle>-0       [000] d.h.1 4698978.291300: bpf_trace_printk: DSQ[4] usage_avg: 0 vtime_now; 0 ravg: 0
          <idle>-0       [000] d.h.1 4698978.291301: bpf_trace_printk: DSQ[5] usage_avg: 0 vtime_now; 0 ravg: 0
          <idle>-0       [000] d.h.1 4698978.291302: bpf_trace_printk: DSQ[6] usage_avg: 0 vtime_now; 9345761902 ravg: 2277506371
          <idle>-0       [000] d.h.1 4698978.291303: bpf_trace_printk: DSQ[7] usage_avg: 0 vtime_now; 9345730072 ravg: 2179806149
          <idle>-0       [000] d.h.1 4698978.291304: bpf_trace_printk: DSQ[1024] usage_avg: 784 vtime_now; 899095518 ravg: 171967308
          <idle>-0       [000] d.h.1 4698978.291305: bpf_trace_printk: DSQ[1025] usage_avg: 0 vtime_now; 747609880 ravg: 83412288
```

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -201,6 +201,22 @@ static struct node_ctx *lookup_node_ctx(u32 node)
 
 struct {
 	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, int);
+	__type(value, struct dsq_ctx);
+	__uint(max_entries, LO_FALLBACK_DSQ);
+	__uint(map_flags, 0);
+} dsq_data SEC(".maps");
+
+static struct dsq_ctx *lookup_dsq_ctx(u64 dsq_idx)
+{
+	struct dsq_ctx *dsqc;
+
+	dsqc = bpf_map_lookup_elem(&dsq_data, &dsq_idx);
+	return dsqc;
+}
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
 	__type(key, u32);
 	__type(value, struct cache_ctx);
 	__uint(max_entries, MAX_LLCS);
@@ -301,6 +317,45 @@ static struct cpumask *lookup_layer_cpumask(int idx)
 	}
 }
 
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, int);
+	__type(value, struct lock_wrapper);
+	__uint(max_entries, LO_FALLBACK_DSQ);
+	__uint(map_flags, 0);
+} dsq_ctx_locks SEC(".maps");
+
+static void calc_dsq_usage(u64 dsq_id)
+{
+	struct dsq_ctx *dsqc;
+	struct lock_wrapper *lockw;
+
+	if (!(dsqc = lookup_dsq_ctx(dsq_id)) ||
+	    !(lockw = bpf_map_lookup_elem(&dsq_ctx_locks, &dsq_id)))
+		return;
+
+	bpf_spin_lock(&lockw->lock);
+	dsqc->usage_avg = (dsqc->usage_avg / 8) * 7;
+	bpf_spin_unlock(&lockw->lock);
+}
+
+static void update_dsq_usage(u64 dsq_id, u64 vtime_now, u64 usage_avg_adj)
+{
+	struct dsq_ctx *dsqc;
+	struct lock_wrapper *lockw;
+	u64 now = bpf_ktime_get_ns();
+
+	if (!(dsqc = lookup_dsq_ctx(dsq_id)) ||
+	    !(lockw = bpf_map_lookup_elem(&dsq_ctx_locks, &dsq_id)))
+		return;
+
+	bpf_spin_lock(&lockw->lock);
+	dsqc->vtime_now = vtime_now;
+	dsqc->usage_avg += usage_avg_adj;
+	ravg_accumulate(&dsqc->vtime_rd, dsqc->vtime_now, now, USAGE_HALF_LIFE_SEC);
+	bpf_spin_unlock(&lockw->lock);
+}
+
 /*
  * Returns if any cpus were added to the layer.
  */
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -301,6 +317,45 @@ static struct cpumask *lookup_layer_cpumask(int idx)
...
+{
+	struct dsq_ctx *dsqc;
+	struct lock_wrapper *lockw;
+
+	if (!(dsqc = lookup_dsq_ctx(dsq_id)) ||
+	    !(lockw = bpf_map_lookup_elem(&dsq_ctx_locks, &dsq_id)))
+		return;
+
+	bpf_spin_lock(&lockw->lock);
+	dsqc->usage_avg = (dsqc->usage_avg / 8) * 7;
```
hodgesds, line 338: @htejun I'm not convinced this is better than using `ravg_accumulate`, but wasn't sure if it is ok to call it during stopping or if it should be called periodically.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -199,6 +199,22 @@ static struct node_ctx *lookup_node_ctx(u32 node)
...
+	__type(value, struct dsq_ctx);
+	__uint(max_entries, LO_FALLBACK_DSQ);
+	__uint(map_flags, 0);
+} dsq_data SEC(".maps");
+
+static struct dsq_ctx *lookup_dsq_ctx(u64 dsq_idx)
+{
+	struct dsq_ctx *dsqc;
+
+	dsqc = bpf_map_lookup_elem(&dsq_data, &dsq_idx);
```
htejun, line 214: Does it want to trigger error when lookup fails? Otherwise, it can just be `return bpf_map_lookup_elem(..)`

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: `ravg` can be used here. It doesn't need periodic calls but it does need timestamp on each accumulation and when the value is read. Here, timestamp is already acquired anyway, so shouldn't add much overhead.

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: Yeah, probably should error out.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: But why does it need to average out usage? Can't it just track absoulate usage and control accordingly?

JakeHillion: We were concerned that tracking absolute usage over the lifetime of the scheduler would cause massive priority inflation for layers with moderate weights that go unused for a while, in that we'd end up seriously indebted to these processes in the long term because we didn't serve them when they had no tasks queued. Looking at usage with a big recency bias seemed to solve this case nicely, while still helping processes that would get less than their weight in the short term. Would be interested in more of your thoughts!

htejun: The usual strategy to handle that is capping the maximum deltas. e.g. Imagine putting the layers on a vtime line like we do with tasks, the reason why task vtime lines enforce min_vtime when a task wakes up is for the same reason, as the negative delta represents the budget it has accumulated while it wasn't running and we don't want it to dominate. This way of expressing it in vtime scheme maps closely to maximum accumulated token limit on token based distribution schemes.

Using running avg also does limit the amount of deviation but as an almost secondary effect. It will be conceptually cleaner and easier to modulate if the implemented mechanism directly addresses the issue (here, accumulating too much budget delta).

hodgesds:
going to try something slightly different

PR #812, opened 2024-10-17
scx_layered: add a flag to have layered emit logs as json

likewhatevs:
add a flag to have layered emit logs as ndjson, have stats use this. ndjson is newline delimited json. It simplifies consuming logs in the right manner (i.e. get a line, or a few lines, publish, repeat).

if we decide to go with this, the json-log flag would have all logs emitted via json like it did with stats.

I would also add some formatting cleanup in json outputs (trimming of spaces on strings,  typing of entries, some issue wrt/ nesting of spans).

Here is how this PR changes how output looks without the `--json-log` flag:
new-logs-no-json

Here is how output looks when `--json-log` is set:
new-logs-json

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1131,6 +1131,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3e2e65a1a2e43cfcb47a895c4c8b10d1f4a61097f9f254f183aee60cad9c651d"
 
 [[package]]
+name = "matchers"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8263075bb86c5a1b1427b5ae862e8889656f126e9f77c484496e8b47cf5c5558"
+dependencies = [
+ "regex-automata 0.1.10",
+]
+
+[[package]]
 name = "memchr"
 version = "2.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1275,6 +1284,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "nu-ansi-term"
+version = "0.46.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77a8165726e8236064dbb45459242600304b42a5ea24ee2948e18e023bf7ba84"
+dependencies = [
+ "overload",
+ "winapi",
+]
+
+[[package]]
 name = "num-conv"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


likewhatevs:
My thinking is, if we want to go this route, I'd add something like this to user_exit_into:
```
    // Return none if the scheduler exited 0,
    // Return UserExitInfo if exit non-zero.
    pub fn uei_return(&self) -> Some(UserExitInfo) {
        if self.kind == 0 {
            return None;
        }

        if self.kind <= ScxExitKind::UnregKern as i32 {
            return None;
        }

        return Some(self);
    }
```
and just emit that like how all other logs are emitted.

htejun:
I'm a bit confused on why this PR is adding json outputs for stats. `scx_stats` already have programmable interface and it should be possible to build generic `scx_stats` -> `ndjson` tool in the style of [`scxstats_to_openmetrics.py`](https://github.com/sched-ext/scx/blob/main/rust/scx_stats/scripts/scxstats_to_openmetrics.py) if necessary.

For exit info, it does have some structured data - backtrace addresses and exit codes are available as data values. However, the backtrace then would require symbolization. It's a bit nasty but it might just be easier to parse the text output using a postprocessing script.

likewhatevs:
> I'm a bit confused on why this PR is adding json outputs for stats. scx_stats already have programmable interface and it should be possible to build generic scx_stats -> ndjson tool in the style of [scxstats_to_openmetrics.py](https://github.com/sched-ext/scx/blob/main/rust/scx_stats/scripts/scxstats_to_openmetrics.py) if necessary.

The idea was to skip that step and emit json directly. 

So like, to use `scxstats_to_openmetrics.py` to get real time metrics, one would need to deploy and run multiple packages (the scheduler and a scheduler monitor). 

If this binary were able to emit json, the logs could be sank to a tool like elasticsearch/ELK using a generic json log exporter and could be ingested roughly as-is, shortening the path (and lift) to enable real time monitoring of scheduler configuration and it's performance.

> For exit info, it does have some structured data - backtrace addresses and exit codes are available as data values. However, the backtrace then would require symbolization. It's a bit nasty but it might just be easier to parse the text output using a postprocessing script.

I wasn't thinking of adding symbolization to that. I wanted to store the dump that is printed when schedulers are kicked in json so that it could be processed like how I mentioned stats logged as json could be above.

I think that, ideally, metrics would be exported directly from the scheduler binary and logs would be able emitted to stdout in a structured format (to simplify log ingest), and I think this is a step in that direction.

likewhatevs:
All that being said, if we want to go that direction, the formatting etc. for the non-json bits should be good and IDK if it is now  (kinda mixed wrt/ how clear those logs are to read in that format, but I think it is more popular so I left it in).

htejun:
So, even if we wanted to be able to output json directly from scheduler binary, wouldn't it be better to bulid that into `scx_stats` so that each scheduler doesn't have to format into json separately?

likewhatevs:
> So, even if we wanted to be able to output json directly from scheduler binary, wouldn't it be better to bulid that into scx_stats so that each scheduler doesn't have to format into json separately?

I was going for "make changes where blast radius is smallest" but that IMO is definitely the way to do this.

likewhatevs:
will implement this in scx_stats

PR #813, opened 2024-10-17
scx_bpfland: rework lowlatency mode

arighi:
Rework lowlatency mode introducing the concept of task's "latency weight".

The latency weight is a dynamic priority, evaluated in function of the task's static weight and the average amount of voluntary context switches. When lowlatency mode is enabled tasks will use their latency weight to evaluate their vruntime and deadline, instead of their static weight.

The latency weight also determines the minimum vruntime for each tasks, enabling a natural priority boost to those that have a higher latency weight over those with a lower latency weight. This makes the classification of "interactive" and "regular" tasks unnecessary, and we can use a single global DSQ to dispatch all tasks when latency mode is enabled.

The new `--lowlatency` mode gives a better level of performance predictability, making this option more suitable for multimedia and real-time audio processing workloads.

### DIFF
--- a/scheds/rust/scx_bpfland/src/stats.rs
+++ b/scheds/rust/scx_bpfland/src/stats.rs
@@ -21,8 +21,10 @@ pub struct Metrics {
     pub nr_cpus: u64,
     #[stat(desc = "Number of running interactive tasks")]
     pub nr_interactive: u64,
-    #[stat(desc = "Average amount of tasks waiting to be dispatched")]
-    pub nr_waiting: u64,
+    #[stat(desc = "Average amount of regular tasks waiting to be dispatched")]
+    pub nr_shared_waiting: u64,
+    #[stat(desc = "Average amount of interactive tasks waiting to be dispatched")]
+    pub nr_prio_waiting: u64,
     #[stat(desc = "Average of voluntary context switches")]
     pub nvcsw_avg_thresh: u64,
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_bpfland/src/stats.rs
```diff
@@ -39,12 +41,13 @@ impl Metrics {
     fn format<W: Write>(&self, w: &mut W) -> Result<()> {
         writeln!(
             w,
-            "[{}] tasks -> run: {:>2}/{:<2} int: {:<2} wait: {:<4} | nvcsw: {:<4} | dispatch -> kth: {:<5} dir: {:<5} pri: {:<5} shr: {:<5}",
+            "[{}] tasks -> r: {:>2}/{:<2} i: {:<2} pw: {:<4} w: {:<4} | nvcsw: {:<4} | dispatch -> k: {:<5} d: {:<5} p: {:<5} s: {:<5}",
```
hodgesds, line 44: Maybe have multiple lines to keep it more readable?

arighi: yeah... I was aiming to keep the output on a single line for easier parsing, but I'm not sure if that's relevant anymore, considering that we don't print the output to the logs by default now. I'll probably split that to multiple lines in a separate PR, thanks!

PR #817, opened 2024-10-17
remove apt fast from ci setup

likewhatevs:
remove apt fast from ci setup to reduce non-core dependencies and fix ci

### DIFF
--- a/.github/actions/install-deps-action/action.yml
+++ b/.github/actions/install-deps-action/action.yml
@@ -4,25 +4,20 @@ runs:
   using: 'composite'
   steps:
     ### OTHER REPOS ####
-    # turn off interactive, refresh pkgs, use apt fast
+    # turn off interactive, refresh pkgs
     - run: |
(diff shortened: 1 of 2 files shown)


likewhatevs:
> Thanks! Also for fixing the trailing spaces.

Thanks for letting me know about this issue! Also adding a .editorconfig to do that (and specify tabs, not spaces, for edited lines only) is on my to-do list (I think that's how those got automatically removed).

arighi:
> > Thanks! Also for fixing the trailing spaces.
> 
> Thanks for letting me know about this issue! Also adding a .editorconfig to do that (and specify tabs, not spaces, for edited lines only) is on my to-do list (I think that's how those got automatically removed).

I always notice those because of this in my .vimrc :)
```
let c_space_errors=1
highlight WhitespaceEOL ctermbg=red guibg=red
match WhitespaceEOL /\s\+$/
```

PR #818, opened 2024-10-18
scx_lavd: add missing reset_lock_futex_boost()

multics69:
reset_lock_futex_boost() should be called every context switch of a
task. Otherwise, in the worst case, a task and that CPU could block
the preemption. To avoid such a situation, add missing
reset_lock_futex_boost() calls.


multics69:
Performance regression was reported. I will try to address the problem now.

htejun (approved):

htejun (approved):

PR #822, opened 2024-10-20
scx_lavd: misc performance tuning

multics69:
This PR contains the following two performance tunings:
- give more penalty to long-running tasks
- direct dispatch when there is an idle CPU

With these two commits, now `scx_lavd` show more stable performance on `osu!` with heavy background workloads.


1Naim:
This definitely shows an improvement on my system compared to `main`.

```
```

htejun (approved):

PR #824, opened 2024-10-20
scx_layered: fix exit_task ctx lookup err

likewhatevs:
scx_layered: fix exit_task ctx lookup err

when exit_task is called with the cancelled arg, tasks have not been run on sched_ext, so task_ctx may not exist.

do not attempt hard failing lookup when this occurs, just return.


hodgesds (approved):
I think this makes sense

PR #825, opened 2024-10-20
scx_layered: add affinity dsq

likewhatevs:
Add a dsq for tasks with affinity set

If i got this right, it should put all tasks with affinity set into the affinity dsq which should be executable by all CPUs.

This dsq should be consumed after the high fallback dsq.

I think, because I didn't set any cpumask on this new dsq (presuming that's how this works), tasks can be consumed on any cpu (enabling the affinity that was set to be respected)?

This is to fix a tricky to repro issue where tasks w/ affinity set stall.

I left the following commands running for a few minutes and no stalls, so, I don't think this will cause stalls when affinity is set (and it might fix the stalls that happen on things w/ affinity set, sometimes):
```
patso in 🌐 gentoo in scheds/rust/scx_layered on  layered-affinity-stall-fix [$!] is 📦 v1.0.5 via 🦀 v1.83.0-nightly took 2s
❯ sudo ../../../target/debug/scx_layered --run-example --stats 1
23:41:58 [INFO] CPUs: online/possible=32/32 nr_cores=16
23:41:58 [INFO] Topology awareness not specified, selecting enabled based on hardware
23:41:58 [INFO] Layered Scheduler Attached. Run `scx_layered --monitor` for metrics.
###### Sun, 20 Oct 2024 19:41:59 -0400 ######
tot=     10 local=10.00 open_idle= 0.00 affn_viol= 0.00 proc=0ms nodes=1
```

```
patso in 🌐 gentoo in ~
❯ stress-ng -f 10
stress-ng: info:  [1309823] defaulting to a 1 day run per stressor
stress-ng: info:  [1309823] dispatching hogs: 10 fork
```

```
patso in 🌐 gentoo in ~/scratch2 via C v13.3.1-gcc
❯ taskset -p -c 29 1309824
pid 1309824's current affinity list: 0-31
pid 1309824's new affinity list: 29
```

WezTerm 2024-10-20 20 02 49


likewhatevs:
Part of my thinking behind this is like, if something is setting affinity, respect it (i.e. don't migrate) and execute earlier (right after high fallback) because it's probably important and complex.

hodgesds:
I think you can test this with:
```
stress-ng --affinity `nproc`
```

likewhatevs:
doing this a different way

PR #826, opened 2024-10-21
layered: bpf: add layer kind to layer

JakeHillion:
Currently we have an approximation of LayerKind in the BPF code with `open` on
the layer, but it is difficult/impossible to tell the difference between an
Open and a Grouped layer. Add a `kind` field to the BPF `layer` and plumb
through an enum from the Rust side.


hodgesds (approved):
LGTM

PR #829, opened 2024-10-22
Add script for generating per-arch vmlinux.h files

minosfuture:
This script iterates over a list archs and generates vmlinux.h for each. Generated files are put under the corresponding arch directory.

### DIFF
--- a/scripts/gen_vmlinux_h.sh
+++ b/scripts/gen_vmlinux_h.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+
+# Prerequisite
+#
+# * .config is ready in the linux repo
+# * binary bpftool is installed
+# * can `sudo apt install` cross-compile toolchains
+#
+# Example
+#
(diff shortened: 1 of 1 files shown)


scripts/gen_vmlinux_h.sh
```diff
@@ -0,0 +1,81 @@
...
+}
+
+if ! command -v bpftool &> /dev/null
+then
+    echo "bpftool could not be found. Please install it first."
+    exit 1
+fi
+
+echo "Install cross-compile toolchains:"
+sudo apt install \
```
frelon, line 63: Maybe this could be moved to a comment or readme? To make this work for people who are not using `apt` :wink:

hodgesds: Maybe this could be encapsulated with a set of distro specific wrapper functions.

minosfuture: updated!

htejun (approved):

PR #833, opened 2024-10-22
ci: enable verbose mode when testing schedulers

arighi:
Always run schedulers in verbose mode to be able to catch the details of potential BPF failures.


htejun (approved):

arighi:
I still don't understand why we're getting all these failures, but at least we see the errors now. Will investigate more tomorrow morning if it's not solved yet.

PR #834, opened 2024-10-23
fix ci errors due to __str update in kfunc signature

likewhatevs:
fix ci errors due to __str update in kfunc signature

per the following doc on verifier annotations:

https://www.kernel.org/doc/html/next/bpf/kfuncs.html#str-annotation

the argument to fmt__str needs to be const now.


likewhatevs:
somewhat confident this is the right fix and ci not broken is better than ci broken so merging if signals pass.

arighi (approved):

arighi:
Hm... it looks like all schedulers are a bit broken now.

arighi:
With #836 we should be able to support both old and new kernels.

likewhatevs:
> they're broken if you have an old kernel without the __str kfunc change. I'd be nice to support both ways (somehow), but I'm not sure how

Thanks for catching this. Once 6.12 is out, I plan on adding CI jobs for supported kernel versions (with that being the first).

PR #836, opened 2024-10-23
scx/common.bpf.h: Compatible kfunc string arguments

arighi:
Kernel commit 3e99aee7ce48 ("sched-ext: Use correct annotation for strings in kfuncs") introduced proper annotations for kfunc string arguments, that are now properly considered as const strings.

Commit 29918c0 ("fix ci errors due to __str update in kfunc signature") applied the same change to common.bpf.h, but this broke the compatibility with old kernels that don't have the annotation change.

Support both formats by relying on the compiler to determine the type of the format string.

Fixes: 29918c0 ("fix ci errors due to __str update in kfunc signature")

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -102,7 +102,7 @@ ___scx_bpf_bstr_format_checker(const char *fmt, ...) {}
  * refer to the initialized list of inputs to the bstr kfunc.
  */
 #define scx_bpf_bstr_preamble(fmt, args...)                                    \
-  static const char ___fmt[] = fmt;                                                  \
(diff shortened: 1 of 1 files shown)


arighi:
Nope, it doesn't seem to fix it.... bummer.

multics69 (approved):
Thank you for catching this!

arighi:
Yeah, I haven't been able to figure out a proper solution. Let me close this PR, since it doesn't really fix anything. We can open a new issue for this.

arighi:
Yeah... honestly I'm not a big fan of introducing autoconf mechanics, but at the same time it'd be nice to support old and new kernels... I'll think more about this, I don't have a clean solution at the moment.

PR #838, opened 2024-10-23
scx_common: Make scx_bpf_bstr_preamble declarations global

hodgesds:
Make scx_bpf_bstr_preamble declarations use a non static declaration to make verification easier on older kernels.


hodgesds:
nope!

PR #841, opened 2024-10-23
scx_layered: Add monitor

hodgesds:
Add a monitor timer for scx_layered. For now the monitor is a noop, but can be made more generic to add extra warnings stats when the scheduler reaches a bad state.


htejun (approved):

htejun:
BTW, scx core has stall detection which is also called watchdog. Might be a good idea to use a different name to avoid confusion.

likewhatevs:
I hit the rebase button @hodgesds (perhaps I shouldn't have now that I see it force pushes which kinda makes sense) but yeah just btw

PR #848, opened 2024-10-24
scx_loader: update zbus to next major version

vnepogodin:


vnepogodin:
Cannot use it for now due to bumped MSRV requirements of `zbus` crate to 1.81

PR #850, opened 2024-10-24
fix lsp to work after multiarch support

likewhatevs:
Fix lsp to work after multiarch support

I think this is a good way to enable lsp etc. to keep working with multiarch support.

If someone is working on arm and wants lsp etc. to work, they'll can overwrite the symlink to the right one for their arch (and that override is gitignored/the default is what will work best for most folks). 

CleanShot 2024-10-24 at 10 04 29@2x

### DIFF
--- a/scheds/include/.gitignore
+++ b/scheds/include/.gitignore
@@ -1,0 +1,1 @@
+vmlinux.h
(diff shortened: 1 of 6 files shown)


likewhatevs:
updated to make tooling work the best it can on the timer library (imo that should be in main.bpf.c, then tooling could work perfect, but as is things are pretty good):

CleanShot 2024-10-24 at 10 45 38@2x

PR #851, opened 2024-10-24
scx_layered: Fix declarations in timer

hodgesds:
Add a forward declaration to be more kind to tools and fix int->u64 casting.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/timer.bpf.h
+++ b/scheds/rust/scx_layered/src/bpf/timer.bpf.h
@@ -12,7 +12,7 @@ enum timer_consts {
 
 struct layered_timer {
 	// if set to 0 the timer will only be scheduled once
-	int interval_ns;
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/bpf/timer.bpf.h
```diff
@@ -22,4 +22,7 @@ enum layer_timer_callbacks {
 	NOOP_TIMER,
 	MAX_TIMERS,
 };
+
+static bool run_timer_cb(int key);
```
JakeHillion, line 26: I expect it needs one for `layered_timers` too to make tooling happy.
```suggestion
static bool run_timer_cb(int key);
struct layered_timer *layered_timers;
```

likewhatevs (approved):
trying to figure out how to rebase a pr atop a pr to get screenshot of lsp stuff working but lgtm

PR #852, opened 2024-10-25
layered/timers: support verifying on older kernels and fix logic

JakeHillion:
Some of the new timer code doesn't verify on older kernels like 6.9. Modify the
code a little to get it verifying again.

Also applies some small fixes to the logic. Error handling was a little off
before and we were using the wrong key in lookups.

Test plan:
- CI


hodgesds (approved):
LGTM

PR #855, opened 2024-10-28
scx_layered: Add cost accounting

hodgesds:
Add cost accounting for layers to make weights work on the BPF side. This is done at both the CPU level as well as globally. When a CPU runs out of budget it acquires budget from the global context. If a layer runs out of global budgets then all budgets are reset. Weight handling is done by iterating over layers by their available budget. Layers budgets are proportional to their weights.

There's still some issues to maybe make the budget updates to not have races, but it seems to work ok for now.

### Tests
Simple two layer config:
```
[{
  "name":"hodgesd",
  "comment":"hodgesd user",
  "matches":[
     [{"UIDEquals":224791}]
  ],
  "kind": {
    "Open":{
      "slice_us": 1000,
      "preempt": false,
      "preempt_first": false,
      "exclusive": false,
      "weight": 100
     }
  }
},{
  "name":"rest",
  "comment":"da rest",
  "matches":[[]],
  "kind": {
    "Open":{
      "slice_us": 1000,
      "preempt": false,
      "preempt_first": false,
      "exclusive": false,
      "weight": 100
     }
  }
}]
```
Roughly even output on a 80 core machine when using half cores:
image
Compared to eevdf the performance is similar:
image
### Weights
Changing the `rest` layer's weight to `1000` and rerunning with 40 cores each yields similar results:
image
When running with 80 cores (`nproc`) the weighted layer gets more throughput:
image
### Three layer test
Testing with three layers with the following config:
```
[
  {
    "name": "hodgesd",
    "comment": "hodgesd user",
    "matches": [
      [
        {
          "UIDEquals": 224791
        }
      ]
    ],
    "kind": {
      "Open": {
        "slice_us": 1000,
        "preempt": false,
        "preempt_first": false,
        "exclusive": false,
        "weight": 100
      }
    }
  },
  {
    "name": "stress-ng",
    "comment": "stress-ng slice",
    "matches": [
      [
        {
          "CommPrefix": "stress-ng"
        }
      ],
      [
        {
          "PcommPrefix": "stress-ng"
        }
      ]
    ],
    "kind": {
      "Open": {
        "slice_us": 1000,
        "preempt": false,
        "preempt_first": false,
        "exclusive": false,
        "weight": 100
      }
    }
  },
  {
    "name": "normal",
    "comment": "the rest",
    "matches": [
      []
    ],
    "kind": {
      "Open": {
        "slice_us": 1000,
        "preempt": false,
        "preempt_first": false,
        "exclusive": false,
        "weight": 200
      }
    }
  }
]
```
Fully loading the system with `stress-ng` using all cores across multiple layers still works:
image
However, when saturating the machine (`nproc * 2`) the results are skewed for one layer:
image
This can be a future optimization.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/cost.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/cost.bpf.c
@@ -1,0 +1,324 @@
+/* Copyright (c) Meta Platforms, Inc. and affiliates. */
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+
+/*
+ * Cost accounting struct that is used in both the per CPU and global context.
+ * Budgets are allowed to recurse to parent structs.
+ */
+struct cost {
+	s64		budget[MAX_LAYERS];
+	s64		capacity[MAX_LAYERS];
+	u32		pref_layer;
+	u32		idx;
+	bool		overflow;
+	bool		has_parent;
+};
+
+
+/*
+ * Map used for global cost accounting. Can be extended to support NUMA nodes.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, u32);
+	__type(value, struct cost);
+	__uint(max_entries, MAX_NUMA_NODES + 1);
+	__uint(map_flags, 0);
+} cost_data SEC(".maps");
+
+/*
+ * CPU map for cost accounting. When budget is expired it requests budget from
+ * global entries.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__type(key, u32);
+	__type(value, struct cost);
+	__uint(max_entries, 1);
+} cpu_cost_data SEC(".maps");
+
+static __always_inline struct cost *lookup_cost(u32 cost_id)
+{
+	struct cost *costc;
+
+	costc = bpf_map_lookup_elem(&cost_data, &cost_id);
+	if (!costc) {
+		scx_bpf_error("cost not found");
+		return NULL;
+	}
+
+	return costc;
+}
+
+static __always_inline struct cost *lookup_cpu_cost(s32 cpu)
+{
+	struct cost *costc;
+	u32 zero = 0;
+
+	if (cpu < 0)
+		costc = bpf_map_lookup_elem(&cpu_cost_data, &zero);
+	else
+		costc = bpf_map_lookup_percpu_elem(&cpu_cost_data,
+						   &zero, cpu);
+	if (!costc) {
+		scx_bpf_error("cost not found");
+		return NULL;
+	}
+
+	return costc;
+}
+
+/*
+ * Initializes a cost.
+ */
+static struct cost *initialize_cost(u32 cost_idx, u32 parent_idx,
+				    bool is_cpu, bool has_parent,
+				    bool overflow)
+{
+	struct cost *costc;
+
+	if (is_cpu) {
+		if (!(costc = lookup_cpu_cost(cost_idx)))
+			return NULL;
+	} else {
+		if (!(costc = lookup_cost(cost_idx)))
+			return NULL;
+	}
+
+	if (has_parent)
+		costc->idx = parent_idx;
+	else
+		costc->idx = cost_idx;
+
+	costc->has_parent = has_parent;
+	costc->overflow = overflow;
+	costc->pref_layer = bpf_get_smp_processor_id() % nr_layers;
+
+	return costc;
+}
+
+/*
+ * Initializes the cost of a layer.
+ */
+static void initialize_cost_layer(struct cost *costc, u32 layer_id, s64 capacity)
+{
+	costc->capacity[layer_id] = capacity;
+	costc->budget[layer_id] = capacity;
+}
+
+/*
+ * Returns the preferred layer based on the layer with the maximum budget.
+ */
+static u32 preferred_cost(struct cost *costc)
+{
+	u32 layer_id, id, max_layer = 0;
+	s64 max_budget = 0;
+	u32 rotation = bpf_get_smp_processor_id() % nr_layers;
+
+	bpf_for(id, 0, nr_layers) {
+		// If there is two equally weighted layers that have the same
+		// budget we rely on rotating the layers based on the cpu. This
+		// may not work well on low core machines.
+		layer_id = rotate_layer_id(id, rotation);
+		if (layer_id > nr_layers) {
+			scx_bpf_error("invalid layer");
+			return 0;
+		}
+		if (costc->budget[layer_id] > max_budget) {
+			max_budget = costc->budget[layer_id];
+			max_layer = layer_id;
+		}
+	}
+
+	return max_layer;
+}
+
+/*
+ * Refreshes the budget of a cost.
+ */
+int refresh_budget(int cost_id)
+{
+	struct cost *costc;
+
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,327 @@
...
+	u32 layer_id, id;
+	u32 rotation = bpf_get_smp_processor_id() % nr_layers;
+	bpf_for(id, 0, nr_layers) {
+		layer_id = rotate_layer_id(id, rotation);
+		if (layer_id > nr_layers) {
+			scx_bpf_error("invalid layer");
+			return 0;
+		}
+		s64 capacity = costc->capacity[layer_id];
+		__sync_lock_test_and_set(MEMBER_VPTR(*costc, .budget[layer_id]),
```
hodgesds, line 157: Probably should check these atomics

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+	__type(value, struct cost);
+	__uint(max_entries, 1);
+} cpu_cost_data SEC(".maps");
+
+
+static __always_inline struct cost *lookup_cost(u32 cost_id)
+{
+	struct cost *costc;
+
+	costc = bpf_map_lookup_elem(&cost_data, &cost_id);
```
htejun, line 48: Maybe trigger ops error on lookup failure?

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+static __always_inline struct cost *lookup_cpu_cost(s32 cpu)
+{
+	struct cost *costc;
+	u32 zero = 0;
+
+	if (cpu < 0)
+		costc = bpf_map_lookup_elem(&cpu_cost_data, &zero);
+	else
+		costc = bpf_map_lookup_percpu_elem(&cpu_cost_data,
+						   &zero, cpu);
```
htejun, line 62: Ditto.

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+
+	return costc;
+}
+
+/*
+ * Initializes the cost of a layer.
+ */
+static void initialize_cost_layer(struct cost *costc,
+				  u32 layer_id,
+				  bool overflow,
```
htejun, line 101: `overflow` is not used. Is that intended?

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+
+
+/*
+ * Map used for global cost accounting. Can be extended to support LLCs.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, u32);
+	__type(value, struct cost);
+	__uint(max_entries, MAX_NUMA_NODES * MAX_LLCS * MAX_LAYERS);
```
htejun, line 28: Each `cost` tracks all layers. Is `* MAX_LAYERS` intended?

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+		// If there is two equally weighted layers that have the same
+		// budget we rely on rotating the layers based on the cpu. This
+		// may not work well on low core machines.
+		layer_id = rotate_layer_id(id, rotation);
+		if (layer_id > nr_layers) {
+			scx_bpf_error("invalid layer");
+			return 0;
+		}
+		if (costc->budget[layer_id] > max_budget) {
+			max_budget = *MEMBER_VPTR(*costc, .budget[layer_id]);
```
htejun, line 127: If `MEMBER_VPTR` can be deref'd without checking for NULL, `MEMBER_VTPR` can probably be removed.

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+/*
+ * Returns the preferred layer based on the layer with the maximum budget.
+ */
+static u32 preferred_cost(struct cost *costc)
+{
+	u32 layer_id, id, max_layer = 0;
+	s64 max_budget = 0;
+	u32 rotation = bpf_get_smp_processor_id() % nr_layers;
+
+	bpf_for(id, 0, nr_layers) {
```
htejun, line 117: This is fine for now but, down the line, I think we'll likely need to restrict the CPUs that each CPU looks at in the dispatch path for scalability. This will likely need a similar treatment.

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+
+	if (!(costc = lookup_cost(cost_id))) {
+		scx_bpf_error("failed to lookup cost %d", cost_id);
+		return 0;
+	}
+
+	u32 layer_id, id;
+	u32 rotation = bpf_get_smp_processor_id() % nr_layers;
+	bpf_for(id, 0, nr_layers) {
+		layer_id = rotate_layer_id(id, rotation);
```
htejun, line 150: Does rotation matter here?

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+}
+
+/*
+ * Acquires a budget from a parent cost account.
+ */
+s64 acquire_budget(struct cost *costc, u32 layer_id, s64 amount)
+{
+	s64 budget = 0;
+
+	if (layer_id > MAX_LAYERS || layer_id < 0) {
```
htejun, line 180: Here and in other similar places, shouldn't this be `layer_id >= MAX_LAYERS`?

scheds/rust/scx_layered/src/bpf/cost.bpf.c
```diff
@@ -0,0 +1,326 @@
...
+ * refreshing all budgets.
+ */
+static int record_cpu_cost(struct cost *costc, u32 layer_id, s64 amount)
+{
+	if (layer_id > MAX_LAYERS || !costc) {
+		scx_bpf_error("invalid layer %d", layer_id);
+		return 0;
+	}
+
+	__sync_fetch_and_sub(MEMBER_VPTR(*costc, .budget[layer_id]), amount);
```
htejun, line 214: Ditto with `MEMBER_VPTR()` usage.

htejun (approved):
Left some nits but generally looks great to me. Thanks!

scheds/rust/scx_layered/src/bpf/cost.bpf.c
hodgesds: I think it's leftover from some stuff I ripped out, good catch.

scheds/rust/scx_layered/src/bpf/cost.bpf.c
hodgesds: I wasn't sure if refreshing the global budget in layer order would lead to the first layer always getting capacity first and potentially being biased to being the preferred layer. Could do some more testing though as I think there is some other issue under heavy load.

likewhatevs:
no stalls w/ this cmd on a 9950x

```
stress-ng -c 128 -M --affinity 128
```

good stuff

samuelnair:
Could we update the docs to explain how to use this.

PR #862, opened 2024-10-30
scx_layered: Refactor dispatch

hodgesds:
Refactor dispatch to use a separate set of global helpers for topo aware dispatch. This change only refactors dispatch to make it more maintainable, without any functional changes.


JakeHillion (approved):
Nice. It looks like we're doing some checks that don't rely on the LLC number in the per-LLC loops, but it's unrelated to this PR and I expect it to change in a follow up anyway.

PR #874, opened 2024-11-01
scx_layered: Add additional drain to fallback DSQs

hodgesds:
Fallback DSQs are not accounted with costs. If a layer is saturating the machine it is possible to not consume from the fallback DSQ and stall the task. This introduces and additional consumption from the fallback DSQ when a layer runs out of budget. In addition, tasks that use partial CPU affinities should be placed into the fallback DSQ. This change was tested with stress-ng --cacheline `nproc` for several minutes without causing stalls (which would stall on main).

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1067,7 +1067,7 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
 	 * usually important for system performance and responsiveness.
 	 */
 	if (!layer->preempt &&
-	    (p->flags & PF_KTHREAD) && p->nr_cpus_allowed == 1) {
+	    (p->flags & PF_KTHREAD) && p->nr_cpus_allowed < nr_possible_cpus) {
 		struct cpumask *layer_cpumask;
 
 		if (!layer->open &&
@@ -1442,6 +1442,18 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 	    sib_cctx->current_exclusive) {
 		gstat_inc(GSTAT_EXCL_IDLE, cctx);
 		return;
+	}
+
+	/*
+	 * Fallback DSQs don't have cost accounting. When the budget runs out
+	 * for a layer we do an extra consume of the fallback DSQ to ensure
+	 * that it doesn't stall out when the system is being saturated.
+	 */
+	if (costc->drain_fallback) {
+		costc->drain_fallback = false;
+		dsq_id = cpu_hi_fallback_dsq_id(cpu);
+		if (scx_bpf_consume(dsq_id))
+			return;
 	}
 
 	u32 my_llc_id = cpu_to_llc_id(cpu);
--- a/scheds/rust/scx_layered/src/bpf/cost.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/cost.bpf.c
@@ -15,6 +15,7 @@ struct cost {
 	u32		idx;
 	bool		overflow;
 	bool		has_parent;
+	bool		drain_fallback;
 };
 
 
@@ -217,6 +218,7 @@ static int record_cpu_cost(struct cost *costc, u32 layer_id, s64 amount)
 	__sync_fetch_and_sub(&costc->budget[layer_id], amount);
 
 	if (costc->budget[layer_id] <= 0) {
+		costc->drain_fallback = true;
 		if (costc->has_parent) {
 			s64 budget = acquire_budget(costc, layer_id,
 						    costc->capacity[layer_id] + amount);


hodgesds:
I was able to run `stress-ng` for ~20 min without stalls:
```
$ stress-ng --cacheline `nproc`
stress-ng: info:  [3201232] defaulting to a 1 day, 0 secs run per stressor
stress-ng: info:  [3201232] dispatching hogs: 176 cacheline
^Cstress-ng: info:  [3201232] skipped: 0
stress-ng: info:  [3201232] passed: 176: cacheline (176)
stress-ng: info:  [3201232] failed: 0
stress-ng: info:  [3201232] metrics untrustworthy: 0
stress-ng: info:  [3201232] successful run completed in 19 mins, 13.15 secs
```

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1444,6 +1444,20 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
...
+	 * for a layer we do an extra consume of the fallback DSQ to ensure
+	 * that it doesn't stall out when the system is being saturated.
+	 */
+	if (costc->drain_fallback) {
+		dsq_id = cpu_hi_fallback_dsq_id(cpu);
+		if (scx_bpf_consume(dsq_id)) {
+			costc->drain_fallback = false;
+			return;
+		}
+		costc->drain_fallback = false;
```
JakeHillion, line 1458:
```suggestion
		costc->drain_fallback = false;
		dsq_id = cpu_hi_fallback_dsq_id(cpu);
		if (scx_bpf_consume(dsq_id))
			return;
```

JakeHillion (approved):
Makes sense! I'm still wondering if we can rejig the flow to always hit these DSQs when budget is constrained, and I think that's necessary for fairness (see #870), but this definitely helps with the stalling.

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: Yeah, that's cleaner!

hodgesds:
I still see some occasional stalls when tasks are setting affinity, which probably needs a fix similar to [86db45f](https://github.com/sched-ext/scx/commit/86db45f855a78ddfe3101ad40be0c38bb2867995).
```
  R drgn#native-mai[4006993] -37645ms
      sticky/holding_cpu=-1/-1 dsq_id=0x15 dsq_vtime=2735410207548
      cpus=0000,00000000,00000000,00000000,00000000,00000001
    __set_cpus_allowed_ptr+0x3e/0x80
    __sched_setaffinity+0x67/0x120
    sched_setaffinity+0x104/0x1e0
    __se_sys_sched_setaffinity+0x6d/0xa0
    do_syscall_64+0x67/0x140
    entry_SYSCALL_64_after_hwframe+0x4b/0x53
    ```

likewhatevs:
> I still see some occasional stalls when tasks are setting affinity, which probably needs a fix similar to [86db45f](https://github.com/sched-ext/scx/commit/86db45f855a78ddfe3101ad40be0c38bb2867995).
> 
> ```
>   R drgn#native-mai[4006993] -37645ms
>       scx_state/flags=3/0x9 dsq_flags=0x1 ops_state/qseq=0/0
>       sticky/holding_cpu=-1/-1 dsq_id=0x15 dsq_vtime=2735410207548
>       cpus=0000,00000000,00000000,00000000,00000000,00000001
>     __set_cpus_allowed_ptr+0x3e/0x80
>     __sched_setaffinity+0x67/0x120
>     sched_setaffinity+0x104/0x1e0
>     __se_sys_sched_setaffinity+0x6d/0xa0
>     do_syscall_64+0x67/0x140
>     entry_SYSCALL_64_after_hwframe+0x4b/0x53
>     ```
> ```

will add that if timer + dsq walking alone isn't sufficient to fix those, thx for heads up

likewhatevs (approved):

hodgesds:
> 
> will add that if timer + dsq walking alone isn't sufficient to fix those, thx for heads up

fwiw I think adding a kprobe similar to what @arighi did should be sufficient. May have to use one of the `scx_bpf_dispatch_from_dsq` helpers.

htejun:
Would it make more sense to add cost accounting for HI and LO fallbacks and assign them weights rather than consuming them as a implementation side-effect?

htejun:
Also, once we know we can reliably control how much HI/LO fallbacks run, we can start using LO again. We currently end up prioritizing user tasks with custom affinities in confined layers by putting them in HI because LO gets easily starved but we don't really want to prioritize them.

hodgesds:
> Would it make more sense to add cost accounting for HI and LO fallbacks and assign them weights rather than consuming them as a implementation side-effect?

Yeah, I think that makes sense... this was more a short term hack to get things working until that can be done properly.

PR #876, opened 2024-11-01
lavd: adjust padding definition on private(LAVD)

etsal:
Attempting to load lavd while running a recent kernel (commit 0fc810ae3ae110f9e2fcccce80fc8c8d62f97907) fails with verifier errors on the cpdom_cpumask array. The array is already padded to appease the verifier, and turning the padding into its own section appeases the verifier. Spin out the existing array padding into a separate section.


hodgesds:
Looks like CI is still failing. I got most of the verifier issues worked out with [this](https://gist.github.com/hodgesds/f7ee5fab6c1412a4cf7960440a1f1fff). I think adding a few bounds checks on the array access of `cpdom_cpumask` instead of using `MEMBER_VPTR` makes it easier to verify.

multics69:
I am still not confident about removing that workaround. I will further take a look after/with fixing the CI issue.

PR #878, opened 2024-11-01
check if CI regression is due to missing *mutex_lock* symbols in compiled kernel

etsal:
Continuation of #877


multics69:
Close the PR as https://github.com/sched-ext/scx/pull/885 is posted.

PR #880, opened 2024-11-02
ci: remove CONFIG_PROVE_LOCKING to avoid missing mutex_lock()

multics69:
For some reason, when CONFIG_PROVE_LOCKING is enabled, many mutex and rt_mutex symboles are not traceable. So disable it for now.


multics69:
Nope. It does not work on CI. Why?

arighi:
I'd be nice to keep CONFIG_PROVE_LOCKING enabled in CI, because it can help to catch potential kernel bugs in advance. Ideally we would like to consider all these kprobes optional, like failing to attach a kprobe shouldn't be considered a critical failure (maybe a warning), but I'm not sure if we can do this in a clean way (probably in libbpf-rs / libbpf) without also excluding other load errors.

arighi:
The downside of relying on many kprobes is that it makes the scheduler less portable across kernels, because the code is more dependent on the in-kernel ABI...

Maybe a temporary workaround for this particular case is to attach to `__mutex_lock`, that is defined both with CONFIG_DEBUG_LOCK_ALLOC on and off. The downside is that you won't intercept the fast path mutex locking when lock debugging is disabled, but if you can tolerate locking imbalance maybe it's acceptable...?

likewhatevs:
> Nope. It does not work on CI. Why?

@multics69 Kernels are cached on CI, the cache key is kernel commit hash, so this change wouldn't have been tested.

LMK, if we end up needing to have a separate kernel config for lavd and I can update CI/caching, etc., to make that work if needed.

PR #884, opened 2024-11-03
scx_rusty: Add core type to domain

hodgesds:
Add core type to scx_rusty domain to identify if a domain contains big or little cores. This can be extended to make scx_rusty more compatible with hybrid architectures.

### DIFF
--- a/scheds/rust/scx_rusty/src/domain.rs
+++ b/scheds/rust/scx_rusty/src/domain.rs
@@ -4,7 +4,9 @@
 // GNU General Public License version 2.
 use std::collections::BTreeMap;
 
+use anyhow::anyhow;
 use anyhow::Result;
+use scx_utils::CoreType;
 use scx_utils::Cpumask;
 use scx_utils::Topology;
 
@@ -12,6 +14,7 @@ use scx_utils::Topology;
 pub struct Domain {
     id: usize,
     mask: Cpumask,
+    core_type: CoreType,
 }
 
 impl Domain {
@@ -60,7 +63,15 @@ impl DomainGroup {
             for mask_str in cpumasks.iter() {
                 let mask = Cpumask::from_str(&mask_str)?;
                 span |= mask.clone();
-                doms.insert(dom_id, Domain { id: dom_id, mask });
+                let core_type = CoreType::Big { turbo: false };
+                doms.insert(
+                    dom_id,
+                    Domain {
+                        id: dom_id,
+                        mask,
+                        core_type,
+                    },
+                );
                 dom_numa_map.insert(dom_id, 0);
                 dom_id += 1;
             }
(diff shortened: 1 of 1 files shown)


likewhatevs (approved):
like this direction

scheds/rust/scx_rusty/src/domain.rs
```diff
@@ -71,7 +82,20 @@ impl DomainGroup {
                 for (_, llc) in node.llcs().iter() {
                     let mask = llc.span().clone();
                     span |= mask.clone();
-                    doms.insert(dom_id, Domain { id: dom_id, mask });
+                    let core_type = &llc
+                        .cores()
+                        .values()
+                        .next()
+                        .ok_or(anyhow!("no core"))?
+                        .core_type;
```
hodgesds, line 90: This assumes a LLC will have the same core type. I think maybe that is a valid assumption, but maybe on some of the newer Intel hardware it is different.

JakeHillion: It looks like this isn't true on Alder Lake[1]. The L3 (LLC) is stated as being shared between P- and E- cores, which holds with my understanding. The more complex Low Power E- cores aren't necessary to violate this.

[1]: https://www.notebookcheck.net/Intel-Alder-Lake-architecture-overview-Heterogeneous-ISA-dynamic-Thread-Director-shared-30-MB-L3-cache-and-more.575759.0.html

hodgesds: Ahh... well luckily I have a Alder Lake CPU to confirm:
```
Machine (47GB total)
  Package L#0
    NUMANode L#0 (P#0 47GB)
    L3 L#0 (12MB)
      L2 L#0 (1280KB) + L1d L#0 (48KB) + L1i L#0 (32KB) + Core L#0
        PU L#0 (P#0)
        PU L#1 (P#1)
      L2 L#1 (1280KB) + L1d L#1 (48KB) + L1i L#1 (32KB) + Core L#1
        PU L#2 (P#2)
        PU L#3 (P#3)
      L2 L#2 (2048KB)
        L1d L#2 (32KB) + L1i L#2 (64KB) + Core L#2 + PU L#4 (P#4)
        L1d L#3 (32KB) + L1i L#3 (64KB) + Core L#3 + PU L#5 (P#5)
        L1d L#4 (32KB) + L1i L#4 (64KB) + Core L#4 + PU L#6 (P#6)
        L1d L#5 (32KB) + L1i L#5 (64KB) + Core L#5 + PU L#7 (P#7)
      L2 L#3 (2048KB)
        L1d L#6 (32KB) + L1i L#6 (64KB) + Core L#6 + PU L#8 (P#8)
        L1d L#7 (32KB) + L1i L#7 (64KB) + Core L#7 + PU L#9 (P#9)
        L1d L#8 (32KB) + L1i L#8 (64KB) + Core L#8 + PU L#10 (P#10)
        L1d L#9 (32KB) + L1i L#9 (64KB) + Core L#9 + PU L#11 (P#11)
```

hodgesds: I'm not sure what to do then... I guess this won't work.

JakeHillion: We might need to split the scheduling domains to `LLCs * TypesOfCore` but then we might hit the current issues with layered around when to work steal. I'll leave that to you to figure out!

htejun: Note that this needs to be handled at both layers. Rusty currently assumes that each CPU in every domain has the same computing power, which isn't true for some hybrid setups, so at LB layer, it would need to consider the difference in compute and power consumption differences.

For cases where hybrid cores share the same L3, the timeline part of the scheduler needs to consider the compute and power consumption differences e.g. as scx_lavd does.

ie. We need hybrid awareness at both layers and this is also why we should work on composability because there's going to be lots of duplicate work everywhere otherwise.

PR #885, opened 2024-11-04
scx_lavd: fix CI error for missing kernel symbols

multics69:
Revised the lock tracking code, relying on stable symbols with various kernel configurations. There are two changes:

- Entirely drop tracing rt_mutex, which can be on and off with kconfig

- Replace mutex_lock() families to __mutex_lock(), which is stable across kernel configs. The downside of such a change is it is now possible to trace the lock fast path, so lock tracing is a bit less accurate. But let's live with it for now until a better solution is found.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/lock.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/lock.bpf.c
@@ -12,7 +12,6 @@
 #define LAVD_TRACE_SEM
 #define LAVD_TRACE_MUTEX
 #define LAVD_TRACE_WW_MUTEX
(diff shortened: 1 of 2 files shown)


arighi (approved):
Looks good to me, lavd is much more portable now, thanks!

Thinking more about this, it might be actually a good thing to not track the fastpath, if there's only one task that acquired the lock without any other concurrent task there's no contention, so no risk of priority inversion, hence it's probably fine if the task follows the "normal scheduling rules". If we want to be more precise, an option could be to inspect `lock->owner` in the slowpath and increment the lock_boost of the owner task from there.

PR #887, opened 2024-11-04
scx_lavd: entirely drop kernel lock tracing

multics69:
The combination of kernel versions and kerenl configs generates different kernel symbols. For example, in an old kernel version,  __mutex_lock() is not generated. Also, there is no workaround  from the fentry/fexit/kprobe side currently. Let's entirely drop the kernel locking for now and revisit it later.


likewhatevs:
Yeah this sounded like the coolest thing I've heard in a while.

htejun:
Note that which symbols are actually generated also depends on the compiler and sometimes just build instances (unrelated changes in code / config changes leading to different inlining decisions). Also, the existence of a symbol doesn't necessarily indicate that the extern definition is always used as at least clang does partial inlining too - some callers get inlined while others use the global definition.

There are TP and functions which are marked to be always global. If the benefits are clear, we can push to introduce those to locking paths if there aren't enough already.

multics69:
One idea would be investigatingif fexit/entry can be improved. Currently, I found that there is a symbol that cannot be bound to a kernel trace point, the loading of the entire BPF program fails.

multics69:
FTW, the futex tracing is still here which is the major part.

PR #890, opened 2024-11-04
scx_layered: add timer antistall

likewhatevs:
This PR adds a timer to layered that will, once per second, replace dispatch logic with logic to check all DSQs for tasks which have not been ran within the past 5 seconds.

When this check finds such a task, dispatch consumes that DSQ.

This is continued for all runs of dispatch until there are no such tasks.

This makes `stress-ng -c 80 -M --affinity 80 -t 60` (which reproduces stalls on my machine on main) not reproduce stalls.

Screenshot_20241104_133231

The configs used were:
```
❯ cat repos/layered-config.json 
[
{
  "name":"patso",
  "comment":"patso user",
  "matches":[
     [{"UIDEquals":1000}]
  ],
  "kind": {
    "Open":{
      "slice_us": 1000,
      "preempt": false,
      "preempt_first": false,
      "exclusive": false,
      "weight": 100
     }
  }
},{
  "name":"stress-ng",
  "comment":"stress-ng slice",
  "matches":[
     [
	    {"CommPrefix":"stress-ng"}
     ],[
	    {"PcommPrefix":"stress-ng"}
     ]],
  "kind": {
    "Open":{
      "slice_us": 1000,
      "preempt": false,
      "preempt_first": false,
      "exclusive": false,
      "weight": 100
     }
  }
},
{
  "name":"normal",
  "comment":"the rest",
  "matches":[[]],
  "kind":{
    "Open": {
      "slice_us": 1000,
      "preempt": false,
      "preempt_first": false,
      "exclusive": false,
      "weight": 200
    }
  }
}]

❯ cat repos/new-layered-config.json 
[
  {
    "name": "random",
    "comment": "random threads",
    "matches": [
      [
        {
          "CommPrefix": "dontmatter"
        }
      ],
      [
        {
          "CommPrefix": "DONTMATTER"
        }
      ]
    ],
    "kind": {
      "Open": {
        "min_exec_us": 0,
        "slice_us": 1000,
        "preempt": false,
        "exclusive": false,
        "growth_algo": "Random",
        "weight": 200
      }
    }
},
  {
    "name": "all",
    "comment": "all threads",
    "matches": [
      []
    ],
    "kind": {
      "Open": {
        "min_exec_us": 50,
        "slice_us": 1000,
        "preempt": false,
        "exclusive": false,
        "growth_algo": "Random",
        "weight": 100
      }
    }
  }
]

```

Note -- this does not fix this issue when run-example config is used with that stress test. I *think* weights working properly with fallbacks solves that though (and that is another thing WIP), because I think the issue with that is that it effectively dumps all load into a single layer (i.e. I think this could fix it if timer frequency was higher, but also like, that effectively caps the ability to prioritize so weights are probably better for preventing stalls in the "bad config" case).
```
DEBUG DUMP

stress-ng-cpu[191725] triggered exit kind 1026:
  runnable task stall (watchdog failed to check in for 30.004s)

Backtrace:
  scx_tick+0x81/0x90
  sched_tick+0xdf/0x2d0
  update_process_times+0x96/0xb0
  tick_nohz_handler+0x8b/0x130
  __hrtimer_run_queues+0x11b/0x280
  hrtimer_interrupt+0xfa/0x230
  __sysvec_apic_timer_interrupt+0x4f/0x110
  sysvec_apic_timer_interrupt+0x38/0x90

LAYER[0][batch] nr_cpus=2 nr_queued=26 -22963ms cpus=01......|........|........|........|
LAYER[1][immediate] nr_cpus=1 nr_queued=0 -0ms cpus=
LAYER[2][stress-ng] nr_cpus=27 nr_queued=0 -0ms cpus=
LAYER[3][normal] nr_cpus=2 nr_queued=101 -24731ms cpus=
HI_FALLBACK[1024] nr_queued=30 -21313ms
LO_FALLBACK nr_queued=0 -0ms
COST GLOBAL[0][batch] budget=8000000000 capacity=8000000000
COST GLOBAL[1][immediate] budget=0 capacity=0
COST GLOBAL[2][stress-ng] budget=0 capacity=0
COST GLOBAL[3][normal] budget=0 capacity=0
COST CPU[0][0][batch] budget=4999998739034397 capacity=5000000000000000
COST CPU[0][1][batch] budget=5000000000000000 capacity=5000000000000000
COST CPU[0][2][batch] budget=199972794717495 capacity=200000000000000
COST CPU[0][3][batch] budget=4999999494460853 capacity=5000000000000000
COST CPU[1][0][immediate] budget=4999998739034397 capacity=5000000000000000
COST CPU[1][1][immediate] budget=5000000000000000 capacity=5000000000000000
COST CPU[1][2][immediate] budget=199972794717495 capacity=200000000000000
COST CPU[1][3][immediate] budget=4999999494460853 capacity=5000000000000000
COST CPU[2][0][stress-ng] budget=4999998739034397 capacity=5000000000000000
COST CPU[2][1][stress-ng] budget=5000000000000000 capacity=5000000000000000
COST CPU[2][2][stress-ng] budget=199972794717495 capacity=200000000000000
COST CPU[2][3][stress-ng] budget=4999999494460853 capacity=5000000000000000
COST CPU[3][0][normal] budget=4999998739034397 capacity=5000000000000000
COST CPU[3][1][normal] budget=5000000000000000 capacity=5000000000000000
COST CPU[3][2][normal] budget=199972794717495 capacity=200000000000000
COST CPU[3][3][normal] budget=4999999494460853 capacity=5000000000000000
```

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1,6 +1,8 @@
 /* Copyright (c) Meta Platforms, Inc. and affiliates. */
 #ifdef LSP
+#ifndef __bpf__
 #define __bpf__
+#endif
 #define LSP_INC
 #include "../../../../include/scx/common.bpf.h"
 #include "../../../../include/scx/ravg_impl.bpf.h"
@@ -39,12 +41,18 @@ const volatile s32 __sibling_cpu[MAX_CPUS];
 const volatile bool monitor_disable = false;
 const volatile unsigned char all_cpus[MAX_CPUS_U8];
 const volatile u32 layer_iteration_order[MAX_LAYERS];
+/* Flag to enable or disable antistall feature */
+const volatile bool enable_antistall = true;
+/* Delay permitted, in seconds, before antistall activates */
+const volatile u64 antistall_sec = 3;
 
 private(all_cpumask) struct bpf_cpumask __kptr *all_cpumask;
 private(big_cpumask) struct bpf_cpumask __kptr *big_cpumask;
 struct layer layers[MAX_LAYERS];
 u32 fallback_cpu;
 static u32 preempt_cursor;
+
+extern unsigned CONFIG_HZ __kconfig;
 
 #define dbg(fmt, args...)	do { if (debug) bpf_printk(fmt, ##args); } while (0)
 #define trace(fmt, args...)	do { if (debug > 1) bpf_printk(fmt, ##args); } while (0)
@@ -1233,6 +1241,107 @@ no:
 	return false;
 }
 
+/* Mapping of cpu to most delayed DSQ it can consume */
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__type(key, u32);
+	__type(value, u64);
+	__uint(max_entries, 1);
+} antistall_cpu_dsq SEC(".maps");
+
+/* Mapping cpu to delay of highest delayed DSQ it can consume */
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__type(key, u32);
+	__type(value, u64);
+	__uint(max_entries, 1);
+} antistall_cpu_max_delay SEC(".maps");
+
+/**
+ * get_delay_sec() - get runnable_at delay of a task_struct in seconds.
+ * @p: task_struct *p
+ * @jiffies_now: time from which to measure delay, in jiffies.
+ *
+ * Return: runnable_at delay, if any exists, in seconds.
+ */
+int get_delay_sec(struct task_struct *p, u64 jiffies_now) 
+{
+	u64 runnable_at, delta_secs;
+	runnable_at = READ_ONCE(p->scx.runnable_at);
+
+	if (vtime_before(runnable_at, jiffies_now)) {
+		delta_secs = (jiffies_now - runnable_at) / CONFIG_HZ;
+	} else {
+		delta_secs = 0;
+	}
+
+	return delta_secs;
+}
+
+/**
+ * antistall_consume() - consume delayed DSQ
+ * @cpu: cpu number
+ * @cctx: cpu context
+ * 
+ * This function consumes a delayed DSQ. This is meant to be called
+ * from dispatch, before any other logic which could result in a 
+ * DSQ being consumed.
+ *
+ * This is meant to prevent issues such as DSQs with affinitized tasks
+ * stalling when all CPUs which can process them are continually consuming
+ * other DSQs.
+ *
+ * Return: bool indicating if a DSQ was consumed or not.
+ */
+bool antistall_consume(s32 cpu, struct cpu_ctx *cctx)
+{
+	u64 *antistall_dsq, jiffies_now, cur_delay;
+	u32 zero;
+	bool consumed;
+	struct task_struct *p;
+	
+	cur_delay = 0;	
+	consumed = false;
+	zero = 0;
+
+	if (!enable_antistall)
+		return false;
+
+	if (!cctx || !cctx->layer_idx || !cpu)
+		return false;
+	
+	antistall_dsq = bpf_map_lookup_elem(&antistall_cpu_dsq, &zero);
+	
+	if (!antistall_dsq) {
+		scx_bpf_error("cant happen");
+		return false;
+	}
+
+	if (*antistall_dsq == SCX_DSQ_INVALID)
+		return false;
+	
+	consumed = scx_bpf_consume(*antistall_dsq);
+	
+	if (!consumed) 
+		goto reset;
+	
+	jiffies_now = bpf_jiffies64();
+
+	bpf_for_each(scx_dsq, p, *antistall_dsq, 0) {
+		cur_delay = get_delay_sec(p, jiffies_now);
+
+		if (cur_delay > antistall_sec)
+			return consumed;
+		
+		goto reset;	
+	}
+	
+reset:
+	trace("antistall reset DSQ[%llu] SELECTED_CPU[%llu] DELAY[%llu]", *antistall_dsq, cpu, cur_delay);
+	*antistall_dsq = SCX_DSQ_INVALID;
+	return consumed;
+}
+
 static __noinline
 void layered_dispatch_no_topo(s32 cpu, struct task_struct *prev)
 {
@@ -1244,6 +1353,9 @@ void layered_dispatch_no_topo(s32 cpu, struct task_struct *prev)
 	s32 sib = sibling_cpu(cpu);
 
 	if (!(cctx = lookup_cpu_ctx(-1)) || !(costc = lookup_cpu_cost(cpu)))
+		return;
+
+	if (antistall_consume(cpu, cctx))
 		return;
 
 	/*
@@ -1455,6 +1567,9 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 	    !(costc = lookup_cpu_cost(cpu)))
 		return;
 
+	if (antistall_consume(cpu, cctx))
+		return;
+	
 	/*
 	 * if @prev was on SCX and is still runnable, we are here because @prev
 	 * has exhausted its slice. We may want to keep running it on this CPU
@@ -2078,7 +2193,7 @@ void BPF_STRUCT_OPS(layered_exit_task, struct task_struct *p,
 	struct cpu_ctx *cctx;
 	struct task_ctx *tctx;
 
-	if(args->cancelled){
+	if (args->cancelled) {
 		return;
 	}
 
@@ -2258,12 +2373,133 @@ static bool layered_monitor(void)
 	return true;
 }
 
+/**
+ * antistall_set() - set antistall flags.
+ * @dsq_id: Dsq to check for delayed tasks.
+ * @jiffies_now: Time to check against in jiffies.
+ *
+ * This function sets entries in antistall_cpu_dsq used by antistall.
+ * It checks the given DSQ to see if delay exceeds antistall_sec.
+ * It tries to find a CPU satisfying the constraints of "can run the first
+ * task in the provided DSQ" and "is not already flagged for use in antistall".
+ * If it cannot find such a CPU to flag, it will try to flag a CPU flagged to
+ * process another with a lesser delay if one exists.
+ */
+u64 antistall_set(u64 dsq_id, u64 jiffies_now) 
+{
+	struct task_struct *p;
+	struct task_ctx *tctx;
+	s32 cpu;
+	u64 *antistall_dsq, *delay, cur_delay;
+	bool first_pass;
+	u32 zero;
+	
+	zero = 0;
+
+	if (!dsq_id || !jiffies_now)
+		return 0;
+	
+	// verifier
+	bpf_rcu_read_lock();	
+	bpf_for_each(scx_dsq, p, dsq_id, 0) {
+		
+		if (!(tctx = lookup_task_ctx(p)))
+			goto unlock;
+		
+		cur_delay = get_delay_sec(p, jiffies_now);
+		if (cur_delay <= antistall_sec)
+			// check head task in dsq
+			goto unlock;
+		
+		first_pass = true;
+look_for_cpu:
+		bpf_for(cpu, 0, nr_possible_cpus) {
+			if (!tctx->layered_cpumask)
+				goto unlock;
+			
+			if (!bpf_cpumask_test_cpu(cpu,   cast_mask(tctx->layered_cpumask)))
+				continue;
+			
+			antistall_dsq = bpf_map_lookup_percpu_elem(&antistall_cpu_dsq, &zero, cpu);
+			delay = bpf_map_lookup_percpu_elem(&antistall_cpu_max_delay, &zero, cpu);
+			
+			if (!antistall_dsq || !delay) {
+				scx_bpf_error("cant happen");
+				goto unlock;
+			}
+
+			if (*antistall_dsq == SCX_DSQ_INVALID) {
+				trace("antistall set DSQ[%llu] SELECTED_CPU[%llu] DELAY[%llu]", dsq_id, cpu, cur_delay);
+				*delay = cur_delay;
+				*antistall_dsq = dsq_id;
+				goto unlock;
+			}
+
+			if (first_pass)
+				continue;
+				
+			if (*delay < cur_delay) {
+				trace("antistall set DSQ[%llu] SELECTED_CPU[%llu] DELAY[%llu]", dsq_id, cpu, cur_delay);
+				*delay = cur_delay;
+				*antistall_dsq = dsq_id;
+				goto unlock;
+			}
+		}		
+		
+		if (first_pass) {
+			first_pass = false;
+			goto look_for_cpu;
+		}
+
+		goto unlock;
+	}
+
+unlock:
+	bpf_rcu_read_unlock();
+	return 0;
+}
+
+/**
+ * antistall_scan() - call antistall_set on all DSQs.
+ *
+ * This function calls antistall_set on all DSQs.
+ * This is where antistall figures out what work, if any, needs
+ * to be prioritized to keep runnable_at delay at or below antistall_sec.
+ */
+static bool antistall_scan(void)
+{
+	s32 cpu;
+	u64 dsq_id;
+	u64 jiffies_now;
+
+	if (!enable_antistall)
+		return true;
+
+	jiffies_now = bpf_jiffies64();
+	
+	bpf_for(dsq_id, 0, nr_layers) {
+		antistall_set(dsq_id, jiffies_now);
+	}
+
+	bpf_for(cpu, 0, nr_possible_cpus) {
+		dsq_id = cpu_hi_fallback_dsq_id(cpu);
+		antistall_set(dsq_id, jiffies_now);
+	}
+	
+	antistall_set(LO_FALLBACK_DSQ, jiffies_now);
+	
+	antistall_set(HI_FALLBACK_DSQ_BASE, jiffies_now);
+	
+	return true;
+}
 
 static bool run_timer_cb(int key)
 {
 	switch (key) {
 	case LAYERED_MONITOR:
 		return layered_monitor();
+	case ANTISTALL_TIMER:
+		return antistall_scan();
 	case NOOP_TIMER:
 	case MAX_TIMERS:
 	default:
@@ -2273,6 +2509,7 @@ static bool run_timer_cb(int key)
 
 struct layered_timer layered_timers[MAX_TIMERS] = {
 	{15LLU * NSEC_PER_SEC, CLOCK_BOOTTIME, 0},
+	{1LLU * NSEC_PER_SEC, CLOCK_BOOTTIME, 0},
 	{0LLU, CLOCK_BOOTTIME, 0},
 };
 
@@ -2285,6 +2522,10 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
 	struct bpf_cpumask *cpumask, *tmp_big_cpumask;
 	struct cpu_ctx *cctx;
 	int i, j, k, nr_online_cpus, ret;
+	u64 *init_antistall_delay, *init_antistall_dsq;
+	u32 zero;
+
+	zero = 0;
 
 	ret = scx_bpf_create_dsq(LO_FALLBACK_DSQ, -1);
 	if (ret < 0)
@@ -2303,6 +2544,11 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
 	nr_online_cpus = 0;
 	bpf_for(i, 0, nr_possible_cpus) {
 		const volatile u8 *u8_ptr;
+		
+	  init_antistall_dsq = bpf_map_lookup_percpu_elem(&antistall_cpu_dsq, &zero, i);
+		if (init_antistall_dsq) {
+			*init_antistall_dsq = SCX_DSQ_INVALID;
+		}
 
 		if (!(cctx = lookup_cpu_ctx(i))) {
 			bpf_cpumask_release(cpumask);
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -187,6 +187,7 @@ jobs:
           matrix:
             scheduler: [ scx_layered ]
             topo: ['--disable-topology=false', '--disable-topology=true']
+            antistall: ['', '--disable-antistall']
           fail-fast: false
     steps:
       # prevent cache permission errors
@@ -228,7 +229,7 @@ jobs:
         run: exit -1
 
       # The actual build:
-      - run: meson setup build -Dkernel=../linux/arch/x86/boot/bzImage -Dkernel_headers=../linux -Denable_stress=true -Dvng_rw_mount=true -Dextra_sched_args=" ${{ matrix.topo }}"
+      - run: meson setup build -Dkernel=../linux/arch/x86/boot/bzImage -Dkernel_headers=../linux -Denable_stress=true -Dvng_rw_mount=true -Dextra_sched_args=" ${{ matrix.topo }} ${{ matrix.antistall }}"
       - run: meson compile -C build ${{ matrix.scheduler }}
 
       # Print CPU model before running the tests (this can be useful for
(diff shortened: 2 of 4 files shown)


likewhatevs:
same results as above (including the bit w/ run example), but ci failure, odd:
Screenshot_20241104_140825

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2207,6 +2301,7 @@ static bool run_timer_cb(int key)
 struct layered_timer layered_timers[MAX_TIMERS] = {
 	{15LLU * NSEC_PER_SEC, CLOCK_BOOTTIME, 0},
 	{0LLU, CLOCK_BOOTTIME, 0},
+	{1LLU * NSEC_PER_SEC, CLOCK_BOOTTIME, 0},
```
hodgesds, line 2304: This changes the timer interval of the `NOOP_TIMER` and `MAX_TIMER` from the ordering of `layer_timer_callbacks`. I think the switch on 2289 needs to match the order of the `layered_timers`

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1211,80 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+	u64 dsq_id, now;
+	u32 idx;
+	
+	// verifier
+	if (!cctx)
+		return false;
+	if(!cctx->layer_idx)
+		return false;
+
+	if(dsq_timer_check_flag){
```
hodgesds, line 1226: nit: just return earlier in this case, less nesting.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1211,80 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
 }
 
+bool timer_antistall(s32 cpu, struct cpu_ctx *cctx){
+	struct task_struct *p;
+	struct task_ctx *tctx;
+	u64 dsq_id, now;
+	u32 idx;
+	
+	// verifier
+	if (!cctx)
```
hodgesds, line 1221: nit:
```
if (!cctx || !cctx->layer_idx)
    return false;
```

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1211,80 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+	if(!cctx->layer_idx)
+		return false;
+
+	if(dsq_timer_check_flag){
+		bpf_for(idx, 0, nr_layers) {
+			dsq_id = rotate_layer_id(cctx->layer_idx, idx);
+
+			// check non-fallback dsqs.
+			bpf_for_each(scx_dsq, p, dsq_id, 0) {
+				now = bpf_ktime_get_ns();
```
hodgesds, line 1232: I'd maybe move out the `now` collection  to outside the loops. It's probably not worth trying to get up to the nanosecond level precision.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: think the last commit fixes that

hodgesds:
One other thing to maybe add is a CLI arg that enables/disables the timer. Maybe we should have some macros to make creating timers easier.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1211,79 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+
+	bpf_for(idx, 0, nr_layers) {
+		dsq_id = rotate_layer_id(cctx->layer_idx, idx);
+
+		// check non-fallback dsqs.
+		bpf_for_each(scx_dsq, p, dsq_id, 0) {
+
+			if ((tctx = lookup_task_ctx(p))) {
+				if(tctx->running_at) {
+					if(now - tctx->running_at > NSEC_PER_MSEC * 5) {
```
htejun, line 1236: `running_at` is the last the the task got on the CPU. `p->scx.runnable_at` is the time when the task became runnable (ie. the task has been waiting to go on a CPU since this time) which is the right timestamp to calculate delta from. Note that `p->scx.runnable_at` is in jiffies. You can get the current jiffies with `bpf_jiffies64()` and the frequency of the current kernel with `extern unsigned CONFIG_HZ __kconfig;`.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1211,79 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+						break;
+					}
+				}
+			}
+		}
+	}
+
+	// check fallback dsqs
+	dsq_id = cpu_hi_fallback_dsq_id(cpu);
+	bpf_for_each(scx_dsq, p, dsq_id, 0) {
```
htejun, line 1250: Would it make sense to make this into a subroutine to reduce code repetition?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -46,6 +48,7 @@ struct layer layers[MAX_LAYERS];
 u32 fallback_cpu;
 static u32 preempt_cursor;
 
+static volatile uint32_t dsq_timer_check_flag = 0;
```
htejun, line 51: Maybe use the same word consistently? ie. if the feature is antistall, use the word in the flag too?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2191,13 +2272,25 @@ static bool layered_monitor(void)
 	return true;
 }
 
+static bool timer_antistall_reset(void){
+	if(dsq_timer_check_flag) {
+		// flag was not cleared, ignore this tick.
+		return true;
+	} else {
+		// flag was cleared, reset it.
+		__sync_fetch_and_or(&dsq_timer_check_flag, 1);
```
htejun, line 2281: Atomic ops isn't adding anything here. I'd just do `WRITE_ONCE(dsq_timer_check_flag, 1);`.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2191,13 +2272,25 @@ static bool layered_monitor(void)
 	return true;
 }
 
+static bool timer_antistall_reset(void){
```
htejun, line 2275: `set` and `reset` usually indicating turning a bit to 1 and 0 respectively which makes this naming rather confusing.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1211,79 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+				} else {
+					// if the first entry in a dsq is ok, the dsq is ok.
+					break;
+				}
+			}
+		}
+	}
+
+	// all dsqs are ok, clear check flag.
+	__sync_fetch_and_and(&dsq_timer_check_flag, 0);
```
htejun, line 1283: Control reaching here doesn't indicate that there are no stalled tasks. e.g. if a task which is affined to one CPU is stalled and a different CPU ended up running antistall before it, that CPU could reach here and clear the flag before the one which can run the stalled task can run antistall.

The code needs to track whether it saw any stalled task or not separately from consumption succeeding or not.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2191,13 +2272,25 @@ static bool layered_monitor(void)
...
+}
 
 static bool run_timer_cb(int key)
 {
 	switch (key) {
 	case LAYERED_MONITOR:
 		return layered_monitor();
 	case NOOP_TIMER:
+	case HACK_DSQ_CHECK:
+		return timer_antistall_reset();
```
htejun, line 2293: As this can put multiple CPUs into the antistall code path, it'd be better if the timer itself can scan and schedule antistall only when it's necessary. Also, note that, in newer kernels, timer can dispatch tasks directly to each CPU's local DSQs or the HI_FALLBACK. Given that we need to deploy to older kernels too, the current implementation is better for now.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: ooh that makes sense wrt/ the outstanding stalls I think, thanks!

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: thanks!

likewhatevs: hmm this works locally, I think I need to edit a kconfig to make it work on ci (will figure it out and add it to PR):

```
09:50:59 [DEBUG] libbpf: object 'bpf_bpf': failed (-22) to create BPF token from '/sys/fs/bpf', skipping optional step...

09:50:59 [WARN] libbpf: failed to open system Kconfig
```

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1213,77 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+	struct task_struct *p;
+	struct task_ctx *tctx;
+	u64 runnable_at_sec;
+	
+	if (!dsq_id || !now_sec)
+		return false;
+	
+	bpf_for_each(scx_dsq, p, dsq_id, 0) {
+		if ((tctx = lookup_task_ctx(p))) {
+			if (tctx->runnable_at) {
```
htejun, line 1226: 0 is a valid timestamp. In fact, jiffies start at a small negative offset from 0 to force wrapping sometime after boot. No need to test for 0.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1213,77 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+	}
+	return false;
+}
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, s32);
+	__type(value, bool);
+	__uint(max_entries, 4096);
+} antistall_cpus SEC(".maps");
```
htejun, line 1246: I'd use `BPF_MAP_TYPE_PERCPU_ARRAY` which contains one bool element.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2191,13 +2272,80 @@ static bool layered_monitor(void)
...
+	struct task_struct *p;
+	struct task_ctx *tctx;
+	u64 runnable_at_sec;
+	s32 cpu;
+	bool in_antistall;
+	in_antistall = true;
+
+	if (!dsq_id || !now_sec)
+		return;
+	bpf_rcu_read_lock();
```
htejun, line 2286: I suppose verifier complains without RCU read lock?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2191,13 +2272,80 @@ static bool layered_monitor(void)
...
+	in_antistall = true;
+
+	if (!dsq_id || !now_sec)
+		return;
+	bpf_rcu_read_lock();	
+	bpf_for_each(scx_dsq, p, dsq_id, 0) {
+		if (!(tctx = lookup_task_ctx(p)))
+			goto unlock;
+	
+		if (!tctx->runnable_at)
```
htejun, line 2291: I don't think this is necessary.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2191,13 +2272,80 @@ static bool layered_monitor(void)
...
+		
+		if (now_sec - runnable_at_sec < runnable_at_sec)
+			goto unlock;
+
+		bpf_for(cpu, 0, nr_possible_cpus) {
+			if (!tctx->layered_cpumask)
+				goto unlock;
+			if (bpf_cpumask_test_cpu(cpu,   cast_mask(tctx->layered_cpumask))) {
+			// when multiple cpus can help prevent a stall,
+			// and dsqs aren't clearing fast enough, let them
```
htejun, line 2304: Unfinished sentence? White spaces in the if block and indentation of the comment look off too.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2191,13 +2272,80 @@ static bool layered_monitor(void)
...
+		
+		runnable_at_sec = tctx->runnable_at / CONFIG_HZ;
+		
+		if (now_sec - runnable_at_sec < runnable_at_sec)
+			goto unlock;
+
+		bpf_for(cpu, 0, nr_possible_cpus) {
+			if (!tctx->layered_cpumask)
+				goto unlock;
+			if (bpf_cpumask_test_cpu(cpu,   cast_mask(tctx->layered_cpumask))) {
```
htejun, line 2302: Neat.

htejun (commented):
Left some comments but in general looks great to me. Does this now prevent all stalls in your testsing?

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: I wanted to do this but I did not see a way to write to `BPF_MAP_TYPE_PERCPU_ARRAY` from another CPU (i.e. I saw a way to read from other CPUs, but not write).

I also didn't see anything on timers indicating that they are per cpu, so I figured the timer is running on one CPU, and that CPU needs to write to the maps for other CPUs.

Is my understanding wrong?

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: yep

likewhatevs:
> Left some comments but in general looks great to me. Does this now prevent all stalls in your testsing?

Not yet, need to keep debugging, thx for the feedback!

htejun:
`bpf_map_lookup_percpu_elem()` returns pointer to the element, so I think it allows updating any CPU's value from any other CPU.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1208,6 +1213,81 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+	u64 runnable_at_sec;
+	
+	if (!dsq_id || !now_sec)
+		return false;
+	
+	bpf_for_each(scx_dsq, p, dsq_id, 0) {
+		if ((tctx = lookup_task_ctx(p))) {
+			if (tctx->runnable_at) {
+				runnable_at_sec = tctx->runnable_at / CONFIG_HZ;
+					scx_bpf_error("%llu %llu", now_sec, runnable_at_sec);
```
likewhatevs, line 1228: debugging and this looks odd?

this line of log prints bpf_jiffies64/CONFIG_HZ and tctx->runnable_at/CONFIG_HZ
```
Error: EXIT: scx_bpf_error (14335776 65064116979)
```

likewhatevs: I think this is relevant to that (making sense of that still):
```
❯ sudo cat /proc/timer_list  | grep jif 
  .idle_jiffies   : 4301162074
  .last_jiffies   : 4301162075
jiffies: 4301162075
  .idle_jiffies   : 4301161701
  .last_jiffies   : 4301161701
```

likewhatevs: no idea what task_ctx.runnable_at is but the one to use is the one on the task struct

htejun: Yeah, `task_ctx->runnable_at` is set by layered from `bpf_ktime_get_ns()`. `p->scx.runnable_at` is the one in jiffies. Alternatively, as `task_ctx->runnable_at` exists already, you can just use `bpf_ktime_get_ns()` as `now` too.

likewhatevs:
> Left some comments but in general looks great to me. Does this now prevent all stalls in your testsing?

It does now :smile:  (although I still need to clean things up and change to percpu array's).

I noticed that like, under severe contention the timer wasn't firing (nor was some of the userspace logic).

I added in a metric called pressure and started dividing slice time by that in addition to preempting usual dispatch logic when tasks take too long to run. This seems to work well at 3s (i.e. setting it to 3s means after around 5s of effectively stalled kinda all-over, slice times start getting smaller to enable the scheduler to recover w/o getting kicked).

likewhatevs:
> `bpf_map_lookup_percpu_elem()` returns pointer to the element, so I think it allows updating any CPU's value from any other CPU.

I'm pretty sure that works but the problem I had was initializing that right. I *think* initialization has to be done on the CPU that owns the map because lookup returns null when the key or cpu doesn't exist. Idk if trying to do that in dispatch is better or worse than using a regular map.

likewhatevs:
Added a disable flag to make testing easier.

This doesn't totally fix stalls, but it does increase how overloaded a system can be before the scheduler is kicked for stalling.

One thing I noticed w/ this is that like, at some point beyond the amount of load where the scheduler would be kicked before this PR (within the range where this PR can like, keep delay under control), some code in (I think, I really don't know) task init hangs hard (i.e. I could reliably cause a stall by typing htop into konsole and hitting enter. Once the scheduler was kicked, htop would open).

I also think at some point in time we maybe need to like, prioritize the userspace component and (if possible) prioritize the execution of bpf code because these were also things I saw start having gaps in logs as the amount of load was increased.

likewhatevs:
This PR paired w/ the latest weight commits and this config to leverage them: https://gist.github.com/likewhatevs/8e960c3f3dcca3f6835cd551fdc0b97a

Enabled `stress-ng -c 8192 -M --affinity 8192 -t 60` to run for 33 minutes before stalling:
Screenshot_20241107_133101

I saw mouse move sometime while waiting for that and the logs showed delays bouncing up and down so I think it worked.

likewhatevs:
The latest changes remove pressure/messing with time slice size.

That was doing a lot wrt/ preventing stalls, getting this to continue to prevent stalls under reasonably near-full (but not over) load with affinitized tasks required:

* Adding CPUs to those working on clearing delayed DSQs on each timer tick (and prioritizing the most delayed DSQs if all are in use).
* Having CPUs work on clearing delayed DSQs in a sticky manner (i.e. keep overriding dispatch until delayed DSQs are under antistall_sec).

This seems to work reasonably well. 

Here are some logs of it working on a 32 thread machine running `stress-ng -c 31 -M --affinity 100 -t 60` (which did not stall), w/ extra logs around consume (removed that because it's noisy):

```
   stress-ng-cpu-378832  [010] ..s1. 14008.755962: bpf_trace_printk: antistall set DSQ[3] SELECTED_CPU[2] 
DELAY[19]
   stress-ng-cpu-378841  [002] dN.31 14008.756372: bpf_trace_printk: antistall consume 1
     HangWatcher-369698  [002] dN.31 14008.756385: bpf_trace_printk: antistall consume 1
 xfsaild/nvme1n1-1352    [002] d..31 14008.756395: bpf_trace_printk: antistall consume 1
 xfsaild/nvme1n1-1352    [002] d..31 14008.756396: bpf_trace_printk: antistall reset DSQ[3] SELECTED_CPU[2
] DELAY[1]
   stress-ng-cpu-378832  [010] ..s1. 14009.755957: bpf_trace_printk: antistall set DSQ[3] SELECTED_CPU[2] 
DELAY[20]
   stress-ng-cpu-378841  [002] dN.31 14009.756353: bpf_trace_printk: antistall consume 1
   stress-ng-cpu-378841  [002] dN.31 14009.756355: bpf_trace_printk: antistall reset DSQ[3] SELECTED_CPU[2
] DELAY[2]
```

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1218,6 +1225,73 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+	u32 zero;
+	bool consumed;
+	struct task_struct *p;
+	
+	cur_delay = 0;	
+	consumed = false;
+	zero = 0;
+	empty = -1;
+
+	if(!enable_antistall)
```
htejun, line 1257: Can you cleanup whitespace usages here and other places?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1218,6 +1225,73 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+		return false;
+
+	if (!cctx || !cctx->layer_idx || !cpu)
+		return false;
+	
+	antistall_dsq = bpf_map_lookup_elem(&antistall_cpu_dsq, &zero);
+	antistall_cpu_max = bpf_map_lookup_elem(&antistall_cpu_max_delay, &zero);
```
htejun, line 1264: What does this variable do? It looks like it's just used to verify that `antistall_dsq` is valid. If that's the case, you can just use `SCX_DSQ_INVALID` to indicate that the value is not valid.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2218,12 +2298,114 @@ static bool layered_monitor(void)
...
+		return 0;
+	bpf_rcu_read_lock();	
+	bpf_for_each(scx_dsq, p, dsq_id, 0) {
+		
+		if (!(tctx = lookup_task_ctx(p)))
+			goto unlock;
+		
+		runnable_at_sec = p->scx.runnable_at / CONFIG_HZ;
+		cur_delay = now_sec - runnable_at_sec;
+		if (cur_delay <= antistall_sec || cur_delay > 100)
```
htejun, line 2323: Why does it jump to unlock if `cur_delay > 100`?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2218,12 +2298,114 @@ static bool layered_monitor(void)
...
+				trace("antistall set DSQ[%llu] SELECTED_CPU[%llu] DELAY[%llu]", dsq_id, cpu, cur_delay);
+				*delay = cur_delay;
+				*antistall_dsq = dsq_id;
+				goto unlock;
+			}
+
+			if (first_pass)
+				continue;
+				
+			if (*delay < cur_delay){
```
htejun, line 2354: Ah, okay, you're picking the more delayed one. Still, I think it'd be less confusing to use `SCX_DSQ_INVALID` as the empty marker and use `antistall_cpu_max_delay` only in this function.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2218,12 +2298,114 @@ static bool layered_monitor(void)
...
+	bool first_pass;
+	u32 zero;
+	
+	zero = 0;
+	empty = -1;
+
+	if (!dsq_id || !now_sec)
+		return 0;
+	bpf_rcu_read_lock();	
+	bpf_for_each(scx_dsq, p, dsq_id, 0) {
```
htejun, line 2316: Comment explaining what's going on would be great.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1218,6 +1225,73 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
+	u64 *antistall_dsq, *antistall_cpu_max, 
+			empty, runnable_at_sec, now_sec, cur_delay;
+	u32 zero;
+	bool consumed;
+	struct task_struct *p;
+	
+	cur_delay = 0;	
+	consumed = false;
+	zero = 0;
+	empty = -1;
```
htejun, line 1255: Note that while `(u64)-1` is an invalid DSQ ID, it might not necessarily stay that way. Please use `SCX_DSQ_INVALID`.

htejun (approved):
Generally looks great to me. Left some minor comments.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: Sometimes delay numbers make no sense (i.e. I've seen it go as high as 2k IIRC). I figured that happens because this is inherently racy maybe. Will post logs of this in a bit.

htejun: So, jiffies wrap and I don't think you can convert it to secs and then calculate delta. You should first calculate delta and then divide so that wrapping doesn't affect the result.

htejun: Something like:
```
now = bpf_jiffies64();
runable_at = READ_ONCE(p->scx.runnable_at);

if (vtime_after(now, runnable_at)) // note that runnable_at can be in the future
        delta_secs = (now - runnable_at) / CONFIG_HZ;
else
        delta_secs = 0;
```

likewhatevs: Thanks, that did it, no weird delay numbers anymore.

PR #891, opened 2024-11-04
scx_layered: point costc to global struct when initializing budgets

etsal:
During ```initialize_budgets()``` the ```costc``` pointer is used to initialize the global and local cost structs with the layer weights. However, the code only initializes the global cost struct for the first loop, after which the ```costc``` pointer points to a local cost struct instead. Initialize ```costc``` on every loop to point to the global cost struct and initialize all of its weights.


hodgesds (approved):
LGTM, might be worth putting some of the trace values in the PR for tracking purposes.

PR #892, opened 2024-11-04
scx_layered: remove ->open from layer struct

etsal:
The ```struct layer``` that represents each ```scx_layered``` layer includes both an enum ```kind``` that represents whether the layer is open, grouped or confined, and a boolean ```open``` that represents whether the layer can expand past its current CPU mask. Whether the ```open``` variable is set depends completely on its kind, so the variable is not necessary. Remove it and update checks that use it to directly use ```kind``` instead.


hodgesds (approved):
LGTM, thanks for plumbing the layer types through!

Issue #863, opened 2024-10-30
scx_utils: Provide scheduler flags in a compatible way


htejun:
Some of SCX flags are derived from scheduler core flags and are subject to change. It's less likely but other flags may change too. When the flag values are used directly, they're substituted to the value that was defined at the time of compilation and the binary wouldn't be able to function correctly when loaded on a kernel which uses different values for these flags.

All the SCX flags are enums and can be obtained using CO-RE helpers. Provide compat helpers that put the flag values in `const volatile` variables which are set at load time so that scheduler binaries can work across kernels with different flag values.

etsal:
Completed with #894

PR #894, opened 2024-11-05
introduce CO:RE enum readers and use them for scx_central

etsal:
The values of kernel enums get hardcoded into the BPF scheduler at compile time, which possibly prevents us from deploying the compiled BPF program on kernels with modified values for these enums (as explained in #863). CO:RE methods provide a solution to this by allowing the scheduler to read enums at load time from the running kernel.

Introduce an ```scx_enums``` struct that is populated at runtime with the enum values from the running kernel, and replace hardcoded enum values in the code with accesses to that struct. This diff demonstrates how to do so in existing schedulers using ```scx_central``` as a proof of concept.


scheds/include/scx/enums.bpf.h
```diff
@@ -0,0 +1,75 @@
...
+#ifndef __SCX_ENUMS_BPF_H
+#define __SCX_ENUMS_BPF_H
+
+#include <scx/enums.h>
+
+#define SCX_ENUM_DEFINE() \
+	struct scx_enums _scx_enum;
+
+#define SCX_ENUM(name) \
+	(_scx_enum.SCX_##name)
```
htejun, line 16: This makes each enum access a memory dereference. If you make each enum a `const volatile` variable and set them from the userspace side during init, then the verifier will handle them as consts and we can init them as a part of the `SCX_OPS_LOAD()` or `SCX_OPS_ATTACH()`.

For example, there can be `const volatile u64 __SCX_ENQ_WAKEUP;` which is initialized during load and then `#define SCX_ENQ_WAKEUP __SCX_ENQ_WAKEUP` would make it transparent for the scheduler implementations.

etsal: I tried setting it that way, but there was the issue where the userspace component didn't have access to the ```vmlinux.h``` definitions and so could not set the initial value of the variables. I tried including ```vmlinux.h``` and got a bunch of errors but they must have been due to improperly structuring the header includes - I will go back to that and update.

htejun: I think it's fine to fail loading if the enums are not available through BTF.

etsal:
New version that sets all the enums before loading the scheduler. Two points:

1) I removed the enums out of the struct just in case gathering them all into a struct prevented the verifier from turning them to consts.
2) ``SCX_ENUM _INIT`` can be rolled into ``SCX_OPS_INIT`` if all the schedulers call ``SCX_ENUM_DEFINE()`` in their BPF program.

scheds/c/scx_central.bpf.c
```diff
@@ -57,13 +58,15 @@ enum {
...
 const volatile u32 nr_cpu_ids = 1;	/* !0 for veristat, set during init */
-const volatile u64 slice_ns = SCX_SLICE_DFL;
+const volatile u64 slice_ns;
 
 bool timer_pinned = true;
 u64 nr_total, nr_locals, nr_queued, nr_lost_pids;
 u64 nr_timers, nr_dispatches, nr_mismatches, nr_retries;
 u64 nr_overflows;
 
+SCX_ENUM_DEFINE();
```
htejun, line 68: `SCX_OPS_LOAD()` already assumes `UEI_DEFINE()`, so adding another dependency is just fine. We can add e.g. `SCX_COMMON_DEFINE()` to make it easier.

htejun (commented):
I left some minor comments but it's looking pretty good. Thanks. The userspace portion would need to be duplicated in `scx_utils` for rust scheds and it'd be nice to have a sanity check on BPF side so that we don't accidentally run with 0 for these enums.

hodgesds:
It might be worth hooking up the script to generate the enums through meson as well. If this is purely for rust schedulers then having the bpf_builder generate the enum could also be rather nice

scripts/gen_enums.py
```diff
@@ -0,0 +1,84 @@
...
+            ("scx_kick_flags", "SCX_KICK_PREEMPT"),
+            ("scx_kick_flags", "SCX_KICK_WAIT"),
+
+            ("scx_enq_flags", "SCX_ENQ_WAKEUP"),
+            ("scx_enq_flags", "SCX_ENQ_HEAD"),
+            ("scx_enq_flags", "SCX_ENQ_PREEMPT"),
+            ("scx_enq_flags", "SCX_ENQ_REENQ"),
+            ("scx_enq_flags", "SCX_ENQ_LAST"),
+            ("scx_enq_flags", "SCX_ENQ_CLEAR_OPSS"),
+            ("scx_enq_flags", "SCX_ENQ_DSQ_PRIQ"),
```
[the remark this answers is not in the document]
etsal: That would make maintaining the autogen'ed files significantly easier, but I think it would be better to do it in a separate diff on top because this one is getting on the large side

etsal:
> I left some minor comments but it's looking pretty good. Thanks. The userspace portion would need to be duplicated in `scx_utils` for rust scheds and it'd be nice to have a sanity check on BPF side so that we don't accidentally run with 0 for these enums.

For the userspace side, it turns out we only need a couple enums - mostly ```SCX_SLICE_*``` macros. We could autogenerate all enum definitions for Rust - it would be a matter of generating a lazy_static! macro for each one - but as I said in the other comment, I am not sure if it is worth it.

Regarding the sanity check, we could add one specifically for ```SCX_SLICE_*``` and embed it in SCX_OPS_OPEN(). For the more general case do we want to distinguish between an enum that is explicitly 0 and one that is uninitialized? One way is to assume that 0 is an error and fix this if we get a false positive in the future.

etsal:
Right now the Rust code requires including the new ```Enums``` struct with all the scx userspace enums, and the ```import_enums!``` macro used in ```scx_ops_open!``` to set the BPF-side enums. It would be nice if we could instead embed the set routine into the BPF skeleton so that scheduler author shouldn't have to ever interact with ```import_enums```.

While we could also embed the userspace enum struct in the BPF skeleton, I think it is useful for the schedulers themselves. Right now we manually define in ```compat.rs``` a variable for each enum we want to use in userspace, so I think it's nice to provide all possible enums with a single struct.

htejun (approved):
Looks great. Thanks!

PR #900, opened 2024-11-06
add CONFIG_IKCONFIG_PROC to ci kconfig

likewhatevs:
Add CONFIG_IKCONFIG_PROC to ci kconfig

This should fix:
```
09:50:59 [DEBUG] libbpf: object 'bpf_bpf': failed (-22) to create BPF token from '/sys/fs/bpf', skipping optional step...

09:50:59 [WARN] libbpf: failed to open system Kconfig
```
when
```
extern unsigned CONFIG_HZ __kconfig;
```

is present in a bpf program in a good manner.

By a good manner, I mean this should have libbpf use the guest's config in `/proc/` instead of the host's config (if present) via the cow mount of `/`.

If this is merged, I'll delete the latest cached kernel to trigger a rebuild with this config.


.github/workflows/sched-ext.config
```diff
@@ -54,3 +54,4 @@ CONFIG_UPROBE_EVENTS=y
 CONFIG_DEBUG_FS=y
 # more bpftrace to make that work
 CONFIG_IKHEADERS=y
+CONFIG_IKHEADERS_PROC=y
```
arighi, line 57: You mean `CONFIG_IKCONFIG_PROC` maybe?

likewhatevs: yes, thank you.

likewhatevs:
found a way to not need this atm (we probably do want to enable this sometime though, I think).

likewhatevs:
reopening this because this approach was more reliable than the alternative I was testing (so if this isn't problematic to enable, I think it would be pretty useful).

arighi (approved):

Issue #897, opened 2024-11-06
scx_lavd: cur_logical_clk update will be lost when concurrent updating 


archerbroler:
In advance_cur_logical_clk, when multiple CPU run lavd_running the same time, the cur_logical_clk will lost update.
Each CPU will read the same cur_logical_clk, and update with their own taskc->vdeadline_log_clk. Then  cur_logical_clk result will be the largest one, which make the cur_logical_clk's value not meaningful.
I think update to cur_logical_clk' should use somethin like this:
```
__sync_fetch_and_add(&cur_logical_clk, taskc->vdeadline_delta_ns);
```
Perhaps this is a reason to remove nr_queued as I mentioned in https://github.com/sched-ext/scx/issues/893

multics69:
That's a good finding. I should update it using CAS. Will fix it soon.

PR #903, opened 2024-11-07
scx_lavd: update cur_logical_clk atomically

multics69:
Previously, the cur_logical_clk is updated with WIRTE_ONCE(), which does not guarantee the atomicity when concurrent writes happen -- which is possible. So change it using CAS (compare-and-swap).

This addresses https://github.com/sched-ext/scx/issues/897


htejun (approved):

multics69:
Thanks @htejun for the review.

PR #889, opened 2024-11-04
[bug-fix] fixes #810

abrehman94:
* increasing the number of bits for cpumask

PR #904, opened 2024-11-07
scx_lavd: drop padding in cpdom_cpumask, which was a workaround

multics69:
The verifier error seems to stem from the wrong vmlinux.h. Also, PR #889 seems to completely fix the problem. So, drop the workaround.


multics69:
Thanks @multics69 for the review.

PR #908, opened 2024-11-07
layered/topo: lift layer specific checks out of per-LLC loop

JakeHillion:
The loops in topology aware mode were recently refactored to place the -per-LLC
loops inside the per-layer loops. However, the layer specific checks were left
in the inner loops, slowing this down unnecessarily.

Pull the layer specific checks from the inner loop into the outer loop.

Also changes these functions to `__weak` to ensure they don't get inlined -
they're expected to be verified as global functions.

Note to reviewers: this looks good to me, but I'd appreciate if you reviewed
the De Morgan applications in detail.

Test plan:
- `cargo build --release && sudo target/release/scx_layered --run-example` on a
  machine with multiple LLCs. It's possible to stall it quite easily with
  stress-ng but I believe this is the case on main.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1326,7 +1326,7 @@ void layered_dispatch_no_topo(s32 cpu, struct task_struct *prev)
 	scx_bpf_consume(LO_FALLBACK_DSQ);
 }
 
-int consume_preempting(struct cost *costc, u32 my_llc_id)
+__weak int consume_preempting(struct cost *costc, u32 my_llc_id)
 {
 	struct layer *layer;
 	u64 dsq_id;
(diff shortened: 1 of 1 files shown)


hodgesds:
This is a good start, I still see high X-LLC migrations on bergamo:
```
open_idle= 0.00 mig=94.94 xnuma_mig= 0.00 xllc_mig=64.02 affn_viol= 0.00
```

hodgesds (approved):

hodgesds:
@JakeHillion I'm also trying to track down an issue with the dsq cost accounting so if you run: `stress-ng -c $(nproc) -t 35` it may stall out if preemption is enabled. We probably need some better tests that cover paths with preemption enabled/disabled.

hodgesds:
> > This is a good start, I still see high X-LLC migrations on bergamo (same issue I saw in 909 as well):
> > ```
> > 
> > open_idle= 0.00 mig=94.94 xnuma_mig= 0.00 xllc_mig=64.02 affn_viol= 0.00
> > ```
> 
> Is this the right PR for that comment? This change should be a no-op except for potentially making the dispatch function a little faster in topology mode.

I believe so since it will be more likely to consume from a DSQ in the local LLC.

hodgesds:
> > > This is a good start, I still see high X-LLC migrations on bergamo (same issue I saw in 909 as well):
> > > ```
> > > 
> > > open_idle= 0.00 mig=94.94 xnuma_mig= 0.00 xllc_mig=64.02 affn_viol= 0.00
> > > ```
> > 
> > 
> > Is this the right PR for that comment? This change should be a no-op except for potentially making the dispatch function a little faster in topology mode.
> 
> I believe so since it will be more likely to consume from a DSQ in the local LLC.

I've misread this diff, it's not flipping the loops around so maybe after this lands I can try #909 again.

PR #909, opened 2024-11-07
scx_layered: Consume from local LLCs for dispatch

hodgesds:
When dispatching consume from DSQs in the local LLC first before trying remote DSQs. This should still be fair as the layer iteration order will be maintained.


hodgesds:
I think this is the same a #908

PR #911, opened 2024-11-07
scx_layered: Use layer idle_smt option

hodgesds:
When selecting the idle CPU use the idle_smt option on the layer. This may improve cache locality in some cases by placing tasks on CPUs that are on closer cache lines.


htejun (approved):

htejun:
I wonder whether when `!pref_idle_smt`, it should check whether the sibling is free first before going looking for other CPUs.

hodgesds:
> I wonder whether when `!pref_idle_smt`, it should check whether the sibling is free first before going looking for other CPUs.

Yeah, that makes sense.

etsal (approved):

PR #916, opened 2024-11-08
scx_layered: Work around older kernels choking on function calls from…

htejun:
… sleepable progs

Verifier in older kernels choke on function calls from sleepable progs triggering non-sensical RCU state error:

   frame1: R1_w=scalar(id=674,smin=smin32=0,smax=umax=smax32=umax32=51,var_off=(0x0; 0x3f)) R10=; return *llc_ptr;
  1072: (61) r0 = *(u32 *)(r2 +0)       ; frame1: R0_w=scalar(smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) R2_w=map_value(map=bpf_bpf.rodata,ks=4,vs=9570,off=4400,smin=smin32=0,smax=umax=smax32=umax32=204,var_off=(0x0; 0xfc)) refs=13,647
  ; }
  1073: (95) exit
  bpf_rcu_read_unlock is missing
  processed 10663 insns (limit 1000000) max_states_per_insn 8 total_states 615 peak_states 281 mark_read 20
  -- END PROG LOAD LOG --

Work around by adding and using __always_inline variant of cpu_to_llc_id() from layered_init(). Note that we can't switch everyone to __always_inline as that can lead to verification failure due to ins limit.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -104,7 +104,12 @@ static __noinline u64 layer_dsq_id(u32 layer_id, u32 llc_id)
 	return (layer_id * nr_llcs) + llc_id;
 }
 
-static __noinline u32 cpu_to_llc_id(s32 cpu_id)
+// XXX - cpu_to_llc_id() must not be inlined to not blow past ins limit when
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
I think in some of the back port patches those helpers were inlined. Maybe those got lost along the way.

PR #917, opened 2024-11-10
CFS -> EEVDF

next-ramses:
replaces references to CFS by EEVDF as is default scheduler since 6.6


htejun:
Can we use "fair" or "kernel scheduler" instead of CFS/EEVDF? "fair" would need an introduction - the fair scheduling class is the default kernel scheduler, so probably "kernel scheduler" is better.

PR #918, opened 2024-11-11
scx_layered: Add helper for layer slice duration

hodgesds:
Add a helper for returning the appropriate slice duration for a layer and replace a various instances where the slice value was being recalculated.


etsal (approved):
Looks good  to me, assuming that the instance of  ```slice_ns``` in main.bpf.c:1114 being shadowed by the new value is fine as you mention.

I would agree that it makes sense for the kthread to use the slice of the layer it's in. IMO if we were treating kthreads as a special case within the layer, then just passing them the standard  slice would not be enough - we'd need a specific separate policy we can tune. In that case, we might as well put them in their own layer.

PR #924, opened 2024-11-15
scheds: introduce scx_flash

arighi:
Introduce scx_flash (Fair Latency-Aware ScHeduler), a scheduler that focuses on ensuring fairness among tasks and performance predictability.

This scheduler is introduced as a replacement of the "lowlatency" mode in scx_bpfland, that has been dropped in commit 78101e4 ("scx_bpfland: drop lowlatency mode and the priority DSQ").

scx_flash operates based on an EDF (Earliest Deadline First) policy, where each task is assigned a latency weight. This weight is adjusted dynamically, influenced by the task's static weight and how often it releases the CPU before its full assigned time slice is used: tasks that release the CPU early receive a higher latency weight, granting them a higher priority over tasks that fully use their time slice.

The combination of dynamic latency weights and EDF scheduling ensures responsive and stable performance, even in overcommitted systems, making the scheduler particularly well-suited for latency-sensitive workloads, such as multimedia or real-time audio processing.


ptr1337:
Tested-by: Peter Jung <ptr1337@cachyos.org>

sirlucjan:
Tested-by: Piotr Gorski [piotrgorski@cachyos.org](mailto:piotrgorski@cachyos.org)

arighi:
> Can we use a different, a bit more specific, name? "fair" is used in commits / documentations to refer to cfs/eevdf, so it can get pretty confusing.

Ack, I'll come up with a different name, this concern has been also raised in the CachyOS discord. I'll keep this PR open and I'll adjust the subject and commit message once we figure out a better name. :)

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1068 @@
...
+		tctx->lock_boost = true;
+}
+
+static void task_unlock(struct task_struct *p)
+{
+	struct task_ctx *tctx;
+
+	tctx = try_lookup_task_ctx(p);
+	if (tctx)
+		tctx->lock_boost = false;
```
multics69, line 174: The current implementation does not seem to consider the case of holding multiple futexes.

multics69 (approved):

scheds/rust/scx_flash/src/bpf/main.bpf.c
multics69: Overall very nice work!

arighi: > The current implementation does not seem to consider the case of holding multiple futexes.

Correct, I'm just keeping things simple for now. And BTW, thanks for figuring out all these kprobes in lavd! :)

ptr1337 (approved):

sirlucjan (approved):

htejun (approved):

Issue #637, opened 2024-09-09
scx_layered: Add per-layer sched-delay metric (and maybe others too)


htejun:
As scheduler behavior metrics can vary widely across different layers, system level metrics aren't that useful in understanding how the scheduler is behaving. Add more per-layer metrics including per-layer scheduling delay metric.

minosfuture:
hi @htejun, I'm interested in working on this. I'm trying to get familiar with the repo by helping with good "first" issues.

I'm assuming there is no existing API that provides the scheduling delay metric for a task, so it needs to be implemented at per-scheduler basis? It'll be the accumulated delay between `enqueued` and `running`, if I understand correctly.

Is there an example implementation in other schedulers that I can refer to? If not, are we interested in adding such metric for other schedulers as well (or we will just rely on kernel's overall scheduling delay metric)?

htejun:
The information is useful for all schedulers but overall system metrics are easily observable with e.g. `btftrace` or `bcc` tools. How the metrics should be aggregated would depend on the specifc scheduler - e.g. `scx_layered` needs to collect the metrics per layer. `scx_bpfland` would probably want to aggregate depending on whether the task is classified interactive or not and so on. One altnerative approach could be coming up with a shared way of "tagging" tasks so that generic BPF tool can aggregate the numbers according to the tags.

We can start with baked-in impelmentation is each scheduler. I'd measure the durations whenever the task is runnable but not running - ie. `ops.runnable()` to `ops.running()` transition durations and `ops.stopping()` to the subsequent `ops.running()` transitions.

PR #929, opened 2024-11-17
scx_layered: Perf improvements and a bug fix

htejun:
- Fix layered treating all layers as preempting w/ `--local-llc-iterations`.
- `layered_select_cpu()` considered the waking CPU's llc/node as local instead of the wakee's previous CPU's. Fixed.
- Performance optimizations leading ~1.5% less cpu consumption while running saturating rd-hashd on ryzen 3900x.
- Stat output updates.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1515,7 +1515,7 @@ __weak int consume_preempting(struct cost *costc, u32 my_llc_id)
 					return -EINVAL;
 				}
 				layer = MEMBER_VPTR(layers, [layer_idx]);
-				if (has_budget(costc, layer) == 0)
+				if (!layer->preempt || has_budget(costc, layer) == 0)
 					continue;
 				dsq_id = layer_dsq_id(layer_idx, llc_id);
 				if (scx_bpf_consume(dsq_id))
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -226,7 +226,8 @@ static struct node_ctx *lookup_node_ctx(u32 node)
 {
 	struct node_ctx *nodec;
 
-	nodec = bpf_map_lookup_elem(&node_data, &node);
+	if (!(nodec = bpf_map_lookup_elem(&node_data, &node)))
+		scx_bpf_error("no node_ctx");
 	return nodec;
 }
 
@@ -242,7 +243,8 @@ static struct cache_ctx *lookup_cache_ctx(u32 cache_idx)
 {
 	struct cache_ctx *cachec;
 
-	cachec = bpf_map_lookup_elem(&cache_data, &cache_idx);
+	if (!(cachec = bpf_map_lookup_elem(&cache_data, &cache_idx)))
+		scx_bpf_error("no cache_ctx");
 	return cachec;
 }
 
@@ -413,6 +415,8 @@ struct task_ctx {
 	pid_t			last_waker;
 	bool			refresh_layer;
 	struct cached_cpus	layered_cpus;
+	struct cached_cpus	layered_cpus_llc;
+	struct cached_cpus	layered_cpus_node;
 	bool			all_cpus_allowed;
 	u64			runnable_at;
 	u64			running_at;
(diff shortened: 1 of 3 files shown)


htejun:
Not sure why neither layer weighs nor antistall could prevent the HI fallback stalls. @hodgesds @likewhatevs can you guys look into what's going on?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -325,7 +327,7 @@ static struct cpumask *lookup_layer_cpumask(int idx)
 	struct layer_cpumask_wrapper *cpumaskw;
 
 	if ((cpumaskw = bpf_map_lookup_elem(&layer_cpumasks, &idx))) {
-		return cast_mask(cpumaskw->cpumask);
+		return (struct cpumask *)cpumaskw->cpumask;
```
hodgesds, line 330: Went a little crazy trying to remove as many casts, thanks for cleaning up.

hodgesds:
> However, even if all layers are treated as preempting, both mechanisms should have been able to protect against HI fallback stalls, right?

Yeah, I think there is a bit more to debug. Not sure if on the antistall path it is checking all fallback DSQs as that seems to be where the stalls are localized.

hodgesds (approved):
Good to see all those temp cpumasks gone.

likewhatevs:
>  However, even if all layers are treated as preempting, both mechanisms should have been able to protect against HI fallback stalls, right?

I'm pretty certain antistall checks all DSQs and while IDK about weights, antistall wouldn't protect against stalls when all CPUs are saturated. 

When all CPUs are continually fully saturated, and affinitized tasks are present, antistall as it is now moves how long it takes for a stall to occur from ~30 seconds to 38 (or 45) seconds. 

I noticed while working on antistall that getting antistall to work when a system is continually fully saturated and slice size is big (which the default is, I think) required scaling slice size by some factor, but I removed that because proper timeline management would address that better.

I could add that slice scaling back in if we want antistall to be more effective until we have proper timeline management though?

Alternatively, I *think* (but didn't see) that increasing the frequency of the timer at which antistall runs (i.e. having it run every .1 second instead of every second) could have a similar effect if that is preferable?

The ahh sticky part IMO of either approach is that like, neither would be great for throughput I think (although in testing, I did not see the slice scaling approach hurt throughput much).

likewhatevs (approved):

likewhatevs:
> Not sure if on the antistall path it is checking all fallback DSQs as that seems to be where the stalls are localized.

I *think* this might be why that seems to be the case:

https://github.com/sched-ext/scx/blob/66223bf2350ec54d557ae9ef8b71a8c1c5d3d67e/scheds/rust/scx_layered/src/bpf/main.bpf.c#L2565-L2573

Antistall looks for the most delayed DSQ a CPU can consume and updates a map with that value. It does the per-DSQ delay comparison in jiffies, *but* it does cross-DSQ delay comparison at 1s resolution. IIUC, this means that if per-DSQ delays are within 1s of each other, because layer DSQs are walked before fallback DSQs, fallback DSQs would always be those to stall first (while non-fallback DSQs would be close behind).

I could swap those the order of those two (or keep all comparisons in jiffies), but IDK if that would help much (it wouldn't if delays are close).

This is all presuming there isn't some bug (or race condition) I'm missing, but IDK if there there is yet (i.e. everything I'm seeing atm kind of makes sense).

htejun:
> When all CPUs are continually fully saturated, and affinitized tasks are present, antistall as it is now moves how long it takes for a stall to occur from ~30 seconds to 38 (or 45) seconds.
> 
> I noticed while working on antistall that getting antistall to work when a system is continually fully saturated and slice size is big (which the default is, I think) required scaling slice size by some factor, but I removed that because proper timeline management would address that better.
> 
> I could add that slice scaling back in if we want antistall to be more effective until we have proper timeline management though?

Here, the machine is saturated but not to the point where the raw amount is naturally inducing long enough scheduling delays. The workload is self-regulating based on latency much like how e.g. a load balanced web server would behave. The machine is saturated but the magnitude of over saturation is continuously monitored and limited so that a task waking up still can get on a CPU in some milliseconds.

> Alternatively, I _think_ (but didn't see) that increasing the frequency of the timer at which antistall runs (i.e. having it run every .1 second instead of every second) could have a similar effect if that is preferable?
> 
> The ahh sticky part IMO of either approach is that like, neither would be great for throughput I think (although in testing, I did not see the slice scaling approach hurt throughput much).

I dont' think this is a frequency issue either. The per-layer DSQs are all being consumed with only a short delay while the fallback DSQs are hitting tens of seconds delays.

likewhatevs:
> Here, the machine is saturated but not to the point where the raw amount is naturally inducing long enough scheduling delays. 

> The per-layer DSQs are all being consumed with only a short delay while the fallback DSQs are hitting tens of seconds delays.

Ahh yeah something definitely sounds wrong, will debug this and implement #637 along the way to facilitate that.

PR #930, opened 2024-11-17
[RFC] scx_layered: Add netdev IRQ balancing

hodgesds:
Add support for netdev IRQ balancing on available CPUs . This adds a module to the `scx_utils` crate that parses network devices and is able to update the associated IRQ affinity masks. This idea is was inspired from conversations with @danobi.

After running layered with the flag enabled IRQs from watching `/proc/interrupts` are balanced across available CPUs from the scheduler's perspective. This can be observed where the large number of interrupts were routed to single cores.
```
            CPU0       CPU1       CPU2       CPU3       CPU4       CPU5       CPU6       CPU7       CPU8       CPU9       CPU10      CPU11
 168:          0          0          0          0          0          0          0          0          0          0          1          0  PCI-MSIX-0000:56:00.0    0-edge      eth1
 171:          0          0          0      63185          0          0          0          0          0          0          0    2651043  PCI-MSIX-0000:56:00.0    1-edge      eth1-TxRx-0
 173:    3314882          0          0          0          0      90830          0          0          0          0          0          0  PCI-MSIX-0000:56:00.0    2-edge      eth1-TxRx-1
 175:          0    3670760          0          0      84690          0          0          0          0          0          0          0  PCI-MSIX-0000:56:00.0    3-edge      eth1-TxRx-2
 176:          0          0    3721104          0          0          0          0          0          0          0          0          0  PCI-MSIX-0000:56:00.0    4-edge      eth1-TxRx-3
 ```
 
 This is a very crude implementation and does not take into account hinted affinity, which can be added later.

### DIFF
--- a/rust/scx_utils/src/netdev.rs
+++ b/rust/scx_utils/src/netdev.rs
@@ -13,13 +13,31 @@ use anyhow::Result;
 
 #[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct NetDev {
-    pub iface: String,
-    pub node: usize,
+    iface: String,
+    node: usize,
     pub irqs: BTreeMap<usize, Cpumask>,
-    pub irq_hints: BTreeMap<usize, Cpumask>,
+    irq_hints: BTreeMap<usize, Cpumask>,
 }
 
 impl NetDev {
+    pub fn iface(&self) -> &str {
+        &self.iface
+    }
+
+    pub fn node(&self) -> usize {
+        self.node
+    }
+
+    pub fn irq_hints(&self) -> &BTreeMap<usize, Cpumask> {
+        &self.irq_hints
+    }
+
+    pub fn update_irq_cpumask(&mut self, irq: usize, cpumask: Cpumask) {
+        if let Some(cur_cpumask) = self.irqs.get_mut(&irq) {
+            *cur_cpumask = cpumask;
+        }
+    }
+
     pub fn apply_cpumasks(&self) -> Result<()> {
         for (irq, cpumask) in self.irqs.iter() {
             let irq_path = format!("/proc/irq/{}/smp_affinity", irq);
--- a/rust/scx_utils/src/netdev.rs
+++ b/rust/scx_utils/src/netdev.rs
@@ -1,0 +1,83 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use std::collections::BTreeMap;
+use std::fs;
+use std::path::Path;
+
+use crate::misc::read_file_usize;
+use crate::Cpumask;
+use anyhow::Result;
+
+#[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct NetDev {
+    pub iface: String,
+    pub node: usize,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by hodgesds** [comment]
Visual representation with the following config:
```
[
  {
    "name": "powersave",
    "comment": "powersave",
    "matches": [[]],
    "kind": {
      "Confined": {
        "util_range": [
          0.4,
          0.8
        ],
        "slice_us": 800,
        "preempt": false,
        "growth_algo": "LittleBig",
        "idle_smt": true,
        "perf": 256,
        "preempt_first": false,
        "exclusive": false
      }
    }
  }
]
```
`stress-ng` runs on the little cores, while running a network speedtest sends IRQs towards the unallocated CPUs:

File: scheds/rust/scx_layered/src/main.rs [hunk 1]
```diff
@@ -471,6 +473,10 @@ struct Opts {
     #[clap(long, default_value = "false")]
     disable_antistall: bool,
 
+    /// Disable antistall
```
likewhatevs, line 476: comment goes up a few lines

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1542,6 +1556,26 @@ impl<'a> Scheduler<'a> {
...
+            return Ok(());
+        }
+
+        for (iface, netdev) in self.netdevs.iter_mut() {
+            for (irq, irqmask) in netdev.irqs.iter_mut() {
+                irqmask.clear();
+                for cpu in available_cpus.iter_ones() {
+                    let _ = irqmask.set_cpu(cpu);
+                }
+                trace!("{} updating irq {} cpusmask {:?}", iface, irq, irqmask);
```
likewhatevs, line 1571: typo

likewhatevs (approved):
pretty cool

htejun (approved):

htejun:
Not opposed to merging this as an experiment but I'm a bit doubtful that this would be useful as-is. It's likely that we will need more sophistication in determining how IRQs are assigned to CPUs. Maybe add a comment explaining that this is a proof-of-concept implementation?

rust/scx_utils/src/netdev.rs
```diff
@@ -0,0 +1,83 @@
...
+use crate::misc::read_file_usize;
+use crate::Cpumask;
+use anyhow::Result;
+
+#[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct NetDev {
+    pub iface: String,
+    pub node: usize,
+    pub irqs: BTreeMap<usize, Cpumask>,
+    pub irq_hints: BTreeMap<usize, Cpumask>,
```
danobi, line 19: Maybe worthwhile to tighten up data model. For example, `irq_hints` is exported from driver to userspace, so this should probably be a read-only getter that returns a const reference to private member (and not available as a publicly modifiable member). Same goes for `iface` and `node`.

`irqs` looks like it's doing double duty here - `read_netdevs()` uses it to return initial config. Then callers will modify the member and trigger a write thru `apply_cpumasks`. Perhaps better to make `irqs` a private member and then have a getter/setter pair, where getter returns whatever is cached and setter is a write thru cache. Setter could just be `apply_cpumasks()` accepting a cpumask.

danobi (commented):
Echoing what Tejun said - good to get this infra in but will probably need more algorithm work to realize wins

PR #933, opened 2024-11-18
scx_layered: Fix verifier issues on older kernels

hodgesds:
On 6.9 kernels the verifier is not able to track `struct bpf_cpumasks` properly on nested structs. Move the cpumasks from the `cached_cpus` struct back to the `task_ctx` struct so older versions of the verifier can pass.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -405,7 +405,6 @@ int BPF_PROG(sched_tick_fentry)
 struct cached_cpus {
 	s64			id;
 	u64			seq;
-	struct bpf_cpumask __kptr *mask;
 };
 
 struct task_ctx {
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1564,7 +1578,8 @@ static __noinline int consume_non_open(struct cost *costc, s32 cpu, u32 my_llc_i
 					return -EINVAL;
 				}
 				layer = MEMBER_VPTR(layers, [layer_idx]);
-				if (has_budget(costc, layer) == 0)
+				if (layer->kind == LAYER_KIND_OPEN ||
```
etsal, line 1581: Is this related to the main change of the diff?

etsal (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: I can pull this into a separate diff, good call.

htejun (approved):

PR #934, opened 2024-11-18
scx_layered: Cleanups around topology handling

htejun:
- Fix a couple bugs around layer -> DSQ ID mapping.
- Use fields cached in `cpu_ctx` instead of recalculating them repeatedly.
- Supply matching topology information when `--disable-topology` is specified so that it matches how the scheduler is going to behave. Reduce code branches.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -112,6 +112,7 @@ struct cpu_ctx {
 	u64			gstats[NR_GSTATS];
 	u64			lstats[MAX_LAYERS][NR_LSTATS];
 	u64			ran_current_for;
+	u64			hi_fallback_dsq_id;
 	u32			layer_idx;
 	u32			cache_idx;
 	u32			node_idx;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -115,7 +115,10 @@ u32 rotate_llc_id(u32 base_llc_id, u32 rotation)
 // return the dsq id for the layer based on the LLC id.
 static __noinline u64 layer_dsq_id(u32 layer_id, u32 llc_id)
 {
-	return (layer_id * nr_llcs) + llc_id;
+	if (nr_llcs == 1)
+		return layer_id;
+	else
+		return (layer_id * nr_llcs) + llc_id;
 }
 
 // XXX - older kernels get confused by RCU state when subprogs are called from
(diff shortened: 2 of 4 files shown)


scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -112,6 +112,7 @@ struct cpu_ctx {
 	u64			gstats[NR_GSTATS];
 	u64			lstats[MAX_LAYERS][NR_LSTATS];
 	u64			ran_current_for;
+	u64			hi_fallback_dsq_id;
```
hodgesds, line 115: This makes a lot of sense, probably should have added that awhile ago

hodgesds (approved):
LGTM, good to see so many `if (disable_topology)` branches die.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1725,8 +1671,6 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 		return;
 	}
 
-	u32 my_llc_id = cpu_to_llc_id(cpu);
```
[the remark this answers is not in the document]
htejun: Here, I think cpu and cctx are pointing to the same cpu. It's actually confusing that we're doing `lookup_cpu_ctx(-1)` instead of `lookup_cpu_ctx(cpu)`. I'll comment on the part which was buggy.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2638,23 +2575,21 @@ u64 antistall_set(u64 dsq_id, u64 jiffies_now)
...
 	u64 jiffies_now;
 
 	if (!enable_antistall)
 		return true;
 
 	jiffies_now = bpf_jiffies64();
 
-	bpf_for(dsq_id, 0, nr_layers) {
-		antistall_set(dsq_id, jiffies_now);
-	}
```
htejun, line 2652: This loop was buggy as it wouldn't visit all layer DSQs when topology is enabled.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1247,21 +1204,11 @@ static bool keep_running(struct cpu_ctx *cctx, struct task_struct *p)
...
-			if (!scx_bpf_dsq_nr_queued(layer->idx)) {
-				p->scx.slice = slice_ns;
-				lstat_inc(LSTAT_KEEP, layer, cctx);
-				return true;
-			}
-		} else {
-			u32 dsq_id = cpu_to_llc_id(tctx->last_cpu >= 0 ?
-						   tctx->last_cpu :
-						   bpf_get_smp_processor_id());
-			if (!scx_bpf_dsq_nr_queued(dsq_id)) {
```
htejun, line 1260: And this was buggy. It stores `llc_id` into `dsq_id` and uses it as DSQ ID.

likewhatevs (approved):

PR #935, opened 2024-11-18
scx_layered: Add workaround for 6.9 verifier

hodgesds:
For some 6.9 kernels the verifier is not able to handle cpumasks properly. Disable a path in the antistall logic on old kernels.

I don't know if this is a good idea, but it does unblock testing on some older kernels.


etsal (approved):
Pretty nice workaround imo, especially if there is no clear way of getting the code to run in older kernels.

hodgesds:
Eh, I think we can use the CLI flag to disable antistall though...

likewhatevs:
This effectively disables antistall which while less problematic than verifier errors is still kinda problematic...

PR #936, opened 2024-11-19
ci: Add benchmark github action

hodgesds:


.github/workflows/caching-build.yml
```diff
@@ -171,6 +171,27 @@ jobs:
       # no symlink following here (to avoid cycle`s)
       - run: sudo find '/home/runner/' -iname '*.ci.log' -exec mv {} ./log_save/ \;
         if: always()
+      # Download previous benchmark result from cache (if exists)
+      - name: Download previous benchmark data
```
likewhatevs, line 175: You're probably gonna want to use artifacts for this and not cache (we use cache alot, data's gonna get rotated out).

scripts/parse_stress_results.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+yaml_file="$1"
```
likewhatevs, line 2: I could be wrong, but merging these all into a single json (and merging new results w/ old) is probably gonna be enough of a pain you wanna do it in python tbh.

hodgesds: I think maybe we should just make it output in the criterion output format, then when we add Rust specific benchmarks it can all use the same output.

PR #939, opened 2024-11-19
scx_layered: Work around verification failure in antistall_set() on o…

htejun:
…ld kernels

In earlier kernels, the iterator variable wasn't trusted making the verifier choke on calling kfuncs on its dereferences. Work around by re-looking up the task by PID.


likewhatevs (approved):
LGTM

PR #940, opened 2024-11-19
verifier workarounds

likewhatevs:
make scx code work on older kernel code in the for-next branch of the sched_ext kernel tree.


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2243,6 +2244,20 @@ s32 BPF_STRUCT_OPS(layered_init_task, struct task_struct *p,
...
 		return -ENOMEM;
 	}
+	const struct cpumask * all_mask = cast_mask(all_cpumask);
+	if (!all_mask)
+		return -EINVAL;
+
+	if (!(cpumask = bpf_cpumask_create()))
+		return -ENOMEM;
+
+	bpf_cpumask_and(cpumask, all_mask, p->cpus_ptr);
```
htejun, line 2254: Yeah, this would work too but you'd also want to update the mask on `ops.set_cpumask()`.

PR #943, opened 2024-11-19
replace goto with unrolled loop in antistall_set

JakeHillion:


etsal (approved):

likewhatevs:
this looks good (passes tests) just ahh, first entry of `#pragma` into a bpf.c file in this repo, hesitant of anything new wrt/ verifier etc.

etsal:
My 2c: Since the pragma just copy-pastes the loop a bunch of times before starting compilation the verifier would never see the difference compared to what we are doing right now. According to the eBPF docs, this pragma was also the only way of doing loops early on so it's been used with eBPF code before.

JakeHillion:
> My 2c: Since the pragma just copy-pastes the loop a bunch of times before starting compilation the verifier would never see the difference compared to what we are doing right now. According to the eBPF docs, this pragma was also the only way of doing loops early on so it's been used with eBPF code before.

Yep! It's one of the prescribed ways to do loops in the ebpf docs: https://docs.ebpf.io/linux/concepts/loops/

This code seems quite challenging to verify, and my theory is that not having to track the state of the loop variable makes it fractionally easier, even if the end result should be the same. We could remove the unroll in future but I don't see much benefit either way.

hodgesds (approved):
LGTM, we used to have to pragma all the loops before `bpf_for` was a thing.

likewhatevs:
> This code seems quite challenging to verify, and my theory is that not having to track the state of the loop variable makes it fractionally easier

Makes sense, nice!

PR #948, opened 2024-11-20
arena-based radix tree allocator

etsal:
Add an arena-based memory allocator to the repository. The allocator uses a radix tree to index the allocated data. During an allocation the code carves out new allocated data instances out of arena pages and attaches them to the chunk. As an optimization, freed data instances are kept in their location in the chunk to be reused on a subsequent allocation.


scheds/include/arch/mips/vmlinux-v6.12-rc2-g5b7c893ed5ed.h
hodgesds, line 1: Is this deletion intentional? I guess I'm not sure if we test on mips architecture.

scheds/include/scx/bpf_arena_common.h
```diff
@@ -0,0 +1,116 @@
...
+
+#if defined(__BPF_FEATURE_ADDR_SPACE_CAST) && !defined(BPF_ARENA_FORCE_ASM)
+#define __arena __attribute__((address_space(1)))
+#define __arena_global __attribute__((address_space(1)))
+#define cast_kern(ptr) /* nop for bpf prog. emitted by LLVM */
+#define cast_user(ptr) /* nop for bpf prog. emitted by LLVM */
+#else
+
+/* emit instruction:
+ * rX = rX .off = BPF_ADDR_SPACE_CAST .imm32 = (dst_as << 16) | src_as
```
hodgesds, line 43: Some more comments here would be helpful.

scheds/include/scx/bpf_arena_htab.h
```diff
@@ -0,0 +1,100 @@
...
+	list_for_each_entry(l, head, hash_node)
+		if (l->hash == hash && l->key == key)
+			return l;
+
+	return NULL;
+}
+
+static int htab_hash(int key)
+{
+	return key;
```
hodgesds, line 53: will this do something different in the future?

scheds/include/arch/mips/vmlinux-v6.12-rc2-g5b7c893ed5ed.h
etsal: No this is an error, this was  rolled into the merge at the base of this pull request, and is a workaround for that lavd verification error. I will bring the header back in.

scheds/include/scx/bpf_arena_common.h
etsal: I imported this from the arena BPF selftests without modifying anything, I will add some comments to make clear why we need the macros from here and what they do.

scheds/include/scx/sdt_task_impl.bpf.h
```diff
@@ -0,0 +1,482 @@
...
+
+	chunk = desc->chunk;
+	cast_kern(chunk);
+
+	/* populate leaf node if necessary */
+
+	pos = idx & (SDT_TASK_ENTS_PER_CHUNK - 1);
+	data = chunk->data[pos];
+	if (!data) {
+		bpf_spin_unlock(&sdt_task_lock);
```
hodgesds, line 458: I guess `sdt_task_alloc_from_pool` needs the lock as well. I think that should be fine.

scheds/include/scx/sdt_task_impl.bpf.h
```diff
@@ -0,0 +1,482 @@
...
+	mval = bpf_task_storage_get(&sdt_task_map, p, 0, 0);
+	if (!mval)
+		return;
+
+	sdt_task_free_idx(mval->tid.idx);
+	mval->data = NULL;
+}
+
+
+static SDT_TASK_FN_ATTRS int sdt_task_find_empty(struct sdt_task_desc __arena *desc, struct sdt_task_desc * __arena *descp, __u64 *idxp)
```
hodgesds, line 349: nit: could be formatted nicer

scheds/c/scx_sdt.c
```diff
@@ -0,0 +1,104 @@
+/* SPDX-License-Identifier: GPL-2.0 */
```
hodgesds, line 1: you can put your name on your code 😃

scheds/include/scx/sdt_task_impl.bpf.h
etsal: ```sdt_task_alloc_from_pool``` uses the ```sdt_pool_alloc_lock```, so we cannot keep the task lock even if we use the fast path in the pool allocator since IIRC we cannot hold multiple spinlocks. The other reason is that on the slow path we have to use the ```bpf_arena_alloc_pages``` kfunc so we cannot be holding any locks.

scheds/include/scx/sdt_task_impl.bpf.h
```diff
@@ -392,6 +398,24 @@ static SDT_TASK_FN_ATTRS void sdt_task_free_idx(__u64 idx)
...
+
+	sdt_arena_verify();
+
+	mval = bpf_task_storage_get(&sdt_task_map, p, 0, 0);
+	if (!mval)
+		return NULL;
+
+	data = mval->data;
+
+	return (void __arena *)data->payload;
```
hodgesds, line 415: At some point was there a `cast_arena` helper? If not, it may be useful like the `cast_mask` for bpf  cpumasks.

etsal: I don't think there was one, would it be used to ```const```ify the arena pointer? Or would it be like cast_{kern, user}() and transform a pointer back to an arena pointer?

scheds/include/scx/sdt_task_impl.bpf.h
```diff
@@ -0,0 +1,579 @@
...
+
+static SDT_TASK_FN_ATTRS
+int sdt_set_idx_state(struct sdt_task_desc __arena *desc, __u64 pos, bool state)
+{
+	__u64 __arena *allocated = desc->allocated;
+	__u64 bit;
+
+	cast_kern(allocated);
+
+	if (pos >= SDT_TASK_ENTS_PER_CHUNK) {
```
htejun, line 292: `unlikely()`?

scheds/include/scx/sdt_task_impl.bpf.h
```diff
@@ -0,0 +1,579 @@
...
+
+		chunk = desc->chunk;
+		cast_kern(chunk);
+
+		desc_children = (struct sdt_task_desc * __arena *)chunk->descs;
+		desc = desc_children[pos];
+
+		/* Someone else is populating the subtree. */
+		if (desc == (void *)SDT_TASK_ALLOC_RESERVE)
+			return -EAGAIN;
```
htejun, line 476: Wouldn't this make everyone else who's racing busy loop on the same slot over and over again? The allocation can take a substantial amount of time if it involves reclaim.

htejun (commented):
Left some comments. Also, you can squash the commit history as you see fit. We probably don't want to commit my WIP commits without any explanations into history.

etsal:
New version of the patch that addresses the feedback. 

- Most important change is that we preallocate any arena pages we possibly need on the alloc path, so at no point do we need to drop the spinlock during allocation. This in turn means we do not need to account for the race condition edge cases that were not being handled gracefully in the previous version, because they are now impossible. We can preallocate because there is a bounded number of arena page allocations we can make during a single descriptor allocation call (2 * number of levels + 1 for the data itself) .

- Allocations now always succeed. Allocations require that we have enough arena pages and free descriptors in the allocator. We ensure both at the very beginning of the operation, so we remove now unnecessary cleanup. This included descriptor/chunk frees, which only happened during error handling. Pools are now alloc only - no need to free back to the pool.

- Since pools are now alloc only, we can remove linked lists from the allocator completely. Pool allocation calls stack allocate out of the preallocated arena pages, and avoid the additional complexity of the linked lists.

scheds/include/scx/sdt_task_impl.bpf.h
etsal: I have added unlikely() on all the sanity checks that should never fail under normal operation.

scheds/include/scx/sdt_task_impl.bpf.h
```diff
@@ -0,0 +1,579 @@
...
+ * to access the arena and help the verifier.
+ */
+static bool sdt_verify_once;
+
+static SDT_TASK_FN_ATTRS void sdt_arena_verify(void)
+{
+	if (sdt_verify_once)
+		return;
+
+	bpf_printk("%s: arena pointer %p", __func__, &arena);
```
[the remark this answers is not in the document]
etsal: Yeah this is more of a band-aid, we can remove it if the verifier is able to automatically associate an arena with all programs in a single load object.

etsal:
@htejun Also wrt the commit history: I've kept it as-is for the review process to make it easier to track which code got added where and when. When we are ready to merge I will squash it into two diffs for brevity,  one with your initial prototype and the final one.

htejun (approved):

etsal:
Squashed history and fixed corruption-inducing bug.

PR #950, opened 2024-11-20
scx_layered: Use timer for cpumask updates

hodgesds:
Use a timer (100ms) for cpumask updates rather than relying on scheduler ticks.

Mostly idle system (PR):
image
Mostly idle system (main):
image


htejun:
Maybe mention that this is a perf optimization. I wonder 100ms might be too long. Depending on the timing, cpumask update can be delayed by almost a full interval. It should be possible to do this synchronously with PROG_RUN, right?

hodgesds:
> It should be possible to do this synchronously with PROG_RUN, right?

My only concern would be with a bad scheduler config that delays scheduler userspace from running.

hodgesds:
I tested with once every millisecond and it didn't seem to have any overhead:
image

I didn't see any overhead from `bpftop`, which I wonder if it able to trace bpf timers.

htejun:
> > It should be possible to do this synchronously with PROG_RUN, right?
> 
> My only concern would be with a bad scheduler config that delays scheduler userspace from running.

Yeah, but then there's no cpumask update to apply anyway.

hodgesds:
> > My only concern would be with a bad scheduler config that delays scheduler userspace from running.
> 
> Yeah, but then there's no cpumask update to apply anyway.

Oh yeah... that makes sense then

htejun (approved):

hodgesds:
PROG_RUN works really well, basically no overhead unless layers are growing/shrinking:
image

PR #951, opened 2024-11-21
layered: update for old verifier compatibility

JakeHillion:
Update layered to increase kernel backwards compatibility.

Change the check in `calc_preferred_cost` to `>MAX_LLCS` rather than `>nr_llcs`.
Both are constant at verification time but the old verifiers are happier with
the macro constant. Note that neither of these should be necessary because of
the inlined `rotate_layer_id` ending with a mod.

Add a `MEMBER_VPTR` to finish making this happy.

Make `initialize_budgets` weak. This prevents it being inlined and makes it
verify as a global function. It's only called from init so this is fine, and
it makes reading verifier errors significantly simpler than when this very
large function gets inlined.

Test plan:
- Tested on a 6.4.3 verifier kernel and a 6.9 verifier kernel. CI will test on
  6.12+.
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/sched-ext/scx/pull/951).
* __->__ #951
* #988

### DIFF
--- a/.github/actions/install-deps-action/action.yml
+++ b/.github/actions/install-deps-action/action.yml
@@ -15,6 +15,13 @@ runs:
 
     ### DOWNLOAD AND INSTALL DEPENDENCIES ###
 
+    # Get some packages from nixpkgs for newer revisions
+    - uses: DeterminateSystems/nix-installer-action@main
+    - run: |
+        nix registry add nixpkgs github:NixOS/nixpkgs/8edf06bea5bcbee082df1b7369ff973b91618b8d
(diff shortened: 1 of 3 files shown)


htejun (approved):

JakeHillion (commented):
Update: pinned the clang version here and rebased to test against a version of clang-17 that's relevant to me. Turns out this works with that version of clang-17 but not with clang 18.1.8 here. Discussing with the BPF clang maintainers to see whether this is a bug or a feature.

JakeHillion:
Closing after discussion. A bisect confirms this is fixed in later revisions of clang, but for distro packaging we need to support 18. clang-18 is closed for fixes, and this specific verifier fix doesn't work with clang-18.

PR #958, opened 2024-11-22
scx_layered: Reimplement CPU allocation and some other fixes

htejun:
The existing load calculation was incorrect. It could enter feedback loops where a layer which is using more CPU cycles keeps getting more CPUs and vice-versa regardless of their weights. Also, as the growth and shrinking decisions are made locally in each layer, they couldn't account for global competition state which is necessary to consider for weight based distribution.

Reimplement CPU allocation so that CPUs are distributed globally and layers are converged to the target state.

While at it, make a few misc changes including printing stats in config order and correctly accounting for layer CPU time.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2120,11 +2120,6 @@ void BPF_STRUCT_OPS(layered_stopping, struct task_struct *p, bool runnable)
 		return;
 
 	used = bpf_ktime_get_ns() - tctx->running_at;
-	if (used < layer->min_exec_ns) {
-		lstat_inc(LSTAT_MIN_EXEC, layer, cctx);
-		lstat_add(LSTAT_MIN_EXEC_NS, layer, cctx, layer->min_exec_ns - used);
-		used = layer->min_exec_ns;
-	}
 
 	// If the task ran on the hi fallback dsq then the cost should be
 	// charged to it.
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/main.rs
```diff
@@ -443,13 +442,15 @@ struct Opts {
     #[clap(short = 'e', long)]
     example: Option<String>,
 
-    /// Disables preemption if the weighted load fraction of a layer (load_frac_adj) exceeds the
-    /// threshold. The default is disabled (0.0).
+    /// ***DEPRECATED*** Disables preemption if the weighted load fraction
```
hodgesds, line 445: These might be able to just be deleted, I don't think they are in use anywhere but I guess backwards compatibility is good. Maybe next release we can clean them out.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2104,11 +2104,6 @@ void BPF_STRUCT_OPS(layered_stopping, struct task_struct *p, bool runnable)
 		return;
 
 	used = bpf_ktime_get_ns() - tctx->running_at;
-	if (used < layer->min_exec_ns) {
```
hodgesds, line 2107: Ahh yeah, this makes sense now for how the accounting is kind of messed up.

scheds/rust/scx_layered/src/main.rs
likewhatevs: > Maybe next release we can clean them out.

Idk. As a rule I want to say the cycle is:
1) Deprecate
2) Deploy
3) Delete

So if there are binaries running deployed where (1) has not happened yet, can't move to (3).

PR #964, opened 2024-11-22
scx_layered: select_cpu() fixes and updates

htejun:
- Both `select_cpu[_no_topo]()` were broken. Each missing different idle CPU search steps. This is likely cause of some of the regressions seen with new versions. Fixed.
- There are no meaningful differences between `no_topo` and regular versions after dead code elimination. Drop the duplicate.
- Short-circuit `select_cpu()` when layers' CPUs are saturated. This saves ~0.6% cputime when running saturating `rd-hashd` on Ryzen 3900x.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -621,68 +621,10 @@ bool should_try_preempt_first(s32 cand, struct layer *layer,
 }
 
 static __always_inline
-s32 pick_idle_no_topo(struct task_struct *p, s32 prev_cpu,
-		      struct cpu_ctx *cctx, struct task_ctx *tctx,
-		      struct layer *layer, bool from_selcpu)
-{
-	const struct cpumask *idle_smtmask;
-	struct cpumask *layer_cpumask, *layered_cpumask;
-	s32 cpu;
-
-	/* look up cpumasks */
-	if (!(layered_cpumask = (struct cpumask *)tctx->layered_mask) ||
-	    !(layer_cpumask = lookup_layer_cpumask(tctx->layer)))
-			return -1;
-
-	/* not much to do if bound to a single CPU */
-	if (p->nr_cpus_allowed == 1 && scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
-		if (layer->kind == LAYER_KIND_CONFINED &&
-		    !bpf_cpumask_test_cpu(prev_cpu, layer_cpumask))
-			lstat_inc(LSTAT_AFFN_VIOL, layer, cctx);
-		return prev_cpu;
-	}
-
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -703,15 +650,37 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
-	if (from_selcpu && should_try_preempt_first(prev_cpu, layer, cpumask)) {
+	if (from_selcpu && should_try_preempt_first(prev_cpu, layer, layered_cpumask)) {
 		cctx->try_preempt_first = true;
 		return -1;
 	}
 
-	if (!(prev_cctx = lookup_cpu_ctx(prev_cpu)) ||
-	    !(idle_smtmask = scx_bpf_get_idle_smtmask()))
+	/*
+	 * When a layer stays saturated, there's no point in repeatedly
```
hodgesds, line 661: Nice!

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -703,15 +650,37 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
+
+		scx_bpf_put_idle_cpumask(cpumask);
+		if (!has_idle)
+			return -1;
+	}
+
+	if ((nr_llcs > 1 || nr_nodes > 1) &&
+	    !(prev_cctx = lookup_cpu_ctx(prev_cpu)))
+		return -1;
+	if (!(idle_smtmask = scx_bpf_get_idle_smtmask()))
```
hodgesds, line 683: If a layer sets [`idle_smt`](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/intf.h#L196) should this set `idle_smtmask` to `scx_bpf_get_idle_cpumask`? I think that's what it used to do, which is a little hacky of overriding `idle_smtmask`. Otherwise it may skip idle smt cpus in [`pick_idle_cpu_from`](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/main.bpf.c#L593).

hodgesds (approved):
LGTM

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: The mask isn't used at all by `pick_idle_cpu_from()` when `pref_idle_smt` is not set. I don't know why the mask itself needs to be overridden too.

hodgesds: > I don't know why the mask itself needs to be overridden too.

Guh... I'm blind, I see it's using `cand`. I've always gotten confused by that function. Sorry for the noise!

PR #965, opened 2024-11-22
scx_layered: Refactor layer growth order

hodgesds:
Refactor layer growth order to be more sticky to cores by using the core growth order to shrink layers. Move layer core growth order creation to a separate helper that has context on all layers. This will be used in the future for more topology aware core growth.

Layers have a tendency to "walk" across cores when growing/shrinking which can lead to suboptimal layouts over time. Consider the following simple configuration for a machine with Big/Little cores:
```
[
  {
    "name": "powersave",
    "comment": "powersave",
    "matches": [[]],
    "kind": {
      "Confined": {
        "util_range": [
          0.4,
          0.8
        ],
        "slice_us": 800,
        "preempt": false,
        "growth_algo": "LittleBig",
        "idle_smt": true,
        "perf": 256,
        "preempt_first": false,
        "exclusive": false
      }
    }
  }
]
```
When load is placed on the layer the growth should happen towards the lower cores on the system:
image
After load is removed off the system the layer should compact towards it's preferred set of cores:
image

In contrast, on `main` after several iterations load ends up on Big cores:
image

When layers shrink they do not consider the core growth order and end up compacting the `LittleBig` layer completely onto Big cores:
image

### DIFF
--- a/scheds/rust/scx_layered/src/layer_core_growth.rs
+++ b/scheds/rust/scx_layered/src/layer_core_growth.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::sync::Arc;
 
 use clap::Parser;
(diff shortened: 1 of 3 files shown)


hodgesds:
After rebasing this doesn't seem to work correctly, will investigate more.

PR #967, opened 2024-11-23
topology: Make it more consistent

htejun:
- Drop trivial accessors.
- Use `Arc` to build skip indices consistently at each level replacing `Topology.cores`, `Topology.cpus` and some lower level ad-hoc helpers.
- Make top level also use `BTreeMap` instead of `Vec`.

No functional changes intended.


rust/scx_utils/src/topology.rs
```diff
@@ -110,116 +111,41 @@ pub enum CoreType {
...
+    pub base_freq: usize,
+    pub trans_lat_ns: usize,
+    pub l2_id: usize,
+    pub l3_id: usize,
+    pub llc_id: usize,
     pub core_type: CoreType,
 }
 
-impl Cpu {
-    /// Get the ID of this Cpu
```
arighi, line 125: Maybe we should move these comments to the attributes instead of dropping them (rephrasing them a bit, like "ID of this CPU"). Most of these descriptions are trivial but some can be useful to keep.

rust/scx_utils/src/topology.rs
```diff
@@ -110,116 +111,41 @@ pub enum CoreType {
...
-
-    /// Get the maximum scaling frequency of this CPU
-    pub fn max_freq(&self) -> usize {
-        self.max_freq
-    }
-
-    /// Get the base operational frequency of this CPU
-    ///
-    /// This is only available on Intel Turbo Boost CPUs, if not available this will simply return
-    /// maximum frequency.
```
arighi, line 143: This one for example would be nice to keep it.

rust/scx_utils/src/topology.rs
```diff
@@ -110,116 +111,41 @@ pub enum CoreType {
...
-        self.id
-    }
-
-    /// Get the map of CPUs inside this Core
-    pub fn cpus(&self) -> &BTreeMap<usize, Cpu> {
-        &self.cpus
-    }
-
-    /// Get a Cpumask of all SMT siblings in this Core
-    pub fn span(&self) -> &Cpumask {
```
arighi, line 191: Ditto.

rust/scx_utils/src/topology.rs
```diff
@@ -110,116 +111,41 @@ pub enum CoreType {
...
-    /// Get the map of cores inside this LLC
-    pub fn cores(&self) -> &BTreeMap<usize, Core> {
-        &self.cores
-    }
+pub struct Llc {
+    pub id: usize,
+    pub cores: BTreeMap<usize, Arc<Core>>,
+    pub span: Cpumask,
 
-    /// Get a Cpumask of all CPUs in this LLC
```
arighi, line 214: Ditto.

arighi (approved):

multics69 (approved):
Looks good to me. Thanks!

PR #970, opened 2024-11-24
scx_layered: Reimplement `try_preempt()`

htejun:
`try_preempt()` is rather complex for doing something which isn't all that dynamic. Instead of doing the tiered trial dynamically, pre-compute the proximity order for each CPU and just try preemption following that order. This maintains or improves the behavior while significantly simplifying the code. It's no longer necessary to separate out `no_topo` variant afterwards. Drop it.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2671,6 +2671,33 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
 			return ret;
 	}
 
+	dbg("CFG: Dumping CPU proxity map");
+	bpf_for(i, 0, nr_possible_cpus) {
+		struct cpu_ctx *cpuc;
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2671,6 +2529,33 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
 			return ret;
 	}
 
+	dbg("CFG: Dumping CPU proxity map");
```
hodgesds, line 2532: Could maybe kick this out to a helper function and initialize the prox map above, but should be good for now as long as the verifier is happy.

hodgesds (approved):
Nice, I'm guessing there's a bit of performance gain from saving all those cpumask operations.

htejun:
I *think* so, it's difficult to measure as preemptions tend to be more chaotic but the main benefit is being able to unify the code paths to ease future changes.

PR #973, opened 2024-11-24
scx_layered: More cleanups and an antistall fix

htejun:
- Remove `disable_toplogy` special case from DSQ init path.
- Rename `cpu_ctx->layer_cycles` to `cpu_ctx->layer_usages`.
- Rename `idx` to `id` and other name related cleanups.
- `layer_id` type and uninit state cleanups.
- Fix `antistall_consume()` skipping incorrectly.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -84,13 +84,13 @@ static inline s32 sibling_cpu(s32 cpu)
 		return -1;
 }
 
-static struct layer *lookup_layer(int idx)
+static struct layer *lookup_layer(u32 id)
 {
-	if (idx < 0 || idx >= nr_layers) {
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1159,10 +1156,7 @@ bool antistall_consume(s32 cpu, struct cpu_ctx *cctx)
 	consumed = false;
 	zero = 0;
 
-	if (!enable_antistall)
-		return false;
-
-	if (!cctx || !cctx->layer_idx || !cpu)
```
hodgesds, line 1165: nice catch, so this means even if the antistall cpu doesn't have a layer it can still try to consume stalled out tasks.

hodgesds (approved):
Finally there is no more confusion between id and idx.

PR #975, opened 2024-11-24
scx_lavd: add tunables for adjusting time slices

multics69:
Add two command line options, --slice-max-us and --slice-min-us, to adjust task's maximum and minimum time slice values, respectively.


hodgesds (approved):
This makes sense.

multics69:
Thanks @hodgesds for the review!

PR #976, opened 2024-11-24
scx_layered: Refactor layered initialization

hodgesds:
Refactor layer initialization into a separate helper function.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2452,12 +2452,135 @@ struct layered_timer layered_timers[MAX_TIMERS] = {
 // TODO: separate this out to a separate compilation unit
 #include "timer.bpf.c"
 
+/*
+ * Initializes per-layer specific data structures.
+ */
+static s32 init_layer(int layer_id, u64 *fallback_dsq_id)
+{
+	struct bpf_cpumask *cpumask;
+	struct layer_cpumask_wrapper *cpumaskw;
+	struct layer *layer = &layers[layer_id];
+	int j, k, ret;
+
+	dbg("CFG LAYER[%d][%s] min_exec_ns=%lu open=%d preempt=%d exclusive=%d",
+	    layer_id, layer->name, layer->min_exec_ns,
+	    layer->kind != LAYER_KIND_CONFINED,
+	    layer->preempt, layer->exclusive);
+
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2559,126 +2680,11 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
...
-		 */
-		bpf_cpumask_setall(cpumask);
-
-		cpumask = bpf_kptr_xchg(&cpumaskw->cpumask, cpumask);
-		if (cpumask)
-			bpf_cpumask_release(cpumask);
-
-		// create the dsqs for the layer
-		bpf_for(j, 0, nr_llcs) {
-			int node_id = llc_node_id(i);
```
hodgesds, line 2673: This looks like a really bad bug.... the llc node id was being looked up based on the layer id and not the LLC.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2558,127 +2678,13 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
 	dbg("CFG: Dumping configuration, nr_online_cpus=%d smt_enabled=%d little_cores=%d",
 	    nr_online_cpus, smt_enabled, has_little_cores);
 
+	u64 fallback_dsq_id = 0;
```
likewhatevs, line 2681: nit: shouldn't that be some FALLBACK_DSQ_ID defined somewhere?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2452,12 +2452,132 @@ struct layered_timer layered_timers[MAX_TIMERS] = {
 // TODO: separate this out to a separate compilation unit
 #include "timer.bpf.c"
 
+static s32 init_layer(int layer_id, u64 *fallback_dsq_id)
```
likewhatevs, line 2455: nit: a nice well formatted comment here would be nice

likewhatevs (approved):
LGTM I think

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: So this is just the counter that is used to sequentially allocate the hi fallback DSQs.

PR #977, opened 2024-11-25
scx_layered: Refactor CPU initialization

hodgesds:
Refactor CPU initialization into a separate helper function and clean up the use of zero values with a common const.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -46,6 +46,7 @@ const volatile u32 layer_iteration_order[MAX_LAYERS];
 const volatile bool enable_antistall = true;
 /* Delay permitted, in seconds, before antistall activates */
 const volatile u64 antistall_sec = 3;
+const u32 zero = 0;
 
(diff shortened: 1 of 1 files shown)


likewhatevs (approved):
LGTM

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2576,15 +2571,66 @@ static s32 init_layer(int layer_id, u64 *fallback_dsq_id)
...
+		*init_antistall_dsq = SCX_DSQ_INVALID;
+	}
+
+	if (!(cctx = lookup_cpu_ctx(cpu))) {
+		return -ENOMEM;
+	}
+	cctx->task_layer_id = MAX_LAYERS;
+
+	if ((u8_ptr = MEMBER_VPTR(all_cpus, [cpu / 8]))) {
+		if (*u8_ptr & (1 << (cpu % 8))) {
```
likewhatevs, line 2598: nit: this was here before so, it's right, but like, that feels like the kind of thing that should get it's own helper because of ease of getting it wrong.

PR #978, opened 2024-11-25
scx_layered: Fix missing layered->id init

htejun:
While refactoring layer init path, 0edbc05e6da1 ("scx_layered: Refactor layered initialization") dropped layer->id init which breaks a lot of stuff. Fix it. While at it, replace j, k iterators with i, j in init_layer().


hodgesds (approved):
LGTM, feels like some of that should have got caught with tests. Will think about that some more.

PR #980, opened 2024-11-25
scx_layered: Track owned/open execution times and per-LLC-layer stats

htejun:
- Track owned/open execution times per layer and base layer growth decision on owned time only.
- Track per-LLC-layer number of scheduling events and running avg of queue latencies and report them. Will be used to modulate open executions.
- Fix the problem where a zeor sized layer couldn't grow due to core quantization in CPU free/alloc paths.
- Code cleanups.
- Bail early with a clear error message if node, LLC or core IDs are not consecutive. This should be fixed in the future.

### DIFF
--- a/scheds/rust/scx_layered/src/stats.rs
+++ b/scheds/rust/scx_layered/src/stats.rs
@@ -54,12 +54,10 @@ pub struct LayerStats {
     pub index: usize,
     #[stat(desc = "Total CPU utilization (100% means one full CPU)")]
     pub util: f64,
+    #[stat(desc = "Open CPU utilization %")]
+    pub util_open: f64,
     #[stat(desc = "fraction of total CPU utilization")]
     pub util_frac: f64,
-    #[stat(desc = "Open CPU utilization (100% means one full CPU)")]
-    pub open_util: f64,
-    #[stat(desc = "fraction of open CPU utilization")]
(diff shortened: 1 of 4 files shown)


likewhatevs:
current err msg is:
```
thread 'main' panicked at scheds/rust/scx_layered/src/main.rs:798:82:
```

code is (not sure why I can't link it):
```
let vec_sub = |l: &[u64], r: &[u64]| l.iter().zip(r.iter()).map(|(l, r)| *l - *r).collect();
```

`(*l).wrapping_sub_unsigned(*r)` should make that rust error go away I think, but that wrapping calls be needed on all manipulations like this.

scheds/rust/scx_layered/src/stats.rs
```diff
@@ -127,6 +130,10 @@ pub struct LayerStats {
     pub max_nr_cpus: u32,
     #[stat(desc = "slice duration config")]
     pub slice_us: u64,
+    #[stat(desc = "Per-LLC scheduling event fractions")]
+    pub llc_fracs: Vec<f64>,
+    #[stat(desc = "Per-LLC average latency")]
```
hodgesds, line 135: could be a little more descriptive

hodgesds (approved):
LGTM

PR #981, opened 2024-11-25
topology: handle LLCs spanning multiple NUMA nodes

arighi:
On physical hardware the same LLC should never span across multiple NUMA nodes. However, this scenario can happen in virtualized environments, particularly when multiple NUMA nodes are defined without specifying an exact CPU topology.

Example:

 $ vng --cpu 8 -m 4G --numa 2G,cpus=0-3 --numa 2G,cpus=4-8
 ...
 $ lscpu -e
 CPU NODE SOCKET CORE L1d:L1i:L2:L3 ONLINE
   0    0      0    0 0:0:0:0          yes
   1    0      0    1 1:1:1:0          yes
   2    0      0    2 2:2:2:0          yes
   3    0      0    3 3:3:3:0          yes
   4    1      0    4 4:4:4:0          yes
   5    1      0    5 5:5:5:0          yes
   6    1      0    6 6:6:6:0          yes
   7    1      0    7 7:7:7:0          yes

In this case, an error like the following is reported, causing scx schedulers relying on topology to fail:

 > sudo scx_bpfland
 ...
 thread 'main' panicked at scheds/rust/scx_bpfland/src/main.rs:261:36:
 called `Result::unwrap()` on an `Err` value: Duplicate LLC ID 0

We can handle this particular configuration by introducing an artificial LLC for each NUMA node. In the example above, an additional LLC (ID 1) is created to cover CPU4..CPU7.

Result with this fix applied:

 > sudo scx_bpfland
 ...
 08:09:29 [INFO] L2 cache ID 4: sibling CPUs: [4]
 08:09:29 [INFO] L2 cache ID 0: sibling CPUs: [0]
 08:09:29 [INFO] L2 cache ID 6: sibling CPUs: [6]
 08:09:29 [INFO] L2 cache ID 1: sibling CPUs: [1]
 08:09:29 [INFO] L2 cache ID 5: sibling CPUs: [5]
 08:09:29 [INFO] L2 cache ID 3: sibling CPUs: [3]
 08:09:29 [INFO] L2 cache ID 2: sibling CPUs: [2]
 08:09:29 [INFO] L2 cache ID 7: sibling CPUs: [7]
 08:09:29 [INFO] L3 cache ID 0: sibling CPUs: [0, 1, 2, 3]
 08:09:29 [INFO] L3 cache ID 1: sibling CPUs: [4, 5, 6, 7]

Fixes: e3b623d ("topology: Build uniform skip indicies")

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -196,16 +196,27 @@ impl Topology {
             let mut node_cores = BTreeMap::new();
             let mut node_cpus = BTreeMap::new();
 
-            for (&llc_id, llc) in node.llcs.iter_mut() {
+            for (&(mut llc_id), llc) in node.llcs.iter_mut() {
                 let llc_mut = Arc::get_mut(llc).unwrap();
                 let mut llc_cpus = BTreeMap::new();
 
+                // The same LLC ID should not span across multiple NUMA nodes. If it does, create a
+                // new artificial LLC for each node. While this should never happen on physical
+                // hardware, it can occur in virtualized environments.
+                if topo_llcs.contains_key(&llc_id) {
+                    llc_id += 1;
+                }
+
                 for (&core_id, core) in llc_mut.cores.iter_mut() {
-                    for (&cpu_id, cpu) in core.cpus.iter() {
+                    let core_mut = Arc::get_mut(core).unwrap();
+                    for (&cpu_id, cpu) in core_mut.cpus.iter_mut() {
+                        let cpu_mut = Arc::get_mut(cpu).unwrap();
+                        cpu_mut.llc_id = llc_id;
+                        cpu_mut.l3_id = llc_id;
                         if topo_cpus
-                            .insert(cpu_id, cpu.clone())
-                            .or(node_cpus.insert(cpu_id, cpu.clone()))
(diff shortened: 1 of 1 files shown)


rust/scx_utils/src/topology.rs
```diff
@@ -196,16 +196,27 @@ impl Topology {
...
-            for (&llc_id, llc) in node.llcs.iter_mut() {
+            for (&(mut llc_id), llc) in node.llcs.iter_mut() {
                 let llc_mut = Arc::get_mut(llc).unwrap();
                 let mut llc_cpus = BTreeMap::new();
 
+                // The same LLC ID should not span across multiple NUMA nodes. If it does, create a
+                // new artificial LLC for each node. While this should never happen on physical
+                // hardware, it can occur in virtualized environments.
+                if topo_llcs.contains_key(&llc_id) {
+                    llc_id += 1;
```
htejun, line 207: Would it be easier to understand if this is done in a separate loop before this one? Also, it probably should scan forward and allocate the first unallocated ID to avoid shifting all subsequent IDs and generate a warning message.

htejun:
Maybe this can be solved together in #982?

arighi:
Closing this one for now, I'll try to address this issue using a different approach.

PR #982, opened 2024-11-25
scx_utils: Add core_id/llc_id to topology as a unique identifiers

hodgesds:
On systems with multiple NUMA nodes core_ids can be reused. Add monotonically increasing fields `core_id`/`llc_id` to uniquely identify cores/llcs. This does not implement the problem in #981 however.

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -126,6 +126,7 @@ pub struct Cpu {
     pub core_id: usize,
     pub llc_id: usize,
     pub node_id: usize,
+    pub hw_id: usize,
 }
 
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
@@ -139,6 +140,8 @@ pub struct Core {
     /// Ancestor IDs.
     pub llc_id: usize,
     pub node_id: usize,
+    /// A monotonically increasing id for each core that is unique.
+    pub hw_id: usize,
 }
 
 #[derive(Debug, Clone)]
@@ -200,7 +203,7 @@ impl Topology {
                 let llc_mut = Arc::get_mut(llc).unwrap();
                 let mut llc_cpus = BTreeMap::new();
 
-                for (&core_id, core) in llc_mut.cores.iter_mut() {
+                for (&hw_id, core) in llc_mut.cores.iter_mut() {
                     for (&cpu_id, cpu) in core.cpus.iter() {
                         if topo_cpus
                             .insert(cpu_id, cpu.clone())
@@ -213,11 +216,11 @@ impl Topology {
                     }
 
                     if topo_cores
-                        .insert(core_id, core.clone())
-                        .or(node_cores.insert(core_id, core.clone()))
+                        .insert(hw_id, core.clone())
+                        .or(node_cores.insert(hw_id, core.clone()))
                         .is_some()
                     {
-                        bail!("Duplicate CORE ID {}", core_id);
+                        bail!("Duplicate hardware ID {}", hw_id);
                     }
                 }
 
@@ -246,21 +249,23 @@ impl Topology {
     /// Build a complete host Topology
     pub fn new() -> Result<Topology> {
         let span = cpus_online()?;
+        let mut hw_id: usize = 0;
         // If the kernel is compiled with CONFIG_NUMA, then build a topology
         // from the NUMA hierarchy in sysfs. Otherwise, just make a single
         // default node of ID 0 which contains all cores.
         let nodes = if Path::new("/sys/devices/system/node").exists() {
-            create_numa_nodes(&span)?
+            create_numa_nodes(&span, &mut hw_id)?
         } else {
-            create_default_node(&span, false)?
+            create_default_node(&span, &mut hw_id, false)?
         };
 
         Self::instantiate(span, nodes)
     }
 
     pub fn with_flattened_llc_node() -> Result<Topology> {
+        let mut hw_id = 0;
         let span = cpus_online()?;
-        let nodes = create_default_node(&span, true)?;
+        let nodes = create_default_node(&span, &mut hw_id, true)?;
         Self::instantiate(span, nodes)
     }
 
@@ -329,9 +334,9 @@ impl Topology {
 /// let topo = Topology::new().unwrap();
 /// let topo_map = TopologyMap::new(&topo).unwrap();
 ///
-/// for (core_id, core) in topo_map.iter().enumerate() {
+/// for (hw_id, core) in topo_map.iter().enumerate() {
 ///     for cpu in core {
-///         println!("core={} cpu={}", core_id, cpu);
+///         println!("core={} cpu={}", hw_id, cpu);
 ///     }
 /// }
 /// ```
@@ -365,12 +370,12 @@ impl TopologyMap {
     /// corresponding `BitVec` represents whether a logical core belongs to that physical core.
     pub fn core_cpus_bitvec(&self) -> Vec<BitVec> {
         let mut core_cpus = Vec::<BitVec>::new();
-        for (core_id, core) in self.iter().enumerate() {
-            if core_cpus.len() < core_id + 1 {
-                core_cpus.resize(core_id + 1, bitvec![0; *NR_CPUS_POSSIBLE]);
+        for (hw_id, core) in self.iter().enumerate() {
+            if core_cpus.len() < hw_id + 1 {
+                core_cpus.resize(hw_id + 1, bitvec![0; *NR_CPUS_POSSIBLE]);
             }
             for cpu in core {
-                core_cpus[core_id].set(*cpu, true);
+                core_cpus[hw_id].set(*cpu, true);
             }
         }
         core_cpus
@@ -381,12 +386,12 @@ impl TopologyMap {
     /// represents whether the physical core id of the logical core.
     pub fn cpu_core_mapping(&self) -> Vec<usize> {
         let mut cpu_core_mapping = Vec::new();
-        for (core_id, core) in self.iter().enumerate() {
+        for (hw_id, core) in self.iter().enumerate() {
             for cpu in core {
                 if cpu_core_mapping.len() < cpu + 1 {
                     cpu_core_mapping.resize(cpu + 1, 0);
                 }
-                cpu_core_mapping[*cpu] = core_id;
+                cpu_core_mapping[*cpu] = hw_id;
             }
         }
         cpu_core_mapping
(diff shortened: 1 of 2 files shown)


hodgesds:
This may be related to the issues #981 is trying to resolve.

hodgesds:
Example from a multi socket machine:
```
$ lscpu | grep 'Model name'
Model name:                           Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz
$ cat /sys/devices/system/cpu/cpu*/topology/core_id  | sort -n | uniq -c
      4 0
      4 1
      4 2
      4 3
      4 4
      4 8
      4 9
      4 10
      4 11
      4 12
      4 16
      4 17
      4 18
      4 19
      4 20
      4 24
      4 25
      4 26
      4 27
      4 28
$ cat /sys/devices/system/cpu/cpu*/topology/cluster_id  | sort -n | uniq -c
      2 0
      2 2
      2 4
      2 6
      2 8
      2 16
      2 18
      2 20
      2 22
      2 24
      2 32
      2 34
      2 36
      2 38
      2 40
      2 48
      2 50
      2 52
      2 54
      2 56
      2 64
      2 66
      2 68
      2 70
      2 72
      2 80
      2 82
      2 84
      2 86
      2 88
      2 96
      2 98
      2 100
      2 102
      2 104
      2 112
      2 114
      2 116
      2 118
      2 120
$ cat /sys/devices/system/cpu/cpu*/topology/die_id  | sort -n | uniq -c
     40 0
     40 1
```

rust/scx_utils/src/topology.rs
```diff
@@ -439,7 +442,16 @@ fn create_insert_cpu(
...
     // Physical core ID
     let top_path = cpu_path.join("topology");
-    let core_id = read_file_usize(&top_path.join("core_id"))?;
+    let core_id_raw = read_file_usize(&top_path.join("core_id"))?;
+    let cluster_id = read_file_usize(&top_path.join("cluster_id")).unwrap_or(0);
+    let die_id = read_file_usize(&top_path.join("die_id")).unwrap_or(0);
+
+    // core_id can be reused across nodes. replace core_id with a hash of topo ids
+    let mut hasher = DefaultHasher::new();
+    core_id_raw.hash(&mut hasher);
```
etsal, line 450: Can we just hash a (core_id_raw, cluster_id, die_id) tuple instead of passing the hasher through each usize var? Other than that this change looks strictly better than what we currently have so LGTM.

etsal (approved):

rust/scx_utils/src/topology.rs
htejun: I'm not sure about hashing. For, CPU and node IDs, there are direct kernel interfaces that consume them and we probably should keep the same IDs but for LLC and core, we can just sequentially assign IDs and then add another field - e.g. hw_id - in case hardware IDs are necessary.

For node and CPU, we should provide iteration interfaces.

hodgesds:
I think this should be good to go now.

rust/scx_utils/src/topology.rs
```diff
@@ -126,6 +126,7 @@ pub struct Cpu {
     pub core_id: usize,
     pub llc_id: usize,
     pub node_id: usize,
+    pub hw_id: usize,
```
htejun, line 129: Why is `cpu.hw_id` necessary?

rust/scx_utils/src/topology.rs
```diff
@@ -139,6 +140,8 @@ pub struct Core {
     /// Ancestor IDs.
     pub llc_id: usize,
     pub node_id: usize,
+    /// A monotonically increasing id for each CPU that is unique.
+    pub hw_id: usize,
```
htejun, line 144: Let's make `core.id` monotonically increasing and `core.hw_id` the kernel provided one. Maybe `core.kerrnel_id` is a better name.

hodgesds: Sounds good.

rust/scx_utils/src/topology.rs
hodgesds: In layered it uses the Cpu struct to lookup the core cpumask

hodgesds:
Update with sequential LLC ids:
```
$ lstopo-no-graphics | grep L3
    L3 L#0 (28MB)
    L3 L#1 (28MB)
$ cargo build --release && sudo ./target/release/scx_layered f:/home/hodgesd/user2.json -v 2>&1 | grep -i llc
...
21:41:51 [DEBUG] configuring node 0, LLCs 1
21:41:51 [DEBUG] configuring llc 0 for node 0
21:41:51 [DEBUG] configuring node 1, LLCs 1
21:41:51 [DEBUG] configuring llc 1 for node 1
```

rust/scx_utils/src/topology.rs
```diff
@@ -124,6 +125,7 @@ pub struct Cpu {
 
     /// Ancestor IDs.
     pub core_id: usize,
+    pub kernel_id: usize,
```
htejun, line 128: Hmm... why do we need `kernel_id` for cpu?

rust/scx_utils/src/topology.rs
```diff
@@ -139,6 +141,8 @@ pub struct Core {
     /// Ancestor IDs.
     pub llc_id: usize,
     pub node_id: usize,
+    /// The sysfs value of core_id
```
htejun, line 144: I'd put this right below `id` and explain why there are two separate IDs.

rust/scx_utils/src/topology.rs
```diff
@@ -153,12 +157,15 @@ pub struct Llc {
 
     /// Skip indices to access lower level members easily.
     pub all_cpus: BTreeMap<usize, Arc<Cpu>>,
+    /// The kernel id of the llc
+    pub kernel_id: usize,
```
htejun, line 161: Ditto.

rust/scx_utils/src/topology.rs
```diff
@@ -460,13 +474,26 @@ fn create_insert_cpu(
...
+        *llc_id += 1;
+    }
+
+    let inc_llc = if node.llc_kernel_ids.len() == 0 || node.llc_kernel_ids.contains(&llc_kernel_id)
+    {
+        false
+    } else {
+        true
+    };
+    node.llc_kernel_ids.insert(llc_kernel_id);
```
htejun, line 487: Is this correct? Let's say there are four cpus 0 1 2 3. Kernel says 0 2 are on llc 0 and 1 3 or llc 4. We want to assign llc 0 to cpu 0 2 and llc 1 to 1 3, right? Wouldn't the above code do that?

rust/scx_utils/src/topology.rs
```diff
@@ -483,14 +510,21 @@ fn create_insert_cpu(
         None => CoreType::Big { turbo: false },
     };
 
-    let core = llc_mut.cores.entry(core_id).or_insert(Arc::new(Core {
-        id: core_id,
+    let inc_core = if llc_mut.cores.contains_key(core_id) || !flatten_llc {
+        true
+    } else {
+        false
+    };
```
htejun, line 517: Ditto with cores.

So, for LLCs, there are two problems to solve:
1. LLC IDs may not be consecutive.
2. A single LLC ID may straddle across multiple nodes incorrectly in some virt setups.

The second one probably should be solved by reassigning a different kernel_id to the llc. Let's ignore that for a bit. For the first issue, we should be building kernel ID -> consecutive ID mapping. I don't see how keeping using the ssame ID when the same kernel ID was seen would assign the correct consecutive IDs.

For core IDs, the problem is that they may repeat across node boundaries, so this is a bit different from the LLC situation - ie. we should be assigning each node_id:core_kernel_id pair a unique consecutive ID.

hodgesds: > I don't see how keeping using the ssame ID when the same kernel ID was seen would assign the correct consecutive IDs.

It only increments when the mapping hasn't been seen before, but I guess on repeats it would be incorrect. I was trying to avoid having to create that extra mapping, but it looks like it's required. I might leave private to the structs though.

htejun: We don't have to keep it around either and can just use temporary context struct containing necessary mappings and cursors during init.

hodgesds:
Seems to be working now on multi socket/multi LLC machines.

rust/scx_utils/src/topology.rs
htejun: Let's move this up right below `.id`.

rust/scx_utils/src/topology.rs
```diff
@@ -460,13 +494,41 @@ fn create_insert_cpu(
...
 
-    let llc = node.llcs.entry(llc_id).or_insert(Arc::new(Llc {
-        id: llc_id,
+    let multi_numa = topo_ctx.node_llc_kernel_ids.len() > 0;
+    let node_llc_kernel_ids: &mut BTreeMap<usize, usize> =
+        if topo_ctx.node_llc_kernel_ids.contains_key(&node.id) {
+            topo_ctx
+                .node_llc_kernel_ids
+                .get_mut(&node.id)
+                .context("impossible")?
```
htejun, line 503: If impossible, just unwrap?

rust/scx_utils/src/topology.rs
```diff
@@ -460,13 +494,41 @@ fn create_insert_cpu(
...
+        };
+    if !node_llc_kernel_ids.contains_key(&llc_kernel_id) {
+        if node_llc_kernel_ids.len() > 0 || multi_numa {
+            topo_ctx.llc_id += 1;
+        }
+        node_llc_kernel_ids.insert(llc_kernel_id, topo_ctx.llc_id);
+    }
```
htejun, line 519: Wouldn't it be easier to make it `BTreeMap<(node_id, llc_kernel_id), llc_id>`? Then there is only one map and the new `llc_id` would just be `map.len()`.

rust/scx_utils/src/topology.rs
```diff
@@ -483,14 +545,41 @@ fn create_insert_cpu(
...
+                .node_core_kernel_ids
+                .get_mut(&node.id)
+                .context("impossible")?
+        };
+    if !node_core_kernel_ids.contains_key(&core_kernel_id) {
+        if node_core_kernel_ids.len() > 0 || multi_numa {
+            topo_ctx.core_id += 1;
+        }
+        node_core_kernel_ids.insert(core_kernel_id, topo_ctx.core_id);
+    }
```
htejun, line 569: Ditto for core IDs.

rust/scx_utils/src/topology.rs
hodgesds: I think with this way it gets included in the rust doc though. Or did you mean the location and not the comment... sorry I misinterpreted...

rust/scx_utils/src/topology.rs
hodgesds: Yeah, that's more reasonable 👍

rust/scx_utils/src/topology.rs
htejun: Oh yeah, the location of the field as `.id` and `.kernel_id` are closely related, I think it'd be better to colocate them.

htejun (approved):

Issue #937, opened 2024-11-19
scx_loader: Does not restart the scheduler upon scheduler stall


1Naim:
Usually when running an scx scheduler with the scx.service systemd service, it will try to restart itself after a stall. However, when running an scx scheduler using `scx_loader`, upon a stall, scx_loader will keep itself running but the stalled scheduler isn't restarted, leaving an "empty" loader.

cc @vnepogodin

PR #983, opened 2024-11-25
scx_loader: restart scheduler upon fail

vnepogodin:
fixes https://github.com/sched-ext/scx/issues/937

### DIFF
--- a/rust/scx_loader/src/main.rs
+++ b/rust/scx_loader/src/main.rs
@@ -480,6 +480,56 @@ async fn start_scheduler(
     args: Vec<String>,
     child_id: Arc<AtomicU32>,
 ) -> Result<()> {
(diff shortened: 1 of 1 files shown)


sirlucjan (approved):

rust/scx_loader/src/main.rs
```diff
@@ -480,6 +480,56 @@ async fn start_scheduler(
     args: Vec<String>,
     child_id: Arc<AtomicU32>,
 ) -> Result<()> {
+    // Ensure the child process exit is handled correctly in the runtime
+    tokio::spawn(async move {
```
hodgesds, line 484: I would think you want some sort of backoff/delay after consecutive failures.

Issue #966, opened 2024-11-22
Introduce `for_each_possible_cpu()` `for_each_online_cpu()` iterators


htejun:
Some schedulers still assume that all possible CPUs are consecutively allocated which is not true on some machines. sched_ext exposes `scx_bpf_nr_cpu_ids()`, `scx_bpf_get_possible_cpumask()` and `scx_bpf_get_online_cpumask()`. Provide helpers to cache the result and macros to iterate over them.

abrehman94:
@htejun could you please review the implementation? 

Example use 
```
static inline void example_use() {
    int *cpu;

    for_each_possible_cpu(cpu){
        bpf_printk("Updated CPU %d is possible", *cpu);
    }

    for_each_online_cpu(cpu){
        bpf_printk("CPU %d is online", *cpu);
    }
}
```

PR #985, opened 2024-11-26
Introduce for_each_possible_cpu() for_each_online_cpu() iterators

abrehman94:
fixes: #966

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -338,6 +338,66 @@ u32 bpf_cpumask_any_and_distribute(const struct cpumask *src1,
                                    const struct cpumask *src2) __ksym;
 u32 bpf_cpumask_weight(const struct cpumask *cpumask) __ksym;
 
+int bpf_iter_bits_new(struct bpf_iter_bits *it, const u64 *unsafe_ptr__ign,
+                      u32 nr_words) __ksym;
+int *bpf_iter_bits_next(struct bpf_iter_bits *it) __ksym;
+void bpf_iter_bits_destroy(struct bpf_iter_bits *it) __ksym;
+
+#define def_iter_struct(name)                                                  \
+  struct bpf_iter_##name {                                                     \
+    struct bpf_iter_bits it;                                                   \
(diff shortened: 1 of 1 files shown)


likewhatevs:
I think this needs to be used in a scheduler to know if it passes the verifier.

Comments on these would be nice for hover docs but perhaps not needed because self explanitory.

abrehman94:
@likewhatevs I have added the doc comment. 

@likewhatevs, @htejun I tried the following code in bpfland. 

```
static inline void resetbuf(char *buf, int len, int cpucount) {
    int i;
    for (i = 0; i < len; i++) {
        if (i < cpucount) {
          buf[i] = ' ';
        }else{
          buf[i] = 0;
        }
    }
}

/*
 * Print out the online and possible CPU map using bpf_printk() as a
 * demonstration of using the cpumask kfuncs and ops.cpu_on/offline().
 */
static void print_cpus(void)
{
    int *cpu;
    int cpucount = 0;
    char buf[128] = "", *p;

    for_each_possible_cpu(cpu){
        if ((p = MEMBER_VPTR(buf, [*cpu]))){
            *p = 'X';
        }
        cpucount = *cpu;
    }
    bpf_printk("Possible CPUs: %s", buf);

    resetbuf( buf, sizeof(buf), cpucount + 1 );
    for_each_online_cpu(cpu){
        if ((p = MEMBER_VPTR(buf, [*cpu]))){
            *p = 'O';
        }
    }
    bpf_printk("Online: %s", buf);
}
```

After make cpu 45 offline on my machine, I get the following output. 
```
     scx_bpfland-2045639 [033] ...11 3474039.329624: bpf_trace_printk: Possible CPUs: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
     scx_bpfland-2045639 [033] ...11 3474039.329628: bpf_trace_printk: Online: OOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO OO
```

I wasn't sure if should make this code part of the PR. Should I?

PR #986, opened 2024-11-26
scx_utils: introduce topology-aware IDs

arighi:
On systems with multiple NUMA nodes, physical cores or caches on different nodes can share the same ID, as IDs might be reused across nodes.

To properly handle this, introduce the concept of topology IDs, that are composed of two parts:
 - lower half bits represent the local ID (e.g., core or cache ID) within a NUMA node,
 - higher half bits represent the NUMA node ID.

Moreover, provide helper functions to construct a topology ID and extract the local ID and node ID components from a given topology ID:
 - topology_id(node_id: usize, id: usize) -> usize
 - topology_id_local(topo_id: usize) -> usize
 - topology_id_node(topo_id: usize) -> usize

This should address all the issues addressed in #981 and #982.

### DIFF
--- a/rust/scx_utils/src/lib.rs
+++ b/rust/scx_utils/src/lib.rs
@@ -61,6 +61,9 @@ pub use libbpf_logger::init_libbpf_logging;
 pub mod ravg;
 
 mod topology;
+pub use topology::topology_id;
+pub use topology::topology_id_local;
+pub use topology::topology_id_node;
 pub use topology::Core;
 pub use topology::CoreType;
 pub use topology::Cpu;
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -420,6 +420,46 @@ fn cpus_online() -> Result<Cpumask> {
     Ok(mask)
 }
 
+// This constant defines the bit shift used to create unique topology IDs.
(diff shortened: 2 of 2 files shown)


hodgesds:
This still might be problematic on some systems as `core_id` can be reused within a NUMA node:
```
$ ls  /sys/devices/system/cpu/cpu{0,40} | grep node
firmware_node
node0
firmware_node
node0
$ cat  /sys/devices/system/cpu/cpu{0,40}/topology/core_id
0
0
```

arighi:
> This still might be problematic on some systems as `core_id` can be reused within a NUMA node:
> 
> ```
> $ ls  /sys/devices/system/cpu/cpu{0,40} | grep node
> firmware_node
> node0
> firmware_node
> node0
> $ cat  /sys/devices/system/cpu/cpu{0,40}/topology/core_id
> 0
> 0
> ```

Ah... what a pain, I thought the overlap was only among different NUMA nodes. Then we still need the logic to increments the IDs.

hodgesds:
Yeah, let me reattempt to get that working properly.

arighi:
But is it correct? Are cpu0 and cpu40 part of the same SMT core?

arighi:
This seems to be fixed by #982 so I'm closing it for now. Thanks @hodgesds!

PR #992, opened 2024-11-27
fix document generation

likewhatevs:
fix document generation by disabling document generation for all features.

looks like a bug in the tooling around optional features and the all-features flag.


hodgesds (approved):
LGTM

PR #996, opened 2024-11-27
scx_layered: Implement in-layer execution protection to replace cost based fairness

htejun:
cost based fairness mechanism tries to implement weight-based fairness by accounting the consumption of each layer and changing preemption and dispatch behaviors accordingly on each scheduling event. While this approach can work, it's unnecessarily finely grained given that `scx_layered` already has control over which layers get how many CPUs.

Previous changes made CPU allocations to follow the layer weights under system saturation. This PR implement in-layer execution protection so that layers can reliably grow and fully occupy its CPUs unaffected by preempt layers. Weight-based fairness is enforced through the combination of CPU allocation and in-layer execution protection - a layer is guaranteed to have its fair share number of CPUs according to its weight under contention and that it can use all its CPUs.

While there are a couple missing pieces, the basic mechanism seems to work as expected. Drop cost based fairness mechanism. This reduces code complexity and `.stopping()` overhead and makes future changes easier.


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2738,10 +2674,12 @@ SCX_OPS_DEFINE(layered,
...
 	       .set_weight		= (void *)layered_set_weight,
 	       .set_cpumask		= (void *)layered_set_cpumask,
+	       .update_idle		= (void *)layered_update_idle,
 	       .cpu_release		= (void *)layered_cpu_release,
 	       .init_task		= (void *)layered_init_task,
 	       .exit_task		= (void *)layered_exit_task,
 	       .dump			= (void *)layered_dump,
 	       .init			= (void *)layered_init,
 	       .exit			= (void *)layered_exit,
+	       .flags			= SCX_OPS_KEEP_BUILTIN_IDLE,
```
hodgesds, line 2684: Is this flag available for backports as well?

hodgesds (approved):
I think this makes sense, it's definitely far easier to verify and maintain.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: Yeah, that flag has been around for a long time.

likewhatevs (approved):
I like complexity being pushed to rust side and I can follow what this is doing. Also fewer bpf.c and bpf.h files is nice too.

likewhatevs:
I've been looking at this PR merging for what feels like all day so I'm going to just merge it, something weird up w/ bpfland job, rest is good.

PR #997, opened 2024-11-28
scx_layered: Reimplement layered_dispatch()

htejun:
Layer and LLC orderings are now determined by userspace and BPF uses two helpers to iterate over the combinations. The overhead is lower for both topo and no-topo cases. See the reimplement commit for details.

Remaining todos:
- Layer growth doesn't always work as expected likely due to owned execution accounting and protection issue.
- scx_layered process itself should be prioritized by queueing to hi fallback.
- Implement DSQ latency based in-layer execution protection so that latency sensitive workloads can be protected from preempt workloads.
- Integrate stickiness mechanism.
- Split on node boundaries.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -160,6 +160,10 @@ struct cpu_ctx {
 	u32			llc_id;
 	u32			node_id;
 	u32			perf;
+
+	u32			open_preempt_layer_order[MAX_LAYERS];
+	u32			open_layer_order[MAX_LAYERS];
+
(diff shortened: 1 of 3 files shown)


likewhatevs (approved):
love all this hoisting complexity into rust code where it is harder to mess things up and there is no verifier

likewhatevs:
This isn't exactly composability but like, if the definitions accessed via bindgen (i.e. cpu_ctx/stuff in bpf_intf) were moved into common, and multiple schedulers used them, the logic (like creating cpu_ctx's, dealing w/ prox maps) could be moved there.

JakeHillion (commented):
Much happier to see this logic in rust, thank you!

PR #999, opened 2024-11-28
scx_bpfland: dump cache_id_map in ascending order

mmz-zmm:
Use BTreeMap to store cache_id_map so that prog can show L2/L3 cache info in ascending order, make it easy to lookup by human.


arighi (approved):
LGTM, thanks!

PR #1000, opened 2024-11-28
scx_lavd: Load balancing across compute domains

multics69:
Perform load balancing across compute domains (e.g., hybrid cores, CCX, or NUMA). The load balancing tries to make the average length of per-compute domain DSQ similar (+/- 12.5%). If not, an under-loaded domain (stealer) will steal a task from an over-loaded domain (stealee). To avoid a central hotspot, all the operations are done in a distributed and probabilistic manner.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -1108,7 +1108,7 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 	}
 }
 
-static bool consume_dsq(s32 cpu, u64 dsq_id, u64 now)
+static bool consume_dsq(u64 dsq_id)
 {
 	struct cpdom_ctx *cpdomc;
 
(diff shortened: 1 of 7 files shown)


scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
```diff
@@ -63,10 +64,66 @@ static void init_sys_stat_ctx(struct sys_stat_ctx *c)
...
+	 */
+	bpf_for(dsq_id, 0, nr_cpdoms) {
+		if (dsq_id >= LAVD_CPDOM_MAX_NR)
+			break;
+
+		nr_q_tasks = scx_bpf_dsq_nr_queued(dsq_id);
+		c->nr_queued_task += nr_q_tasks;
+
+		cpdomc = MEMBER_VPTR(cpdom_ctxs, [dsq_id]);
+		cpdomc->nr_q_tasks_per_cpu = (nr_q_tasks * 1000) / cpdomc->nr_cpus;
```
vax-r, line 85: Small question, why do we need `* 1000` here? guessing you want to keep the value to 2 decimal places?
`avg_nr_q_tasks_per_cpu >> LAVD_CPDOM_MIGRATION_SHIFT` is going be performed, wouldn't `nr_q_tasks << LAVD_CPDOM_MIGRATION_SHIFT` be more intuitive ?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1129,18 +1129,109 @@ static bool consume_dsq(u64 dsq_id)
...
+	/*
+	 * Traverse neighbor compute domains in distance order.
+	 */
+	nuance = bpf_get_prandom_u32();
 	for (int i = 0; i < LAVD_CPDOM_MAX_DIST; i++) {
 		nr_nbr = min(cpdomc->nr_neighbors[i], LAVD_CPDOM_MAX_NR);
 		if (nr_nbr == 0)
 			break;
 
-		nuance = bpf_get_prandom_u32();
```
vax-r: Want to know what's the reason for moving this out of the for-loop?
wouldn't tasking a pseudo random number every time increase randomness?

scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
multics69: Yes, it is to keep 2 decimal places for better accuracy. LAVD_CPDOM_MIGRATION_SHIFT is to maintain +/- 12.5% delta.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: It is the tradeoff between overhead and randomness.

multics69:
Thank you for the review @htejun and @vax-r !

PR #1006, opened 2024-11-28
scx_rusty: Refactor access operation of nodemask

vax-r:
## Summary
Within `task_set_preferred_mempolicy_dom_mask()`, it tries to access the nodemask of the task directly. BPF verifier might not allow it under all situation.

Use BPF_CORE_READ() macro to let libbpf handle CO-RE relocate to make sure it can pass the verifier.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -1264,7 +1264,6 @@ static void task_set_preferred_mempolicy_dom_mask(struct task_struct *p,
 {
 	u32 node_id;
 	u32 val = 0;
-	nodemask_t *node_mask = &p->mempolicy->nodes;
 	void *mask;
 
 	taskc->preferred_dom_mask = 0;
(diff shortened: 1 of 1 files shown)


etsal (approved):
Looks good to me. Thank you!

vax-r:
> Did it actually fail verification? BPF should CO-RE relocate `p->` deref automatically. I don't think it needs an explicit `BPF_CORE_READ()`.

Yes , Mr.Chaturvedi posted his failed log [here](https://bin.tablaster.dev/zJ0D4PNG), he mentioned it in slack channel.
I think he haven't opened an issue yet.

likewhatevs:
I opened GH-1009 because of this issue fwiw (no idea which kernel version this is for).

Blaster4385:
Hi @likewhatevs this was on Mainline kernel 6.12.1

htejun (approved):

vax-r:
> Thank you @Blaster4385 . I wasn't able to reproduce that error though and it makes me wonder what additional information we need to track in debug output to be able to repro... Also I tried on 6.12.1 w/o clang/lto kernel (not that that should matter tbh afaik).

@likewhatevs -
Could it possibly be the version of libbpf or its configuration ? ( Since the problem is the verifier doesn't allow the operation , I guess it's more relate to libbpf ? )

PR #1008, opened 2024-11-28
scx_layered: Prioritize sched userspace and fix owned execution protection

htejun:
- Always put scx_layered userspace in hi fallback DSQ to guarantee timely execution of scheduler userspace code. This is required as reasonable forward progress depends on userspace code running. This assumes that scx_layered is running in the root PID namespace. It might be useful to add a warning when that's not the case - e.g. look up the task from the BPF side and verify that it actually is scx_layered.
- Owned execution protection had a couple bugs which could weaken protection and prevent a layer with high util growth bound from growing. Fix them. After this change, a saturating workload can reliably grow from empty under competition from preempting layers.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1805,52 +1805,67 @@ void BPF_STRUCT_OPS(layered_stopping, struct task_struct *p, bool runnable)
 {
 	struct cpu_ctx *cpuc;
 	struct task_ctx *taskc;
-	struct layer *layer;
+	struct layer *task_layer, *cpu_layer = NULL;
 	u64 now = bpf_ktime_get_ns();
-	s32 lid;
-	u64 used;
+	bool is_fallback;
+	s32 task_lid, target_ppk;
+	u64 used, cpu_slice;
 
 	if (!(cpuc = lookup_cpu_ctx(-1)) || !(taskc = lookup_task_ctx(p)))
 		return;
+	is_fallback = cpuc->cpu == fallback_cpu;
 
-	lid = taskc->layer_id;
-	if (!(layer = lookup_layer(lid)))
+	task_lid = taskc->layer_id;
+	if (!(task_layer = lookup_layer(task_lid)))
+		return;
+
+	if (cpuc->layer_id != MAX_LAYERS &&
+	    !(cpu_layer = lookup_layer(cpuc->layer_id)))
 		return;
 
 	used = now - taskc->running_at;
 
-	u64 slice_ns = layer_slice_ns(layer);
-
 	if (cpuc->running_owned) {
-		cpuc->layer_usages[lid][LAYER_USAGE_OWNED] += used;
+		cpuc->layer_usages[task_lid][LAYER_USAGE_OWNED] += used;
 		if (cpuc->protect_owned)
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/main.rs
```diff
@@ -1584,6 +1584,9 @@ impl<'a> Scheduler<'a> {
         skel.struct_ops.layered_mut().exit_dump_len = opts.exit_dump_len;
 
         skel.maps.rodata_data.debug = opts.verbose as u32;
+        // Running scx_layered inside a PID namespace would break the
+        // following.
+        skel.maps.rodata_data.layered_tgid = unsafe { libc::getpid() };
```
JakeHillion, line 1589: nit: Pretty sure this is the same and is safe. It may need a cast though and in that case up to you.
```suggestion
        skel.maps.rodata_data.layered_tgid = std::process::id();
```

JakeHillion (approved):
Glad to see more moves in the Rust direction.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1584,6 +1584,9 @@ impl<'a> Scheduler<'a> {
         skel.struct_ops.layered_mut().exit_dump_len = opts.exit_dump_len;
 
         skel.maps.rodata_data.debug = opts.verbose as u32;
+        // Running scx_layered inside a PID namespace would break the
```
hodgesds, line 1587: I think if we write the namespace id as well we can translate it on the bpf side, we already have the [helpers](https://github.com/sched-ext/scx/blob/main/scheds/include/scx/namespace.bpf.h).

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1005,13 +1006,13 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
...
-	 * to give reasonable boost to per-cpu kthreads by default as they are
-	 * usually important for system performance and responsiveness.
+	 * Special-case per-cpu kthreads and scx_layered userspace so that they
+	 * run before preempting layers. This is to guarantee timely execution
+	 * of layered userspace code and give boost to per-cpu kthreads as they
+	 * are usually important for system performance and responsiveness.
 	 */
-	if (!layer->preempt &&
-	    (p->flags & PF_KTHREAD) && p->nr_cpus_allowed < nr_possible_cpus) {
+	if (((p->flags & PF_KTHREAD) && p->nr_cpus_allowed < nr_possible_cpus) ||
```
hodgesds, line 1014: Maybe we can have a helper function since this is getting pretty complicated, something like `is_fallback_task`?

hodgesds (approved):
Few nits, can be cleaned up later. LGTM

likewhatevs (approved):
LGTM

etsal (approved):

scheds/rust/scx_layered/src/main.rs
htejun: Ah, cool, I'll do that later.

scheds/rust/scx_layered/src/bpf/main.bpf.c
hodgesds: Yeah, that's fair.

Issue #1010, opened 2024-11-29
Enable Link-Time Optimization (LTO)


zamazan4ik:
Hi!

I noticed that in the build scripts Link-Time Optimization (LTO) for the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve the application's performance. If you want to read a bit more about LTO, I can recommend starting from [this](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto) Rustc documentation (similar pages exist for C compilers like GCC and Clang - `-flto` compiler switch).

I suggest enabling LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project since LTO consumes an additional amount of time to finish the compilation routine. If you think that a regular Release build should not be affected by such a change as well, then I suggest adding an additional `dist` or `release-lto` profile where in addition to regular `release` optimizations LTO will also be added. Such a change simplifies life for maintainers and others interested in the project persons who want to build the most performant version of the application. E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48).

Basically, it can be enabled with the following lines for the Rust part:

```
[profile.release]
lto = true
```

I have made quick tests (Fedora 41, Rustc 1.83) by adding `lto = true` to the Release profile. The binary size reduction is the following:

* `scp_bpflang`: from 5 Mib to 3.9 Mib
* `scx_flash`: from 4.9 Mib to 3.9 Mib
* `scx_lavd`: from 5.3 Mib to 4.2 Mib
* `scx_layered`: from 6 Mib to 4.7 Mib
* `scx_loader`: from 6 Mib to 4.5 Mib
* `scx_mitosis`: from 2.1 Mib to 1.8 Mib
* `scx_rlfifo`: from 3.5 Mib to 2.8 Mib
* `scx_rustland`: from 5 to 3.9 Mib
* `scx_rusty`: from 5.5 Mib to 4.4 Mib

I also suggest enabling LTO for the C part of the project - didn't perform such tests though but it should work too.

We discussed a bit this topic with @arighi in Discord - I will provide a PR for the Rust part of the project with enabling LTO. Regarding the C part - I am not sure since I am not an expert in Meson.

Thank you.

hodgesds:
> I suggest enabling LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project since LTO consumes an additional amount of time to finish the compilation routine.

[...]

PR #1011, opened 2024-11-29
Enable Link-Time Optimization (LTO) for the Rust part

zamazan4ik:
Related to https://github.com/sched-ext/scx/issues/1010 but doesn't completely resolve it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,6 @@ members = ["rust/scx_stats",
            "scheds/rust/scx_layered",
            "scheds/rust/scx_mitosis"]
 resolver = "2"
+
+[profile.release]
+lto = true


arighi (commented):
What is the unresolved part? The C schedulers? Also it'd be nice to have a longer description in the commit message (like a summary of what you wrote in #1010), thanks

zamazan4ik:
> What is the unresolved part? The C schedulers?

Yep.

> Also it'd be nice to have a longer description in the commit message (like a summary of what you wrote in https://github.com/sched-ext/scx/issues/1010), thanks

Just did it :)

Cargo.toml
```diff
@@ -13,3 +13,6 @@ members = ["rust/scx_stats",
            "scheds/rust/scx_layered",
            "scheds/rust/scx_mitosis"]
 resolver = "2"
+
+[profile.release]
```
hodgesds, line 17: Maybe it's better to add another [profile instead](https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles):
```
[profile.release-lto]
inherits = "release"
lto = true
```

ptr1337: Personally I would prefer, that the default would use LTO, and for the CI we pass LTO off or so.

sirlucjan: The meson flag can be added, then we activate this in PKGBUILD.

ptr1337:
What is pending here?
I think the default release profile should use LTO and the debug variant just disables it.

zamazan4ik:
I was waiting for the final decision about how exactly LTO should be enabled for the project: in a dedicated profile or in the regular Release Cargo profile. As far as I understand, no final agreement is met yet.

htejun (approved):

htejun:
I don't see a reason to not enable it by default on release builds. Let's enable it and see whether anyone complains.

htejun (approved):

PR #1012, opened 2024-11-29
scx_layered: State tracking updates and layer sizing related fixes

htejun:
- Drop now unused load metric.
- Implement per-task avg runtime tracking.
- Layer sizing related bug fixes.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -352,6 +352,8 @@ static bool refresh_cpumasks(u32 layer_id)
 				cpuc->layer_id = layer_id;
 				bpf_cpumask_set_cpu(cpu, layer_cpumask);
 				total++;
+
+				scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
 			} else {
 				if (cpuc->layer_id == layer_id)
 					cpuc->layer_id = MAX_LAYERS;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -38,7 +38,6 @@ const volatile u32 nr_nodes = 32;	/* !0 for veristat, set during init */
 const volatile u32 nr_llcs = 32;	/* !0 for veristat, set during init */
 const volatile bool smt_enabled = true;
 const volatile bool has_little_cores = true;
-const volatile bool disable_topology = false;
 const volatile bool xnuma_preemption = false;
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -386,6 +352,8 @@ static bool refresh_cpumasks(u32 layer_id)
 				cpuc->layer_id = layer_id;
 				bpf_cpumask_set_cpu(cpu, layer_cpumask);
 				total++;
+
+				scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
```
hodgesds, line 356: Silly question, what does the kick idle do in this path?

likewhatevs (approved):
LGTM

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2074,6 +2078,14 @@ s32 BPF_STRUCT_OPS(layered_init_task, struct task_struct *p,
 	taskc->last_cpu = -1;
 	taskc->layer_id = MAX_LAYERS;
 	taskc->refresh_layer = true;
+	taskc->prev_llc_id = MAX_LAYERS;
+
+	/*
+	 * Start runtime_avg at some arbitrary sane-ish value. If this becomes a
```
hodgesds, line 2084: I guess another alternative if the task were to be assigned a `prev_llc_id` it could just be initialized to the current avg for the `llcc->queued_runtime[layer_id]`, but this should work.

hodgesds (approved):
LGTM. I think there still might be a layer shrinking issue. When I run a `stress-ng -c 2 -t 15 -M` it grows to two CPUs and then gets compacted down to a single CPU.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: e.g. Let's say a system has a confined layer which is currently the only active one. A lot of tasks woke up to the layer just now and the layer expanded quite a bit. Let's say that the tasks don't sleep much - they just wake up, run for a long time and then go back to sleep. Without the above kick, when the layer expands with those tasks queued, the new CPUs will sit idle until some tasks wake up to them.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: `queued_runtime` is the average of the sum of all queued tasks' avg runtimes, which can run up quite high under saturation. Ideally, this I think should be the avg of the first runtimes of the children for each parent.

PR #1013, opened 2024-11-30
scx_layered: Deprecate idle_smt layer config

htejun:
24fba4ab8de9 ("scx_layered: Add idle smt layer configuration") added the idle_smt layer config but

- It flipped the default from preferring idle cores to not having preference.

- Misdocumented what it meant. It doesn't only pick idle cores. It tries to pick an idle core first and if that fails pick any idle CPU.

A follow-up commit 637fc3f6e136 ("scx_layered: Use layer idle_smt option") made it more confusing. If idle_smt is set, the idle core prioritizing logic is disabled.

The first commit disables idle core prioritization by overriding idle_smtmask to be idle_cpumask if idle_smt is *clear* and the second commit disables the same by disabling the code path when the flag is *set*. ie. Both options did exactly the same thing.

Recently, 75dd81e3e657 ("scx_layered: Improve topology aware select_cpu()") restored the function of the flag by dropping the cpumask override. However, this made the actual behavior the opposite of the documented one by leaving only the behavior of the second commit which implemented the reverse behavior.

This flag is hopeless. History aside, the name itself is too confusing. idle_smt - is it saying that the flag is going to prefer idle smt *thread* or idle smt *core*? While the name is transferred from idle_cpumask/smtmask, there, the meaning of the former is clear which also makes it difficult to confuse what the latter means.

Preferring idle cores was one of the drivers of performance gain identified during earlier ads experiments. Let's just drop the flag to restore the previous behavior and retry if necessary.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -105,7 +105,7 @@ lazy_static! {
                         preempt: false,
                         preempt_first: false,
                         exclusive: false,
-                        idle_smt: false,
+                        idle_smt: None,
                         slice_us: 20000,
                         weight: DEFAULT_LAYER_WEIGHT,
                         growth_algo: LayerGrowthAlgo::Sticky,
@@ -129,7 +129,7 @@ lazy_static! {
                         preempt: true,
                         preempt_first: false,
                         exclusive: true,
-                        idle_smt: false,
+                        idle_smt: None,
                         slice_us: 20000,
                         weight: DEFAULT_LAYER_WEIGHT,
                         growth_algo: LayerGrowthAlgo::Sticky,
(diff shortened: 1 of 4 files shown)


hodgesds:
I think the problem of this flag was an initial misunderstanding of what `pick_idle_cpu_from` does. What it was trying to achieve was keep big cores hot compared to spreading load more evenly, which did improve some benchmarks a bit.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -2315,6 +2312,12 @@ fn main() -> Result<()> {
         );
     }
 
+    for spec in &layer_config.specs {
```
hodgesds, line 2315: Maybe we could have a macro for deprecating fields to generate these warnings, could be a good help wanted task.

hodgesds (approved):
LGTM

likewhatevs (approved):
LGTM

Issue #810, opened 2024-10-16
scx_lavd: verifier complains invalid access


abrehman94:
Verifier Log

```
    ; bpf_cpumask_and(t2_cpumask, cast_mask(t_cpumask), cast_mask(cpdom_mask_prev)); @ main.bpf.c:836
    392: (bf) r1 = r7                     ; R1_w=rcu_ptr_bpf_cpumask() R7_w=rcu_ptr_bpf_cpumask()
    393: (79) r2 = *(u64 *)(r10 -120)     ; R2_w=rcu_ptr_bpf_cpumask() R10=fp0 fp-120=rcu_ptr_bpf_cpumask()
    394: (79) r3 = *(u64 *)(r10 -88)      ; R3_w=map_value(map=.data.LAVD,ks=4,vs=1320,off=40,smin=smin32=0,smax=umax=smax32=umax32=1240,var_off=(0x0; 0x7f8)) R10=fp0 fp-88=map_value(map=.data.LAVD,ks=4,vs=1320,off=40,smin=smin32=0,smax=umax=smax32=umax32=1240,var_off=(0x0; 0x7f8))
    395: (85) call bpf_cpumask_and#65708
    invalid access to map value, value_size=1320 off=1280 size=1024
    R3 max value is outside of the allowed memory range
    processed 365 insns (limit 1000000) max_states_per_insn 1 total_states 35 peak_states 35 mark_read 19
```

Sched Config
```
    Opts {                               
        autopilot: true,                 
        autopower: false,                
        performance: false,              
        powersave: false,                
        balanced: false,                 
        no_core_compaction: false,       
        prefer_smt_core: false,          
        prefer_little_core: false,       
        no_prefer_turbo_core: false,     
        no_freq_scaling: false,          
        stats: None,                     
        monitor: None,                   
        monitor_sched_samples: None,     
        verbose: 1,                      
        version: false,                  
        help_stats: false,               
    }                                    
```

Setup:
* **Linux Version:** `Linux v-021 6.11.0-rc1-scx1+ #6 SMP PREEMPT_DYNAMIC Mon Oct 14 12:34:19 EDT 2024 x86_64 x86_64 x86_64 GNU/Linux`
* Machine: Server with [Intel® Xeon® Platinum 8160 Processor](https://www.intel.com/content/www/us/en/products/sku/120501/intel-xeon-platinum-8160-processor-33m-cache-2-10-ghz/specifications.html)
** it's a dual socket machine with 24 cores each socket
[...]

PR #1016, opened 2024-12-02
vmlinux: Increase cpumask size in vmlinux.h for all arch

multics69:
The size of CPU mask (struct cpumask) is defined as too small in all architectures (e.g., unsigned long * 4) except for x86. It triggers a verifier error in lavd (e.g., arm64), so increase its size to the same as x86. The changes were tested on an arm64 platform.

This solves the following issues (on arm64).

   https://github.com/sched-ext/scx/issues/810

### DIFF
--- a/scheds/include/arch/arm/vmlinux-v6.12-rc2-g5b7c893ed5ed.h
+++ b/scheds/include/arch/arm/vmlinux-v6.12-rc2-g5b7c893ed5ed.h
@@ -2507,7 +2507,7 @@ struct sched_statistics {
 };
 
 struct cpumask {
-	unsigned long bits[1];
+	unsigned long bits[128];
 };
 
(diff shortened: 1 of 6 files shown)


likewhatevs:
I say lets do this to fix CI because *maybe* the bad vmlinux files were from an RC kernel w/ a bug around this (the size 1 arm cpumask looks really wrong). We can fix this properly when we regenerate vmlinux files (where CI would fail if this issue is still present, I think).

multics69:
Thanks for the review @likewhatevs !

hodgesds:
> We can fix this properly when we regenerate vmlinux files (where CI would fail if this issue is still present, I think).

I wonder if this is a problem with `bpftool` making assumptions when generating vmlinux. Might be worth reporting upstream.

PR #1017, opened 2024-12-02
[RFC] scx_lavd: Fetch active profile from power-profiles-daemon dbus interface when using autopower

ruihe774:
Current implementation of `--autopower` reads `energy_performance_preference` to determine which power profile is active. IMO this is sub-optimal and does not play well with CPUs without EPP and alternative power management daemons like `tuned-ppd`. This PR implements a better method that fetching active power profile from the dbus interface [`net.hadess.PowerProfiles`](https://hadess.fedorapeople.org/power-profiles-daemon-docs/gdbus-net.hadess.PowerProfiles.html), which is exposed by `power-profiles-daemon` and `tuned-ppd`. (There is a standardized interface [`org.freedesktop.UPower.PowerProfiles`](https://upower.pages.freedesktop.org/power-profiles-daemon/gdbus-org.freedesktop.UPower.PowerProfiles.html); unfortunately, it is currently not implemented by `tuned-ppd`.)

This is a RFC. Maybe it is better to move this functionality into scx_utils and provide options to choose which probing methods to use.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-executor"
+version = "1.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "30ca9a001c1e8ba5149f91a74362376cc6bc5b919d92d988668657bd570bdcec"
+dependencies = [
+ "async-task",
+ "concurrent-queue",
+ "fastrand",
+ "futures-lite",
+ "slab",
+]
+
+[[package]]
+name = "async-fs"
+version = "2.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ebcd09b382f40fcd159c2d695175b2ae620ffa5f3bd6f664131efff4e8b9e04a"
+dependencies = [
+ "async-lock",
+ "blocking",
+ "futures-lite",
+]
+
+[[package]]
 name = "async-io"
 version = "2.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -863,8 +887,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d6401deb83407ab3da39eba7e33987a73c3df0c82b4bb5813ee871c19c41d48"
 dependencies = [
  "futures-core",
+ "futures-io",
  "futures-sink",
  "futures-task",
+ "memchr",
  "pin-project-lite",
  "pin-utils",
  "slab",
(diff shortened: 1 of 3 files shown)


multics69 (approved):
Looks good. Thanks!

multics69:
In my opinion, it makes sense moving this to scx_utils. Could you please open another PR for that?

ruihe774:
> It triggers the following error. Can we fix this somehow?
> 
> error: package `zvariant v5.1.0` cannot be built because it requires rustc 1.80 or newer, while the currently active rustc version is 1.75.0 Either upgrade to rustc 1.80 or newer, or use cargo update zvariant@5.1.0 --precise ver where `ver` is the latest version of `zvariant` supporting rustc 1.75.0

It's related to [MSRV](https://rust-lang.github.io/rfcs/2495-min-rust-version.html).

The scx_* Rust packages do not specify a MSRV. So I used the Rust toolchain specified in `rust-toolchain.toml` (nightly-2024-09-10), which can compile `zvariant v5.1.0` successfully. In order to support rustc 1.75.0, we can find an older version of `zvariant`. Or you can upgrade your Rust toolchain to the one specified in `rust-toolchain.toml`.

ruihe774:
> In my opinion, it makes sense moving this to scx_utils. Could you please open another PR for that?

Ok.

I find scx_bpfland also does power profile detection. I'm going to extract the reusable part.

multics69:
It would be nice to find an older version of zvariant, so a bit old version of rustic can handle it.

ruihe774:
> It would be nice to find an older version of zvariant, so a bit old version of rustic can handle it.

I'll do it in another PR.

Jackaed:
@ruihe774 Is it possible to get this to also support `tlp`? Seems like `tlp` is increasingly popular.

ruihe774:
> @ruihe774 Is it possible to get this to also support `tlp`? Seems like `tlp` is increasingly popular.

tlp makes decision based on whether the computer is connected to AC or using battery. If we want to achieve the same effect, we can just read `/sys/class/power_supply/`. I do not find tlp exposing an interface.

Jackaed:
> > @ruihe774 Is it possible to get this to also support `tlp`? Seems like `tlp` is increasingly popular.
> 
> tlp makes decision based on whether the computer is connected to AC or using battery. If we want to achieve the same effect, we can just read `/sys/class/power_supply/`. I do not find tlp exposing an interface.

Ah, makes sense. Sorry to disturb you! Been using the scheduler pretty flawlessly for months now.

multics69:
> @ruihe774 Is it possible to get this to also support `tlp`? Seems like `tlp` is increasingly popular.

Thanks for the suggestion! Checking if the AC power is connected or not is one in my backlog. Could you open an issue so I can push myself to implement it? :-)

PR #1019, opened 2024-12-02
deps: Fix compilation on older Rust toolchain

ruihe774:
This PR does the following to allow compilation with older Rust toolchain:

- Downgrade Rust version specified in `rust-toolchain.toml` to 1.77. (`sched_mitosis` requires `std::mem::offset_of`, so the Rust version cannot be further downgraded.)
- Downgrade the dependency `zbus` to v4.

@multics69 @JakeHillion

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1767,7 +1767,7 @@ dependencies = [
  "serde",
  "simplelog",
  "static_assertions",
- "zbus 5.1.1",
+ "zbus",
 ]
 
 [[package]]
@@ -1808,8 +1808,8 @@ dependencies = [
  "sysinfo",
  "tokio",
  "toml",
- "zbus 4.4.0",
- "zvariant 4.2.0",
+ "zbus",
+ "zvariant",
 ]
 
 [[package]]
@@ -2866,9 +2866,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bb97012beadd29e654708a0fdb4c84bc046f537aecfde2c3ee0a9e4b4d48c725"
 dependencies = [
  "async-broadcast",
+ "async-executor",
+ "async-fs",
+ "async-io",
+ "async-lock",
  "async-process",
  "async-recursion",
+ "async-task",
  "async-trait",
+ "blocking",
  "enumflags2",
  "event-listener",
  "futures-core",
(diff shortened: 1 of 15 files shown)


ruihe774:
`scx_mitosis` requires Rust v1.77 or newer; however, the Rust in CI is v1.75 (it is not fixed and will be upgraded along CI runner image upgrades). We must solve this issue before `scx_mitosis` can be tested in CI. Please consider using an action (a commonly used one is [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain)) to setup a specific Rust toolchain.

multics69 (commented):
Do we need to downgrade the rust toolchain when downgrading zbus?

multics69 (commented):
Is reformatting the rust code irrelevant to this PR??

JakeHillion:
> Is reformatting the rust code irrelevant to this PR??

@multics69 I think it might be necessary, for some reason we have to use the nightly toolchain to format our rust code and it ends up being quite toolchain specific. We have to stay in sync with whatever the Rust toolchain wants which is why the formatting toolchain is pinned properly (apparently the build isn't).

@ruihe774 isn't the real fix here to pin the toolchain in the CI, like we do for the formatting toolchain? I would rather accept a PR doing that than pinning everyone else to the version the toolchain uses (can do it myself if you don't have time).

likewhatevs:
the pinning (which was present) wasn't working btw, https://github.com/sched-ext/scx/pull/1020 fixes that + the build

ruihe774:
> Do we need to downgrade the rust toolchain when downgrading zbus?

This is to avoid developers writing code that does not supported by older Rust toolchains. It is usual to have the same version of Rust in local workspace and in CI.

> @multics69 I think it might be necessary, for some reason we have to use the nightly toolchain to format our rust code and it ends up being quite toolchain specific. We have to stay in sync with whatever the Rust toolchain wants which is why the formatting toolchain is pinned properly (apparently the build isn't).

Yes.

> the pinning (which was present) wasn't working btw, #1020 fixes that + the build

@likewhatevs A problem about #1020 is that it raise the minimal supported Rust version. As we've discussed in https://github.com/sched-ext/scx/pull/1017#issuecomment-2510607780, it may be better to keep supporting older Rust toolchains.

likewhatevs:
> @likewhatevs A problem about https://github.com/sched-ext/scx/pull/1020 is that it raise the minimal supported Rust version. As we've discussed in https://github.com/sched-ext/scx/pull/1017#issuecomment-2510607780, it may be better to keep supporting older Rust toolchains.

Supporting older toolchains is good also tbh, no pushback on that

ruihe774:
> > @likewhatevs A problem about #1020 is that it raise the minimal supported Rust version. As we've discussed in [#1017 (comment)](https://github.com/sched-ext/scx/pull/1017#issuecomment-2510607780), it may be better to keep supporting older Rust toolchains.
> 
> Supporting older toolchains is good also tbh, no pushback on that

Great.

likewhatevs:
Ahh should I open PR to do this or do you want to @ruihe774 , kinda thought this one was going to do that, LMK.

PR #1020, opened 2024-12-02
scx: fix build

likewhatevs:
bump pinned rust toolchain to fix the following issue w/ lavd build:
```
  Downloaded anstyle-query v1.1.1
error: package `zbus_names v4.1.0` cannot be built because it requires rustc 1.80 or newer, while the currently active rustc version is 1.75.0
Either upgrade to rustc 1.80 or newer, or use
cargo update zbus_names@4.1.0 --precise ver
```

also only set toolchain version, leave the rest at defaults.


likewhatevs:
hmm, weird how that didn't work

.github/actions/install-deps-action/action.yml
```diff
@@ -23,7 +23,12 @@ runs:
         linux-headers-generic linux-tools-common linux-tools-generic make \
         ninja-build pahole pkg-config python3-dev python3-pip python3-requests \
         qemu-kvm rsync stress-ng udev zstd libseccomp-dev libcap-ng-dev \
-        llvm clang python3-full curl meson bpftrace cargo rustc dwarves
+        llvm clang python3-full curl meson bpftrace dwarves rustup
+      shell: bash
+
+    # ensure some toolchain is installed
+    - run: |
+        rustup default nightly
```
JakeHillion, line 31: This was pinned in `lint` before but not in install deps.

Can we do the same tas here so we don't install  a redundant toolchain in this step:

https://github.com/sched-ext/scx/blob/c1690c3dbd93a6aec2c34ad42ea91669e4cc2895/.github/workflows/caching-build.yml#L20

likewhatevs: kinda prefer default nightly so a default is set for the virtiofsd install later on in the compile kernel path when the cache misses

JakeHillion: Wouldn't it be more effective to copy our rust-toolchain.toml into that build too? It's not going to be reproducible either way due to the caching, but it's at least then using a toolchain we might have used to build scx.

etsal:
LGTM, can we maybe spin off the (I assume automatic) header reorg into a separate patch?

etsal (approved):

likewhatevs:
> LGTM, can we maybe spin off the (I assume automatic) header reorg into a separate patch?

Some backstory on that, what cargo fmt does is tied to the version of the nightly compiler we're using. The lint job fails w/o the formatting so rust version changes require formatting changes.

likewhatevs:
split the patches, good call, thx.

PR #1025, opened 2024-12-02
scx: fix pages build

likewhatevs:
this fixes the pages build here: 

https://github.com/likewhatevs/scx/actions/runs/12124307848/job/33801740919


etsal (approved):
LGTM

PR #1026, opened 2024-12-02
scx_layered: Make vtime_now per-LLC

htejun:
On multi-LLC machines, a layer uses one DSQ per LLC; however, the DSQs all
share one per-layer vtime_now. Cacheline contention aside, this makes
vtime_now progression erratic as a single DSQ executing can easily progress
the per-layer vtime_now past the heads of other DSQs which in turn can
collapse vtime based ordering in other DSQs.
    
Make vtime_now per DSQ and translate between vtimelines when a tasks jumps
between LLCs. In rd-hashd CPU saturation scenario on ryzen 3900x, this
reduces the gap from EEVDF from >10% to ~5%.


likewhatevs (approved):
cool and makes sense

PR #1027, opened 2024-12-02
scx_lavd: have ci use an older rust release, make that build

likewhatevs:
@ruihe774 @multics69 this should get CI working with an older version of rust for lavd. 

Not sure what version of rust is preferable, but this is a way to make CI run any version.

The first commit in this PR should fail CI (confirming this way of pinning versions works on CI), and the second should pass (fixing the build after this change).


likewhatevs:
alright, this works as expected now (misspelled scheduler first time), LMK if it looks good (or bad), I think this is a *reasonable* way to test against an older versions of rust on CI.

This is kinda-sorta a workaround over the limitations of rustup override (which I would recommend for folks working on schedulers targeting older versions of rust, so that local builds can fail like CI ones would if versions do not work out etc.).

ruihe774:
I think we should test all schedulers with an older Rust toolchain, not just scx_lavd.

I think nightly-2024-04-09 is still too new. We should at least test a stable toolchain.

IMO 1.77 is a good choice, giving scx_mitosis requires `offset_of`.

What's the meaning of `RUSTUP_TOOLCHAIN`? Can we just use an action like [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) to setup the Rust toolchain with specific version?

ruihe774 (changes requested):
I think we should test all schedulers with an older Rust toolchain, not just scx_lavd.

I think nightly-2024-04-09 is still too new. We should at least test a stable toolchain.

IMO 1.77 is a good choice, giving scx_mitosis requires `offset_of`.

What's the meaning of `RUSTUP_TOOLCHAIN`? Can we just use an action like [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) to setup the Rust toolchain with specific version?

I find you have downgraded zbus version. Did you update `Cargo.lock`?

likewhatevs:
> I think we should test all schedulers with an older Rust toolchain, not just scx_lavd.

I don't know what versions folks want for other schedulers, but I am reasonably (if not completely) confident that scx_layered and scx_rusty want latest nightly.

> IMO 1.77 is a good choice, giving scx_mitosis requires offset_of.

nightly-2024-04-09 is 1.77, RUST_TOOLCHAIN="x" works with `nightly-2024-04-09`, but not `1.77`, but those two are the same.

> Did you update Cargo.lock? Please also commit it.

will do

> Can we just use an action like [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) to setup the Rust toolchain with specific version?

I wouldn't recommend this tbh. Complexity here is high enough that doesn't enough sufficiently simplify things to be worth using/adding into the mix IMO.

> I have a question that how to use the older Rust toolchain instead of the Rust toolchain specified in rust-toolchain.toml in CI? AFAIK the latter will override the toolchain version.

RUST_TOOLCHAIN overrides the rust_toolchain file.

ruihe774:
> I don't know what versions folks want for other schedulers, but I am reasonably (if not completely) confident that scx_layered and scx_rusty want latest nightly (or roughly, whatever is easiest/reasonably new/in the top level rust-toolchain tbh).

Hmm, considering this, I don't think it makes sense to *only* require older Rust compatibility for scx_lavd. If scx_layered and scx_rusty require a latest Rust toolchain, I think it is also okay for scx_lavd to require it. And it's better for all schedulers to stick to a same Rust toolchain. @multics69 What's your opinion?

> nightly-2024-04-09 is 1.77, RUST_TOOLCHAIN="x" works with nightly-2024-04-09, but not 1.77, but those two are the same.

It's not:
```console
$ rustc +nightly-2024-04-09 --version
rustc 1.79.0-nightly (ab5bda1aa 2024-04-08)
```

likewhatevs:
> It's not:

good catch, thx, fixed that (got the month wrong before I think).

> Maybe you can try RUST_TOOLCHAIN="1.77.0".

this did not work

> Hmm, considering this, I don't think it makes sense to only require older Rust compatibility for scx_lavd.

I kinda disagree. If folks want lavd to use an older version of rustc (i.e. because perhaps places where lavd is used downstream require that), which I think is the case due to https://github.com/sched-ext/scx/pull/1017#issuecomment-2510607780, I'm all for this, because it makes CI more useful for folks.

likewhatevs:
also I set the version of cargo used to 1.79/04-09 because that's the oldest month I could get lavd to build locally (before that, I get compiler crashes or lockfile format invalid).

ruihe774:
> I kinda disagree. If folks want lavd to use an older version of rustc (i.e. because perhaps places where lavd is used downstream require that), which I think is the case due to https://github.com/sched-ext/scx/pull/1017#issuecomment-2510607780, I'm all for this, because it makes CI more useful for folks.

https://github.com/sched-ext/scx/pull/1017#issuecomment-2510607780 is just a discussion. We can re-consider this issue.

IMO it does not make sense to have different MSRVs giving all schedulers are built and distributed as a single package.

likewhatevs:
> IMO it does not make sense to have different MSRVs giving all schedulers are built and distributed as a single package.

I don't know that's the case.

> https://github.com/sched-ext/scx/pull/1017#issuecomment-2510607780 is just a discussion. We can re-consider this issue.

Alright, I'll go ahead and close this PR.

My main interest in this PR is ensuring that schedulers are not constrained by each others requirements (and that we'll find some way to make whatever is preferred work well, because issues like this will come up over time), so long as that's clear, I'm good.

PR #1029, opened 2024-12-02
scx_layered: Implement layer config xllc_mig_min_us

htejun:
Now that per-DSQ queued runtimes are tracked, we can modulate cross-LLC
migrations based on them. Add a per-layer config option to skip cross-LLC
migrations if the queued runtime is below the configured value. Default to
100us which should be a safe value in most cases.

### DIFF
--- a/scheds/rust/scx_layered/src/stats.rs
+++ b/scheds/rust/scx_layered/src/stats.rs
@@ -117,6 +117,8 @@ pub struct LayerStats {
     pub xnuma_migration: f64,
     #[stat(desc = "% migrated across LLCs")]
     pub xllc_migration: f64,
+    #[stat(desc = "% migration skipped across LLCs due to xllc_mig_min_us")]
+    pub xllc_migration_skip: f64,
     #[stat(desc = "% wakers across layers")]
     pub xlayer_wake: f64,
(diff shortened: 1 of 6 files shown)


hodgesds (approved):
Might need to appease the linters, otherwise LGTM.

JakeHillion (approved):
Nice! Looking forward to trying this out. Will close #944 which attempted something similar but is way out of date.

scheds/rust/scx_layered/src/stats.rs
```diff
@@ -301,12 +304,13 @@ impl LayerStats {
 
         writeln!(
             w,
-            "  {:<width$}  open_idle={} mig={} xnuma_mig={} xllc_mig={} affn_viol={}",
+            "  {:<width$}  open_idle={} mig={} xnuma_mig={} xllc_mig/skip={}/{} affn_viol={}",
```
hodgesds, line 307: the `skip` might get confusing down the road, but I think it's fine for now as long as numa/llc related states are grouped together.

htejun:
> Nice! Looking forward to trying this out. Will close #944 which attempted something similar but is way out of date.

We'll probably need a better way to configure this than an absolute limit. Hopefully, we'll have a better idea on what that should be once we gain more experience with this.

Issue #783, opened 2024-10-11
scx_lavd: improve --help-stats option


multics69:
Currently "--help-stats" prints only the the description of `SysStats`.
It should prints also `SchedSample` defined in `stats.rs`

```
$>scx_lavd  --help-stats                                                                                                                                                                                                                                                     
[SysStats]
  avg_svc_time   (u64)    : Average runtime per schedule
  mseq           (u64)    : Sequence ID of this messge
  nr_active      (u64)    : Number of active CPUs when core compaction is enabled
  nr_queued_task (u64)    : Number of runnable tasks in runqueues
  nr_sched       (u64)    : Number of context switches
  pc_balanced    (float)  : % of balanced mode
  pc_big         (float)  : % of tasks scheduled on big cores
  pc_greedy      (float)  : % of greedy tasks
  pc_lc          (float)  : % of latency-critical tasks
  pc_lc_on_big   (float)  : % of latency-critical tasks scheduled on big cores
  pc_lhp         (float)  : % lock holder preemption
  pc_migration   (float)  : % of task migration
  pc_pc          (float)  : % of performance-critical tasks
  pc_pc_on_big   (float)  : % of performance-critical tasks scheduled on big cores
  pc_performance (float)  : % of performance mode
  pc_powersave   (float)  : % of powersave powersave
  pc_preemption  (float)  : % of task preemption
  power_mode     (string) : Current power mode

```

multics69:
Thanks @luigidematteis for the work.

PR #1030, opened 2024-12-02
scx_stats,scx_lavd: describe non-nested stats meta

luigidematteis:
In cases where non-nested structs are used for statistics (such as in `scx_lavd`), the `describe_metadata` method cannot display metadata for all of them. This is because the current recursion mechanism does not directly support multiple monitoring modes that use different, unrelated structs.

In `scx_lavd`, two distinct monitoring modes rely on separate structs. Therefore, it was necessary to create a wrapper to utilize the existing `describe_metadata` method in a way that accommodates both these separate statistics.

This makes it possible to display the metadata description either selectively (in cases where we want to have separate flags to display the metadata of different statistics) or for `n` statistics simultaneously, leaving untouched the way `server_data` provides data to initialize the server, which requires all statistics to be included in the same `StatsServerData`, 

The issue is reported in [sched-ext#783](https://github.com/sched-ext/scx/issues/783)


rust/scx_stats/src/server.rs
```diff
@@ -280,6 +280,27 @@ where
         self.visit_meta(self.top.as_ref().unwrap(), &mut |_| Ok(()))
     }
 
+    pub fn describe_nstats_meta<W: Write>(
```
htejun, line 283: I'd just update `describe_meta()` to take array instead.

rust/scx_stats/src/server.rs
```diff
@@ -280,6 +280,27 @@ where
...
+        &self,
+        w: &mut W,
+        stats_meta_names: Option<&[&str]>,
+    ) -> Result<()> {
+        let meta_names = match stats_meta_names {
+            Some(v) => v,
+            None => {
+                return Err(anyhow!(
+                    "Expected stats meta names to look up in server data"
```
htejun, line 292: Especially given that this function doesn't accept `None` as input and most existing users of `describe_meta()` don't specify `from`.

htejun (changes requested):

htejun (approved):

PR #1032, opened 2024-12-03
scx_utils: Extend topology to support Snapdragon X Elite

multics69:
Snapdragon X Elite processor has an unusual CPU and cache topologies. It does not have an L3 cache, and a CPU cluster with four CPUs shares the same L2 cache -- so the L2 cache is the LLC. Moreover,  CPUs in a single core span across multiple L2 domains. Lastly, the sysfs does not exports an cache id file. Specifically, it has the following topology from the scx's point of view:

```
  LLC0 - core 0 - cpu 0
       \ core 1 - cpu 1
       \ core 2 - cpu 2
       \ core 3 - cpu 3

  LLC1 - core 0 - cpu 4
       \ core 1 - cpu 5
       \ core 2 - cpu 6
       \ core 3 - cpu 7

  LLC2 - core 0 - cpu 8
       \ core 1 - cpu 9
       \ core 2 - cpu 10
       \ core 3 - cpu 11
```

To support the above architecture, changed three things:
- When there is no L3 cache, use L2 cache ID as LLC ID.
- When there is no cache id file under sysfs, assign an id based on `shared_cpu_list` of the cache.
- Allow the case that a core ID is duplicated across LLC domains.

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -79,6 +79,7 @@ use glob::glob;
 use sscanf::sscanf;
 use std::collections::BTreeMap;
 use std::path::Path;
+use std::path::PathBuf;
 use std::slice::Iter;
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


hodgesds:
On the `Core` struct does it need to have a `llc_ids` field now instead of a single field? What kind of machine did you end up getting for testing?

multics69:
> On the `Core` struct does it need to have a `llc_ids` field now instead of a single field? What kind of machine did you end up getting for testing?

I am using a Thinkpad T14s Gen6 Snapdragon laptop with Ubuntu.

hodgesds (approved):
LGTM

multics69:
> The change looks fine to me but what does it mean for a core to span multiple caches? How does locality actually work out?

I don't know exactly. I guess it would be a bug of information exposed by a hardware?

PR #1034, opened 2024-12-03
layered: actually get slice_ns and max_exec_ns from CO:RE

JakeHillion:
Layered previously hardcoded the defaults for slice_ns and max_exec_us. This
was changed in a recent PR to get these from CO:RE in the running kernel. This
change doesn't work because the old values are still set a few lines below.

Check for the presence of the CLI options and default to the CO:RE values
otherwise. Remove setting the layer struct's slice_ns as we check if it's equal
to 0 anyway and default to the global in the BPF code.

Test plan:
- CI

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -401,15 +401,15 @@ lazy_static! {
 #[command(verbatim_doc_comment)]
 struct Opts {
     /// Scheduling slice duration in microseconds.
-    #[clap(short = 's', long, default_value = "20000")]
-    slice_us: u64,
+    #[clap(short = 's', long)]
+    slice_us: Option<u64>,
 
     /// Maximum consecutive execution time in microseconds. A task may be
     /// allowed to keep executing on a CPU for this long. Note that this is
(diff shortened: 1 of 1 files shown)


htejun (approved):

etsal:
I think ideally we should have --slice-us and --max-exec-us to not have a default and only override the system default if set. The current code definitely does not use the builtin values because there is both a default and a possibly user-defined value provided from the scheduler, and it makes sense for both to override the system default. I think it makes sense to removes the scheduler-provided defaults and use the system ones instead, but the current code AFAICT breaks the --slice-us and --max-exec-us switches.

JakeHillion:
> I think ideally we should have --slice-us and --max-exec-us to not have a default and only override the system default if set. The current code definitely does not use the builtin values because there is both a default and a possibly user-defined value provided from the scheduler, and it makes sense for both to override the system default. I think it makes sense to removes the scheduler-provided defaults and use the system ones instead, but the current code AFAICT breaks the --slice-us and --max-exec-us switches.

Ah sure, you're right! Will update and test this properly. Spotted going through that the current behaviour (setting it twice) clearly wasn't what was intended and thought it was an easy fix.

PR #1035, opened 2024-12-03
Cargo: move dependencies into workspace

ruihe774:
This PR moves dependencies of individual Cargo packages into Cargo workspace, so that shared dependencies can be managed at the same place. This PR also upgrades dependencies.


ruihe774:
cc @likewhatevs @JakeHillion

likewhatevs:
Not sure how I feel about this tbh. There are pros and cons to either way. I think keeping dependencies separate makes things simpler for each scheduler, idk.

ruihe774:
> Not sure how I feel about this tbh. There are pros and cons to either way. I think keeping dependencies separate makes things simpler for each scheduler, idk.

It is not the case that different schedulers depend on different versions. They all depend on the same versions. I consider them as code duplications.

htejun:
Up until now, each scheduler subdirectory could be separated out easily. There was no cross-scheduler interlocking and the only thing the higher level workspace did was making it more convenient to build / package and so on. This PR creates dependency across schedulers and each scheduler subdirectory is no longer self-contained.

I understand the motivation but with `version-tool.py`, it's not that difficult to manage dependency versions and I think it's useful to retain the ability for each scheduler to depend on different versions of the same crate if necessary for whatever reason.

PR #1036, opened 2024-12-03
scx: set rust to stable

likewhatevs:
set rust version to stable and remove rustfmt.toml files (these were all the same and were the reason we were depending on nightly rust afaik).

also run fmt so lint job passes


htejun (approved):

ruihe774 (changes requested):
IMO it'll be better to also set the Rust toolchain used in CI to stable.

likewhatevs:
> IMO it'll be better to also set the Rust toolchain used in CI to latest stable.

this is the toolchain used in CI.

ruihe774 (approved):

likewhatevs:
Actually, I think I see what you are saying, let me explain.

So like, rust-toolchain ensures that for our code, the rust toolchain used will be the specified one.

On ci, when nightly is set to default, it makes everything else outside of that "just work". For example, doc generation or installing system-ish packages which may depend upon that. That's why there is rustup default nightly in the dependency install script.

ruihe774:
> Actually, I think I see what you are saying, let me explain.
> 
> So like, rust-toolchain ensures that for our code, the rust toolchain used will be the specified one.
> 
> On ci, when nightly is set to default, it makes everything else outside of that "just work". For example, doc generation or installing system-ish packages which may depend upon that. That's why there is rustup default nightly in the dependency install script.

Do we ensure that the code can be built with stable in CI? If there is already such test, I think it's okay.

PR #1040, opened 2024-12-04
scx_layered: Refactor userspace layer cpumask update logic

hodgesds:
Create separate helper functions for growing/shrinking layers and add some various helper functions. I'm still debugging some of the layer growth issues, but this make it a little easier to follow the code.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -1041,6 +1041,13 @@ impl Layer {
         self.nr_cpus += nr_new_cpus;
         Ok(nr_new_cpus)
     }
+
+    fn is_open(&self) -> bool {
+        return match self.kind {
+            LayerKind::Open { .. } => true,
+            _ => false,
+        };
+    }
 }
 
 struct Scheduler<'a> {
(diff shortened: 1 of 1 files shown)


likewhatevs (approved):
LGTM

likewhatevs:
also, misc btw, when refactoring I'd lean towards whatever the name of "have impl's on struct that wrap calls to standalone fn's that take everything as args" because that makes adding tests later on simpler.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1966,54 +1970,78 @@ impl<'a> Scheduler<'a> {
...
+    }
+
+    fn refresh_cpumasks(&mut self) -> Result<()> {
+        let targets = self.calc_target_nr_cpus();
+        let targets = self.weighted_target_nr_cpus(&targets);
+
+        let mut ascending: Vec<(usize, usize)> = targets.iter().copied().enumerate().collect();
+        ascending.sort_by(|a, b| a.1.cmp(&b.1));
+
+        let shrunk_layers = self.shrink_layers(&ascending, &targets)?;
```
htejun, line 2039: shrank?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1966,54 +1970,78 @@ impl<'a> Scheduler<'a> {
...
+        let _ = prog.test_run(input);
+    }
 
-                // Open layers need the intersection of allowed cpus and
-                // available cpus.
-                layer.cpus.copy_from_bitslice(&available_cpus);
-                layer.nr_cpus = nr_available_cpus;
-                Self::update_bpf_layer_cpumask(layer, bpf_layer);
+    /// Distributes remaining CPUs across open layers.
+    fn grow_open_layers(&mut self) -> Result<()> {
```
htejun, line 1986: This is a misnomer. The function doesn't just grow open layers. It also performs all the post-update operations including triggering BPF side updates and recording layer cpus_range and so on. It also can shrink open layers -ie. the right verb would be something like `fit`. It probably would make sense to split this into two separate functions.

htejun (changes requested):
Left a comment around `grow_open_layers()`. Personally, I'm not a huge fan of splitting sequential procedural function into multiple. There's nothing repeating or being reused here. It's just a step by step thing where context from previous steps may be useful for following steps. Putting function boundaries seem to just add friction around passing around context. That said, this may be a minority opinion, and splitting is fine too.

scheds/rust/scx_layered/src/main.rs
JakeHillion: Apparently "shrank" is the past simple in UK English (https://dictionary.cambridge.org/dictionary/english/shrank) while "shrunk" is the past participle (https://dictionary.cambridge.org/dictionary/english/shrunk). Fortunately shrunk is both in US English so I don't have to learn what the difference is :)

htejun: Oh, so the next line is:
```
        let grew_layers = self.grow_non_open_layers(&ascending)?;
```
It's a bit odd for one to be pp and the other p.

Issue #1039, opened 2024-12-04
scx_lavd: cpuc's stopping_tm_est_ns and lat_cri not updated when schedule to real time schedule class


archerbroler:
cpuc's stopping_tm_est_ns and lat_cri is used in preemption compare. But its not cleared when schedule to a real time task(only idle case cleared), this make task can preemption compare to a current running real time task with last scx task's preemption value, which is not correct.

multics69:
What a good catch! Than you! I will fix it today or tomorrow.

multics69:
BTW, @archerbroler , did you have a chance to test recent improvement on load balancing on your hardware set up?

https://github.com/sched-ext/scx/pull/1000
https://github.com/sched-ext/scx/pull/1032

archerbroler:
oh, This is great. 
I have not update my code to your latest NUMA change, instead I prefer to use NUMA binding for all NUMA aware need. Which I dont think will by accepted by LAVD.
I just found your task steal code, but I think what I need is threads in a process do its best to stay in a same NUMA node, not NUMA balance. NUMA balance should only be handled as a starving like case.
BTW, I think most NUMA machine will run server as k8s like with a  In NUMA cpuset. Recently hardware will share IO die with multiple CPU die, but no direct link between CPU dies. Too many communication between multiple CPU die in such CPU is not acceptble.
I found in your commit you are testing Snapdragon X Elite processor, I think this hardware should not do balance too. These clusters is more or little like a sub-NUMA. Ampere also has a no L3 cache CPU.
I dont't like the whole Numa balance idea. But I also has no better idea to handle this except NUMA binding. If the task load is calculated accurate enough, predict place may be a way. But that's a long way to go.

PR #1041, opened 2024-12-04
scx_lavd: Reset per-CPU preemption information when a CPU is released

multics69:
When a CPU is released to serve a higher-priority scheduler class, reset per-CPU preemption information of that CPU so not choose the CPU as a victim by mistake. This solves the issue https://github.com/sched-ext/scx/issues/1039


hodgesds (approved):
Makes sense. LGTM

multics69:
Thanks @hodgesds !

PR #1044, opened 2024-12-04
scx_layered: Tweak usage decay constant

hodgesds:
Tweak the usage decay to be slightly less aggressive. This was causing layers not to grow properly. The strange thing is that on `main` it's not always reproducible, so maybe the decay is just enough to prevent layers from growing.

`main`:
image
`pr`:
image

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -86,7 +86,7 @@ const NR_LLC_LSTATS: usize = bpf_intf::llc_layer_stat_id_NR_LLC_LSTATS as usize;
 const NR_LAYER_MATCH_KINDS: usize = bpf_intf::layer_match_kind_NR_LAYER_MATCH_KINDS as usize;
 
 lazy_static! {
-    static ref USAGE_DECAY: f64 = 0.5f64.powf(1.0 / USAGE_HALF_LIFE_F64);
+    static ref USAGE_DECAY: f64 = 0.2f64.powf(1.0 / USAGE_HALF_LIFE_F64);
     static ref EXAMPLE_CONFIG: LayerConfig = LayerConfig {
         specs: vec![
             LayerSpec {


JakeHillion (changes requested):
I think this is the wrong number to change. I believe the `USAGE_DECAY` here is converting the half life to the loopback window, and by memory that's (1/2)^(1/r) where r is the halflife. The intention here is probably to change `USAGE_HALF_LIFE` instead which I believe is a BPF constant.

Feel free to correct me if I've misunderstood!

hodgesds:
> I think this is the wrong number to change. I believe the `USAGE_DECAY` here is converting the half life to the loopback window, and by memory that's (1/2)^(1/r) where r is the halflife. The intention here is probably to change `USAGE_HALF_LIFE` instead which I believe is a BPF constant.
> 
> Feel free to correct me if I've misunderstood!

Let me try that, that seems reasonable.

hodgesds:
I wonder if the decay needs to be based on the per layer timeslice, something like `layer.slice_ns * 3` or something. With a 20ms slice the default of [100ms](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/intf.h#L40) is pretty reasonable, but if the slice is smaller maybe it doesn't decay as well.

likewhatevs:
> I wonder if the decay needs to be based on the per layer timeslice, something like layer.slice_ns * 3 or something. With a 20ms slice the default of [100ms](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/intf.h#L40) is pretty reasonable, but if the slice is smaller maybe it doesn't decay as well.

if it isn't clear which is best, I'd say add a flag until we know.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -86,7 +86,7 @@ const NR_LLC_LSTATS: usize = bpf_intf::llc_layer_stat_id_NR_LLC_LSTATS as usize;
 const NR_LAYER_MATCH_KINDS: usize = bpf_intf::layer_match_kind_NR_LAYER_MATCH_KINDS as usize;
 
 lazy_static! {
-    static ref USAGE_DECAY: f64 = 0.5f64.powf(1.0 / USAGE_HALF_LIFE_F64);
+    static ref USAGE_DECAY: f64 = 0.2f64.powf(1.0 / USAGE_HALF_LIFE_F64);
```
likewhatevs, line 89: pulled out calculator btw, I think `f64.powf(1.0 / USAGE_HALF_LIFE_F64)` is maybe always just 1 btw.

JakeHillion: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=172222a94132f6336d0fedb6db887745

Doesn't look like 1 to me.

likewhatevs: I missed the divide by 1 billion while grepping for the consts involved, but tl;dr, as that is, it's a complicated way of writing `0.5*1^(1/(100 million/1 billion))`

JakeHillion: Yeah, the HALF_LIFE_F64 has a pretty bad name given it's not just HALF_LIFE as f64! Would probably be better if they were named `HALF_LIFE_NS` and `HALF_LIFE_S_F64` or something.

PR #1045, opened 2024-12-04
introduce common BPF library for C schedulers

etsal:
Add to the build system the option to link a common library of BPF functions into a scheduler. This patch includes the arena-based allocator as the first component of this common library, and will automatically add into the library any code added inside the lib/ directory.

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -2,27 +2,23 @@ libs = ['sdt_task']
 
 objs = []
 
-foreach lib : libs
-  bpf_o = gen_bpf_o.process(lib + '.bpf.c', extra_args: bpf_includes)
-  tgt = custom_target(lib,
-    output: lib + '.bpf.o',
-    input: lib + '.bpf.c',
-    command: [bpf_clang, bpf_base_cflags, bpf_includes, '-target', 'bpf',
-        libbpf_c_headers, '-c', '@INPUT@', '-o', '@OUTPUT@'],
-    depends: [libbpf],
-    build_by_default: true,
-    build_always_stale: true,
-  )
+foreach src: libs
+  # Copy of gen_bpf_o(), because custom targets and generators don't mix
+  bpf_o = custom_target(src,
+            output: src + '.bpf.o',
+            input: src + '.bpf.c',
+            command: [bpf_clang, bpf_base_cflags, '-target', 'bpf', libbpf_c_headers, bpf_includes,
+                        '-c', '@INPUT@', '-o', '@OUTPUT@'],
+            depends: [libbpf]
+            )
 
-  objs += [tgt]
+  objs += [bpf_o]
 endforeach
 
-custom_target(scx_lib_name,
+scx_lib = custom_target(scx_lib_name,
   output: scx_lib_name + '.bpf.o',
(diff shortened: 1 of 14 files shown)


scheds/c/scx_flatcg.bpf.c
htejun, line 1: For schedulers which are also carried in the kernel tree under `tools/sched_ext`, I don't think we want to add dependencies. Let's leave them alone.

htejun (approved):

lib/meson.build
```diff
@@ -0,0 +1,29 @@
...
+    depends: [libbpf],
+    build_by_default: true,
+    build_always_stale: true,
+  )
+
+  objs += [tgt]
+endforeach
+
+custom_target(scx_lib_name,
+  output: scx_lib_name + '.bpf.o',
```
hodgesds, line 21: Would this work if there are two separate `util.bpf.c` files?

hodgesds (approved):

lib/meson.build
etsal: Unfortunately no, the library compiles the files specified in a target list at the top of the lib/meson.build and Meson does not allow us including paths in that list. To do that we would have to add subdirectories in lib/.

arighi:
Trying to build a single scheduler (i.e., `meson compile -C build scx_simple`) seems to fail with the following error:
```
libbpf: failed to open file '/home/arighi/src/scx/build/lib/lib.bpf.o': -ENOENT
Error: failed to link '/home/arighi/src/scx/build/lib/lib.bpf.o': No such file or directory (2)
```

I haven't investigated much, maybe the scheds are missing an explicit dependency on the new libs?

lib/sdt_task.bpf.c
```diff
@@ -269,7 +271,7 @@ static SDT_TASK_FN_ATTRS int sdt_pool_set_size(struct sdt_task_pool __arena *poo
 }
 
 /* initialize the whole thing, maybe misnomer */
-static SDT_TASK_FN_ATTRS int sdt_task_init(__u64 data_size)
+__hidden SDT_TASK_FN_ATTRS int sdt_task_init(__u64 data_size)
```
multics69, line 274: I am confused with the following code. The code seems to return with the lock on failure (-ENOMEM) not on success. Am I missing something? Also, in case of error (or success), why is it reasonable to hold the lock?

```.c
	/* On success, returns with the lock taken. */
	ret = sdt_alloc_attempt(prealloc_stack);
```

etsal: The call unlocks the spinlock on all error paths this is also largely enforced by the verifier. The reason ```sdt_alloc_attempt```  keepsthe lock on success is because success means the preallocation stack has enough data to satisfy all the memory needs of the ```sdt_task_allloc``` without having to make a sleepable arena allocation. Dropping the lock would invalidate this guarantee.

multics69: One question. Many scx schedulers use `cpu_ctx`. Do you plan to extend the current task slab allocator to a bit more general slab allocator covering both task and CPU?

etsal: Yes, the plan is to evolve the allocator towards being able to define multiple suballocators. It should be a matter of adding an extra layer of indirection in the code, but I went for a simpler approach for starters.

PR #1046, opened 2024-12-05
scx_layered: Fix vtime and llc_id handling bugs

htejun:
Fix:

- vtime could go wildly wrong.
- queue_runtime tracking was broken for LLC IDs >= 16  because MAX_LAYERS was incorrectly used as the invalid value instead of MAX_LLCS.


hodgesds (approved):
Good catch!

Issue #721, opened 2024-10-02
topology: Use cpumasks to iterate all possible or online CPUs


htejun:
CPU IDs aren't guaranteed to be consecutive. To iterate all possible or currently online CPUs, the respective cpumask should be iterated. There are several issues in this area:

- Some are iterating 0 to the number of online CPUS. This is incorrect as CPU IDs can have gaps and thus can miss higher numbered CPUs.
- Some are iterating 0 to `nr_cpu_ids`. This is correct but not ideal in that the CPU IDs can have substantial gaps not online or unallocated.
- The possible and online cpumasks can be accessed from BPF side using `scx_bpf_get_possible_cpumask()` and `scx_bpf_get_online_cpumask()`, respectively. However, `scx_utils::toplogy` doesn't expose the current possible or online CPU masks that are available through `/sys/devices/system/cpu/possible` and `online`.

Make `scx_utils::topology` expose the cpumasks and make all schedulers use the appropriate cpumask for iteration in both userspace and BPF code.

uran0sH:
If no one works on it, I would like to work.

PR #1047, opened 2024-12-05
scx_utils: expose cpus_online vec

luigidematteis:
Currently, `scx_utils::Topology` already exposes the number of CPUs that may be onlined via `nr_cpu_ids`.

By providing a vector that contains all the online CPUs, we enable schedulers to iterate only over the effectively available CPUs.

Related to #721

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -183,7 +183,7 @@ pub struct Topology {
     /// Cpumask all CPUs in the system.
     pub span: Cpumask,
     pub nr_cpus_online: usize,
-
+    pub cpus_online: Vec<usize>,
     /// Skip indices to access lower level members easily.
     pub all_llcs: BTreeMap<usize, Arc<Llc>>,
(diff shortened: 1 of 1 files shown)


rust/scx_utils/src/topology.rs
```diff
@@ -238,10 +238,12 @@ impl Topology {
         }
 
         let nr_cpus_online = span.weight();
+        let cpu_online_ids = cpu_online_ids(&span);
         Ok(Topology {
             nodes,
             span,
             nr_cpus_online,
+            cpu_online_ids,
```
htejun, line 246: Can't it just be Cpumask?

htejun: Also, can you name it `cpus_online`?

luigidematteis: Hi @htejun,

the idea was to have a precomputed value that could be used instead of iterating over `nr_cpu_ids` and then calling `test_cpu()`. 

However, the Cpumask is already sufficient for this. 

I'll close this draft.

PR #1048, opened 2024-12-05
scx_layered: Fix idle cpu iteration

hodgesds:
Fix a bug in idle CPU selection where all possible CPUs were being considered.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -987,7 +987,7 @@ void try_preempt(s32 task_cpu, struct task_struct *p, struct task_ctx *taskc,
 
 	pmap = &task_cpuc->prox_map;
 
-	bpf_for(i, 1, MAX_CPUS) {
+	bpf_for(i, 1, nr_possible_cpus) {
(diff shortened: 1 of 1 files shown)


likewhatevs (approved):

etsal (approved):
Makes sense, do we know why this was causing the bug? Was it that the loop was too long, and making preemption slow, or maybe reading the blank entries did something? 

Also, for the ```(i >= pmap->sys_end)``` check below: i is used as an index in the CPU list, but ```sys_end``` seems to be hte amount of LLCs in the topology. Is this expected?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -987,7 +987,7 @@ void try_preempt(s32 task_cpu, struct task_struct *p, struct task_ctx *taskc,
 
 	pmap = &task_cpuc->prox_map;
 
-	bpf_for(i, 1, MAX_CPUS) {
+	bpf_for(i, 1, nr_possible_cpus) {
 		if (i >= pmap->sys_end)
```
hodgesds, line 991: Maybe this isn't as useful as this should get set properly.

PR #1049, opened 2024-12-05
scx: add ci job to test rolling stable

likewhatevs:
scx: add ci job to test rolling stable

this should run every 4 hours

will merge once I see this run on my fork, should just work (err well just fail because I think this is broke atm).

@ptr1337

### DIFF
--- a/.github/workflows/stable.yml
+++ b/.github/workflows/stable.yml
@@ -1,0 +1,158 @@
+name: stable-test
+
+on:
+  # only runs on main, every 6 hours
+  schedule:
+    - cron: "0 */6 * * *"
+
+jobs:
+  build-kernel:
+    runs-on: ubuntu-24.04
+    steps:
+      # prevent cache permission errors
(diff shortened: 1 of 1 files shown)


ptr1337:
Thanks for adding this!
I think it would be quite good to test again the stable-rc branch, or?

arighi (approved):

arighi:
Thanks for doing this, feel free to merge when it works

likewhatevs:
> I think it would be quite good to test again the stable-rc branch, or?

tbh I just picked the one that sounded right, let me look into what the branches are and confirm it is the right one. we want to build against the equivalent of for-next so that we are aware of breakages before they get cut into a release.

likewhatevs:
I'm not certain there's anything better to use than `linux-rolling-stable`, but am open to suggestions if anyone has one.

Looking at:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git

It looks like commits are pushed the same time as they are to `linux-stable-rolling`.

htejun (approved):

PR #1050, opened 2024-12-05
scx_layered: Extend confined preempting checks

hodgesds:
When a confined layer is determining whether or not a task should preempt it should also consider if the layer is able to grow. Add an additional check on the number of idle CPUs when determinig if confined layers should preempt.


etsal (approved):

htejun:
I'm not sure the explanation matches the code change. `keep_running()` is only used to decide whether to keep running the current task instead of switching to a task which is waiting in one of the DSQs. Can you explain the rationale behind the change in more details?

hodgesds:
I think with latest `main` this is no longer an issue.

PR #1052, opened 2024-12-06
scx: fix bug in llc mask creation

likewhatevs:
When creating an llc mask for a node by iterating all cpus, OR llc_mask with the new value instead of AND'ing it.


hodgesds (approved):
That mask might not be used anymore, but seems right.

PR #1062, opened 2024-12-06
scx_layered: Fix a silly bug in `select_cpu()` and implement per-layer fifo mode

htejun:
- Fix a silly bug critical bug in `select_cpu()`.
- Fix layer sizing when weights are not specified.
- Implement per-layer fifo mode.


likewhatevs (approved):
LGTM

PR #1066, opened 2024-12-07
scx_bpfland: do not rely on scx_utils::autopower

arighi:
Relying on scx_utils::autopower reuires power-profiles-daemon running, this has introduced a regression in certain systems that don't have this service available.

Restore the old behavior for now, we will fix scx_utils::autopower later.

Fixes: 939d4f5 ("scx_utils: move autopower into scx_utils")


hodgesds (approved):
LGTM, I think once we have multiple compilation units this could be done with a few probes + rust library instead.

PR #1070, opened 2024-12-07
scx_loader: provide library crate to be used by other crates

vnepogodin:
provide library crate to be used by other crates, e.g CachyOS KernelManager

also provides Rust typed-interface for schedulers, with changed error for callers:
```
❯ dbus-send --system --print-reply --dest=org.scx.Loader /org/scx/Loader org.scx.Loader.SwitchScheduler string:scx_lvd uint32:2
Error org.freedesktop.zbus.Error: unknown variant `scx_lvd`, expected one of `scx_bpfland`, `scx_rusty`, `scx_lavd`, `scx_flash`
```
instead of
```
❯ dbus-send --system --print-reply --dest=org.scx.Loader /org/scx/Loader org.scx.Loader.SwitchScheduler string:scx_lvd uint32:2
Error org.freedesktop.DBus.Error.Failed: scx_lvd is not supported
```

### DIFF
--- a/rust/scx_loader/Cargo.toml
+++ b/rust/scx_loader/Cargo.toml
@@ -19,3 +19,10 @@ tokio = { version = "1.39", features = ["macros", "sync", "rt-multi-thread", "pr
 toml = "0.8.19"
 zbus = { version = "5", features = ["tokio"], default-features = false }
 zvariant = "5.1"
+
+[lib]
+path = "src/lib.rs"
+
+[[bin]]
+name = "scx_loader"
+path = "src/main.rs"
--- a/rust/scx_loader/src/config.rs
+++ b/rust/scx_loader/src/config.rs
@@ -11,12 +11,12 @@ use std::path::Path;
 
 use anyhow::Result;
 use serde::Deserialize;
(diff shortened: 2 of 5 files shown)


rust/scx_loader/src/utils.rs
```diff
@@ -0,0 +1,18 @@
...
+//
+// Copyright (c) 2024 Vladislav Nepogodin <vnepogodin@cachyos.org>
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use crate::SupportedSched;
+
+/// Get the scx name from the given scx trait
+pub fn get_name_from_scx(supported_sched: &SupportedSched) -> &'static str {
```
hodgesds, line 11: silly question, could this function be an `impl` method on the `SupportedSched` enum?

htejun (approved):

rust/scx_loader/src/utils.rs
vnepogodin: I think implementing Into/IntoStr would be more native

hodgesds:
Nice!

PR #1067, opened 2024-12-07
meson: Add scx_lib dep to gen_bpf_o

frelon:
This PR introduces a dependency on scx_lib in gen_bpf_o.

Without this dependency the build fails intermittently with the
following error:

```
FAILED: scheds/c/scx_pair.p/scx_pair.bpf.skel.h scheds/c/scx_pair.p/scx_pair.bpf.subskel.h
/home/frelon/src/scx/meson-scripts/bpftool_build_skel /usr/sbin/bpftool scheds/c/scx_pair.p/scx_pair.bpf.o scheds/c/scx_pair.p/scx_pair.bpf.skel.h scheds/c/scx_pair.p/scx_pair.bpf.subskel.h /home/frelon/src/scx/build/lib/lib
libbpf: failed to get ELF header for /home/frelon/src/scx/build/lib/lib.bpf.o: invalid `Elf' handle
/home/frelon/src/scx/meson-scripts/bpftool_build_skel: line 25: 3400506 Segmentation fault      (core dumped) "$bpftool" gen object "$stem".l1o "$input" "$lib".bpf.o
```

Also updates a usage of bpftool to use the passed in path.

PR #1071, opened 2024-12-08
scx_lib: use the proper bpftool

arighi:
Make sure to use the right bpftool when building scx_lib, instead of always relying on the system's bpftool.

Fixes: 3d72a8e ("add library infrastructure and turn allocator into a library")

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -9,7 +9,7 @@ foreach src: libs
             input: src + '.bpf.c',
             command: [bpf_clang, bpf_base_cflags, '-target', 'bpf', libbpf_c_headers, bpf_includes,
                         '-c', '@INPUT@', '-o', '@OUTPUT@'],
-            depends: [libbpf]
+            depends: [libbpf, bpftool_target]
             )
 
   objs += [bpf_o]
--- a/meson-scripts/compile_scx_lib
+++ b/meson-scripts/compile_scx_lib
@@ -9,4 +9,4 @@ dir="$PWD/lib"
 
 libs=`find $dir -type f -name *.bpf.o | grep -v $lib.bpf.o`
 
-"$bpftool" gen object "$output" $libs
+${bpftool} gen object "$output" $libs


likewhatevs (approved):

htejun (approved):

likewhatevs:
btw #1067 (being merged atm) fixes this also

arighi:
> btw #1067 (being merged atm) fixes this also

Hm.. nope, it's still broken:
```
FAILED: lib/lib.bpf.o
/home/nvidia/arighi/src/scx/meson-scripts/compile_scx_lib /home/nvidia/arighi/src/scx/build/bpftool/src/bpftool lib/lib.bpf.o lib lib/sdt_task.bpf.o
/home/nvidia/arighi/src/scx/meson-scripts/compile_scx_lib: line 12: /home/nvidia/arighi/src/scx/build/bpftool/src/bpftool: No such file or directory
```

I think bash doesn't like the double quotes around "$bpftool".

arighi:
Updated on top of #1067, @likewhatevs can you check if this works for you as well?

arighi:
@likewhatevs actually scratch that, it's still not working, something else is broken now... looking.

likewhatevs:
this worked for me without needing to run compile twice?:
```
rm -rf build
meson setup build
meson compile -C build scx_simple
```

arighi:
@likewhatevs I see the problem, scx_lib should depend on bpftool_target. It seems to be fixed now.

likewhatevs:
I tested the latest commit on my system and it worked also (but it worked before for me so maybe there's something non-standard on my system making this always work).

arighi:
> I tested the latest commit on my system and it worked also (but it worked before for me so maybe there's something non-standard on my system making this always work).

It could be also a timing / parallelization issue and in your case it was always building bpftool before any scheduler (or you are using the system's bpftool).

Issue #1072, opened 2024-12-08
scx_layered: subtraction overflow in nr_to_free calculation


luigidematteis:
The variable `nr_to_free` in `scx_layered/src/main.rs` is assigned an incorrect value due to a subtraction overflow occurring in the following code:

```rust
                let nr_freed = layer.free_some_cpus(&mut self.cpu_pool, max_to_free)?;
                if nr_freed == 0 {
                    break;
                }
                nr_to_free -= nr_freed;
                freed = true;
```

### Reproduction steps

1. Building the scheduler without `-Dbuildtype=release` flag
2. Run the scheduler with either of the following commands: 
```
./scx_layered --run-example --disable-topology=true -v
./scx_layered --run-example --disable-topology=false -v
```
The program panics with the following `output`

```
             xlayer_wake= 0.00 xlayer_rewake= 0.00
             keep/max/busy= 0.00/ 0.00/ 0.00 kick= 0.00 yield/ign=85.71/    0 
             open_idle= 0.00 mig= 0.00 xnuma_mig= 0.00 xllc_mig/skip= 0.00/114.3 affn_viol= 0.00
             preempt/first/xllc/xnuma/idle/fail= 0.00/ 0.00/ 0.00/ 0.00/ 0.00/ 0.00
             slice=20ms min_exec=85.71/   0.97ms
             cpus=  2 [  0,  2] 00000c00
             LLC sched%/lat_ms [100.0/   1.61] [ 0.00/   1.46]
thread 'main' panicked at scheds/rust/scx_layered/src/main.rs:1921:17:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
### Details

`nr_to_free` is of type `usize`. When the scheduler is built with `-Dbuildtype=release` flag, Rust will not panic on that, and `nr_to_free` will assign a wrong value:
```
// Added debug logs
07:11:49 [DEBUG] nr_to_free: 1 - nr_freed: 2
07:11:49 [DEBUG] result nr_to_free: 18446744073709551615
07:11:49 [DEBUG] Layer "normal" free some CPUs. Max to free: 18446744073709551615
```

PR #1073, opened 2024-12-08
scx_layered: handle nr_to_free calculation

luigidematteis:
The variable `nr_to_free` in `scx_layered/src/main.rs` is assigned an incorrect value due to a subtraction overflow, as reported in this issue #1072.

Ensure `nr_to_free` is correctly handled to prevent subtraction overflow, adding a check to set `nr_to_free` to 0 if `nr_freed` > `nr_to_free`.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -1904,7 +1904,8 @@ impl<'a> Scheduler<'a> {
                 if nr_freed == 0 {
                     break;
                 }
-                nr_to_free -= nr_freed;
+
+                nr_to_free = nr_to_free.saturating_sub(nr_freed);
                 freed = true;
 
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/main.rs
```diff
@@ -1911,7 +1911,12 @@ impl<'a> Scheduler<'a> {
                 if nr_freed == 0 {
                     break;
                 }
-                nr_to_free -= nr_freed;
+
+                if nr_freed > nr_to_free {
+                    nr_to_free = 0
+                } else {
+                    nr_to_free -= nr_freed;
+                }
```
htejun, line 1919: Can we do `nr_to_free.saturating_sub(nr_freed)`?

htejun (approved):

PR #1076, opened 2024-12-08
scx_layered: Make low fallback DSQs useful and other changes

htejun:
- Implement starvation prevention for low fallback DSQs and use them to execute tasks with custom affinities and from empty layers.
- Simplify owned execution protection.
- Guarantee layer growth of high util_range layers.
- Other changes.


likewhatevs (approved):
LGTM, rly like the stats cleanup

PR #1077, opened 2024-12-08
Fixes for dsq_lat.bt

samuelnair:
- Fix to ensure that we can track dsq id 0
- Cleans up some warnings
- Clean up some maps to avoid them printing out a bunch of output that was not really helpful

### DIFF
--- a/scripts/dsq_lat.bt
+++ b/scripts/dsq_lat.bt
@@ -30,10 +30,10 @@ kprobe:scx_bpf_dispatch_vtime,
 	$task = (struct task_struct *)arg0;
 	$dsq = arg1;
 
-	if ($1 > 0 && $task->tgid != $1) {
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
LGTM, I have another bpftrace script that I'm working on where I just add 1 to the DSQ id to get around handling dsq 0 for zero value map entries. It's kind of annoying but you end up just subtracting 1 when printing and it sort of works.

likewhatevs:
> Longer term, it'd probably make sense to invest a bit more effort into making these scripts more customized to our purposes using e.g. bcc. The default histogram implementation is really not great, DSQ IDs are a lot better printed in hex and so on.

I have no idea how practical this is, but I'm gonna take a swing at writing a rust cli that uses small bpf functions to show info these scripts retrieve in a easier to read manner.

PR #1080, opened 2024-12-09
only compile library when using recent bpftool versions

etsal:
Some scx users have bpftool versions < 7.5.0, causing Meson system to throw errors when linking in the BPF library to schedulers. Add a workaround by only compiling the library when the bpftool is new enough.

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -15,11 +15,13 @@ foreach src: libs
   objs += [bpf_o]
 endforeach
 
-scx_lib = custom_target(scx_lib_name,
-  output: scx_lib_name + '.bpf.o',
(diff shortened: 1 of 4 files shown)


lib/meson.build
```diff
@@ -15,11 +15,13 @@ foreach src: libs
   objs += [bpf_o]
 endforeach
 
-scx_lib = custom_target(scx_lib_name,
-  output: scx_lib_name + '.bpf.o',
-  command: [compile_scx_lib, bpftool_exe_path, '@OUTPUT@', scx_lib_name, objs],
-  depends: objs,
+if bpftool_maj < 7 or (bpftool_maj == 7 and bpftool_min >= 5)
```
hodgesds, line 18: Probably worth adding a comment on why this is needed?

PR #1085, opened 2024-12-09
cpumask, scx_layered: Clean up Cpumask iterator

htejun:
No reason to consume the cpumask for iteration. Replace `.into_iter()` with `.iter()` and use it in layered.


hodgesds (approved):
Nice cleanup!

Issue #1074, opened 2024-12-08
scx schedulers installed from cargo are segfaulting


arighi:
```
ubuntu@ubuntu:~$ sudo ~/.cargo/bin/scx_bpfland
08:40:08 [INFO] scx_bpfland 1.0.8 x86_64-unknown-linux-gnu SMT off
...
Segmentation fault
ubuntu@ubuntu:~$ sudo ~/.cargo/bin/scx_lavd
08:40:13 [INFO] Autopilot mode is enabled by default.
...
Segmentation fault
```

It seems to happen both on Ubuntu and CachyOS.

likewhatevs:
gentoo also (so probably everywhere)

JohnRTitor:
I believe we had the same problem on NixOS (https://github.com/sched-ext/scx/issues/927), it was solved by adding proper include values of compiled libbpf. 

https://github.com/NixOS/nixpkgs/blob/909365d2203fe5570203a5032af3a564b6ab4705/pkgs/os-specific/linux/scx/scx_rustscheds.nix#L37-L48

If that's hard to read, we are just adding include paths of the compiled libbpf. 

When using meson to compile scx, this is done automatically. https://github.com/sched-ext/scx/blob/b4db5ee9f6611c2f853dcd2682995702a42a99cf/meson.build#L307-L377

etsal:
> gentoo also (so probably everywhere)

Yeah can reproduce on arch too.

likewhatevs:
BTW i think the "fix" is folks need to install scx schedulers using `cargo install --locked`. Not the first time I've seen this on a project (iirc zellij does that also), so maybe this is why.

likewhatevs:
something is pulling in `libbpf-sys v1.5.0+v1.5.0` when cargo install runs so I think pinning to versions specified here might fix this: https://github.com/sched-ext/scx/issues/1015

etsal:
I think that's fair as a temporary measure, however we should also check why 1.5.0 is failing because we need it to support multiple compilation units in Rust.

etsal:
Turns out that the versions specified in #1015 are perfectly fine for multiple compilation units, so there is no downside to going with them.

Issue #1015, opened 2024-12-01
Segfault when linked with libbpf-sys version 1.5.0


abrehman94:
When a rust scx scheduler is built with libbpf-sys version 1.5.0+v1.5.0 it causes a seg fault.
```
17:38:01 [INFO] scx_flash 1.0.7-gfa1e12a-dirty x86_64-unknown-linux-gnu/debug SMT off
17:38:01 [INFO] L3 cache ID 1: sibling CPUs: [24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47]
17:38:01 [INFO] L3 cache ID 0: sibling CPUs: [0, 1, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 2, 20, 21, 22, 23, 3, 4, 5, 6, 7, 8, 9]
./clean_build.sh: line 6: 96847 Segmentation fault      sudo ./target/debug/scx_flash
```

As far as I could identify, the previous version that works is 1.4.6.
```
libbpf-rs = "=0.24.8"
libbpf-sys = "=1.4.6"
```

etsal:
HI Abdul, did you happen to look more closely into why this is happening?

abrehman94:
Hi Emil, no I didn't, as far as I saw was the difference on git in v1.4.6 and v1.5.0 of the libbpfsys - couldn't spot anything that would catch my eye.

PR #1087, opened 2024-12-09
scx_utils: specify libbpf-sys to 1.4.6, bump libbpf-rs versions

etsal:
The libbpf-sys package for the Rust schedulers has buggy behavior for versions 1.4.1 and 1.5.0: 1.4.1 crashes when linking multiple BPF objects, while using 1.5.0 causes all Rust schedulers to crash (#1074 #1015). Pin libbpf-sys 1.4.6 to avoid either problem. To use 1.4.6, also update libbbpf-rs to the most recent version.

### DIFF
--- a/rust/scx_utils/src/cpumask.rs
+++ b/rust/scx_utils/src/cpumask.rs
@@ -244,9 +244,28 @@ impl Cpumask {
         new.mask ^= other.mask.clone();
         new
     }
-}
 
-impl Cpumask {
+    /// Iterate over each element of a Cpumask, and return the indices with bits
+    /// set.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use log::info;
+    /// use scx_utils::Cpumask;
+    /// let str = String::from("all");
+    /// let mask = Cpumask::from_str(&str).unwrap();
+    /// for cpu in mask.iter() {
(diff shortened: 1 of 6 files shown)


multics69 (approved):

likewhatevs:
I kind of think we should set all instances of libbpf-rs, libbpf-sys and libbpf-cargo to the same version everywhere.

That being said, I tried doing exactly that (using 1.4.6 iirc) and 1.5.0 was still being pulled in somehow during cargo install (which is like cargo build without a Cargo.lock) despite 1.5.0 being absent from cargo-tree so idk...

etsal:
Did you try ```libbpf-sys = '1.4.6'```, or ```libbpf-sys = '=1.4.6'```? If it was the latter then it's very weird, it's supposed to be pinning the semantic version. The former is just a preferred version so Cargo is allowed to pull a more recent one.

likewhatevs:
> Did you try libbpf-sys = '1.4.6', or libbpf-sys = '=1.4.6'? If it was the latter then it's very weird, it's supposed to be pinning the semantic version. The former is just a preferred version so Cargo is allowed to pull a more recent one.

the latter, after seeing that not work I just added `cargo install --locked` to the readme instead of trying to get `cargo install` without `--locked` to work.

PR #1088, opened 2024-12-10
scripts: Refactor scxtop

hodgesds:
Fix a few bugs with calculations and add dsq vtime.


likewhatevs (approved):
this is both wildly cool and mind-bending. did not know one could do this with bpftrace.

PR #1089, opened 2024-12-10
bump libbpf-rs version to 0.24.8, specify libbpf-sys version 1.4.6+1.4.7

etsal:
This is  a redo of #1087 which was accidentally closed.


likewhatevs:
I think you need to remove the revert commit to make this do what is intended (no changes atm).

likewhatevs (approved):
signals are solid wrt/ what this touches, once green lets do it.

etsal:
> I think you need to remove the revert commit to make this do what is intended (no changes atm).

Oh yeah, I added a revert commit to try to see whether the CI breakage we were seeing was caused by the commit or if it was an issue with the ppas.

likewhatevs:
Ahh btw, because we're using head of the sched_ext repo on kernel.org, main is built hourly to help us see issues like this one (i.e. if you ever suspect that, just check CI status on main, it'll be red if that's the case).

htejun (approved):

PR #1101, opened 2024-12-13
Sync with kernel sched_ext/for-6.14

htejun:
- Update to vmlinux-v6.13-rc1.h
- Apply the API renames.
- Receive a couple fixes from tools/sched_ext.
- Make lib/sdt_task.h inclusion optional.

`scx_rusty`'s `p->mempolicy` optimization is temporarily disabled as the field has been renamed in v6.13. Will address separately.


likewhatevs:
If we're gonna be doing another release soon, I think we want to do the next release before merging this PR.

IIUC, this will break the stable CI job (not 100% sure on how to deal with that yet beyond disabling it) and downstream packaging builds until downstream build systems upgrade to 6.13.

likewhatevs (approved):

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1262,6 +1261,12 @@ static u64 node_dom_mask(u32 node_id)
 static void task_set_preferred_mempolicy_dom_mask(struct task_struct *p,
 						  struct task_ctx *taskc)
 {
+#warning "FIXME - See the comment"
+	/*
+	 * XXX - p->mempolicy no longer exists on v6.13. Disable mempolicy
```
hodgesds, line 1266: I guess this could use `if (bpf_core_field_exists(p->mempolicy))` , but most likely not being used. Maybe we can make a help wanted task to fix it.

hodgesds (approved):
> Apply the API renames.

Hero work, I thought we'd have to do a round of updates... this is nice!

arighi (approved):
Thanks for this work @htejun, I agree with @likewhatevs that it'd be nice to cut a new release before merging this, since everything seems to be pretty stable now and this has the potential to introduce some breakage.

htejun:
This shouldn't break stable CI. The binaries should stay compatible with kernels before the API change, at least theoretically, but yeah I'll hold it off until a new release is cut.

likewhatevs:
> This shouldn't break stable CI. The binaries should stay compatible with kernels before the API change, at least theoretically, but yeah I'll hold it off until a new release is cut.

Oh right, yeah we build with the included vmlinux.h (and not the system's one) so yeah that's right, no breakage.

multics69 (commented):
Thank you @htejun ! Everything looks great except for one thing:
It seems that the size of cpumask in vmlinux should be manually updated like https://github.com/sched-ext/scx/pull/1016

etsal (approved):

multics69 (approved):

PR #1112, opened 2024-12-17
scx_lavd: Optimize autopilot mode for ARM platform

multics69:
In the Snapdraon X Elite platform (Thinkpad T14s Gen6), changing  CPU frequency is more expensive than in x86. This PR optimizes the autopilot implementation to reduce the CPU frequency changes and the overhead of power mode switches.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/power.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/power.bpf.c
@@ -416,10 +416,14 @@ static int reinit_active_cpumask_for_performance(void)
 			goto unlock_out;
 		}
 
-		if (cpuc->big_core)
+		if (cpuc->big_core) {
 			bpf_cpumask_set_cpu(cpu, active);
-		else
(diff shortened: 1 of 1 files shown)


htejun (approved):

hodgesds:
Could you check the values of [`Cpu.trans_lat_ns`](https://github.com/sched-ext/scx/blob/main/rust/scx_utils/src/topology.rs#L118C9-L118C21) on different platforms. I wonder if you could use that as a multiplier on how much to boost performance.

multics69:
> Could you check the values of [`Cpu.trans_lat_ns`](https://github.com/sched-ext/scx/blob/main/rust/scx_utils/src/topology.rs#L118C9-L118C21) on different platforms. I wonder if you could use that as a multiplier on how much to boost performance.

`trans_lat_ns` values are as follows:
- Snapdagon X Elite: 30,000 ns
- Intel Raptor Lake: 0ns
- AMD Ryzen 9: 0ns

PR #1115, opened 2024-12-17
common.bpf.h: Suppress generated prototypes in vmlinux.h

htejun:
The generated helper and kfunc prototypes in vmlinux.h are missing address
space attributes leading to build failures. Until the generation is fixed,
suppress the generated prototypes.


etsal (approved):
awesome!

likewhatevs (approved):
this is pretty nice

htejun:
Turns out we can't do this yet due to address space attributes not being generated. For now, I did 180 and suppressed the autogenerated prototypes. We'll try again some other time.

hodgesds (approved):

PR #1116, opened 2024-12-17
scx_layered: Add stats for cpu release

hodgesds:
Add stats for when the BPF scheduler is released from a CPU. This is to better understand the performance of the scheduler when CPUs are released. Not sure if this is super useful, but running on a test machine showed a handful of counts so probably little overhead in collecting these stats.

`scx_layered --run-example --stats 1` output:
```
###### Tue, 17 Dec 2024 14:39:47 -0800 ######
tot=  16171 cpus_released=1 local=76.62 open_idle= 1.61 affn_viol= 0.00 hi/lo= 1.26/ 0.00 proc=5ms nodes=2
busy=  2.2 util/hi/lo=  172.8/ 0.35/ 0.00 fallback_cpu/util=  1/ 0.0
excl_coll=0.00 excl_preempt=0.00 excl_idle=0.00 excl_wakeup=0.00
```


etsal (approved):
Looks reasonable, I assume the lookup is most likely on per-cpu data?

scheds/rust/scx_layered/src/stats.rs
```diff
@@ -547,8 +551,9 @@ impl SysStats {
     pub fn format<W: Write>(&self, w: &mut W) -> Result<()> {
         writeln!(
             w,
-            "tot={:7} local={} open_idle={} affn_viol={} hi/lo={}/{} proc={:?}ms nodes={}",
+            "tot={:7} cpus_released={} local={} open_idle={} affn_viol={} hi/lo={}/{} proc={:?}ms nodes={}",
```
hodgesds, line 554: kind of hate the name, but couldn't think of something better

htejun:
Isn't this already captured by `LSTAT_ENQ_REENQ`?

hodgesds:
> Isn't this already captured by `LSTAT_ENQ_REENQ`?

Yeah, that's a more discreet counter. I guess the other benefit of `LSTAT_ENQ_REENQ` is that it maps to the layer.

hodgesds:
Probably don't need this, but was interesting to observe.

PR #1118, opened 2024-12-19
[RFC] scx_flash: Favor prev task more in dispatch

vax-r:
## Summary
Originally we choose a new task to run each time we trigger "flash_dispatch", unless there's a task holding user-space lock. This leads to enormous triggers of "flash_dispatch" when we encounter heavy workloads, because each time the prev task may still be able to run, but instead we choose another task to run and let prev task wait for more.

We know shorter runtime tasks ,which are the ones we wish to prioritize, are the ones who tends to still be able to run, putting them back to origin CPU to let them finish faster and those CPU won't have to trigger "flash_dispatch" so frequently when we're dealing with heavy workloads.

## Tests
We utilize postgreSQL with pgbench to generate heavy workloads for the system and monitor the latency for queries, TPS, and number of transactions actually processed.
```
$ pgbench -h localhost -p 5432 -U postgres -c 100 -j 16 -T 120 -M prepared --protocol extended testdb
```
We compare EEVDF, scx_flash before the change and after the change.
### EEVDF
```
$ pgbench -h localhost -p 5432 -U postgres -c 100 -j 16 -T 120 -M prepared --protocol extended testdb
Password: 
pgbench (16.6 (Ubuntu 16.6-0ubuntu0.24.04.1), server 15.10 (Debian 15.10-1.pgdg120+1))
starting vacuum...end.
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1000
query mode: extended
number of clients: 100
number of threads: 16
maximum number of tries: 1
duration: 120 s
number of transactions actually processed: 663015
number of failed transactions: 0 (0.000%)
latency average = 18.066 ms
initial connection time = 254.886 ms
tps = 5535.153684 (without initial connection time)
```

### scx_flash before change
```
$ pgbench -h localhost -p 5432 -U postgres -c 100 -j 16 -T 120 -M prepared --protocol extended testdb
Password:
pgbench (16.6 (Ubuntu 16.6-0ubuntu0.24.04.1), server 15.10 (Debian 15.10-1.pgdg120+1))
starting vacuum...end.
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1000
query mode: extended
number of clients: 100
number of threads: 16
maximum number of tries: 1
duration: 120 s
number of transactions actually processed: 298540
number of failed transactions: 0 (0.000%)
latency average = 40.133 ms
initial connection time = 249.265 ms
tps = 2491.686311 (without initial connection time)
```

### scx_flash after change
```
$ pgbench -h localhost -p 5432 -U postgres -c 100 -j 16 -T 120 -M prepared --protocol extended testdb
Password:
pgbench (16.6 (Ubuntu 16.6-0ubuntu0.24.04.1), server 15.10 (Debian 15.10-1.pgdg120+1))
starting vacuum...end.
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1000
query mode: extended
number of clients: 100
number of threads: 16
maximum number of tries: 1
duration: 120 s
number of transactions actually processed: 408782
number of failed transactions: 0 (0.000%)
latency average = 29.312 ms
initial connection time = 280.524 ms
tps = 3411.620821 (without initial connection time)
```

Though both perform worse than EEVDF, we can see some improvements in both number of transactions actually processed , latency average and TPS.

Btw, using bpftop to observe `flash_dispatch`, it seems that before the change, the handler's behavior tends to variate much more severse than after the change, I guess this is because before the change, CPU are too busy grabbing for new tasks from the global queue , and left tasks which is about to finish to wait unnecessary.

The tests ran on x86_64 architecture with AMD Ryzen 7 5700X3D 8-Core Processor.

*Maybe more workloads are need to be tested, please let me know I'll be happy to test them.*

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -825,7 +825,6 @@ void BPF_STRUCT_OPS(flash_dispatch, s32 cpu, struct task_struct *prev)
 	    is_task_locked(prev)) {
 		task_unlock(prev);
 		task_refill_slice(prev);
-		return;
 	}
 
 	/*


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -818,29 +818,21 @@ void BPF_STRUCT_OPS(flash_enqueue, struct task_struct *p, u64 enq_flags)
...
+	 * (i.e. shorter runtime), so let it finish earlier leads to better
+	 * latency outcome.
 	 */
-	if (prev && (prev->scx.flags & SCX_TASK_QUEUED) &&
-	    is_task_locked(prev)) {
-		task_unlock(prev);
+	if (prev && (prev->scx.flags & SCX_TASK_QUEUED)) {
+		if (is_task_locked(prev))
+			task_unlock(prev);
 		task_refill_slice(prev);
```
arighi, line 829: Hm.. but in this way we are refilling the time slice for all the tasks that still want to run. Shouldn't we do this only if the task has a lock?  Basically moving the refill inside is_task_locked()

arighi (commented):
Looks like a sane change. I left a comment, let me know if it makes sense. Thanks!

scheds/rust/scx_flash/src/bpf/main.bpf.c
vax-r: Yeah I agree, I should change this part and test it again.
My original intent was more aggressive, as long as the task can still run we let it run one more time (on an assumption that most task still has `slice > 0` wouldn't run for too long.) But the assumption indeed is too optimistic and too aggressive.

Though I have a question, what kind of task would be holding a user-space lock ? does it indicate some process calling `pthread_mutex_lock()` or something like that ?

arighi: The idea here is to trace when user-space tasks are holding a futex. If a task holding a futex uses all its assigned time slice it'd be de-scheduled in favor of another task, but the other task may still need to acquire the same futex, introducing a priority-inversion issue.

So, the idea is to trace when a task is holding a futex and if its time slice expires give it another time slice "for free", so that it can (hopefully) complete its critical section and release the lock. Of course we also want to prevent tasks from abusing this "bonus", so when we give another "free" timeslice we also mark the task a "unlocked".

vax-r:
@arighi -
Sorry for the late reply.
I've made the corresponding change and only prioritize task holding futex this time , then after refilling it's slice, we then another new task. 
For the rest of task which isn't holding futex but still has slice left, let it wait for another round for now.
I ran the same test. The test result shown as below, still can see some improvements on TPS, latency avg and number of transactions actually processed.

```
pgbench -h localhost -p 5432 -U postgres -c 100 -j 16 -T 120 -M prepared --protocol extended testdb
Password:
pgbench (16.6 (Ubuntu 16.6-0ubuntu0.24.04.1), server 15.10 (Debian 15.10-1.pgdg120+1))
starting vacuum...end.
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1000
query mode: extended
number of clients: 100
number of threads: 16
maximum number of tries: 1
duration: 120 s
number of transactions actually processed: 502252
number of failed transactions: 0 (0.000%)
latency average = 23.870 ms
initial connection time = 250.037 ms
tps = 4189.323434 (without initial connection time)
```

Beside of it, I'm thinking maybe add some mechanism to determine whether a task is latency-sensitive task or not will also help as a hint here ( I saw something similar done within `scx_bpfland` ) , as currently my guess for the reason why `flash_dispatch` would be triggered so frequently is that we're queueing most of tasks into Global DSQ to let CPUs to grab task from it.
Maybe let `scx_flash` queue latency-sensitive task into Local DSQ could help us decrease the times we have to enter `flash_dispatch` when handling heavier workloads.

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -825,7 +825,6 @@ void BPF_STRUCT_OPS(flash_dispatch, s32 cpu, struct task_struct *prev)
 	    is_task_locked(prev)) {
 		task_unlock(prev);
 		task_refill_slice(prev);
-		return;
```
arighi, line 828: Hm... if we just remove this `return` here, we would refill prev's time slice, but then call `scx_bpf_dsq_move_to_local()`, which can move the next task to the local DSQ (if there's any task in `SHARED_DSQ`). This can nullify refilling prev's time slice, as we would simply execute the next task upon returning from `ocps.dispatch()`.

In practice, this effectively ignores whether the task is holding a lock or not, which could potentially be better in some scenarios. Perhaps it might even make sense to remove the entire futex priority inversion detection logic altogether...

Can you check if you get the same performance by completely removing the whole `is_task_locked()` logic? I expect that you get the same performance, because your last change is pretty much doing this. Thanks!

vax-r: No problem thanks for the idea ! Let me test it real soon, I should look into what `scx_bpf_dsq_move_to_local()` actually performed deeper, I think I misunderstood some parts.

Removing futex detection to be better is in fact the opposite of my expectation because I do think making futex owner to finish faster will be a benefit , maybe the workload is heavy but can't create enough contention ? I'll check it for more detailed and get back to you ASAP. Thanks!

vax-r: After removing `is_task_locked()` part, the output shows similar performance.
```
$ pgbench -h localhost -p 5432 -U postgres -c 100 -j 16 -T 120 -M prepared --protocol extended testdb
Password: 
pgbench (16.6 (Ubuntu 16.6-0ubuntu0.24.04.1), server 15.10 (Debian 15.10-1.pgdg120+1))
starting vacuum...end.
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1000
query mode: extended
number of clients: 100
number of threads: 16
maximum number of tries: 1
duration: 120 s
number of transactions actually processed: 520080
number of failed transactions: 0 (0.000%)
latency average = 23.071 ms
initial connection time = 275.107 ms
tps = 4334.413322 (without initial connection time)
```

with the current implementation still perform worse ( I even execute more tests on it trying to avoid cache issue. )
```
$ pgbench -h localhost -p 5432 -U postgres -c 100 -j 16 -T 120 -M prepared --protocol extended testdb
Password: 
pgbench (16.6 (Ubuntu 16.6-0ubuntu0.24.04.1), server 15.10 (Debian 15.10-1.pgdg120+1))
starting vacuum...end.
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1000
query mode: extended
number of clients: 100
number of threads: 16
maximum number of tries: 1
duration: 120 s
number of transactions actually processed: 388329
number of failed transactions: 0 (0.000%)
latency average = 30.847 ms
initial connection time = 248.118 ms
tps = 3241.852846 (without initial connection time)
```

Also I found something interesting that in `scx_flash`, actually only `fexit_futex_wake` and `fexit__futex_wait` are triggered, others are almost silence, at least for this workload.

Screenshot 2024-12-24 at 12 44 21 AM


Maybe the deadline policy itself is already good enough for futex owner to finish their critical section in 1 or 2 period so the benefit of it isn't really sufficient in this scenario ?

arighi: Probably for this particular workload the futex tracking is likely not very significant and might actually degrade performance.

Maybe we should introduce a command line option to enable/disable it, so that people can decided if they want the futex tracking or not.

vax-r:
Closed by #1136 .

PR #1123, opened 2024-12-21
scx_flash: Explicitly return in "fexit_vfs_fsync_range"

vax-r:
## Summary
To elimate warning `src/bpf/main.bpf.c:268:2: error: non-void function '____fexit_vfs_fsync_range' should return a value [-Wreturn-type]` during compilation, explicitly return 0 within `fexit_vfs_fsync_range()`.


hodgesds (approved):
LGTM

PR #1124, opened 2024-12-22
scx_layered: Fix open layer handling

htejun:
Open layers get all the CPUs that aren't currently allocated to other types
of layers. If there are multiple open layers, they all share the unallocated
CPUs - ie. an otherwise unallocated CPU belongs to all open layers in the
system.
    
This wasn't handled correctly. An unallocated CPU was assigned to the first
open layer vai cpuc->layer_id and owned execution tracking and protection
only applied to the first open layer. Fix it by implementing dedicated paths
for open layers and treating all of them as the owners.


hodgesds (approved):
This makes sense.

PR #1126, opened 2024-12-22
scx_lavd: Prioritize a migration-disabled task

multics69:
A migration-disabled task has restrictions in task placement, so it has less chance to be scheduled. To address such issue, prioritize a migration-disabled task.


arighi (approved):
LGTM, scx_bpfland is doing something similar. However, keep in mind that this will make the scheduler incompatible with systems where CONFIG_SMP is disabled, as `p->migration_disabled` is not defined.

In fact I was wondering if we should provide a kfunc to access this information. It can be quite useful for making scheduling decisions while maintaining compatibility with !CONFIG_SMP kernels.

What do you think?

htejun:
> In fact I was wondering if we should provide a kfunc to access this information. It can be quite useful for making scheduling decisions while maintaining compatibility with !CONFIG_SMP kernels.
> 
> What do you think?

I don't think you need to add a kfunc for that. You should be able to test whether the field exists and then pick between two alternative implementations accordingly using CO-RE.

arighi:
> > In fact I was wondering if we should provide a kfunc to access this information. It can be quite useful for making scheduling decisions while maintaining compatibility with !CONFIG_SMP kernels.
> > What do you think?
> 
> I don't think you need to add a kfunc for that. You should be able to test whether the field exists and then pick between two alternative implementations accordingly using CO-RE.

Ah! I forgot about `bpf_core_field_exists()`. So we should be able to do something like this (untested) to support both CONFIG_SMP and !CONFIG_SMP:
```
static bool is_migration_disabled(struct task_struct *p)
{
        if (bpf_core_field_exists(p->migration_disabled))
                return p->migration_disabled;
        return false;
}
```

multics69:
Thank you @arighi and @htejun for the comments. I will add the CO-RE checking and merge it.

PR #1129, opened 2024-12-26
scx_layered: Implement layer property disallow_preempt_after_us

htejun:
This is used to modulate owner execution protection against preemptions
separately from open/grouped executions so that preemptions can be further
prioritized.


hodgesds (approved):
LGTM

etsal (approved):

PR #1132, opened 2024-12-27
scx_utils: expose possible and online cpus from Topology

evanrittenhouse:
This exposes possible and online CPUs to schedulers.

Related to #721.


htejun (commented):
Given that `topology.span` is the online CPUs, it's rather odd to add separate online mask. Maybe just add `topology.span_possible`?

evanrittenhouse:
> Given that `topology.span` is the online CPUs, it's rather odd to add separate online mask. Maybe just add `topology.span_possible`?

Ah, I had this in draft as I was still exploring, sorry about that! Does it make sense to move the `span = cpus_online()` call into `instantiate`, like here?

htejun:
> > Given that `topology.span` is the online CPUs, it's rather odd to add separate online mask. Maybe just add `topology.span_possible`?
> 
> Ah, I had this in draft as I was still exploring, sorry about that! Does it make sense to move the `span = cpus_online()` call into `instantiate`, like here?

Hmm.. are there reasons it can't be done in `new()`? `instantiate()` is there to post process the collected topology information. I don't know why online / possible cpumasks would need to be processed there.

evanrittenhouse:
> > > Given that `topology.span` is the online CPUs, it's rather odd to add separate online mask. Maybe just add `topology.span_possible`?
> > 
> > 
> > Ah, I had this in draft as I was still exploring, sorry about that! Does it make sense to move the `span = cpus_online()` call into `instantiate`, like here?
> 
> Hmm.. are there reasons it can't be done in `new()`? `instantiate()` is there to post process the collected topology information. I don't know why online / possible cpumasks would need to be processed there.

Sure thing. I still need to figure out how to integrate the `possible` cpumasks into the schedulers, though maybe that could/should be a followup

PR #1134, opened 2024-12-28
lib/sdt_alloc: fix LLVM19 compilation error

etsal:
Fix a compilation error on Clang 19 caused by defining multiple arena pointers in a single comma separated statement. Also fix a duplicate use of ```__arena``` that causes LLVM to emit a warning.


hodgesds (approved):
LGTM

PR #1139, opened 2024-12-31
scx_rusty: prevent scheduling bubbles on kernels >= 6.12

arighi:
Selecting an idle CPU without dispatching a task to it can lead to scheduling bubbles, as the CPU remains incorrectly marked as busy and is excluded from the ops.select_cpu() selection, resulting in inefficient core utilization.

To prevent this, always check the idle state of the CPUs that we want to wake up, but avoid marking them as busy, unless we are certain that a task will be dispatched on them.

Note that this issue occurs only in kernel 6.12 and later (following the pick_next_task() refactoring), as the in-kernel CPU idle state update happens only during transitions between a non-idle task and the idle task, whereas earlier kernels refreshed the CPU idle state on every idle task cycle (as a result of the put_prev_task() call).

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -1092,10 +1092,13 @@ s32 BPF_STRUCT_OPS(rusty_select_cpu, struct task_struct *p, s32 prev_cpu,
 	 * stalls as all in-domain CPUs may be idle by the time @p gets
 	 * enqueued.
 	 */
-	if (prev_domestic)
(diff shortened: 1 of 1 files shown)


etsal (approved):
LGTM

htejun (approved):

htejun:
Ah, that's unfortunate. We probably should make at least kick_cpu() to refresh the state.

arighi:
I agree, we can fix the schedulers for now (this happens also with scx_layered BTW), but it'd nicer if to force an idle refresh when we explicitly kick idle CPUs. Will look at that.

PR #1142, opened 2025-01-01
scx_flash: prevent scheduling bubbles on kernels >= 6.12

arighi:
Similar to commit 809ce6c ("scx_rusty: prevent scheduling bubbles on kernels >= 6.12"), avoid allocating idle CPUs unless it is certain that a task will be dispatched to them.

This ensures the CPU idle state reflects its actual state and is not allocated without running any tasks.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -73,6 +73,11 @@ const volatile bool smt_enabled = true;
 static u64 vtime_now;
 
 /*
+ * Maximum possible CPU number.
+ */
+static u64 nr_cpu_ids;
+
+/*
  * Per-CPU context.
  */
 struct cpu_ctx {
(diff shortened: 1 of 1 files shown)


ptr1337 (approved):

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -703,13 +719,21 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
...
+	if (is_prev_llc_affine) {
 		cpu = prev_cpu;
+		goto out_put_cpumask;
+	}
+
+	/*
+	 * Otherwise, return a random CPU in the task's LLC domain.
+	 */
+	cpu = bpf_cpumask_any_distribute(llc_mask);
+	if (cpu < nr_cpu_ids)
```
etsal, line 733: nit: can't this if-else be 
```
if (cpu >= nr_cpu_ids)
	cpu = prev_cpu;
```

etsal (approved):

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: oh yes, definitely, will change that, thanks

PR #1146, opened 2025-01-01
scx_utils: emit compilation warning when compiling multiple *.bpf.c files

etsal:
Commit a4739e9 introduces code in ```scx_utils``` to capture and emit Clang warnings when compiling the BPF schedulers.  Schedulers like ```scx_layered``` use a separate code path in ```scx_utils``` in which we still do not emit Clang warnings. Reuse the code from a4739e9 in those code paths to emit warnings for schedulers that use multiple compilation units.


JakeHillion (approved):
Thanks for the catch!

PR #1148, opened 2025-01-02
scx_bpfland: do not allow per-CPU kthread to preempt other tasks

arighi:
While prioritizing per-CPU kthreads is generally a good idea due to their critical role in the system, allowing them to preempt other tasks indiscriminately can result in significant scheduling overhead sometimes.

A notable example is starting a VM, which can be quite inefficient if we allow with per-CPU kthread preemption.

For instance, running a simple `time vng -- uname -r` produces the following results:

 - EEVDF:              1.53 seconds
 - bpfland:            3.30 seconds
 - bpfland-no-preempt: 1.32 seconds

Moreover, disabling per-CPU kthread preemption does not appear to introduce any regressions in typical latency-sensitive benchmarks.

Therefore, it makes sense to disable preemption for the per-CPU kthreads.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -690,8 +690,7 @@ void BPF_STRUCT_OPS(bpfland_enqueue, struct task_struct *p, u64 enq_flags)
 	 * If local_kthread is specified dispatch all kthreads directly.
 	 */
 	if (is_kthread(p) && (local_kthreads || p->nr_cpus_allowed == 1)) {
-		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, SCX_SLICE_DFL,
-				   enq_flags | SCX_ENQ_PREEMPT);
+		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, SCX_SLICE_DFL, enq_flags);
 		__sync_fetch_and_add(&nr_kthread_dispatches, 1);
 		return;
 	}


arighi:
Closing for now as it seems to introduce regressions / stalls.

arighi:
Actually... let me re-open it, so that I'm forced to look at this. :) We need to figure out what is happening, because it looks like a bug that may potentially affect other schedulers as well.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -690,8 +690,7 @@ void BPF_STRUCT_OPS(bpfland_enqueue, struct task_struct *p, u64 enq_flags)
 	 * If local_kthread is specified dispatch all kthreads directly.
 	 */
 	if (is_kthread(p) && (local_kthreads || p->nr_cpus_allowed == 1)) {
-		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, SCX_SLICE_DFL,
-				   enq_flags | SCX_ENQ_PREEMPT);
+		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, SCX_SLICE_DFL, enq_flags);
```
hodgesds, line 693: On the `enq_flags` should it forcibly set `SCX_ENQ_LAST` and `SCX_ENQ_PREEMPT`?

arighi: Does it make any difference? My understanding is that SCX_ENQ_LAST is set by the kernel when SCX_OPS_ENQ_LAST is enabled and the task is the only one running on a CPU. I don't think the kernel is actually using it if we explicitly set this flags in the scheduler, unless I'm missing something...

arighi: Hm.. it seems that, for me, only SCX_ENQ_PREEMPT has an impact:
```
enq_flags | SCX_ENQ_PREEMPT | SCX_ENQ_LAST => 2.40-2.70s
enq_flags | SCX_ENQ_PREEMPT                => 2.40-2.70s
enq_flags | SCX_ENQ_LAST                   => 1.30-1.40s
enq_flags                                  => 1.30-1.40s
```

hodgesds: >  I don't think the kernel is actually using it if we explicitly set this flags in the scheduler, unless I'm missing something.

Wasn't sure just more a shot in the dark based on the comment for `SCX_ENQ_LAST`:
>  The BPF scheduler is responsible for triggering a follow-up scheduling event. Otherwise, Execution may stall.

hodgesds: It does automatically set [`SCX_ENQ_DSQ_PRIQ`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/sched/ext.c#n6423), what I think probably would make the most difference is getting the vtime set ideally (not sure what ideal is in this situation) if dispatching to a vtime dsq.

arighi: I've also tried to dispatch the per-cpu kthreads to the SHARED_DSQ, either setting the vtime to make sure they were dispatched before any other task, and even setting vtime=0 (to be 100% sure they are dispatched before any other task), no difference... we're still getting the weird stall trace showing for example a kthread being the first in the queue, yet it has been sitting in the DSQ for > 5 sec.

It almost seems that the CPU is going idle incorrectly or it's stolen by a higher scheduling class (like an RT task monopolizing the CPU?). The fact that I can't reproduce it locally doesn't help unfortunately....

arighi:
Merging this one, since it has been extensively tested now and it also solves #1153.

PR #1150, opened 2025-01-02
lib/sdt_data: Support allocators for large structs

etsal:
We can currently create allocators for structs that are up to a page in size, but that prevents us from combining them with structs like ```scx_rusty```'s domain structs. Expand the allocator code to support struct sizes larger than a page.


lib/sdt_alloc.bpf.c
```diff
@@ -231,20 +260,20 @@ sdt_desc_t *sdt_alloc_chunk(struct sdt_alloc_stack __arena *stack)
 	return out;
 }
 
-static SDT_TASK_FN_ATTRS int sdt_pool_set_size(struct sdt_pool *pool, __u64 data_size)
+static SDT_TASK_FN_ATTRS int sdt_pool_set_size(struct sdt_pool *pool, __u64 data_size, __u64 alloc_size)
```
htejun, line 263: Maybe sth like `nr_pages` instead of `alloc_size`? The units can be confusing if len/size are used for page count.

htejun (approved):

PR #1154, opened 2025-01-03
scxtop: Add scxtop tui

hodgesds:
Add `scxtop` as a TUI utility that collects and renders system perf events as well as scheduler events via BPF. This is a rework of the existing bpftrace script, which doesn't really work that well. 

This is still a WIP, but it's far enough along to be useful and should be able to be extended rather easily. It provides aggregations at the CPU, LLC, Node level for perf events and at the DSQ level for bpf events. This still some bugs with how vtime/dsq latency is tracked. Other bonus features include a keymap (which can later be customized) and color themes. This uses `scx_utils` for generating topology and if needed could be put into a separate repo.

Example usage:
image

https://github.com/user-attachments/assets/3ae1c0f6-c724-46c6-a688-79aac9788fa7

Scales reasonably well to systems with large core counts:
image


htejun (approved):

multics69:
Thanks @hodgesds ! This is awesome!

PR #1156, opened 2025-01-06
scx_lavd: Add a fast path for a migration-disabled task in pick_idle_cpu()

multics69:
If a task is not migratable, do not go through the complete idle CPU selection logic.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -292,13 +292,7 @@ static u64 calc_weight_factor(struct task_struct *p, struct task_ctx *taskc,
 	 * Prioritize a migration-disabled task since it has restrictions
 	 * in placement so it tends to be delayed.
 	 */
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -748,7 +748,7 @@ static s32 pick_idle_cpu(struct task_struct *p, struct task_ctx *taskc,
 	 * If a task can run only on a single CPU (e.g., per-CPU kworker), we
 	 * simply check if a task is still pinned on the prev_cpu and go.
 	 */
-	if (p->nr_cpus_allowed == 1 &&
+	if ((p->nr_cpus_allowed == 1 || is_migration_disabled(p)) &&
```
arighi, line 751: We can probably move `p->nr_cpus_allowed == 1` into `is_migration_disabled()`.

multics69: Thanks for the comment. I will change it as suggested.

arighi (approved):

PR #1157, opened 2025-01-06
[scx] fix ci

likewhatevs:
fix ci, looks like vng was updated making `--net` ambiguous.


arighi:
@likewhatevs thanks for fixing this!

PR #1158, opened 2025-01-06
meson: do not link BPF library into C schedulers by default

etsal:
Meson currently defaults to building and linking the BPF library into all existing C schedulers, but that has caused build errors even for schedulers that do not use it (most recently scx_vder). Split the existing Meson build path in two, depending on whether the library should be linked into the scheduler. 

This change does not affect Rust schedulers.

### DIFF
--- a/meson-scripts/bpftool_build_skel
+++ b/meson-scripts/bpftool_build_skel
@@ -6,20 +6,14 @@ bpftool="$1"
 input="$2"
 skel="$3"
 subskel="$4"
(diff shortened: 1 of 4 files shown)


etsal:
@arighi Hopefully this fixes the build error, if not please let me know and I will look further into it.

arighi (approved):
LGTM

Thanks for looking at this @etsal and sorry everyone for the intense the git push activty in the past days...

PR #1163, opened 2025-01-08
docs: update nix install docs

JohnRTitor:
`scx` has been available on Nixpkgs for a while now.


JohnRTitor:
CC @htejun @JakeHillion

JakeHillion:
Why is there any reference to cachyos here at all now? In my NixOS configs I have something like:

```
boot.kernelPackages = if pkgs.linuxPackages.kernelAtLeast "6.12" then pkgs.linuxPackages else pkgs.linuxPackages_latest;
```

Which works totally fine for sched_ext - no deps outside of nixpkgs needed. These docs could say `pkgs.linuxPackages_latest` and remove any reference to Chaotic Nyx/flakes.

An aside that I missed commenting when this change went into nixpkgs: I am not a fan of shipping all the schedulers as one package in Nix. Building them together makes sense, but is there a way we can provide subpackages with a single scheduler? It seems very strange to me to pollute the nix store with schedulers that aren't used after a NixOS option has explicitly requested one scheduler.

JohnRTitor:
> Why is there any reference to cachyos here at all now?

Removed the flake and cachyos kernel reference.

> but is there a way we can provide subpackages with a single scheduler

There are `scx.rustscheds`, and `scx.cscheds` for Rust and C schedulers respectively. We can definitely expose `scx.rustland`, `scx.bpfland` single scheduler packages, but as new schedulers are added and removed/disabled (AFAIK rusty had a build issue in a 1.0.4 or so), it may become a hassle for downstream maintainers.

For now you can modify `services.scx.package` and set it to something like `scx.rustscheds` if you wish to only use the rust schedulers. We will keep `scx.full` as default for obvious reasons though.

JakeHillion (approved):
Thanks for the updates! Really nice that this is available by default in Nix 24.11 and doesn't need anything unstable/any external deps.

PR #1164, opened 2025-01-08
scxtop: Add additional bpf kprobes

hodgesds:
Add kprobes for handling various events so that dispatch latency and vtime tracking can be more accurate.


etsal (approved):
LGTM

PR #1170, opened 2025-01-09
scx_rusty: Move domains into arenas

etsal:
Continue the transition of scx_rusty to using arenas for its data structures by adjusting domains to be allocated from arena memory.


etsal:
The code is still in WIP state because Clang 19 breaks some of the underlying assumptions made in it.

scheds/rust/scx_rusty/src/bpf/sdt_dom.h
```diff
@@ -0,0 +1,127 @@
...
+	dom_ptr domc;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, u32);
+	__type(value, struct sdt_dom_map_val);
+	__uint(max_entries, MAX_DOMS);
+	__uint(map_flags, 0);
+} sdt_dom_map SEC(".maps");
```
htejun, line 22: I wonder whether using global array (.bss/.data) would be better for this as that would get mmapped to userspace and then userspace can follow the pointers directly into arena without going through map lookups.

htejun: and for the longer term, it'd be nice to abstract this style of usage where the application may allocate N arena elements into a static array.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -337,14 +312,14 @@ static void task_load_adj(struct task_ctx *taskc,
...
-static struct bucket_ctx *lookup_dom_bucket(struct dom_ctx *dom_ctx,
+static struct bucket_ctx *lookup_dom_bucket(dom_ptr dom_ctx,
 					    u32 weight, u32 *bucket_id)
 {
 	u32 idx = weight_to_bucket_idx(weight);
 	struct bucket_ctx *bucket;
 
 	*bucket_id = idx;
-	bucket = MEMBER_VPTR(dom_ctx->buckets, [idx]);
+	bucket = (struct bucket_ctx *)&dom_ctx->buckets[idx];
```
htejun, line 322: Is the explicit casting necessary?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -502,21 +466,21 @@ static void dom_dcycle_xfer_task(struct task_struct *p, struct task_ctx *taskc,
 			   to_dcycle[1] >> RAVG_FRAC_BITS);
 }
 
-static u64 dom_min_vruntime(struct dom_ctx *domc)
+static u64 dom_min_vruntime(dom_ptr domc)
 {
-	return READ_ONCE(domc->min_vruntime);
+	return domc->min_vruntime;
```
htejun, line 471: Hmm... why no READ_ONCE?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1193,20 +1161,22 @@ static void place_task_dl(struct task_struct *p, struct task_ctx *taskc,
...
 
 	if (!(taskc = lookup_task_ctx_mask(p, &p_cpumask)) || !p_cpumask)
 		return;
 
+	domc = task_domain(taskc);
+	if (!domc)
+		return;
+
 	/*
 	 * Migrate @p to a new domain if requested by userland through lb_data.
```
htejun, line 1176: Stray reference to lb_data. There are also references to lb_data in the top comment.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1193,20 +1161,22 @@ static void place_task_dl(struct task_struct *p, struct task_ctx *taskc,
...
 	 */
-	new_dom = bpf_map_lookup_elem(&lb_data, &key);
-	if (new_dom && *new_dom != taskc->dom_id &&
-	    task_set_domain(p, *new_dom, false)) {
+	if (domc->id != taskc->dom_id &&
```
htejun, line 1178: Maybe rename taskc->dom_id to something which indicates that it may not represent the current assignment? e.g. taskc->target_dom_id or something like that?

htejun (commented):
Overall looks great to me. Left a few comments.

scheds/rust/scx_rusty/src/bpf/sdt_dom.h
etsal: I agree, I main.bpf.c I expose an array with all the arena-allocated domain contexts anyway so if we move the privileged data there we can completely remove the map. IMO whether we use a map or an array for the privileged data depends on how many instances of the type we need to support. For example, tasks should use maps while domains/nodes should use static arrays.

As an aside, I thought BPF_MAP_TYPE_ARRAY maps are basically arrays hidden behind a BPF map interface? Is the extra machinery in bpf_array a source of overhead/extra locking?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
etsal: That's to get rid of the __arena attribute, Clang 19 prevents assigning arena pointers to non-arena pointers including passing them as function arguments. AFAICT typecasting and assigning is how we turn arena pointers to kernel pointers without explicit cast_kern() calls that are no-ops for 19.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
etsal: This was a hacky fix that slipped in, READ_ONCE/WRITE_ONCE don't work with __arena so I need to look into why and fix it. Adding *_ARENA variants of the macros may do the trick but I haven't checked.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
etsal: I will do a pass, the comments are now quite out of date, e.g..

scheds/rust/scx_rusty/src/bpf/main.bpf.c
etsal: Will do!

scheds/rust/scx_rusty/src/bpf/main.bpf.c
htejun: I see. It'd be nice if we can come up with a casting macro which can drop __arena attribute. IIRC, there's newish compiler feature to get the attributeless type of a variable. Anyways, we can figure that out later.

scheds/rust/scx_rusty/src/bpf/sdt_dom.h
htejun: Using global array is nice because they get auto mmapped and bpftool generates typed access fields for them in the skeleton. Other than that, they're the same but the mmaps do give you the extra ability for partial updates.

etsal:
Addressed most feedback except for the first. An unexpected problem with going from a map to a BSS array is that structs in maps work well with the verifier. Statically allocating domains in the BSS means the verifier turns all our kptrs into scalars. Not sure if this is programming error or a more fundamental limitation of the verifier.

htejun:
Ah, they're mmapped to userspace so can't carry kptrs. Oh well, I guess the index would have to be a map.

etsal:
We can still define a programming/data model for the various abstractions we have. Before, we attempted to do this:

```
struct abstraction {
    struct lock lock;
    struct pointer __kptr ptr;

    struct abstraction_arena __arena *arena_data;
}

struct abstraction abstractions[MAX_ABSTRACTIONS];

```

The above hits the issue I described. We can instead have the following:

```

struct abstraction_privileged {
    struct lock lock;
    struct pointer __kptr ptr;
};

struct abstraction {
    [data fields]
    u64 id; // Used to access the map
};

typedef struct abstraction __arena * abs_ptr;

abs_ptr abstractions[MAX_ABSTRACTIONS];

```

This is close to what we have but a bit tidier. We still have to access privileged data through a map, but we can hide that from the users of the API.

etsal:
(Pressed the wrong button trying to comment)

etsal:
I will be reorganizing the diffs after the review to keep the history clean

scheds/rust/scx_rusty/src/bpf/sdt_dom.h
```diff
@@ -0,0 +1,132 @@
...
+	return domc;
+}
+
+__hidden
+void sdt_dom_free(dom_ptr domc)
+{
+	struct lb_domain *mval;
+	u32 key = domc->id;
+
+	sdt_arena_verify();
```
hodgesds, line 70: I wonder if it makes sense to have macro like `__arena_verified` or something that ensures arena state of the function without having to explicitly verify the state.

etsal: So this is not really to verify the arena variables, it is to touch the arena map from the subprogram otherwise the verifier fails. This was the case with scx_sdt where a function that did not use the arena map somehow could not use any arena variables.

etsal: There is a comment on the definition lib/sdt_alloc.bpf.c:38 right above the definition of sdt_arena_verify. I can add one at the call site if that helps

etsal: Actually the problem is the name, it is misleading.  I wlil update it to make it explicit that it is a verifier workaround.

hodgesds: Ah, sorry that was me being lazy and not looking at the definition. It's probably fine.

etsal: No it's a fair point sdt_arena_verify sounds exactly like it verifies arena variables :P It needs changing especially considering it's there to handle an edge case in the verifier's detail.

htejun (approved):

PR #1177, opened 2025-01-11
scxtop: Add CPU frequency to default sparkline view

hodgesds:
Add CPU frequency to the sparkline default view.

sparkline:

barchart:

Can be toggled with the `f` key.


likewhatevs (approved):
nice

PR #1178, opened 2025-01-11
scheds: Sync scx schedulers with the kernel changes: time_helpers and scx_bpf_now()

multics69:
As the kernel patch for scx_bpf_now() and time helpers are merged (https://lore.kernel.org/lkml/20250109131456.7055-1-changwoo@igalia.com/), sync up the relevant headers and scx schedulers to use time helpers.

### DIFF
--- a/scheds/include/scx/compat.bpf.h
+++ b/scheds/include/scx/compat.bpf.h
@@ -125,6 +125,11 @@ bool scx_bpf_dispatch_vtime_from_dsq___compat(struct bpf_iter_scx_dsq *it__iter,
 	false;									\
 })
 
+#define scx_bpf_now()								\
+	(bpf_ksym_exists(scx_bpf_now) ?						\
+	 scx_bpf_now() :							\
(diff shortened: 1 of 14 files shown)


scheds/include/scx/compat.bpf.h
```diff
@@ -125,6 +125,11 @@ bool scx_bpf_dispatch_vtime_from_dsq___compat(struct bpf_iter_scx_dsq *it__iter,
 	false;									\
 })
 
+#define scx_bpf_now()								\
```
arighi, line 128: Maybe we could add a little description of what `scx_bpf_now()` does, or even something like "Return a monotonically non-decreasing time in nanoseconds, use `bpf_ktime_get_ns()` if you have high-precision timing requirements" or something similar.

arighi (approved):
I left a small comment, but LGTM, thanks!

hodgesds (approved):
Good stuff!

PR #1180, opened 2025-01-11
scxtop: Fix attach on newer kernels

hodgesds:
Updated kernels and noticed `scxtop` wasn't working properly on new probes. This simplifies attachment and works with the `for-6.14` branch:


etsal:
LGTM

PR #1181, opened 2025-01-11
scx_utils: Fix UserExitInfo::new() usage.

devnexen:
it is dealing with pointer with various sizes/types thus dereferencing these could be considered UB. Thus marking it as unsafe overall. reported by clippy.

### DIFF
--- a/rust/scx_utils/src/user_exit_info.rs
+++ b/rust/scx_utils/src/user_exit_info.rs
@@ -58,13 +58,15 @@ macro_rules! uei_read {
                 _ => std::ptr::null(),
             };
 
-            scx_utils::UserExitInfo::new(
-                &bpf_uei.kind as *const _,
-                exit_code_ptr,
-                bpf_uei.reason.as_ptr() as *const _,
-                bpf_uei.msg.as_ptr() as *const _,
-                bpf_dump,
-            )
+            unsafe {
+                scx_utils::UserExitInfo::new(
+                    &bpf_uei.kind as *const _,
+                    exit_code_ptr,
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
This seems more correct, thanks for the PR!

htejun:
`UserExitInfo::new()` is given pointer values and then reads them out using `unsafe` which is expectedly required by the compiler. Given that `new()` just takes pointer values, I don't understand why `unsafe` would be needed here. What is clippy saying exactly and why is that not aligning with the compiler?

devnexen:
> `UserExitInfo::new()` is given pointer values and then reads them out using `unsafe` which is expectedly required by the compiler. Given that `new()` just takes pointer values, I don't understand why `unsafe` would be needed here. What is clippy saying exactly and why is that not aligning with the compiler?

image

would you prefer to revert it (or another change) ?

htejun:
I'm having a hard time seeing the rationale for the warning. Yes, the inner function dereferences the passed-in pointers which is unsafe and thus using `unsafe` blocks. I don't see why the callers of the function need to be annotated `unsafe`. We sure don't want to make `unsafe` markings to become transitive.

To my mind, `unsafe` is mostly about telling the compiler to disable safety checks because the code in question can't be vetted for memory safety using rust rules. As such, the more targeted and small unsafe blocks are, the less we're disabling safety checks, which is what we want. Even in this conversion, we ended up putting the whole `new` block inside `unsafe` which makes the code *less* safe by disabling safety checks when evaluating the arguments to the `new` function.

So, let's revert this and disable that particular lint option.

devnexen:
ok doing it.

PR #1183, opened 2025-01-12
sdt_task: Cleanup formatting

hodgesds:


scheds/include/lib/sdt_task.h
```diff
@@ -74,28 +74,28 @@ struct sdt_chunk {
...
+	__u64		free_ops;
+	__u64		active_allocs;
+	__u64		arena_pages_used;
 };
 
 struct sdt_allocator {
-	struct sdt_pool pool;
-	sdt_desc_t *root;
+	struct sdt_pool	pool;
+	sdt_desc_t *	root;
```
etsal, line 98: Nit: Can we do

```
	sdt_desc_t     *root;
```

so that it is consistent with the other definitions?

etsal (approved):
Nit:

PR #1187, opened 2025-01-13
lib/sdt_alloc: rename sdt_arena_verify to avoid confusion

etsal:
The sdt_arena_verify function has a misleading name, leading to the mistaken assumption it verifies arena variables. The function is in fact a workaround that ensures all subprograms explicitly access the arena map, even if they do not need to. This is because the verifier finds the arena associated with a subprogram by identifying these accesses in the bytecode, and forbids use of arena variables if it finds none.

### DIFF
--- a/lib/sdt_alloc.bpf.c
+++ b/lib/sdt_alloc.bpf.c
@@ -48,7 +48,7 @@ static bool sdt_verify_once;
 
 #define SDT_TASK_FN_ATTRS	inline __attribute__((unused, always_inline))
 
-__hidden void sdt_arena_verify(void)
+__hidden void sdt_subprog_init_arena(void)
 {
 	if (sdt_verify_once)
 		return;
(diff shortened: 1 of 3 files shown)


lib/sdt_alloc.bpf.c
```diff
@@ -48,7 +48,7 @@ static bool sdt_verify_once;
 
 #define SDT_TASK_FN_ATTRS	inline __attribute__((unused, always_inline))
 
-__hidden void sdt_arena_verify(void)
+__hidden void sdt_associate_subprog_with_arena(void)
```
hodgesds, line 51: Maybe `sdt_subprog_init_arena`, it's maybe a little less verbose.

hodgesds (approved):
It's probably fine as is so I'll leave it up to you, I think this is more a reading comprehension error than anything.

etsal:
I think there was a legitimate issue with the naming being misleading, appreciate the feedback

PR #1189, opened 2025-01-13
Revert gh1181

devnexen:


rust/scx_utils/src/user_exit_info.rs
```diff
@@ -136,7 +134,11 @@ impl UserExitInfo {
     /// user_exit_info, so we can't take the type directly. Instead, this
     /// method takes each member field. Use the macro uei_read!() on the C
     /// type which then calls this method with the individual fields.
-    pub unsafe fn new(
+    ///
+    /// NOTE: do not remove, pointer accesses here are safe and self
+    /// contained into unsafe blocks
+    #[allow(clippy::not_unsafe_ptr_arg_deref)]
```
htejun, line 140: Is there a way to apply this to the whole code base?

devnexen: Well ... if this is what you want, we can try sure.

htejun (approved):

PR #1191, opened 2025-01-14
scxtop: Collect uncore frequency

hodgesds:
Add support for collecting uncore frequency. This only works with [Intel uncore freq scaling](https://docs.kernel.org/admin-guide/pm/intel_uncore_frequency_scaling.html)
image


arighi (approved):
Works great, thanks!

PR #1193, opened 2025-01-14
scx_rustland_core: fix static mutable reference usage warning (BUF).

devnexen:


htejun:
Hmm... can you describe the warning in detail? What's the linter not happy about?

devnexen:
image

htejun:
Ah, okay, that's BPF ringbuffer access. Can the `allow` directive attached to the actual unsafe block or at least the callback function?

@arighi I googled a bit and it seems like the preferred to way to handle situations like this seems to be using pointers instead of references. Adding `allow` is fine too, so no need to change it necessarily.

htejun (approved):

arighi:
> Ah, okay, that's BPF ringbuffer access. Can the `allow` directive attached to the actual unsafe block or at least the callback function?
> 
> @arighi I googled a bit and it seems like the preferred to way to handle situations like this seems to be using pointers instead of references. Adding `allow` is fine too, so no need to change it necessarily.

I think I tried at some point to replace this unsafe access using a Mutex, but it didn't work. I'll look at this again, for now it's probably fine to add the allow directive. Thanks for looking at this @devnexen.

devnexen:
@arighi was just thinking if a AtomicPtr would work then ?

PR #1194, opened 2025-01-14
scx_layered: Fix dump_layer_cpumask()

htejun:
It was incorrectly using layer @id to index into $buf. Use a separate $idx and drop confusing noop ++'s.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2475,18 +2475,12 @@ static void dump_layer_cpumask(int id)
 		return;
 
 	bpf_for(cpu, 0, scx_bpf_nr_cpu_ids()) {
-		if (!(p = MEMBER_VPTR(buf, [id++])))
+		if (!(p = MEMBER_VPTR(buf, [cpu])))
 			break;
 		if (bpf_cpumask_test_cpu(cpu, layer_cpumask))
-			*p++ = '0' + cpu % 10;
+			*p = '0' + cpu % 10;
 		else
-			*p++ = '.';
-
-		if ((cpu & 7) == 7) {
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2468,24 +2468,24 @@ static u64 dsq_first_runnable_for_ms(u64 dsq_id, u64 now)
...
-	s32 cpu;
+	s32 cpu, idx = 0;
 	char buf[128] = "", *p;
 
 	if (!(layer_cpumask = lookup_layer_cpumask(id)))
 		return;
 
 	bpf_for(cpu, 0, scx_bpf_nr_cpu_ids()) {
-		if (!(p = MEMBER_VPTR(buf, [id++])))
+		if (!(p = MEMBER_VPTR(buf, [idx++])))
```
eddyz87, line 2478: Note, that such separate counter might be problematic for verifier:
- verifier knows it's initial value
- verifier tracks it's increment in the loop

Because of this verifier can't deduce that different states at the end of the loop body are equal (as idx value differs), and verifies all 128 iterations (limited by `if ... MEMBER_VPTR ...`).
Under certain conditions, verifier state caching might be not deep enough for this code to converge.

It appears that using `cpu` would have same effect as `idx`, but verifier "forgets" it's range at the start of each iteration, so the loop should converge faster and w/o relying on deep state caching.

htejun: This being a debug function we can drop it but the reason why it was using a separate variable was because it was inserting `|` every 8 cpus, so the index doesn't quite coincide with `cpu`. I guess I can substitute it with arithmetic on `cpu`. For now, I'll just drop the `|` separators.

etsal (approved):

hodgesds (approved):

PR #1198, opened 2025-01-15
scx_central: fix cpu affinity corrupted allocation (glibc).

devnexen:
at startup it crashes directly:

```shell
% sudo scx_central
malloc(): unsorted double linked list corrupted
Aborted
```

it appears it comes down to the cpu binding.
Note: another fix would be to use the stack.

### DIFF
--- a/scheds/c/scx_central.c
+++ b/scheds/c/scx_central.c
@@ -96,7 +96,7 @@ restart:
 	 */
 	cpuset = CPU_ALLOC(skel->rodata->nr_cpu_ids);
 	SCX_BUG_ON(!cpuset, "Failed to allocate cpuset");
-	CPU_ZERO(cpuset);
+	CPU_ZERO_S(CPU_ALLOC_SIZE(skel->rodata->nr_cpu_ids), cpuset);
 	CPU_SET(skel->rodata->central_cpu, cpuset);
 	SCX_BUG_ON(sched_setaffinity(0, sizeof(*cpuset), cpuset),
 		   "Failed to affinitize to central CPU %d (max %d)",


htejun:
Hmm... do you see anything wrong with the existing code? What's your environment?

devnexen:
Not really, that is a surprising bug. Just using regular debian.

htejun:
So, I'm not necessarily against working around but let's at least clearly document what's being worked around and why such conclusion was reached. Also, it'd probably be a good idea to investigate a bit further - e.g. glibc malloc has a bunch of debug options that can be enabled with env vars. Can you try them and see whether that points to the source of the error?

It's okay to not be able to root cause it too but if it comes to that, can you please put the code being worked around in an `ifdef` block so that we can easily test the original code easily later?

devnexen:
For now, this is what I got when ran through a debugger

```shell
__pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
warning: 44	./nptl/pthread_kill.c: No such file or directory
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007ffff7e06def in __pthread_kill_internal (threadid=<optimized out>, signo=6) at ./nptl/pthread_kill.c:78
#2  0x00007ffff7db2d02 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff7d9b4f0 in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007ffff7d9c32d in __libc_message_impl (fmt=fmt@entry=0x7ffff7f1e303 "%s\n") at ../sysdeps/posix/libc_fatal.c:132
#5  0x00007ffff7e10925 in malloc_printerr (str=str@entry=0x7ffff7f219e8 "malloc(): unsorted double linked list corrupted") at ./malloc/malloc.c:5772
#6  0x00007ffff7e13dc4 in _int_malloc (av=av@entry=0x7ffff7f5aac0 <main_arena>, bytes=bytes@entry=40) at ./malloc/malloc.c:4086
#7  0x00007ffff7e15e81 in __libc_calloc (n=<optimized out>, elem_size=<optimized out>) at ./malloc/malloc.c:3754
#8  0x0000555555587527 in bpf_map__attach_struct_ops (map=0x5555555d8360) at libbpf.c:12980
#9  0x000055555555c918 in main (argc=1, argv=0x7fffffffe428) at ../scheds/c/scx_central.c:106
```
```shell
(gdb) f 8
#8  0x0000555555587527 in bpf_map__attach_struct_ops (map=0x5555555d8360) at libbpf.c:12980
12980		link = calloc(1, sizeof(*link));
(gdb) l
12975		if (map->fd < 0) {
12976			pr_warn("map '%s': can't attach BPF map without FD (was it created?)\n", map->name);
12977			return libbpf_err_ptr(-EINVAL);
12978		}
12979	
12980		link = calloc(1, sizeof(*link));
```

After investigation, if I remove the cpu binding part, it no longer crashes. But I ll see further a bit later.

devnexen:
I finally got to the bottom of this, this kind of fix is not needed, the cpu set is just not properly zeroed. committing soonish.

htejun (approved):

arighi (approved):
Thanks for figuring this out!

PR #1203, opened 2025-01-15
scx_layered: add rule for matching thread to name change

etsal:
Add a new layer matching rule that matches tasks with a special flag set in their task_ctx. As a proof of concept this flag is set when the task's comm name is set using prctl() (e.g., because of a userspace call to pthread_setname_np()), but can be expanded to be set from other events.


etsal:
Note: This code currently crashes some 6.9 kernels.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -432,6 +435,34 @@ static struct task_ctx *lookup_task_ctx(struct task_struct *p)
 	return taskc;
 }
 
+/* HACK opcode imported wholesale from linux/prctl.h */
+#define PR_SET_NAME 15
+
+SEC("ksyscall/prctl")
+int BPF_KSYSCALL(prctl_enter, int option, unsigned long arg2,
+	unsigned long arg3, unsigned long arg4, unsigned long arg5)
```
htejun, line 443: We're already attaching to `tp_task_rename()`, so no need to create another one.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -432,6 +435,34 @@ static struct task_ctx *lookup_task_ctx(struct task_struct *p)
...
+
+	p = bpf_get_current_task_btf();
+	if (!p)
+		return -EINVAL;
+
+	taskc = lookup_task_ctx_may_fail(p);
+	if (!taskc)
+		return -EINVAL;
+
+	taskc->flagged = true;
```
htejun, line 459: `->flagged` is too generic a name.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1768,6 +1799,14 @@ static __noinline bool match_one(struct layer_match *match,
 		bpf_rcu_read_unlock();
 		return nsid == match->nsid;
 	}
+	case MATCH_NS_FLAGGED: {
```
htejun, line 1802: Hmm... why is this "namespace flagged"?

etsal: I was thinking of making this rule more generic but didn't end up implementing it, same reason why the flag is called  "flagged" in the layer sturct.  It makes no sense for the patch as it is right now, I will rename everything to make it clear what is going on.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -432,6 +435,34 @@ static struct task_ctx *lookup_task_ctx(struct task_struct *p)
...
+	if (!taskc)
+		return -EINVAL;
+
+	taskc->is_renamed = true;
+	taskc->refresh_layer = true;
+
+	return 0;
+}
+
+
```
JakeHillion, line 465: This is covered by L532 now right?

```suggestion
```

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1768,6 +1802,14 @@ static __noinline bool match_one(struct layer_match *match,
 		bpf_rcu_read_unlock();
 		return nsid == match->nsid;
 	}
+	case MATCH_IS_RENAMED: {
+		struct task_ctx *taskc;
+		taskc = lookup_task_ctx_may_fail(p);
+		if (!taskc)
+			return false;
+
```
JakeHillion, line 1810: Covered by the next line, unless it's needed to appease the verifier...

```suggestion
```

JakeHillion (commented):
Looks good! Is it tested & have the weird crashes gone? I'd hope this is pretty harmless now.

JakeHillion (approved):
Nice, thanks for addressing the feedback!

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -233,6 +233,7 @@ enum layer_match_kind {
 	MATCH_TGID_EQUALS,
 	MATCH_NSPID_EQUALS,
 	MATCH_NS_EQUALS,
+	MATCH_IS_RENAMED,
```
htejun, line 236: Would `MATCH_ON_RENAME` be better?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -496,8 +499,11 @@ int BPF_PROG(tp_task_rename, struct task_struct *p, const char *buf)
 {
 	struct task_ctx *taskc;
 
-	if ((taskc = lookup_task_ctx_may_fail(p)))
+	if ((taskc = lookup_task_ctx_may_fail(p))) {
 		taskc->refresh_layer = true;
+		taskc->is_renamed = true;
```
htejun, line 504: I wonder whether it should capture `comm` here for matching. Imagine a scenario like the following:
- User defines an edge triggered rule matching a special token.
- User renames the thread temporarily to the token to put the task in the layer.
- User restores the original thread name.
- The task gets enqueued and matches the restored name.

`comm` is only 16 bytes long, so while the copy buffer would add some overhead it's not prohibitive. Alternatively, we can run the matching here too. Note that supporting the above means that the rule would need to match its own string.

hodgesds: > Alternatively, we can run the matching here too.

Is there a downside to matching on rename? If it's nothing major then that'd be my vote.

etsal: > I wonder whether it should capture `comm` here for matching. Imagine a scenario like the following:
> 
> * User defines an edge triggered rule matching a special token.
> * User renames the thread temporarily to the token to put the task in the layer.
> * User restores the original thread name.
> * The task gets enqueued and matches the restored name.
> 
> `comm` is only 16 bytes long, so while the copy buffer would add some overhead it's not prohibitive. Alternatively, we can run the matching here too. Note that supporting the above means that the rule would need to match its own string.

Makes sense, I assume we want this to be edge triggered so only the first rename will be considered, then never reset? 

Also, do we always know what the task is being renamed to? I guess if we don't we can have the rule be "have the empty string as a prefix".

scheds/rust/scx_layered/src/bpf/intf.h
etsal: If we only consider the first rename, MATCH_ON_RENAME sounds like we match on every rename; which one do we prefer? MATCH_IS_RENAMED was meant to convey that the task has been renamed at least once.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: Hmmm... I was thinking edge triggered on each rename. ie. the latest rename should be the determinant one.

htejun: @hodgesds the only difference is the overhead of storing the name, I think.

etsal: If we match on each rename then, can't we just trigger taskc->refresh_layer and call maybe_refresh_layer? Then we don't need to store the comm (AFAICT that's what Dan is also proposing).

etsal: Also, rn in the code we already mark the task for a layer refresh on a name change, so what would we be gaining by triggering the refresh on the rename? The only difference is we will be moving the task to a different layer slightly sooner.

htejun: Let's say the user wants the thread name to be "group1" and there will be many threads with that name, but this one thread among them is special and should be put into its own layer. One way to do that could be defining an on-rename rule for, say, "IAMSPECIAL" and then the thread can rename itself first to "IAMSPECIAL" and then to "group1" expecting the edge-triggered to match on rename. If we don't store the name on rename, by the time matching runs, the name might have gone back to "group1".

htejun: > Also, rn in the code we already mark the task for a layer refresh on a name change, so what would we be gaining by triggering the refresh on the rename? The only difference is we will be moving the task to a different layer slightly sooner.

It's the same. Users would expect an edge-triggered rule to apply to the name that triggered that edge. If we don't store the name or match synchronously, we may lose the name that triggered the match.

etsal: In that case, if the thread gets renamed "group1" --> "IAMSPECIAL", we put it on its own layer, then it renames itself back "IAMSPECIAL" --> "group1" -  this counts as a second edge and we put it back into its original layer. Or do we write down the comm only on the first edge we see? That would be equivalent to the "has_been_renamed" flag we had in the first place, only with the added comm so that threads renaming themselves to different names can be put on their own special layers.

We can also make "sticky" layers from which tasks cannot be removed once entered. In that case we trigger the edge, run the match rule and insert the task to the layer, and subsequent renames do not affect it.

htejun: So, the target layer would have an or rule for "IAMSPECIAL" which only triggers on renames to "IAMSPECIAL" and the rename back wouldn't matter.

etsal:
New version that stores the name triggering the rule. I think that on a subsequent name change after the trigger, the task will be removed from the layer because comm_renamed will be overwritten; we can add a special case that a task in a RENAMED layer should not be removed from it even if it matches a COMM rule.

likewhatevs:
> New version that stores the name triggering the rule. 

This makes sense. My thinking w/ this was throw task struct pointers serialized as u64's (i.e. to avoid pid/tgid/ppid fun) that matched the rename rule into a BPF map that would live forever.

> I think that on a subsequent name change after the trigger, the task will be removed from the layer because comm_renamed will be overwritten; we can add a special case that a task in a RENAMED layer should not be removed from it even if it matches a COMM rule.

That makes sense to me and is the same as how I kinda thought this could be done above.

etsal:
New version that incorporates using thread rename operations to pass commands to the scheudler, as requested by @htejun. Right now we have two commands: One sets a string that lets us join a layer, and one to unset it.

Note: This requires the threads to explicitly request to be moved to layers. This means we are no longer edge triggered on all renames.

htejun (approved):

PR #1204, opened 2025-01-16
scx_layered: Add matcher for when tgid is or is not equal to pid

likewhatevs:
Add a matcher, `TgidPidEq`, which accepts a c boolean (i.e. 0 is false, non-zero is true), for when TGID is (or is not) equal to PID.

The intent is to be able to use this matcher to express either of the following:
* "Scope matched tasks to those where PID is TGID"
* "Scope matched tasks to those where PID is NOT TGID"

in combination with other matchers (i.e. `CommPrefix`).

Need to write a cfg to test this still.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -243,6 +243,7 @@ enum layer_match_kind {
 	MATCH_SCXCMD_JOIN,
 
 	NR_LAYER_MATCH_KINDS,
+	MATCH_PID_TGID_EQUALS,
 };
 
 struct layer_match {
@@ -257,6 +258,7 @@ struct layer_match {
 	u32		ppid;
 	u32		tgid;
 	u64		nsid;
+	u32		pid_tgid_eq;
 };
 
 struct layer_match_ands {
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -241,9 +241,9 @@ enum layer_match_kind {
 	MATCH_NSPID_EQUALS,
 	MATCH_NS_EQUALS,
 	MATCH_SCXCMD_JOIN,
+	MATCH_IS_GROUP_LEADER,
 
 	NR_LAYER_MATCH_KINDS,
-	MATCH_IS_GROUP_LEADER,
 };
 
 struct layer_match {
@@ -258,7 +258,7 @@ struct layer_match {
 	u32		ppid;
 	u32		tgid;
 	u64		nsid;
-	bool	is_group_leader;
+	bool		is_group_leader;
 };
 
 struct layer_match_ands {
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -243,7 +243,7 @@ enum layer_match_kind {
 	MATCH_SCXCMD_JOIN,
 
 	NR_LAYER_MATCH_KINDS,
-	MATCH_PID_TGID_EQUALS,
+	MATCH_IS_GROUP_LEADER,
 };
 
 struct layer_match {
@@ -258,7 +258,7 @@ struct layer_match {
 	u32		ppid;
 	u32		tgid;
 	u64		nsid;
-	u32		pid_tgid_eq;
+	bool	is_group_leader;
 };
 
 struct layer_match_ands {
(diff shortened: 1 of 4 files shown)


etsal (approved):
Looks good, only nits I have are with the naming:

- We could possibly rename  PID_TGID_EQUALS to something more descriptive like IS_MAIN_PROCESS_THREAD but I'm 50-50 on whether it's worth it. 
- Can we rename polarity to should_be_equal or something equivalent?

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -249,6 +250,7 @@ struct layer_match {
 	u32		ppid;
 	u32		tgid;
 	u64		nsid;
+	u32		pid_tgid_eq;
```
hodgesds, line 253: should this be a bool?

hodgesds:
> We could possibly rename PID_TGID_EQUALS to something more descriptive like IS_MAIN_PROCESS_THREAD but I'm 50-50 on whether it's worth it.

`PID_IS_TGID` is probably most descriptive.

scheds/rust/scx_layered/src/bpf/intf.h
likewhatevs: bool's a macro not a type, the ffi w/ this looked weird.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1768,6 +1768,11 @@ static __noinline bool match_one(struct layer_match *match,
 		bpf_rcu_read_unlock();
 		return nsid == match->nsid;
 	}
+	case MATCH_PID_TGID_EQUALS: {
+		// There is nuance to this around exec(2)s and group leader swaps.
+		// See https://github.com/sched-ext/scx/issues/610 for more details.
+		return p->tgid == p->pid && match->pid_tgid_eq;
```
htejun, line 1774: I wonder whether the better name for this match is something like `is_group_leader`.

scheds/rust/scx_layered/src/bpf/intf.h
htejun: I'd prefer `bool` too. How is ffi weird?

scheds/rust/scx_layered/src/config.rs
```diff
@@ -73,6 +73,7 @@ pub enum LayerMatch {
     TGIDEquals(u32),
     NSPIDEquals(u64, u32),
     NSEquals(u32),
+    TgidPidEq(u32),
```
htejun, line 76: Here too, why not bool?

scheds/rust/scx_layered/src/bpf/intf.h
likewhatevs: It picks a funny type for bool (`MaybeUninit<bool>`, vs just the primitive type for `u32`). That being said playing with it a bit, it works as well as `u32`. Changed to bool.

likewhatevs:
I made the cfg parameter a bool, renamed the flag to is_group_leader, and tested via the following:

no scx:
Screenshot from 2025-01-19 02-19-59

scx with test cfg:
Screenshot from 2025-01-19 02-20-54

test cfg:
https://gist.github.com/likewhatevs/76f5caa47000d794d7a421c4859ac477

test program:
https://gist.github.com/likewhatevs/97bf407067c70c87d55791e3b0769cc6

The test results look close enough to what I expect that I think this change is good, but far enough from what I expect that I think maybe my test cfg is bad or there's something off somewhere else (i.e. why are 3 CPUs (and not 2 CPUs) pegged w/ a cfg like that)?

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -243,6 +243,7 @@ enum layer_match_kind {
 	MATCH_SCXCMD_JOIN,
 
 	NR_LAYER_MATCH_KINDS,
+	MATCH_IS_GROUP_LEADER,
```
hodgesds, line 246: `NR_LAYER_MATCH_KINDS`  has to come last because it's used to check bounds on the match enum.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -257,6 +258,7 @@ struct layer_match {
 	u32		ppid;
 	u32		tgid;
 	u64		nsid;
+	bool	is_group_leader;
```
hodgesds, line 261: alignment?

hodgesds (approved):
LGTM

PR #1215, opened 2025-01-18
scx_layered: Implement `allow_node_aligned` layer property

htejun:
While cpu allocations aren't node-aware yet, an open layer can work fine with tasks with a node aligned custom cpumask as long as the machine doesn't stay fully saturated for a long time. Add allow_node_aligned layer property which puts node-aligned tassk into layer DSQs instead of the lo fallbacks.

This isn't by any chance a full-fledged NUMA support but setting `allow_node_aligned` on an open layer restores the old behavior which assumed that an open layer can almost always consume from all nodes on the system. This can still lead to stalls but can be useful in specific situations. This should be deprecated once proper numa support is implemented.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -286,6 +286,7 @@ struct layer {
 	bool			preempt;
 	bool			preempt_first;
 	bool			exclusive;
+	bool			allow_node_aligned;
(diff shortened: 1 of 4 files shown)


likewhatevs (approved):
makes sense

hodgesds (approved):
Seems legit, we should be able to track the migrations as well.

etsal (approved):

Issue #1214, opened 2025-01-17
scx_lavd: got an scx_bpf_error() while playing CyberPunk 2077


multics69:
The problem is reported by @yuby64

```
❯ sudo scx_lavd --balanced
18:15:22 [INFO] scx_lavd scheduler is initialized (build ID: 1.0.8-g8313b752 x86_64-unknown-linux-gnu)
18:15:22 [INFO] scx_lavd scheduler starts running.

DEBUG DUMP
================================================================================

rcub/0[19] triggered exit kind 1025:
  scx_bpf_error (run_time_ns is negative: 0x5772484607211 - 0x5772484608241)

Backtrace:
  scx_bpf_error_bstr+0x65/0x80
  bpf_prog_24288482c3037da4_lavd_stopping+0x214/0x248
  bpf__sched_ext_ops_stopping+0x4b/0xa8
  dequeue_task_scx.llvm.15304548555706679011+0xd8/0x270
  rt_mutex_setprio+0x153/0x460
  task_blocks_on_rt_mutex+0x385/0x3f0
  rt_mutex_slowlock+0x113/0x290
  rcu_boost_kthread+0x126/0x330
  kthread+0x14a/0x170
  ret_from_fork+0x37/0x50
  ret_from_fork_asm+0x1a/0x30

CPU states
----------

CPU 0   : nr_run=1 flags=0x9 cpu_rel=1 ops_qseq=4930858 pnt_seq=15980341
          curr=rcub/0[19] class=rt_sched_class

  R redDispatcher6[28815] +0ms
      scx_state/flags=3/0x9 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x0 dsq_vtime=1596513577 slice=5000000
      cpus=ffff

    bpf_trampoline_6442538215+0x41/0xe8
    futex_wait_multiple+0x9/0x170
    __se_sys_futex_waitv+0x2a7/0x330
    do_syscall_64+0x88/0x170
    entry_SYSCALL_64_after_hwframe+0x76/0x7e

CPU 1   : nr_run=2 flags=0x1 cpu_rel=0 ops_qseq=4818417 pnt_seq=17045343
          curr=redDispatcher3[28812] class=ext_sched_class

 *R redDispatcher3[28812] -1ms
      scx_state/flags=3/0xd dsq_flags=0x0 ops_state/qseq=0/0
[...]

PR #1216, opened 2025-01-18
scx_lavd: Use time_delta() for clock difference calculation

multics69:
As time_delta() is added to the common code, use time_delta() to calculate the clock difference.

This fixes the following issue:
    https://github.com/sched-ext/scx/issues/1214


multics69:
Needs some more testing, so close the PR for now.

PR #1217, opened 2025-01-18
scxtop: proposing to add bus-cycles event.

devnexen:
if the hardware does not have PMUs, the graph just does not provide any data.
Can have its use for memory intensive tasks for example.


hodgesds (approved):
Nice!

PR #1219, opened 2025-01-18
meson: Install rust projects in tools dir

### LINKED ISSUE #1182: scxtop: add to meson.build
Maybe we should consider adding scxtop to meson.build? It's a very useful tool and it would be good to add it to the package by default.

frelon:
Update meson-scripts/install_rust_user_scheds to also install
rust projects found in the tools/ directory.

This will install scxtop when running 'meson install'.

Fixes #1182


hodgesds (approved):
LGTM

PR #1223, opened 2025-01-19
scx_lavd: Optimize the use of idle CPUs

multics69:
This PR optimizes the use of idle CPUs in ops.select_cpu() and ops.enqueue() paths. It harvests opportunities for direct dispatch in ops.enqueue() when an ops.select_cpu() is skipped. Also, the ops.enqueue() path proactively kicks the idle CPU to reduce the latency, but it does not mark it busy since there is no guarantee that the found idle CPU schedules the task.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/util.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/util.bpf.c
@@ -227,6 +227,11 @@ static bool is_migration_disabled(const struct task_struct *p)
 	return false;
 }
 
+static bool have_idle_cpus(const struct cpumask *idle_mask)
+{
+	return !bpf_cpumask_empty(idle_mask);
+}
+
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_lavd/src/bpf/util.bpf.c
```diff
@@ -227,6 +227,25 @@ static bool is_migration_disabled(const struct task_struct *p)
 	return false;
 }
 
+static __always_inline
+const struct cpumask *get_idle_cpumask(void)
+{
+	if (is_smt_active)
+		return scx_bpf_get_idle_smtmask();
```
arighi, line 234: Hm... but in this way when `is_smt_active == true` we'll only consider full-idle cores and never allocate idle CPUs with a thread running. Is this intended? Maybe we want to check the idle smtmask and, if all the cores are busy, use idle cpumask?

PR #1226, opened 2025-01-20
docs: Update tumbleweed installation steps

frelon:
The scx package was accepted into Factory and is able to be installed
without adding any repositories.

Signed-off-by: Fredrik Lönnegren <fredrik@frelon.se>


sirlucjan (approved):

arighi (approved):
Awesome, thanks!

PR #1228, opened 2025-01-21
scx_layered: initial attempt at adding gpu layer match filter

likewhatevs:
scx_layered: initial attempt at adding gpu layer match filter

I figure once this works/we use it a bit, we'll know a decent interface (and impl. tbh) and can pull it out into scx_utils.

ATM I've got pids making it through to where I try to write it to the bpf.

I kind of wanted use user ringbuf's for this and started trying to use tests to figure out how to do that, but the docs scared me off a bit: https://docs.ebpf.io/linux/map-type/BPF_MAP_TYPE_USER_RINGBUF/

I'm not sure if how I'm doing this would work at all or if there's a bug in it atm causing pids to not get to bpf side (nor if this approach would be performant enough to be usable if it worked, tbh).

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -64,6 +64,7 @@ enum consts {
 
 	SCXCMD_PREFIX		= 0x5C10,
 	SCXCMD_COMLEN		= 13,
+	MAX_GPU_PIDS 		= 100000,
 };
 
 static inline void ___consts_sanity_check___(void) {
@@ -242,6 +243,8 @@ enum layer_match_kind {
 	MATCH_NS_EQUALS,
 	MATCH_SCXCMD_JOIN,
 	MATCH_IS_GROUP_LEADER,
+	MATCH_USING_GPU,
+	MATCH_USED_GPU,
 
 	NR_LAYER_MATCH_KINDS,
 };
@@ -259,6 +262,8 @@ struct layer_match {
 	u32		tgid;
 	u64		nsid;
 	bool		is_group_leader;
+	bool	using_gpu;
+	bool	used_gpu;
 };
 
 struct layer_match_ands {
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -182,7 +182,7 @@ struct {
 	__type(key, u32);
 	__type(value, u32);
 	__uint(max_entries, MAX_GPU_PIDS);
-	__uint(map_flags, 0);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
 } all_gpu_pid SEC(".maps");
 
 
@@ -192,7 +192,7 @@ struct {
 	__type(key, u32);
 	__type(value, u32);
 	__uint(max_entries, MAX_GPU_PIDS);
-	__uint(map_flags, 0);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
 } cur_gpu_pid SEC(".maps");
 
 // XXX - Converting this to bss array triggers verifier bugs. See
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -54,6 +54,7 @@ const volatile u32 lo_fb_share_ppk = 128;	/* !0 for veristat */
 
 /* Flag to enable or disable antistall feature */
 const volatile bool enable_antistall = true;
+const volatile bool enable_gpu_support = false;
 /* Delay permitted, in seconds, before antistall activates */
 const volatile u64 antistall_sec = 3;
 const u32 zero_u32 = 0;
@@ -173,6 +174,26 @@ static bool cpuc_in_layer(struct cpu_ctx *cpuc, struct layer *layer)
 	else
 		return cpuc->layer_id == layer->id;
 }
+
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	// u32 comes from nvml
+	__type(key, u32);
+	__type(value, u32);
+	__uint(max_entries, MAX_GPU_PIDS);
+	__uint(map_flags, 0);
+} all_gpu_pid SEC(".maps");
+
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	// u32 comes from nvml
+	__type(key, u32);
+	__type(value, u32);
+	__uint(max_entries, MAX_GPU_PIDS);
+	__uint(map_flags, 0);
+} cur_gpu_pid SEC(".maps");
 
 // XXX - Converting this to bss array triggers verifier bugs. See
 // BpfStats::read(). Should also be cacheline aligned which doesn't work with
@@ -1841,6 +1862,39 @@ static __noinline bool match_one(struct layer_match *match,
 		// See https://github.com/sched-ext/scx/issues/610 for more details.
 		return (p->tgid == p->pid) == match->is_group_leader;
 	}
+	case MATCH_USING_GPU: {
+			u32 pid;
+			u32 gpu_pid;
+			bool pid_present = false;
+
+			if (!enable_gpu_support)
+				return match->used_gpu;
+
+			pid = p->pid;
+			gpu_pid = bpf_map_lookup_elem(&cur_gpu_pid, &pid) == 0;
+
+			if (gpu_pid)
+				pid_present = true;
+
+			return pid_present == match->used_gpu;
+	}
+	case MATCH_USED_GPU: {
+			u32 pid;
+			u32 gpu_pid;
+			bool pid_present = false;
+
+			if (!enable_gpu_support)
+				return match->used_gpu;
+
+			pid = p->pid;
+			gpu_pid = bpf_map_lookup_elem(&all_gpu_pid, &pid) == 0;
+
+			if (gpu_pid)
+				pid_present = true;
+
+			return pid_present == match->used_gpu;
+	}
+
 	default:
 		scx_bpf_error("invalid match kind %d", match->kind);
 		return result;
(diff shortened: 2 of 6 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -67,6 +68,9 @@ u32 layered_root_tgid = 0;
 u32 empty_layer_ids[MAX_LAYERS];
 u32 nr_empty_layer_ids;
 
+u32 all_gpu_pids[MAX_GPU_PIDS];
+u32 cur_gpu_pids[MAX_GPU_PIDS];
```
hodgesds, line 72: You probably want a variable to track `nr_gpu_pids` as well so that indexing into the array on the bpf side is possible.

Another thought is it is probably also worth storing the GPU/NUMA node mappings as well on the bpf side. That way on the bpf side we know what resources the pid is using in relation to the topology.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1870,6 +1896,84 @@ impl<'a> Scheduler<'a> {
...
+
+        let device_count = device_count_r.unwrap();
+
+        let mut pids = HashSet::new();
+
+        // Iterate over all devices and gather running processes
+        for i in 0..device_count {
+            let device_r = nvml.device_by_index(i);
+            if device_r.is_err() {
+                warn!("err getting gpu");
```
hodgesds, line 1925: nit: throw the index as well, might be useful for debugging multiple gpus.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1870,6 +1896,84 @@ impl<'a> Scheduler<'a> {
...
+                continue;
+            }
+            let device = device_r.unwrap();
+
+            let procs_r = device.running_compute_processes();
```
hodgesds, line 1930: We could add to the procs hashset the graphics processes as well? wdyt?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1870,6 +1896,84 @@ impl<'a> Scheduler<'a> {
...
+        if !self.opts.enable_gpu_support {
+            return;
+        }
+        let nvml_r = NVML_CELL.get_or_try_init(|| Nvml::init());
+
+        if nvml_r.is_err() {
+            return;
+        }
+
+        let nvml = nvml_r.unwrap();
```
hodgesds, line 1909: since this is called in `step` it might be nice to return a `Result`.

etsal:
> scx_layered: initial attempt at adding gpu layer match filter
> 
> I figure once this works/we use it a bit, we'll know a decent interface (and impl. tbh) and can pull it out into scx_utils.
> 
> ATM I've got pids making it through to where I try to write it to the bpf.
> 
> I kind of wanted use user ringbuf's for this and started trying to use tests to figure out how to do that, but the docs scared me off a bit: https://docs.ebpf.io/linux/map-type/BPF_MAP_TYPE_USER_RINGBUF/
> 
> I'm not sure if how I'm doing this would work at all or if there's a bug in it atm causing pids to not get to bpf side (nor if this approach would be performant enough to be usable if it worked, tbh).

Wrt ringbufs and user-kernel communication, we can replace with an arena-backed region later once we get this landed.

likewhatevs:
> Wrt ringbufs and user-kernel communication, we can replace with an arena-backed region later once we get this landed.

That makes sense. Writing to bss_data like this should work though, right?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1843,6 +1847,38 @@ static __noinline bool match_one(struct layer_match *match,
...
+					break;
+				if (cur_gpu_pids[i] == p->pid) {
+					pid_match = true;
+					break;
+				}
+			}
+			// to see why bpf side isn't seeing pids.
+			if(pid_match)
+				scx_bpf_error("error");
+			return pid_match && match->is_using_gpu;
```
etsal, line 1865: Should this not be 

```
return pid_match == match->is_using_gpu;```

? Otherwise leaving is_using_gpu just prevents the rule from ever being matched and we can't set the polarity as we describe in the Rust part.

File: scheds/rust/scx_layered/src/bpf/main.bpf.c [hunk 6]
```diff
@@ -1843,6 +1847,38 @@ static __noinline bool match_one(struct layer_match *match,
...
+			bool pid_match = false;
+			for(int i = 0; i < MAX_GPU_PIDS; i++) {
+				if (all_gpu_pids[i] == 0)
+					break;
+				if (all_gpu_pids[i] == p->pid) {
+					pid_match = true;
+					break;
+				}
+			}
+			return pid_match && match->is_or_has_used_gpu;
```
etsal, line 1879: Same here

etsal:
> > Wrt ringbufs and user-kernel communication, we can replace with an arena-backed region later once we get this landed.
> 
> That makes sense. Writing to bss_data like this should work though, right?

Yeah it should be fine, especially since we overwrite it in every period so we don't have to worry about going over the size of the array.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: right, I need to fix that in tgid/pid matcher also, thx!

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1843,6 +1847,38 @@ static __noinline bool match_one(struct layer_match *match,
...
 		return p->tgid == p->pid && match->is_group_leader;
 	}
+	case MATCH_IS_USING_GPU: {
+			if (!enable_gpu_support)
+				return match->is_using_gpu;
+			bool pid_match = false;
+			for(int i = 0; i < MAX_GPU_PIDS; i++) {
+				if (cur_gpu_pids[i] == 0)
+					break;
+				if (cur_gpu_pids[i] == p->pid) {
```
htejun, line 1857: Why not use a hashmap?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1843,6 +1847,38 @@ static __noinline bool match_one(struct layer_match *match,
...
+					pid_match = true;
+					break;
+				}
+			}
+			// to see why bpf side isn't seeing pids.
+			if(pid_match)
```
htejun, line 1863: `if (` here and other places.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -242,6 +243,8 @@ enum layer_match_kind {
 	MATCH_NS_EQUALS,
 	MATCH_SCXCMD_JOIN,
 	MATCH_IS_GROUP_LEADER,
+	MATCH_IS_USING_GPU,
+	MATCH_IS_OR_HAS_USED_GPU,
```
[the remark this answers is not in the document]
likewhatevs: MATCH_USING_GPU / MATCH_USED_GPU makes it sound like used and using are disjoint (but used is really a superset of using). I couldn't think of any shorter names that avoid being unclear in that particular manner.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: I *think* you have to use `BPF_MAP_TYPE_USER_RINGBUF` to write to write to a map user space side and read from a that map bpf side. 

I'm gonna try using UserRingbuf again. I couldn't find an example of using that outside of the tests for it so I was trying to avoid it.

arighi: IIUC the list of pids is generated in user-space and the BPF part is the consumer. In this case you can definitely use a `BPF_MAP_TYPE_HASH` and you can update the items via libbpf-rs.

If you need an example of how to use `BPF_MAP_TYPE_USER_RINGBUF` you can take a look at https://github.com/sched-ext/scx/blob/dfe03fea83ba53d60884051f97d13fc2413a3f23/rust/scx_rustland_core/assets/bpf/main.bpf.c#L154, however in this case I think you can simply use a hash map.

likewhatevs:
btw @htejun you called it earlier w/ nvidia-smi/nvml not returning what we need (user space pid).

This would return thread ID: https://developer.nvidia.com/cupti-ctk12_2 , so I'm gonna swap out nvml w/ some bindgen around that and then I *think* this should work (provided there's no in-process initialization/opt-in from observed processes required, which afaict there isn't).

likewhatevs:
ahh looked at cupti a bit more, more optimistic about finding some place to add a tracepoint, will try that.

likewhatevs:
I need to go over feedback and clean things up but I think this works now (which is great, learning!),

Using a config to restrict GPU processes with the CommPrefix python to 2 cores yielded the following:
Screenshot from 2025-01-26 21-15-47

Flipping the polarity on that match (i.e. those two cores can only run python processes which have not used GPUs) seems to work also (i.e. the python threads all get pushed to the open/not core layer):
Screenshot from 2025-01-26 21-18-22

This doesn't do what I had initially hoped for (matching threads, not pids), but it might be enough for the moment.

etsal (approved):

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -64,6 +64,7 @@ enum consts {
 
 	SCXCMD_PREFIX		= 0x5C10,
 	SCXCMD_COMLEN		= 13,
+	MAX_GPU_PIDS 		= 100000,
```
likewhatevs, line 67: I think MAX_GPU_PIDS = 100000 is reasonable, not sure about that though.

likewhatevs: added __uint(map_flags, BPF_F_NO_PREALLOC); to make that fine.

PR #1231, opened 2025-01-21
scx_utils: Get power profile at every fetch_power_profile() call

multics69:
fetch_power_profile() read the power profile at its first call. Now we change it to read the power profile at every fetch_power_profile() call.

### DIFF
--- a/rust/scx_utils/src/autopower.rs
+++ b/rust/scx_utils/src/autopower.rs
@@ -14,11 +14,10 @@ trait PowerProfiles {
     fn active_profile(&self) -> Result<String>;
 }
 
-static POWER_PROFILES_PROXY: OnceLock<Option<PowerProfilesProxyBlocking<'static>>> =
-    OnceLock::new();
-
 pub fn fetch_power_profile() -> String {
-    POWER_PROFILES_PROXY
+    let power_profile: OnceLock<Option<PowerProfilesProxyBlocking<'static>>> = OnceLock::new();
+
+    power_profile
         .get_or_init(|| {
             let system_bus = Connection::system().ok()?;
             let system_bus = Box::leak(Box::new(system_bus));


multics69:
@ruihe774 -- Please take a look if you are available.

hodgesds (approved):
Seems reasonable!

rust/scx_utils/src/autopower.rs
```diff
@@ -14,11 +14,10 @@ trait PowerProfiles {
     fn active_profile(&self) -> Result<String>;
 }
 
-static POWER_PROFILES_PROXY: OnceLock<Option<PowerProfilesProxyBlocking<'static>>> =
-    OnceLock::new();
-
 pub fn fetch_power_profile() -> String {
-    POWER_PROFILES_PROXY
+    let power_profile: OnceLock<Option<PowerProfilesProxyBlocking<'static>>> = OnceLock::new();
```
ruihe774, line 18: Doesn't it mean we need to reconnect to dbus at every call to `fetch_power_profile()`?

`zbus` actually caches properties. It listens to property change event. If the property is no changed, it does not make a dbus call.

The problem of the current implementation may be that the dbus or ppd daemon is not yet started when we first time call `fetch_power_profile()`. This dependency requirement can be enforced by setting in the service file.

multics69: > Doesn't it mean we need to reconnect to dbus at every call to `fetch_power_profile()`?

Right. The change re-establishes the dbus connection every function call.

> `zbus` actually caches properties. It listens to property change event. If the property is no changed, it does not make a dbus call.

Yes, that's the expected behavior but ... hmm..

> The problem of the current implementation may be that the dbus or ppd daemon is not yet started when we first time call `fetch_power_profile()`. This dependency requirement can be enforced by setting in the service file.

In my testing, dbus should be started already before starting the scx_lavd. The test scenario is as follows:
   1. Login in to any regular desktop (GNOME in my case)
   2. Launch the scx_lavd in a terminal with an option `--autopower -v`
   3. Change the power mode from GNOME UI and see scx_lavd prints out the proper log.

Could you take a look what the problem is in your implementation? @ruihe774

ruihe774: > In my testing, dbus should be started already before starting the scx_lavd. The test scenario is as follows:
> 
>     1. Login in to any regular desktop (GNOME in my case)
> 
>     2. Launch the scx_lavd in a terminal with an option `--autopower -v`
> 
>     3. Change the power mode from GNOME UI and see scx_lavd prints out the proper log.
> 
> 
> Could you take a look what the problem is in your implementation? @ruihe774

I tested it in the exact same way. Weird. Let me investigate it later this week.

multics69:
Close this since https://github.com/sched-ext/scx/pull/1245 should resolve the problem.

PR #1232, opened 2025-01-21
scx_layered: fixes for MATCH_SCXCMD_JOIN

etsal:
Add three fixes for the MATCH_SCXCMD_JOIN command:

- Remove stray printf when processing a user command
- Add debug printf command for the layer type during layer initialization
- Use the right enum when initializing the layer from the Rust side


hodgesds (approved):
LGTM

PR #1234, opened 2025-01-21
meson.build: enable o3 by default

JohnRTitor:
Should be safe to enable. bpf compilation used `-o2` by default, that is changed to `-o3` to allow for better optimization.

CC @JakeHillion @htejun


htejun:
I'd rather not change the optimization level for compiling BPF.

PR #1235, opened 2025-01-21
scx_layered: Document the CmdJoin layer

etsal:
Document the CmdJoin layer type and add a usage example.

### DIFF
--- a/scheds/rust/scx_layered/examples/cmdjoin.json
+++ b/scheds/rust/scx_layered/examples/cmdjoin.json
@@ -1,0 +1,34 @@
+[
+  {
+    "name": "cmd",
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/examples/cmdjoin.json
```diff
@@ -0,0 +1,37 @@
...
+        }
+      ]
+    ],
+    "kind": {
+      "Confined": {
+        "util_range": [
+          0.4,
+          0.90
+        ],
+        "llcs": [
```
hodgesds, line 18: Can probably remove this as I don't think this config cares about which LLC.

hodgesds (approved):
LGTM

PR #1236, opened 2025-01-21
scx_layered: add HAS_BEEN_RENAMED matching rule

etsal:
Add a rule that tests whether a task has been renamed using pthread_rename_np() or through /proc. This has two uses:
- Using task renames to identify specific tasks in a large application, without requiring modifications for that application.
- Preventing tasks from renaming themselves to match into a more privileged layer for a performance boost.

### DIFF
--- a/scheds/rust/scx_layered/examples/renamed.json
+++ b/scheds/rust/scx_layered/examples/renamed.json
@@ -1,0 +1,83 @@
+[
+  {
+    "name": "renamed_to_stress_ng",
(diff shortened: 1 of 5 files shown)


htejun:
Hmm... this is rather trivial to defeat, so I'm not sure how meaningful this is. There are other gating mechanisms already (e.g. cgroup membership), wouldn't that be sufficient?

likewhatevs:
I think this is more about covering a case that cmdjoin doesn't cover:

* I don't know the name of the threads I care about (or don't always know), but I do know that rename will be called on it.

PR #1237, opened 2025-01-21
scxtop: Fill render area by setting max events

### LINKED ISSUE #1167: scxtop: Make max_cpu_events scale to fit window
### Issue
From the following screenshot in the right panel the perf event pane displays the cycles. However, the sparkline does not fill the entire panel because the [`max_cpu_events`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/app.rs#L49) is not dynamic.
image
In `scxtop` various aggregations (ex: LLC/CPU) use the [`EventData`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/event_data.rs#L12) struct for storing event data. This should be sized dynamically so that the number of events fills the window size. This probably will require using the [`ratatui::layout::Rect`](https://docs.rs/ratatui/latest/ratatui/layout/struct.Rect.html) to calculate the appropriate size. 

### How to reproduce
```
cargo build --release
# run scheduler of choice
sudo ./target/release/scxtop
# press "v" to toggle sparkline view
```
If the issue doesn't appear shrink the terminal until the length doesn't fill various panes in the sparkline view. This is an issue on different view contexts (llc, node etc) so the ideal solution should work for all contexts.

### What you'll learn
This is purely a Rust issue, so it should be something that someone new to the project could pick up. You'll learn a little about how `scxtop` can be used to observe schedulers and learn a little Rust TUI handling. There's still a lot of work/cleanup that needs to be done for `scxtop`, but this is a good starter task.

hodgesds:
Scale the max number of events by setting the max number of events based on the rendered area. This ensures that events fill the window space when rendering. fixes #1167

image

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -52,6 +52,7 @@ pub struct App<'a> {
     keymap: KeyMap,
     scheduler: String,
     max_cpu_events: usize,
+    max_sched_events: usize,
     state: AppState,
     prev_state: AppState,
     theme: AppTheme,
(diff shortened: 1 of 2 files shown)


likewhatevs (approved):
nice

Issue #1238, opened 2025-01-22
scxtop: Add support for scheduler stats


hodgesds:
Create a view that is able to read scheduler stats and render them in a structured way. See the [`scx_stats`](https://github.com/sched-ext/scx/tree/main/rust/scx_stats) crate for details on the the protocol and client implementation. The scheduler stats could be displayed as part of the [`render_scheduler`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/app.rs#L1179) state.

PR #1241, opened 2025-01-24
scxtop: Add scx_stats support

hodgesds:
Add scx_stats support for scxtop to the scheduler view. Implements #1238 

Right now the stats rendering is super basic and just prints the json, but this can be made into more structured output in the future. For schedulers with few numbers of stats the output is reasonable:
image

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -36,19 +36,27 @@ use ratatui::{
     Frame,
 };
 use regex::Regex;
+use scx_stats::prelude::StatsClient;
 use scx_utils::misc::read_file_usize;
 use scx_utils::Topology;
+use serde_json;
+use serde_json::Value as JsonValue;
 use std::collections::BTreeMap;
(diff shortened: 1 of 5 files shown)


tools/scxtop/src/app.rs
```diff
@@ -348,8 +370,51 @@ impl<'a> App<'a> {
...
+        match self.state {
+            AppState::Scheduler => {
+                if !self.scheduler.is_empty() {
+                    let stats_client_read = self.stats_client.clone();
+                    let stats_reconnect = self.stats_reconnect.clone();
+
+                    let tx = self.action_tx.clone();
+                    tokio::spawn(async move {
+                        let mut client = stats_client_read.write().await;
+                        if stats_reconnect.load(Ordering::Relaxed) {
```
hodgesds, line 389: I think if this gets moved into the error handling the bool isn't needed now that I think about it more.

htejun (approved):

PR #1245, opened 2025-01-25
scx_utils: add retries and fallback path for autopower

ruihe774:
This PR adds retries and fallback path for `fetch_power_profile()` to fix issues in the case ppd is not available or not started yet. This is generally a better solution than that in #1231. I also re-enable use of autopower in scx_bpfland.

@multics69 @arighi PTAL

### DIFF
--- a/rust/scx_utils/src/autopower.rs
+++ b/rust/scx_utils/src/autopower.rs
@@ -41,8 +41,8 @@ fn read_energy_profile() -> String {
         .to_string()
 }
 
-pub fn fetch_power_profile() -> String {
-    if std::env::var("SCX_NO_PPD").is_ok_and(|s| s == "1") {
+pub fn fetch_power_profile(no_ppd: bool) -> String {
+    if no_ppd || std::env::var("SCX_NO_PPD").is_ok_and(|s| s == "1") {
         return read_energy_profile();
     }
     let proxy = POWER_PROFILES_PROXY.get();
@@ -61,7 +61,7 @@ pub fn fetch_power_profile() -> String {
             })();
             if let Some(proxy) = proxy {
                 let _ = POWER_PROFILES_PROXY.set(proxy);
-                fetch_power_profile()
+                fetch_power_profile(false)
             } else {
                 warn!("failed to communicate with ppd: retry {retries}");
                 read_energy_profile()
--- a/rust/scx_utils/src/autopower.rs
+++ b/rust/scx_utils/src/autopower.rs
@@ -1,5 +1,4 @@
-use std::fs::File;
-use std::io::Read;
+use std::fs;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::OnceLock;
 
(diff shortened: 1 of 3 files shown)


rust/scx_utils/src/autopower.rs
```diff
@@ -14,18 +19,57 @@ trait PowerProfiles {
...
+                    file.read_to_string(&mut contents)?;
+                    Ok(contents.trim().to_string())
+                })
+                .or_else(|_| Ok("none".to_string()))
+        });
+
+    res.unwrap_or_else(|_: String| "none".to_string())
+}
 
 pub fn fetch_power_profile() -> String {
```
arighi, line 49: Can we have an option to exclude the dbus way and rely directly on the sysfs info? Personally I don't like much adding dependencies to external system components if we can get the same information via sysfs.

arighi (changes requested):

rust/scx_utils/src/autopower.rs
ruihe774: > if we can get the same information via sysfs.

Actually we cannot. It's merely an assumption that `energy_performance_preference` reflects the user's indention on power settings. In some systems there's no `energy_performance_preference` and `scaling_governor` is `scheutil` (this is very common for arm64 cpus), and in some configurations `energy_performance_preference` is raw integer values. The ground truth is always the ppd if the user enables it.



> Personally I don't like much adding dependencies to external system components

It does not *depend* on it. In this PR, if ppd is not enabled, the code goes into the fallback path that reads energy profiles.

> Can we have an option to exclude the dbus way and rely directly on the sysfs info?

I added an envvar option `SCX_NO_PPD` in `03c4417`. Giving autopower is used by multiple schedulers, I think this is simpler than adding options to each scheduler.

arighi: The env var is fine, but wouldn't it be better to have an option like `fetch_power_profile(no_ppd: bool)`, so that schedulers can introduce options like `--no-ppd` to explicitly disable it if they want?

ruihe774: > The env var is fine, but wouldn't it be better to have an option like `fetch_power_profile(no_ppd: bool)`, so that schedulers can introduce options like `--no-ppd` to explicitly disable it if they want?

Done in e19132f.

rust/scx_utils/src/autopower.rs
```diff
@@ -14,18 +18,56 @@ trait PowerProfiles {
...
         })
-        .as_ref()
-        .map(|power_profiles_proxy| power_profiles_proxy.active_profile().ok())
-        .flatten()
         .unwrap_or_default()
+        .to_string()
+}
+
+pub fn fetch_power_profile(no_ppd: bool) -> String {
+    if no_ppd || std::env::var("SCX_NO_PPD").is_ok_and(|s| s == "1") {
```
arighi, line 45: Sorry for nitpicking, but with the `no_ppd` option isn't the `SCX_NO_PPD` env variable redundant? Does it bring any benefit having an env variable to override the option?

That aside, LGTM.

arighi (approved):

etsal:
This is causing the scx_bpfland CI tests to fail.

arighi:
The CI is probably unhappy about this:
```
00:00:32 [WARN] failed to communicate with dbus (retry 0): I/O error: No such file or directory (os error 2)
1028
```

I'd be better to let the caller decide whether to print these warnings, instead of having them printed by the autopower crate. Also, there are people running bpfland in servers or inside containers and it's totally ok to not have dbus in those systems.

Maybe we can pass the warning to the caller (somehow) or we add a `verbose: bool` option to print them?

ruihe774:
> The CI is probably unhappy about this:
> 
> ```
> 00:00:32 [WARN] failed to communicate with dbus (retry 0): I/O error: No such file or directory (os error 2)
> 1028
> ```
> 
> I'd be better to let the caller decide whether to print these warnings, instead of having them printed by the autopower crate. Also, there are people running bpfland in servers or inside containers and it's totally ok to not have dbus in those systems.
> 
> Maybe we can pass the warning to the caller (somehow) or we add a `verbose: bool` option to print them?

Now it uses `debug` instead of `warn` in 60e04ca.

I did not figure out how to make it display debug level log, though.

Issue #1246, opened 2025-01-25
scx_rusty tests not passing with most recent bpf-next


etsal:
The most recent pushes on the bpf-next kernel branch combined with Clang 18 are causing verification failures for scx_rusty, breaking the CI. Identify and fix the exact cause of the error. This is not an issue that manifests with the mainline kernels.

etsal:
Exact timeline is:

- lb_domain arenas gets merged last week. This is the patch that probably breaks the bpf-next CI. Alternatively, an unrelated commit to bpf-next breaks the build, but that is way less probable.
- There is a merge into the sched_ext tree from bpf-next, and that introduces the breakage to our build-and-test workflow.

etsal:
Also: #1243 fixes the CI breakage, but makes scx_rusty fail verification on pre-6.14 kernels. We need a build switch to select the right looping construct at compilation time.

etsal:
Mitigated with #1243

PR #1251, opened 2025-01-27
[scx_rusty] make ci signal always green

likewhatevs:
make ci signal always green for scx_rusty.

not sure we should do this (it'll probably make breakages in rusty harder to sort out), but at the same time, rusty is kind-of the proving ground for how we're refactoring things (i.e. best to keep breakages there from being problematic w/ overall ci signal/merge queue etc.).

### DIFF
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -153,15 +153,31 @@ jobs:
 
       # Test schedulers
       - run: meson compile -C build test_sched_${{ matrix.scheduler }}
-      # this is where errors we want logs on start occurring, so always generate debug info and save logs
-        if: always()
+      # exclude flaky/known broken schedulers to keep signal good
+        if: ${{ matrix.scheduler != 'scx_rusty' }}
       # Stress schedulers
       - uses: cytopia/shell-command-retry-action@v0.1.2
         name: stress test
-        if: always()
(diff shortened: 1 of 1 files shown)


multics69:
If possible, I think it would be better to fix the real verification error.

etsal:
@multics69 This addresses issue #1246 and just makes the CI usable for now. I am looking into a more permanent fix.

etsal (approved):

likewhatevs:
spoke to @etsal , there's a decent chance this is a kernel issue, gonna see if I can find the commit which caused it/table this workaround for a day

likewhatevs:
Was able to find a good commit, bisect says https://kernel.googlesource.com/pub/scm/linux/kernel/git/bpf/bpf-next.git/+/d2102f2f5d75a84dbab6ff890359f0bd4a18ca22 is where the breakage started. IDK if the fix is easy or complex, but once the fix is in bpf-next it should make it's way to sched-ext for-next (which CI uses) not too long after.

likewhatevs:
btw we're going w/ moving sched_ext for-next to a periodic job (like bpf-next), and having the main job use a stable branch of the sched_ext repo (I'm thinking sched-ext-LATEST_RELEASE_KERNEL-fixes atm.) to fix this issue.

PR #1252, opened 2025-01-28
scxtop: Add support for generating perfetto traces

hodgesds:
Add support for adding perfetto traces. Still a little bit of a WIP but has some basic functionality. The `SchedSwitchFtraceEvent` protos still need to be fixed, but the basic track line for the DSQ latency works.

Press `a` to record a trace:

Trace file generated (had to rename to `.log` to appease GH):
[scxtop_trace.log](https://github.com/user-attachments/files/18567579/scxtop_trace.log)
Download trace and open [perfetto](https://ui.perfetto.dev/)
Upload the trace and view results:


likewhatevs (approved):
this is a nice ui

hodgesds:
Update ring buffer event handling to run in separate threads and it seems to help a bit with dropped events. It still takes ~2 cores on a 16 core system in processing the events off the buffer. The protobuf handling isn't very optimized, so that could be improved in the future.

JakeHillion:
Been testing this for a while and I really like how stable the output is - have taken several traces on very noisy machines and every one loads into perfetto (with some non-fatal errors) while the existing ftrace script creates invalid output most times which need post-processing to fix. I think we can make the output more complete but this is an excellent start!

PR #1254, opened 2025-01-28
scx/common.bpf.h: Add helper to check task migration state

arighi:
Introduce a new helper for BPF schedulers to determine whether a task can migrate or not (supporting both SMP and UP systems).

This re-aligns the code with kernel commit 5f52bbf2f6e0 ("tools/sched_ext: Add helper to check task migration state").

Moreover, update the schedulers that are relying on p->migration_disabled to use the common helper instead.


etsal (approved):
Looks good to me!

htejun (approved):

PR #1257, opened 2025-01-28
ci: add build-kernel-nix job using dedicated server

JakeHillion:
Begin migration of CPU heavy tasks from GitHub free runners to self hosted dedicated Linux runners. Start with the build-kernel job as it's pretty simple and doesn't run very often. Will come back for the other copies of `build-kernel` once this is proven.

Changes to enable this:
- Set `runs-on` correctly for this job.
- Switch dependency management to a Nix develop shell for this job. This means we get the same dependencies whether we stay on a self-hosted runner or switch back to GitHub. This is much easier than chasing the ever moving target of software installed on the GitHub runners, and has the added benefit of pinning dependencies.
- Use my branch of nixpkgs with `virtme-ng` packaged. Will upstream this once `virtme-ng` is confirmed working for all of our use cases.
- Bump the cache version number. This isn't really necessary but will mean if this does cause any problems that a revert is cleaner.
- Enables `lookup-only` for the cache kernel step. This means that the cache is never downloaded, which is a good idea given the dedicated server will likely take longer to download the cache. It has the added benefit of being much faster on a hit, and has the same behaviour on a miss.

On request, landing this as an additional job and not using the cache artifacts in this initial merge. Will leave this running for a short time before switching.

|||
|-|-|
|Old cache miss| [13m2s](https://github.com/sched-ext/scx/actions/runs/13000153631/job/36256954083) |
|New cache miss| [2m55s](https://github.com/sched-ext/scx/actions/runs/13021263130/job/36322236211) |
|Old cache hit | [12s  ](https://github.com/sched-ext/scx/actions/runs/13016947090/job/36308397960) |
|New cache hit | [6s   ](https://github.com/sched-ext/scx/actions/runs/13021532927/job/36323025625) |

Test plan:
- Performance looks good.
- If CI passes we're good (excl rusty at the minute). Already checked the hit and miss cases separately.

### DIFF
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -26,6 +26,65 @@ jobs:
       - run: cargo fmt
       - run: git diff --exit-code
 
+  build-kernel-nix:
+    runs-on: [ "self-hosted", "linux", "x64" ]
+    steps:
+      # Uncomment these if reverting to GitHub hosted runners
+      # - name: Install Nix
+      #   uses: DeterminateSystems/nix-installer-action@main
+
+      # Make very basic dependencies available in PATH
+      - name: Install basic dependencies
+        run: |
+          nix profile install nixpkgs#{git,openssl,gawk,gnutar,zstd}
+          echo "$HOME/.nix-profile/bin" >> $GITHUB_PATH
+
+      # get latest head commit of sched_ext for-next
+      - run: echo "SCHED_EXT_KERNEL_COMMIT=$(git ls-remote https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git heads/for-6.13 | awk '{print $1}')" >> $GITHUB_ENV
+
+      # check for cached kernel without downloading
+      - name: Cache Kernel
+        id: cache-kernel
+        uses: actions/cache@v4
+        with:
+          path: |
+            linux/arch/x86/boot/bzImage
+            linux/usr/include
+            linux/**/*.h
+          key: kernel-build-${{ env.SCHED_EXT_KERNEL_COMMIT }}-7
+          lookup-only: true
+
+      - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
+        uses: actions/checkout@v4
+
+      - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
+        uses: nicknovitski/nix-develop@v1
+        with:
+          arguments: ./.github/workflows#build-kernel
+
+      - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
+        name: Clone Kernel
+        uses: cytopia/shell-command-retry-action@v0.1.2
+        with:
+          retries: 10
+          pause: 18
+          command: git clone --single-branch -b for-6.13 --depth 1 https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git linux
+
+      - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
+        name: Select correct commit for cache hash
+        run: |
+          cd linux
+          git switch --detach ${{ env.SCHED_EXT_KERNEL_COMMIT }}
+          git log -1 --pretty=format:"%h %ad %s" --date=short
+
+      - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
+      # Build a minimal kernel (with sched-ext enabled) using virtme-ng
+        run: cd linux && virtme-ng -v --build --config ../.github/workflows/sched-ext.config
+
+      - if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
+      # Generate kernel headers
+        run: cd linux && make headers
+
   build-kernel:
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/flake.lock
+++ b/.github/workflows/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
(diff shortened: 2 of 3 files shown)


JakeHillion:
Had to move the kernel cache restore before the setuid on tar (which we obviously shouldn't need anyway). Seems like the random user in the GitHub self hosted runner doesn't give all read permissions to the directories so with setuid the files are owned by a non-existent user and inaccessible.

likewhatevs:
This seems like a lot of complexity to save 10 mins (i.e. pip/cargo/meson/apt => pip/cargo/meson/apt/nix). 

IIUC, nix is most complex out of those.

Could we use the big server with the current build scripts/package managers instead of adding a new package manager? 

This would keep upkeep simple while making builds faster I think.

JakeHillion:
> This seems like a lot of complexity to save 10 mins (i.e. pip/cargo/meson/apt => pip/cargo/meson/apt/nix).

Fortunately the complexity is already sorted with this finished PR :)

> IIUC, nix is most complex out of those.
> 
> Could we use the big server with the current build scripts/package managers instead of adding a new package manager?

Hard disagree here. This, and the remaining work, are a removal of apt, so it's a replacement, not an additional package manager. It also works everywhere with a simple install of Nix.

Part of the reason for choosing Nix here is the improved security posture of running GitHub Actions under NixOS than Debian. It runs in a restricted environment that isn't any sort of Docker-like environment, so getting apt there would be tricky. The alternatives of running under something like Debian are unsupported and we'd likely have to containerise fully there which there aren't any good docs for. This leaves Nix as the choice for package manager, which is an excellent one anyway - it works in all environments including in the future people's development servers, and makes the CI reproducible locally. It's a great experience.

In the long run Nix will be the least complicated here anyway. All the complexity of caching in these workflows can be handled by something like DetSys's flakehub, leaving us with a drastically simpler CI workflow. But we have to take things one step at a time.

> This would keep upkeep simple while making builds faster I think.

The beauty of a Nix based CI is that this will never rot. As long as the install Nix action works we will get an identical CI environment every time, and we can get that environment on a local machine. Upkeep should be great from now on :)

likewhatevs:
I don't understand this:

> Part of the reason for choosing Nix here is the improved security posture of running GitHub Actions under NixOS than Debian. It runs in a restricted environment that isn't any sort of Docker-like environment, so getting apt there would be tricky.

So the reason for using nix package manager is because our CI box is running NixOS and because it is using NixOS for CI to work well it must also use nix?

> The alternatives of running under something like Debian are unsupported and we'd likely have to containerise fully there which there aren't any good docs for. 

Unsupported by what/whom? Unsupported by NixOS? If so use a standard distribution like Ubuntu/Fedora etc. then the need for this additional complexity goes away.

JakeHillion:
> So the reason for using nix package manager is because our CI box is running NixOS and because it is using NixOS for CI to work well it must also use nix?

The reason for using Nix is determinstic builds that can be well reproduced on a local machine, taking away a lot of the GitHub Actions complexity. The reason for making this change at the same time as moving to the big box is because that machine is running NixOS. Nix is a positive change either way.

> > The alternatives of running under something like Debian are unsupported and we'd likely have to containerise fully there which there aren't any good docs for.
> 
> Unsupported by what/whom? Unsupported by NixOS? If so use a standard distribution like Ubuntu/Fedora etc. then the need for this additional complexity goes away.

That was the wrong way around. NixOS has semi-official support for running GitHub Actions runners. The instructions from GitHub for running on other distributions are to build the binary from source and run it, with no explanation of how to do that securely. NixOS has a built in module that creates a relatively secure GitHub Actions runner, is open source, and theoretically that's much better than us guessing how to do it securely once and ignoring it.

likewhatevs:
not supporting this change, but if we decide to go w/ this, can we wait a day or so for me to update CI to resolve the issue w/ signal failing due to kernel issue w/ rusty. 

That way I can make that change quick and easy without having to figure out nix etc.

JakeHillion:
Sure, I can wait until Monday to land. There really isn't much to figure out here once the dependencies are right (they appear to be), and the job doesn't really change much except for installing the dependencies differently. What issue are you foreseeing?

I don't want to wait much longer after the weekend though as the win comes from speeding up the other jobs that take longer, this is just the starting point.

likewhatevs:
> What issue are you foreseeing?

It took me like 15 minutes to make the PR to change how builds etc. go earlier because I didn't use anything fancy/complex/new like nix.

> I don't want to wait much longer after the weekend though as the win comes from speeding up the other jobs that take longer, this is just the starting point.

Yeah this is conflating the win from using a big server for builds with migrating build jobs to use nix. I'd guess the CI config as it is is probably just as fast w/o the overhead of nix if ran on a larger server also. 

> Sure, I can wait until Monday to land.

Again, I'd really rather we not do this, I don't see a tangible benefit from this change to offset it's overhead. Nobody wants to have to learn a new build tool/language when fixing some random breakage and I don't believe there's a magical tool that makes all upkeep/random breakages disappear like that.

hodgesds (approved):

JakeHillion:
> One option is run parallel for a little while to get the feel for it (non land blocking) and then we can go from there? It would be good to get a few people familiar with the setup before cutting over.

Good suggestion. I'll add back the old job with the old cache key and reset the cache keys for the consumers, then switch over fully once we know the builds are clean for a while. I don't expect getting confidence in this to take long, and it unblocks me too.

Will update then land.

JakeHillion:
> One option is run parallel for a little while to get the feel for it (non land blocking) and then we can go from there? It would be good to get a few people familiar with the setup before cutting over.

Good suggestion. I'll add back the old job with the old cache key and reset the cache keys for the consumers, then switch over fully once we know the builds are clean for a while. I don't expect getting confidence in this to take long, and it unblocks me too.

Will update then land.

PR #1261, opened 2025-01-29
scx_central: alert for 'silly' values for the chosen cpu from command…

devnexen:
… line.

### DIFF
--- a/scheds/c/scx_central.c
+++ b/scheds/c/scx_central.c
@@ -10,6 +10,7 @@
 #include <unistd.h>
 #include <inttypes.h>
 #include <signal.h>
+#include <assert.h>
 #include <libgen.h>
 #include <bpf/bpf.h>
 #include <scx/common.h>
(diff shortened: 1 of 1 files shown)


scheds/c/scx_central.c
```diff
@@ -60,14 +61,23 @@ int main(int argc, char **argv)
...
 	while ((opt = getopt(argc, argv, "s:c:pvh")) != -1) {
 		switch (opt) {
 		case 's':
 			skel->rodata->slice_ns = strtoull(optarg, NULL, 0) * 1000;
 			break;
-		case 'c':
-			skel->rodata->central_cpu = strtoul(optarg, NULL, 0);
+		case 'c': {
+			u32 central_cpu = strtoul(optarg, NULL, 0);
+			if (__builtin_expect((central_cpu >= skel->rodata->nr_cpu_ids), 0)) {
```
htejun, line 73: I don't think we need to hint the compiler for optimization here.

devnexen: ok

scheds/c/scx_central.c
```diff
@@ -60,14 +61,23 @@ int main(int argc, char **argv)
...
 			break;
-		case 'c':
-			skel->rodata->central_cpu = strtoul(optarg, NULL, 0);
+		case 'c': {
+			u32 central_cpu = strtoul(optarg, NULL, 0);
+			if (central_cpu >= skel->rodata->nr_cpu_ids) {
+				fprintf(stderr, "invalid central CPU id value, %u given (%u max)\n", central_cpu, skel->rodata->nr_cpu_ids);
+				return -1;
+			}
+			/* does central_cpu have to be a signed type ? */
```
htejun, line 77: Yeah, scheduler internally uses negative values for error indication, so using u32 for cpus can become pretty confusing. We used to use u32 for cpus and actually had subtle bugs due to it.

scheds/c/scx_central.c
```diff
@@ -60,14 +61,23 @@ int main(int argc, char **argv)
 	skel->rodata->nr_cpu_ids = libbpf_num_possible_cpus();
 	skel->rodata->slice_ns = __COMPAT_ENUM_OR_ZERO("scx_public_consts", "SCX_SLICE_DFL");
 
+	assert(skel->rodata->nr_cpu_ids <= INT32_MAX);
```
htejun, line 64: Why not but I wouldn't worry too much about this.

scheds/c/scx_central.c
```diff
@@ -60,14 +61,23 @@ int main(int argc, char **argv)
 	skel->rodata->nr_cpu_ids = libbpf_num_possible_cpus();
```
htejun, line 61: Not from this PR but as a side note, this is wrong, strictly speaking. `nr_cpu_ids` should be `MAX_CPU_ID + 1` while `libbpf_num_possible_cpus()` returns the number of all possible CPUs. The two agree on most systems but if possible cpumask can have holes, the two deviate and there are systems with holes. So, strictly speaking, we should be getting nr_cpu_ids from topology information and then testing the cpus against possible mask.

PR #1262, opened 2025-01-29
scx_rusty: add option for setting scx_bpf_cpuperf_set()

etsal:
Add functionality in scx_rusty to set CPU perf levels when initializing its domains. Doing so has provided performance improvements for scx_layered.


etsal:
Note that the default option is to set the perf level to 1024, the minimum. Maybe a better idea is to set it to -1 by default, which notifies the BPF scheduler not to set the event.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1851,6 +1864,7 @@ static s32 create_dom(u32 dom_id)
 	bpf_cpumask_copy(dom_mask, cast_mask(node_mask));
 	bpf_rcu_read_unlock();
 
+
```
htejun, line 1867: stray new line?

htejun (approved):

scheds/rust/scx_rusty/src/bpf/intf.h
```diff
@@ -64,6 +64,9 @@ enum consts {
 	 * anyway and will be retried until loads are balanced.
 	 */
 	MAX_DOM_ACTIVE_TPTRS	= 1024,
+
+	/* Maximum allowed value for scx_bpf_cpuperf_set(). */
+	PERF_MAX		= 1024,
```
arighi, line 69: Can we use `SCX_CPUPERF_ONE` as `PERF_MAX`?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1759,6 +1760,7 @@ static s32 create_dom(u32 dom_id)
 	struct bpf_cpumask *dom_mask, *node_mask, *all_mask;
 	struct lb_domain *lb_domain;
 	u32 cpu, node_id;
+	int perf;
```
arighi, line 1763: Isn't this shadowing the global `perf`?

scheds/rust/scx_rusty/src/main.rs
```diff
@@ -225,6 +225,10 @@ struct Opts {
     /// Show descriptions for statistics.
     #[clap(long)]
     help_stats: bool,
+
+    /// Use scx_bpf_cpuperf_set() to set CPU frequency.
```
arighi, line 229: Should we provide a more abstracted description, since this is also shown to the users, maybe something like:
```
Set the performance level of the schedutil CPU frequency governor. The value ranges from 0
(minimum performance) to 1024 (maximum performance). A lower value favors energy efficiency,
while a higher value prioritizes performance.
```

etsal: I agree, I was also debating just having two levels, 0 and 1 and making this a boolean to make it simpler. That would lock users out of being able to try out settings for themselves, so making it a u32 and recommending they use 0 or 1024 if they're not sure is better.

PR #1264, opened 2025-01-29
scx_central: account for offline cpus when choosing cpu != 0.

devnexen:
max number of cpu ids does not necessarily equate all available. Thus checking the current topology and accounting manually.

### DIFF
--- a/scheds/c/scx_central.c
+++ b/scheds/c/scx_central.c
@@ -49,7 +49,17 @@ int main(int argc, char **argv)
 	struct bpf_link *link;
 	__u64 seq = 0, ecode;
 	__s32 opt;
-	cpu_set_t *cpuset;
+	__u32 max_cpu_ids, nr_cpu_ids = 0, i = 0;
+	cpu_set_t *cpuset, curcpuset;
+
+	max_cpu_ids = libbpf_num_possible_cpus();
+	CPU_ZERO(&curcpuset);
+	SCX_BUG_ON(sched_getaffinity(0, sizeof(curcpuset), &curcpuset),
+		   "Failed to get current CPU topology");
+	for (i = 0; i < max_cpu_ids; i ++) {
+		if (CPU_ISSET(i, &curcpuset))
+			nr_cpu_ids ++;
+	}
 
 	libbpf_set_print(libbpf_print_fn);
 	signal(SIGINT, sigint_handler);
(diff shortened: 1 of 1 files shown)


scheds/c/scx_central.c
```diff
@@ -49,7 +49,17 @@ int main(int argc, char **argv)
...
+	__u32 max_cpu_ids, nr_cpu_ids = 0, i = 0;
+	cpu_set_t *cpuset, curcpuset;
+
+	max_cpu_ids = libbpf_num_possible_cpus();
+	CPU_ZERO(&curcpuset);
+	SCX_BUG_ON(sched_getaffinity(0, sizeof(curcpuset), &curcpuset),
+		   "Failed to get current CPU topology");
+	for (i = 0; i < max_cpu_ids; i ++) {
+		if (CPU_ISSET(i, &curcpuset))
+			nr_cpu_ids ++;
```
htejun, line 61: `nr_cpu_ids` is maximum possible CPU + 1. The above is counting the number of CPUs in the current task's affinity, which is a different thing. You'd have to go by `/sys/devices/system/cpu/possible`.

scheds/c/scx_central.c
```diff
@@ -43,13 +43,18 @@ static void sigint_handler(int dummy)
 	exit_req = 1;
 }
 
+extern int parse_cpu_mask_file(const char *, bool **, int *);
```
htejun, line 46: Where is this function defined?

devnexen: it s a libbpf internal.

scheds/c/scx_central.c
```diff
@@ -72,6 +77,11 @@ int main(int argc, char **argv)
 			u32 central_cpu = strtoul(optarg, NULL, 0);
 			if (central_cpu >= skel->rodata->nr_cpu_ids) {
 				fprintf(stderr, "invalid central CPU id value, %u given (%u max)\n", central_cpu, skel->rodata->nr_cpu_ids);
+				free(cpumask);
+				return -1;
+			} else if (central_cpu >= ncpu || !cpumask[central_cpu]) {
```
htejun, line 82: I'm not sure how valuable it is to add just CPU online test here. Overall, the logic is still wrong as it's still assuming possible cpumasks don't have gaps.

devnexen: So what would be your suggestion ?

htejun: As it's more of a simple example to demonstrate an idea, I wonder whether we should just leave it be. We can add possible / online distinction and hotplug handling but that seems rather too much given that the scheduler isn't a very practical one to begin with. We have similar issues on rust scheds and there we should do it properly but it'd also be a lot easier to do so as it's easy to grow libraries and helpers there.

PR #1265, opened 2025-01-29
scx_rusty: Use bpf_skel::types::* instead of bpf_intf::*

htejun:
libbpf-rs now provides all types defined in .bpf.c. Use those definitions instead of bindgen generated ones. This removes the need for explicit type casts.


likewhatevs:
I think rebasing fixes build failure, still waiting to see the changes earlier work to know.

arighi (approved):
Ah that's nice! I need to do something similar in scx_rustland_core.

etsal (approved):

PR #1267, opened 2025-01-30
scxtop: Fix handling non scx scheduled switch events

hodgesds:
Fix handling non scx handling sched_switch events by properly setting the header on the event so that it can be processed by userspace correctly. Remove the use of tokio for ringbuffer callbacks . Will add some counters later for collecting dropped events.

traces are now working for when a scheduler is loaded or when eevdf is running.

no scheduler:

layered:

example trace:
[scxtop_trace_0.log](https://github.com/user-attachments/files/18596118/scxtop_trace_0.log)

### DIFF
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -372,8 +372,8 @@ int BPF_PROG(on_sched_wakeup_new, struct task_struct *p)
 }
 
 
-static int on_sched_switch_non_scx(bool preempt, struct task_struct *prev,
-				   struct task_struct *next, u64 prev_state)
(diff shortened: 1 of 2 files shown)


tools/scxtop/src/bpf/main.bpf.c
```diff
@@ -384,6 +384,10 @@ static int on_sched_switch_non_scx(bool preempt, struct task_struct *prev,
 	if (!event)
 		return -ENOENT;
 
+	u64 now = bpf_ktime_get_ns();
+	event->type = SCHED_SWITCH;
```
hodgesds, line 388: @etsal it was something painfully stupid :laughing:

likewhatevs (approved):

JakeHillion:
Tested this with both layered and EEVDF, super happy with the output now. Thanks a lot!

PR #1274, opened 2025-01-30
scx_rustland_core: re-introduce ops.update_idle()

arighi:
Re-introduce a softer version of idle polling to keep re-scheduling the user-space scheduler from ops.update_idle() if it still has pending tasks waiting to be dispatched.

This allows to achieve good core utilization both with v6.12 and v6.13 kernels.


JakeHillion (approved):
LGTM.

PR #1277, opened 2025-01-30
ci: only trigger slack notification on failure of main for scheduled runs

JakeHillion:
Currently the Slack notification gets sent even if the failing branch is not
`main`. This means that any testing done on these workflows (by temporarily
enabling `push`) triggers the notification if they fail or are cancelled.

Replace the `always()` condition with `failure()`. This omits the previous
`cancelled()` case, but that should be fine. Will add it back if there are
objections. Also add the same filter as the `pages` job to only run on `main`.

Test plan:
- __shrugs__

### DIFF
--- a/.github/workflows/bpf-next-test.yml
+++ b/.github/workflows/bpf-next-test.yml
@@ -157,7 +157,7 @@ jobs:
           compression-level: 9
   notify-job:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


JakeHillion:
There's not an obvious way to test this and it seems like the last few sets of testing this feature happened on `main` anyway, so best approach here seems to be to merge and keep an eye on the next several runs.

hodgesds (approved):
LGTM

PR #1278, opened 2025-01-30
ci: create reusable workflow for nix kernel builds

JakeHillion:
Adds a reusable workflow `build-kernel.yml` to cover the build-kernel job using the new Nix based build process on the dedicated runner.

All of the `build-kernel` jobs are identical except for their git repo and git branch name. Factor these out into a reusable workflow to reduce code duplication.

This also removes any suffixes from the cache, which might (unlikely) increase hit rates. Each build was suffixing their kernels separately even though the build was identical, which was slightly wasteful. It is unlikely any of these repos/branches have identical states though.

Test plan:
- Ran the CI and waited for build-kernel-nix to succeed in each workflow (added a temporary `push:` condition to make sure they all build). Cancelled the rest as this built kernel isn't used yet so there's no point causing extra queuing.

### DIFF
--- a/.github/workflows/bpf-next-test.yml
+++ b/.github/workflows/bpf-next-test.yml
@@ -6,6 +6,12 @@ on:
       - cron: "0 */6 * * *"
 
 jobs:
+  build-kernel-nix:
+    uses: ./.github/workflows/build-kernel.yml
+    with:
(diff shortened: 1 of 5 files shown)


hodgesds (approved):
LGTM

PR #1280, opened 2025-01-30
ci: switch scheduled jobs to nix built kernel

JakeHillion:
Create a reusable GitHub action to restore the kernel cache. This takes an
argument for `git-repo` and `branch`, like the equivalent reusable workflow for
building a Nix kernel. Enable this action for the Nix built kernels in the
scheduled runs.

This reduces the possibility for a mis-copy and significantly simplifies the
files compared to duplicating the code.

Note that this is deliberately ordered before the `tar` setuid bit in the
actions, as the new cache and that change don't interact well. Will drop that
setuid step in future changes but leaving it there for now to load the other
caches.

Drive by: remove lint from the for-next scheduled run. It's already checked on
PRs, in the merge queue, and on the scheduled main run.

Test plan:
- Changed one of the cache loads to the `master` branch to guarantee a miss.
    Workflow fails as expected.
- Tested scheduled workflows by enabling them temporarily on `push`. They pass.

### DIFF
--- a/.github/actions/restore-kernel-cache/action.yml
+++ b/.github/actions/restore-kernel-cache/action.yml
@@ -1,0 +1,26 @@
+name: 'Restore kernel cache'
+description: 'Restore a previously built kernel cache, failing if not available'
+inputs:
+  git-repo:
+    required: true
+    type: string
+  branch:
+    required: true
+    type: string
+
+runs:
+  using: "composite"
+  steps:
+    - name: Get hash from repo/branch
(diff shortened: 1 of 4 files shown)


JakeHillion:
Whoops, this doesn't work until after switching to the Nix kernels because of the currently different suffixes. Will hold off until that gets switched on.

JakeHillion:
Still using the generic action but now applying it to the scheduled runs. They've been successful at building the kernel - I looked at the last 20 or so, and one even built when the GH actions runner failed (a timeout I think). Will switch them over to the new kernel and keep an eye, then follow with the more relevant `caching-build` probably tomorrow.

hodgesds (approved):
LGTM

PR #1281, opened 2025-01-31
scx_utils: adding cpu affinity data to Gpu type.

devnexen:


devnexen:
Good question, it should work correctly on 64 bits, as a follow up I ll try to think of something for 32 bits wdyt ?

PR #1282, opened 2025-01-31
scx_utils: #1281 follow-up for 32 bits.

devnexen:


hodgesds (approved):
LGTM

rust/scx_utils/src/gpu.rs
```diff
@@ -51,6 +51,9 @@ pub fn create_gpus() -> BTreeMap<usize, Vec<Gpu>> {
             };
 
             let cpu_mask = if let Ok(cpu_affinity) = nvidia_gpu.cpu_affinity(*NR_CPUS_POSSIBLE) {
```
htejun, line 53: BTW, this should use `NR_CPU_IDS`. Possible CPUs are not guaranteed to be consecutive.

rust/scx_utils/src/gpu.rs
```diff
@@ -51,6 +51,9 @@ pub fn create_gpus() -> BTreeMap<usize, Vec<Gpu>> {
...
+                // Note: nvml returns it as an arch dependent array of integrals
+                #[cfg(target_pointer_width = "32")]
+                let cpu_affinity = cpu_affinity.into_iter().map(|aff| aff as u64).collect();
```
htejun, line 56: Wouldn't the values need to be packed? Let's say there are 64 CPUs:
- On 64bit, it'd return `0xffff_ffff_ffff_ffffu64`.
- On 32bit, it'd return `0xffff_ffffu32, 0xffff_ffffu32` and then the above will make it into `0x0000_0000_ffff_ffffu64, 0x0000_0000_ffff_ffffu64`, no?

devnexen: ok, valid point let me see this..

PR #1285, opened 2025-02-01
scx_layered: refactor gpu support to workaround nvml issues

likewhatevs:
Refactor gpu support to workaround nvml issues.

This changes default behavior when `--enable-gpu-support` is set to use NVML less. It also adds a flag `--aggressive-nvml-polling` to restore prior behavior (poll NVML on each iteration of the userspace update loop).

By default, NVML will now be called super conservatively. In addition to only initializing NVML once  (which is done either way), the existence of PIDs seen previously by NVML will be used to limit updates to the map of GPU pids (and, more importantly, reading PIDs using GPUs via NVML).

The ruleset is now as follows:
* If all previously seen PIDs using GPUs are not dead, check NVML again in 10 minutes.
* If no PIDs were previously seen using GPUs, check NVML every 30 seconds.
* When a PID using a GPU dies, check NVML.
* "dead" wrt/ PIDs in this context means start_time > previous start time or the PID is gone.

The goal of this is to reduce NVML calls to avoid issues with concurrent calls (on occasion) crashing systems.

The GPU pinning works the same as last time, this just controls when it's done. Seems to work fine, the following screenshot shows a 30 second interval check being skipped (while test job was running), a NVML update after I killed the test job, and another NVML update after that (when no GPU jobs were running).

Screenshot from 2025-02-01 00-47-13

cc @valentinandrei

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2237,6 +2237,7 @@ dependencies = [
  "serde",
  "serde_json",
  "simplelog",
+ "sysinfo 0.33.1",
 ]
 
 [[package]]
@@ -2250,7 +2251,7 @@ dependencies = [
  "log",
  "nix 0.29.0",
  "serde",
- "sysinfo",
+ "sysinfo 0.31.4",
  "tokio",
  "tokio-util",
  "toml",
@@ -2717,6 +2718,20 @@ name = "sysinfo"
 version = "0.31.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "355dbe4f8799b304b05e1b0f05fc59b2a18d36645cf169607da45bde2f69a1be"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+ "memchr",
+ "ntapi",
+ "rayon",
+ "windows",
+]
+
+[[package]]
+name = "sysinfo"
+version = "0.33.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4fc858248ea01b66f19d8e8a6d55f41deaf91e9d495246fd01368d99935c6c01"
 dependencies = [
  "core-foundation-sys",
  "libc",
--- a/scheds/rust/scx_layered/Cargo.toml
+++ b/scheds/rust/scx_layered/Cargo.toml
@@ -27,6 +27,7 @@ serde_json = "1.0"
 simplelog = "0.12"
 nvml-wrapper = "0.10.0"
 once_cell = "1.20.2"
+sysinfo = "0.33.1"
 
 [build-dependencies]
 scx_utils = { path = "../../../rust/scx_utils", version = "1.0.10" }
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -65,6 +65,7 @@ use stats::LayerStats;
 use stats::StatsReq;
 use stats::StatsRes;
 use stats::SysStats;
+use sysinfo::System;
 
 const MAX_PATH: usize = bpf_intf::consts_MAX_PATH as usize;
 const MAX_COMM: usize = bpf_intf::consts_MAX_COMM as usize;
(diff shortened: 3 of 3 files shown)


hodgesds (commented):
Pulling in the `sysinfo` crate might be another big dependency considering it already has the `fb_procfs` crate, which has similar [api](https://docs.rs/fb_procfs/latest/fb_procfs/struct.ProcReader.html#method.read_all_pids). I think a more lightweight approach would be something like spinning up a background thread if monitoring the GPU pids using `pidfd_open` would be a more lightweight/better approach. From the `pidfd_open` man page:

> The pidfd_open() system call is the preferred way of obtaining a
       PID file descriptor for an already existing process.  The
       alternative is to obtain a file descriptor by opening a /proc/pid
       directory.  However, the latter technique is possible only if the
       [proc(5)](https://www.man7.org/linux/man-pages/man5/proc.5.html) filesystem is mounted; furthermore, the file descriptor
       obtained in this way is not pollable and can't be waited on with
       [waitid(2)](https://www.man7.org/linux/man-pages/man2/waitid.2.html).The pidfd_open() system call is the preferred way of obtaining a
       PID file descriptor for an already existing process.  The
       alternative is to obtain a file descriptor by opening a /proc/pid
       directory.  However, the latter technique is possible only if the
       [proc(5)](https://www.man7.org/linux/man-pages/man5/proc.5.html) filesystem is mounted; furthermore, the file descriptor
       obtained in this way is not pollable and can't be waited on with
       [waitid(2)](https://www.man7.org/linux/man-pages/man2/waitid.2.html).
       
       
The big advantage of using `pidfd_open` is that the fd can be polled on and when the process dies you can get notifications directly off the fd and send the result across a channel.

likewhatevs:
> Pulling in the `sysinfo` crate might be another big dependency

IDK, this feels like early micro optimization.

Here's data on the change on binary size w/ this change:
```
gentoo in scheds/rust/scx_layered on  main [$] is 📦 v1.0.9 via 🦀 v1.84.1
❯ ls -lah ../../../target/release/scx_layered
.rwxr-xr-x patso patso 4.9 MB Sat Feb  1 11:56:33 2025  ../../../target/release/scx_layered

gentoo in scheds/rust/scx_layered on  add-gpu-safe [$] is 📦 v1.0.9 via 🦀 v1.84.1
❯ ls -lah ../../../target/release/scx_layered
.rwxr-xr-x patso patso 5.1 MB Sat Feb  1 11:57:41 2025  ../../../target/release/scx_layered
```

Here's data on the change on semi-warm build times (what folks will see most of the time checking out code/doing release builds for local tests):
```
gentoo in scheds/rust/scx_layered on  add-gpu-safe [$] is 📦 v1.0.9 via 🦀 v1.84.1
❯ cargo build --release
warning: /home/patso/repos/scx/rust/scx_rustland_core/Cargo.toml: unused manifest key: lib.include
   Compiling scx_utils v1.0.10 (/home/patso/repos/scx/rust/scx_utils)
   Compiling scx_layered v1.0.9 (/home/patso/repos/scx/scheds/rust/scx_layered)
    Finished `release` profile [optimized] target(s) in 24.57s

gentoo in scheds/rust/scx_layered on  main [$] is 📦 v1.0.9 via 🦀 v1.84.1
❯ cargo build --release
warning: /home/patso/repos/scx/rust/scx_rustland_core/Cargo.toml: unused manifest key: lib.include
   Compiling scx_utils v1.0.10 (/home/patso/repos/scx/rust/scx_utils)
   Compiling scx_layered v1.0.9 (/home/patso/repos/scx/scheds/rust/scx_layered)
    Finished `release` profile [optimized] target(s) in 23.84s
```

> I think a more lightweight approach would be something like spinning up a background thread if monitoring the GPU pids using pidfd_open. From the pidfd_open man page:

A background thread sounds more heavy (both in terms of work and complexity of the code, etc) than doing a periodic check like this tbh.

> The big advantage of using pidfd_open is that the fd can be polled on and when the process dies you can get notifications directly off the fd and send the result across a channel.

Seeing as we would act upon this synchronously/in the update loop anyhow, not seeing the benefit of adding an async executor nor threading this.

etsal:
Just as a clarification, this code assumes that the processes in the working set are all long lived and do not spawn short lived processes right? 

Also, AFAICT this code will force an NVML lookup if an existing process dies, but it doesn't check whether a new process has started using the GPU code. This should be fine because we assume a model where all worker threads are spawned at once when an application that uses the GPU starts up.

etsal (approved):

likewhatevs:
> Just as a clarification, this code assumes that the processes in the working set are all long lived and do not spawn short lived processes right?

Correct. Not super sure how I feel about that, but it's a reasonable idea to reduce NVML calls.

> Also, AFAICT this code will force an NVML lookup if an existing process dies, but it doesn't check whether a new process has started using the GPU code. This should be fine because we assume a model where all worker threads are spawned at once when an application that uses the GPU starts up.

Hmm let me test this. I'm reasonably confident that after a GPU process dies, things work like they did before a GPU process was found (which is the intent).

likewhatevs:
> Hmm let me test this. I'm reasonably confident that after a GPU process dies, things work like they did before a GPU process was found (which is the intent).

This works like that when tested (while GPU jobs were running, once found, htop showed things looking as they should, and the skipped/not skipped NVML checks in the following screenshot confirm the "when to check" logic works as was intended also):

image

PR #1286, opened 2025-02-01
scx_util: Fix incorrect time delta calculation in time_delta()

multics69:
When (s64)(after - before) > 0, the code returns the result of (s64)(after - before) > 0 while the intended result should be (s64)(after - before). That happens because the middle operand of the ternary operator was omitted incorrectly, returning the result of (s64)(after - before) > 0. Thus, add the middle operand -- (s64)(after - before) -- to return the correct time calculation.

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -444,7 +444,7 @@ void bpf_rcu_read_unlock(void) __ksym;
  */
 static inline s64 time_delta(u64 after, u64 before)
 {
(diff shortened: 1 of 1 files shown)


multics69:
I will also submit a kernel patch.

arighi (approved):
At some point we could also introduce `max_t()/min_t()/clamp()` macros (similar to those that we have in `include/linux/kernel.h`) and use `max_t()` here. And also use them in bpfland and flash, instead of the custom `MAX()/MIN()/CLAMP()` macros.

But for now looks good to me, thanks!

PR #1287, opened 2025-02-02
scx_lavd: Revert "scx_lavd: Preemption when ops.select_cpu() was skipped"

multics69:
This reverts commit 22057be29a4308861e0c05759fcc80c00a915e47

That particular commit has more harm than good. It causes latency spikes and performance drops in many workloads, including WebGL Aquarium, osu-wine, Bitwig, and slowroads.io. Thus revert the commit.


arighi (approved):

htejun (approved):

multics69:
@htejun -- This fixes the performance regression, so `scx_lavd` is ready for the new release.

PR #1289, opened 2025-02-03
ci: disable DEBUG_INFO_REDUCED for arm64

luigidematteis:
Disable CONFIG_DEBUG_INFO_REDUCED, as it blocks CONFIG_DEBUG_INFO_BTF, preventing BTF from being generated.


arighi (approved):
@luigidematteis have you tried to cleanup the build, re-generate the config via vng and see if `CONFIG_DEBUG_INFO_BTF` is set?

In that case LGMT!

luigidematteis:
> @luigidematteis have you tried to cleanup the build, re-generate the config via vng and see if `CONFIG_DEBUG_INFO_BTF` is set?
> 
> 
> 
> In that case LGMT!

Yes 👍

PR #1290, opened 2025-02-03
scx_utils: not hardcoding BPF version anymore proposal.

devnexen:
kernel and toolchain still need to match, eBPF v4 supported only from llvm 18.x.

### DIFF
--- a/rust/scx_utils/src/clang_info.rs
+++ b/rust/scx_utils/src/clang_info.rs
@@ -33,11 +33,13 @@ pub struct ClangInfo {
     pub clang: String,
     pub ver: String,
     pub arch: String,
+    pub bpfver: u8,
 }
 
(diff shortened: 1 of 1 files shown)


JakeHillion:
Looks like the diff between v3 and v4 is here: https://lore.kernel.org/bpf/4bfe98be-5333-1c7e-2f6d-42486c8ec039@meta.com/

This change makes sense to me. @etsal @htejun @arighi any issues with using the v4 instruction set when available? It seems like the v3 instructions pre-data sched-ext, at least upstream.

htejun:
The problem is that code which is fine in v4 won't verify in v3 - e.g. the signed div/mod. It might be less confusing to wait till we can switch over to clang19 and just require v4.

PR #1291, opened 2025-02-03
lib/sdt_alloc: introduce typeless init-time arena allocator

etsal:
Add a simple typeless, single threaded arena allocator to be used at initialization time. This is useful because:

1) Arena allocators are currently typed. We have to create a new allocator for each new type we want to add into the arena.
2) The existing allocator's retry and tree traversal logic is complex enough that naively making too many allocator calls in a single function can cause verification to fail, e.g., allocating an array of structs in a loop.

With this new allocator we avoid these problems for initialization-time struct instances.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -1909,6 +1909,10 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init)
 {
 	s32 i, ret;
 
+	ret = sdt_static_init(STATIC_ALLOC_PAGES_GRANULARITY);
+	if (ret)
+		return ret;
+
 	ret = sdt_task_init(sizeof(struct task_ctx));
 	if (ret)
 		return ret;
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -94,10 +94,14 @@ struct bpfmask_wrapper {
 	struct bpf_cpumask __kptr *instance;
 };
 
+struct rusty_percpu_storage {
+	struct bpf_cpumask __kptr *bpfmask;
+};
+
 struct {
 	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
 	__type(key, u32);
-	__type(value, struct bpfmask_wrapper);
+	__type(value, struct rusty_percpu_storage);
 	__uint(max_entries, 1);
 } scx_percpu_bpfmask_map SEC(".maps");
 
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1908,7 +1909,7 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init)
 {
 	s32 i, ret;
 
-	ret = scx_percpu_tmpmask_init();
+	ret = sdt_static_init(PAGE_SIZE);
```
htejun, line 1912: Isn't this allocating 4k pages? Is that intentional? Also, does it need to pre-allocate pages? Wouldn't something like the following be fairly simple while not requiring users to pre-calculate how much they may need?
- If allocation size >= PAGE_SIZE, just forward the requrest to `bpf_arena_alloc_pages()`.
- For allocations < PAGE_SIZE, let the allocator hold onto one page and allocate from it. When the next allocation won't fit in the page, get a new page, allocate from it and pick between the previous and new page based on number of bytes remaining.

htejun:
Also, can't we hide the allocator details from the verifier with `__weak` so that how many times it's called doesn't matter?

scheds/rust/scx_rusty/src/bpf/main.bpf.c
etsal: That's pretty much what the code does, it sets the allocation size. The alloc_pages parameter is how many pages we allocate out of the backing arena every time we run out of memory. We stack allocate out of these pages on every requested allocation. I have not implemented the second bullet point for simplicity, because the code is making the assumption alloc_size is way larger than the individual allocation request made by the scheduler so internal fragmentation should be small.

etsal:
> Also, can't we hide the allocator details from the verifier with `__weak` so that how many times it's called doesn't matter?

The allocator 

> Also, can't we hide the allocator details from the verifier with `__weak` so that how many times it's called doesn't matter?

We do so for the free path, but the allocation path returns a pointer and global functions can only return scalars.

htejun:
> > Also, can't we hide the allocator details from the verifier with `__weak` so that how many times it's called doesn't matter?
> 
> We do so for the free path, but the allocation path returns a pointer and global functions can only return scalars.

Hmm... this sounds like something which can be easily fixed from the verifier side for arena pointers, right?

etsal:
Will send a patch out for sdt_alloc(), checked typecasting the pointer to and from a u64 and it works fine. Do we still want the simpler allocator from this patch or is it not worth it for now?

htejun:
I think it's generally useful to have a typeless allocator, so please go ahead.

htejun (approved):

etsal:
Sounds good I'll merge it then. I think now that we have ways of writing larger and larger programs that pass the verifier it may be time for a full-fledged malloc() style allocator soon.

scheds/rust/scx_rusty/src/bpf/main.bpf.c
etsal: Oh yeah it is a leftover from when I was passing the size in bytes.

sirlucjan:
```[45/47] Compiling C object scheds/c/scx_sdt.p/scx_sdt.c.o
FAILED: scheds/c/scx_sdt.p/scx_sdt.c.o 
ccache cc -Ischeds/c/scx_sdt.p -Ischeds/c -I../scheds/c -I../scheds/include -I../build/libbpf/src/usr/include -I../build/libbpf/include/uapi -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -march=native -O3 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fPIE -pthread -MD -MQ scheds/c/scx_sdt.p/scx_sdt.c.o -MF scheds/c/scx_sdt.p/scx_sdt.c.o.d -o scheds/c/scx_sdt.p/scx_sdt.c.o -c ../scheds/c/scx_sdt.c
In file included from ../scheds/c/scx_sdt.c:14:
scheds/c/scx_sdt.p/scx_sdt.bpf.skel.h:145:35: error: field ‘sdt_static’ has incomplete type
  145 |                 struct sdt_static sdt_static;
      |                                   ^~~~~~~~~~
scheds/c/scx_sdt.p/scx_sdt.bpf.skel.h: In function ‘scx_sdt__assert’:
scheds/c/scx_sdt.p/scx_sdt.bpf.skel.h:3561:24: error: expression in static assertion is not an integer
 3561 |         _Static_assert(sizeof(s->bss->sdt_static) == 24, "unexpected size of 'sdt_static'");
```
@multics69 has the same error.

PR #1293, opened 2025-02-03
scx_rusty: use arena-based CPU masks

etsal:
The scx_rusty scheduler currently represents domains and tasks with arena-based structs, but cannot embed cpumasks into these fields as they are represented as kptrs. Introduce a cpumask type that is implemented as a BPF library and so can be embedded/pointed to by arena structs. Also add wrappers around kfuncs that take struct bpf_cpumask * arguments to transparently translate between our new internal cpumask type and the bpf_cpumask type expected by the calls.

Note: This code currently has really suboptimal bpf_cpumask to scxmask conversions that I am in the process of fixing. We'll need a kfunc for that but the rest of the code compiles and verifies just fine so it's a matter of solving this one isolated issue.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/cpumask.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/cpumask.bpf.c
@@ -4,30 +4,30 @@
 #include "cpumask.h"
 #include "percpu.h"
 
-static struct sdt_allocator scx_mask_allocator;
+static struct sdt_allocator scx_bitmap_allocator;
 size_t mask_size;
 
 __weak
-int scx_mask_init(__u64 total_mask_size)
+int scx_bitmap_init(__u64 total_mask_size)
 {
 	mask_size = div_round_up(total_mask_size, 8);
 
-	return sdt_alloc_init(&scx_mask_allocator, mask_size * 8 + sizeof(union sdt_id));
+	return sdt_alloc_init(&scx_bitmap_allocator, mask_size * 8 + sizeof(union sdt_id));
 }
 
 __weak
-u64 scx_mask_alloc_internal(void)
+u64 scx_bitmap_alloc_internal(void)
 {
 	struct sdt_data __arena *data = NULL;
-	scx_cpumask_t mask;
+	scx_bitmap_t mask;
 	int i;
 
-	data = sdt_alloc(&scx_mask_allocator);
+	data = sdt_alloc(&scx_bitmap_allocator);
 	cast_kern(data);
 	if (!data)
 		return (u64)(NULL);
 
-	mask = (scx_cpumask_t)data->payload;
+	mask = (scx_bitmap_t)data->payload;
 	mask->tid = data->tid;
 	bpf_for(i, 0, mask_size) {
 		mask->bits[i] = 0ULL;
@@ -42,16 +42,16 @@ u64 scx_mask_alloc_internal(void)
  */
 
 __weak
-int scx_mask_free(scx_cpumask_t __arg_arena mask)
+int scx_bitmap_free(scx_bitmap_t __arg_arena mask)
 {
 	sdt_subprog_init_arena();
 
-	sdt_free_idx(&scx_mask_allocator, mask->tid.idx);
+	sdt_free_idx(&scx_bitmap_allocator, mask->tid.idx);
 	return 0;
 }
 
 __weak
-int scxmask_copy_to_stack(struct scx_cpumask *dst, scx_cpumask_t __arg_arena src)
+int scx_bitmap_copy_to_stack(struct scx_bitmap *dst, scx_bitmap_t __arg_arena src)
 {
 	int i;
 
@@ -70,27 +70,27 @@ int scxmask_copy_to_stack(struct scx_cpumask *dst, scx_cpumask_t __arg_arena src
 }
 
 __weak
-int scxmask_set_cpu(u32 cpu, scx_cpumask_t __arg_arena mask)
+int scx_bitmap_set_cpu(u32 cpu, scx_bitmap_t __arg_arena mask)
 {
 	mask->bits[cpu / 64] |= 1 << (cpu % 64);
 	return 0;
 }
 
 __weak
-int scxmask_clear_cpu(u32 cpu, scx_cpumask_t __arg_arena mask)
+int scx_bitmap_clear_cpu(u32 cpu, scx_bitmap_t __arg_arena mask)
 {
 	mask->bits[cpu / 64] &= 1 << ~(cpu % 64);
 	return 0;
 }
 
 __weak
-bool scxmask_test_cpu(u32 cpu, scx_cpumask_t __arg_arena mask)
+bool scx_bitmap_test_cpu(u32 cpu, scx_bitmap_t __arg_arena mask)
 {
 	return mask->bits[cpu / 64] & (1 << (cpu % 64));
 }
 
 __weak
-int scxmask_clear(scx_cpumask_t __arg_arena mask)
+int scx_bitmap_clear(scx_bitmap_t __arg_arena mask)
 {
 	int i;
 
@@ -102,7 +102,7 @@ int scxmask_clear(scx_cpumask_t __arg_arena mask)
 }
 
 __weak
-int scxmask_and(scx_cpumask_t __arg_arena dst, scx_cpumask_t __arg_arena src1, scx_cpumask_t __arg_arena src2)
+int scx_bitmap_and(scx_bitmap_t __arg_arena dst, scx_bitmap_t __arg_arena src1, scx_bitmap_t __arg_arena src2)
 {
 	int i;
 
@@ -114,7 +114,7 @@ int scxmask_and(scx_cpumask_t __arg_arena dst, scx_cpumask_t __arg_arena src1, s
 }
 
 __weak
-bool scxmask_empty(scx_cpumask_t __arg_arena mask)
+bool scx_bitmap_empty(scx_bitmap_t __arg_arena mask)
 {
 	int i;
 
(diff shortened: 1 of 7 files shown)


scheds/rust/scx_rusty/src/bpf/cpumask.h
```diff
@@ -0,0 +1,199 @@
+extern const volatile u32 nr_cpu_ids;
+size_t mask_size;
+
+/* XXX HACK hardcoded MAX_CPUS / 8 from src/bpf/intf.h, right now this header is a mess */
+struct scx_cpumask {
+	union sdt_id tid;
+	u8 mask[64];
```
htejun, line 7: Maybe use u64's? BPF is a lot more efficient with 64bit operations. With smaller types, it'd be generating a lot of masking and shifting instructions.

scheds/rust/scx_rusty/src/bpf/cpumask.h
```diff
@@ -0,0 +1,199 @@
...
+{
+	struct sdt_data __arena *data = NULL;
+	scx_cpumask_t mask;
+	int i;
+
+	data = sdt_alloc(&scx_mask_allocator);
+	cast_kern(data);
+
+	mask = (scx_cpumask_t)data->payload;
+	mask->tid = data->tid;
```
htejun, line 33: It's fine for now but I wonder whether allocation ID linking can be hidden inside `sdt_alloc()` down the line. It's a bit weird because `sdt` originally meant `scx data for task` and the `tid` is unique `task ID`. The names and explicit `tid` handling make sense for task data allocations but get confusing when generalized like this. Something to clean up down the line.

scheds/rust/scx_rusty/src/bpf/cpumask.h
```diff
@@ -0,0 +1,199 @@
...
+{
+	sdt_subprog_init_arena();
+
+	sdt_free_idx(&scx_mask_allocator, mask->tid.idx);
+	return 0;
+}
+
+
+__weak
+int scxmask_set_cpu(u32 cpu, scx_cpumask_t __arg_arena mask)
```
htejun, line 57: There are three prefixes in use - `scx_cpumask`, `scx_mask` and `scxmask`. If this can be generalized in terms of number of bits (which we do need), maybe use `scx_bitmap`? In kernel, cpumasks are just wrapping bitmaps too.

scheds/rust/scx_rusty/src/bpf/cpumask.h
```diff
@@ -0,0 +1,199 @@
...
+	return 0;
+}
+
+static
+bool scxmask_subset_cpumask(scx_cpumask_t __arg_arena big, const struct cpumask *small)
+{
+	int i;
+
+	bpf_for(i, 0, nr_cpu_ids) {
+		if (!scxmask_test_cpu(i, big) && bpf_cpumask_test_cpu(i, small))
```
htejun, line 164: This can be done word by word, right?  `~mask[i] & mask[j]`

scheds/rust/scx_rusty/src/bpf/cpumask.h
```diff
@@ -0,0 +1,199 @@
...
+	return true;
+}
+
+static
+bool scxmask_intersects_cpumask(scx_cpumask_t __arg_arena scx, const struct cpumask *cpu)
+{
+	int i;
+
+	bpf_for(i, 0, nr_cpu_ids) {
+		if (scxmask_test_cpu(i, scx) && bpf_cpumask_test_cpu(i, cpu))
```
htejun, line 177: `mask[i] & mask[j]`

scheds/rust/scx_rusty/src/bpf/cpumask.h
```diff
@@ -0,0 +1,199 @@
...
+
+static
+int scxmask_and_cpumask(scx_cpumask_t dst __arg_arena,
+			       scx_cpumask_t scx __arg_arena,
+			       const struct cpumask *cpu __arg_trusted)
+{
+	int i;
+
+	bpf_for(i, 0, nr_cpu_ids) {
+		if (scxmask_test_cpu(i, scx) && bpf_cpumask_test_cpu(i, cpu))
```
htejun, line 192: `mask[i] &= mask[j]`

scheds/rust/scx_rusty/src/bpf/cpumask.h
etsal: For sure, cpumask_t also uses an array of longs for the mask so this would be consistent with the kernel bitmasks

scheds/rust/scx_rusty/src/bpf/cpumask.h
etsal: These are comparisons between a const struct cpumask * and an scxmask so we can't access the cpumask's bitmap directly AFAICT.

scheds/rust/scx_rusty/src/bpf/cpumask.h
etsal: Ditto

scheds/rust/scx_rusty/src/bpf/cpumask.h
etsal: These are different mask types and one of them is a kptr that we cannot dereference directly (there may be a way to do that with a kernel patch though? IIRC we can direclty read some fields from task_struct. Looking into it rn)

scheds/rust/scx_rusty/src/bpf/cpumask.h
etsal: The tid is definitely a misnomer, because what we really care about is the idx of the allocation within the radix tree.

The reason the tid is embedded in the struct is to avoid having a separate KV map of arena pointers to idx'es. The sdt_task allocator must be able to match a struct proc to the arena allocation we've made for it, and that map also holds the TID in the value. For cpumasks we don't need that kind of map if we embed the tid in the struct.

The type-specific code in cpumask.h doesn't do anything with the TID, it just stores it. We can make a wrapper struct for the cpumasks to remove the tid from the actual struct we give out during allocations and  completely hide tids from outside the allocator code.

htejun: Yeah, later, we can create an allocation wrapper which puts the ID after the actual allocation (to not disturb alignment requirement) so that the detail can be hidden from allocator users.

scheds/rust/scx_rusty/src/bpf/cpumask.h
htejun: Ah, can't we convert them and then just operate on the scx bitmasks?

etsal: We can, I will adjust these accordingly.

scheds/rust/scx_rusty/src/bpf/cpumask.bpf.c
```diff
@@ -0,0 +1,230 @@
...
+}
+
+__weak int
+scx_bitmap_to_bpf(struct bpf_cpumask __kptr *bpfmask __arg_trusted,
+		   scx_bitmap_t __arg_arena scx_bitmap)
+{
+	struct scx_bitmap *tmp;
+	int ret, i;
+
+	if (bpf_ksym_exists(bpf_cpumask_from_bpf_mem)) {
```
etsal, line 136: This is the only kfunc required for this patch. It populates a bpf cpumask from BPF memory.

scheds/rust/scx_rusty/src/bpf/cpumask.bpf.c
```diff
@@ -0,0 +1,230 @@
...
+	return 0;
+}
+
+__weak int
+scx_bitmap_from_bpf(scx_bitmap_t __arg_arena scx_bitmap, const cpumask_t *bpfmask __arg_trusted)
+{
+	int i;
+
+	#pragma unroll
+	for (i = 0; i < sizeof(cpumask_t) / 8; i++) {
```
etsal, line 177: / 8 because the bitmap is in longs.

scheds/rust/scx_rusty/src/bpf/cpumask.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+#include <scx/common.bpf.h>
+
+#include <lib/sdt_task.h>
+
+int bpf_cpumask_from_bpf_mem(struct bpf_cpumask *dst, void *src, size_t src__sz) __ksym __weak;
+
+#define SCXMASK_NLONG (512 / 8)
```
etsal, line 8: This should ideally be taken from intf.h where we define the maximum bitmask size supported by userspace.

scheds/rust/scx_rusty/src/bpf/cpumask.h
```diff
@@ -0,0 +1,44 @@
...
+
+struct scx_bitmap {
+	union sdt_id tid;
+	u64 bits[SCXMASK_NLONG];
```
etsal, line 12: This is not actually the real size of the bitmap. Since the bitmap is in arena memory, we can overlay multiple instances so that the unused bits of a struct instance are shared with another instance. This is done in the allocator logic which allocate space for each scx_bitmap based on the value of mask_size, not based on the size of the struct.

scheds/rust/scx_rusty/src/bpf/lb_domain.h
```diff
@@ -6,9 +6,9 @@ struct lb_domain {
 	union sdt_id		tid;
 
 	struct bpf_spin_lock vtime_lock;
-	struct bpf_cpumask __kptr *cpumask;
-	struct bpf_cpumask __kptr *direct_greedy_cpumask;
-	struct bpf_cpumask __kptr *node_cpumask;
+	scx_bitmap_t cpumask;
+	scx_bitmap_t direct_greedy_cpumask;
+	scx_bitmap_t node_cpumask;
```
etsal, line 11: These can be moved into the arena dom_ctx struct in a subsequent patch.

scheds/rust/scx_rusty/src/bpf/percpu.h
```diff
@@ -0,0 +1,163 @@
+#pragma once
```
etsal, line 1: factored out the per-CPU storage routines we use in scx_rusty, because they are currently used by the bitmap allocation logic (we need a non-arena scx_bitmap to pass to the kfunc, and we cannot stack allocate it in the function because it is too large).

scheds/rust/scx_rusty/src/bpf/cpumask.bpf.c
```diff
@@ -0,0 +1,230 @@
...
+int scx_bitmap_free(scx_bitmap_t __arg_arena mask)
+{
+	sdt_subprog_init_arena();
+
+	sdt_free_idx(&scx_bitmap_allocator, mask->tid.idx);
+	return 0;
+}
+
+__weak
+int scx_bitmap_copy_to_stack(struct scx_bitmap *dst, scx_bitmap_t __arg_arena src)
```
etsal, line 54: NOTE: This function fails to verify because of a libbpf CO:RE bug fixed in this commit: https://lore.kernel.org/bpf/58436ca32a9ba1fb1cad6d822d6dbbd926ac2375.camel@gmail.com/T/#t

PR #1295, opened 2025-02-04
scx_layered: various code optimisations/simplifications.

devnexen:


scheds/rust/scx_layered/src/main.rs
```diff
@@ -2477,14 +2474,10 @@ fn verify_layer_specs(specs: &[LayerSpec]) -> Result<()> {
...
-                bail!("Non-terminal spec {:?} has NULL matches", spec.name);
-            }
-        } else {
-            if spec.matches.len() != 1 || spec.matches[0].len() != 0 {
-                bail!("Terminal spec {:?} must have an empty match", spec.name);
-            }
+        if idx < nr_specs - 1 && spec.matches.is_empty() {
+            bail!("Non-terminal spec {:?} has NULL matches", spec.name);
+        } else if spec.matches.len() != 1 || !spec.matches[0].is_empty() {
+            bail!("Terminal spec {:?} must have an empty match", spec.name);
```
htejun, line 2480: This is not equivalent and I think would trigger spuriously.

devnexen: yes I can see in the logs now.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -2477,12 +2474,10 @@ fn verify_layer_specs(specs: &[LayerSpec]) -> Result<()> {
     }
 
     for (idx, spec) in specs.iter().enumerate() {
-        if idx < nr_specs - 1 {
-            if spec.matches.len() == 0 {
-                bail!("Non-terminal spec {:?} has NULL matches", spec.name);
-            }
+        if idx < nr_specs - 1 && spec.matches.is_empty() {
```
htejun, line 2477: This changes the condition for the else block which breaks the test.

htejun (approved):

PR #1298, opened 2025-02-04
scx_lavd: Use BPF CO-RE for SCX_ENQ_CPU_SELECTED

multics69:
There could be a case such that the scheduler is compiled against the latest vmlinux.h, which has SCX_ENQ_CPU_SELECTED, but runs on the old kernel, which does not have SCX_ENQ_CPU_SELECTED. In this case, !(enq_flags & SCX_ENQ_CPU_SELECTED) will be always true, which is unintended. So use BPF CO-RE to check if SCX_ENQ_CPU_SELECTED exists in runtime.


arighi (approved):
We should do this also for bpfland and qmap. Should we provide a helper in `scx/common.bpf.h`? Something like `bool is_cpu_selected(u64 enq_flags)`?

But this looks good for now.

multics69:
Thanks @arighi for the suggestion. I will send out a separate PR.

PR #1299, opened 2025-02-04
ci: avoid synchronising scheduled jobs

JakeHillion:
Every 6 hours 4 scheduled workflows trigger at exactly the same time, and because
each currently requires our full quantity of 20 scheduled runners in parallel
this guarantees a lot of queueing if you push a change just after.

Set each scheduled workflow to a random minute and manually specify the "every
6 hours" part with a rotation for the 3 jobs that run every 6 hours. This means
that there's a far better chance you'll be behind 1 job instead of 4 when you
push.

Drive by: make the alignment of the `on:` section consistent with the other
files and with the `jobs:` section.

### DIFF
--- a/.github/workflows/bpf-next-test.yml
+++ b/.github/workflows/bpf-next-test.yml
@@ -1,9 +1,9 @@
 name: bpf-next-test
 
 on:
-  # only runs on main, every 6 hours
-    schedule:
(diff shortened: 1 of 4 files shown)


hodgesds (approved):
LGTM, would be nice if the `schedule` action could take a random value.

PR #1301, opened 2025-02-04
scx_utils: Introduce is_enq_cpu_selected() for BPF compatibility

multics69:
Introduce a helper function, is_enq_cpu_selected(), to test SCX_ENQ_CPU_SELECTED in a compatible way.

BPF schedulers that directly use SCX_ENQ_CPU_SELECTED would have a compatibility bug if the schedulers are compiled against vmlinux.h that has SCX_ENQ_CPU_SELECTED but run on an old kernel that does not have SCX_ENQ_CPU_SELECTED. 

is_enq_cpu_selected() addresses such a problem using BPF CO-RE. Also, this PR changes BPF schedulers -- scx_qmap, scx_bpfland, and scx_lavd -- to use is_enq_cpu_selected().

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -529,6 +529,23 @@ static inline bool time_in_range_open(u64 a, u64 b, u64 c)
  * Other helpers
  */
 
+static inline bool is_enq_cpu_selected(u64 enq_flags)
+{
+	u64 flag;
+
+	/*
+	 * When the kernel did not have SCX_ENQ_CPU_SELECTED,
+	 * select_task_rq_scx() has never been skipped. Thus, this case
+	 * should be considered that the CPU has already been selected.
+	 */
+	if (!bpf_core_enum_value_exists(enum scx_enq_flags,
+					SCX_ENQ_CPU_SELECTED))
+		return true;
+
+	flag = bpf_core_enum_value(enum scx_enq_flags, SCX_ENQ_CPU_SELECTED);
+	return enq_flags & flag;
+}
+
 /* useful compiler attributes */
 #define likely(x) __builtin_expect(!!(x), 1)
 #define unlikely(x) __builtin_expect(!!(x), 0)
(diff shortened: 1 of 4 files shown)


multics69:
@htejun -- Once this PR lands, I will submit the corresponding patch for the kernel.

scheds/include/scx/common.bpf.h
```diff
@@ -529,6 +529,23 @@ static inline bool time_in_range_open(u64 a, u64 b, u64 c)
  * Other helpers
  */
 
+static inline bool is_enq_cpu_selected(u64 enq_flags)
```
htejun, line 532: Can you put this into `compat.bpf.h` and name it `__COMPAT_is_enq_cpu_selected()` and set the expiration release target similar to other compat features?

multics69: I am a bit confused what the COMPAT macro for in this case. I think we can just change the implementation of is_enq_cpu_selected() as follows at some point:

```
static inline bool is_enq_cpu_selected(u64 enq_flags)
{
	return enq_flags & SCX_ENQ_CPU_SELECTED;
}
```

Am I missing something?

multics69:
Once #1307 lands, we can further improve is_enq_cpu_selected() something like this:

```.c
static inline bool is_enq_cpu_selected(u64 enq_flags)
{
#ifdef SCX_ENQ_CPU_SELECTED
	u64 flag;

	/*
	 * When the kernel did not have SCX_ENQ_CPU_SELECTED,
	 * select_task_rq_scx() has never been skipped. Thus, this case
	 * should be considered that the CPU has already been selected.
	 */
	if (!bpf_core_enum_value_exists(enum scx_enq_flags,
					SCX_ENQ_CPU_SELECTED))
		return true;

	flag = bpf_core_enum_value(enum scx_enq_flags, SCX_ENQ_CPU_SELECTED);
	return enq_flags & flag;
#else
	return true;
#endif
}

```

multics69:
Let me close this PR for now. I will open a new PR once a better approach is discovered.

scheds/include/scx/common.bpf.h
htejun: `__COMPAT` just indicates that it's a temporary workaround for backward and forward compatibility and will be dropped in the future. We don't want to keep accumulating permanent helpers for these things, so it's just a way to explicitly time-limit these helpers.

multics69: @htejun -- ah, I got it.

PR #1302, opened 2025-02-04
ci: move caching-build to nix built kernel

JakeHillion:
This has been building for a while and is stable, and now has been running on the scheduled jobs for a day or so. Merging to enable the `caching-build` workflow which covers push/PRs/merge queue events, and removes the non-nix non-self-hosted kernel build from the final workflow.

Test plan:
- CI

### DIFF
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -26,74 +26,11 @@ jobs:
       - run: cargo fmt
       - run: git diff --exit-code
 
-  build-kernel-nix:
+  build-kernel:
     uses: ./.github/workflows/build-kernel.yml
     with:
       git-repo: "https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git"
       branch: for-6.13
-
-  build-kernel:
-    runs-on: ubuntu-24.04
-    steps:
-      # prevent cache permission errors
-      - run: sudo chown root /usr/bin/tar && sudo chmod u+s /usr/bin/tar
-      # redundancy to exit fast
-      - run: echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-selections
-      - run: sudo apt-get update
-      - run: sudo apt-get install -y git --no-install-recommends
-      - run: echo "SCHED_EXT_KERNEL_COMMIT=$(git ls-remote https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git heads/for-6.13 | awk '{print $1}')" >> $GITHUB_ENV
-
-      - uses: actions/checkout@v4
-
-      # use cached kernel if available, create after job if not
-      - name: Cache Kernel
-        id: cache-kernel
-        uses: actions/cache@v4
-        with:
-          path: |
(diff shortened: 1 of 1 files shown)


likewhatevs:
misc. btw -- iiuc how the nix host works etc, add ccache to it and maybe you can delete most of the cache logic (although probably best to keep it around for when jobs overflow to gh runners).

.github/workflows/caching-build.yml
```diff
@@ -103,9 +40,15 @@ jobs:
             scheduler: [ scx_bpfland, scx_lavd, scx_layered, scx_rlfifo, scx_rustland, scx_rusty ]
           fail-fast: false
     steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/restore-kernel-cache
+        with:
+          git-repo: "https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git"
+          branch: for-6.13
```
hodgesds, line 48: do we want `for-6.14` now?

.github/workflows/caching-build.yml
```diff
@@ -192,9 +117,15 @@ jobs:
             antistall: ['', '--disable-antistall']
           fail-fast: false
     steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/restore-kernel-cache
+        with:
+          git-repo: "https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git"
+          branch: for-6.13
```
hodgesds, line 125: should branch also be a variable?

hodgesds (approved):

JakeHillion:
> misc. btw -- iiuc how the nix host works etc, add ccache to it and maybe you can delete most of the cache logic

`ccache` is possible inside Nix but I had pretty bad experiences with it in a previous job, and it kind of goes against the grain of being reproducible (it can work though). I think the better approach here is going to be doing some of the builds with Nix instead of inside a devshell and caching the results. For example, we can do the kernel build in Nix then use that result later with some Cachix or FlakeHub action and everything will work correctly.

> (although probably best to keep it around for when jobs overflow to gh runners).

This doesn't happen currently and I don't think it's possible. It's a bit weird but I guess it makes sense, if you request a certain class of runner you'll either queue forever and fail or run on that runner (as we saw with forks). To mitigate this I'm tempted to get a second smaller runner in a different region to make sure jobs can always complete, but again do I really want the kernel job to run on a smaller runner when it's a miss? Need to have a think about what's best here.

.github/workflows/caching-build.yml
JakeHillion: @likewhatevs what do you think? I'm not sure what the logic behind picking 6.13 was and whether 6.14 RCs make sense, I just copied this across for now.

.github/workflows/caching-build.yml
JakeHillion: A variable at the top of this workflow you mean? It is duplicated a lot which is kind of annoying and will get changed in the future, so could make sense. I'm not sure what the best way to do that is though - will have a think.

PR #1303, opened 2025-02-04
lib/sdt_alloc: move sdt_static definition to sdt_task.h header

etsal:
The definition for the sdt_static allocator is currently in the lib/sdt_task.bpf.c. The definition is not picked up into the skeleton for scx_sdt, causing build failures. Move the definition of the struct into scheds/include/lib/sdt_task.h.


multics69 (approved):
Thanks @etsal ! It works for me.

JakeHillion (approved):

sirlucjan (approved):
Compiled fine

PR #1308, opened 2025-02-05
scx_utils: accessing GPU SMs (aka Streaming Multiprocessor).

devnexen:
the rust wrapper, while binding it, does not expose it and the wrapper does lazy library call loadings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2383,6 +2383,7 @@ dependencies = [
  "libc",
  "log",
  "nvml-wrapper",
+ "nvml-wrapper-sys",
  "paste",
  "regex",
  "scx_stats",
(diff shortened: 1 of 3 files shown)


likewhatevs:
Not super familiar with this, but IIUC loading nvml twice might not be a good idea.

I'll soon-ish be adding some code to scx_layered to enable using kprobes (nvidia_mmap in particular, I think) to get PIDs using GPUs instead of using NVML to get PIDs using GPUs due to how problematic using NVML is. 

WRT/ NVML being problematic to use, idk the details beyond "if you call it concurrently, some percentage of the time, it'll crash the system".

devnexen:
> Not super familiar with this, but IIUC loading nvml twice might not be a good idea.
> 
> I'll soon-ish be adding some code to scx_layered to enable using kprobes (nvidia_mmap in particular, I think) to get PIDs using GPUs instead of using NVML to get PIDs using GPUs due to how problematic using NVML is.
> 

Ah will like to see that.
> WRT/ NVML being problematic to use, idk the details beyond "if you call it concurrently, some percentage of the time, it'll crash the system".

Fair point, yes my approach is a bit hacky.

PR #1309, opened 2025-02-05
scxtop: clean up duplicated entry insertion code

JakeHillion:
Clean up uses of `.entry` by creating the `Vec` empty first then always pushing
in the same way.

Test plan:
- It builds.
- It generated a valid trace.


hodgesds (approved):
order of operations, nice 👍

PR #1311, opened 2025-02-05
scxtop: use struct-in-tuples for Action enum instead of structs

JakeHillion:
Currently we have a few functions that take an entire `&Action` but only match
on one specific case and `panic` otherwise. The invariant is that they're only
called with that case.

This can be checked by the type system, but unfortunately it's not possible to
specify that you're taking an `&Action::SetState`. To get the correct behaviour
here we can instead use tuple enum variants and place a named struct in there.
This gets rid of those ugly panics. Admittedly it's quite a noisy change, but
should make things better for when we inevitably forget those invariants.

Test plan:
- It builds.


hodgesds (approved):
LGTM

PR #1312, opened 2025-02-05
rust/scx_utils: rebuild scheduler when headers are modified

etsal:
The schedulers that use multiple compilation units only generate Clang rerun-if-changed directives for the *.bpf.c files explicitly added to build.rs. Expand the code to also rebuild the scheduler if any of its local *.h files are modified.


htejun (approved):

likewhatevs (approved):
lgtm

PR #1313, opened 2025-02-06
scxtop: adding one perf event type `L1-dcache-load-misses`.

devnexen:
Gives a hint if the there is a suboptimal CPU cache locality, forcing
 to slower CPU levels.


hodgesds (approved):
LGTM, one thing I think we should check is on hardware that doesn't support that event type if it is handled properly.

PR #1314, opened 2025-02-06
scripts, scx_utils, scheds: Compatible testing of `SCX_ENQ_CPU_SELECTED`

multics69:
This PR is for testing `SCX_ENQ_CPU_SELECTED` in a compatible way. More specifically, it handles two cases:
1. a BPF scheduler is compiled against `vmlinux.h` where `SCX_ENQ_CPU_SELECTED` is defined but it runs on a kernel that does not have `SCX_ENQ_CPU_SELECTED`
2. a BPF scheduler is compiling against `vmlinux.h` where `SCX_ENQ_CPU_SELECTED` is not defined.

The PR consists of three parts:
1. Add a script, `scripts/gen_enum_defs.py`, which generates a header file `enum_defs.autogen.h` from a given `vmlinux.h`. `enum_defs.autogen.h` has macros denoting whether SCX enums are defined in the `vmlinux.h` or not.
3. Implement `__COMPAT_is_enq_cpu_selected()`, which provide the test of `SCX_ENQ_CPU_SELECTED` in a compatible way.
4. Use  `__COMPAT_is_enq_cpu_selected()` in scx schedulers: scx_bpfland, scx_lavd, and scx_qmap.


scheds/c/scx_qmap.bpf.c
```diff
@@ -231,7 +231,7 @@ void BPF_STRUCT_OPS(qmap_enqueue, struct task_struct *p, u64 enq_flags)
 	}
 
 	/* if select_cpu() wasn't called, try direct dispatch */
-	if (!(enq_flags & SCX_ENQ_CPU_SELECTED) &&
+	if (!__COMPAT_is_enq_cpu_selected(SCX_ENQ_CPU_SELECTED) &&
```
htejun, line 234: Shouldn't this be `__COMPAT_is_enq_cpu_selected(enq_flags)`?

scheds/include/scx/compat.bpf.h
```diff
@@ -125,6 +125,58 @@ bool scx_bpf_dispatch_vtime_from_dsq___compat(struct bpf_iter_scx_dsq *it__iter,
...
+	 * where the enum SCX_ENQ_CPU_SELECTED exists.
+	 */
+
+	/*
+	 * We should temporarily suspend the macro expansion of
+	 * 'SCX_ENQ_CPU_SELECTED'. This avoids 'SCX_ENQ_CPU_SELECTED' being
+	 * rewritten to '__SCX_ENQ_CPU_SELECTED' when 'SCX_ENQ_CPU_SELECTED'
+	 * is defined in 'scripts/gen_enums.py'.
+	 */
+#pragma push_macro("SCX_ENQ_CPU_SELECTED")
```
htejun, line 150: Ooh, fancy. Yeah, this is a nice way to work around the problem.

etsal:
Really nice! I wonder if the gen_enum_defs.py and gen_enums.py are worth merging at some point since gen_enum_defs.py.

etsal (approved):

scheds/c/scx_qmap.bpf.c
multics69: Haha, a mistake in copy & paste.

multics69:
> I wonder if the gen_enum_defs.py and gen_enums.py are worth merging at some point since gen_enum_defs.py gets all SCX definitions from vmlinux.h.

Thank you for bringing this @etsal ! I had also considered it, but I decided to create a separate script. However, it is worth to discuss.

- The enums listed in `gen_enum_defs.py` must be present in the kernel. Otherwise, the loading of a BPF scheduler will fail. 
   https://github.com/sched-ext/scx/blob/c01587da4eb25d45d70987e608b16100d1011251/scheds/include/scx/enums.h#L17)
- So I think it would be proper to have a helper function (like `__COMPAT_is_enq_cpu_selected`). After 2-3 kernel releases, remove the helper function and add the new enum to  `gen_enum_defs.py`. 

Another approach would be changing `SCX_BUG_ON()` to `SCX_WARN_ON()` in the following, allowing the loading of a BPF scheduler in an old kernel. I thought that approach would be risky because the BPF scheduler may mal-function silently. 

   https://github.com/sched-ext/scx/blob/c01587da4eb25d45d70987e608b16100d1011251/scheds/include/scx/enums.h#L17)


What do you think?

htejun:
I'd prefer to keep `tools/sched_ext` in the kernel tree as simple as possible.

htejun (approved):

htejun:
> I'd prefer to keep `tools/sched_ext` in the kernel tree as simple as possible.

That said, we need to sync back the `.autogen.h` file anyway, so might as well commit the script to generate it together. Either way is fine by me.

multics69:
> > I'd prefer to keep `tools/sched_ext` in the kernel tree as simple as possible.
> 
> That said, we need to sync back the `.autogen.h` file anyway, so might as well commit the script to generate it together. Either way is fine by me.

Okay.  Then I will upload the `.autogen.h` file without the script for now. Once we have more clear idea how to manage `gen_enums.py` and `gen_enum_defs.py`, let's revisit it.

PR #1316, opened 2025-02-07
scx_p2dq: Add scx_p2dq scheduler

hodgesds:
Add p2dq scheduler. The p2dq scheduler is a simple load balancing scheduler that uses a pick two algorithm for load balancing. A fixed number of DSQs are created per LLC with incremental slice intervals. If a task is able to consume the majority of the assigned slice is it dispatched to a DSQ with a longer slice. Tasks that do not consume more than half the slice are moved to shorter slice DSQs. The DSQs with the shortest slice lengths are then determined to be "interactive". All DSQs on the same LLC share the same vtime and there is special handling for (non)interactive tasks.

It's kind of far enough along to merge, but there's still a lot of edge cases that aren't handled yet.


scheds/rust/scx_p2dq/src/bpf/intf.h
```diff
@@ -0,0 +1,108 @@
...
+	u64 			last_run_at;
+
+	/* The task is a workqueue worker thread */
+	bool			is_kworker;
+
+	/* Allowed on all CPUs and eligible for DIRECT_GREEDY optimization */
+	bool			all_cpus;
+
+	/* select_cpu() telling enqueue() to queue directly on the DSQ */
+	bool			dispatch_local;
```
JakeHillion, line 70: Shouldn't be needed on modern kernels

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1006 @@
...
+	stat_add(idx, 1);
+}
+
+/*
+ * Returns if the task is interactive based on the last DSQ.
+ */
+static bool is_interactive(struct task_ctx *taskc)
+{
+	if (nr_dsqs_per_llc <= 1)
+		return false;
```
JakeHillion, line 194: Nothing is interactive in the single DSQ case? A comment here explaining why `false` is picked over `true` would be nice.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1006 @@
...
+
+	if (!(llcx = lookup_llc_ctx(cpuc->llc_id)))
+		return false;
+
+	int nr_queued = 0;
+	bpf_for(i, 0, nr_dsqs_per_llc) {
+		// ignore interactive tasks
+		if (i == 0)
+			continue;
```
JakeHillion, line 224: This doesn't match the `is_interactive` check. That says if there is 1 queue per LLC nothing is interactive, whereas this says if there is 1 queue per LLC everything is interactive. Maybe that's okay but it's not clear why.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1006 @@
...
+	u64 now = scx_bpf_now();
+	last_dsq_slice_ns = dsq_time_slice(taskc->dsq_index);
+	used = now - taskc->last_run_at;
+	taskc->last_run_at = now;
+	taskc->last_dsq_id = taskc->dsq_id;
+	cpuc->dsq_index = taskc->dsq_index;
+
+	// On stopping determine if the task can move to a longer DSQ by
+	// comparing the used time to the scaled DSQ slice.
+	if (used >= ((9 * last_dsq_slice_ns) / 10)) {
```
JakeHillion, line 608: Is it worth making this `7/8`s or `15/16`s to make the division easier? The 9/10ths seems arbitrary but perhaps there's a good reason.

scheds/rust/scx_p2dq/src/main.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        let mut cpu_ctxs: Vec<bpf_intf::cpu_ctx> = vec![];
+        let cpu_ctxs_vec = skel
+            .maps
+            .cpu_ctxs
+            .lookup_percpu(&0u32.to_ne_bytes(), libbpf_rs::MapFlags::ANY)?
+            .unwrap();
+
+        for (cpu_id, cpu_arc) in topo.all_cpus {
+            let cpu = cpu_arc.clone();
```
JakeHillion, line 196: nit: do we need this clone/variable? It's cloning the `Arc<Cpu>`, incrementing the reference count, doing read only things, then dropping it at the end. These operations should all be possible on the `&Arc<Cpu> cpu_arc` (which could be renamed to `cpu`).

JakeHillion (approved):
Nice, great to see this landing!

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1006 @@
...
+	taskc->runnable = true;
+
+	if (!all_cpumask) {
+		scx_bpf_error("NULL all_cpumask");
+		return -EINVAL;
+	}
+
+	bpf_rcu_read_lock();
+	if (p->cpus_ptr && all_cpumask)
+		taskc->all_cpus = bpf_cpumask_subset(cast_mask(all_cpumask),
```
arighi, line 754: Maybe we could simply check `p->nr_cpus_allowed == nr_cpus`? And in this way we can also get rid of the RCU locking.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1006 @@
...
+				   enq_flags | SCX_ENQ_PREEMPT);
+		return;
+	}
+
+	/* 
+	 * Affinitized tasks just get dispatched directly, need to handle this better 
+	 */
+	if ((!taskc->all_cpus)) {
+		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, slice_ns, enq_flags);
+	}
```
multics69, line 534: Missing return?

multics69 (approved):
Thank you for sharing the code. The core idea looks promising. I am eager to see how it works. I left a few comments.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1006 @@
...
+			max_dsq = i;
+		}
+	}
+
+	// First try the DSQ with the most load.
+	if (max_dsq >= 0 && max_load > 0) {
+		dsq_id = *MEMBER_VPTR(cpuc->dsqs, [max_dsq]);
+		if (scx_bpf_dsq_move_to_local(dsq_id))
+			return;
+	}
```
multics69, line 681: What is the idea behind picking the most loaded DSQ?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1006 @@
...
+ * This software may be used and distributed according to the terms of the
+ * GNU General Public License version 2.
+ *
+ * scx_p2dq is a scheduler where the load balancing is done using a pick 2
+ * algorithm.
+ */
+
+#ifdef LSP
+#define __bpf__
+#include "../../../../include/scx/common.bpf.h"
```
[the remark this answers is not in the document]
likewhatevs: that enables fancy code completion/go-to-definition/type checking while editing bpf.c files:

Screenshot from 2025-02-07 09-30-44

if you remove that those features largely stop working:

Screenshot from 2025-02-07 09-34-57

scheds/rust/scx_p2dq/src/bpf/intf.h
hodgesds: Yeah, lemme remove that. bad copy/paste

scheds/rust/scx_p2dq/src/main.rs
```diff
@@ -0,0 +1,343 @@
...
+struct Opts {
+    /// Disables per-cpu kthreads directly dispatched into local dsqs.
+    #[clap(short = 'k', long, action = clap::ArgAction::SetTrue)]
+    disable_kthreads_local: bool,
+
+    /// If specified, only tasks which have their scheduling policy set to
+    /// SCHED_EXT using sched_setscheduler(2) are switched. Otherwise, all
+    /// tasks are switched.
+    #[clap(short = 'p', long, action = clap::ArgAction::SetTrue)]
+    partial: bool,
```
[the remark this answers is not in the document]
hodgesds: Probably not, let me remove.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: This is only for the case to allow a task to keep running beyond it's slice. I guess I was thinking that before a task should be allowed to run for more than it's slice it should at least have completed the most interactive slice.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: Thanks, nice catch!

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1006 @@
...
+
+
+void BPF_STRUCT_OPS(p2dq_set_weight, struct task_struct *p, u32 weight)
+{
+	struct task_ctx *taskc;
+
+	if (!(taskc = lookup_task_ctx(p)))
+		return;
+
+	taskc->weight = weight;
```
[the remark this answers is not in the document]
hodgesds: Yeah, it's not being used currently either, still a TODO so maybe I'll remove for now.

PR #1318, opened 2025-02-07
scxtop: track sched_waking events

JakeHillion:
Trace sched_waking and output them in the scxtop perfetto traces. This fills in
the per-process timelines with "Uninterruptible Sleep" and "Runnable".

It has the annoying side effect of sometimes putting "Wake Kill" at the end of
the trace. I think this is because we see the thread as not-runnable for the
last time then the trace terminates, but am unsure.

Drive by: removes the `run` function that is wrapped by `main` in favour of
calling it `main`.

Test plan:

Now some processes show states like this in places that fit the timeline:

image

I think we may be dropping some events because of this. I'm adding monitoring for it and hopefully have some solutions in follow up PRs.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -21,7 +21,9 @@ use crate::ViewState;
 use crate::APP;
 use crate::LICENSE;
 use crate::SCHED_NAME_PATH;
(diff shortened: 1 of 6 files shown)


hodgesds (approved):
LGTM

PR #1319, opened 2025-02-08
scx_p2dq: Fix vtime handling and cleanup

hodgesds:
- Fix vtime handling so that it progresses properly and tasks get vtime initialized.

- Remove workaround logic in dispatch for broken vtime handling.

- Add variable to control whether or not to store stats. If metric collection is disabled then stat collection is paused.

- Cleanup unused/unimplemented code.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -41,6 +41,9 @@ const volatile u64 dsq_shift = 2;
 const volatile int init_dsq_index = 0;
 const volatile u64 min_slice_us = 100;
 
+const volatile bool interactive_sticky = false;
+const volatile bool keep_running_enabled = true;
+const volatile bool eager_load_balance = true;
 const volatile bool smt_enabled = true;
 const volatile bool has_little_cores = false;
 const volatile bool kthreads_local;
@@ -177,7 +180,6 @@ struct {
 static inline void stat_add(enum stat_idx idx, u64 amount)
 {
 	u32 idx_v = idx;
-
 	u64 *cnt_p = bpf_map_lookup_elem(&stats, &idx_v);
 	if (cnt_p)
 		(*cnt_p) += amount;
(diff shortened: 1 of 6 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -655,44 +676,29 @@ void BPF_STRUCT_OPS(p2dq_dispatch, s32 cpu, struct task_struct *prev)
...
 	bpf_for(i, 0, nr_dsqs_per_llc) {
-		nr_queued = scx_bpf_dsq_nr_queued(*MEMBER_VPTR(cpuc->dsqs, [i]));
-		if (nr_queued <= 0)
-			continue;
-
-		dsq_load = i <= nr_queued ? nr_queued : nr_queued / i;
-		if (dsq_load > max_load) {
-			max_load = dsq_load;
-			max_dsq = i;
+		if (llcx->dsq_max_vtime[i] < max_vtime) {
```
arighi, line 681: Isn't this the other way around? (> instead of <)

hodgesds: The naming is kind of bad, it's the max vtime for that DSQ, but it should try to consume from the DSQ with the vtime that is most behind is the thinking.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -578,10 +583,13 @@ void BPF_STRUCT_OPS(p2dq_running, struct task_struct *p)
 	trace("RUNNING %d prev_cpu %d prev_llc %d llc %d", p->pid, taskc->cpu, taskc->llc_id, llcx->id);
 	taskc->last_run_at = scx_bpf_now();
 	taskc->llc_id = llcx->id;
-	llcx->vtime = p->scx.dsq_vtime;
 	cpuc->dsq_index = taskc->dsq_index;
 	cpuc->ran_for = 0;
 	taskc->cpu = task_cpu;
+	if (p->scx.dsq_vtime > llcx->vtime)
+		llcx->vtime = p->scx.dsq_vtime;
```
multics69, line 590: Would this be racy? If you don't care, better to comment on it?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -611,6 +619,8 @@ void BPF_STRUCT_OPS(p2dq_stopping, struct task_struct *p, bool runnable)
 	taskc->last_dsq_id = taskc->dsq_id;
 	taskc->last_dsq_index = taskc->dsq_index;
 	cpuc->dsq_index = taskc->dsq_index;
+	// racy, don't care
+	llcx->vtime += used;
```
multics69, line 623: It would be better to use atomic increment since vtime should be critical in making scheduling decision. The contention on atomically updating llcx->vtime should be ok since the size of llcx should be not that big.

multics69 (approved):

PR #1321, opened 2025-02-09
scx_lavd: Revert "scx_lavd: Kick an idle cpu as early as possible on ops.select_cpu()"

multics69:
Kicking a CPU at ops.select_cpu() causes performance regression in some scenarios. The reported case is as follows:

  - Game: Space Marine 2
  - Kernel:  6.13.1-2-cachyos
  - CPU: i9-13900KF
  - GPU: RTX 3700 Ti 570 drivers closed source GSP disabled
  - KDE Plasma Wayland with VRR Automatic

Since kicking a CPU at ops.select_cpu() is redundant -- after the ops.select_cpu(), the kernel will kick the CPU if necessary, it is safe to drop the kick CPU here.


hodgesds (approved):
LGTM

PR #1323, opened 2025-02-09
scx_lavd, scx_top: inlining builtin copies.

devnexen:
clang can still generate a memcpy call with __builtin_memcpy __builtin_memcpy_inline takes it a step further with compile time size value in exchange of a little binary size increase. supported since clang 15 and we want clang 16 min.


multics69 (commented):
Since the commit changes both scx_lavd and scx_top, please update the commit message to reflect this, for example, "scx_lavd, scx_top: inlining builtin copies". Also, please add SoB line.

multics69 (approved):

PR #1324, opened 2025-02-09
scx_layered: addressing TODOs.

devnexen:
topology creation seems to already ignore holes in topology though.


htejun:
For trivial / identity cleanups, omitting most of descriptions is fine but I think this would need more explanations on what changes are being made and why they're safe. e.g. The comment there doesn't just concern the immediate code block. There are iteration code in BPF side which also assumes that CPUs are consecutive. Are they vetted and tested? If so, how?

htejun:
Closing for now due to inactivity. Please feel free to reopen. Thanks.

PR #1325, opened 2025-02-10
scx_layered: making unexpected gpu error smoother.

devnexen:
nvml attempts to load the C library being, but scx_layered exits in a AMD setting.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -1935,7 +1935,10 @@ impl<'a> Scheduler<'a> {
             return Ok(());
         }
 
-        let nvml = NVML_CELL.get_or_try_init(|| Nvml::init())?;
+        let nvml = NVML_CELL.get_or_try_init(|| {
+            Nvml::init().context("enabling GPU support requires a nvidia device")
+        })?;
+
         let mut missing_gpu_pid = false;
 
         if !self.opts.aggressive_nvml_polling {


JakeHillion (changes requested):
This only happens if you run with `--enable-gpu-support` right? Perhaps that's a misnomer and should be `--enable-nvidia-gpu-support` at the minute, but I'm not sure we should silently ignore errors initialising the library when you've explicitly requested the support.

devnexen:
Well, would you like a log entry ? otherwise the scheduler just crash due to nvml.

JakeHillion:
> Well, would you like a log entry ? otherwise the scheduler just crash due to nvml.

An error message like "enabling gpu support requires an nvidia gpu" seems reasonable to me. You can add it to the existing error that crashes the scheduler with Anyhow's `.context()` trait. IMO the crash itself is the intended behaviour when the scheduler is started with a config/flags that can't be satisfied, though admittedly it'd be cleaner if we crashed in userspace before attaching the scheduler.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1935,7 +1935,10 @@ impl<'a> Scheduler<'a> {
             return Ok(());
         }
 
-        let nvml = NVML_CELL.get_or_try_init(|| Nvml::init())?;
+        let nvml = NVML_CELL.get_or_try_init(|| {
+            Nvml::init().context(format!("enabling GPU support requires a nvidia device"))
```
JakeHillion, line 1939:
```suggestion
            Nvml::init().context("enabling GPU support requires a nvidia device")
```

What's the `format` for?

devnexen: nothing.. was trying something else and forgot to remove it.

JakeHillion (approved):
Thanks for persevering!

PR #1329, opened 2025-02-11
scx_layered: Remove CPU shuffling when idle CPU is not found in layered_select_cpu()

htejun:
layered_select_cpu() returns a random in-layer CPU if an idle CPU cannot be found for a task which can run anywhere. The intention was to preventing putting a task into an LLC where the layer doesn't have a CPU as that triggers LLC draining which isn't free. However, as a random CPU is picked from the layer, this means that the task is likely to end up in an LLC which wasn't its previous LLC whenever there is no idle CPU which likely will lead to a lot more cross-LLC migrations.

Draining occassionally isn't that expensive and is definitely cheaper than triggering cross-LLC migrations all the time. Just return @prev_cpu if no idle CPU can be found.

In preliminary web workload saturation test at meta, this seems to meaningfully lower cross-llc migration rate and increase service bandwidth.


likewhatevs (approved):
makes sense, running some extra tests on this, should have results sometime tomorrow.

PR #1331, opened 2025-02-12
scx_bpfland: minor optimisations/modernisation.

devnexen:
- cpu mask instead of allocating with format!, we accumulate all.


rust/scx_rustland_core/assets/bpf.rs
```diff
@@ -585,7 +585,7 @@ impl<'cb> BpfScheduler<'cb> {
 }
 
 // Disconnect the low-level BPF scheduler.
-impl<'a> Drop for BpfScheduler<'a> {
+impl Drop for BpfScheduler<'_> {
```
arighi, line 588: This is scx_rustland_core, not bpfland. It should go in a separate patch.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -234,7 +234,7 @@ static int calloc_cpumask(struct bpf_cpumask **p_cpumask)
 	struct bpf_cpumask *cpumask;
 
 	cpumask = bpf_cpumask_create();
-	if (!cpumask)
+	if (unlikely(!cpumask))
```
arighi, line 237: Is there any actual benefit using likely/unlikely in BPF code? I think the JIT will just ignore this hint. Let's drop this for now.

scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -562,7 +562,7 @@ impl<'a> Scheduler<'a> {
     }
 }
 
-impl<'a> Drop for Scheduler<'a> {
+impl Drop for Scheduler<'_> {
```
arighi, line 565: This change could be applied pretty much to all the Rust schedulers. Maybe send a separate PR with this particular change targeting all schedulers, instead of just bpfland?

arighi (changes requested):

scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -99,11 +100,10 @@ fn cpus_to_cpumask(cpus: &Vec<usize>) -> String {
...
 
     // Convert the byte vector to a hexadecimal string.
-    let hex_str: String = bitmask
-        .iter()
-        .rev()
-        .map(|byte| format!("{:02x}", byte))
-        .collect();
+    let hex_str: String = bitmask.iter().rev().fold(String::new(), |mut f, byte| {
+        let _ = write!(&mut f, "{:02x}", byte);
+        f
```
htejun, line 105: I get that this is equivalent, but is this necessarily better?

devnexen: Not exactly the same, the change avoid a string allocation for each byte written.

arighi (approved):

PR #1333, opened 2025-02-12
build: Add script for pgo/lto builds

hodgesds:
Add a script to generate PGO/LTO builds. This may improve performance for some of the more userspace oriented schedulers.

If everything goes right and the toolchain supports it then a binary should be produced:
```
$ ./pgo-lto.sh /tmp/pgo-data scx_rustland 10
<lots of output>
   Compiling ordered-float v3.9.2
   Compiling simplelog v0.12.2
   Compiling scx_rustland_core v2.2.8 (/root/scx/rust/scx_rustland_core)
   Compiling fb_procfs v0.7.1
   Compiling scx_rustland v1.0.9 (/root/scx/scheds/rust/scx_rustland)
warning: scx_rustland_core.307c2d1e9a16060e-cgu.0: function control flow change detected (hash mismatch) __rust_realloc Hash = 2202711501531239 up to 467244 count discarded

warning: scx_rustland_core.307c2d1e9a16060e-cgu.0: function control flow change detected (hash mismatch) __rust_alloc_zeroed Hash = 2202711501531239 up to 198 count discarded

warning: `scx_rustland_core` (lib) generated 2 warnings
    Finished `release` profile [optimized] target(s) in 1m 59s
```

### DIFF
--- a/pgo-lto.sh
+++ b/pgo-lto.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+set -eou pipefail
+
+PGO_TMPDIR=${1:-/tmp/pgo-data}
+SCHED=${2:-scx_rustland}
+PGO_DUR=${3:-30}
+
+rm -rf "$PGO_TMPDIR"
+mkdir -p "$PGO_TMPDIR"
+
+RUSTFLAGS="-C profile-generate=$PGO_TMPDIR -C link-arg=-lgcov" \
+    cargo build --release --bin "$SCHED"
+
+echo "Running sched to generate PGO"
+for i in {0..3}; do
+	sudo "./target/release/$SCHED" &
+	sleep "$PGO_DUR"
+	sudo kill -9 $! || echo "$SCHED already dead"
+	sleep 1
+done
+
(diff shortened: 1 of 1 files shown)


JakeHillion (approved):
I don't see any reason not to do this, though I think to get the most benefit you'd have to run a realistic workload instead of the `sleep "$PGO_DUR"`. This one might end up optimising some of the early returns instead of the important bits if the machine is quiet for example.

Maybe we should throw this in a directory like `build`? Not sure if that name is already taken.

etsal:
Seems reasonable to me!

> I don't see any reason not to do this, though I think to get the most benefit you'd have to run a realistic workload instead of the `sleep "$PGO_DUR"`. This one might end up optimising some of the early returns instead of the important bits if the machine is quiet for example.

Given that most schedulers mostly do a bunch of initialization at startup then a bunch of periodic jobs in userspace that wouldn't be that big of an issue, e.g., scx_rusty still makes an attempt to load balance and traverses the relevant code paths even if the machine is relatively idle.

PR #1334, opened 2025-02-12
scxtop: make q exit help if in help instead of quitting

JakeHillion:
I cannot wrap my head around not pressing 'q' to close the help screen. Make
pressing 'q' while help is open quit the help screen rather than the
application.

Test plan:
- Opened help, pressed q, help closed. Pressed q again and the app quit.
- Pressed q without opening help, app quit.


hodgesds (approved):
LGTM

PR #1335, opened 2025-02-12
scxtop: track number of dropped events and render it

JakeHillion:
Adds a stat to track events which can't be inserted into the ringbuf. I can't
get this to hit on my big AMD machines which is a good sign. I've tried a
couple of prod services and some stress-ng configurations. Seems like dropping
events isn't an issue at the minute, but we should keep an eye on it.

There are likely better ways to render this, but at least this way we keep
track of it and can render it more appropriately in the future.

Test plan:
- Artificially added failures on the BPF side. They render as expected in
  `scxtop` during a trace.

### DIFF
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -33,6 +33,31 @@ struct {
 	__uint(max_entries, 10 * 1024 * 1024 /* 10Mib */);
 } events SEC(".maps");
 
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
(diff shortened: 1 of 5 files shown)


tools/scxtop/src/bpf/main.bpf.c
```diff
@@ -122,9 +147,9 @@ int BPF_KPROBE(scx_sched_reg)
 	if (!enable_bpf_events)
 		return 0;
 
-	event = bpf_ringbuf_reserve(&events, sizeof(struct bpf_event), 0);
-	if (!event)
+	if (!(event = try_reserve_event())) {
```
hodgesds, line 150: nit: might not need the extra brackets

PR #1336, opened 2025-02-12
aarch64 build fixes

etsal:
Fix the issues highlighted in #1320 to compile on systems with 16~KiB page sizes and for kernel configs without NUMA support.


hodgesds (approved):
LGTM, thanks for cleaning this up!

JohnRTitor:
At this point, I think we should have some kind of CI for aarch64 builds.

PR #1338, opened 2025-02-12
scx_layered: Update enqueue path to use direct dispatch into foreign local DSQs

htejun:
sched_ext now allows direct dispatch into foreign local DSQs from `ops.enqueue()`. Use it to make idle selection handling and preemption in `layered_enqueue()` simpler and more efficient.


hodgesds (approved):
Makes sense!

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -958,30 +962,16 @@ s32 BPF_STRUCT_OPS(layered_select_cpu, struct task_struct *p, s32 prev_cpu, u64
...
-		return false;
-	}
-}
-
-static __always_inline
-bool try_preempt_cpu(s32 cand, struct task_struct *p, struct cpu_ctx *cpuc,
-		     struct task_ctx *taskc, struct layer *layer,
-		     bool preempt_first)
+/*
+ * XXX - It'd be better to get @cpuc and @enq_flags from the caller but that
```
hodgesds, line 966: Maybe we need a `struct context_ctx` of all the different contexts or something to make it easier to maintain/verify.  This should be good for now.

htejun: Yeah, maybe.

PR #1340, opened 2025-02-13
scxtop: Add softirq events

hodgesds:
Add softirq event collection for perfetto traces.

Screenshot of trace:

### DIFF
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -27,6 +27,7 @@ struct bpf_event _event = {0};
 bool enable_bpf_events = true;
 u32 sample_rate = 128;
 
+const int zero_int = 0;
 
 struct {
 	__uint(type, BPF_MAP_TYPE_RINGBUF);
@@ -79,6 +80,19 @@ struct {
 	__uint(max_entries, 1000000);
 	__uint(map_flags, 0);
 } task_data SEC(".maps");
+
+
+struct __softirq_event {
+	u32		pid;
+	u64		start_ts;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__type(key, int);
+	__type(value, struct __softirq_event);
+	__uint(max_entries, 1);
+} softirq_events SEC(".maps");
 
 
 static __always_inline u64 t_to_tptr(struct task_struct *p)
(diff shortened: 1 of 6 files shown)


tools/scxtop/src/bpf/main.bpf.c
```diff
@@ -526,3 +541,56 @@ int BPF_PROG(on_sched_switch, bool preempt, struct task_struct *prev,
...
+		return 0;
+
+	if (!(event = try_reserve_event()))
+		return -ENOENT;
+
+	event->type = SOFTIRQ;
+	event->cpu = softirq_event->cpu;
+	event->ts = exit_ts;
+	event->event.softirq.pid = softirq_event->pid;
+	event->event.softirq.entry_ts = softirq_event->start_ts;
```
JakeHillion, line 589: Do we need `cpu` and `pid` from `softirq_event`, or could it simply be the entry timestamp? Wondering if it'd be kinder to the `irq` to do as much work as possible on `exit` rather than `entry`.

tools/scxtop/src/main.rs
```diff
@@ -228,8 +243,9 @@ async fn main() -> Result<()> {
             _ => {}
         }
         0
-    })?;
-    let rb = rbb.build()?;
+    };
+    event_rbb.add(&skel.maps.events, event_handler.clone())?;
```
JakeHillion, line 247: nit: does this need the `.clone()`?

tools/scxtop/src/main.rs
```diff
@@ -158,6 +162,17 @@ async fn main() -> Result<()> {
                 tx.send(action).ok();
             }
             #[allow(non_upper_case_globals)]
+            event_type_SOFTIRQ => unsafe {
```
JakeHillion, line 165: nit: this `unsafe` block is bigger than it needs to be. Could we settle on something like:

```
let softirq_event = unsafe { &event.event.softirq };
```

So we don't accidentally use other unsafe things in this block?

JakeHillion (approved):
LGTM. Going forward we might need the option to disable some of these probes when collecting a perfetto trace, the probe effect is already pretty high and when you don't need something that might end up being a problem. Can get to that later though.

tools/scxtop/src/main.rs
hodgesds: Ahh, guess it doesn't right now had left that over from testing mulitple ringbuffers.

tools/scxtop/src/perfetto_trace.rs
```diff
@@ -275,6 +275,31 @@ impl<'a> PerfettoTraceManager<'a> {
...
+                exit_event.set_vec(action.softirq_nr as u32);
+
+                entry_ftrace_event.set_timestamp(action.entry_ts);
+                entry_ftrace_event.set_softirq_entry(entry_event);
+                entry_ftrace_event.set_pid(action.pid);
+                exit_ftrace_event.set_timestamp(action.exit_ts);
+                exit_ftrace_event.set_softirq_exit(exit_event);
+                exit_ftrace_event.set_pid(action.pid);
+
+                vec![entry_ftrace_event, exit_ftrace_event]
```
JakeHillion, line 299: nit: I think (untested) that you can do this with a tuple and avoid allocating.

```suggestion
                (entry_ftrace_event, exit_ftrace_event)
```

tools/scxtop/src/main.rs
hodgesds: I think anything accessing the union has to be unsafe so the only thing that isn't unsafe is `cpu`. I'll maybe live lazy this time 🗡️

tools/scxtop/src/perfetto_trace.rs
hodgesds: Annoying, it seems the tuple doesn't implement `IntoIterator`:
```
error[E0277]: `(FtraceEvent, FtraceEvent)` is not an iterator
   --> tools/scxtop/src/perfetto_trace.rs:284:21
    |
284 |               .extend({
    |  ______________------_^
    | |              |
    | |              required by a bound introduced by this call
285 | |                 let mut entry_ftrace_event = FtraceEvent::new();
286 | |                 let mut exit_ftrace_event = FtraceEvent::new();
287 | |                 let mut entry_event = SoftirqEntryFtraceEvent::new();
...   |
299 | |                 (entry_ftrace_event, exit_ftrace_event)
    | |                 --------------------------------------- this tail expression is of type `(FtraceEvent, FtraceEvent)`
300 | |             });
    | |_____________^ `(FtraceEvent, FtraceEvent)` is not an iterator
    |
    = help: the trait `Iterator` is not implemented for `(FtraceEvent, FtraceEvent)`, which is required by `(FtraceEvent, FtraceEvent): IntoIterator`
    = note: required for `(FtraceEvent, FtraceEvent)` to implement `IntoIterator`
note: required by a bound in `extend`
```

hodgesds: I am dumb... `.into_iter()` exists

JakeHillion: I was dumb! Of course it doesn't, the types can be different. An array should actually work.

PR #1342, opened 2025-02-13
scxtop: drop unnecessary Arc and RwLock on tui

JakeHillion:
This showed up because we're holding an RwLock across an await point. Rather
than changing this to a tokio aware lock, let's remove the Arc and lock
altogether as this variable is only used from a single thread.

Thank you Rust for confirming this is safe :)


hodgesds (approved):
Nice!

PR #1344, opened 2025-02-13
scx_rusty: further aarch64 fixes

etsal:
Two further fixes for scx_rusty on aarch: 

1) 7f992e3 left a remaining reference to p->mempolicy in the call checking whether the field exists in the first place. This has been removed and replaced with BPF_CORE_READ().
2) Adjust arena size on aarch64 to account for 16KiB pages. The code currently assumes 4KiB pages when calculate how many of them fit into a mapping.


kkdwivedi:
IIRC pages on aarch64 can also be 64kb. It might be better to do 4GB/PAGE_SIZE? Then it does the right thing everywhere.

etsal:
> IIRC pages on aarch64 can also be 64kb. It might be better to do 4GB/PAGE_SIZE? Then it does the right thing everywhere.

Is there a way to get the PAGE_SIZE at compile time? AFAICT it's a boot time variable now and not available to BPF programs. I tried to use it to solve a similar problem and it was reported back as 0.

kkdwivedi:
I guess not. One option is to use libbpf to set max entries between open and load of the BPF object. That's some work to precisely size it to 4GB max.

The other (easier) one is specifying 1 << 16, that should be fine for any PAGE_SIZE from 4KB to 64KB, and gives at minimum 262MB on x86 (which I think should be enough).

etsal:
> I guess not. One option is to use libbpf to set max entries between open and load of the BPF object. That's some work to precisely size it to 4GB max.
> 
> The other (easier) one is specifying 1 << 16 everywhere, that should be fine for any PAGE_SIZE from 4KB to 64KB, and gives at minimum 262MB on x86 (which I think should be enough).
> 
> Or just change 1 << 18 to 1 << 16 for arm and leave it be to cover 64kb PAGE_SIZE.

Yeah I'll just adjust the arena size for aarch64 for now, we cannot do much else about it.

PR #1346, opened 2025-02-13
scx_wd40: add cpumask library

etsal:
Pull request #1293 adds support for arena-backed CPU masks to scx_rusty, but depends on yet-unmerged upstream fixes and features to function properly. This means we cannot merge yet without breaking downstream consumers. The wd40 scheduler is experimental by design and can be used to safely iterate on the code without risking breakage. Port the CPU mask library to wd40 for this purpose.


scheds/rust/scx_wd40/src/bpf/cpumask.bpf.c
```diff
@@ -0,0 +1,229 @@
+#include <scx/common.bpf.h>
```
hodgesds, line 1: I assume at some point we'd pull this out to a common library?

scheds/rust/scx_wd40/src/bpf/percpu.h
```diff
@@ -0,0 +1,163 @@
+#pragma once
+
+#include <scx/common.bpf.h>
+
+struct wd40_percpu_storage {
```
hodgesds, line 5: Wonder if it makes sense to throw a spinlock in here, if there's ever going to be concurrent updates at the same time of any of this data it might be useful.

hodgesds (approved):
Looks good overall

scheds/rust/scx_wd40/src/bpf/cpumask.bpf.c
etsal: Yeah, it's only part of wd40 because it's relatively unstable.

scheds/rust/scx_wd40/src/bpf/percpu.h
etsal: Good point, even though it's per-CPU data it may still be used by multiple threads if one of them sleeps during an operation. Ideally we would prevent writing code that can sleep while a reference to per-CPU storage is held.

etsal: Update to this, adding the spinlock is not necessary right now because none of the paths sleep. It's still an interface with sharp edges but ideally we'll rip it out soon, or roll it into the CPU mask library where threads are guaranteed to not sleep between referencing and using their per-cpu variables.

PR #1347, opened 2025-02-13
Scx central mem leak

devnexen:


scheds/c/scx_central.c
```diff
@@ -82,6 +82,7 @@ int main(int argc, char **argv)
 			break;
 		default:
 			fprintf(stderr, help_fmt, basename(argv[0]));
+			scx_central__destroy(skel);
```
htejun, line 85: The program is exiting. Memory can't leak past program exit. I don't think this is necessary.

devnexen: Ok. are you ok with the other change though ?

htejun: That doesn't do anything either, does it?

PR #1349, opened 2025-02-13
scx_layered: Fix pick_idle_cpu() sometimes returning a disallowed CPU

htejun:
Fix two issues which can lead to pick_idle_cpu() occasionally returning a disallowed CPU triggering ops error and ejecting the scheduler.


hodgesds (approved):
LGTM

PR #1354, opened 2025-02-14
kernel: Sync with kernel at ad3b301aa05a ("sched_ext: Provides a sysfs 'events' to expose core event counters")

htejun:
Sync both vmlinux.h's and tools/sched_ext.


hodgesds (approved):
I'm too lazy to check, but I'm guessing `CONFIG_NUMA` is set for all the vmlinux's :smile:

htejun:
I don't think arm64 has it. It refreshes the kernel config from the x86 one, so I don't know why it ends up without CONFIG_NUMA. Problem to solve some other time.

PR #1356, opened 2025-02-14
enum updates

htejun:
- Sync enum definitions with kernel.
- Add missing `scx_rq_flags` to autogenerated enums.
- Make scx_utils not panic when kernel is missing some enums.


hodgesds (approved):
LGTM

etsal (approved):

PR #1359, opened 2025-02-15
scx_rusty: turn off mempolicy setting for arm64 until we regen the headers

etsal:
Using BPF_CORE_READ() still fails for arm64 using our current vmlinux.h, because the macro still assumes the field exists. Temporarily turn off mempolicy setting for arm64 until we regenerate the arm64 header with NUMA support.


hodgesds (approved):
LGTM

Issue #1351, opened 2025-02-14
scxtop: Add process views


hodgesds:
Add a process view that allows for viewing scheduling metrics at an aggregated level (similar to other top like tools) as well as a process details view. The process details view should provide detailed process/thread level scheduling stats.

Some inspirations:

Tree like view from `htop`:

![Image](https://github.com/user-attachments/assets/35b34621-107a-4a5f-aec3-aca35052ad3b)

Utilization from `btop`:

![Image](https://github.com/user-attachments/assets/547a59f0-3f6b-4d63-8293-85c0d0d2c222)

hodgesds:
implemented in #2434

PR #1361, opened 2025-02-16
scxtop: propose to give the ability to focus on one process.

devnexen:
checking if scheduler A fits well with a specific application. In that case, it is probably fair to think that child processes created (e.g. web servers) should be included.


hodgesds:
At some point I'd like to add process views (#1351) so maybe if this could be done in a less static way it would be less refactoring in the future.

hodgesds (approved):
LGTM

PR #1366, opened 2025-02-17
scxtop: add experimental_long_tail_tracing

JakeHillion:
Adding an experimental feature to automatically start a perfetto trace when a
named function takes a specified long time to return. This can be used to get a
trace of an event when a group of long tail events happen together.

Prefacing this all with experimental as the implementation is a little funky.

Implementation notes:
- Starts the trace entirely from BPF space and relies on ordering of the
  ringbuffer to make sure userspace notices before the first events come
  through.
- Misses early softirq events because they require userspace to attach the
  programs.
- Starts the trace from `trace_tick_warmup` ticks to events aren't dropped.
- Extends the trace if another such event comes from, meaning this can generate
  an infinitely long trace and probably get scxtop OOM killed if you get the
  parameters wrong.

The concept here is that many traces can be recorded by leaving scxtop running,
but currently it doesn't seem to be able to get more than 2 without the TUI
crashing and leaving it in a terminal output that needs Ctrl+C'ing.

Test plan:
- Attached to a production binary and it hits when expected, gathering traces
  that look reasonable.

### DIFF
--- a/tools/scxtop/src/main.rs
+++ b/tools/scxtop/src/main.rs
@@ -3,11 +3,6 @@
 // This software may be used and distributed according to the terms of the
 // GNU General Public License version 2.
 
-use anyhow::Result;
-use clap::Parser;
-use libbpf_rs::skel::OpenSkel;
-use libbpf_rs::skel::SkelBuilder;
-use libbpf_rs::RingBufferBuilder;
 use scxtop::bpf_intf::*;
 use scxtop::bpf_skel::types::bpf_event;
 use scxtop::bpf_skel::*;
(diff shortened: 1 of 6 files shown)


tools/scxtop/src/main.rs
```diff
@@ -62,6 +65,25 @@ struct Args {
...
+    /// Symbol to automatically trace the long tail of.
+    #[arg(long)]
+    experimental_long_tail_tracing_symbol: Option<String>,
+    /// Binary to attach the uprobe and uretprobe to.
+    #[arg(long)]
+    experimental_long_tail_tracing_binary: Option<String>,
+    /// Minimum latency to trigger a trace. Defaults to 0ns which launches a
+    /// trace on every return, this probably isn't what you want!
+    #[arg(long, default_value_t = 100000000)]
+    experimental_long_tail_tracing_min_latency_ns: u64,
```
hodgesds, line 86: Should this error if the value is 0?

hodgesds (approved):
Seems pretty useful!

tools/scxtop/src/main.rs
JakeHillion: Ah, forgot I changed the default, will update the comment. I think tracing every time a uprobe is hit is fine with the warning, even if the implementation that times it anyway is a bit silly.

PR #1369, opened 2025-02-18
scxtop: __bpf_ringbuf_reserve() failure set to ENOMEM instead.

devnexen:
Think that ENOENT is not necessarily appropriate in this case. Seems more likely a potential error occurs when we run out of space in the ring buffer.


hodgesds (approved):
LGTM

PR #1370, opened 2025-02-18
ci: use for-next for regular ci jobs to enable testing of new kernel …

likewhatevs:
ci: use for-next for regular ci jobs to enable testing of new kernel functions (numa ones)


etsal (approved):

JakeHillion (commented):
Can you drop the `for-next-test.yml` please.

likewhatevs:
> Can you drop the for-next-test.yml please.

I can but realistically the moment for-next breaks we're gonna switch the main job to using the latest release and add for-next-test.yml back so that's just more chores.

Still drop for-next-test.yml @JakeHillion ?

JakeHillion:
> > Can you drop the for-next-test.yml please.
> 
> I can but realistically the moment for-next breaks we're gonna switch the main job to using the latest release and add for-next-test.yml back so that's just more chores.
> 
> Still drop for-next-test.yml @JakeHillion ?

I think that's definitely preferable. Fewer places to copy and paste everything is definitely a win, even if we do end up having to add it later.

PR #1372, opened 2025-02-18
scx_wd40: start simplifying

etsal:
Initial series of refactorings for scx_wd40. The two most important are:

- Domains now directly point to the CPU mask of their respective NUMA node instead of each having their own copy.
- Domain CPU masks are now directly in the dom_ctx struct, so the lb_domain is almost removed from the codebase.


JakeHillion (approved):
Neat.

PR #1375, opened 2025-02-18
scx_layered: change sysinfo version to 0.30

likewhatevs:
change sysinfo version to 0.30

this will make downstream consumption simpler in places with this version and less-flexible dep management than cargo.


htejun:
I'm late but why do we need this? We generally don't want to depend on multiple versions of the same crate in the repo, so does this mean that everyone should be using the old version?

PR #1376, opened 2025-02-19
scxtop: Make tokio runtime options configurable

hodgesds:
Make tokio runtime options configurable via the CLI. On machines with large number of cores (100+) the default number of tokio threads (`nproc`) is excessive for what `scxtop` requires.


tools/scxtop/src/main.rs
```diff
@@ -59,6 +59,9 @@ struct Args {
     /// Number of ticks for traces.
     #[arg(long, default_value_t = 5)]
     trace_ticks: usize,
+    /// Number of worker threads
+    #[arg(long, default_value_t = 4, value_parser = clap::value_parser!(u16).range(2..128))]
```
JakeHillion, line 63: Why do we upper limit it? Do we really care if someone requests 400 threads?

JakeHillion (approved):

PR #1379, opened 2025-02-19
scx_layered: fix gpu matcher bugs

likewhatevs:
scx_layered: fix gpu matcher bugs ~and add flag to enable affinitizing gpu tasks~

this fixes pid serialization across ffi, confirmed via bpftool that the gpu pid map has all GPU pids in it:
```
[{
        "key": 558103,
        "value": 0
    },{
        "key": 558106,
        "value": 0
    },{
...
```

~this also adds a flag, ` --affinitize-gpu-tasks`, which has the scheduler grab some topo info when it uses nvml and affine gpu processes to those CPUs nvml recommends as ideal.~

I think this tested well (it looked like there was enough load for the results to be valid),

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -40,7 +40,6 @@ use log::debug;
 use log::info;
 use log::trace;
 use log::warn;
-use nix::sched::{sched_setaffinity, CpuSet};
 use nvml_wrapper::Nvml;
 use scx_layered::*;
 use scx_stats::prelude::*;
@@ -92,8 +91,6 @@ const NR_LSTATS: usize = bpf_intf::layer_stat_id_NR_LSTATS as usize;
 const NR_LLC_LSTATS: usize = bpf_intf::llc_layer_stat_id_NR_LLC_LSTATS as usize;
 
 const NR_LAYER_MATCH_KINDS: usize = bpf_intf::layer_match_kind_NR_LAYER_MATCH_KINDS as usize;
-
-const CPUMASK_MAX_LONGS: usize = 8;
 
 lazy_static! {
     static ref USAGE_DECAY: f64 = 0.5f64.powf(1.0 / USAGE_HALF_LIFE_F64);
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/main.rs
```diff
@@ -1990,17 +2000,39 @@ impl<'a> Scheduler<'a> {
...
                     pids.insert(proc_info.pid);
+                    if self.opts.affinitize_gpu_tasks
+                        && !self.gpu_mon_data.gpu_to_affinity.is_empty()
+                    {
+                        let cpu_set = self.gpu_mon_data.gpu_to_affinity.get(&i);
+                        if let Some(cpu_set) = cpu_set {
+                            sched_setaffinity(
+                                nix::unistd::Pid::from_raw(proc_info.pid.try_into()?),
+                                cpu_set,
+                            )?;
```
JakeHillion, line 2033: Is this a good idea, especially in layered? As I understand it any task with affinity != a NUMA node will go into the lo_fallback_dsq, which should have the effect of deprioritising these tasks. These are likely to be NUMA aligned but I don't see any checks here, or a comment saying they will always be NUMA aligned (I don't think that's possible, especially with AMD's ignore NUMA BIOS feature). If they're NUMA aligned the layer they fall into needs to be Open, have `allow_node_aligned` set, and the machine can't saturate for risk of stalls.

I think some additional checks or warnings here would be a good idea, even if this is behind a flag.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1990,17 +2000,39 @@ impl<'a> Scheduler<'a> {
...
-
             // Iterate over all devices and gather running processes
             for i in 0..device_count {
                 let device = nvml.device_by_index(i)?;
+                // this feels expensive and static, so do it once.
+                if self.gpu_mon_data.gpu_to_affinity.is_empty() {
+                    let affinity = device.cpu_affinity(CPUMASK_MAX_LONGS)?;
+                    let mut cpu_set = CpuSet::new();
+                    for (x, &sub_cpumask) in affinity.iter().enumerate() {
+                        for cpu_index in 0..64 {
```
etsal, line 2012: Why * 32? Are we keeping the low 5 bits for something? Is it possible this is supposed to be << 32?

htejun:
It'd be better to do this through layers or implicit cpu allocations rather than going through explicit process affinity changes. In general, I don't think it's a good idea for system software, whether kernel or userspace, to be changing and possibly overriding user-visible configurations. This is one of the big problems with cpuset btw.

scheds/rust/scx_layered/src/main.rs
likewhatevs: > is supposed to be << 32

Yep, thanks.

likewhatevs:
> In general, I don't think it's a good idea for system software, whether kernel or userspace, to be changing and possibly overriding user-visible configurations. This is one of the big problems with cpuset btw.

That makes sense, will remove that part.

etsal (approved):

PR #1381, opened 2025-02-19
scx_utils: make cpumask example no_run

JakeHillion:
This example creates a Cpumask from an explicit string which required certain
CPUs to be present. Previously this succeeded on machines with 2/3 CPUs but not
4/5/6/7. This wasn't testing anything useful and reduced the number of machines
that can cleanly run these tests, so stop running it. Use `no_run` instead of
`ignore` so the snippet still gets compiled.

Drive-by: remove unnecessary String and use a str literal instead.


hodgesds (approved):
LGTM

PR #1386, opened 2025-02-20
scx_utils: use clsuter_id to detect if big.LITTLE CPUs

arighi:
Using CPU frequency alone to identify hybrid cores can lead to incorrect classifications. Some processors (i.e., some NVIDIA Grace CPUs) may have cores operating at different frequency ranges, due to power management or other design factors, but that doesn't necessarily mean they are hybrid cores, in the sense of having different microarchitectural designs (e.g., big.LITTLE or Intel's P/E cores).

To achieve more accurate big.LITTLE core detection, rely both on the cluster_id and CPU frequency information.

Moreover, add the "cluster_id" property to struct Core and struct Cpu.

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -125,6 +125,7 @@ pub struct Cpu {
     pub llc_id: usize,
     pub node_id: usize,
     pub package_id: usize,
+    pub cluster_id: usize,
 }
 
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
Makes sense.

multics69:
Looks good. Thanks!

PR #1387, opened 2025-02-20
scx_loader: removing only usage of nix's crate.

devnexen:
everywhere else use directly native calls, so let's do there too.


rust/scx_loader/Cargo.toml
```diff
@@ -12,7 +12,7 @@ clap = { version = "4.1", features = ["derive", "env", "unicode", "wrap_help"] }
 colored = "2"
 ctrlc = { version = "3.1", features = ["termination"] }
 log = "0.4.17"
-nix = { features = ["process", "signal"], default-features = false, version = "0.29" }
+libc = "0.2.137"
```
hodgesds, line 15: In the [lock file](https://github.com/sched-ext/scx/blob/main/Cargo.lock#L1370) it seems to be a newer version.

rust/scx_loader/src/main.rs
```diff
@@ -467,10 +467,11 @@ async fn start_scheduler(
                         log::debug!("Received cancellation signal");
                         // Send SIGINT
                         if let Some(child_id) = child.id() {
-                            nix::sys::signal::kill(
-                                nix::unistd::Pid::from_raw(child_id as i32),
-                                nix::sys::signal::SIGINT,
-                            ).context("Failed to send termination signal to the child")?;
+                            let ret = unsafe { libc::kill(child_id as i32, libc::SIGINT) };
```
vnepogodin, line 470: Hmm looks less convenient, as `nix` provides good "safe" wrapper around `libc`.

IMHO I'd want more usage of "safe" code

devnexen: hmmm yes and no, ultimately nix calls libc::kill under the hood, there is a tiny safety net for the arguments types. nix strength relies more on interface unification across platforms IMHO but if scx folks wants me to close it I won t argue :)

likewhatevs: Not huge on safe/unsafe until it becomes problematic, but wrt/ maintainability I'm all for using crates to make things being done simpler.

Consistency (doing the same thing in more places) is good if not great, but I think the overall direction should be "use more crates and make things simpler" not "do things in harder manners to increase consistency".

devnexen: Fair point, note I m not against nix personally, I contribute to it. Anyhow, let s close it then :)

PR #1388, opened 2025-02-20
scx_bpfland: avoid creating redundant cache domains

arighi:
Skip creating cache domains when they don't provide any benefit:
- when a cache domain contains only a single CPU,
- when a cache domain contains all available CPUs,
- when all CPUs in the cache domain belong to the same SMT core.


hodgesds (approved):
Nice!

multics69:
Looks good!

PR #1389, opened 2025-02-20
scx_bpfland: reduce task migrations

arighi:
Implement a more conservative task migration strategy:
 - consider cache locality for proactive CPU wake up in ops.enqueue(),
 - perform local direct dispatch on non-wakeup paths when the local DSQ has just one or zero tasks waiting.

This change improves cache locality while still maintaining a good level of work conservation.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -227,6 +227,24 @@ static bool is_queued(const struct task_struct *p)
 }
 
 /*
+ * Return true if @cpu is a full-idle SMT core, false otherwise.
+ */
+static bool is_idle_smt(s32 cpu)
+{
+	const struct cpumask *idle_smtmask;
+	bool is_idle;
+
+	if (!smt_enabled)
+		return false;
+
+	idle_smtmask = scx_bpf_get_idle_smtmask();
+	is_idle = bpf_cpumask_test_cpu(cpu, idle_smtmask);
+	scx_bpf_put_cpumask(idle_smtmask);
+
+	return is_idle;
+}
+
+/*
  * Allocate/re-allocate a new cpumask.
  */
 static int calloc_cpumask(struct bpf_cpumask **p_cpumask)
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -660,21 +659,45 @@ s32 BPF_STRUCT_OPS(bpfland_select_cpu, struct task_struct *p,
...
 	 * Look for any idle CPU usable by the task that can immediately
 	 * execute the task.
 	 *
 	 * Note that we do not want to mark the CPU as busy, since we don't
 	 * know at this stage if we will actually dispatch any task on it.
 	 */
-	idle_cpumask = scx_bpf_get_idle_cpumask();
-	cpu = bpf_cpumask_any_and_distribute(p->cpus_ptr, idle_cpumask);
-	scx_bpf_put_cpumask(idle_cpumask);
+	mask = cast_mask(tctx->l2_cpumask);
```
hodgesds, line 674: The one thing I've thought of for the the selection is it is better to do something like this in the SMT case:
L2 -> check idle SMT -> check idle CPU
L3 -> check idle SMT -> check idle CPU

On partially loaded SMT systems I wonder if it's better to try to get better SMT isolation vs L2 locality. I haven't done any testing, but curious if you've thought of that. The downside of checking if there's an idle SMT at the L2/L3 level first is a bunch of extra code though.

hodgesds (approved):
LGTM

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: That is probably a good idea, I'll do some tests!

arighi:
I've done more tests trying to disrupt a CPU hog, running another task on the same CPU or the SMT sibling, forcing the hog to bounce on another CPU, and the SMT-aware proactive kick seems to help preventing tasks to stay on the same SMT core.

So, I've updated this PR adding more SMT-aware logic. Thanks or the suggestion @hodgesds !

multics69 (approved):
Looks good!

PR #1391, opened 2025-02-20
scxtop: Pass actions by reference

hodgesds:
Cleanup some inefficient code taht would send copies of Actions.


tools/scxtop/src/app.rs
```diff
@@ -2073,10 +2073,10 @@ impl<'a> App<'a> {
             Action::PageDown => self.on_pg_down(),
             Action::Enter => self.on_enter(),
             Action::SetState(state) => {
-                if state == self.state {
+                if *state == self.state {
```
hodgesds, line 2076: For now it's just pushing the clones further down, but this can be cleaned up later.

tools/scxtop/src/app.rs
```diff
@@ -2073,10 +2073,10 @@ impl<'a> App<'a> {
...
                     self.set_state(self.prev_state.clone());
                 } else {
-                    self.set_state(state);
+                    self.set_state(state.clone());
```
JakeHillion, line 2079: nit: if you derive `Copy` on `AppState` this becomes a `*`, which is good because it shows we don't need to clean it up in future.

JakeHillion (approved):

PR #1392, opened 2025-02-20
scx_wd40: using nix instead of libc.

devnexen:
just follow-up on #1387 discussions.


likewhatevs:
I like the idea of using high level things to make code simpler and I really appreciate this change. I think I added a dep on nix on a random PR a few days back. I ended up removing that for unrelated reasons.

All that being said, I would probably end up undoing this change for layered in particular due to downstream pain it will likely cause, so this change needs to provide utility beyond making the code more consistent and simper to outweigh that.

If nix worked w/o any features and was on a particular (and likely, older) version that I don't have on-hand atm., that might be fine.

likewhatevs:
also -- I don't think this constraint applies to other schedulers to the same degree as it does to layered (idk), so maybe see if others don't have such constraints etc.?

etsal:
I don't have strong feelings for this change either way, but about this in particular:

> doing only 1 scheduler on purpose until that s the agreed direction we should aim for.

In general I don't think we should be aiming to enforce code conventions between schedulers, because they are not really a shared codebase even though they share a build system and dependencies.

etsal (approved):

devnexen:
> In general I don't think we should be aiming to enforce code conventions between schedulers, because they are not really a single codebase even though they share a build system and dependencies. Enforcing common conventions between them wouldn't give us anything extra, expecially since most of them are actively evolving in terms of design.

ok will remember that :)

PR #1394, opened 2025-02-21
scxtop: Add config file

hodgesds:
Add option to load default values from a config file defined following XDG base directory specification. This will allow for custom keymaps to be defined in config files. The handling between the clap CLI args and the config file is a little awkward, but that can be cleaned up in the future.

After running `scxtop` the default config file will be created:
```
$ cat $HOME/.config/scxtop/scxtop.toml
tick_rate_ms = 250
debug = false
exclude_bpf = false
stats_socket_path = "/var/run/scx/root/stats"
trace_file_prefix = "scxtop_trace"
trace_ticks = 5
worker_threads = 4
trace_tick_warmup = 3
```

### DIFF
--- a/tools/scxtop/src/config.rs
+++ b/tools/scxtop/src/config.rs
@@ -1,0 +1,194 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+//
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use crate::cli::Cli;
+use crate::AppTheme;
+use crate::STATS_SOCKET_PATH;
+use crate::TRACE_FILE_PREFIX;
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use std::fs;
+use std::path::{Path, PathBuf};
+use xdg;
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct Config {
+    /// TUI theme.
+    theme: Option<AppTheme>,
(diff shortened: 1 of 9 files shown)


likewhatevs (approved):

likewhatevs:
this is cool but if cfg options are gonna get non-self-explanitory, might make sense to do something like w/ zellij does. 

it's cfg files are nice:
https://github.com/zellij-org/zellij/blob/6f8e9cc6125cb31de710ef756b0ab00cd6b84f07/example/config.kdl#L5

tools/scxtop/src/config.rs
```diff
@@ -0,0 +1,94 @@
...
+            tick_rate_ms: 250,
+            debug: false,
+            exclude_bpf: false,
+            stats_socket_path: STATS_SOCKET_PATH.to_string(),
+            trace_file_prefix: TRACE_FILE_PREFIX.to_string(),
+            trace_ticks: 5,
+            worker_threads: 4,
+            trace_tick_warmup: 3,
+        }
+    }
```
JakeHillion, line 54: Could we add these defaults using the `serde::default` tags like in layered? Then if the config isn't present we could parse an empty string and get the default config, else fields would be filled in with defaults if they aren't specified. What do you think? I prefer the idea of being able to encode only what I want to specify in the config file and leaving the rest to the app defaults that may change over time.

hodgesds: That sounds good!

PR #1398, opened 2025-02-21
scx_bpfland: Always re-enqueue tasks that are running on a busy SMT core

arighi:
In ops.dispatch() we normally extend the time slice for tasks running on primary domain CPUs that have no contention, allowing them to continue their execution without the overhead of going through ops.enqueue().

However, the CPU might be part of an SMT core with another task running on the sibling CPU. In this case, migrating the task to another CPU, preferably a fully idle SMT core, can lead to better performance.

Therefore, in such cases, rather than keeping the task running on its current CPU, direct it through the standard ops.enqueue() path to enable potential migration.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -131,6 +131,7 @@ struct cpu_ctx {
 	u64 tot_runtime;
 	u64 prev_runtime;
 	u64 last_running;
+	struct bpf_cpumask __kptr *smt_cpumask;
 	struct bpf_cpumask __kptr *l2_cpumask;
 	struct bpf_cpumask __kptr *l3_cpumask;
 };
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -851,10 +852,45 @@ void BPF_STRUCT_OPS(bpfland_enqueue, struct task_struct *p, u64 enq_flags)
...
+	/* Do not keep running if the task doesn't need to run */
+	if (!is_queued(p))
+		return false;
+
+	/* Do not keep running if the CPU is not in the primary domain */
+	if (!primary || !bpf_cpumask_test_cpu(cpu, primary))
+		return false;
 
+	/*
+	 * Keep running only if the task is on a full-idle SMT core (or SMT
```
hodgesds, line 871: I think this gets interesting as the machine becomes more saturated where there are no idle SMTs left. For low saturation this is probably the right thing to do.

hodgesds (approved):

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: @hodgesds precisely! I have another patch that I'm still testing to skip searching for idle CPUs when the system is busy :)

multics69:
Nice! Looks good to me.

PR #1399, opened 2025-02-21
scx_wd40: simplify BPF component

etsal:
Introduce 2 refactoring steps to simplify the code:

- Split arena allocation and kernel-side initialization into two steps to remove the intermediate masks we are currently using to communicate between userspace and BPF.
- Move the tuner completely in userspace by directly updating arena memory from the Rust module

### DIFF
--- a/scheds/rust/scx_wd40/src/bpf/lb_domain.bpf.c
+++ b/scheds/rust/scx_wd40/src/bpf/lb_domain.bpf.c
@@ -21,8 +21,6 @@
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_tracing.h>
 
-const volatile u64 dom_cpumasks[MAX_DOMS][MAX_CPUS / 64];
-
 struct lock_wrapper {
 	struct bpf_spin_lock lock;
 };
@@ -42,7 +40,7 @@ struct {
 	__uint(map_flags, 0);
 } dom_dcycle_locks SEC(".maps");
 
-scx_bitmap_t node_data[MAX_NUMA_NODES];
+volatile scx_bitmap_t node_data[MAX_NUMA_NODES];
 
 struct {
 	__uint(type, BPF_MAP_TYPE_ARRAY);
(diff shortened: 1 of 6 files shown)


scheds/rust/scx_wd40/src/bpf/main.bpf.c
```diff
@@ -902,12 +856,15 @@ static s32 initialize_cpu(s32 cpu)
 		}
 	}
 
+	scx_bpf_error("dom%d not found for CPU %d", nr_doms, cpu);
+
 	return -ENOENT;
 }
 
-s32 BPF_STRUCT_OPS_SLEEPABLE(wd40_init)
+SEC("syscall")
```
hodgesds, line 864: Nice!

scheds/rust/scx_wd40/src/bpf/main.bpf.c
```diff
@@ -941,21 +898,42 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(wd40_init)
...
 			return ret;
 	}
+
+	bpf_for(i, 0, nr_doms) {
+		ret = alloc_dom(i);
+		if (ret)
+			return ret;
+	}
+
+	return (0);
```
hodgesds, line 920: nit: why the extra ()?

hodgesds (approved):

scheds/rust/scx_wd40/src/bpf/main.bpf.c
etsal: [Muscle memory](https://man.freebsd.org/cgi/man.cgi?style(9)) :P I'll fix it for uniformity.

scheds/rust/scx_wd40/src/bpf/lb_domain.bpf.c
```diff
@@ -338,66 +336,49 @@ int dom_xfer_task(struct task_struct *p __arg_trusted, u32 new_dom_id, u64 now)
...
-	cast_user(dom_ctxs[dom_id]);
+
+	return 0;
+}
+
+__hidden
+s32 create_dom(u32 dom_id)
+{
+	dom_ptr domc;
+	u32 node_id = 0;
```
arighi, line 361: nit: we don't have to initialize node_id to 0

scheds/rust/scx_wd40/src/tuner.rs
```diff
@@ -138,8 +138,15 @@ impl Tuner {
         avg_util /= self.dom_group.weight() as f64;
         self.fully_utilized = avg_util >= 0.99999;
 
+        let write_to_bpf = |target: &mut [u64; 64], mask: &Cpumask| {
```
arighi, line 141: I'm wondering if we should make this block unsafe for better safety encapsulation and maybe add a `SAFETY:` comment explaining why it's safe.

arighi (approved):
I left a couple of comments, but overall LGTM!

scheds/rust/scx_wd40/src/tuner.rs
etsal: I agree, that would reduce the number of unsafe blocks around the code by moving the unsafe task to this lambda. I also think this lambda should be a proper function for the CPU mask because I've found variations of this lambda in other places in the code, e.g.,  when initializing the scheduler.

PR #1401, opened 2025-02-22
scx_bpfland: Skip idle CPU selection when system is fully busy

arighi:
When the system is fully busy, immediately give up searching for an idle CPU and just dispatch tasks on the first CPU available.

Moreover, do not attempt to migrate a task to a full-idle SMT core if all the SMT cores are busy.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -413,21 +413,31 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
 	bool is_prev_llc_affine = false;
 	s32 cpu;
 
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -429,6 +429,14 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
 	idle_smtmask = scx_bpf_get_idle_smtmask();
 	idle_cpumask = scx_bpf_get_idle_cpumask();
 
+	/*
```
hodgesds, line 432: Might be able to save a tctx lookup if this is placed before line 418. This is a good pattern!

hodgesds (approved):
LGTM

PR #1405, opened 2025-02-23
scx_bpfland: fix non-SMT CPU busy state detection

arighi:
Rename is_full_smt() to is_fully_idle() and return true when SMT is disabled, to represent the fact that the CPU doesn't have any busy shared resources.

This solves an issue in non-SMT cases where CPUs were incorrectly considered always busy.

Fixes: a480d880 ("scx_bpfland: reduce task migrations")


hodgesds (approved):
Makes sense!

multics69:
Looks good to me.

PR #1408, opened 2025-02-24
scx_lavd: Use tracepoints for futex for reliable tracing

multics69:
Using ftrace is handy. However, it could be unreliable across kernel versions and kernel configs because ftraced-functions can be renamed or inlined. Thus, for reliable tracing, replace ftrace to tracepoint. With tracepoint, there is no easy/efficient way to get arguments and return values, so drop tracking uaddr was used to distinguish indivisual futexes.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/lock.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/lock.bpf.c
@@ -7,37 +7,52 @@
 /*
  * To be included to the main.bpf.c
  */
-
-static void inc_futex_boost(u32 *uaddr)
+static void __inc_futex_boost(struct cpu_ctx *cpuc)
 {
 	struct task_struct *p = bpf_get_current_task_btf();
 	struct task_ctx *taskc = try_get_task_ctx(p);
-	struct cpu_ctx *cpuc = get_cpu_ctx();
 
-	if (taskc && cpuc && (taskc->futex_uaddr != uaddr)) {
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_lavd/src/bpf/lock.bpf.c
```diff
@@ -7,46 +7,60 @@
...
 	struct task_struct *p = bpf_get_current_task_btf();
 	struct task_ctx *taskc = try_get_task_ctx(p);
-	struct cpu_ctx *cpuc = get_cpu_ctx();
 
-	if (p && taskc && cpuc && taskc->futex_boost > 0) {
-		taskc->futex_boost--;
-		taskc->futex_uaddr = NULL;
-		cpuc->lock_holder = is_lock_holder(taskc);
+	if (taskc && taskc->futex_boost) {
+		if (!cpuc)
```
hodgesds, line 35: nit: you might be able to push this lookup into `dec_futex_boost` and that keeps this function a little cleaner where it can return early if `!cpuc`.

hodgesds (approved):
LGTM

multics69:
> LGTM, I think this could also be interesting on the wakeup path as well.

Currently, LAVD boosts a lock holder at the enqueue path.

https://github.com/sched-ext/scx/blob/972c65b6255bd60033e7e7156ffc92e8550547a3/scheds/rust/scx_lavd/src/bpf/main.bpf.c#L297

PR #1409, opened 2025-02-24
scx_p2dq: Handle invalid load sum calculations

hodgesds:
When calculating load sums if the load is zero then reset the load count. Handle edge case when resetting load to only update time slices if autoslice is enabled.


multics69 (approved):
Makes sense to me.

PR #1414, opened 2025-02-24
scx_utils: add helper for updating BPF masks from userspace

etsal:
As @arighi mentioned in #1399, it would be nice to have a more structured way to update BPF masks from userspace. Ideally, the userspace CPU mask type would have a helper for this operation. However, each scheduler directly or indirectly defines its own CPU mask types, so such a common helper would have to work with all of them.

Introduce a Cpumask method for setting BPF CPU masks from userspace and use it from scx_wd40 as an example. The helper itself is unsafe as it takes a raw pointer to the BPF proram, casts it into an array, and updates it with the contents of the Cpumask. Since the function is unsafe, it can handle any scheduler-specific types by turning them into [u64] to make it compatible with Cpumask's own representation.

On the WD40 side, add a convenience helper for passing the scheduler-specific scx_bitmap type into the method.


JakeHillion (commented):
Thinking about this a little, would it be better to have a type like `BitmapSlice` that is unsafe to create with something like `::new<T>(bpfptr: mut T*)` but that has safe read/write methods? The unsafe bit here should be describing the pointer, and only if that's wrong will the other methods not be. Not sure if I'm missing something though!

rust/scx_utils/src/cpumask.rs
```diff
@@ -266,6 +266,26 @@ impl Cpumask {
         }
     }
 
+    /// Write out a CPU mask to a raw memory pointer. We normally use this as part of updating
+    /// the CPU masks on the BPF side.
+    pub unsafe fn write_to_ptr(&self, bpfptr: *mut u64, len: usize) -> Result<()> {
+        let cpumask_slice = self.as_raw_slice();
+        if len < cpumask_slice.len() {
```
arighi, line 273: I would even do len != cpumask_slice.len(), is there any reasonable scenario to pass a bigger buffer?

arighi (approved):
Nice! I left a comment, but overall looks good. I need to test it out. Thanks for working on this!

etsal:
> Thinking about this a little, would it be better to have a type like `BitmapSlice` that is unsafe to create with something like `::new<T>(bpfptr: mut T*)` but that has safe read/write methods? The unsafe bit here should be describing the pointer, and only if that's wrong will the other methods not be. Not sure if I'm missing something though!

Mainly to keep it very explicit that this is an unsafe operation. Adding an extra type that will also never be used apart from this call would add extra boilerplate for not much gain imo.

JakeHillion (approved):

multics69 (approved):
Thank you @etsal for the work! cpumask handling was a real pain. Overall looks good to me.

PR #1415, opened 2025-02-24
scxtop: make event_data faster

JakeHillion:
EventData has a lot of string copies and a `Vec::remove(0)`.

Changes made:
- Replace the `std::collections::BTreeMap` with a `hashbrown::HashMap`. This is
  very similar to `std::collections::HashMap` now but has the benefit of not
  needing a nightly compiler to use the `raw_entry_mut` API.
- Use the `event_data_mut` API to avoid cloning the `String` in the common
  (hit) case.
- Proliferate that String change everywhere.
- Change the `Vec` to a `std::collections::VecDeque` to make popping from the
  front cheap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2516,6 +2516,7 @@ dependencies = [
  "directories",
  "futures",
  "glob",
+ "hashbrown 0.15.2",
  "json5",
  "lazy_static",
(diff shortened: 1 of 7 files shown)


hodgesds (approved):
Nice cleanup!

PR #1417, opened 2025-02-25
build: Add release-fast build mode

hodgesds:
Add `release-fast` build mode for fast local builds for testing.

`release` build:
image

`release-quick` build:
image

Build timings generated with:
```
cargo build --timings --profile release
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,3 +18,11 @@ resolver = "2"
 
 [profile.release]
 lto = true
+
+[profile.release-fast]
+inherits = "release"
+target-cpu = "native"
+lto = false
+incremental = true
+opt-level = 0
+debuginfo = 0


likewhatevs (approved):

Cargo.toml
```diff
@@ -18,3 +18,11 @@ resolver = "2"
 
 [profile.release]
 lto = true
+
+[profile.release-quick]
```
arighi, line 22: Is it release-fast or release quick?

arighi (approved):
Looks useful! Maybe clarify the profile name in the commit message and report the cargo build command.

Cargo.toml
hodgesds: Whoops, I couldn't decide on a name so I guess I used both. I think fast is better (less keys to type).

PR #1420, opened 2025-02-25
scxtop: using TopologyLevel::Internal nearest gpu discovery mode.

devnexen:
For models that support multi gpu cards, NVML_TOPOLOGY_LEVEL_INTERNAL is faster as it focus only on looking up in this context.


hodgesds:
I think the title of the PR should include scx_utils and not scxtop?

devnexen:
hehe true..

multics69 (approved):
Looks good to me. Thanks!

hodgesds (approved):

PR #1425, opened 2025-02-26
scx_layered: revamp gpu support

likewhatevs:
redo gpu support for layered

* remove the hacks to avoid using nvml and replace that with a kprobe on nvidia-open.
* have the kprobe be conditionally loaded such that layered loads/runs with and without the symbol nvidia_open present.
* remove no_prealloc from the gpu pid map and make the gpu pid map lru.

this seems to work better than the prior solution:

radeon machine:
Pasted image

nvidia machine:
Pasted image (2)

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -178,22 +178,35 @@ static bool cpuc_in_layer(struct cpu_ctx *cpuc, struct layer *layer)
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	// u32 comes from nvml
 	__type(key, u32);
 	__type(value, u32);
 	__uint(max_entries, MAX_GPU_PIDS);
 	__uint(map_flags, BPF_F_NO_PREALLOC);
-} all_gpu_pid SEC(".maps");
-
+} gpu_tgid SEC(".maps");
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	// u32 comes from nvml
 	__type(key, u32);
 	__type(value, u32);
 	__uint(max_entries, MAX_GPU_PIDS);
 	__uint(map_flags, BPF_F_NO_PREALLOC);
-} cur_gpu_pid SEC(".maps");
+} gpu_tid SEC(".maps");
+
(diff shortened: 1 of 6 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1831,37 +1844,33 @@ static __noinline bool match_one(struct layer_match *match,
...
-				return match->using_gpu;
-			
-			pid = (u32) p->tgid;
-			gpu_pid = bpf_map_lookup_elem(&cur_gpu_pid, &pid);
+				scx_bpf_error("UsedGpuTid requires --enable_gpu_support");
+
+			tid = p->pid;
 
-			if (gpu_pid && *gpu_pid == 0)
+			if (bpf_map_lookup_elem(&gpu_tgid, &tid) == 0)
```
etsal, line 1856: Shouldn't this be searching gpu_tid?

likewhatevs: Yes, thank you!!!! 

I wasn't sure why that was getting 0 matches but didn't know what to expect w/ that one (unlike gpu pid which is clearer wrt/ what working right looks like).

etsal: Sg! Just making sure, looks great overall.

etsal (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -178,22 +178,35 @@ static bool cpuc_in_layer(struct cpu_ctx *cpuc, struct layer *layer)
...
 	__uint(type, BPF_MAP_TYPE_HASH);
-	// u32 comes from nvml
 	__type(key, u32);
 	__type(value, u32);
 	__uint(max_entries, MAX_GPU_PIDS);
 	__uint(map_flags, BPF_F_NO_PREALLOC);
-} cur_gpu_pid SEC(".maps");
+} gpu_tid SEC(".maps");
+
+SEC("?kprobe/nvidia_open")
```
arighi, line 195: This is really cool! TIL! Thanks for figuring this out @likewhatevs!

PR #1426, opened 2025-02-26
scx_layered: have allow_node_aligned use hi fallback dsqs

likewhatevs:
was experimenting and think this made an improvement. not sure if it is a bad idea nor how it would handle affinitized tasks (not certain if any of the test workloads I've been working with have those).


htejun:
It would be better to ensure that the threads that are important don't end up in the fallback DSQs in the first place. Hi or lo, you lose a lot of control by bouncing them there.

PR #1427, opened 2025-02-26
scxtop: add config merging interface

JakeHillion:
The `Config::merge_tui_args` method is quite complicated and needs modifying
each time. It leaves the `Config` in a mostly-Some state, meaning combining
more than one `TuiArgs` for whatever reason wouldn't work.

Generalise this into a `Config::merge` method which prefers the left-most
value. This is done with an extension of the `Option::or` method for all fields
in the config struct, meaning the left-most set value will win. We can then
replace `Config::merge_tui_args` with a combination of `From<TuiArgs> for Config`
and `Config::merge`. This turns the quite complicated function into two simple
ones.

Test plan:
- Wrote a test to confirm fields get merged as expected.

### DIFF
--- a/tools/scxtop/src/cli.rs
+++ b/tools/scxtop/src/cli.rs
@@ -25,7 +25,7 @@ pub struct Cli {
     pub tui: TuiArgs,
 }
 
-#[derive(Parser, Debug)]
+#[derive(Clone, Parser, Debug)]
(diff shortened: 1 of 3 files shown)


hodgesds (approved):
Nice, this is much more idiomatic!

PR #1431, opened 2025-02-26
scxtop: Set correct max on LLC sparkline view

hodgesds:
Set the max on the LLC sparkline view to be the maximum aggregated across all LLCs. This makes the TUI far more easier to intrepret across LLCs.

image


JakeHillion (approved):
Very nice.

PR #1436, opened 2025-02-27
scxtop: adding gpu_mem_total tracepoint.

devnexen:
when for a given GPU, its memory activity fluctuates, this event is triggered. Could be on a process basis or globally.


hodgesds (approved):
LGTM, sorry if tracing is broken on main. Currently in the middle of a refactor to make a separate `scxtop trace` subcommand.

PR #1437, opened 2025-02-27
scx_layered: expand kprobes used for gpu support

likewhatevs:
add kprobes for nvidia_mmap, nvidia_poll, nvidia_open.

these in combination with IsGroupLeader seem to identify pids using gpus reasonably well (including when layered is started after the PIDs using GPUs are).

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -192,7 +192,6 @@ struct {
 	__uint(map_flags, BPF_F_NO_PREALLOC);
 } gpu_tid SEC(".maps");
 
-SEC("?kprobe/nvidia_open")
 int save_gpu_tgid_pid() {
 	if (!enable_gpu_support)
 		return 0;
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -208,6 +207,21 @@ int save_gpu_tgid_pid() {
 	return 0;
 }
 
+SEC("?kprobe/nvidia_open")
+int kprobe_nvidia_open() {
+	return save_gpu_tgid_pid();
+}
+
+SEC("?kprobe/nvidia_poll")
```
hodgesds, line 215: Is this probe used frequently? If it's on every GPU request it might have some overhead.

likewhatevs: I can't tell w/ my current test setup (too noisy) but mmap alone works fine.

likewhatevs: I messed up my test earlier -- no gpu support w/o poll it looks like, adding it back in.

likewhatevs:
Compat's ksym_exists doesn't work for this (just checked). 

Gonna try fixing that to get the helper from this PR moved into compat, but not sure how to make that work other than grepping kallsyms like this does atm and not sure why compat doesn't just do that because IIUC it would always work.

etsal (approved):
LGTM, if tracking all poll requests is too much we can try going for less frequent events inside the syscall

likewhatevs:
> LGTM, if tracking all poll requests is too much we can try going for less frequent events inside the syscall

Yeah, this is gonna be a balancing act. 

ATM, this seems like a reasonable place to do it (i.e. the performance benefit we've seen is outweighing cost incurred). 

If we have perf issues in the future and have to kprobe component syscalls, that might be when we start having to worry more about symbol changes across upgrades (i.e. mmap, poll and open don't have to be stable, but I think they would be).

PR #1438, opened 2025-02-28
scxtop: Conditionally attach GPU probes

hodgesds:
Only attempt to attach GPU probes if the symbol exists. Fixes failure to launch on systems without GPU tracepoints.

Happy trace:

### DIFF
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -677,7 +677,7 @@ int BPF_PROG(on_ipi_send_cpu, u32 cpu, void *callsite, void *callback)
 	return 0;
 }
 
-SEC("tp_bpf/gpu_memory_total")
+SEC("?tp_btf/gpu_mem_total")
 int BPF_PROG(on_gpu_memory_total, u32 gpu, u32 pid, u64 size)
 {
 	struct bpf_event *event;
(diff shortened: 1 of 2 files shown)


devnexen (approved):

tools/scxtop/src/bpf/main.bpf.c
```diff
@@ -677,11 +677,10 @@ int BPF_PROG(on_ipi_send_cpu, u32 cpu, void *callsite, void *callback)
 	return 0;
 }
 
-SEC("tp_bpf/gpu_memory_total")
+SEC("?tp_bpf/gpu_memory_total")
```
devnexen, line 680: note: I think there is a typo (my bad), should not it be tp_b**t**f/gpu_memory_total ?

hodgesds: Should be good now, thanks for catching!

hodgesds: @devnexen I think the tracepoint is `gpu_mem_total` from the [definition](https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpu/trace/trace_gpu_mem.c#n13).

devnexen: correct

JakeHillion (approved):

PR #1439, opened 2025-02-28
scxtop: adding cpu hotplug (enter) event tracepoint.

devnexen:
For a given cpu, we capture its activity, e.g. online/offline, some states more in between like about to be teardown, before full activation.


hodgesds (approved):
Good stuff!

PR #1441, opened 2025-02-28
scxtop: stop traces with bpf timer

### LINKED ISSUE #1317: scxtop: Hang during perfetto trace capture
I have been getting hangs when trying to capture perfetto traces using scxtop. The progress bar gets stuck at around40%, and during the hang the scxtop process is permanently stuck at exactly 200% CPU util.

JakeHillion:
Currently scxtop schedules the end of a trace from userspace. With an earlier
setup this end of a trace would never be reached if userspace couldn't keep up
with the bpf events. In the current setup it should be reached, but still
relies on userspace getting scheduled to slow the BPF side back down.

Set this instead with a bpf timer. The bpf timer stops the flood of events by
itself and informs userspace that the event flow has finished, keeping all of
the timestamps in BPF ktime_get_ns instead of ticks. This allows generally more
flexibility too.

The new error condition here is the BPF side not sending the event informing
userspace that the trace has finished, causing it to get stuck in the tracing
state forever. This can be fixed by setting a flag on the BPF side that we need
to terminate if we are unable to get the event into the ring and sending it
later, but this PR is already quite large.

Closes #1317

Test plan:
- Ran several traces on several different sizes of machine. Other than the
  non-termination mentioned, this seems good.


JakeHillion:
I took a proper look at making sure the event to terminate the trace always gets through (restarting the timer if it fails) and it turns out this is pretty complicated. Making sure the event always gets through at some time is easy enough, but the userspace changes to forget events past when the trace really stopped (to avoid the sample rate breaking the perfetto trace) are more invasive. Would rather merge this with the bug and fixup separately if that's okay.

tools/scxtop/src/app.rs
```diff
@@ -2130,21 +2123,82 @@ impl<'a> App<'a> {
         Ok(())
     }
 
+    /// Request the BPF side start a trace.
+    fn request_start_trace(&mut self) -> Result<()> {
+        self.skel.maps.data_data.trace_duration_ns = self.config.trace_duration_ns();
```
hodgesds, line 2128: nit: should these be after the check on the `AppState`?

hodgesds (approved):
LGTM

PR #1442, opened 2025-03-01
scx_wd40: optimisations/simplifications

devnexen:
- mostly remove unneeded clone() calls.
- remove unneeded references.
- apply lifetime simplification like other schedulers.


etsal:
Looks great, thank you for taking the time.

etsal (approved):

etsal:
Also, just to better understand the patch: The clone() calls were unnecessary because dereferencing a reference to a base type returns it by value, and similarly passing a base type to a function is the same as passing a reference to it for the sae reason/

PR #1443, opened 2025-03-01
scxtop: Add docs for configuration

hodgesds:
Update the `scxtop` README to include documentation about the config file.


JakeHillion (approved):
Do you think it's worth putting this in the help docs like layered instead? I wonder whether that's closer to where users might need it. Your call.

hodgesds:
> Do you think it's worth putting this in the help docs like layered instead? I wonder whether that's closer to where users might need it. Your call.

I think both is a good option 😄

PR #1444, opened 2025-03-01
scx_flash: drop user lock priority boosting

arighi:
Drop the logic that tries to track user-space futexes and extends task time slices when tasks is holding a user-space lock.

This improves scheduler portability and maintainability across kernel versions. For the scheduler's primary use case (audio/multimedia), the user lock priority boosting provided minimal benefits and sometimes negatively impacted performance predictability by over-prioritizing lock-heavy tasks.

Moreover, conditionally load the remaining kprobe (vfs_fsync_range) to improve scheduler portability and allow to run the scheduler also in kernels where this function is not available.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -47,12 +47,6 @@ const volatile u64 slice_lag = 20ULL * NSEC_PER_MSEC;
 const volatile bool local_kthreads;
 
 /*
- * When enabled, threads who's hold a user-space lock and still able to run
- * will be dispatch first. See flash_dispatch() for details.
- */
-const volatile bool user_lock_boost;
-
-/*
  * Scheduling statistics.
  */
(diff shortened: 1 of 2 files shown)


likewhatevs (approved):
Nice, also I'm gonna move a wrapper around copies of that that if-else-unsafe block that are in layered also to a helper in compat which you can use some time later today/tomorrow btw.

multics69 (approved):

arighi:
@likewhatevs I was also wondering why in layered you're not using compat::ksym_exists() and you check /proc/kallsyms directly. Is it because nvidia_open() is in an external module? In that case we should probably fix compat.

likewhatevs:
> @likewhatevs I was also wondering why in layered you're not using compat::ksym_exists() and you check /proc/kallsyms directly. Is it because nvidia_open() is in an external module? In that case we should probably fix compat.

compat may just work -- I need to test that and will fix it if it doesn't.

PR #1446, opened 2025-03-02
scheds: Introduce scx_tickless

arighi:
scx_tickless is a server-oriented scheduler, derived from scx_central, designed to minimize OS noise and maximize performance isolation.

The scheduler works by routing all scheduling events through a pool of primary CPUs assigned to handle these events. This allows disabling the scheduler's tick on other CPUs, reducing OS noise.

Tasks are added to a global DSQ, and the primary CPUs distribute them across the other "tickless" CPUs, sending preemption events via IPC in case multiple tasks are competing for the same CPU.

In situations where the system is not overcommitted, tasks can run uninterrupted on their assigned CPUs, leading to better performance isolation and lower OS noise.

Typical use cases include cloud computing, virtualization and high performance computing workloads. This scheduler is *not* designed for latency-sensitive workloads.

NOTE: in order to effectively disable ticks on the "tickless" CPUs the kernel must be booted with nohz_full.

Example:

 - start a VM running a CPU-intensive workload:
```
   $ vng -vr -- stress-ng -c 0
```
 - on the host measure the rate of sched_tick events/sec:
```
   $ bpftrace -e 'kprobe:sched_tick { @ticks++; } interval:s:1 { print(@ticks); clear(@ticks); }'
```
With the default scheduler:
```
 @ticks: 1027
 @ticks: 1021
 @ticks: 1016
 @ticks: 1016
 @ticks: 1029
 @ticks: 1024
 @ticks: 1015
 @ticks: 1022
 @ticks: 1014
 @ticks: 1056
 ...
```

With scx_tickless:
```
 @ticks: 1000
 @ticks: 1001
 @ticks: 1001
 @ticks: 1000
 @ticks: 1001
 @ticks: 1000
 @ticks: 1000
 @ticks: 1002
 @ticks: 1000
 @ticks: 1000
 ...
```


multics69 (commented):
The code mixes scx_bpf_now() and bpf_ktime_get_ns()., Is this intentional?

arighi:
> The code mixes scx_bpf_now() and bpf_ktime_get_ns()., Is this intentional?

Nope, I'll fix that, thanks for catching it!

scheds/rust/scx_tickless/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,696 @@
...
+ *
+ * Return true if a task was dispatched, false otherwise.
+ */
+static bool dispatch_cpu(s32 cpu, bool same_cpu)
+{
+	struct task_struct *p;
+	bool dispatched = false;
+
+	bpf_for_each(scx_dsq, p, SHARED_DSQ, 0) {
+		p = bpf_task_from_pid(p->pid);
```
htejun, line 267: Maybe add a comment explaining that this is working around a BPF issue and isn't needed on newer kernels?

scheds/rust/scx_tickless/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,696 @@
...
+		/*
+		 * This tickless CPU is about to go idle: kick and preempt
+		 * a primary CPU so that it can dispatch a task here.
+		 *
+		 * We need to send a preemption IPI since primary CPUs may
+		 * also run a task with infinite time slice if all the
+		 * non-primary CPUs were busy and a task was dispatched
+		 * there.
+		 */
+		scx_bpf_kick_cpu(pick_primary_cpu(), SCX_KICK_PREEMPT);
```
htejun, line 403: Wouldn't it be cheaper for !primary cpu to dispatch directly if it's about to go idle anyway?

scheds/rust/scx_tickless/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,696 @@
...
+			continue;
+
+		if (time_before(now, cctx->started_at + slice_ns))
+			continue;
+
+		if (!scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | cpu) &&
+		    !scx_bpf_dsq_nr_queued(SHARED_DSQ))
+			continue;
+
+		scx_bpf_kick_cpu(cpu, SCX_KICK_PREEMPT);
```
htejun, line 559: I *think* it should be possible to call `dispatch_all_cpus()` directly from the timer.

htejun (approved):

sirlucjan:
@arighi Is there any chance that before the next stable release we will add this to scx_loader?

arighi:
@sirlucjan sure, I'll add the change to include it in scx_loader in this PR

scheds/rust/scx_tickless/src/bpf/main.bpf.c
arighi: Actually we're already dispatching directly, we've just called `scx_bpf_dsq_move_to_local(SHARED_DSQ)`. So we didn't consume any task from SHARED_DSQ it means either that SHARED_DSQ is empty (and in this case it's pointless to kick a primary CPU) or none of remaining tasks in the SHARED_DSQ can be consumed on this CPU. But also in this case a primary CPU won't do much, because if there are CPUs that can consume tasks from the SHARED_DSQ they can run them from their ops.dispatch().

In conclusion, I think we can just remove this `scx_bpf_kick_cpu()` and let the CPU go idle without doing anything special.

scheds/rust/scx_tickless/src/bpf/main.bpf.c
arighi: Almost... I can't check the available dispatch slots from the timer:
```
swapper/0[0] triggered exit kind 1024:
  runtime error (kfunc with mask 0x2 called from an operation only allowing 0x0)

Backtrace:
  scx_bpf_dispatch_nr_slots+0x52/0x70
  bpf_prog_93597a2c6163323b_dispatch_all_cpus+0x1b7/0x224
  bpf_prog_4cfe5fb148bda16b_sched_timerfn+0xd1/0x28e
  bpf_timer_cb+0x6b/0xc0
```

sirlucjan:
```lucjan at cachyos ~/DEV/github/archlinux-lucjan/archlinux-packages 17:38:20 2fdacd24 master    
❯ scxctl list                        
supported schedulers: ["bpfland", "flash", "lavd", "p2dq", "tickless", "rusty"]
```
Everything is fine! Thanks.

scheds/rust/scx_tickless/src/bpf/main.bpf.c
htejun: You don't need to check nr_slots from other paths. Dispatches don't go through the dispatch buffer outside of ops.dispatch().

arighi: yep, I'll investigate more and see if we can move the dispatches in the timer handler

scheds/rust/scx_tickless/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// Print scheduler version and exit.
+    #[clap(short = 'V', long, action = clap::ArgAction::SetTrue)]
+    version: bool,
+
+    /// Show descriptions for statistics.
+    #[clap(long)]
+    help_stats: bool,
+}
+
+fn is_smt_active() -> bool {
```
hodgesds, line 111: As a later cleanup this should be moved into `scx_utils` so it doesn't have to be reimplemented so many places.

hodgesds (approved):

scheds/rust/scx_tickless/src/main.rs
arighi: yes, good point!

arighi:
Alright, as discussed in office hours, now we have a BPF timer on each primary CPU that is distributing tasks among the tickless CPUs.

We still need to call `dispatch_all_cpus()` from `ops.dispatch()` on the primary CPUs, otherwise they would directly consume tasks (since they're often awake) and we don't want that. Ideally we want to bounce tasks to the available tickless CPUs first, so basically the dispatch can happen via the timer or through `ops.dispatch()` (bounce from a primary CPU or directly consuming tasks from the shared DSQ).

I'm not super happy how we pin the BPF timers to each primary CPU. In practice each timer is initialized and started from `ops.dispatch()` to make sure that we're running on each primary CPU. However, once the timer is started we still keep checking if it's initialized each time `ops.dispatch()` is invoked, which is ugly, but I haven't found a better way...

We can probably improve a bunch of things, but in general it seems to work.

PR #1447, opened 2025-03-02
scx_bpfland: fix cache domain detection

arighi:
Properly differentiate between CPUs that share the same cache and those that belong to the same SMT core.

This fixes a bug where L2/L3 caches were incorrectly identified as belonging to the same SMT core, causing them to be skipped.

Fixes: def6d305 ("scx_bpfland: avoid creating redundant cache domains")

### DIFF
--- a/scheds/rust/scx_bpfland/src/main.rs
+++ b/scheds/rust/scx_bpfland/src/main.rs
@@ -498,6 +498,22 @@ impl<'a> Scheduler<'a> {
         Ok(())
     }
 
+    fn are_smt_siblings(topo: &Topology, cpus: &[usize]) -> bool {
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
LGTM, I think if using `Cpu` structs from topology you could check the `package_id`/`cluster_id` but this works as well.

arighi:
> LGTM, I think if using `Cpu` structs from topology you could check the `package_id`/`cluster_id` but this works as well.

I considered that, but I was concerned about handling ID conflicts in the presence of multiple nodes or sockets. The sibling_cpus() function already ensures unique IDs, so it seemed safer/easier. :)

multics69:
LGTM

PR #1449, opened 2025-03-03
scxtop: Allow attaching hardware specific perf events

hodgesds:
Allow attaching hardware specific perf events using the CLI or the config file. This allows for a range of hardware specific events to be collected. The list of available perf events can be found in the perf tool configuration in the kernel source:
https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/perf/pmu-events/arch/x86

An example of the config file is as follows:
```
$ head ~/.config/scxtop/scxtop.toml
perf_events = ["ls_dmnd_fills_from_sys.local_l2:0x243:4", "ucode_ops:0x1c2:4", "ls_smi_rx:0x2b:4"] theme = "MidnightGreen"
tick_rate_ms = 250
debug = false
exclude_bpf = false
trace_ticks = 5
worker_threads = 4
trace_tick_warmup = 3
```

### DIFF
--- a/tools/scxtop/src/cli.rs
+++ b/tools/scxtop/src/cli.rs
@@ -55,6 +55,9 @@ pub struct TuiArgs {
     /// Process to monitor or all.
     #[arg(long, default_value_t = -1)]
     pub process_id: i32,
+    /// Custom perf events colon delimited (ex: "<event_name>:<event and umask ex: 0x023>:<event_type ex: 4>")
+    #[arg(long)]
+    pub perf_events: Vec<String>,
 
     /// Automatically start a trace when a function takes too long to return.
     #[arg(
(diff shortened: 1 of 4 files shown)


tools/scxtop/src/cli.rs
```diff
@@ -55,6 +55,9 @@ pub struct TuiArgs {
     /// Process to monitor or all.
     #[arg(long, default_value_t = -1)]
     pub process_id: i32,
+    /// Custom perf events colon delimited (ex: "<event_name>:<event and umask ex: 0x023>:<event_type ex: 4>")
```
likewhatevs, line 58: Is there a risk of an observer effect from supporting multiple perf events like this? Do we need a mitigation for this?

likewhatevs (approved):

tools/scxtop/src/cli.rs
hodgesds: There's only ever one perf event attached at a time. This just is expanding the available list so that the selectable perf events has a bigger range. On newer processors there's [plenty of counters](https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/programmer-references/58550-0.01.pdf):

> There are six core performance event counters per thread, six performance events counters per L3 complex and sixteen Data Fabric performance events counters mapped to the RDPMC instruction

The thing that will probably need to be added in the future is a differentiation between a CPU and a L3 perf event (since aggregating L3 perf events doesn't really make sense).

PR #1450, opened 2025-03-03
compat: add kprobe helper and update ksym helper

likewhatevs:
This adds a helper, `cond_kprobe_enable` to common. 

This will enable the passed in kprobe or print a warning if the symbol is missing. 

To support this, this switches `compat::ksym_exists` to using `/proc/kallsyms` (which, IIUC, will always work, vs the current approach which does not work with symbols in (some?) modules).

This also updates layered to use this for one conditional kprobe it has (nvidia_open) and updates stress-test config to have CI cover this code.

nvidia gpu machine (w/ symbol present) using this:
image

amd gpu machine (w/ symbol missing) using this:
image

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -5,10 +5,13 @@
 
 use anyhow::{anyhow, bail, Context, Result};
 use libbpf_rs::libbpf_sys::*;
+use libbpf_rs::{AsRawLibbpf, OpenProgramImpl};
+use log::warn;
 use std::ffi::c_void;
 use std::ffi::CStr;
 use std::ffi::CString;
 use std::io;
+use std::io::BufRead;
 use std::mem::size_of;
 use std::slice::from_raw_parts;
 
@@ -150,6 +153,32 @@ pub fn ksym_exists(ksym: &str) -> Result<bool> {
     let ksym_name = CString::new(ksym).unwrap();
     let tid = unsafe { btf__find_by_name(btf, ksym_name.as_ptr()) };
     Ok(tid >= 0)
+}
+
+pub fn in_kallsyms(ksym: &str) -> Result<bool> {
+    let file = std::fs::File::open("/proc/kallsyms")?;
+    let reader = std::io::BufReader::new(file);
+
+    for line in reader.lines() {
+        for sym in line.unwrap().split_whitespace() {
+            if ksym == sym {
+                return Ok(true);
+            }
+        }
+    }
+
+    Ok(false)
+}
+
+pub fn cond_kprobe_enable<T>(sym: &str, prog_ptr: &OpenProgramImpl<T>) -> Result<()> {
+    if in_kallsyms(sym)? {
+        unsafe {
+            bpf_program__set_autoload(prog_ptr.as_libbpf_object().as_ptr(), true);
+        }
+    } else {
+        warn!("symbol {} is missing, kprobe not loaded", sym);
+    }
+    Ok(())
 }
 
 pub fn is_sched_ext_enabled() -> io::Result<bool> {
--- a/meson-scripts/stress_tests.ini
+++ b/meson-scripts/stress_tests.ini
@@ -30,7 +30,7 @@ timeout_sec: 15
 
 [scx_layered]
 sched: scx_layered
-sched_args: --run-example -v --stats 1
+sched_args: --run-example -v --stats 1 --enable-gpu-support
 stress_cmd: stress-ng -t 14 --aggressive -M -c `nproc` -f `nproc`
 timeout_sec: 15
 bpftrace_scripts: dsq_lat.bt,process_runqlat.bt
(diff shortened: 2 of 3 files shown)


rust/scx_utils/src/compat.rs
```diff
@@ -145,11 +149,28 @@ pub fn struct_has_field(type_name: &str, field: &str) -> Result<bool> {
 }
 
 pub fn ksym_exists(ksym: &str) -> Result<bool> {
-    let btf: &btf = *VMLINUX_BTF;
+    let file = std::fs::File::open("/proc/kallsyms")?;
```
hodgesds, line 152: This is interesting in that trusting vmlinux is probably not right in all circumstances, but would this work without  `CONFIG_DEBUG_INFO` being set? Not quite sure what the right thing to do here is, maybe having a fallback to vmlinux would be good.

rust/scx_utils/src/compat.rs
```diff
@@ -145,11 +149,28 @@ pub fn struct_has_field(type_name: &str, field: &str) -> Result<bool> {
...
-    let ksym_name = CString::new(ksym).unwrap();
-    let tid = unsafe { btf__find_by_name(btf, ksym_name.as_ptr()) };
-    Ok(tid >= 0)
+    for line in reader.lines() {
+        for sym in line.unwrap().split_whitespace() {
+            syms_present.insert(sym.into());
+        }
+    }
+
+    Ok(syms_present.contains(ksym))
```
JakeHillion, line 162: There's no reason to create a set here. Return true from the for loop if the string matches, else return false at the end of the function.

rust/scx_utils/src/compat.rs
htejun: Is this necessary to detect symbols in modules? I'd much prefer to stick with BTF given that the rest of compat machinery depends on compat - ie. it's not just testing "is that ksymbol listed somewhere?" but also "is it known to BTF and linkable / relocatable?". Maybe use `btf__load_module_btf()` or make this a separate path for kprobes?

likewhatevs: > This is interesting in that trusting vmlinux is probably not right in all circumstances, but would this work without CONFIG_DEBUG_INFO being set? 

My thinking is "if it's present on ubuntu server and bazzite w/o a custom kernel, it's probably standard," and it is so it's probably standard?

> Is this necessary to detect symbols in modules?

"Yes" as in -- yes confirmed experimentally (i.e. w/o doing anything fancy/non-standard, yes/this works).

> I'd much prefer to stick with BTF given that the rest of compat machinery depends on compat - ie. it's not just testing "is that ksymbol listed somewhere?" but also "is it known to BTF and linkable / relocatable?". Maybe use btf__load_module_btf() or make this a separate path for kprobes?

Ahh I follow, will add a separate path for the ksym_exists part of this.

rust/scx_utils/src/compat.rs
likewhatevs: The premise of this assertion is us knowing that checking every entry in this split list with duplicates is cheaper than inserting this split list with duplicates into a set and checking if that set contains the passed in string. Do we know this?

likewhatevs:
Avoided editing ksym_exists and instead added ksym_probeable (open to better names, that one seems clear-ish). 

Still works as expected across symbol present/symbol missing:
image

rust/scx_utils/src/compat.rs
```diff
@@ -152,6 +155,32 @@ pub fn ksym_exists(ksym: &str) -> Result<bool> {
     Ok(tid >= 0)
 }
 
+pub fn ksym_probeable(ksym: &str) -> Result<bool> {
```
htejun, line 158: How about `in_kallsyms()`?

rust/scx_utils/src/compat.rs
```diff
@@ -152,6 +155,32 @@ pub fn ksym_exists(ksym: &str) -> Result<bool> {
...
+    let file = std::fs::File::open("/proc/kallsyms")?;
+    let reader = std::io::BufReader::new(file);
+
+    for line in reader.lines() {
+        for sym in line.unwrap().split_whitespace() {
+            if ksym == sym {
+                return Ok(true);
+            }
+        }
+    }
```
htejun, line 168: Not necessary right now but it'd probably be better to build a hashset and keep it in a lazy_static. Or is the worry that we wanna re-parse the whole thing in case modules have been loaded inbetween?

htejun (approved):
Sans function naming, looks fine to me.

rust/scx_utils/src/compat.rs
JakeHillion: I think minimally this is going to use >24MB of memory, probably not a big deal as it's swappable.

likewhatevs: I like this idea. Will do it in another PR though because I think it will also need a timer/expiration (or something) to periodically reset the map (for when layered starts before modules are loaded, which I think is more realistic of a use case than it sounds like due to the popularity of AMD handhelds and nvidia eGPU's (presuming eGPU's work like this on linux at all -- I've only done this w/ windows)).

htejun: Oh, let's not overcomplicate it. Reading and discarding is fine.

PR #1452, opened 2025-03-03
scxtop: always send bpf event to terminate trace

### LINKED ISSUE #1317: scxtop: Hang during perfetto trace capture

JakeHillion:
Although this BPF timer work started with the intention of making sure the
trace always terminated, the initial implementation makes it less likely to
terminate. This is because the userspace ticks are guaranteed to happen since
#1433, but the single timer callback cannot guarantee it will insert into the ring
buffer.

Add an additional mode called TRACE_STOPPING for the case where there was a
previous trace that should have stopped but userspace hasn't been informed yet.
Retrigger the timer every 5ms until it manages to enter the ringbuf, which is
made more likely by having already reduced the sample rate from BPF side.
Finally, clean up any events that came after the true end of the trace in
userspace, as these events will be partial due to the sample rate.

There's an exceedingly unlikely case where practically all events are being
dropped that means the trace takes a very long time to terminate. This will
have the previous issue of consuming an unbounded amount of userspace memory,
and will probably result in something (hopefully scxtop) getting OOM killed.
If this shows up we should solve it by having a userspace cutoff too that stops
storing events, but I haven't seen it yet.

Closes #1317

Test plan:
- Tested on a busy machine with lots of dropped events. The trace terminates.
- Artificially introduced a large amount of latency into `TryFrom<&bpf_event for Action>`.
  The trace eventually terminates.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -26,7 +26,7 @@ use crate::LICENSE;
 use crate::SCHED_NAME_PATH;
 use crate::{
     Action, CpuhpAction, GpuMemAction, IPIAction, SchedCpuPerfSetAction, SchedSwitchAction,
-    SchedWakeupAction, SchedWakingAction, SoftIRQAction, TraceStartedAction,
+    SchedWakeupAction, SchedWakingAction, SoftIRQAction, TraceStartedAction, TraceStoppedAction,
 };
 
 use anyhow::Result;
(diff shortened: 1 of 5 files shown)


hodgesds (approved):
LGTM

PR #1453, opened 2025-03-04
scxtop: adding hw_pressure event trace.

devnexen:
tracks hardware related contention in a cpu scheduling context. could cpu frequency related, high cpu cache miss rate, poor cpu locality
 but also memory ; thermal throttling.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -25,8 +25,9 @@ use crate::APP;
 use crate::LICENSE;
 use crate::SCHED_NAME_PATH;
 use crate::{
-    Action, CpuhpAction, GpuMemAction, IPIAction, SchedCpuPerfSetAction, SchedSwitchAction,
-    SchedWakeupAction, SchedWakingAction, SoftIRQAction, TraceStartedAction, TraceStoppedAction,
+    Action, CpuhpAction, GpuMemAction, HwPressureAction, IPIAction, SchedCpuPerfSetAction,
+    SchedSwitchAction, SchedWakeupAction, SchedWakingAction, SoftIRQAction, TraceStartedAction,
+    TraceStoppedAction,
 };
 
 use anyhow::Result;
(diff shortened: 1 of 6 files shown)


tools/scxtop/src/app.rs
```diff
@@ -2351,6 +2352,21 @@ impl<'a> App<'a> {
         }
     }
 
+    /// Handles hardware pressure events.
+    pub fn on_hw_pressure(&mut self, action: &HwPressureAction) {
+        let HwPressureAction { cpu, hw_pressure } = action;
+
+        let cpu_data = self.cpu_data.entry(*cpu as usize).or_insert(CpuData::new(
```
hodgesds, line 2359: This is probably not needed unless it's worth adding it to the TUI.

hodgesds:
Few other questions, I don't see the event being attached in userspace, I think you might need something like:
```
diff --git a/tools/scxtop/src/main.rs b/tools/scxtop/src/main.rs
index 42057b14..970c2761 100644
--- a/tools/scxtop/src/main.rs
+++ b/tools/scxtop/src/main.rs
@@ -115,20 +115,25 @@ fn attach_progs(skel: &mut BpfSkel) -> Result<Vec<Link>> {
         links.push(link);
     }
     if let Ok(link) = skel.progs.on_cpuhp_enter.attach() {
         links.push(link);
     }
     if compat::ksym_exists("gpu_memory_total").is_ok() {
         if let Ok(link) = skel.progs.on_gpu_memory_total.attach() {
             links.push(link);
         }
     }
+    if compat::ksym_exists("hw_pressure_update").is_ok() {
+        if let Ok(link) = skel.progs.on_hw_pressure_update.attach() {
+            links.push(link);
+        }
+    }

     Ok(links)
 }

 fn run_trace(trace_args: &TraceArgs) -> Result<()> {
     TermLogger::init(
         match trace_args.verbose {
             0 => simplelog::LevelFilter::Info,
             1 => simplelog::LevelFilter::Debug,
             _ => simplelog::LevelFilter::Trace,
```
Is the metric a per-cpu metrics or is the pressure system wide? It seems a bit strange to report it per CPU as I don't think it is at that granularity. With that patch applied I still don't see any changes even if I `stress-ng -c $(nproc)`:
image

devnexen:
> Is the metric a per-cpu metrics or is the pressure system wide? It seems a bit strange to report it per CPU as I don't think it is at that granularity. With that patch applied I still don't see any changes even if I `stress-ng -c $(nproc)`: 

According to the kernel tracepoint it s cpu and pressure level.
When you set debug true, is it skipped ?

tools/scxtop/src/app.rs
```diff
@@ -578,6 +592,11 @@ impl<'a> App<'a> {
                             format!(" {}", format_hz(cpu_freq))
                         } else {
                             "".to_string()
+                        },
+                        if hw_pressure > 0 {
+                            format!("hw/pressure({})", hw_pressure)
```
hodgesds, line 597: nit: `format!(" hw_pressure({})", hw_pressure)`

hodgesds (approved):

Issue #1418, opened 2025-02-25
scx_lavd missing kernel options?


dim-geo:
Hi, thanks for the great work!
In my desktop I use scx_lavd with very nice results.

I use the latest git to make sure that scheduling improvements are used.
When I build this version 
`commit 211512acbee86982c1637a32f9472946d7762146 (HEAD -> main, origin/main, origin/HEAD)`

scx_lavd fails to run with:

```
libbpf: map '.rodata.str1.1': created successfully, fd=14
libbpf: map 'lavd_ops': created successfully, fd=15
libbpf: failed to open '/sys/kernel/debug/tracing/events/syscalls/sys_enter_futex/id': -ENOENT
libbpf: failed to determine tracepoint 'syscalls/sys_enter_futex' perf event ID: -ENOENT
libbpf: prog 'rtp_sys_enter_futex': failed to create tracepoint 'syscalls/sys_enter_futex' perf event: -ENOENT
libbpf: prog 'rtp_sys_enter_futex': failed to auto-attach: -ENOENT
Error: Failed to attach non-struct_ops BPF programs
```

am I missing some kernel configs?
I am on kernel 6.12.13

scx_lavd worked fine a few days ago.

dim-geo:
Yes
```
zcat /proc/config.gz | grep -i kprobes
CONFIG_KPROBES=y
CONFIG_KPROBES_ON_FTRACE=y
CONFIG_HAVE_KPROBES=y
CONFIG_HAVE_KPROBES_ON_FTRACE=y
# CONFIG_SAMPLE_KPROBES is not set

```

multics69:
Hi @dim-geo, thank you for reporting the problem.
Does your kernel have `CONFIG_FTRACE_SYSCALLS`?
Otherwise, you should turn it on.

multics69:
@dim-geo  -- I submitted a PR (https://github.com/sched-ext/scx/pull/1454) to gracefully disable futex holder boosting when tracepoints are available.

dim-geo:
> Hi [@dim-geo](https://github.com/dim-geo), thank you for reporting the problem. Does your kernel have `CONFIG_FTRACE_SYSCALLS`? Otherwise, you should turn it on.
> 
> BTW, what's the distribution you are using? I didn't expect `CONFIG_FTRACE_SYSCALLS` is not on. If necessary, I can change `lavd` to launch successfully even when there are no tracepoints to hook on.

[...]

PR #1454, opened 2025-03-05
scx_utils, scx_lavd: Enable futex holder boosting conditionally.

multics69:
When a kernel is compiled without CONFIG_FTRACE_SYSCALLS, the relevant tracepoints for futex boosting do not exist. For this case, conditionally enable futex holder boost when the tracepoints exist and '--no-futex-boost' option is not provided.

This PR addresses **https://github.com/sched-ext/scx/issues/1418**

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -170,17 +170,49 @@ pub fn in_kallsyms(ksym: &str) -> Result<bool> {
     Ok(false)
 }
 
-pub fn cond_kprobe_enable<T>(sym: &str, prog_ptr: &OpenProgramImpl<T>) -> Result<()> {
+pub fn tracepoint_exists(tracepoint: &str) -> Result<bool> {
+    let file = std::fs::File::open("/sys/kernel/debug/tracing/available_events")?;
+    let reader = std::io::BufReader::new(file);
+
+    for line in reader.lines() {
+        for tp in line.unwrap().split_whitespace() {
+            if tracepoint == tp {
+                return Ok(true);
+            }
+        }
+    }
+
+    Ok(false)
+}
+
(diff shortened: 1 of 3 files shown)


rust/scx_utils/src/compat.rs
```diff
@@ -170,17 +170,49 @@ pub fn in_kallsyms(ksym: &str) -> Result<bool> {
...
+        return Ok(true);
     } else {
         warn!("symbol {} is missing, kprobe not loaded", sym);
     }
-    Ok(())
+
+    Ok(false)
 }
 
+pub fn cond_tracepoint_enable<T>(tracepoint: &str, prog_ptr: &OpenProgramImpl<T>) -> Result<bool> {
```
arighi, line 201: Since it's a lib, instead of printing the warning directly here, I'd return an error to the caller, like we do with the `tracepoint_exists()` and then change the function to just return a Result.

Something like:
```
pub fn cond_tracepoint_enable<T>(
    tracepoint: &str,
    prog_ptr: &OpenProgramImpl<T>,
) -> Result<()> {
    if tracepoint_exists(tracepoint)? {
        unsafe {
            bpf_program__set_autoload(prog_ptr.as_libbpf_object().as_ptr(), true);
        }
        Ok(())
    } else {
        Err(Error::new(
            ErrorKind::NotFound,
            format!("tracepoint {} is missing, tracepoint not loaded", tracepoint),
        ))
    }
}
```

But this can be improved later, the same applies also to `cond_kprobe_enable()`.

arighi (approved):
I left a comment, but as a future improvement. This one LTGM.

rust/scx_utils/src/compat.rs
multics69: That's a good suggestion. After cutting the release, I will submit a separate PR cleaning this up for lavd and layered.

PR #1456, opened 2025-03-05
scx_utils: Fix kernels without CONFIG_DEBUG_FS

anh0516:
tracefs is usually mounted on both /sys/kernel/tracing and /sys/kernel/debug/tracing, I guess for compatibility reasons.

On kernels without CONFIG_DEBUG_FS, /sys/kernel/debug doesn't exist, so we should use /sys/kernel/tracing instead to avoid errors such as "Error: No such file or directory (os error 2)" from scx_lavd with the default futex holder boosting enabled.

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -171,7 +171,7 @@ pub fn in_kallsyms(ksym: &str) -> Result<bool> {
 }
 
 pub fn tracepoint_exists(tracepoint: &str) -> Result<bool> {
-    let file = std::fs::File::open("/sys/kernel/debug/tracing/available_events")?;
+    let file = std::fs::File::open("/sys/kernel/tracing/available_events")?;
     let reader = std::io::BufReader::new(file);
 
     for line in reader.lines() {
--- a/scheds/rust/scx_bpfland/src/main.rs
+++ b/scheds/rust/scx_bpfland/src/main.rs
@@ -211,7 +211,7 @@ struct Opts {
     #[clap(long)]
     monitor: Option<f64>,
 
-    /// Enable BPF debugging via /sys/kernel/debug/tracing/trace_pipe.
+    /// Enable BPF debugging via /sys/kernel/tracing/trace_pipe.
     #[clap(short = 'd', long, action = clap::ArgAction::SetTrue)]
     debug: bool,
 
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -94,7 +94,7 @@ struct Opts {
     exit_dump_len: u32,
 
     /// Enable verbose output, including libbpf details. Moreover, BPF scheduling events will be
-    /// reported in debugfs (e.g., /sys/kernel/debug/tracing/trace_pipe).
+    /// reported in tracefs (e.g., /sys/kernel/tracing/trace_pipe).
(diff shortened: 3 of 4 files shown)


multics69 (approved):
Nice catch! Thanks!

JohnRTitor:
Unfortunately this broke other kernels that have `CONFIG_DEBUG_FS` enabled.

```

[root@Ainz-NIX:~]# stat /sys/kernel/debug/tracing/available_events
  File: /sys/kernel/debug/tracing/available_events
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 0,13	Inode: 2049        Links: 1
Access: (0440/-r--r-----)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-03-10 10:10:39.298000000 +0530
Modify: 2025-03-10 10:10:39.298000000 +0530
Change: 2025-03-10 10:10:39.298000000 +0530
 Birth: -

[root@Ainz-NIX:~]# stat /sys/kernel/tracing/available_events 
stat: cannot statx '/sys/kernel/tracing/available_events': No such file or directory

[root@Ainz-NIX:~]# zcat /proc/config.gz | grep DEBUG_FS
# CONFIG_XEN_DEBUG_FS is not set
CONFIG_BLK_DEBUG_FS=y
# CONFIG_SCSI_SNIC_DEBUG_FS is not set
# CONFIG_SCSI_LPFC_DEBUG_FS is not set
# CONFIG_USB_GADGET_DEBUG_FS is not set
# CONFIG_OCFS2_DEBUG_FS is not set
CONFIG_DEBUG_FS=y
CONFIG_DEBUG_FS_ALLOW_ALL=y
# CONFIG_DEBUG_FS_DISALLOW_MOUNT is not set
# CONFIG_DEBUG_FS_ALLOW_NONE is not set
```

Since this patch is present in v1.10, upgrading to that version is broken (https://github.com/NixOS/nixpkgs/pull/388603, https://github.com/chaotic-cx/nyx/pull/989).

```
❯  sudo ./result-bin/bin/scx_lavd 
05:50:02 [INFO] Autopilot mode is enabled by default.
Error: No such file or directory (os error 2)
```

This doesn't affect other schedulers of course. Tested with `scx_rustland`, rusty, bpfland, central...

JohnRTitor:
CC @htejun @JakeHillion

anh0516:
Hmm. What's `systemctl status sys-kernel-tracing.mount` say? Is tracefs mounted on `/sys/kernel/tracing?`

anh0516:
I was just about to say check both and I see you already fixed it. It's strange to me that tracefs isn't always mounted on both (it is on Debian and RHEL, for example), but as long as it works everywhere that's good.

JohnRTitor:
> Hmm. What's `systemctl status sys-kernel-tracing.mount` say? Is tracefs mounted on `/sys/kernel/tracing?`

I suppose this is just an issue with our (NixOS)'s end. We don't enable `tracefs` by default for some reason. I will try to raise a PR for this.

JakeHillion:
> I suppose this is just an issue with our (NixOS)'s end. We don't enable `tracefs` by default for some reason. I will try to raise a PR for this.

Definitely have my upvote for this. I have to mount tracefs on NixOS every time I use scxtop, having it enabled in the distro kernel would be great.

JohnRTitor:
https://github.com/NixOS/nixpkgs/pull/388751

Issue #1373, opened 2025-02-18
[Feature request] Include scxctl into scx/sched-ext


sirlucjan:
It would be a great idea to integrate [scxctl](https://github.com/frap129/scxctl) in scx or add this as a separate repository in sched-ext. @frap129  has done a great job and this is a very useful tool. Tested on CachyOS and Fedora 41 Kinoite.

```
lucjan at cachyos ~ 12:35:48    
❯ scxctl --help
A cli interface for scx_loader

Usage: scxctl <COMMAND>

Commands:
  get     Get the current scheduler and mode
  list    List all supported schedulers
  start   Start a scheduler in a mode or with arguments
  switch  Switch schedulers or modes, optionally with arguments
  stop    Stop the current scheduler
  help    Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
lucjan at cachyos ~ 12:35:53    
❯ scxctl list  
supported schedulers: ["bpfland", "flash", "lavd", "p2dq", "rusty"]  
lucjan at cachyos ~ 12:36:11    
❯ scxctl start -s bpfland
started scx_bpfland in auto mode
lucjan at cachyos ~ 12:36:39    
❯ scxctl get                                                             
running bpfland in auto mode
lucjan at cachyos ~ 12:36:49    
❯ systemctl status scx_loader.service
● scx_loader.service - DBUS on-demand loader of sched-ext schedulers
     Loaded: loaded (/usr/lib/systemd/system/scx_loader.service; disabled; preset: disabled)
     Active: active (running) since Tue 2025-02-18 11:44:08 CET; 52min ago
 Invocation: 17b213e4700c4574a728925d02b5b046
   Main PID: 115114 (scx_loader)
      Tasks: 22 (limit: 37392)
     Memory: 31.5M (peak: 32M)
        CPU: 54ms
     CGroup: /system.slice/scx_loader.service
             ├─115114 /usr/bin/scx_loader
             └─120351 scx_bpfland

lut 18 12:36:39 cachyos scx_loader[120351]: 11:36:39 [INFO] L2 cache ID 0: sibling CPUs: [0, 8]
lut 18 12:36:39 cachyos scx_loader[120351]: 11:36:39 [INFO] L2 cache ID 1: sibling CPUs: [1, 9]
lut 18 12:36:39 cachyos scx_loader[120351]: 11:36:39 [INFO] L2 cache ID 2: sibling CPUs: [2, 10]
[...]

PR #1458, opened 2025-03-05
Add scxctl into tools

frap129:
This PR adds `scxctl` code to the main scx repo, resolving #1373 
- Added it in `tools/`, though it could arguably go in `rust/` with `scx_loader`
- Relicenced to GPL 2 to match the rest of this repo
- Updated the repo link in Cargo.toml to point to this repo, rather than mine
- Updated the version to be in sync with scx


sirlucjan:
```
ls: cannot access '/home/lucjan/workroom/Repo/sched-ext/scx-scheds-git/src/scx/build/*/scxctl': No such file or directory exists

FAILED: install script '/home/lucjan/Workroom/Repo/sched-ext/scx-scheds-git/src/scx/meson-scripts/install_rust_user_scheds' failed with exit code 2.
```
I got this error during building.

htejun:
Let's merge this after release.

tools/scxctl/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+edition = "2021"
+description = "A cli dbus client for scx_loader"
+authors = ["Joe Maples <joe@maples.dev>"]
+repository = "https://github.com/sched-ext/scx"
+readme = "README.md"
+keywords = ["linux", "scx", "scx_scheds", "scx_loader", "sched_ext"]
+categories = ["command-line-utilities"]
+license = "GPL-2.0-only"
+
+[profile.release]
```
hodgesds, line 16: I think this gets picked up by the top level cargo, so this is probably not needed.

tools/scxctl/src/scx_loader.rs
```diff
@@ -0,0 +1,93 @@
+use clap::ValueEnum;
+use zbus::proxy;
+
+/*
+ * D-bus interface info
+ */
+
+#[proxy(
```
vnepogodin, line 8: why not just use `scx_loader` crate, I don't think code duplication is necessary here. Unless you want to minimize dependencies coming from scx_loader

frap129: Initially it was to minimize dependencies, but now that scxctl is becoming part of the same repo it would probably make more sense to use the crate

sirlucjan:
```
lucjan at cachyos ~ 21:12:07    
❯ pacman -Fl scx-scheds-git | grep scxctl
scx-scheds-git usr/bin/scxctl
lucjan at cachyos ~ 21:12:17    
❯ scxctl list
supported schedulers: ["bpfland", "flash", "lavd", "p2dq", "tickless", "rusty"]
lucjan at cachyos ~ 21:12:23    
❯ scxctl get 
no scx scheduler running
lucjan at cachyos ~ 21:12:30    
❯ scxctl start -s flash
started scx_flash in auto mode
lucjan at cachyos ~ 21:12:35    
❯ systemctl status scx_loader.service    
● scx_loader.service - DBUS on-demand loader of sched-ext schedulers
     Loaded: loaded (/usr/lib/systemd/system/scx_loader.service; disabled; preset: disabled)
     Active: active (running) since Wed 2025-03-05 21:12:22 CET; 22s ago
 Invocation: 78a72abac2ed48f298fb4fca67f0e731
   Main PID: 95208 (scx_loader)
      Tasks: 22 (limit: 37391)
     Memory: 21.4M (peak: 22M)
        CPU: 42ms
     CGroup: /system.slice/scx_loader.service
             ├─95208 /usr/bin/scx_loader
             └─95318 scx_flash

mar 05 21:12:22 cachyos systemd[1]: Starting DBUS on-demand loader of sched-ext schedulers...
mar 05 21:12:22 cachyos scx_loader[95208]: [INFO]: Starting as dbus interface
mar 05 21:12:22 cachyos systemd[1]: Started DBUS on-demand loader of sched-ext schedulers.
mar 05 21:12:35 cachyos scx_loader[95208]: [INFO]: starting Flash with mode Auto..
mar 05 21:12:35 cachyos scx_loader[95208]: [INFO]: Got event to start scheduler!
mar 05 21:12:35 cachyos scx_loader[95208]: [INFO]: starting scx_flash command
mar 05 21:12:35 cachyos scx_loader[95318]: 20:12:35 [INFO] scx_flash 1.0.7-gfb111651-dirty x86_64-unknown-linux-gnu SMT on
```
Good job @frap129 and thank you very much!

sirlucjan (approved):
Tested earlier from a separate repo, tested now - works great.

tools/scxctl/src/scx_loader.rs
frap129: After taking a look, its not as simple as just switching to the crate. The zbus interface in `scx_loader `expects sched name to be passed as `SupportedSched` and modes as `SchedMode`, whereas I've been handling them as String and u32. I still think it makes sense to switch to the crate, but would probably be more appropriate as a follow up PR

htejun (approved):

htejun:
Unrelated but we probably should move scx_loader under tools/ too.

multics69:
Thank you @frap129 for contributing this!

tools/scxctl/src/scx_loader.rs
vnepogodin: I can send PR which implements that, after merge

arighi (approved):

htejun:
CI is busted. Merged manually.

PR #1459, opened 2025-03-06
scx_layered: Improve kthread handling

htejun:
Bugs in `keep_running()` and preemption logic made `ksoftirqd` latencies frequently go over hundreds of msecs on moderately loaded systems. Fix the bugs and promote per-cpu kthreads such that they're considered preempting by default. This brings down ksoftirqd latency outliers by more than a couple orders of magnitude.


likewhatevs (approved):
fantastic

PR #1460, opened 2025-03-06
scx_layered: Implement IsKthread match

htejun:
While at it, make nearby code a bit more consistent.


hodgesds (approved):
Yeah, seems like an obvious match type.

PR #1463, opened 2025-03-06
scxtop: Disable bpf conditional attachment

hodgesds:
Userspace correctly handles conditional attachment based on available kfuncs so no need to conditionally attach on the bpf side.

DSQ stats render properly:
image


JakeHillion (approved):
ACK.

PR #1465, opened 2025-03-07
scx_utils, scx_lavd: Add cpu_capacity to Topology and use it in scx_lavd.

multics69:
There are various sysfs sources that somehow represent each CPU's capacity even if the kernel does not properly implement arch_scale_cpu_capacity(), which is used for scx_bpf_cpuperf_cap().

This PR tries to make a best guess about CPU capacity based on available sysfs entries by checking the most precise (e.g., vendor-specific information) to the least precise (CPU max frequency).

Also, scx_lavd is changed to use cpu_capacity exposed by Topology.

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -115,6 +115,8 @@ pub struct Cpu {
     /// Base operational frqeuency. Only available on Intel Turbo Boost
     /// CPUs. If not available, this will simply return maximum frequency.
     pub base_freq: usize,
+    /// The best-effort guessting of cpu_capacity scaled to 1024
+    pub cpu_capacity: usize,
     pub trans_lat_ns: usize,
     pub l2_id: usize,
     pub l3_id: usize,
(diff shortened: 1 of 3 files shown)


multics69:
Ooops, it fails in our CI. I will take a look.

multics69:
> Ooops, it fails in our CI. I will take a look.

Fixed.

arighi (approved):
LGTM. Few comments.

The static approach works in most of the cases, but in some architectures the capacity can change at runtime (see for example #1207). In general it'd be nice to have an event-based API to catch topology changes. But I'm not sure how to implement something like this.

Or maybe just have an API similar to `scx_bpf_cpuperf_cap()` to get a more accurate capacity value, see for example `get_actual_cpu_capacity()` in fair.c, that takes in account the cpufreq pressure, thermal throttling and potentially other factors.

But this is not related to this PR, just some thoughts for the future.

multics69:
> get_actual_cpu_capacity

That's definitely a way to go. Thanks for the suggestion.

PR #1467, opened 2025-03-07
scx_lavd: Optimize core compaction at ops.dispatch() and more

multics69:
Optimize the core compaction to minimize the number of using CPUs and to reduce latency by more conservatively activating a CPU and more proactively finding a task that cannot be run on the current active/overflow set. In addition to that, add SCX_OPS_ENQ_LAST to avoid latency spikes in under-utilized cases.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -2210,7 +2210,8 @@ SCX_OPS_DEFINE(lavd_ops,
 	       .init_task		= (void *)lavd_init_task,
 	       .init			= (void *)lavd_init,
 	       .exit			= (void *)lavd_exit,
-	       .flags			= SCX_OPS_KEEP_BUILTIN_IDLE |
+	       .flags			= SCX_OPS_ENQ_LAST |
+					  SCX_OPS_KEEP_BUILTIN_IDLE |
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1378,91 +1377,79 @@ void BPF_STRUCT_OPS(lavd_dispatch, s32 cpu, struct task_struct *prev)
...
-	 * find and run a task pinned to run on this CPU.
+	 * find and run the task pinned on this CPU.
 	 */
 	bpf_for_each(scx_dsq, p, dsq_id, 0) {
 		/*
-		 * Prioritize kernel tasks because most kernel tasks are pinned
-		 * to a particular CPU and latency-critical (e.g., ksoftirqd,
-		 * kworker, etc).
-		 */
-		if (is_kernel_task(p)) {
```
arighi, line 1399: This also makes kthread prioritization unnecessary, right? Maybe you can mention it in the committee message.

arighi (approved):
Left a comment, but LGTM.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Sure. Thanks!

PR #1468, opened 2025-03-07
scx_bpfland: Fix idle detection when SMT is disabled

ruihe774:
`is_fully_idle()` now checks whether @cpu is in a full-idle physical core dispite SMT is enabled or not.
This fixes a regression introduced by 43c9759f8c8258db66b593879c08c27ad47d216e (tasks may be dispatched to busy cores if SMT is disabled).

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -215,7 +215,7 @@ static bool is_queued(const struct task_struct *p)
 }
 
 /*
- * Return true if @cpu is a full-idle SMT core (or if SMT is disabled),
+ * Return true if @cpu is in a full-idle physical core,
  * false otherwise.
  */
 static bool is_fully_idle(s32 cpu)
@@ -223,9 +223,6 @@ static bool is_fully_idle(s32 cpu)
 	const struct cpumask *idle_smtmask;
 	int node = __COMPAT_scx_bpf_cpu_node(cpu);
 	bool is_idle;
-
-	if (!smt_enabled)
-		return true;
 
 	idle_smtmask = __COMPAT_scx_bpf_get_idle_smtmask_node(node);
 	is_idle = bpf_cpumask_test_cpu(cpu, idle_smtmask);


ruihe774:
> Do you see any difference with this change applied?

Yes. Without this change applied, an application can randomly hang for more than one second on a SMT disabled system.

> in the latter we don't want to check if `prev_cpu` is idle, because it's where the task is currently running and we just check if no other tasks are queued on that CPU.

I do not know very clear about it. If the core is not idle we should not direct dispatch to it, right? Maybe there are some race conditions that prev_cpu may become busy before try_direct_dispatch returns.

arighi:
> > Do you see any difference with this change applied?
> 
> Yes. Without this change applied, an application can randomly hang for more than one second on my SMT disabled system.

If you see hangs longer than 1s there's probably something else going on. What kind of workload are you running? Can you share some info so that I can try to reproduce it? Which options are you using with scx_bpfland (maybe you're using `-p`)?

> 
> > in the latter we don't want to check if `prev_cpu` is idle, because it's where the task is currently running and we just check if no other tasks are queued on that CPU.
> 
> I do not know very clear about it. If the core is not idle we should not direct dispatch to it, right? Maybe there are some race conditions that prev_cpu may become busy before try_direct_dispatch returns.

`try_direct_dispatch()` is called from `ops.enqueue()` when `ops.select_cpu()` is skipped, meaning that the task was likely re-enqueued because it expired its time slice. In this case we don't want to be too aggressive with migrations and we can keep the task on the same CPU if no other tasks have been queued (even if the particular CPU is busy).

ruihe774:
> In this case we don't want to be too aggressive with migrations and we can keep the task on the same CPU if no other tasks have been queued (even if the particular CPU is busy).

If so, why does it consider whether the physical core is idle or not if SMT is enabled? In this way, the logic differs between non-SMT and SMT systems.

> If you see hangs longer than 1s there's probably something else going on. What kind of workload are you running? Can you share some info so that I can try to reproduce it? Which options are you using with scx_bpfland (maybe you're using `-p`)?

I do not specify additional arguments. Here is the output of scx_bpfland:
```
scx_bpfland 1.0.10-g7a18ed78 x86_64-unknown-linux-gnu SMT off
scheduler flags: 0x16
primary CPU domain = 0x003f
cpufreq performance level: max
L2 cache ID 6: sibling CPUs: [6, 7, 8, 9]
L2 cache ID 7: sibling CPUs: [10, 11, 12, 13]
```
(7a18ed78 is the commit with the change applied.)

My system:
OS: `Fedora Linux 42.20250304.n.0 (Silverblue Prerelease) x86_64`
Kernel: `Linux 6.14.0-0.rc3.29.fc42.x86_64`
CPU: `Intel(R) Core(TM) i5-14600KF (14) @ 5.30 GHz`

Some typical workloads that hang frequently include games (especially Windows games on Wine) and coding (especially IDEs that run heavy code analysis in the background like products of JetBrains, with GPU acceleration off).

arighi:
> > In this case we don't want to be too aggressive with migrations and we can keep the task on the same CPU if no other tasks have been queued (even if the particular CPU is busy).
> 
> If so, why does it consider whether the physical core is idle or not if SMT is enabled? In this way, the logic differs between non-SMT and SMT systems.

Yeah that's a valid point, SMT and non-SMT should be consistent. Ok, with this assumption it makes sense to me, let's apply this, I'll do some tests to make sure nothing else is broken. For the SMT case it doesn't change anything, so we should be fine.

Thanks!

PR #1470, opened 2025-03-07
scx_layered: add protected layer flag

etsal:
Add a flag to scx_layered for marking Grouped and Confined layers as protected. The idle CPUs in these layers cannot be used to execute tasks that belong outside of the layer, with the exception of tasks that belong to layers marked with preempt.


etsal:
Heavily WIP, getting it to pass verification and ensure correct behavior, but wanted to get a first sketch out.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -103,6 +104,43 @@ static __always_inline struct layer *lookup_layer(u32 id)
...
+{
+	struct bpf_cpumask *cpumask;
+	const struct cpumask *idlemask;
+
+	if (!(cpumask = bpf_cpumask_create()))
+		return NULL;
+
+	idlemask = scx_bpf_get_idle_cpumask();
+
+	bpf_cpumask_and(cpumask, idlemask, cast_mask(protected_cpumask));
```
JakeHillion, line 118: This function says it returns the unprotected mask but the logic is to AND idle with protected. Wouldn't that return the protected mask? Wondering whether the `protected_cpumask` variable is a misnomer.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -103,6 +104,43 @@ static __always_inline struct layer *lookup_layer(u32 id)
...
+	const struct cpumask *idlemask;
+	struct bpf_cpumask *smtmask;
+
+	if (!(smtmask = bpf_cpumask_create()))
+		return NULL;
+
+	idlemask = scx_bpf_get_idle_smtmask();
+
+	/* XXXETSAL Do we have to also hold a protected smtmask? */
+	bpf_cpumask_and(smtmask, idlemask, cast_mask(protected_cpumask));
```
JakeHillion, line 137: ditto

JakeHillion (commented):
Can we get the documentation of this new flag in the `main.rs` docs please?

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: Good catch I flipped the polairyt of the cpumask because there is no easy way to do x & ~y

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -103,6 +105,53 @@ static __always_inline struct layer *lookup_layer(u32 id)
 	return &layers[id];
 }
 
+static __always_inline 
+struct bpf_cpumask *idle_cpumask_unprotected(struct cpu_ctx *cpuc)
+{
+	struct bpf_cpumask *cpumask;
+	const struct cpumask __kptr *idlemask;
+
+	if (!(cpumask = bpf_cpumask_create()))
```
htejun, line 114: I don't think we want to hot-allocate the cpumask on the spot. We already use cached masks extensively when picking idle CPU. It'd be better if we integrate this mask there instead of layering it on top of `scx_bpf_get_idle_cpumask()`.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -849,29 +899,31 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
+		bpfmask = idle_cpumask_unprotected(cpuc);
+		if (!bpfmask)
+			return -1;
 
 		if (layer->kind == LAYER_KIND_CONFINED)
-			has_idle = bpf_cpumask_intersects(layered_cpumask, cpumask);
+			has_idle = bpf_cpumask_intersects(layered_cpumask, cast_mask(bpfmask));
 		else
-			has_idle = bpf_cpumask_intersects(p->cpus_ptr, cpumask);
+			has_idle = bpf_cpumask_intersects(p->cpus_ptr, cast_mask(bpfmask));
```
htejun, line 909: I'm not sure this is correct. Updating like the above would mean that a layer would consider itself to be without any idle CPUs when only the CPUs it owns are idle.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -887,7 +939,7 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
 			cpu = -1;
 			goto out_put;
 		}
-		if ((cpu = pick_idle_cpu_from(cpumask, prev_cpu, idle_smtmask)) >= 0)
+		if ((cpu = pick_idle_cpu_from(cpumask, prev_cpu, cast_mask(idle_smtmask))) >= 0)
```
htejun, line 942: Ditto. Here, `cpumask` contains the CPUs that are owned by this layer in the previous LLC. Let's say the layer is a protected grouped one; then, `idle_smtmask` from idle_smtmask_unprotected() wouldn't contain any CPUs which intersect with the owned CPUs.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -908,18 +960,18 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
-	if ((cpu = pick_idle_cpu_from(layered_cpumask, prev_cpu, idle_smtmask)) >= 0)
+	if ((cpu = pick_idle_cpu_from(layered_cpumask, prev_cpu, cast_mask(idle_smtmask))) >= 0)
 		goto out_put;
 
 	/*
 	 * If the layer is an open one, we can try the whole machine.
 	 */
 	if (layer->kind != LAYER_KIND_CONFINED &&
-	    (cpu = pick_idle_cpu_from(p->cpus_ptr, prev_cpu, idle_smtmask)) >= 0) {
+	    (cpu = pick_idle_cpu_from(p->cpus_ptr, prev_cpu, cast_mask(idle_smtmask))) >= 0) {
```
htejun, line 974: This is the first time we're looking outside the owned CPUs and should consider whether CPUs are protected or not, no?

htejun (commented):
The preemption path would need to consider protected status too.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: You are right, we need to OR the CPU mask of unprotected CPUs with that of the layer itself before ANDing it with the idle CPU mask , same for line 942 below.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: It does seem like we should AND cand_cpumask in pick_idle_cpu_from because this is a potential problem every time it is called.

etsal:
> The preemption path would need to consider protected status too.

Do you mean that we should be preventing CPUs in protected layers from being preempted? I deliberately did not look into that in this patch for simplicity, making preempting layers be "more important" than protected layers. I will look into preventing protected layers from being preempted if the current behavior is confusing.

likewhatevs:
> > The preemption path would need to consider protected status too.
> 
> Do you mean that we should be preventing CPUs in protected layers from being preempted? I deliberately did not look into that in this patch for simplicity, making preempting layers be "more important" than protected layers. I will look into preventing protected layers from being preempted if the current behavior is confusing.

I could be wrong on this, but I don't think protected layers have to be protected from preempting layers (yet, perhaps). I think that issue is solvable with documentation (i.e. mention that's how this works). I also think changing this from how it is now might be as confusing (I think this is gonna be confusing however it's done, but we need it).

etsal:
- Added flag description to main.rs
- Fixed CPU mask calculation
- Removed CPU mask allocations from the hot path
- Account for the layer's mask when calculating available idle CPUs

etsal:
TODO: Fix pick_idle_cpu_from

etsal:
New version with unprotected_cpumask - aware pick_idle_cpu_from(), now at the  debugging stage. Had to add a percpu cache with preallocated cpumasks.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -157,6 +159,126 @@ struct {
...
+	idlemask = scx_bpf_get_idle_cpumask();
+	if (!idlemask || !unprotected_cpumask) {
+		scx_bpf_put_idle_cpumask(idlemask);
+		return NULL;
+	}
+
+	/* Candidate CPUs are either in this layer or not in a protected layer. */
+	bpf_cpumask_or(cpumask, layermask, cast_mask(unprotected_cpumask));
+
+	bpf_cpumask_and(cpumask, cast_mask(cpumask), idlemask);
```
htejun, line 249: The `unprotected` mask only changes when layer CPU allocations are updated and is only needed when a layer can't find an idle CPU that it owns. Maybe I'm confused but I don't think we need to be computing these masks dynamically. We'll only overflow into unprotected CPUs when there are no owned CPUs, so we don't necessarily need to or them with owned CPUs and even if we did we can calculate the result and cache it using `cached_cpus` and friends.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: We can either do two separate tests or keep the computed mask cached. Let's not hot-compute the masks.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -887,7 +1019,7 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
 			cpu = -1;
 			goto out_put;
 		}
-		if ((cpu = pick_idle_cpu_from(cpumask, prev_cpu, idle_smtmask)) >= 0)
+		if ((cpu = pick_idle_cpu_from(cpumask, prev_cpu, cast_mask(idle_smtmask), layer)) >= 0)
```
htejun, line 1022: So, here, I think the code is trying to pick an idle owned CPU which is in the same LLC as the previous CPU, right? I don't know why we would use `idle_smtmask_unprotected` here. Other places too. We don't care about unprotected cpus until we get to the step where there's no owned CPUs that we can pick, no?

etsal: Ok I think I see - We should applying the unprotected mask directly to each task and using cached_cpus to ensure it isn't stale. We can then use the regular idle cpumask as before. The reason I needed to add the OR'ing was because I did not incorporate unprotected_cpumask into the per-task cpumasks which should be pretty cheap thanks to the caching we are doing. Am I getting this right?

etsal: Actually we never need to worry about whether a CPU is protected until the end of pick_idle_cpu, right? Because this is the only time we do not choose from a subset of the CPUs owned by the layer. Applying the unprotected mask to the task masks would be a no-op so we don't even need to do that.

etsal:
Can test with the new config in examples/ by flooding the machine with stress-ng hogs and ensuring that the CPUs in the first layer are still at almost no utilization.

likewhatevs (approved):
LGTM

likewhatevs:
@etsal this is annoying, but is the case, for ease of things downstream, protected needs to be protected_layer or idk anything but the word protected (chokes something flaky).

etsal:
> @etsal this is annoying, but is the case, for ease of things downstream, protected needs to be protected_layer or idk anything but the word protected (chokes something flaky).
> 
>     * I want to say if protected isn't on intf.h, the issue goes away, but probably best to avoid keyword like strings like `protected` tbh.

ACK, I assume the problem is because of linters that are both for C and C++ and freaking out because protected is a keyword in the latter?

likewhatevs:
> ACK, I assume the problem is because of linters that are both for C and C++ and freaking out because protected is a keyword in the latter?

I think it's actually something related to codegen that isn't used (but unavoidable) tbh.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -848,13 +905,28 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
 			has_idle = bpf_cpumask_intersects(layered_cpumask, cpumask);
-		else
-			has_idle = bpf_cpumask_intersects(p->cpus_ptr, cpumask);
+		} else {
+			maybe_refresh_layered_cpus_unprotected(p, taskc);
+			/*
+			 * Use the task's idle unprotected mask if available, otherwise
+			 * use the global one.
+			 */
+			unprot_mask = taskc->layered_unprotected_mask;
```
htejun, line 918: So, this is `p->cpus_allowed & unprotected_mask`, right?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -848,13 +905,28 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
+			if (!unprot_mask)
+				unprot_mask = unprotected_cpumask;
+
+			if (unlikely(!unprot_mask)) {
+				scx_bpf_error("unprotected_cpumask not initialized");
+				scx_bpf_put_idle_cpumask(cpumask);
+				return -1;
+			}
+
+			has_idle = bpf_cpumask_intersects(cast_mask(unprot_mask), cpumask);
```
htejun, line 928: and then we're saying "if `(p->cpus_allowed & unprotected_mask) & idle_cpumask` is empty, we have no idle CPUs". However, that's not quite the right condition. The right condition would be `(p->cpus_allowed & (layered->cpus | unprotected_cpumask)` because the task has access to both the layer's owned CPUs and unprotected CPUs.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -918,10 +990,16 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
...
 	 */
-	if (layer->kind != LAYER_KIND_CONFINED &&
-	    (cpu = pick_idle_cpu_from(p->cpus_ptr, prev_cpu, idle_smtmask)) >= 0) {
+	if (layer->kind != LAYER_KIND_CONFINED) {
+	    maybe_refresh_layered_cpus_unprotected(p, taskc);
+	    unprot_mask = taskc->layered_unprotected_mask;
+	    if (!unprot_mask)
+		    unprot_mask = unprotected_cpumask;
+
+	    if ((cpu = pick_idle_cpu_from(cast_mask(unprot_mask), prev_cpu, idle_smtmask)) >= 0) {
```
htejun, line 999: However, the usage here is correct, I think, because we already checked all the owned CPUs in the layer and we just need to check the intersection between `p->cpus_allowed` and `unprotected_cpumask`.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -453,6 +484,8 @@ struct task_ctx {
 	struct bpf_cpumask __kptr *layered_llc_mask;
 	struct cached_cpus	layered_cpus_node;
 	struct bpf_cpumask __kptr *layered_node_mask;
+	struct cached_cpus	layered_cpus_unprotected;
+	struct bpf_cpumask __kptr *layered_unprotected_mask;
```
htejun, line 488: `layered_unprotected_mask` is probably a bit of misnomer. The above two masks have `layered` in the name because they're intersection of `p->cpus_allowed`, `layered->cpus` and the llc or node's cpumask. The new one just wants to do `p->cpus_allowed & unprotected_cpumask`, so the `layered` part of the name doesn't really make sense anymore.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: Ah, this is ultimately a problem with maybe_refresh_unprotected mask. This should be (p->cpus_allowed | unprotected_mask), this mask lists the CPUs that either a) are owned by us or b) are not owned by a protected layer. The refresh function is ANDing the two masks instead of ORing. Would this be correct in this case?

etsal: I see, yes this is an error in maybe_refresh_layered_cpus_unprotected. It should be the expression in your comment below.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: I think so, updating the unprotected cpumask calculation to be correct will not affect this either since the layer has no idle CPUs so we can leave this as-is.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: Sorry, missed this comment.  Now that the mask is (layer_cpumask | unprotected_mask) & p->cpumask, is the name more appropriate?

PR #1473, opened 2025-03-08
scx_p2dq: Add config for min runs for pick2

hodgesds:
Add a config that tracks the number of runs the task has run on the local LLC and do not consider tasks below the threshold available to pick2 load balance on the wakeup path.

Test on 80 core dual socket (40 CPUs per socket) machine with `stress-ng`:
```
$ stress-ng -c 60
```

Run with 100 slices before migrating on the local LLC to make tasks sticky:
```
$ scx_p2dq --stats 1  -g -l 100
direct/idle/greedy 208/29342/0
        dsq same/migrate 30415/3400
        keep 0 pick2 4
        migrations llc/node: 1559/1559
direct/idle/greedy 194/31168/0
        dsq same/migrate 32933/3320
        keep 0 pick2 3
        migrations llc/node: 2433/2433
direct/idle/greedy 321/28467/0
        dsq same/migrate 29682/3378
        keep 0 pick2 31
        migrations llc/node: 1467/1467
```
image

image

Run with 1 slice for thrashing:
```
$ scx_p2dq --stats 1  -g -l 1
direct/idle/greedy 143/28412/0
        dsq same/migrate 29553/3158
        keep 0 pick2 727
        migrations llc/node: 1125/1125
direct/idle/greedy 268/30955/0
        dsq same/migrate 32142/4230
        keep 0 pick2 804
        migrations llc/node: 1823/1823
direct/idle/greedy 331/37811/0
        dsq same/migrate 38857/5179
        keep 0 pick2 1036
        migrations llc/node: 1448/1448
```
image
image

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -40,6 +40,7 @@ const volatile u32 nr_dsqs_per_llc = 3;
 const volatile u64 dsq_shift = 2;
 const volatile int init_dsq_index = 0;
 const volatile u64 min_slice_us = 100;
+const volatile u64 min_llc_runs_pick2 = 5;
 const volatile u32 interactive_ratio = 10;
 const volatile u32 min_nr_queued_pick2 = 10;
 
@@ -124,7 +125,9 @@ static struct cpu_ctx *lookup_cpu_ctx(int cpu)
 }
 
 static __always_inline u64 cpu_dsq_id(int dsq_index, struct cpu_ctx *cpuc) {
-	if (dsq_index < 0 || dsq_index > nr_dsqs_per_llc || dsq_index >= MAX_DSQS_PER_LLC) {
+	if (dsq_index < 0 ||
+	    dsq_index > nr_dsqs_per_llc ||
+	    dsq_index >= MAX_DSQS_PER_LLC) {
 		scx_bpf_error("cpuc invalid dsq index: %d", dsq_index);
 		return 0;
 	}
@@ -267,7 +270,8 @@ static bool keep_running(struct cpu_ctx *cpuc, struct task_struct *p)
 	return true;
 }
 
-static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *left, struct llc_ctx *right, bool most_loaded)
+static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *left,
+					struct llc_ctx *right, bool most_loaded)
 {
 	s32 left_queued = 0, right_queued = 0;
 	u64 left_load = 0, right_load = 0;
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_p2dq/src/main.rs
```diff
@@ -91,6 +91,11 @@ struct Opts {
     #[clap(short = 's', long, default_value = "100")]
     min_slice_us: u64,
 
+    /// Number of slices on the LLC before a task becomes eligbile for pick2 migration on the
+    /// wakeup path.
+    #[clap(short = 'l', long, default_value = "5")]
```
JakeHillion, line 96: Maybe default this to 0 to maintain the existing behaviour?

JakeHillion (approved):
Implementation looks good, but I'm not totally convinced by measuring this in slices - perhaps I'm misunderstanding!

Looks like on wakeup we'll consider whether this task has had ample opportunity to run on its current LLC (warm caches etc) before considering another. Because we measure this in slices it can take an unbounded amount of time to get there due to our lack of deadlines. Would it be better to use some sort of wait time estimate here? e.g. if the current LLC has 100 tasks queued with 10ms slices it might take a second to run, and I'm not sure it matters if we ran for one or two slices on this LLC before.

I'm thinking that with this implementation tasks could become doomed by being on an overloaded LLC which would make tasks less likely to move off of it at wake, which is probably a pretty good time to move them to a lighter queue.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -296,6 +297,10 @@ static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *left, struct llc_ctx *ri
 
 static s32 pick_two_cpu(struct task_ctx *taskc, bool *is_idle)
 {
+	if (min_llc_slices_pick2 > 0 && taskc->llc_slices < min_llc_slices_pick2)
+		return -EINVAL;
+
```
multics69, line 302: An extra new line?

multics69 (approved):
I like the idea. The PR overall looks good. It would be a bit easier to read the change if the goal of the changes is described in the commit message, something like this: "The goal is to reduce task migration across LLC domains by enforcing temporal stickiness.

hodgesds:
> I like the idea. The PR overall looks good. It would be a bit easier to read the change if the goal of the changes is described in the commit message, something like this: "The goal is to reduce task migration across LLC domains by enforcing temporal stickiness.

Good suggestion, I was a little bit lazy with the commit messages. Let me clean those up!

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -289,18 +293,21 @@ static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *left, struct llc_ctx *ri
...
+	    (left_queued < min_nr_queued_pick2 &&
+	     right_queued < min_nr_queued_pick2))
 		return NULL;
 
-	if (((left_queued > 0 || right_queued > 0) && left_queued < right_queued) ||
-	    left_load > right_load)
+	if (left_queued < right_queued || left_load > right_load)
 		return most_loaded ? right : left;
 
-	return NULL;
```
hodgesds, line 299: Drive by fix, this was a bug to make this NULL.

PR #1475, opened 2025-03-08
scx_p2dq: Add stats for pick2 on dispatch/select paths

hodgesds:
Add separate stats for accounting for pick2 load balancing on the select and dispatch paths. This is to get better insight into load balancing behavior.

Example output:
```
direct/idle/greedy 268/4059/0
        dsq same/migrate 20951/1131
        keep 0
        pick2 select/dispatch 10/17
        migrations llc/node: 1359/0
```


JakeHillion (approved):
Nice.

multics69 (approved):

PR #1478, opened 2025-03-09
scx_utils: Topology: Provide smt_enabled

arighi:
Add the new property smt_enabled in Topology that can be used by the schedulers to check if SMT is enabled in the system.

Also change all the schedulers to use this property and get rid of all the duplicate code.


hodgesds (approved):
Absolute legend!

multics69 (approved):
Thanks!

PR #1479, opened 2025-03-09
scx_flash: simplify scheduling policy

arighi:
With the recent kernel improvements (such as the topology awareness in the built-in idle CPU selection policy and the NUMA functionalities) we can now get rid of the custom `ops.select_cpu()` implementation and rely on the in-kernel one.

Moreover, we can simplify the overall scheduling policy, aligning more to the scx_bpfland algorithm, also applying some fixes as well.

Overall, with these new changes in place, the scheduler seems to provide a better level of responsiveness, it's also really hard to trigger audio cracking issues, even when the system is massively overloaded and we can reduce the code significantly:

```
 scheds/rust/scx_flash/src/bpf/main.bpf.c | 700 +++--------------------
 scheds/rust/scx_flash/src/main.rs        |  85 ---
 2 files changed, 91 insertions(+), 694 deletions(-)
```

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -14,11 +14,6 @@ char _license[] SEC("license") = "GPL";
 #define MAX_AVG_NVCSW		128
 
 /*
- * Global DSQ used to dispatch tasks.
- */
-#define SHARED_DSQ		0
-
-/*
  * Task time slice range.
  */
 const volatile u64 slice_max = 20ULL * NSEC_PER_MSEC;
@@ -149,9 +144,9 @@ static inline bool is_kthread(const struct task_struct *p)
 /*
  * Return the amount of tasks that are waiting to run.
  */
-static inline u64 nr_tasks_waiting(void)
+static inline u64 nr_tasks_waiting(int node)
 {
-	return scx_bpf_dsq_nr_queued(SHARED_DSQ) + 1;
+	return scx_bpf_dsq_nr_queued(node) + 1;
 }
 
 /*
@@ -199,14 +194,15 @@ static u64 task_deadline(const struct task_struct *p, struct task_ctx *tctx)
  * Evaluate task's time slice in function of the total amount of tasks that
  * are waiting to be dispatched and the task's weight.
  */
-static u64 task_slice(const struct task_struct *p, const struct task_ctx *tctx)
+static u64 task_slice(const struct task_struct *p,
+		      const struct task_ctx *tctx, int node)
 {
 	/*
 	 * Assign a time slice proportional to the task weight and inversely
 	 * proportional to the total amount of tasks that are waiting to be
 	 * scheduled.
 	 */
-	return scale_up_fair(p, slice_max / nr_tasks_waiting());
+	return scale_up_fair(p, slice_max / nr_tasks_waiting(node));
 }
 
 /*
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -697,12 +429,6 @@ void BPF_STRUCT_OPS(flash_enqueue, struct task_struct *p, u64 enq_flags)
 	scx_bpf_dsq_insert_vtime(p, SHARED_DSQ, SCX_SLICE_DFL,
 				 task_vtime(p, tctx), enq_flags);
 	__sync_fetch_and_add(&nr_shared_dispatches, 1);
-
-	/*
-	 * Try to proactively wake up an idle CPU, so that it can
-	 * immediately execute the task in case its current CPU is busy.
-	 */
-	kick_task_cpu(p, tctx);
```
multics69, line 705: I think the proactively waking up an idle CPU would still be necessary, right?

multics69 (approved):
Looks good! A lot simpler! I left one question for clarification.

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: Ah good point. Without the proactive kick, CPU-intensive tasks are more likely staying on the same CPU, but we should at least try to kick their current CPU. I'll add that, thanks!

arighi: And BTW, I intentionally don't want to be too much proactive at kicking CPUs in `ops.enqueue()`, because interactive tasks will go often through `ops.select_cpu()`, so they always get a chance to pick a different idle CPU, while CPU-intensive tasks will be re-enqueued via `ops.enqueue()` on slice exhaustion, skipping `ops.select_cpu()`, so they won't steal potential idle CPUs from the interactive tasks (unless the system is fully saturated, in which case all the CPUs are already up, so there's no need for the proactive kick).

In theory this should give interactive tasks a little extra priority boost when the system is not fully saturated and keeping CPU-intensive tasks on the same CPU should actually help them if they're also cache-sensitive. 

I'll add a comment to clarify this in the code.

multics69: Thank you for the explanation @arighi ! 

I am a bit confused. Anyway, kick_task_cpu() uses scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE), so it should not hinder the progress of CPU-intensive tasks, where their CPUs won't be idle, in my understanding.

arighi: @multics69 basically what I want to do in ops.enqueue() is this:
```
        s32 cpu = scx_bpf_task_cpu(p);

        scx_bpf_dsq_insert_vtime( ... );

       if (scx_bpf_test_and_clear_cpu_idle(cpu))
```

In this way this CPU won't be picked during `ops.select_cpu()`, since it's marked as busy by `scx_bpf_test_and_clear_cpu_idle()`, and the current task will likely keep running on this CPU if it's re-enqueued to `ops.enqueue()`.

Then other tasks can pick other CPUs from `ops.select_cpu()`  and this task will less likely bounce around. If we allow kicking other CPUs than the current one in `ops.enqueue()`, there're more chances for the task to bounce somewhere else.

Instead, tasks that go through `ops.select_cpu()` can easily kick other CPUs, so, when that happens, we increase work conservation globally and tasks are more likely bouncing around, since more CPUs will be awake.

arighi: To be even more conservative we could use per-CPU DSQs and keep adding tasks to the same CPU DSQ in `ops.enqueue()` and only allow migrations from `ops.select_cpu()`, but this is a bit too much conservative and it hurts responsiveness when the system is busy.

multics69: Thank you again @arighi ! I missed the kick CPU part (in below) in the enqueue path. Looks great!

```
       if (scx_bpf_test_and_clear_cpu_idle(cpu))
```

hodgesds (approved):
LGTM, will `scx_flash` still work on older kernels as well? If not it may be worth adding a note about it.

Issue #1482, opened 2025-03-10
scx_lavd is broken on DEBUG_FS kernels


JohnRTitor:
https://github.com/sched-ext/scx/pull/1456 fixed `scx_lavd` on kernels without DEBUG_FS, but broke it on kernels with one.

### Tests done

```

[root@Ainz-NIX:~]# stat /sys/kernel/debug/tracing/available_events
  File: /sys/kernel/debug/tracing/available_events
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 0,13	Inode: 2049        Links: 1
Access: (0440/-r--r-----)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-03-10 10:10:39.298000000 +0530
Modify: 2025-03-10 10:10:39.298000000 +0530
Change: 2025-03-10 10:10:39.298000000 +0530
 Birth: -

[root@Ainz-NIX:~]# stat /sys/kernel/tracing/available_events 
stat: cannot statx '/sys/kernel/tracing/available_events': No such file or directory

[root@Ainz-NIX:~]# zcat /proc/config.gz | grep DEBUG_FS
# CONFIG_XEN_DEBUG_FS is not set
CONFIG_BLK_DEBUG_FS=y
# CONFIG_SCSI_SNIC_DEBUG_FS is not set
# CONFIG_SCSI_LPFC_DEBUG_FS is not set
# CONFIG_USB_GADGET_DEBUG_FS is not set
# CONFIG_OCFS2_DEBUG_FS is not set
CONFIG_DEBUG_FS=y
CONFIG_DEBUG_FS_ALLOW_ALL=y
# CONFIG_DEBUG_FS_DISALLOW_MOUNT is not set
# CONFIG_DEBUG_FS_ALLOW_NONE is not set
```

```
❯  sudo ./result-bin/bin/scx_lavd 
05:50:02 [INFO] Autopilot mode is enabled by default.
Error: No such file or directory (os error 2)
```

Since this patch made to 1.0.10 release, we on Nixpkgs can't upgrade to `1.0.10`.

Also CCing @ptr1337 as it seems to affect `linux-cachyos` as well.

CC: @htejun @JakeHillion

multics69:
Thanks for the report, @JohnRTitor ! I will fix it soon.

multics69:
@JohnRTitor  -- I submitted the PR. If possible, could you please check if it works on your end?

[...]

PR #1483, opened 2025-03-10
scx_utils: Fix the failure in tracepoint_exists() in some kernel configs.

multics69:
tracepoint_exists() assumed that "/sys/kernel/tracing/available_events" always exists, which is not true in some kernel configurations. So let's check both "/sys/kernel/tracing/available_events" and "/sys/kernel/debug/tracing/available_events".

This fixes the following issue:
  https://github.com/sched-ext/scx/issues/1482

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -171,7 +171,10 @@ pub fn in_kallsyms(ksym: &str) -> Result<bool> {
 }
 
 pub fn tracepoint_exists(tracepoint: &str) -> Result<bool> {
-    let file = std::fs::File::open("/sys/kernel/tracing/available_events")?;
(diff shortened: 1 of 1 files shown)


arighi (approved):

JohnRTitor:
Can we get this backported/made into a hotfix release since this bug makes `scx_lavd` unusable?

multics69:
> Can we get this backported/made into a hotfix release since this bug makes `scx_lavd` unusable?

@JohnRTitor -- It seems okay in other distros (like CachyOS, Gentoo). Is there any way to include the fix in the nix package (I have almost zero knowledge on the nix packaging)?

JohnRTitor:
Done, thanks. We also enabled `tracefs` by default. So hopefully bugs like these won't bother again. https://github.com/NixOS/nixpkgs/pull/388751

PR #1486, opened 2025-03-10
scx_utils: Fix cargo warning

hodgesds:
Fix cargo linter warning for a variable being unnecessarily mutable.

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -171,7 +171,7 @@ pub fn in_kallsyms(ksym: &str) -> Result<bool> {
 }
 
 pub fn tracepoint_exists(tracepoint: &str) -> Result<bool> {
-    let mut file = match std::fs::File::open("/sys/kernel/tracing/available_events") {
+    let file = match std::fs::File::open("/sys/kernel/tracing/available_events") {
         Err(_) => std::fs::File::open("/sys/kernel/debug/tracing/available_events")?,
         Ok(file) => file,
     };


arighi (approved):

JohnRTitor:
Hi, this once again broke scx_lavd, causing the same problem as https://github.com/sched-ext/scx/issues/1482. This was fixed in https://github.com/sched-ext/scx/pull/1483 by @multics69, by querying whether the file exists and trying the next possible file path. Removing the mut once again broke it. 

CC @JakeHillion

hodgesds:
I guess it needs a more proper fix like querying the tracefs mount. Maybe I'll move this [helper](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/perf_event.rs#L58) into `scx_utils`.

JakeHillion:
> Hi, this once again broke scx_lavd, causing the same problem as #1482. This was fixed in #1483 by @multics69, by querying whether the file exists and trying the next possible file path. Removing the mut once again broke it.
> 
> CC @JakeHillion

Hey, this specific change didn't break Nix. It's a no-op: Rust doesn't need this variable to be mutable because the match statement selects either the initial result or the second and places that into the variable. This means that the variable is not mutated after its initial assignment.

For future reference, removing `mut` with code that still compiles in Rust will always be safe.

@JohnRTitor

s0me1newithhand7s:
> I am not sure what's going on here, as I can't reproduce it. But hopefully @s0me1newithhand7s can add more context.

firstly: 
```Shell
 ~ 𝑠𝑛𝑜𝑤𝑦 𝑝𝑙𝑎𝑐𝑒, 𝑓𝑢𝑙𝑙 𝑜𝑓 𝑓𝑙𝑎𝑘𝑒𝑠! 

╭──╼ s0mePC-nix on NixOS 25.5.0 ❄️
┆ ~
╰─> uname -a                                                                                                                     ✓  at ❗ 14:45
Linux s0mePC-nix 6.13.6-cachyos #1-NixOS SMP PREEMPT_DYNAMIC Fri Mar  7 17:27:16 UTC 2025 x86_64 GNU/Linux
```
so:
1. NixOS Unstable 
2. cachyos-linux kernel

brief run of all other scheds resulted with no errors;
`scx_lavd` runs trough `scxctl` still ends with same `Error: No such file or directory (os error 2)`;
reverting versions won't help; i've tried, `scx_lavd` behave same way;

i think this is some `boot.kernelParam` or/and `boot.sysctl` that could cause this weird `scx_lavd` behavior.
attaching my current sysctls and kernelParams:
image
image

i can't really find any reasons why `scx_lavd` doesn't works while others do..

JohnRTitor:
Right, you passed `debugfs=off` as a kernel parameter. `scx_lavd` needs either `tracefs` or `debugfs`. NixOS have always had debugfs support, but tracefs by default was only enabled recently. That feature hasn't yet reached unstable channels yet. https://github.com/NixOS/nixpkgs/pull/388751

It should reach nixos-unstable in coming days. For now, you can reenable debugfs to use lavd properly.

PR #1487, opened 2025-03-10
scx_wd40: import kkd's arena spinlocks from the kernel

etsal:
Support for arena spin locks has recently landed on the bpf-next branch of the Linux kernel, allowing us to directly embed spin locks in data structures that reside in arenas. This support is in the form of headers that define a spin lock struct backed by arena memory along with lock/unlock operations that can replace the existing bpf_spin_lock data structure we are using. 

These new spinlocks are useful because we can replace with them all  bpf_spin_lock instances in our schedulers. The bpf_spin_lock type cannot be embedded in arenas, and has required us to use maps along our arena-allocated scheduler structures.

Import the headers defining the spinlock types into WD40 in anticipation of using them to simplify the codebase.

### DIFF
--- a/scheds/rust/scx_wd40/src/bpf/bpf_arena_spin_lock.h
+++ b/scheds/rust/scx_wd40/src/bpf/bpf_arena_spin_lock.h
@@ -1,0 +1,512 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2025 Meta Platforms, Inc. and affiliates. */
+#ifndef BPF_ARENA_SPIN_LOCK_H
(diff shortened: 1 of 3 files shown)


htejun (approved):

multics69:
Thank you @etsal ! It would be great if you could add some description (e.g. why you need this, etc) in the commit message. That way the others can easily catch up what is going on.

etsal:
> Thank you @etsal ! It would be great if you could add some description (e.g. why you need this, etc) to the commit message. That way the others can easily catch up what is going on.

For sure! I was planning on doing this import and following up quickly with a patch that actually uses it which would clear things out. I will add a description to this one for completeness.

PR #1489, opened 2025-03-11
scx_bpfland: handle per-CPU tasks preempted by a higher scheduling class

arighi:
When a CPU is stolen by a higher scheduling class we re-schedule all the tasks in the local DSQ from ops.cpu_release() to give them a chance to be dispatched on a different CPU.

However, this is pointless for per-CPU tasks, since they can only run on that specific CPU, so re-scheduling them is only increasing the risk of triggering stalls.

Therefore, dispatch them directly on reschedule, to maximize their chances of reclaiming the CPU quickly and avoiding stalls.

This approach will be more effective once dl_server support is added to the sched_ext core.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -734,9 +734,24 @@ static bool try_direct_dispatch(struct task_struct *p, struct task_ctx *tctx,
 	 * taken by a higher priority scheduling class, force it to follow
 	 * the regular scheduling path and give it a chance to run on a
 	 * different CPU.
+	 *
+	 * However, if the task can only run on a single CPU, re-scheduling
+	 * is unnecessary, as it can only be dispatched on that specific
+	 * CPU. In this case, dispatch it immediately to maximize its
+	 * chances of reclaiming the CPU quickly and avoiding stalls.
+	 *
+	 * This approach will be effective once dl_server support is added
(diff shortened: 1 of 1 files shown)


multics69 (approved):
Thank you for catching the niche case! I think the other scx schedulers should handle it similarly.

arighi:
> Thank you for catching the niche case! I think the other scx schedulers should handle it similarly. BTW, how does dl_server help in this case? Running the locally dispatched task in its cycle?

Right, all the scx schedulers should probably do the same, I was also wondering if we should implement something like this internally when `ops.cpu_release()` isn't implemented...

The dl_server will work similarly to the fair class: the server will be started in `enqueue_task_scx()` when the first task is being enqueued on a rq and stopped in `dequeue_task_scx()` when the last task of the rq is de-queued. Then the dl_server is basically a `SCHED_DEADLINE` task that will guarantee that `SCHED_EXT` tasks will have at least a small amount of time (50ms over 1sec IIRC) even when higher priority classes (rt and fair) are monopolizing the CPU.

arighi:
BTW once all of this is in place it'll also make partial mode more viable I think, since we won't necessarily need to isolate SCHED_EXT tasks and SCHED_NORMAL tasks on separate subsets of CPUs to avoid stalls (it's still probably a good idea to do the isolation, but at least we won't have immediate stalls).

hodgesds (approved):
Good find!

PR #1494, opened 2025-03-11
scx_p2dq: Use SCX_DSQ_LOCAL_ON for idle CPU selection

hodgesds:
When idle CPUs are found in select_cpu use SCX_DSQ_LOCAL_ON to dispatch to the CPUs local DSQ.


hodgesds:
Whoops, @etsal did the same thing in #1495

PR #1495, opened 2025-03-11
scx_p2dq: enqueue to selected_cpu

etsal:
The .select_cpu() method always enqueues on the local CPU even if another one is selected because it uses SCX_DSQ_LOCAL. Use SCX_DSQ_LOCAL_ON | cpu instead to enqueue to the CPU picked by pick_idle_cpu().


arighi (commented):
Using `SCX_DSQ_LOCAL` in `ops.select_cpu()` is fine, since the target CPU will be determined by the return value of `ops.select_cpu()`, so as long as you "return cpu" this change shouldn't be needed.

PR #1497, opened 2025-03-12
scx_layered: add cpus_pct cfg param

likewhatevs:
scx_layered: add cpus_pct config param.

An issue we have ran into is that we want some "fixed" amount of a machines CPUs to be allocated to different layers.  Fixed in the sense of "I want a few CPUs running ansible and many CPUs running nginx", and I don't want those CPUs changing.

CPUs range works fine for this when the number of CPUs is static across hosts, but does not work when we want to use the same configuration across hosts with different numbers of CPUs.

This PR pulls some of the `cpus_range` setup logic out into a helper function and overrides it when cpus_pct is set, enabling layered configs to better err, "do what we want" if that makes sense.

I get 16 cores on layer 1 when I set a layer's cpus_pct to 50 on a 9950x (32 logical cpu machine).

image

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -1011,8 +1011,8 @@ fn resolve_cpus_pct_range(
             {
                 bail!("cpus_range_frac values must be between 0.0 and 1.0");
             }
-            let cpus_min_count = ((max_cpus as f64) * cpus_frac_min).floor() as usize;
-            let cpus_max_count = ((max_cpus as f64) * cpus_frac_max).floor() as usize;
+            let cpus_min_count = ((max_cpus as f64) * cpus_frac_min).round_ties_even() as usize;
+            let cpus_max_count = ((max_cpus as f64) * cpus_frac_max).round_ties_even() as usize;
             Ok((
                 std::cmp::max(cpus_min_count, 1),
                 std::cmp::min(cpus_max_count, max_cpus),
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -104,6 +104,7 @@ lazy_static! {
                 kind: LayerKind::Confined {
                     util_range: (0.8, 0.9),
                     cpus_range: Some((0, 16)),
+                    cpus_pct: None,
                     protected: false,
                     common: LayerCommon {
                         min_exec_us: 1000,
@@ -166,6 +167,7 @@ lazy_static! {
                     cpus_range: None,
                     util_range: (0.2, 0.8),
                     protected: false,
+                    cpus_pct: None,
                     common: LayerCommon {
                         min_exec_us: 800,
                         yield_ignore: 0.0,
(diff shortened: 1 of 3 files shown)


hodgesds:
Is it valid to have two layers with `cpus_pct` set to 100?

likewhatevs:
> Is it valid to have two layers with cpus_pct set to 100? I guess it's not much different than the existing behavior, so maybe this makes sense.

that's the thinking -- what we have now works well enough but isn't portable across CPUs w/ different core counts, this makes it portable across CPUs w/ different core counts.

hodgesds (approved):

htejun:
Can we just allow "NUMBER%" for `cpus_range` instead of adding this as a separate single number config?

htejun:
Can we then do `cpus_range_pct` and error out if both `cpus_range` and `cpus_range_pct` are specified? That'd be useful in dynamic cases as well.

likewhatevs:
moved cpus_pct to cpus_range_pct.

this shows the pct working when `min=max`, when `min!=max`, and cpus_range still working like it does:
image

checked against main because `cpus_range: [5, 5]` leads to 6 cpus which sounds not-right on PR but main does the same atm:
image

scheds/rust/scx_layered/src/main.rs
```diff
@@ -986,6 +993,39 @@ struct Layer {
     allowed_cpus: Cpumask,
 }
 
+fn resolve_cpus_pct_range(
+    cpus_range: &Option<(usize, usize)>,
+    cpus_range_pct: &Option<(usize, usize)>,
+    max_cpus: usize,
+) -> Result<(usize, usize)> {
+    match cpus_range {
```
htejun, line 1001: Doing `match (cpus_range, cpus_range_pct)` will probably make the code simpler.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -986,6 +993,39 @@ struct Layer {
...
+        Some(cpus_range) => match cpus_range_pct {
+            Some(_cpus_pct) => {
+                bail!("cpus_range cannot be used with cpus_pct.");
+            }
+            None => Ok(*cpus_range),
+        },
+        None => match cpus_range_pct {
+            Some(cpus_range_pct) => {
```
htejun, line 1009: Given that `util_range` is in [0, 1.0], I wonder whether it'd be better to use fraction instead of percentage. ie. `cpus_range_frac` instead of `cpus_range_pct`.

likewhatevs: For consistency w/ util_range, this makes sense. 

I generally don't like floats in cfg files (probably got burned by deserialization/serialization of one forever ago or something), but consistency is good and int for one but float for the other isn't consistent.

scheds/rust/scx_layered/src/main.rs
likewhatevs: never thought to try that, thx.

likewhatevs:
change `cpus_range_pct` to `cpus_range_frac` and make rounding do what seems most reasonable (when tie, pick nearest even).
image

htejun (approved):

PR #1501, opened 2025-03-12
ci: pin kernel revisions

JakeHillion:
Currently we use unpinned kernels. We name a repo and branch and pull whatever is at the head of that at the time the GitHub workflow runs. This changeset proposes and implements locking kernel revision.

Advantages of pinning:
- Disambiguates between a commit that failed a GitHub run becuase of a code change and the kernel updating.
- Breakages in upstream kernels won't land in the repo until it is fixed, meaning people can keep developing.
- No race condition between the `build-kernel` CI job and the `restore-kernel-cache` CI steps. Currently they race because each checks the head cache individually, pinned they are guaranteed to use the same hash.
- Reduced reliance on git.kernel.org. Recently we saw failures because we query the head cache 10-20 times per PR. With this change we query git.kernel.org 0 times on a cache hit, which is very nearly always.
- Local reproducibility is improved (possible in follow up commits).

Disadvantages of pinning:
- There is more noise in the repo. A bot will occasionally commit updating the lock.
- Slight delay in picking up kernel changes. The `update-kernels` workflow needs to run, and if we run it too frequently there will be a lot of noise in the repo.
- With this setup if every branch has changed several CI runs will arrive at once, causing some unpleasant queuing, and undoing some spreading out of runs I made a while ago. I can't find a nice way to do this without spreading out the workflows. Mitigated by choosing the time of day with fewest user scheduled action runs.

This commit:
- Adds a script to update the kernel locks with `nix run ./.github/workflows#update-kernels`. This script is easily extensible if we need more information in the future.
- Adds a GitHub Action to automatically run this script, commit changes, and start PRs with auto merge for kernel updates.
- Uses the lock file in `build-kernel.yml` and the `restore-kernel-cache` action.
- Drops the scheduled run for `sched_ext/for-next`. This is now covered by the PR from the lock update or the PR with the scheduler change.

This commit DOES NOT:
- Test the `bpf/bpf-next` and `stable/linux-rolling-stable` versions before merging them, that happens at an unrelated time with the existing workflows. It does solve the polling git.kernel.org issue for them and is left in for consistency/to enable future work on making these push blocking where appropriate.

Test plan:
- Created a GitHub repository JakeHillion/scx-ci-staging. This can't be setup quite the same as the main repo but it successfully receives PRs from the bot, runs the actions, and auto merges when action conditions pass.
- Created a fork of Linux and broke the build. This change does not get pulled in.

I expect there to be a bit of noise landing this in the main repo, particularly around getting the `GH_TOKEN_FOR_UPDATES` secret right. This is necessary because standard actions can't trigger other actions on PRs, and we need a PR to run the merge queue (the whole point of this). The actual kernel builds are well tested.

[1] https://github.com/DeterminateSystems/update-flake-lock?tab=readme-ov-file#running-github-actions-ci

### DIFF
--- a/.github/actions/restore-kernel-cache/action.yml
+++ b/.github/actions/restore-kernel-cache/action.yml
@@ -1,10 +1,7 @@
 name: 'Restore kernel cache'
 description: 'Restore a previously built kernel cache, failing if not available'
 inputs:
-  git-repo:
-    required: true
-    type: string
-  branch:
+  repo-name:
     required: true
     type: string
 
@@ -12,7 +9,7 @@ runs:
   using: "composite"
   steps:
     - name: Get hash from repo/branch
-      run: echo "SCHED_EXT_KERNEL_COMMIT=$(git ls-remote ${{ inputs.git-repo }} heads/${{ inputs.branch }} | awk '{print $1}')" >> $GITHUB_ENV
+      run: echo "SCHED_EXT_KERNEL_COMMIT=$(jq -r '."${{ inputs.repo-name }}".commitHash' kernel-versions.json)" >> $GITHUB_ENV
       shell: bash
 
     - name: Load cache
--- a/.github/workflows/bpf-next-test.yml
+++ b/.github/workflows/bpf-next-test.yml
@@ -9,8 +9,7 @@ jobs:
   build-kernel:
     uses: ./.github/workflows/build-kernel.yml
     with:
-      git-repo: "https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git"
-      branch: for-next
+      repo-name: bpf/bpf-next
 
   integration-test:
     runs-on: ubuntu-24.04
@@ -24,8 +23,7 @@ jobs:
 
       - uses: ./.github/actions/restore-kernel-cache
         with:
-          git-repo: "https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git"
-          branch: for-next
+          repo-name: bpf/bpf-next
 
       # prevent cache permission errors
       - run: sudo chown root /usr/bin/tar && sudo chmod u+s /usr/bin/tar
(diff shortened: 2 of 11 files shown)


JakeHillion:
Note for reviewers: I'm not planning to land this until Monday as it'll likely need to be babysat a little. The kernels will continue to build but may be slightly out of date while the updater workflow settles. Will manually update them before landing and keep an eye on it.

likewhatevs:
> Breakages in upstream kernels won't land in the repo until it is fixed, meaning people can keep developing.

Idk -- I kind of like upstream kernel breakages being in your face. It's more important those are fixed ASAP than CI runs smoothly for scheduler updates tbh IMO. Also like, this presumes upstream breaking downstream means that upstream is wrong (and complicates+delays fixing downstream when downstream is wrong).

> No race condition between the build-kernel CI job and the restore-kernel-cache CI steps. Currently they race because each checks the head cache individually, pinned they are guaranteed to use the same hash.

This replaces a CI race condition with a bunch of update commits and PR noise (noise like https://github.com/likewhatevs/scx/pulls?q=is%3Apr+is%3Aclosed, which I turned off immediately, because it was annoying). I prefer the race condition.

> Reduced reliance on git.kernel.org. Recently we saw failures because we query the head cache 10-20 times per PR. With this change we query git.kernel.org 0 times on a cache hit, which is very nearly always.

I think there's a better fix for this. This issue is new, it wasn't present on the older CI config somehow, and IIUC that ran more jobs due to it de-duplicating less work or something.

> Local reproducibility is improved (possible in follow up commits).

Strong disagree on this point. CI configs are now using a bunch of GH action features/nix stuff and are no longer simple shell scripts that are (mostly) copy-pasteable onto any rando ubuntu box. I don't think reproducibility is a feature if leveraging it is sufficiently impractical that most folks aren't ever gonna leverage it (which IMO things are now).

hodgesds (approved):

Issue #1502, opened 2025-03-13
scx_lavd Error: No debugfs mount found


anh0516:
I'm looking at `tracepoint_exists()` in `scx_utils`:

`let base_path = tracefs_mount().unwrap_or(debugfs_mount()?);`

It looks like it should be choosing either tracefs or debugfs, whichever succeeds first. I do have tracefs mounted, and that works. But it's still complaining that it can't mount debugfs. I have little familiarity with Rust syntax and I wasn't able to figure out what's going on. Unless I'm completely misunderstanding this, and it's actually checking for both and handling failures for both in one line?

Is debugfs currently used at all or just for tracefs potentially being mounted under it, in which case `tracefs_mount()` should find it anyways?

I was able to get scx_lavd to run again by changing the line to:

`let base_path = tracefs_mount()?;`

and just bypassing checking for a debugfs mount entirely.

multics69:
Thank you @anh0516 for reporting the problem.
I think https://github.com/sched-ext/scx/pull/1500 should fix the problem.
Could you please check if it works on your side?

anh0516:
Forgot to mention that I was building from main. I have already pulled that commit, and it did not fix the problem.

multics69:
Oh, got it.

multics69:
@anh0516 -- Could you post the output of `cat /proc/mounts`?

anh0516:
> [@anh0516](https://github.com/anh0516) -- Could you post the output of `cat /proc/mounts`?

```
/dev/nvme0n1p1 / btrfs rw,noatime,nodiratime,compress-force=zstd:3,ssd,discard=async,space_cache=v2,subvolid=5,subvol=/ 0 0
devtmpfs /dev devtmpfs rw,nosuid,noexec,relatime,size=8088796k,nr_inodes=2022199,mode=755,inode64 0 0
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /dev/shm tmpfs rw,nosuid,nodev,inode64 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000 0 0
tmpfs /run tmpfs rw,nosuid,nodev,size=3236176k,nr_inodes=819200,mode=755,inode64 0 0
[...]

PR #1506, opened 2025-03-13
scx_utils: Fix tracepoint_exists()

multics69:
In tracepoint_exists(), unwrap_or() was used to get a result of either tracefs_mount() or debugfs_mount(). However, in Rust, unwrap_or() is eager so both tracefs_mount() and debugfs_mount() are evaluated. Hence, if one of them does not exist, an error is raised.

So, we replace unwrap_or() with unwrap_or_else(), which conducts lazy evaluation, meaning that if tracefs_mount() succeeds, debugfs_mount() is never evaluated.

Also, added "/tracing" to the debugfs mount point since the debugfs mount point is "/sys/kernel/debug" not "/sys/kernel/debug/tracing".

This should fix the following issue:
  https://github.com/sched-ext/scx/issues/1502

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -211,7 +211,7 @@ pub fn debugfs_mount() -> Result<std::path::PathBuf> {
 }
 
 pub fn tracepoint_exists(tracepoint: &str) -> Result<bool> {
-    let base_path = tracefs_mount().unwrap_or(debugfs_mount()?);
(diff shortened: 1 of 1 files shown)


hodgesds (approved):

JohnRTitor:
@multics69 Let me test before you guys merge this.

JohnRTitor (approved):
Nice, works!

Applied 42a9c0c694579269adc1170e46afdc79b302cfd6 (this commit) on top of 00e4551ef3f29c74d4fe98f8d99e90ce4a5244f0.

PR #1514, opened 2025-03-15
include/scx: turn arena spinlocks into common headers

etsal:
Move the arena spinlocks imported into WD40 to the common headers directory, so that they can be used by other schedulers. This amounts to updating the existing common definitions for the cond_break BPF macro, and renaming the new files' READ_ONCE/WRITE_ONCE into READ_VOLATILE/WRITE_VOLATILE to avoid a naming conflict with the already present macros of the same name.


kkdwivedi:
> READ_ONCE/WRITE_ONCE into READ_VOLATILE/WRITE_VOLATILE

Why not simply drop the conflicting redefinition and just use READ_ONCE/WRITE_ONCE everywhere instead of renaming?

etsal:
> > READ_ONCE/WRITE_ONCE into READ_VOLATILE/WRITE_VOLATILE
> 
> Why not simply drop the conflicting redefinition and just use READ_ONCE/WRITE_ONCE everywhere instead of renaming?

I wasn't sure if the two macros were interchangeable, and READ_ONCE/WRITE_ONCE are already being used in other schedulers. I assume from this the existing definitions can safely replace the ones in the arena code?

etsal:
We have 3 kinds of macros right now:

```
#define READ_ONCE(x)                                    \
({                                                      \
        union { typeof(x) __val; char __c[1]; } __u =   \
                { .__c = { 0 } };                       \
        __read_once_size(&(x), __u.__c, sizeof(x));     \
        __u.__val;                                      \
})

#define READ_ONCE_ARENA(type, x)                                \
({                                                              \
        union { type __val; char __c[1]; } __u =                \
                { .__c = { 0 } };                               \
        __read_once_size((void *)&(x), __u.__c, sizeof(x));     \
        __u.__val;                                              \
})

#define READ_VOLATILE(x) (*(volatile typeof(x) *)&(x))
```

The READ_ONCE macro doesn't take arena-qualified variables. For this reason we introduced READ_ONCE_ARENA that explicitly gets passed the type of the variable without the address space qualifier. Using the _ONCE_ARENA macros however in bpf_atomic.h would require changing the macros throughout bpf_atomic.h to require a type argument which is a no-go imo.  A "strip all attributes/qualifiers from type" compiler directive would solve this problem, but IIRC from our original discussions when we introduced the _ONCE_ARENA macros there is nothing like that available right now. 

Replacing the existing macros with the simpler "just typecast to volatile" versions from bpf_atomic.h would solve the problem, but I am not sure if they have identical behavior. Reading the macros I'd say they do, so unless there is some edge case I'm not seeing I will replace the existing macros with the simpler versions from the new headers.

likewhatevs (approved):
I think the approach of migrate wd40 arena library code to common, but in such a way that other schedulers are not touched, makes sense. 

Misc btw if possible/practical it'd be great if wd40 had a CI job.

kkdwivedi:
@etsal They do have identical behavior, but I would keep the existing ones scx has which are from the kernel. The reason you needed READ_ONCE_ARENA (with an extra type arg) is because __arena tag on scalar local variables isn't allowed. But consider you __unqual_typeof from the bpf_atomic.h header to get the typeof stripped of qualifiers in case of scalars, and the right thing in case of pointers. With that I'm sure we can remove the three versions and pick one.

etsal:
> @etsal They do have identical behavior, but I would keep the existing ones scx has which are from the kernel. The reason you needed READ_ONCE_ARENA (with an extra type arg) is because __arena tag on scalar local variables isn't allowed. But consider using __unqual_typeof from the bpf_atomic.h header to get the typeof stripped of qualifiers in case of scalars, and the right thing in case of pointers. With that I'm sure we can remove the three versions and pick one.
> 
> As you can see in that header, we can accept arena pointers and don't need to pass the extra type for other atomic operations.

Using __unqual_typeof is exactly the kind of macro I was referring to in my explanation and solves the problem. I have embedded it in our existing definitions in the original macro, so now we have a single pair that works with everything.

kkdwivedi (approved):

multics69 (approved):

PR #1518, opened 2025-03-16
scx_layered: make nodespread account for smt

likewhatevs:
node spread wasn't accounting for smt order/topo beyond "spread across numa nodes". fix this to improve locality and make `exclusive` flag work better (err, if not better, make how `NodeSpread + exclusive` work make sense w/ how folks would reason about them).

### core order
Screenshot from 2025-03-16 13-51-16
### likwid topo
Screenshot from 2025-03-16 13-51-25

### core order
Screenshot from 2025-03-16 13-50-28
### likwid topo
Screenshot from 2025-03-16 13-50-55

### DIFF
--- a/scheds/rust/scx_layered/src/layer_core_growth.rs
+++ b/scheds/rust/scx_layered/src/layer_core_growth.rs
@@ -39,9 +39,11 @@ pub enum LayerGrowthAlgo {
     /// LittleBig attempts to first grow across all little cores and then
     /// allocates onto big cores after all little cores are allocated.
     LittleBig,
-    /// Grab CPUs from NUMA nodes, iteratively, in linear order.
+    /// Grab CPUs from NUMA nodes, iteratively, in linear order. Not for use
+    /// on CPUs with heterogenous SMT.
     NodeSpread,
-    /// Grab CPUs from NUMA nodes, iteratively, in reverse order.
+    /// Grab CPUs from NUMA nodes, iteratively, in reverse order. Not for use
+    /// on CPUs with heterogenous SMT.
     NodeSpreadReverse,
     /// RandomTopo is sticky to NUMA nodes/LLCs but randomises the order in which
     /// it visits each. The layer will select a random NUMA node, then a random LLC
@@ -246,7 +248,19 @@ impl<'a> LayerCoreOrderGenerator<'a> {
             .nodes
             .clone()
             .into_iter()
-            .map(|(_, x)| x.all_cpus.into_iter().map(|(_, x)| x.id).collect())
+            .map(|(_, x)| {
+                x.all_cores
+                    .clone()
+                    .into_iter()
+                    .flat_map(|(_, x)| {
+                        if self.topo.all_cores.len() * 2 == self.topo.all_cpus.len() {
+                            vec![x.id, x.id + self.topo.all_cores.len()]
+                        } else {
+                            vec![x.id]
+                        }
+                    })
+                    .collect()
+            })
             .collect();
 
         for i in 0..core_vecs[0].len() {


likewhatevs:
Tests still pass like above for 9950x/epyc dual socket. Per spec sheets, this would work for p/e core xeon's also. Updated docs to note to not use this for mixed-smt systems (i.e. p core e core where p core has smt, e core doesn't, which afaict are only single socket only consumer chips).

likewhatevs:
Can't test w/ smt disabled via grub because `nosmt` causes the following:

```
❯ sudo ./target/release/scx_layered f:/home/patso/gb.json
22:08:48 [INFO] Topology awareness not specified, selecting enabled based on hardware
22:08:48 [INFO] CPUs: online/possible=16/32 nr_cores=16
22:08:48 [INFO] Kernel does not support queued wakeup optimization
thread 'main' panicked at scheds/rust/scx_layered/src/main.rs:1560:52:
called `Option::unwrap()` on a `None` value
```

Can't test on 258V machine (which has no smt) because some mix of idk not-super-standard kconfig's to make that a bit snappier cause the following w/ scx (unrelated to PR):
```
22:11:27 [WARN] libbpf: prog 'refresh_layer_cpumasks': BPF program load failed: Permission denied

22:11:27 [WARN] libbpf: prog 'refresh_layer_cpumasks': -- BEGIN PROG LOAD LOG --
0: R1=ctx() R10=fp0
; bpf_for(id, 0, nr_layers) @ main.bpf.c:461
0: (18) r7 = 0xffffc90007c05008       ; R7_w=map_value(map=bpf_bpf.rodata,ks=4,vs=11400,off=8)
2: (61) r3 = *(u32 *)(r7 +0)          ; R3_w=2 R7_w=map_value(map=bpf_bpf.rodata,ks=4,vs=11400,off=8)
3: (bf) r6 = r10                      ; R6_w=fp0 R10=fp0
4: (07) r6 += -32                     ; R6_w=fp-32
5: (bf) r1 = r6                       ; R1_w=fp-32 R6_w=fp-32
6: (b4) w2 = 0                        ; R2_w=0
7: (85) call bpf_iter_num_new#100864
calling kernel function bpf_iter_num_new is not allowed
processed 7 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
-- END PROG LOAD LOG --

22:11:27 [WARN] libbpf: prog 'refresh_layer_cpumasks': failed to load: -13

22:11:27 [WARN] libbpf: failed to load object 'bpf_bpf'

22:11:27 [WARN] libbpf: failed to load BPF skeleton 'bpf_bpf': -13

Caused by:
    Permission denied (os error 13)
```

scheds/rust/scx_layered/src/layer_core_growth.rs
```diff
@@ -246,7 +248,19 @@ impl<'a> LayerCoreOrderGenerator<'a> {
             .nodes
             .clone()
             .into_iter()
-            .map(|(_, x)| x.all_cpus.into_iter().map(|(_, x)| x.id).collect())
+            .map(|(_, x)| {
+                x.all_cores
+                    .clone()
+                    .into_iter()
+                    .flat_map(|(_, x)| {
+                        if self.topo.all_cores.len() * 2 == self.topo.all_cpus.len() {
```
hodgesds, line 256: Use [`self.topo.smt_enabled`](https://github.com/sched-ext/scx/blob/main/rust/scx_utils/src/topology.rs#L188) for a more robust check.

likewhatevs: That would fall over in (I think) a non-graceful manner on intel P/E core consumer chips (i.e. putting CPUs that don't exist in the vec).

scheds/rust/scx_layered/src/layer_core_growth.rs
```diff
@@ -246,7 +248,19 @@ impl<'a> LayerCoreOrderGenerator<'a> {
...
+                            vec![x.id, x.id + self.topo.all_cores.len()]
```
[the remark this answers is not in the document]
likewhatevs: ahh will fix that, thx!

PR #1522, opened 2025-03-17
scx_layered: cleanup node spread

likewhatevs:
make node spread determine cpu order in a less-fragile manner (i.e. use topo instead of shifting by number of cpus)

### layered node spread order
Screenshot from 2025-03-17 12-54-25

### likwid topo
Screenshot from 2025-03-17 12-54-38

### layered node spread order
Screenshot from 2025-03-17 12-56-56

### likwid topo
Screenshot from 2025-03-17 12-57-08

### DIFF
--- a/scheds/rust/scx_layered/src/layer_core_growth.rs
+++ b/scheds/rust/scx_layered/src/layer_core_growth.rs
@@ -246,17 +246,22 @@ impl<'a> LayerCoreOrderGenerator<'a> {
         let mut node_core_vecs: Vec<Vec<usize>> = Vec::new();
         let mut max_node_cpus: usize = 0;
 
-        for (_, node) in self.topo.nodes.iter() {
+        for (node_id, node) in self.topo.nodes.iter() {
             let mut node_vec = Vec::new();
-            for (_, llc) in node.llcs.iter() {
+            for (llc_id, llc) in node.llcs.iter() {
+                let mut llc_vec = Vec::new();
                 for (core_id, _) in llc.cores.iter() {
-                    node_vec.push(core_id.clone());
+                    llc_vec.push(core_id.clone());
                 }
+                llc_vec.sort();
+                log::debug!("LLC ID {} has cores {:?}", llc_id, llc_vec);
+                node_vec.push(llc_vec.clone());
             }
-            max_node_cpus = std::cmp::max(node_vec.len(), max_node_cpus);
-            // XXX this is a hack to maximize locality when core order is linear.
-            node_vec.sort();
-            node_core_vecs.push(node_vec.clone());
+            node_vec.sort_by(|x, y| (x[0]).cmp(&y[0]));
+            let flat_node_vec: Vec<usize> = node_vec.into_iter().flatten().collect();
+            log::debug!("Node ID {} has cores {:?}", node_id, flat_node_vec.clone());
+            max_node_cpus = std::cmp::max(flat_node_vec.len(), max_node_cpus);
+            node_core_vecs.push(flat_node_vec.clone());
         }
 
         for i in 0..=max_node_cpus {
--- a/scheds/rust/scx_layered/src/layer_core_growth.rs
+++ b/scheds/rust/scx_layered/src/layer_core_growth.rs
@@ -243,28 +243,24 @@ impl<'a> LayerCoreOrderGenerator<'a> {
 
     fn grow_node_spread(&self) -> Vec<usize> {
         let mut cores: Vec<usize> = Vec::new();
-        let core_vecs: Vec<Vec<usize>> = self
-            .topo
-            .nodes
-            .clone()
-            .into_iter()
-            .map(|(_, x)| {
-                x.all_cores
-                    .clone()
-                    .into_iter()
-                    .flat_map(|(_, x)| {
-                        if self.topo.all_cores.len() * 2 == self.topo.all_cpus.len() {
-                            vec![x.id, x.id + self.topo.all_cores.len()]
-                        } else {
-                            vec![x.id]
-                        }
-                    })
-                    .collect()
-            })
-            .collect();
+        let mut node_core_vecs: Vec<Vec<usize>> = Vec::new();
+        let mut max_node_cpus: usize = 0;
 
-        for i in 0..core_vecs[0].len() {
-            for sub_vec in core_vecs.clone() {
+        for (_, node) in self.topo.nodes.iter() {
+            let mut node_vec = Vec::new();
+            for (_, llc) in node.llcs.iter() {
+                for (core_id, _) in llc.cores.iter() {
+                    node_vec.push(core_id.clone());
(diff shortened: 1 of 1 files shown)


likewhatevs:
not sure why this isn't doing what I'd expect (nor if it's right or if what I expect is right).

scheds/rust/scx_layered/src/layer_core_growth.rs
```diff
@@ -243,28 +243,23 @@ impl<'a> LayerCoreOrderGenerator<'a> {
...
-                    })
-                    .collect()
-            })
-            .collect();
+        let mut cpus_vec: Vec<Vec<usize>> = Vec::new();
+
+        // XXX this probably does not work with nosmt.
+        // but layered doesn't work with that at all atm.
+
+        for (_, node) in self.topo.nodes.clone() {
```
htejun, line 251: Here and below, are `.clone()`s necessary? Can't you just `.iter()`?

likewhatevs:
ahh ok -- cache topo in likwid makes it clear how this should work:
image

scheds/rust/scx_layered/src/layer_core_growth.rs
```diff
@@ -243,28 +243,23 @@ impl<'a> LayerCoreOrderGenerator<'a> {
...
+            cpus_vec.push(node_vec);
+        }
 
-        for i in 0..core_vecs[0].len() {
-            for sub_vec in core_vecs.clone() {
+        for i in 0..cpus_vec[0].len() {
+            for sub_vec in cpus_vec.clone() {
                 if i < sub_vec.len() {
                     cores.push(sub_vec[i]);
                 }
```
htejun, line 265: Hmm... this looks complicated and still fragile (what if not all nodes have the same number of CPUs?). Wouldn't something like the following work?
```
    let mut cores: Vec<usize> = vec![];
    let mut iters:Vec<_> = self.topo.nodes.iter().map(|x| x.iter()).collect();
    loop {
        let mut had_some = false;
        for it in &mut iters {
            if let Some(v) = it.next() {
                cores.push(*v);
                had_some = true;
            }
        }
        if !had_some {
            break;
        }
    }
```

htejun: Oh, the above assumes that an iterator keeps returning `None` after returning `None` once which isn't guarnateed. Maybe just `.remove()` the iterator from the vec and replace `loop - if { break; }` with `while (iters.len())`?

htejun:
> ahh ok -- cache topo in likwid makes it clear how this should work: image

Right, the ordering should iterate llc by llc in each node to ensure that llc's are given out together.

likewhatevs:
I *think* this is right wrt/ topology now, but doesn't make a sense wrt/ ordering once idk topological distances are the same.

Should I hack sorting once distances are the same in here *or* open another PR updating topo to have it do that (i.e. have it sort cores by numerical ID once distances are accounted for/all else is equal?

### 7282 dual socket
Screenshot from 2025-03-17 13-55-12
### 9950
Screenshot from 2025-03-17 13-54-13

scheds/rust/scx_layered/src/layer_core_growth.rs
likewhatevs: I did this in a different but (I think) not fragile way.

htejun:
Hmm... topo should already have those sorted as they're all in `BTreeMap`s which are iterated in key order.

likewhatevs:
> Hmm... topo should already have those sorted as they're all in `BTreeMap`s which are iterated in key order.

Yeah I'm messing something up I think, tried sorting them and that did nothing.

likewhatevs:
Ok, I think I follow what this needs to do/it does it now:

9950x:
```
19:57:42 [DEBUG] layer: hhvmworkers algo: NodeSpread core order: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
```

7282 dual socket:
```
19:57:23 [DEBUG] layer: hhvmworkers algo: NodeSpread core order: [0, 16, 1, 17, 2, 18, 3, 19, 4, 20, 5, 21, 6, 22, 7, 23, 8, 24, 9, 25, 10, 26, 11, 27, 12, 28, 13, 29, 14, 30, 15, 31]
```

somewhere along the line I mixed up prox map/likwid output's and growth algo. growth algo runs on cores (not cpus like prox map/likwid output), so for this to be right it needs to not account for CPUs at all.

likewhatevs:
Hmm ok so this is really starting to really confuse me.

I added debug logs because I kept getting kinda non-sensical core ordering when like, trying to do things the right way.

These logs are printing what I'm getting back from topo:
```
20:50:03 [DEBUG] LLC ID 0 has cores [0, 1, 8, 9, 10, 11, 12, 13]
20:50:03 [DEBUG] LLC ID 1 has cores [2, 3, 4, 5, 6, 7, 14, 15]
20:50:03 [DEBUG] Node ID 0 has cores [0, 1, 8, 9, 10, 11, 12, 13, 2, 3, 4, 5, 6, 7, 14, 15]
```

This directly opposes what likwid says topology is.
image

likewhatevs:
Added debug logs to node spread which dump a key of node/llc/core/cpu mappings. I added this because I didn't understand how this core ordering works, and I'm reasonably confident most folks I've seen use layered misunderstand how this works also. 

Wrt/ this misunderstanding, everyone presumes that Linear/Reverse etc. in configs means "Linear/Reverse wrt/ CPUs (maybe w/ some weirdness around SMT)", not Linear/Reverse wrt/ numerical values of Core IDs, which AFAICT doesn't have any meaning outside of like layered/layer growth algo (i.e. vs the meaning that numerical CPU IDs have, for example)?

Per those logs, this does what it is intended to now. Also, make node spread use the rotation logic.

### core order from 9950x
Screenshot from 2025-03-17 21-18-06

### core order from dual socket
Screenshot from 2025-03-17 21-21-11

scheds/rust/scx_layered/src/layer_core_growth.rs
```diff
@@ -243,33 +243,44 @@ impl<'a> LayerCoreOrderGenerator<'a> {
...
+                                    core_id,
+                                    cpu_id
+                                );
+                            }
+                            core_id.clone()
+                        })
+                        .collect::<Vec<usize>>()
+                })
+                .collect();
+            max_node_cpus = std::cmp::max(flat_node_vec.len(), max_node_cpus);
```
etsal, line 272: Is it ever possible we find more CPUs than the node maximum? Why?

etsal (approved):
LGTM

hodgesds:
I think this works now here's a [gist](https://gist.github.com/hodgesds/bee44744206370021620ea57d63a89ab) with a bunch of different growth algos.

hodgesds (approved):

PR #1534, opened 2025-03-19
scxtop: Add checks when updating DSQ latency

hodgesds:
Add additional sanity checks for when DSQ latency is added in the TUI. DSQ latency of 0 should not be considered valid.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -2317,7 +2317,7 @@ impl<'a> App<'a> {
             self.max_cpu_events,
         ));
 
-        if *next_dsq_id != scx_enums.SCX_DSQ_INVALID {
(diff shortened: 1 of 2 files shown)


tools/scxtop/src/app.rs
```diff
@@ -2317,7 +2317,7 @@ impl<'a> App<'a> {
             self.max_cpu_events,
         ));
 
-        if *next_dsq_id != scx_enums.SCX_DSQ_INVALID {
+        if *next_dsq_id != scx_enums.SCX_DSQ_INVALID && *next_dsq_lat_us > 0 {
```
JakeHillion, line 2320: I wonder whether we could see an issue where the `/ 1000` makes this == 0 even though the floating point would be >? I think this is fine now because these events don't form a timeline of their own but something to watch out for maybe.

JakeHillion (approved):

PR #1536, opened 2025-03-20
scx_p2dq: Refactor pick_two_cpu idle selection

hodgesds:
Refactor pick_two_cpu when load balancing to first attempt to find an idle SMT and then an idle cpu. This should make slightly better load balancing decisions.


etsal (approved):
Makes sense!

PR #1539, opened 2025-03-20
scx_p2dq: Update enqueue CPU selection

hodgesds:
Update the enqueue path so that idle CPU selection attempts to select idle cores first.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -436,8 +436,7 @@ static s32 pick_idle_cpu(struct task_struct *p, struct task_ctx *taskc,
 				}
 				// Try an idle core in the LLC.
 				if (llcx->cpumask &&
-				    (cpu = scx_bpf_pick_idle_cpu(cast_mask(llcx->cpumask), SCX_PICK_IDLE_CORE)) &&
-				    cpu >= 0) {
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -643,19 +643,19 @@ void BPF_STRUCT_OPS(p2dq_enqueue, struct task_struct *p __arg_trusted, u64 enq_f
...
+	if (smt_enabled &&
+	    (cpu = scx_bpf_pick_idle_cpu(llc_mask, SCX_PICK_IDLE_CORE)) &&
+	    cpu >= 0) {
+		scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
+		return;
+	}
 
-	// If there is an idle CPU in the LLC kick it.
-	if (cpu < nr_cpus)
+	if ((cpu = scx_bpf_pick_idle_cpu(llc_mask, 0)) && cpu >= 0)
```
etsal, line 658: I think this will spuriously fail on CPU 0 selection.

hodgesds: Explain more, I was trying to debug seeing CPU 0 being left idle.... 😆

etsal: Oh, (cpu = scx_bpf_pick_idle_cpu(llc_mask, 0)) will evaluate to the selected CPU, which when zero so will count as a false-y value and cause the entire expression to evaluate to false. So the code then happily throws the selection away :P

etsal (approved):

PR #1542, opened 2025-03-20
scx_layered: add exclude match rule

etsal:
Add matching rules for excluding tasks/processes/cgroups with a given name from a layer. This lets us more finely specify task membership for a layer without overly depending on the order we specify the layers.

### DIFF
--- a/scheds/rust/scx_layered/examples/exclude.json
+++ b/scheds/rust/scx_layered/examples/exclude.json
@@ -1,0 +1,27 @@
+[
+	{
+		"name": "not stress-ng",
+		"matches": [
+			[{ "CommPrefixExclude": "stress-ng" }]
+		],
+		"kind": {
+			"Grouped": {
+				"cpus_range":  [5, 5],
+				"util_range": [0.01, 0.99],
+				"growth_algo": "RandomTopo",
+				"protected": true
+			}
+		}
+	},
+	{
+		"name": "stress-ng",
+		"matches": [
+			[]
+		],
+		"kind": {
+			"Open": {
+				"growth_algo": "RandomTopo"
+			}
+		}
+	}
+]
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -268,6 +268,7 @@ struct layer_match {
 	bool		is_kthread;
 	bool		used_gpu_tid;
 	bool		used_gpu_pid;
+	bool		exclude;
 };
 
 struct layer_match_ands {
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2044,7 +2044,7 @@ int match_layer(u32 layer_id, pid_t pid, const char *cgrp_path)
 				goto err;
 
 			match = &ands->matches[and_id];
-			if (!match_one(match, p, cgrp_path)) {
+			if (!(match_one(match, p, cgrp_path) == !match->exclude)) {
 				matched = false;
 				break;
 			}
--- a/scheds/rust/scx_layered/src/config.rs
+++ b/scheds/rust/scx_layered/src/config.rs
@@ -62,7 +62,9 @@ impl LayerSpec {
 pub enum LayerMatch {
     CgroupPrefix(String),
     CommPrefix(String),
+    CommPrefixExclude(String),
     PcommPrefix(String),
+    PcommPrefixExclude(String),
     NiceAbove(i32),
     NiceBelow(i32),
     NiceEquals(i32),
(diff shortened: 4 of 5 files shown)


likewhatevs (approved):
Seems reasonable. makes sense

htejun:
Can we generalize this? ie. Let all matchers take `Not` suffix or prefix and have a bool negate flag to `layer_match`?

likewhatevs:
> Can we generalize this? ie. Let all matchers take Not suffix or prefix and have a bool negate flag to layer_match?

I don't like the idea of adding a `Not` prefix (which could probably be done) nor adding a bool (which would need to be a breaking change w/o adding more polymorphic json and the requisite mess of handling with that). 

I think we should probably clean layered config up when we have a hard breaking change, making configs look kinda like this:

```
{
     "matcher_type": "comm_prefix",
     "polarity": true,
     "comm_prefix_matcher_properties": {
          "prefix": "prefix"
     }
}
```

Similar would apply to `layer.kind.matcher_type` and `layer.kind.matcher_type.growth_algo`.

hodgesds:
Maybe a dumb idea, but maybe have a `excludes` field that works similar as `matches`?

likewhatevs:
> Maybe a dumb idea, but maybe have a excludes field that works similar as matches? Then on the bpf side before doing any matches iterate over the set of excludes and bail early if it matches.

Not a bad idea if we want this now w/o needing to break cfg format, but I do think the idk overall of how cfg is handled  gets simpler if cfg entries are flat/simple objects like that. Using objects like that would enable polarity to be handled like how UsedGpuPid does it I think, which is pretty good/minimal doing of stuff.

etsal:
> Can we generalize this? ie. Let all matchers take `Not` suffix or prefix and have a bool negate flag to `layer_match`?

The main problem in doing this is how serdes deserializes the config: Right now the fields have to be the exact enum types we specify in Rust, so modifying any existing types breaks all existing configs. In particular I tried to get a drop-in replacement for Matches that would let us specify an exclude flag for each rule. I think @hodgesds 's exclude list idea is the most painless way to add general rule negation without risking breakage. The alternative would be to muck around with serdes to be able to decouple the JSON syntax from the Rust types.

etsal:
> > Maybe a dumb idea, but maybe have a excludes field that works similar as matches? Then on the bpf side before doing any matches iterate over the set of excludes and bail early if it matches.
> 
> Not a bad idea if we want this now w/o needing to break cfg format, but I do think the idk overall of how cfg is handled gets simpler if cfg entries are flat/simple objects like that. Using objects like that would enable polarity to be handled like how UsedGpuPid does it I think, which is pretty good/minimal doing of stuff.
> 
> This is what I mean -- polarity being a minimal tack on like this (and cfg enabling that) would be good:
> 
> ```
> return pid_present == match->used_gpu_pid;
> ```

This is how it looks internally and what the initial version of this patch was doing, but requires interfering with the serdes parsing to avoid breaking existing configs.

etsal:
Actually nevermind, I think doing generic negation requires no serdes code and s a matter of finding the right field attribute. I will look more into it.

etsal:
Ok, so negating individual terms in the expression without breaking things is deceivingly complex. 

> Maybe a dumb idea, but maybe have a `excludes` field that works similar as `matches`? Then on the bpf side before doing any matches iterate over the set of excludes and bail early if it matches.

I thought about it a bit more, and while that would work it does not give us the ability to negate individual terms in the sum of minterms we specify in matches. Also, it would be nonobvious whether exclude or match rules take precedence if both are applicable to a layer, and we would have to arbitrarily have one take precedence over the other.

htejun:
Maybe we can just add new negated match names and then handle them with a negate bit internally? That way serde sees all the negated matches as plain ones while we don't have to have duplicate code for all the matches that we have.

likewhatevs:
> Ok, unless I'm missing something there is no easy way to accept either a (LayerMatch, bool) JSON string or a LayerMatch instance with just serde attributes and without breaking current configs. I think we should merge the pull request as it is and revisit later. There should be a solution that involves writing extra parsing logic, but imo it is not worth it.

I don't think you're missing anything, the config format has structural issues that like make parsing/extending things a pain and fixing that w/o making the parsing logic worse means a breaking change.

etsal:
> Maybe we can just add new negated match names and then handle them with a negate bit internally? That way serde sees all the negated matches as plain ones while we don't have to have duplicate code for all the matches that we have.

On the BPF side that is exactly what we do rn, I have added a flag for each match target that specifies whether it is an allowlist or blocklist. The new "match types" are not visible from the BPF side.

etsal:
I have included negation rules for Comm, Pcomm, and Cgroup which are imo the most commonly used. Should I add for all of them?

hodgesds:
> The only one that I think would be needed is NSID/NSPID, should I add it?

Probably not necessary, those are for special cases anyways and if it's really needed we can add it later.

PR #1543, opened 2025-03-20
scxtop: trace on_sched_exec event.

devnexen:


tools/scxtop/src/perfetto_trace.rs
```diff
@@ -264,6 +264,28 @@ impl PerfettoTraceManager {
         });
     }
 
+    pub fn on_exec(&mut self, action: &ExecAction) {
```
hodgesds, line 267: This needs to be added to [`on_action`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/perfetto_trace.rs#L520) as well It's kind of clunky :disappointed:

hodgesds (approved):

PR #1545, opened 2025-03-21
scxtop: adding sched_move_numa trace event

devnexen:


devnexen:
odd, so it happens on fork event, can t reproduce, let s see with this change.

hodgesds:
It might be broke on `main`, let me do some more digging.

hodgesds (approved):

PR #1549, opened 2025-03-23
scx_p2dq: Refactor load balancing

hodgesds:
Refactor dispatch load balancing to into separate helper methods. When load balancing in the dispatch path take into account configuration values for load balancing parameters and use a DSQ iterator for selecting tasks to load balance. Refactor enqueue to use pick_idle_cpu rather than reimplementing idle CPU selection.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -128,7 +128,8 @@ static struct cpu_ctx *lookup_cpu_ctx(int cpu)
 }
 
 static __always_inline u64 cpu_dsq_id(int dsq_index, struct cpu_ctx *cpuc) {
-	if (dsq_index < 0 ||
+	if (!cpuc ||
+	    dsq_index < 0 ||
 	    dsq_index > nr_dsqs_per_llc ||
 	    dsq_index >= MAX_DSQS_PER_LLC) {
 		scx_bpf_error("cpuc invalid dsq index: %d", dsq_index);
@@ -240,7 +241,9 @@ static bool is_interactive(struct task_ctx *taskc)
 static bool can_pick2(struct task_ctx *taskc)
 {
 	if (is_interactive(taskc) ||
-	    (max_dsq_pick2 && taskc->dsq_id < nr_dsqs_per_llc))
+	    !taskc->all_cpus ||
+	    taskc->is_kworker ||
+	    (max_dsq_pick2 > 0 && taskc->llc_runs < min_llc_runs_pick2))
 		return false;
 
 	return true;
@@ -293,7 +296,7 @@ static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *left,
 	int i;
 
 	if (!left || !right)
-		return left;
+		return NULL;
 
 	bpf_for(i, 0, nr_llcs) {
 		if (i >= nr_dsqs_per_llc || i < 0)
@@ -312,10 +315,9 @@ static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *left,
 	     right_queued < min_nr_queued_pick2))
 		return NULL;
 
-	if (left_queued < right_queued || left_load > right_load)
+        if (left_queued < right_queued || left_load < right_load)
 		return most_loaded ? right : left;
-
-	return most_loaded ? right : left;
+	return most_loaded ? left : right;
 }
 
 static s32 pick_two_cpu(struct task_ctx *taskc, bool *is_idle)
@@ -325,10 +327,17 @@ static s32 pick_two_cpu(struct task_ctx *taskc, bool *is_idle)
 		return -EINVAL;
 
 	struct llc_ctx *chosen;
-	struct llc_ctx *left = rand_llc_ctx();
-	struct llc_ctx *right = rand_llc_ctx();
+	struct llc_ctx *left, *right;
 	int dsq_index = taskc->dsq_index;
 	s32 cpu;
+
+	if (nr_llcs == 2) {
+		left = lookup_llc_ctx(0);
+		right = lookup_llc_ctx(1);
+	} else {
+		left = rand_llc_ctx();
+		right = rand_llc_ctx();
+	}
 
 	if (!left || !right || dsq_index > nr_dsqs_per_llc) {
 		return -EINVAL;
@@ -353,9 +362,9 @@ static s32 pick_two_cpu(struct task_ctx *taskc, bool *is_idle)
 	     right_queued < min_nr_queued_pick2))
 		return -EINVAL;
 
-	if (((left_queued > 0 || right_queued > 0) &&
+	if (((left_queued > 0 || right_queued > 0) ||
 	    left_queued < right_queued) ||
-	    left_load > right_load) {
+	    left_load < right_load) {
 		chosen = left;
 		goto pick_llc;
 	} else {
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -432,7 +447,8 @@ static s32 pick_idle_cpu(struct task_struct *p, struct task_ctx *taskc,
 				}
 				// Try an idle core in the LLC.
 				if (llcx->cpumask &&
-				    (cpu = scx_bpf_pick_idle_cpu(cast_mask(llcx->cpumask), SCX_PICK_IDLE_CORE)) >= 0) {
+				    (cpu = scx_bpf_pick_idle_cpu(cast_mask(llcx->cpumask), SCX_PICK_IDLE_CORE)) >= 0 &&
+				    bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
```
multics69, line 451: I think you should test ` bpf_cpumask_test_cpu(cpu, p->cpus_ptr)` first. Otherwise, an idle CPU is booked then it turns out that idle CPU is not usable. Since there is no way to cancel the booking, it is better to check it first.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -444,15 +460,17 @@ static s32 pick_idle_cpu(struct task_struct *p, struct task_ctx *taskc,
 					goto out_put_cpumask;
 
 				if (cur_llcx->cpumask &&
-				    (cpu = scx_bpf_pick_idle_cpu(cast_mask(cur_llcx->cpumask), SCX_PICK_IDLE_CORE)) >= 0) {
+				    (cpu = scx_bpf_pick_idle_cpu(cast_mask(cur_llcx->cpumask), SCX_PICK_IDLE_CORE)) >= 0 &&
+				    bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
```
multics69, line 464: Same here.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -444,15 +460,17 @@ static s32 pick_idle_cpu(struct task_struct *p, struct task_ctx *taskc,
...
 					stat_inc(P2DQ_STAT_WAKE_MIG);
 					*is_idle = true;
 					goto out_put_cpumask;
 				}
 
 				// Couldn't find an idle core so just migrate to the CPU
 				if (cur_llcx->cpumask &&
-				    (cpu = scx_bpf_pick_idle_cpu(cast_mask(cur_llcx->cpumask), 0)) >= 0) {
+				    (cpu = scx_bpf_pick_idle_cpu(cast_mask(cur_llcx->cpumask), 0)) >= 0 &&
+				    bpf_cpumask_test_cpu(cpu, p->cpus_ptr)) {
```
multics69, line 473: Same here.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -552,6 +574,7 @@ s32 BPF_STRUCT_OPS(p2dq_select_cpu, struct task_struct *p, s32 prev_cpu, u64 wak
 		stat_inc(P2DQ_STAT_IDLE);
 		u64 slice_ns = dsq_time_slice(taskc->dsq_index);
 		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, slice_ns, 0);
+		scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
```
multics69, line 577: You don't need kick_cpu here since the sched_ext kernel will kick the CPU when the chosen CPU is idle.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -521,12 +541,14 @@ static s32 pick_idle_cpu(struct task_struct *p, struct task_ctx *taskc,
...
 	    !interactive &&
+	    nr_idle > 1 &&
 	    (cpu = pick_two_cpu(taskc, is_idle)) >= 0) {
 		stat_inc(P2DQ_STAT_SELECT_PICK2);
 		goto out_put_cpumask;
 	}
 
-	cpu = prev_cpu;
+	if (llcx->cpumask)
+		cpu = bpf_cpumask_any_distribute(cast_mask(llcx->cpumask));
```
multics69, line 551: It is not just related to this particular line. I think `pick_idle_cpu()` in general does not properly handle when a task is affinitized. For example, I think you should choose an idle or any CPU from `llcx->cpumask` ∩ `p->cpus_ptr` not just from  `llcx->cpumask`. Or is it implemented somewhere else already?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
arighi: I think this requires another tmp cpumask, since we want to pick an idle cpu in `p->cpus_ptr & llcx->cpumask`. Hopefully the new `scx_bpf_select_cpu_and()` API will make stuff like this much easier without having to introduce other tmp cpumasks.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
arighi: I'm wondering if we should explicitly forbid using `scx_bpf_kick_cpu()` from `ops.select_cpu()` in the sched_ext core. I don't see any valid reason why you may want to kick other CPUs than the one that is returned. Even in the proactive wakeup scenario, it's more effective to do that from `ops.enqueue()`, since we still need to go through `ops.enqueue()` if we don't dispatch directly here in `ops.select_cpu()`, so it's pretty much pointless to do a kick here.

multics69: I agree.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: Yeah, that's a good observation. I was trying to figure that pattern out last night but was a little too tired. Let me think this through again.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: Yeah, I added a tmp mask to the llc_ctx. Let me add those checks.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -398,19 +409,37 @@ static s32 pick_idle_cpu(struct task_struct *p, struct task_ctx *taskc,
...
+	    !llcx->cpumask || !llcx->tmp_cpumask)
+		goto found_cpu;
+
+	bpf_cpumask_and(llcx->tmp_cpumask, idle_cpumask, cast_mask(llcx->cpumask));
+	int nr_idle = bpf_cpumask_weight(idle_cpumask);
+	if (llcx->tmp_cpumask)
+		llc_idle = bpf_cpumask_weight(cast_mask(llcx->tmp_cpumask));
+
+
+	// Special handling of tasks with custom affinities
```
hodgesds, line 433: Going to punt on the affinitized tasks for now as when trying to add a test around all the different CPU selection spots it got clunky.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -293,7 +296,7 @@ static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *left,
 	int i;
 
 	if (!left || !right)
-		return left;
+		return NULL;
```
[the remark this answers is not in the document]
hodgesds: Yeah, thanks for catching. I was looking at this code as I wasn't sure it was working properly and it seems to be a bit bugged. I think what you propose make sense. At some point I want to have a separate function for doing the comparison and try load balancing on other values.

hodgesds:
I think this should be a bit better now with the affinitized task handling.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -325,11 +327,18 @@ static s32 pick_two_cpu(struct task_ctx *taskc, bool *is_idle)
...
 	int dsq_index = taskc->dsq_index;
 	s32 cpu;
 
+	if (nr_llcs == 2) {
+		left = lookup_llc_ctx(0);
+		right = lookup_llc_ctx(1);
+	} else {
+		left = rand_llc_ctx();
+		right = rand_llc_ctx();
+	}
```
JakeHillion, line 340: nice! should behave much better on dual socket intels now.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -390,6 +399,7 @@ static s32 pick_two_cpu(struct task_ctx *taskc, bool *is_idle)
 	return -EINVAL;
 }
 
+
```
JakeHillion, line 402: ultra-nit: unnecessary. is this consistent with the style?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -761,6 +793,91 @@ void BPF_STRUCT_OPS(p2dq_stopping, struct task_struct *p, bool runnable)
 	}
 }
 
+static bool dispatch_cpu(u64 dsq_id, s32 cpu, struct llc_ctx *llcx, int dsq_index)
```
JakeHillion, line 796: nit: do you want either `__always_inline` or `__weak` here to prevent the compiler being clever?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -761,6 +793,91 @@ void BPF_STRUCT_OPS(p2dq_stopping, struct task_struct *p, bool runnable)
...
+		stat_inc(P2DQ_STAT_DISPATCH_PICK2);
+		bpf_task_release(p);
+		break;
+	}
+
+	return dispatched;
+}
+
+
+static void dispatch_pick_two(s32 cpu)
```
JakeHillion, line 849: nit: same comment about __always_inline/weak

JakeHillion (approved):

PR #1553, opened 2025-03-23
scx_p2dq: Fix vtime accounting

hodgesds:
When a task stops running fix the fix time by adding the weighted vtime to the current vtime.

Reported-by: Changwoo Min <changwoo@igalia.com>

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -728,7 +728,7 @@ void BPF_STRUCT_OPS(p2dq_stopping, struct task_struct *p, bool runnable)
 	}
 
 	used = now - taskc->last_run_at;
-	p->scx.dsq_vtime = scale_by_task_weight(p, used);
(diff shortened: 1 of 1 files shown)


arighi (approved):

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -728,7 +728,7 @@ void BPF_STRUCT_OPS(p2dq_stopping, struct task_struct *p, bool runnable)
 	}
 
 	used = now - taskc->last_run_at;
-	p->scx.dsq_vtime = scale_by_task_weight(p, used);
+	p->scx.dsq_vtime += scale_by_task_weight(p, used);
```
multics69, line 731: I think `scale_by_task_weight_inverse()` should be used to correctly reflect priority.

multics69 (approved):

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: Yeah, that ended up working well when testing with various nice levels with stress-ng

PR #1555, opened 2025-03-24
add linux .editorconfig for c and related files

JakeHillion:
Add the `.editorconfig` used by Linux, only taking the section related to
C-like files. This lets contributors use their existing `vimrc` or whatever
other editorconfig while overriding the settings for consistent formatting in C
files. Often when you get a new contributor they insert tabs and don't view the
files with the correct indent size, this change fixes that in future.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+root = true
+
+[{*.{awk,c,dts,dtsi,dtso,h,mk,s,S},Kconfig,Makefile,Makefile.*}]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = tab
+indent_size = 8


likewhatevs:
I like this but doesn't this mean files get an auto-reformat on next edit?

JakeHillion:
> I like this but doesn't this mean files get an auto-reformat on next edit?

Should be fine. `charset` would probably change things, but I expect all of our files are ascii or utf-8 already. Same `insert_final_newline`, it'll add any trailing newlines when someone writes a file but that's fine. `tab_width` is view only with hard tabs, and `indent_style` only affects what happens when you press the tab key in my editor.

I've been using the tab parts of this config for a long time and it has never made any noise. Grabbed the rest for consistency with the kernel, as our style guide for C is effectively "use kernel style".

JakeHillion:
@htejun thoughts on this? It matches the file in `linux` so it should make sense for our C too, unless I'm missing anything.

htejun (approved):

htejun:
No real opinions. As long as it doesn't break things, I don't see a reason to not do it.

PR #1558, opened 2025-03-25
scx_bpfland: Enable SCX_OPS_ALLOW_QUEUED_WAKEUP

arighi:
Enable ttwu_queue optimization (SCX_OPS_ALLOW_QUEUED_WAKEUP) if the kernel supports it.

This optimization allows to process a task wakeup on the wakee's CPU (via IPI), saving some locking overhead on the waker's CPU, which can be expensive on large machines, due to cache and node boundaries.

This results in ops.select_cpu() being skipped when this path is taken; however, scx_bpfland explicitly checks whether ops.select_cpu() is skipped and still provides the task an opportunity to be directly dispatched from ops.enqueue(), so it is fair to enable this optimization.


multics69 (approved):

hodgesds (approved):
Nice, thanks for the descriptive message!

PR #1561, opened 2025-03-25
scx_utils: add missing traits to macros and avoid dependency on OpenBpfSkel

JakeHillion:
Two last changes to get these macros working in any context without additional
imports.
- Use `::anyhow::Result::Ok` to get type inference and not rely on `OpenBpfSkel`
  being named.
- Import the traits we rely on so they don't have to imported at every call site.

Test plan:
- They work with my new scheduler now without additional imports. Doesn't mean
  they're completely clean, but better than before.
- `cargo build`
- CI


hodgesds (approved):
LGTM

PR #1562, opened 2025-03-26
scx_bpfland: Introduce --idle-resume-us

arighi:
Introduce an option to configure the CPU idle QoS latency when the scheduler is loaded.

### DIFF
--- a/scheds/rust/scx_bpfland/src/main.rs
+++ b/scheds/rust/scx_bpfland/src/main.rs
@@ -32,6 +32,7 @@ use scx_stats::prelude::*;
 use scx_utils::autopower::{fetch_power_profile, PowerProfile};
 use scx_utils::build_id;
 use scx_utils::compat;
+use scx_utils::pm::{cpu_idle_resume_latency_supported, update_cpu_idle_resume_latency};
 use scx_utils::scx_ops_attach;
 use scx_utils::scx_ops_load;
 use scx_utils::scx_ops_open;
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -605,6 +628,19 @@ impl<'a> Scheduler<'a> {
 impl Drop for Scheduler<'_> {
     fn drop(&mut self) {
         info!("Unregister {} scheduler", SCHEDULER_NAME);
```
hodgesds, line 630: Good call on this, probably should have done it on the other schedulers

hodgesds (approved):
Nice! I haven't tested this API much, but it should be interesting.

scheds/rust/scx_bpfland/src/main.rs
arighi: Yeah, I was also considering to store the default `pm_qos_resume_latency_us` in struct Cpu to make sure we restore the previous value when the scheduler is unloaded. Maybe it can be done in a separate PR.

arighi:
> Nice! I haven't tested this API much, but it should be interesting.

Honestly, I haven't noticed any difference in my initial tests on my laptop. :smile: I'll test this on more architectures, probably it's more effective on certain machines and less noticeable on others.

scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -605,6 +628,19 @@ impl<'a> Scheduler<'a> {
...
+
+        // Restore default CPU idle QoS resume latency.
+        if self.opts.idle_resume_us >= 0 {
+            if cpu_idle_resume_latency_supported() {
+                let default_latency = 5000;
```
multics69, line 635: I am curious if 5000us is a real default value of CPU. In other words, can we revert the CPU state before setting idle resume latency with this?

scheds/rust/scx_bpfland/src/main.rs
multics69: BTW, in my systems (both Intel and AMD), `/sys/devices/system/cpu/cpu<N>/power/pm_qos_resume_latency_us` has 0 when reading it. Does 0 has a special meaning (e.g., not control the resume latency)?

multics69 (approved):

scheds/rust/scx_bpfland/src/main.rs
multics69: BTW, in my systems (both intel and amd), `power/pm_qos_resume_latency_us` has zero when reading it. Do 0 has a special meaning (not controlling the idle resume latency)?

arighi: I also noticed that on an AMD machine the default is 0, my laptop has 5000 by default so probably it's a good idea to read the value at the beginning, save it somewhere and restore the proper default...

PR #1563, opened 2025-03-26
scxtop: fix clock_sync_failure errors

tyroguru:
scxtop generated perfetto traces currently generate clock_sync_errors when you import them in the perfetto traceviewer. For example, a 5 second trace I generated generates the following error volume:

```
      {
        "name": "clock_sync_failure",
        "count": 6157712,
        "source": "SOURCE_ANALYSIS",
        "severity": "SEVERITY_ERROR"
      },
```

The fix given in this change makes everything happy and the data still looks good :-) . With this change another 5 second trace on the same machine gives:

```
      {
        "name": "clock_sync_failure",
        "count": 0,
        "source": "SOURCE_ANALYSIS",
        "severity": "SEVERITY_ERROR"
      },
```

To be honest, I'm not sure why this fixes the problem as I find the perfetto clock management code a bit opaque (currently anyway). I ripped this code off from the systing project (https://github.com/josefbacik/systing) .

All the timestamps for events look sane  when viewed in the perfetto viewer.


hodgesds:
On the bpf side I wonder if `bpf_ktime_get_boot_ns` should be used instead of `bpf_ktime_get_ns`. I rember maybe seeing something about that somewhere in the perfetto docs.

hodgesds (approved):
LGTM

PR #1564, opened 2025-03-26
scx_p2dq: Add support for SCX_OPS_ALLOW_QUEUED_WAKEUP

hodgesds:
Add support if enabled in the kernel for SCX_OPS_ALLOW_QUEUED_WAKEUP.

I believe this should be supported now handles when `select_cpu` isn't called ([code](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_p2dq/src/bpf/main.bpf.c#L663)).


likewhatevs (approved):
nice. this makes fps up and frame time flatter?

htejun (approved):

arighi (approved):

PR #1567, opened 2025-03-27
scx_bpfland: restore the correct value of pm_qos_resume_latency_us

arighi:
Restore the correct value of pm_qos_resume_latency_us when the scheduler is stopped.

### DIFF
--- a/scheds/rust/scx_bpfland/src/main.rs
+++ b/scheds/rust/scx_bpfland/src/main.rs
@@ -234,6 +234,7 @@ struct Scheduler<'a> {
     skel: BpfSkel<'a>,
     struct_ops: Option<libbpf_rs::Link>,
     opts: &'a Opts,
+    topo: Topology,
     power_profile: PowerProfile,
     stats_server: StatsServer<(), Metrics>,
     user_restart: bool,
@@ -343,6 +344,7 @@ impl<'a> Scheduler<'a> {
             skel,
             struct_ops,
             opts,
+            topo,
             power_profile,
             stats_server,
             user_restart: false,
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_bpfland/src/main.rs
```diff
@@ -632,12 +634,9 @@ impl Drop for Scheduler<'_> {
...
             if cpu_idle_resume_latency_supported() {
-                let default_latency = 5000;
-                info!("Restore idle QoS to {} us", default_latency);
-
-                let topo = Topology::new().unwrap();
-                for cpu in topo.all_cpus.values() {
-                    update_cpu_idle_resume_latency(cpu.id, default_latency).unwrap();
+                for cpu in self.topo.all_cpus.values() {
+                    update_cpu_idle_resume_latency(cpu.id, cpu.pm_qos_resume_latency_us as i32)
+                        .unwrap();
```
multics69, line 639: Overall, the change looks good to me.

I have a question. Is this code *guaranteed* to execute in all circumstances?
For example, what happens if the scheduler is forcefully killed via a signal.
I think, ideally, the restoring code should be guaranteed to be executed in all circumstances.

arighi: Yep, this code is executed also if you kill the scheduler via an external signal, excluding SIGKILL... but I don't think we can do much in this case.

hodgesds:
I was thinking in the future maybe we could have something like a `CpuContext`/`CoreContext` etc in topology that allows for mutating system state like this. Then when scheduler is dropped we have a more standardized way of getting the system back to the prior state.

scheds/rust/scx_bpfland/src/main.rs
multics69: Right, there is nothing that a userspace can do for SIGKILL. It would be a reason that bpf APIs is required *if* we think the idle resume latency API is critical.

multics69 (approved):

PR #1569, opened 2025-03-27
scxtop: adding cpu hotplug event on exit trace event.

devnexen:
refactoring existing cpu hotplug enter naming as well.


hodgesds (approved):
LGTM

PR #1574, opened 2025-03-28
scx_chaos: init with basic functionality

JakeHillion:
This commit initialises scx_chaos, a sched_ext scheduler designed for concurrency
testing programs. It is extremely lightly tested at the minute, but should implement
most of the behaviour of concurrency-fuzz-scheduler[0].

scx_chaos uses scx_p2dq as a base, and rather than copy-pasting code it makes
some changes to the structure to be able to use it directly. It turns out this
is pretty simple as we load the scheduler from the skeleton by name. It is then
pretty easy to wrap methods in chaos's BPF side while still maintaining all of
the necessary functionality from p2dq. The biggest change to p2dq comes in
enqueue, where the structure is changed to return a "promise". This is a C
structure detailing either what happened or how to do it, allowing chaos to
choose not to execute the enqueue and delay it until later, as well as allowing
chaos to run the enqueue with `scx_bpf_dsq_move` rather than `scx_bpf_dsq_insert`.

The concurrency testing functionality in this commit is pretty basic, and I
intend to extend it much further with a variety of different algorithms/inference
based techniques. The current method uniformly selects tasks for delay in `enqueue`,
and if selected enqueues them in a special vtime queue where vtime is wall time.
The delay is selected uniformly between a minimum and maximum. This is currently
only tested in dispatch, which appears never to stall on my machine but does
sometimes have very large delays - will be addressed in follow up patches.

[0] https://github.com/parttimenerd/concurrency-fuzz-scheduler

### DIFF
--- a/scheds/rust/scx_chaos/src/main.rs
+++ b/scheds/rust/scx_chaos/src/main.rs
@@ -1,0 +1,239 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+use scx_chaos::Builder;
+use scx_chaos::Scheduler;
+use scx_chaos::Trait;
+
+use scx_p2dq::SchedulerOpts as P2dqOpts;
+
+use anyhow::{anyhow, Context, Result};
+use clap::Parser;
+use log::info;
+
+use std::panic;
+use std::pin::Pin;
+use std::process::Command;
+use std::sync::Arc;
+use std::sync::Condvar;
+use std::sync::Mutex;
+use std::thread;
+use std::time::Duration;
+
+/// Randomly delay a process.
+#[derive(Debug, Parser)]
+pub struct RandomDelayArgs {
+    /// Chance of randomly delaying a process.
+    #[clap(long, requires = "random_delay_min_us")]
+    pub random_delay_frequency: Option<f64>,
+
+    /// Minimum time to add for random delay.
+    #[clap(long, requires = "random_delay_max_us")]
+    pub random_delay_min_us: Option<u64>,
+
+    /// Maximum time to add for random delay.
+    #[clap(long, requires = "random_delay_frequency")]
+    pub random_delay_max_us: Option<u64>,
+}
+
+/// scx_chaos: A general purpose sched_ext scheduler designed to amplify race conditions
+///
+/// WARNING: This scheduler is a very early alpha, and hasn't been production tested yet. The CLI
+/// in particular is likely very unstable and does not guarantee compatibility between versions.
+///
+/// scx_chaos is a general purpose scheduler designed to run apps with acceptable performance. It
+/// has a series of features designed to add latency in paths in an application. All control is
+/// through the CLI. Running without arguments will not attempt to introduce latency and can set a
+/// baseline for performance impact. The other command line arguments allow for specifying latency
+/// inducing behaviours which attempt to induce a crash.
+///
+/// Unlike most other schedulers, you can also run scx_chaos with a named target. For example:
+///     scx_chaos -- ./app_that_might_crash --arg1 --arg2
+/// In this mode the scheduler will automatically detach after the application exits, unless run
+/// with `--repeat-failure` where it will restart the application on failure.
+#[derive(Debug, Parser)]
+pub struct Args {
+    /// Whether to continue on failure of the command under test.
+    #[clap(long, action = clap::ArgAction::SetTrue, requires = "args")]
+    pub repeat_failure: bool,
+
+    /// Whether to continue on successful exit of the command under test.
+    #[clap(long, action = clap::ArgAction::SetTrue, requires = "args")]
+    pub repeat_success: bool,
+
+    /// Enable verbose output, including libbpf details. Specify multiple
+    /// times to increase verbosity.
+    #[clap(short = 'v', long, action = clap::ArgAction::Count)]
+    pub verbose: u8,
+
+    #[command(flatten, next_help_heading = "Random Delays")]
+    pub random_delay: RandomDelayArgs,
+
+    #[command(flatten, next_help_heading = "General Scheduling")]
+    pub p2dq: P2dqOpts,
+
+    /// Stop the scheduler if specified process terminates
+    #[arg(
+        long,
+        short = 'p',
+        help_heading = "Test Command",
+        conflicts_with = "args"
+    )]
+    pub pid: Option<u64>,
+
+    /// Program to run under the chaos scheduler
+    ///
+    /// Runs a program under test and tracks when it terminates, similar to most debuggers. Note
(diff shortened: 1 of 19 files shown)


likewhatevs:
This is cool, but p2dq has had some promising performance results, please do not regress them while editing it to support chaos (i.e. test that performance does not regress across this change).

Also, lib.rs complicates downstream usage w/o adding any value to p2dq so like, keep this in mind as we make changes like this.

@hodgesds

JakeHillion:
The p2dq failure is legit and after the next amend there will be one for chaos too. Interestingly they don't appear to happen with normal load/`stress-ng -c $(nproc)` or a smaller stresser, but do happen with the CI's:

    stress-ng -t 14 --aggressive -M -c `nproc` -f `nproc`

Haven't quite gotten to the bottom of it yet, so will leave this PR as a draft until I find the bug.

scheds/rust/scx_p2dq/src/bpf/intf.h
```diff
@@ -126,4 +126,31 @@ struct node_ctx {
...
 
-#endif /* __INTF_H */
+enum enqueue_promise_kind {
+	P2DQ_ENQUEUE_PROMISE_COMPLETE,
+	P2DQ_ENQUEUE_PROMISE_VTIME,
+	P2DQ_ENQUEUE_PROMISE_FIFO,
+};
+
+struct enqueue_promise_vtime {
+	u32	dsq;
```
hodgesds, line 136: Maybe name this `dsq_index` if it is the index in the `llc_ctx->dsqs`.

hodgesds:
For a rough perf test run with `scx_p2dq -s 1000 -x 1` and the perf should be roughly equal/better than eevdf.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -646,6 +662,8 @@ void BPF_STRUCT_OPS(p2dq_enqueue, struct task_struct *p __arg_trusted, u64 enq_f
 		stat_inc(P2DQ_STAT_DIRECT);
 		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, slice_ns,
 				   enq_flags | SCX_ENQ_PREEMPT);
+
+		ret->kind = P2DQ_ENQUEUE_PROMISE_COMPLETE;
```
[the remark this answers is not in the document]
JakeHillion: Thank you for looking! The theory here (though it's quite unclear) is that `async_p2dq_enqueue` returns a completable promise where appropriate, but that the per-CPU kthread case doesn't make sense to defer so does the enqueue immediately. That means that the kthread case does the enqueue and returns PROMISE_COMPLETE, while the other case below doesn't do the enqueue (I deleted the `_insert` call) and instead tells the completer how to complete it.

The bug that breaks p2dq & chaos is probably somewhere around here, though I've reviewed this code several times and really can't see it. Will keep debugging!

Does that description make sense? I should probably put it in a comment if it does.

JakeHillion:
> Also, lib.rs complicates downstream usage w/o adding any value to p2dq so like, keep this in mind as we make changes like this.

"Downstream" here means people using alternative build systems, so likely Meta? It doesn't complicate things when using Cargo, and it does add value to the scx ecosystem by allowing people to extend the scheduler to do cool things. I think this is a worthy tradeoff.

> For a rough perf test run with `scx_p2dq -s 1000 -x 1` and the perf should be roughly equal/better than eevdf.

Thanks, will perform some tests shortly. Marking this PR as ready for review as the code now seems good and CI passes.

hodgesds:
Some tests with `schbench`:

`scx_p2dq` main:
```
Wakeup Latencies percentiles (usec) runtime 30 (s) (228657 total samples)
          50.0th: 17         (69921 samples)
          90.0th: 781        (89437 samples)
        * 99.0th: 4584       (20563 samples)
          99.9th: 6376       (2048 samples)
          min=1, max=9132
Request Latencies percentiles (usec) runtime 30 (s) (229044 total samples)
          50.0th: 8944       (71752 samples)
          90.0th: 14768      (88664 samples)
        * 99.0th: 27488      (20538 samples)
          99.9th: 40000      (2040 samples)
          min=3888, max=70668
          20.0th: 7256       (7 samples)
        * 50.0th: 7672       (9 samples)
          90.0th: 8208       (12 samples)
          min=6612, max=8399
average rps: 7634.80
```
`scx_p2dq` PR:
```
Wakeup Latencies percentiles (usec) runtime 30 (s) (238912 total samples)
          50.0th: 17         (71770 samples)
          90.0th: 289        (94648 samples)
        * 99.0th: 2092       (21485 samples)
          99.9th: 4344       (2144 samples)
          min=1, max=7731
Request Latencies percentiles (usec) runtime 30 (s) (239467 total samples)
          50.0th: 9008       (73844 samples)
          90.0th: 13680      (93908 samples)
        * 99.0th: 22880      (21447 samples)
          99.9th: 30944      (2121 samples)
          min=4281, max=50764
          20.0th: 7784       (8 samples)
        * 50.0th: 8040       (9 samples)
          90.0th: 8168       (11 samples)
          min=7422, max=8332
average rps: 7982.23
```
Looks like a nice perf improvement!

likewhatevs:
> Looks like a nice perf improvement!

has me sold.

JakeHillion:
updated async_p2dq_enqueue comment and minimum rust version (still stable)

sirlucjan:
Compiled without issues, p2dq with PR works really fine.

JakeHillion:
> `--autopilot` doesn't seem to be implemented yet?

That's correct, did I mention it somewhere? I don't see it in the args. I think there needs to be some planning/discussion about how that's implemented, but I did structure the code in a way that's ready for it (mainly the iterator of different builders from the args).

scheds/rust/scx_chaos/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+
+/// scx_chaos: A general purpose sched_ext scheduler designed to amplify race conditions
+///
+/// WARNING: This scheduler is a very early alpha, and hasn't been production tested yet. The CLI
+/// in particular is likely very unstable and does not guarantee compatibility between versions.
+///
+/// scx_chaos is a general purpose scheduler designed to run apps with acceptable performance. It
+/// has a series of features designed to add latency in paths in an application. All control is
+/// through the CLI. Running without arguments will not attempt to introduce latency and can set a
+/// baseline for performance impact, while `--autopilot` will run a series of mixtures of latency
```
parttimenerd, line 47: You mentioned `--autopilot` here

JakeHillion:
Rebased, amended help message to remove reference to the currently unimplemented autopilot mode, and added the `--repeat-success` flag for apps under test with a bit of a refactor of that code.

parttimenerd (approved):

sirlucjan (approved):

PR #1575, opened 2025-03-28
scx_layered: Fix accounting, reporting and CPU allocation fluctuations

htejun:
- CPU usage time accounting could go wildly wrong on large machines leading to wide fluctuations on CPU allocations. This likely explains some of the reason why we are using `cpus_range` a lot more now. Fix it using `ops.tick()`.
- System-wide util was adding duplicate counters reporting usage number a lot higher than actual.
- Keep decisions wasn't being added to the total sched events which leads to weird percentage numbers for layers with a lot of long running tasks.


likewhatevs (approved):
cool use of tick

PR #1576, opened 2025-03-28
scx_layered: add same_over_idle

likewhatevs:
Add the layer property same_over_idle. When a layer has this property on it set to `true` and a machine has SMT enabled, tasks on this layer will be scheduled to use the same CPU even if that CPU's SMT sibling is not idle.

When this property is false, things will work as they do without this change. That is, tasks such as those in the example above would be scheduler to a CPU with an idle SMT sibling, if any such CPU exists.

the thing to note in these gif's the delta in how utilized CPUs are bouncing around:
disabled:
Kooha-2025-03-28-18-06-51-opt

enabled:
Kooha-2025-03-28-18-08-12-opt

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -751,7 +751,7 @@ static void maybe_refresh_layered_cpus_unprotected(struct task_struct *p, struct
 }
 
 static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
-			      const struct cpumask *idle_smtmask)
+			      const struct cpumask *idle_smtmask, const struct layer *layer)
 {
 	bool prev_in_cand;
 	s32 cpu;
@@ -765,7 +765,7 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
 	 * If CPU has SMT, any wholly idle CPU is likely a better pick than
 	 * partially idle @prev_cpu.
 	 */
-	if (smt_enabled) {
+	if (smt_enabled && !layer->same_over_idle) {
 		if (prev_in_cand &&
 		    bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
 		    scx_bpf_test_and_clear_cpu_idle(prev_cpu))
@@ -830,7 +830,7 @@ s32 pick_idle_big_little(struct layer *layer, struct task_ctx *taskc,
 		bpf_cpumask_and(tmp_cpumask, cast_mask(taskc->layered_mask),
 				cast_mask(big_cpumask));
 		cpu = pick_idle_cpu_from(cast_mask(tmp_cpumask),
-					 prev_cpu, idle_smtmask);
+					 prev_cpu, idle_smtmask, layer);
 		goto out_put;
 	}
 	case GROWTH_ALGO_LITTLE_BIG: {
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -765,7 +765,7 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
 	 * If CPU has SMT, any wholly idle CPU is likely a better pick than
 	 * partially idle @prev_cpu.
 	 */
-	if (smt_enabled) {
+	if (smt_enabled && !layer->same_over_idle) {
```
htejun, line 768: Can you please name it `prev_over_idle_core`? Also, what we probably want to do is trying `prev_cpu` before idle cores, instead of skipping idle cores altogether. ie:
```
if (prev_in_cand && layer->prev_over_idle_core &&
    scx_bpf_test_and_clear_cpu_idle(prev_cpu))
    return prev cpu;
if (smt_enabled) {
    // idle core block
}

if (prev_in_cand && !layer->prev_over_idle_core &&
     scx_bpf_test_and_clear_cpu_idle(...))
    return prev_cpu;
```

likewhatevs: > Also, what we probably want to do is trying prev_cpu before idle cores, instead of skipping idle cores altogether. ie

If I made that change right (which I think I did), that ends up bouncing things around more than the prior version, but less than disabled. That being said, I do think this approach makes more sense.

Kooha-2025-03-28-19-22-47-opt

valentinandrei: Screenshot 2025-03-28 at 7 26 25 PM

This is how I changed it to get the stickiness in the internal tests.

likewhatevs: Thanks! That means that this PR as it is now should have the effect we want.

The relevant part of this change is now (smt is enabled or this change is noop/there is no smt sibling to care about):
image

htejun:
Not a big difference but no need to do `scx_bpf_test_and_clear_cpu_idle(prev_cpu)` again if already performed in the `smt` block, right?

likewhatevs:
> Not a big difference but no need to do scx_bpf_test_and_clear_cpu_idle(prev_cpu) again if already performed in the smt block, right?

If you mean:
```
		// if layer is prev_over_idle_core, try prev
		if (prev_in_cand && layer->prev_over_idle_core &&
				scx_bpf_test_and_clear_cpu_idle(prev_cpu))
				return prev_cpu;

		// try prev if layers smt sibling is idle
		if (prev_in_cand &&
		    bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
		    scx_bpf_test_and_clear_cpu_idle(prev_cpu))
			return prev_cpu;
```
to 
```
                if (scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
    		       // if layer is prev_over_idle_core, try prev
    		        if (prev_in_cand && layer->prev_over_idle_core)
				 return prev_cpu;

		       // try prev if layers smt sibling is idle
		       if (prev_in_cand &&
		                 bpf_cpumask_test_cpu(prev_cpu, idle_smtmask))
			                 return prev_cpu;
                 }
```
I think that's right (i.e. not sure calling `scx_bpf_test_and_clear_cpu_idle` before `bpf_cpumask_test_cpu` like that is OK), and will open a PR w/ it.

likewhatevs:
opened #1593 applying that change

PR #1578, opened 2025-03-29
scx_utils: add fps module

likewhatevs:
This PR adds a module to utils which provides FPS to schedulers via gamescope or mangohud if logging is enabled. It provides the following data:
```
#[derive(Debug, Clone, Deserialize)]
pub struct SystemData {
    pub os: String,
    pub cpu: String,
    pub gpu: String,
    pub ram: u64,
    pub kernel: String,
    pub driver: String,
    pub cpu_scheduler: String,
}

#[derive(Debug, Clone, Deserialize)]
pub struct FrameData {
    pub fps: f64,
    pub frametime: f64,
    pub cpu_load: f64,
    pub cpu_power: f64,
    pub gpu_load: f64,
    pub cpu_temp: f64,
    pub gpu_temp: f64,
    pub gpu_core_clock: f64,
    pub gpu_mem_clock: f64,
    pub gpu_vram_used: f64,
    pub gpu_power: f64,
    pub ram_used: f64,
    pub swap_used: f64,
    pub process_rss: f64,
    pub elapsed: u64,
}
```
### Context

I spoke with @hodgesds recently about p2dq and it showing some promise wrt/ latency sensitive server workloads and gaming fps. We kinda came to the conclusion that like, maybe something cool could be done if we had a way to establish a feedback loop between application metrics and a scheduler. IDK how we would do that, but this provides some application metrics for that, maybe.

I looked into ways that we could make fps available to schedulers and this looked like the best one. It isn't great (you would have to shift-f2 toggle stats logging in either mangohud or gamescope to make this work), but it is something and I saw no other way to obtain this information that didn't involve intercepting graphics libraries.

The stats logging this relies upon looks lightweight enough that, if this itself is lightweight enough, with some extra tricks (i.e. guessing the right log file from timestamps), this could be usable enough to proof-of-concept something we could use to request stats like these be exposed in a better manner (like dbus or a socket or anything that isn't this, lol).

I added a decent-ish test to see this actually works and it seems to:

```
❯ cargo test --features all
...
test fps::tests::test_mangohud_tailer ... ok
...
```


likewhatevs:
@hodgesds LMK if you think this is worth experimenting with and I can re-open/re-visit.

PR #1579, opened 2025-03-29
scx_p2dq: Fix load calculation

hodgesds:
Load values should not be scaled by weight in the same way vtime is progressed for tasks/DSQs. Use absolute values for load metrics so that load balancing works appropriately.


arighi (approved):
Makes sense.

PR #1581, opened 2025-03-31
scx_lavd: Better control and handling of per-CPU tasks.

multics69:
Handling of per-CPU/non-migratable tasks was inconsistent. For non-migratable tasks, the sched_ext core bypassed the BPF scheduler and directly enqueued to the local DSQ with the default time slice. But for the pinned per-CPU tasks were forwarded to the BPF scheduler for enqueuing.

This PR addressed such inconsistent handling of per-CPU/non-migratable tasks. In particular, this PR contains the following changes:

1) Enable SCX_OPS_ENQ_MIGRATION_DISABLED, so the sched_ext core always forwards tasks to the BPF scheduler.
2) Add fast path for per-CPU/non-migratable tasks to ops.dispatch() and ops.enqueue().
3) Add a direct dispatch optimization at ops.enqueue() when the chosen CPU is idle, and there is nothing to do in the chosen domain.
4) As a fallback of the kernel does not support SCX_OPS_ENQ_MIGRATION_DISABLED, override the default time slice to the proper one at ops.running().
5) Clean up the flag setting code for readability.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -310,6 +310,9 @@ static u64 calc_weight_factor(struct task_struct *p, struct task_ctx *taskc)
 		weight_boost += LAVD_LC_WEIGHT_BOOST;
 	}
 
+	/*
+	 * Respect nice priority.
+	 */
 	weight_ft = p->scx.weight * weight_boost;
 	return weight_ft;
 }
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -549,6 +549,16 @@ static void update_stat_for_running(struct task_struct *p,
 	u64 wait_period, interval;
 
 	/*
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -545,6 +548,20 @@ static void update_stat_for_running(struct task_struct *p,
...
+	/*
+	 * If the sched_ext core directly dispatched a task, calculating the
+	 * task's deadline and time slice was also skipped. In this case, we
+	 * set the deadline and time slice here.
+	 *
+	 * Note that this is necessary when the kernel does not support
+	 * SCX_OPS_ENQ_MIGRATION_DISABLED or SCX_OPS_ENQ_MIGRATION_DISABLED
+	 * is not turned on.
+	 */
+	if (p->scx.slice == SCX_SLICE_DFL) {
```
arighi, line 560: Nice! I'm wondering if we should have a per-task flag to determine in a more reliable way when the sched_ext core dispatches a task, so that we can update the slice, deadline, etc. accordingly.

arighi (approved):

multics69:

Issue #1580, opened 2025-03-30
scx_lavd (task_ctx lookup failed)


dim-geo:
while updating my gentoo system:

```
⌂5.18 [root@gentoo:/home/none/scx/build/release] # ./scx_lavd --performance
08:32:17 [INFO] Kernel does not support queued wakeup optimization.
08:32:18 [INFO] scx_lavd scheduler is initialized (build ID: 1.0.10-g8f9ecadd x86_64-unknown-linux-gnu)
08:32:18 [INFO] scx_lavd scheduler starts running.

DEBUG DUMP
================================================================================

python[1058544] triggered exit kind 1025:
  scx_bpf_error (task_ctx lookup failed for python[1058371])

Backtrace:
  scx_bpf_error_bstr+0xf5/0x1a0
  bpf_prog_c05e4715bf23d31a_lavd_select_cpu+0x95/0x314
  bpf__sched_ext_ops_select_cpu+0x4b/0xaf
  select_task_rq_scx+0xab/0x1a0
  try_to_wake_up+0x143/0x980
  __wake_up_sync_key+0x58/0xa0
  do_notify_parent+0x1cd/0x2e0
  do_exit+0x876/0xb40
  do_group_exit+0x2c/0x80
  __x64_sys_exit_group+0x14/0x20
  x64_sys_call+0x714/0x720
  do_syscall_64+0x7e/0x160
  entry_SYSCALL_64_after_hwframe+0x4b/0x53

CPU states
----------

CPU 1   : nr_run=2 flags=0x1 cpu_rel=0 ops_qseq=68562855 pnt_seq=43573099
          curr=python[1058545] class=ext_sched_class

 *R python[1058545] +0ms
      scx_state/flags=3/0x5 dsq_flags=0x0 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=(n/a) dsq_vtime=249635791168 slice=4097015
      cpus=f

  R python[1058371] +0ms
      scx_state/flags=3/0x9 dsq_flags=0x0 ops_state/qseq=2/68562854
      sticky/holding_cpu=-1/-1 dsq_id=(n/a) dsq_vtime=249635351575 slice=4925512
      cpus=f

    kernel_wait4+0x9d/0x130
    do_syscall_64+0x7e/0x160
    entry_SYSCALL_64_after_hwframe+0x4b/0x53

[...]

PR #1582, opened 2025-03-31
scx_lavd: Remove try_get_task_ctx() and try_get_cpu_ctx().

multics69:
Previously, there were two versions of get_cpu/task_ctx(): with and without the 'try_' prefix. The version without 'try_' prefix performs the error check and calls scx_bpf_error(). On the other hand, the version with the prefix does not perform the error checking.

However, checking the error inside get_cpu/task_ctx() is not redundant because the caller should check the error to satisfy the verifier. Moreover, eagerly calls scx_bpf_error() (i.e., killing the BPF scheduler) would be improper in some cases.

Thus, remove the internal error checking and calling scx_bpf_error() inside get_cpu/task_ctx() and remove the 'try_' versions.

This should fix the following issue:
  https://github.com/sched-ext/scx/issues/1580


hodgesds (approved):
LGTM

arighi (approved):

PR #1584, opened 2025-03-31
scx_bpfland: Improve NUMA node locality

arighi:
With these changes, bpfland tries more aggressively to keep tasks running within the same node, prioritizing partially idle SMT cores before considering cross-node migrations.

Moreover, it tries more aggressively to perform migrations + direct dispatch attempts from ops.enqueue() to reduce pressure on the shared per-node DSQs.


multics69 (approved):

hodgesds (approved):
LGTM

PR #1586, opened 2025-03-31
lib: move cpumask into main library code

etsal:
The arena-based cpumask code in wd40 is portable to other schedulers. Move the code into the common lib/ directory so that it is easier to share.


likewhatevs (approved):
LGTM

PR #1587, opened 2025-03-31
scxtop: scx_utils: Correctly handle signed values in CPU topology files

metin-arm:
On some systems, the file `/sys/devices/system/cpu/cpu<N>/topology/cluster_id` may contain `-1` to indicate the lack of a defined cluster ID for a given CPU.

The existing code assumes unsigned integers, which causes parsing failures (e.g., `parse::<usize>()`) when reading such values.

Introduce a generic `read_from_file<T>()` function that supports any type implementing `FromStr`, including `isize`. Update the topology code to use `read_from_file::<isize>()` when reading `cluster_id` to correctly handle negative values.

While we are here, slightly improve the message in `bail()` statement in `read_from_file()`.

Signed-off-by: Metin Kaya <metin.kaya@arm.com>

### DIFF
--- a/rust/scx_utils/src/netdev.rs
+++ b/rust/scx_utils/src/netdev.rs
@@ -7,7 +7,7 @@ use std::collections::BTreeMap;
 use std::fs;
 use std::path::Path;
 
-use crate::misc::read_file_usize;
+use crate::misc::read_from_file;
 use crate::Cpumask;
 use anyhow::Result;
 
@@ -55,7 +55,7 @@ pub fn read_netdevs() -> Result<BTreeMap<String, NetDev>> {
         let iface = entry.file_name().to_string_lossy().into_owned();
         let iface_path_raw = format!("/sys/class/net/{}/device/enable", iface);
         let iface_path = Path::new(&iface_path_raw);
-        let is_enabled = read_file_usize(iface_path).unwrap_or(0);
+        let is_enabled = read_from_file(iface_path).unwrap_or(0_usize);
         if is_enabled < 1 {
             continue;
         }
@@ -67,7 +67,7 @@ pub fn read_netdevs() -> Result<BTreeMap<String, NetDev>> {
 
         let node_path_raw = format!("/sys/class/net/{}/device/numa_node", iface);
         let node_path = Path::new(&node_path_raw);
-        let node = read_file_usize(node_path).unwrap_or(0);
+        let node = read_from_file(node_path).unwrap_or(0_usize);
         let mut irqs = BTreeMap::new();
         let mut irq_hints = BTreeMap::new();
 
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -69,8 +69,8 @@
 //! hierarchy are entirely read-only. If the host topology were to change (due
 //! to e.g. hotplug), a new Topology object should be created.
 
-use crate::misc::read_file_usize;
 use crate::misc::read_file_usize_vec;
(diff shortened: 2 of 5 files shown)


rust/scx_utils/src/netdev.rs
```diff
@@ -55,7 +55,7 @@ pub fn read_netdevs() -> Result<BTreeMap<String, NetDev>> {
         let iface = entry.file_name().to_string_lossy().into_owned();
         let iface_path_raw = format!("/sys/class/net/{}/device/enable", iface);
         let iface_path = Path::new(&iface_path_raw);
-        let is_enabled = read_file_usize(iface_path).unwrap_or(0);
+        let is_enabled = read_from_file(iface_path).unwrap_or(0);
```
JakeHillion, line 58: Do all of these warrant an explicit type name? Did a quick test and it looks like `0` is an `i32` here so we're probably extracting this as an `i32`. This is probably not too bad because we should get an error instead of a clamp, but it feels like a good place to be explicit.

metin-arm: Hi Jake,
Thanks for the review.
I can make the type explicit (e.g., `let is_enabled: usize = read_from_file(iface_path).unwrap_or(0);`).
Please let me know if it looks OK.

htejun (approved):

rust/scx_utils/src/topology.rs
```diff
@@ -630,8 +630,8 @@ fn avg_cpu_freq() -> Option<(usize, usize)> {
     let cpu_paths = glob("/sys/devices/system/cpu/cpu[0-9]*").ok()?;
     for cpu_path in cpu_paths.filter_map(Result::ok) {
         let freq_path = cpu_path.join("cpufreq");
-        let max_freq = read_file_usize(&freq_path.join("scaling_max_freq")).unwrap_or(0);
-        let base_freq = read_file_usize(&freq_path.join("base_frequency")).unwrap_or(max_freq);
+        let max_freq = read_from_file(&freq_path.join("scaling_max_freq")).unwrap_or(0);
+        let base_freq = read_from_file(&freq_path.join("base_frequency")).unwrap_or(max_freq);
```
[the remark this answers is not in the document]
metin-arm: You are right about `0`'s being `i32` literal.
Made the type conversions explicit.
Thanks.

JakeHillion (approved):
Nice, thanks for all the amends!

metin-arm:
Fixed the lint error. Here is the diff in the last push:
```Diff
diff --git i/rust/scx_utils/src/topology.rs w/rust/scx_utils/src/topology.rs
index 0259988d..647448d4 100644
--- i/rust/scx_utils/src/topology.rs
+++ w/rust/scx_utils/src/topology.rs
@@ -455,7 +455,8 @@ fn create_insert_cpu(
     let min_freq = read_from_file(&freq_path.join("scaling_min_freq")).unwrap_or(0_usize);
     let max_freq = read_from_file(&freq_path.join("scaling_max_freq")).unwrap_or(0_usize);
     let base_freq = read_from_file(&freq_path.join("base_frequency")).unwrap_or(max_freq);
-    let trans_lat_ns = read_from_file(&freq_path.join("cpuinfo_transition_latency")).unwrap_or(0_usize);
+    let trans_lat_ns =
+        read_from_file(&freq_path.join("cpuinfo_transition_latency")).unwrap_or(0_usize);

     // Cpu capacity
     let (cap_suffix, _avg_rcap, max_rcap) = capacity_src.unwrap_or(("".to_string(), 1024, 1024));
```

PR #1590, opened 2025-03-31
scx_bpfland: Introduce an option to disable SMT awareness

arighi:
Introduce --disable-smt to explicitly disable SMT support. With this option, the scheduler will select idle CPUs without distinguishing between fully-idle and partially-idle SMT cores.


hodgesds (approved):
LGTM

PR #1592, opened 2025-03-31
Cpumask code fixups

etsal:
Miscellaneous cleanups for the now common cpumask code. The most notable one is the phasing out of cast_kern() from the allocator.


hodgesds (approved):
Nice to see the `cast_kern` disappear 👍

PR #1593, opened 2025-03-31
scx_layered: remove extra call to test_and_clear_cpu in pick_idle_from

likewhatevs:
Remove extra call to scx_bpf_test_and_clear_cpu_idle in pick_idle_cpu_from which would happen if smt_enabled and prev_over_idle_core when prev_cpu is busy and sibling_cpu is idle.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -776,11 +776,11 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
 		}
 
 		// try prev if smt sibling empty
-		if (prev_in_cand &&
-			bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
-			scx_bpf_test_and_clear_cpu_idle(prev_cpu))
-			return prev_cpu;
-		prev_in_cand = false;
+		if (prev_in_cand && bpf_cpumask_test_cpu(prev_cpu, idle_smtmask)) {
+			if (scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+				return prev_cpu;
+			prev_in_cand = false;
+		}
 
 		// try any idle core
 		cpu = scx_bpf_pick_idle_cpu(cand_cpumask, SCX_PICK_IDLE_CORE);
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -766,31 +766,33 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
 	 * partially idle @prev_cpu.
 	 */
 	if (smt_enabled) {
+
+		// try prev if prev_over_idle_core
 		if (prev_in_cand &&
 			layer->prev_over_idle_core) {
-			// try prev if prev_over_idle_core
 			if (scx_bpf_test_and_clear_cpu_idle(prev_cpu))
 				return prev_cpu;
 			prev_in_cand = false;
 		}
 
-		// try prev if smt sibling is empty
+		// try prev if smt sibling empty
 		if (prev_in_cand &&
 			bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
 			scx_bpf_test_and_clear_cpu_idle(prev_cpu))
 			return prev_cpu;
+		prev_in_cand = false;
 
-		// try a wholly idle cpu
+		// try any idle core
 		cpu = scx_bpf_pick_idle_cpu(cand_cpumask, SCX_PICK_IDLE_CORE);
 		if (cpu >= 0)
 			return cpu;
-
-	} else {
-		// if not SMT, try prev
-		if (prev_in_cand &&
-			scx_bpf_test_and_clear_cpu_idle(prev_cpu))
-			return prev_cpu;
 	}
+
+	// try prev if not previously tried and failed
+	if (prev_in_cand &&
+		scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+		return prev_cpu;
+
 	// return any idle cpu
 	return scx_bpf_pick_idle_cpu(cand_cpumask, 0);
 }
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -766,27 +766,32 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
 	 * partially idle @prev_cpu.
 	 */
 	if (smt_enabled) {
-
-		// if layer is prev_over_idle_core, try prev
-		if (prev_in_cand && layer->prev_over_idle_core &&
-				scx_bpf_test_and_clear_cpu_idle(prev_cpu))
-				return prev_cpu;
-
-		// try prev if layers smt sibling is idle
 		if (prev_in_cand &&
-		    bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
-		    scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+			layer->prev_over_idle_core) {
+			// try prev if prev_over_idle_core
+			if (scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+				return prev_cpu;
+			prev_in_cand = false;
+		}
+
+		// try prev if smt sibling is empty
+		if (prev_in_cand &&
+			bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
+			scx_bpf_test_and_clear_cpu_idle(prev_cpu))
 			return prev_cpu;
 
 		// try a wholly idle cpu
 		cpu = scx_bpf_pick_idle_cpu(cand_cpumask, SCX_PICK_IDLE_CORE);
 		if (cpu >= 0)
 			return cpu;
+
+	} else {
+		// if not SMT, try prev
+		if (prev_in_cand &&
+			scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+			return prev_cpu;
 	}
-
-	if (prev_in_cand && scx_bpf_test_and_clear_cpu_idle(prev_cpu))
-		return prev_cpu;
-
+	// return any idle cpu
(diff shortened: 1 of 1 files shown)


etsal (commented):
IIUC the scx_bpf_test_and_clear_cpu_idle must always be the last call because it has side effects. With this change, if the call succeeds but then the checks fail we can set the CPU as in use but then never use it, effectively leaking it.

etsal:
The code below would work, though it is a bit of an abomination.

```
		if (prev_in_cand && 
                    (layer->prev_over_idle_core ||  bpf_cpumask_test_cpu(prev_cpu, idle_smtmask)) &&
		    scx_bpf_test_and_clear_cpu_idle(prev_cpu))
				return prev_cpu;
```

likewhatevs:
> IIUC the scx_bpf_test_and_clear_cpu_idle must always be the last call because it has side effects. With this change, if the call succeeds but then the checks fail we can set the CPU as in use but then never use it, effectively leaking it.

Yeah I didn't know that but it kinda looked like that was the case, thanks (also thx for the boolean gymnastics which would work).

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,16 +766,10 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
 	 * partially idle @prev_cpu.
 	 */
 	if (smt_enabled) {
-
-		// if layer is prev_over_idle_core, try prev
-		if (prev_in_cand && layer->prev_over_idle_core &&
```
hodgesds, line 771: Do you want `prev_over_idle_core` to work when SMT is disabled? It seems to ignore that case, is that intended?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,16 +766,10 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
-				return prev_cpu;
-
-		// try prev if layers smt sibling is idle
-		if (prev_in_cand &&
-		    bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
-		    scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+		// use prev if sibling is free or prev_over_idle_core
+		if (prev_in_cand && (layer->prev_over_idle_core
+			|| bpf_cpumask_test_cpu(prev_cpu, idle_smtmask)) &&
+			scx_bpf_test_and_clear_cpu_idle(prev_cpu))
```
htejun, line 772: `||` being on the previous line would be more consistent. I'd just put each condition on its own line with the right indentation. It's a bit difficult to read as-is.

htejun: Also, below, there's another sometimes duplicate `scx_bpf_test_and_clear_cpu_idle(prev_cpu)` right after the `if (smt_enabled) {}` block. We can skip that if `smt_enabled && prev_over_idle_core`, right? An easier way to structure this may be something like this:
```
if (prev_in_cand && prev_over_idle_core) {
    if (test_idle())
        return prev_cpu;
    prev_in_cand = false;
}

if (smt_enabled) {
    if (prev_in_cand && test_smt_mask() && test_idle())
            return prev_cpu;
   ...
}

if (prev_in_cand && test_idle())
    return prev_cpu; 
```

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: yep, prev_over_idle_core prefers previous CPU even if SMT sibling is busy.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: >     prev_in_cand = false;

ahh that's kinda cool.

likewhatevs: I *thinK* that gets rid of the possible duplicate calls while still doing the right thing?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,27 +766,30 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
-		    scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+		if (prev_in_cand) {
+			if (layer->prev_over_idle_core) {
+				// try prev if prev_over_idle_core
+				if (scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+					return prev_cpu;
+			} else {
+				// try prev if smt sibling is empty
+				if (bpf_cpumask_test_cpu(prev_cpu, idle_smtmask)
+					&& scx_bpf_test_and_clear_cpu_idle(prev_cpu))
```
htejun, line 777: Can you put `&&` in the preceding line to stay consistent with other lines?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,27 +766,30 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
+					return prev_cpu;
+			}
```
htejun, line 779: Don't we still need to do `SCX_PICK_IDLE_CORE` here?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,27 +766,30 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
+		} else {
+			// try a wholly idle cpu
+			cpu = scx_bpf_pick_idle_cpu(cand_cpumask, SCX_PICK_IDLE_CORE);
+			if (cpu >= 0)
+				return cpu;
+		}
+	} else {
+		// if not SMT, try prev
+		if (prev_in_cand
+			&& scx_bpf_test_and_clear_cpu_idle(prev_cpu))
```
htejun, line 789: Ditto with `&&`.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: yes, thanks

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,27 +766,30 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
 	 * partially idle @prev_cpu.
 	 */
 	if (smt_enabled) {
-
-		// if layer is prev_over_idle_core, try prev
-		if (prev_in_cand && layer->prev_over_idle_core &&
+		if (prev_in_cand &&
+			layer->prev_over_idle_core) {
+			// try prev if prev_over_idle_core
```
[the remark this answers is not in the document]
likewhatevs: Ahh ok, thanks!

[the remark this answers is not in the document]
likewhatevs: I think that fixes that while avoiding the extra call?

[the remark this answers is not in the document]
etsal: I think this is correct now, though the changes in lines 769-774 are not strictly necessary now. The key change is adding the else block for the non-SMT case to avoid rechecking, since for SMT we now have the same check in lines 776-778.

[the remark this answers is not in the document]
likewhatevs: Yeah lol looking at it again, I think the main change was just adding an else (and maybe making things easier to follow, although wrt/ that, idk).

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,27 +766,30 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
+			if (scx_bpf_test_and_clear_cpu_idle(prev_cpu))
 				return prev_cpu;
+		}
 
-		// try prev if layers smt sibling is idle
-		if (prev_in_cand &&
-		    bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
-		    scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+		// try prev if smt sibling is empty
+		if (bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
```
htejun, line 777: This wouldn't be true if the preceding `prev_cpu` test failed.

htejun (approved):

likewhatevs:
I might have just made this wrong but I think I follow the logic now, does it still look right?

tl;dr is toggle prev_in_cand if we know not prev, gate other checks on that where possible?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,27 +766,32 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
 		// try a wholly idle cpu
 		cpu = scx_bpf_pick_idle_cpu(cand_cpumask, SCX_PICK_IDLE_CORE);
 		if (cpu >= 0)
 			return cpu;
-	}
-
-	if (prev_in_cand && scx_bpf_test_and_clear_cpu_idle(prev_cpu))
-		return prev_cpu;
 
+	} else {
```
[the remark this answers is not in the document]
likewhatevs: I think that checks out wrt/ doing the right thing with the least calls now. but will wait for confirmation.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,31 +766,33 @@ static s32 pick_idle_cpu_from(const struct cpumask *cand_cpumask, s32 prev_cpu,
...
 			prev_in_cand = false;
 		}
 
-		// try prev if smt sibling is empty
+		// try prev if smt sibling empty
 		if (prev_in_cand &&
 			bpf_cpumask_test_cpu(prev_cpu, idle_smtmask) &&
 			scx_bpf_test_and_clear_cpu_idle(prev_cpu))
 			return prev_cpu;
+		prev_in_cand = false;
```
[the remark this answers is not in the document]
likewhatevs: right can fall through on the middle condition, thanks.

htejun (approved):

PR #1595, opened 2025-03-31
scx_p2dq: Refactor dispatch load balancing

hodgesds:
Refactor pick two load balancing path in dispatch to consider the current LLC load and avoid migrating tasks to LLCs with more load. This provides better load balancing by only migrating tasks when needed.

Ran `stress-ng` to saturation and pick2 dispatches are greatly reduced:
```
direct/idle/greedy 315/1031/0
        dsq same/migrate/keep 63551/2547/0
        wake_prev/llc/mig 0/0/2
        pick2 select/dispatch 174/23
        migrations llc/node: 2453/0
direct/idle/greedy 369/1279/0
        dsq same/migrate/keep 59850/3460/0
        wake_prev/llc/mig 6/0/2
        pick2 select/dispatch 407/24
        migrations llc/node: 3248/0
```
`main` branch has far more dispatch pick2 under saturation:
```
direct/idle/greedy 331/200/0
        dsq same/migrate/keep 64113/2264/0
        wake_prev/llc/mig 0/0/1
        pick2 select/dispatch 57/2140
        migrations llc/node: 4392/0
direct/idle/greedy 331/232/0
        dsq same/migrate/keep 63909/2099/0
        wake_prev/llc/mig 0/0/0
        pick2 select/dispatch 76/1984
        migrations llc/node: 4396/0
```


JakeHillion (approved):

etsal:
IIUC, this code prevents the local LLC from work stealing from left & right if the local LLC is _less_ loaded than either.  It's very possible I'm misreading it, if not, why not flip the condition?

hodgesds:
> IIUC, this code prevents the local LLC from work stealing from left & right if the local LLC is _less_ loaded than either. It's very possible I'm misreading it, if not, why not flip the condition?

I think you are right, `pick_two_llc_ctx` has a `most_loaded` parameter (not currently used) which optionally returns the most loaded DSQ. I was going to replace some of `pick_two_cpu` with `pick_two_llc_ctx` to remove duplicate code, but now this code is in a slight mess. For now I'll just remove the `most_loaded` parameter and I think that should work.

PR #1597, opened 2025-04-01
scx_lavd: Fix incorrect task affinity testing and more.

multics69:
p->nr_cpus_allowed does not always tell the number of allowed CPUs when
p->cpus_ptr does not point to p->cpus_mask [1]. The previous implementation
of is_affinitized() didn't consider that.

So, track the number of allowed CPUs under task_ctx (taskc->nr_cpus_allowed)
and update it whenever task's cpumask is updated. Change is_affinitized() and
its callers accordingly.

[1] https://lore.kernel.org/lkml/Z-sOIwUNgrjbQJkx@slm.duckdns.org/

In addition to this, this PR contains a code cleanup without functional change.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -656,6 +656,20 @@ static void update_stat_for_stopping(struct task_struct *p,
 	taskc->svc_time += task_run_time / p->scx.weight;
 
 	/*
+	 * Count how many times a task completely consumed the assigned time
+	 * slice to boost the task's slice. If not fully consumed, decrease
+	 * the slice boost priority by half.
+	 */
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1406,6 +1395,7 @@ static void init_task_ctx(struct task_struct *p, struct task_ctx *taskc)
 	u64 now = scx_bpf_now();
 
 	__builtin_memset(taskc, 0, sizeof(*taskc));
+	taskc->nr_cpus_allowed = bpf_cpumask_weight(p->cpus_ptr);
```
hodgesds, line 1398: I like this way of handling the nr_cpus_allowed, I'll probably even borrow it :smile:

hodgesds (approved):
LGTM

multics69:
Thanks @hodgesds for the review.

PR #1603, opened 2025-04-01
scx_p2dq: Update default load balancing config

hodgesds:
Update the default load balancing config to allow load balancing from all non interactive DSQs by default. This provides better load balancing characteristics on workloads that have a mix of interactive and CPU bound tasks.

Example from running `schbench` with `scxtop` to show load:
image

### DIFF
--- a/scheds/rust/scx_p2dq/src/main.rs
+++ b/scheds/rust/scx_p2dq/src/main.rs
@@ -109,7 +109,7 @@ struct Opts {
     idle_resume_us: Option<u32>,
 
     /// Only pick2 load balance from the max DSQ.
-    #[clap(long, default_value = "true", action = clap::ArgAction::Set)]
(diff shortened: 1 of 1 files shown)


JakeHillion (approved):
I really wish `clap` let us write `--no-max-dsq-pick2` like compilers. Do you know if users can currently write `--max-dsq-pick2=false` explicitly? Would be good to be able to make changes to defaults like this in future without messing things up for people who explicitly want that behaviour.

hodgesds:
> I really wish `clap` let us write `--no-max-dsq-pick2` like compilers. Do you know if users can currently write `--max-dsq-pick2=false` explicitly? Would be good to be able to make changes to defaults like this in future without messing things up for people who explicitly want that behaviour.

Fair point about not breaking compatibility, updated now so `--max-dsq-pick2=false` still works.

PR #1605, opened 2025-04-01
scx_bpfland: NUMA load balancer

arighi:
Implement a simple node load balancer: the idea is to extend the search for fully idle SMT cores across NUMA nodes (and relax cache optimization) when the current node is overloaded, so that, on wake up, tasks are naturally migrating to less loaded nodes when load distribution across nodes becomes unbalanced.

Also provide a new option `--disable-numa` to disable this rebalancing.logic and fallback to the previous behavior.


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -1259,22 +1339,113 @@ static void init_cpuperf_target(void)
...
+		 * System has at least one idle node if its current
+		 * utilization is 25% or below.
+		 */
+		if (nctx->perf_lvl <= SCX_CPUPERF_ONE / 4)
+			has_idle_nodes = true;
+
+		/*
+		 * Reset partial node statistics.
+		 */
+		nctx->nr_cpus = 0;
```
hodgesds, line 1414: Wouldn't this only need to be calculated once? Could be done in init, but I guess this would handle hotplugging.

arighi: Oh yes indeed, we can evaluate that in ops.init(), I'll change that, thanks!

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -111,13 +111,71 @@ const volatile bool smt_enabled = true;
...
+} numa_timer SEC(".maps");
+
+/*
+ * Per-node context.
+ */
+struct node_ctx {
+	u64 tot_perf_lvl;
+	u64 nr_cpus;
+	u64 perf_lvl;
+	bool need_rebalance;
```
multics69, line 140: I think READ_ONCE() and WRITE_ONCE() needs to be used when accessing `need_rebalance` because that variable can be accessed from any CPUs concurrently and require the exact value.

arighi: That's right, it's true that it might be inaccurate, but I think it's ok, considering that we access `need_rebalance` in some hot paths (`ops.select_cpu()`) and the value is refreshed every second via a timer. Essentially, it's a inherently imprecise by design, so the extra overhead (even if it's negligible) and code complexity would be unnecessary I think.

arighi:
PR udpated to set node's `nr_cpus` once during init, as suggested by @hodgesds, thanks!

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
multics69: That makes sense. Other than that, LGTM.

multics69 (approved):

PR #1606, opened 2025-04-01
scx_p2dq: Refactor freq boost

hodgesds:
Refactor frequency boosting so that the most CPU throughput bound DSQ gets a frequency boost before running.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -738,8 +738,10 @@ void BPF_STRUCT_OPS(p2dq_running, struct task_struct *p)
 	else
 		p->scx.dsq_vtime = llcx->vtime;
 
-	if (sched_mode == MODE_PERFORMANCE && taskc->dsq_index == 0) {
-		scx_bpf_cpuperf_set(task_cpu, 1024);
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -738,8 +738,10 @@ void BPF_STRUCT_OPS(p2dq_running, struct task_struct *p)
 	else
 		p->scx.dsq_vtime = llcx->vtime;
 
-	if (sched_mode == MODE_PERFORMANCE && taskc->dsq_index == 0) {
```
hodgesds, line 741: At some point I'll implement different modes....

hodgesds:
> Is this gated behind a flag btw? It probably should be if not for devices where a power limit is set to avoid jerky frame times due to thermal throttling, IIUC what this does correctly.

No flag currently, IIRC this would only affect performance if the `schedutil` frequency governor is used.

likewhatevs:
> No flag currently, IIRC this would only affect performance if the schedutil frequency governor is used.

thx

likewhatevs (approved):
why boost the least interactive and not the most interactive (i.e. is the most interactive waiting all the time or something, and even then don't those tasks matter most)?

multics69 (approved):

PR #1608, opened 2025-04-01
scx_layered: Use lookup_task_ctx_may_fail in non-critical paths

hodgesds:
In some situations scx_layered is exiting from failing to lookup the task_ctx. Use the more permissive lookup_task_ctx_may_fail in where scheduling decisions are not being made.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -573,7 +573,7 @@ int BPF_PROG(tp_cgroup_attach_task, struct cgroup *cgrp, const char *cgrp_path,
 			break;
 		}
 
-		if ((taskc = lookup_task_ctx(next)))
+		if ((taskc = lookup_task_ctx_may_fail(next)))
 			taskc->refresh_layer = true;
 	}
 
@@ -2915,7 +2915,7 @@ u64 antistall_set(u64 dsq_id, u64 jiffies_now)
 		if (!(p = bpf_task_from_pid(__p->pid)))
 			continue;
 
-		if (!(taskc = lookup_task_ctx(p)))
+		if (!(taskc = lookup_task_ctx_may_fail(p)))
 			goto unlock;
 
 		cur_delay = get_delay_sec(p, jiffies_now);


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2459,7 +2459,7 @@ void BPF_STRUCT_OPS(layered_stopping, struct task_struct *p, bool runnable)
 	s32 task_lid;
 	u64 runtime;
 
-	if (!(cpuc = lookup_cpu_ctx(-1)) || !(taskc = lookup_task_ctx(p)))
+	if (!(cpuc = lookup_cpu_ctx(-1)) || !(taskc = lookup_task_ctx_may_fail(p)))
```
hodgesds, line 2462: I'm not sure on this one, I'd assume stopping is always called so maybe this isn't required.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -573,7 +573,7 @@ int BPF_PROG(tp_cgroup_attach_task, struct cgroup *cgrp, const char *cgrp_path,
 			break;
 		}
 
-		if ((taskc = lookup_task_ctx(next)))
+		if ((taskc = lookup_task_ctx_may_fail(next)))
```
hodgesds, line 576: Since the leader `taskc` uses `lookup_task_ctx_may_fail` this should be fine.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2915,7 +2915,7 @@ u64 antistall_set(u64 dsq_id, u64 jiffies_now)
 		if (!(p = bpf_task_from_pid(__p->pid)))
 			continue;
 
-		if (!(taskc = lookup_task_ctx(p)))
+		if (!(taskc = lookup_task_ctx_may_fail(p)))
```
likewhatevs, line 2918: Strongly oppose this one. It's better we get a failure due to task_ctx failure here than a stall which could happen due to a myriad of reasons. Maybe wait to see this be a problem before making this change.

htejun:
As a workaround, this is fine, but it'd be great if we understand why it's failing first and explain what we're working around in the comment. Just looking at the code, this isn't supposed to happen, no?

PR #1612, opened 2025-04-02
scx_p2dq: Avoid aggressive return on cpu selection

hodgesds:
Avoid returning early if there are no idle CPUs. Returning the previous CPU if there are no idle CPUs is overly aggressive and causes affinitized tasks to hang if there are no idle CPUs.


JakeHillion (approved):
Makes sense! I wonder if it's worth a comment somewhere in the code, given there are some post-conditions here that aren't immediately obvious.

hodgesds:
> Makes sense! I wonder if it's worth a comment somewhere in the code, given there are some post-conditions here that aren't immediately obvious.

I almost want to rename that function, because it's doing far more than picking an idle CPU.

PR #1613, opened 2025-04-02
common: include file and line number in scx_bpf_error

JakeHillion:
scx_bpf_error takes a user provided format string and args to exit the scheduler
with an error. When deploying a scheduler at scale this can be tricky to convert
back to a line of code. Add the `__FILE__` and `__LINE__` macros to give us a
fighting chance of finding the line the error occurred on.

Uses string literal concatenation rather than printf parameters to avoid using
up the 6 argument limit in BPF-C. This string should end up as a compile time
constant instead of on the BPF stack, and is passed directly to the kfunc.

Test plan:
- Built everything (local for Rust & CI for all).

Tested by injecting a failure into p2dq. Before:

    Error: EXIT: scx_bpf_error (failed)

After:

    Error: EXIT: scx_bpf_error (src/bpf/main.bpf.c:1047: failed)

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -105,6 +105,9 @@ void scx_bpf_events(struct scx_event_stats *events, size_t events__sz) __ksym __
 static inline __attribute__((format(printf, 1, 2)))
 void ___scx_bpf_bstr_format_checker(const char *fmt, ...) {}
 
+#define SCX_STRINGIFY(x) #x
+#define SCX_TOSTRING(x) SCX_STRINGIFY(x)
+
 /*
  * Helper macro for initializing the fmt and variadic argument inputs to both
  * bstr exit kfuncs. Callers to this function should use ___fmt and ___param to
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
Seems super useful to me, curious to what others think.

etsal (approved):
Looks good to me, normally scx_bpf_error messages are distinct and self-descriptive enough that the func/line number doesn't make much of a difference but it doesn't hurt either. The func/line number does provide a nice way to find out if we're accidentally running a stale binary.

arighi (commented):
Looks really useful to me!

Maybe there are too many parenthesis in the output, how about having something like this for the error line?
```
scx_bpf_error src/bpf/main.bpf.c:1363 failed
```

JakeHillion:
> Looks really useful to me!
> 
> Maybe there are too many parenthesis in the output, how about having something like this for the error line?
> 
> ```
> scx_bpf_error src/bpf/main.bpf.c:1363 failed
> ```

I've been trawling the code for this, I think the outer set of parantheses come from the kernel? In Rust we get a `*const c_char` seemingly from somewhere else, and the output is in the DEBUG DUMP section (I didn't give enough context in the commit message):

```
DEBUG DUMP

scx_p2dq[1615089] triggered exit kind 1025:
  scx_bpf_error ((src/bpf/main.bpf.c:1047) failed)

Backtrace:
  scx_bpf_error_bstr+0x66/0x80
  bpf_prog_7a8b3652834ed4d2_p2dq_init_task+0x48/0x2a8
  bpf__sched_ext_ops_init_task+0x4b/0xb3
  scx_ops_init_task+0x6d/0x1a0
  scx_ops_enable.isra.0+0x9fa/0xed0
  bpf_struct_ops_link_create+0x167/0x1c0
  __sys_bpf+0x1a41/0x2310
  __x64_sys_bpf+0x1e/0x30
  do_syscall_64+0xb7/0x210
  entry_SYSCALL_64_after_hwframe+0x77/0x7f

CPU states
----------

CPU 7   : nr_run=0 flags=0x11 cpu_rel=0 ops_qseq=28548 pnt_seq=207336
          curr=scx_p2dq[1615089] class=fair_sched_class

```

I can drop the parentheses and make this look like:

    scx_bpf_error (src/bpf/main.bpf.c:1047: failed)

From userspace, unless I'm misunderstanding something here.

JakeHillion:
Updated the output format. Still has one set of parantheses that I don't think I'm adding but is a bit less verbose now. It looks the same in the debug dump and in the error reason, so I think this extra set come from the kernel.

multics69 (approved):

arighi (approved):

PR #1615, opened 2025-04-02
scx_chaos: fix missing kernel symbols on older kernels.

devnexen:


JakeHillion (changes requested):
This change breaks some invariants in this code that I probably should have commented. `scx_bpf_dsq_move` and the previous name `scx_bpf_dispatch_from_dsq` return true if the task was consumed from the iterated DSQ, or false if it had previously been consumed/dequeued. The code in chaos is carefully written such that items are only dequeued in one place, which means we don't need to check the return value.

The compat macros here return `false` if there is no compatible kfunc, which we would need to handle. But I think this is still a step too far for chaos - chaos should not attach if neither name of this kfunc is available because it cannot work - it is guaranteed to stall as soon as anything chaotic happens.

I think the approach here would be to have a custom compat macro for chaos that selects the correct name but fails if `scx_bpf_dispatch_from_dsq` isn't present. What do you think?

devnexen:
Ah ok good to know, I can see to this.

JakeHillion (changes requested):
My bad, I think I wasn't clear before. The `bpf_dsq_move` wrapper for example should look like:

    #define __COMPAT_chaos_scx_bpf_dsq_move(it__iter, p, dsq_id, enq_flags) \
        (bpf_ksym_exists(scx_bpf_dsq_move) ? \
            scx_bpf_dsq_move((it__iter), (p), (dsq_id), (enq_flags) : \
            scx_bpf_dispatch_from_dsq___compat((it__iter), (p), (dsq_id), (enq_flags)))

The formatting in the GitHub comment is a total mess but please make sure the `\` are indented with tabs and lined up.

The behaviour here should be that scx_chaos fails to attach completely on a kernel that has neither kfunc, not that it fails at runtime. The `scx_bpf_error` here will let the scheduler attach then later have it fail - this is inappropriate because some variant of `scx_bpf_dsq_move` is essential for chaos to function, so it shouldn't be able to attach unless the kfunc is available by one name or the other.

If you implement the macros like that and drop the unnecessary `scx_bpf_unknown_ksym` I'll be happy to take this patch - let me know if anything is unclear :)

JakeHillion (approved):
Thanks for the cleanup!

PR #1620, opened 2025-04-03
scx: publish test rpm of schedulers on each commit

likewhatevs:
This PR adds boilerplate/an rpmspec to enable adding a webhook pointed at a copr to trigger commit-versioned builds when PRs are merged.

The end result would look like this: https://copr.fedorainfracloud.org/coprs/likewhatevs/sched_ext/ , with packages generated looking like this:
image

The main benefit of this would be performance testing, but also maybe ease of testing (i.e. bazzite handhelds) overall.

I generated the following performance data on layered a few days ago and it took about 24 hours to compute:

https://rando.likewhatevs.io/layered_walk_81_pct_done.html
image

This would have taken maybe 10 or 12 hours and had less noise if I had this available previously, so I would like to make this available for next time (the download here was cached, but copr is pretty fast even if it weren't):
image


scx.spec
```diff
@@ -0,0 +1,49 @@
+%global debug_package %{nil}
+
+Name: 		scx
+Version:	1%{?dist}
+Release:        %autorelease
+Summary:	sched-ext/scx scheds
+
+
+License:	GPLv2
+URL:		https://github.com/likewhatevs/scx
```
JakeHillion, line 10:
```suggestion
URL:		https://github.com/sched-ext/scx
```

???

scx.spec
```diff
@@ -0,0 +1,49 @@
...
+Source0:	${nil}
+
+BuildRequires: cargo, rust, elfutils-devel, clang
+
+%description
+scx schedulers packaged to simplify perf testing
```
JakeHillion, line 16: Is this a good description? These are available for general use as far as I can tell, it doesn't only need to be perf testing.

JakeHillion (approved):
Neat. What happens if this fails, do we get notified somehow?

likewhatevs:
going a different direction w/ this (want devel headers also)

PR #1624, opened 2025-04-03
scx_utils: Add EnergyModel (energy_model.rs).

multics69:
Schedulers need to know the energy model of a system to make energy-aware
scheduling decisions. The EnergyModel (energy_model.rs) represents the energy
model of a system. It loads the energy model data from debugfs (typically at
/sys/kernel/debug/energy_model), so the kernel should support not only the
energy model (CONFIG_ENERGY_MODEL) but also debugfs.

A system's energy model (EnergyModel) consists of one or more performance
domains (PerfDomain). A group of CPUs belonging to the same performance domain
is frequency-scaled altogether (i.e., the same frequency domain), so it has
the same performance-power characteristics. The performance-power attributes
of a performance domain are represented by a collection of performance states
(PerfState). Each performance state includes
  1) CPU frequency (kHz),
  2) the cost of the transition,
  3) performance scaled by 1024 against the highest performance of
     the fastest CPU,
  4) power consumption (mW), and
  5) whether this state is power-inefficient or not.

### DIFF
--- a/rust/scx_utils/src/energy_model.rs
+++ b/rust/scx_utils/src/energy_model.rs
@@ -1,0 +1,180 @@
+// SPDX-License-Identifier: GPL-2.0
+//
+// Copyright (c) 2025 Valve Corporation.
+// Author: Changwoo Min <changwoo@igalia.com>
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+//! # SCX Energy Model
+//!
+//! A crate that allows schedulers to inspect and model the host's energy model,
+//! which is loaded from debugfs.
+
+use crate::compat;
+use crate::misc::read_from_file;
+use crate::Cpumask;
+use anyhow::bail;
+use anyhow::Result;
+use glob::glob;
+use std::collections::BTreeMap;
+use std::fmt;
+use std::path::Path;
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub struct PerfState {
+    pub cost: usize,
+    pub frequency: usize,
+    pub inefficient: usize,
+    pub performance: usize,
+    pub power: usize,
+}
+
+#[derive(Debug)]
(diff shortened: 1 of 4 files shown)


rust/scx_utils/src/energy_model.rs
```diff
@@ -0,0 +1,201 @@
...
+    let mut ps_vec = vec![];
+    for ps_path in ps_paths.filter_map(Result::ok) {
+        let ps_str = ps_path.into_os_string().into_string().unwrap();
+        ps_vec.push(ps_str);
+    }
+
+    Ok(ps_vec)
+}
+
+fn read_cpumask(path: String) -> Result<Cpumask> {
```
htejun, line 153: The comma separated range list is usually called cpulist while cpumask refers to the hexadecimal format. Maybe rename and make it a part of `Cpumask` crate?

rust/scx_utils/src/energy_model.rs
```diff
@@ -0,0 +1,201 @@
...
+            span,
+            perf_table,
+        })
+    }
+}
+
+impl PerfState {
+    /// Build a PerfState
+    pub fn new(root: String) -> Result<PerfState> {
+        let cost = read_file_usize(Path::new(&(root.clone() + "/cost"))).unwrap();
```
hodgesds, line 92: Might be a little aggressive with the `unwrap` here, since it's returning a `Result` wouldn't a `?` work as well and not cause panics?

rust/scx_utils/src/energy_model.rs
multics69: That makes sense. Fixed it.

rust/scx_utils/src/energy_model.rs
multics69: Yup. Fixed it.

PR #1626, opened 2025-04-03
lib/sdt_alloc: introduce ring buffer-based arena allocator

etsal:
The arena allocator currently uses a tree-based indexing structure that doubles as a key-value store for the allocated values. This is useful for data structures like task contexts that must be uniquely identified, but not for interchangeable structs like cpumask instances. For these structures we only need to allocate and free, and do not need to provide lookups for.

For interchangeable data structures, using a tree to back the allocator has the disadvantage that each allocated structure is permanently associated with a unique key, which is also used for freeing the struct. This has required embedding unique IDs to each data structure which is used solely by the allocator. 

Introduce a ring buffer-based allocator for arena memory that is more appropriate for these interchangeable data structures. The ring buffer holds preallocated instances of a data structure that get popped and pushed by allocation and free operations, respectively. The ring buffer expands as needed to accommodate the preallocated data structures. The ring buffer is designed for allocating large data structures, so the overhead caused by holding a pointer per struct instance is small relative to the total size of the data.

Also the new allocator for domain structs in the WD40 scheduler.

### DIFF
--- a/lib/sdt_alloc.bpf.c
+++ b/lib/sdt_alloc.bpf.c
@@ -1,8 +1,8 @@
 /*
  * SPDX-License-Identifier: GPL-2.0
- * Copyright (c) 2024 Meta Platforms, Inc. and affiliates.
- * Copyright (c) 2024 Tejun Heo <tj@kernel.org>
- * Copyright (c) 2024 Emil Tsalapatis <etsal@meta.com>
+ * Copyright (c) 2024-2025 Meta Platforms, Inc. and affiliates.
+ * Copyright (c) 2024-2025 Tejun Heo <tj@kernel.org>
+ * Copyright (c) 2024-2025 Emil Tsalapatis <etsal@meta.com>
  */
 
 #include <scx/common.bpf.h>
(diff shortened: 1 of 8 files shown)


sirlucjan:
Ive tried to compile it but without success. 

[scx-scheds-git-1.0.11.r31.gcc537468-2-x86_64-build.log](https://github.com/user-attachments/files/19603984/scx-scheds-git-1.0.11.r31.gcc537468-2-x86_64-build.log)

etsal:
> @etsal
> 
> Ive tried to compile it but without success.
> 
> [scx-scheds-git-1.0.11.r31.gcc537468-2-x86_64-build.log](https://github.com/user-attachments/files/19603984/scx-scheds-git-1.0.11.r31.gcc537468-2-x86_64-build.log)

This looks like the vmlinux.h is getting included

> @etsal
> 
> Ive tried to compile it but without success.
> 
> [scx-scheds-git-1.0.11.r31.gcc537468-2-x86_64-build.log](https://github.com/user-attachments/files/19603984/scx-scheds-git-1.0.11.r31.gcc537468-2-x86_64-build.log)

Thanks for the heads up, this is caused by the arena system header getting included in the userspace program and causing vmlinux.h to be included as well. I will look into it.

PR #1629, opened 2025-04-04
scx_utils, scx_lavd: Add a new option, "--cpu-pref-order".

multics69:
Add a new option, "--cpu-pref-order", to `scx_lavd` to easily experiment with various CPU allocation policies.
Additionally, refactor `scx_utils::cpumaks` to make `cpumask::read_cpulist()` public.

### DIFF
--- a/scheds/rust/scx_lavd/src/main.rs
+++ b/scheds/rust/scx_lavd/src/main.rs
@@ -46,12 +46,14 @@ use scx_stats::prelude::*;
 use scx_utils::autopower::{fetch_power_profile, PowerProfile};
 use scx_utils::build_id;
 use scx_utils::compat;
+use scx_utils::read_cpulist;
 use scx_utils::scx_ops_attach;
 use scx_utils::scx_ops_load;
 use scx_utils::scx_ops_open;
 use scx_utils::set_rlimit_infinity;
 use scx_utils::uei_exited;
 use scx_utils::uei_report;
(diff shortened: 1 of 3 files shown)


hodgesds (approved):
LGTM

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -585,17 +595,43 @@ impl<'a> Scheduler<'a> {
...
+            let cpu_list = read_cpulist(&opts.cpu_pref_order).unwrap();
+            let pref_mask = Cpumask::from_cpulist(&opts.cpu_pref_order).unwrap();
+            if pref_mask != topo.all_cpus_mask {
+                panic!("--cpu_pref_order does not cover the whole CPUs.");
+            }
+            cpu_pf_order = cpu_list.clone();
+            cpu_ps_order = cpu_list.clone();
         }
-        debug!("{:#?}", topo);
+        for (pos, cpu) in cpu_pf_order.iter().enumerate() {
```
hodgesds, line 627: I almost wish these were methods on the `Topology` struct 😄, maybe we can add some helper methods that do similar things.

multics69: > I almost wish these were methods on the `Topology` struct 😄, maybe we can add some helper methods that do similar things.

Once we have enough experience with lavd, we can add some to Topology after knowing what to do.

PR #1630, opened 2025-04-04
scx_utils: fix too_many_arguments warning on create_insert_cpu().

devnexen:
and make it the new energy model code more consistent,
 rust idiomatic wise.


htejun:
Making the code more idiomatic looks good to me but I'm not sure on the number of parameters part. Logically, the call being outside makes more sense and that seems a lot more important than some generic lint rules.

PR #1635, opened 2025-04-05
scx_lavd: Simplify CPU frequency scaling.

multics69:
Simplify CPU frequency scaling to decide the frequency based on CPU utilization without considering the task's performance criticality. The running average of tasks' performance criticality converges to a constant, making no significant difference. To respond quickly upon load spikes, use the bigger one between the running average and the recent utilization. When the utilization is greater than LAVD_CPU_UTIL_MAX_FOR_CPUPERF (85%), ceil to 100%.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/power.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/power.bpf.c
@@ -458,93 +458,37 @@ unlock_out:
 	return err;
 }
 
-static int calc_cpuperf_target(struct sys_stat *stat_cur,
-			       struct task_ctx *taskc, struct cpu_ctx *cpuc)
+static void update_cpuperf_target(struct cpu_ctx *cpuc)
 {
-	u64 max_load, cpu_load;
-	u32 cpuperf_target;
-
-	if (!stat_cur || !taskc || !cpuc)
-		return -EINVAL;
-
-	if (no_freq_scaling) {
-		cpuc->cpuperf_task = SCX_CPUPERF_ONE;
-		cpuc->cpuperf_avg = SCX_CPUPERF_ONE;
-		return 0;
-	}
+	u32 util, cpuperf_target;
 
 	/*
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_lavd/src/bpf/power.bpf.c
```diff
@@ -458,93 +458,37 @@ static int reinit_active_cpumask_for_performance(void)
...
-	 * performance criticality is the same as the system-wide average
-	 * criticality, we set the target CPU frequency to the maximum.
-	 *
-	 * In other words, even if CPU utilization is not so high, the target
-	 * CPU frequency could be high when the task's performance criticality
-	 * is high enough (i.e., boosting CPU frequency). On the other hand,
-	 * the target CPU frequency could be low even if CPU utilization is
-	 * high when a non-performance-critical task is running (i.e.,
-	 * deboosting CPU frequency).
+	 * The CPU utilization decides the frequency. The bigger one between
```
hodgesds, line 466: One tricky thing is that accounting for CPU time from a high priority scheduling class isn't accounting for. Not sure there is a good way of doing it but I think what you have makes sense.

hodgesds (approved):
LGTM, definitely easier to understand.

scheds/rust/scx_lavd/src/bpf/power.bpf.c
multics69: Actually, it did. When calculating the utilization of each CPU, it substracts the idle time from the total duration. So, it considers non-idle time (whether used by SCX or RT) as a total compute time. 

https://github.com/sched-ext/scx/blob/c189b288a127bcf6ec14dfdaf0cae03a91c7743a/scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c#L213

hodgesds: I missed, that! That is a clever way of calculating utilization!

PR #1637, opened 2025-04-06
scx_flash: Always bounce tasks to ops.enqueue()

arighi:
Disable auto-dispatch of migration disabled tasks and ensure all tasks are always processed via ops.enqueue().

This seems to prevent stall scenarios where, under certain conditions, migration-disabled tasks are repeatedly rescheduled on the same CPU, leading to starvation of per-CPU tasks queued in the per-node DSQs (reported and tested by the CachyOS community).

### DIFF
--- a/scheds/rust/scx_flash/src/main.rs
+++ b/scheds/rust/scx_flash/src/main.rs
@@ -142,6 +142,15 @@ impl<'a> Scheduler<'a> {
             warn!("vfs_fsync_range symbol is missing")
         }
 
+        // Set scheduler flags.
+        skel.struct_ops.flash_ops_mut().flags = *compat::SCX_OPS_ENQ_EXITING
+            | *compat::SCX_OPS_ENQ_LAST
+            | *compat::SCX_OPS_ENQ_MIGRATION_DISABLED;
+        info!(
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_flash/src/main.rs
```diff
@@ -142,6 +142,15 @@ impl<'a> Scheduler<'a> {
             warn!("vfs_fsync_range symbol is missing")
         }
 
+        // Set scheduler flags.
+        skel.struct_ops.flash_ops_mut().flags = *compat::SCX_OPS_ENQ_EXITING
+            | *compat::SCX_OPS_ENQ_LAST
+            | *compat::SCX_OPS_ENQ_MIGRATION_DISABLED;
+        info!(
```
hodgesds, line 149: Logging flags is a good idea, maybe we should put that in the helper macro.

hodgesds (approved):

scheds/rust/scx_flash/src/main.rs
arighi: Yes! Good idea, I have exactly the same code in bpfland.

multics69 (approved):

PR #1639, opened 2025-04-06
scx_lavd: Further prioritize kworkers and a few more.

multics69:
The main changes of this PR is 1) prioritizing kworkers to improve IO performance, 2) reducing the cost of preemption, and 3) reducing the cost of switching to the performance mode.

Besides these main changes, the PR includes code clean-ups -- including removing unused function args, caching repeatedly calculated time slices to sys_stat,  and relaxing the atomic visibility of sys_stat. Note that there is no functional change in the code clean-ups.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -441,7 +441,7 @@ static u32 clamp_time_slice_ns(u32 slice)
 	return slice;
 }
 
-static u64 calc_time_slice(struct task_struct *p, struct task_ctx *taskc)
+static u64 calc_time_slice(struct task_ctx *taskc)
 {
 	struct sys_stat *stat_cur = get_sys_stat_cur();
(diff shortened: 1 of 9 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -288,6 +287,12 @@ static u64 calc_weight_factor(struct task_struct *p, struct task_ctx *taskc)
 	if (is_kernel_task(p))
 		weight_boost += LAVD_LC_WEIGHT_BOOST;
 
+	/*
+	 * Further prioritize kworkers.
```
hodgesds, line 291: Random idea, I wonder if it makes sense to also boost exiting tasks (`PF_EXITING`) as well, to help reclaim resources.

multics69: If there is an environment where the resource reclaimation speed matters, it would be worth trying. :-)

PR #1643, opened 2025-04-08
scx_p2dq: Update kthread selection on task context

hodgesds:
Update the is_kthread field on the task_context to include checks of if the task is a kworker, io, or wq worker.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -782,7 +782,7 @@ static __always_inline void p2dq_runnable_impl(struct task_struct *p, u64 enq_fl
 	if (!(wakee_ctx = lookup_task_ctx(p)))
 		return;
 
-	wakee_ctx->is_kworker = p->flags & PF_WQ_WORKER;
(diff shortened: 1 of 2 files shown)


multics69 (approved):

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -782,7 +782,7 @@ static __always_inline void p2dq_runnable_impl(struct task_struct *p, u64 enq_fl
 	if (!(wakee_ctx = lookup_task_ctx(p)))
 		return;
 
-	wakee_ctx->is_kworker = p->flags & PF_WQ_WORKER;
+	wakee_ctx->is_kworker = p->flags & (PF_KTHREAD | PF_WQ_WORKER | PF_IO_WORKER);
```
arighi, line 785: I think `PF_KTHREAD` is always set when `PF_WQ_WORKER` is set (a workqueue worker is always a kthread). So I think only `PF_IO_WORKER` is needed here if you want to include the io_uring workers.

hodgesds: I wasn't sure on the `PF_KTHREAD`, thanks for confirming. I'll do some more testing, but I mostly wanted to prevent any kthreads from getting migrated. Maybe the name is no longer accurate.

PR #1653, opened 2025-04-10
add docs to stubs in common with docs

likewhatevs:
Make working w/ scheduler code easier by copying docs with stub signatures.

This isn't perfect (i.e. enums are generated, so no docs for those), but I think it's better than what we have now (jumping between kernel sources and scheduler code).

Screenshot From 2025-04-09 20-09-57

We probably should generate some of this sometime (some new fns in bpf look useful), but that might be excessive and this seems good enough atm, tbh, idk.

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -47,51 +47,591 @@ static inline void ___vmlinux_h_sanity_check___(void)
 		       "bpftool generated vmlinux.h is missing high bits for 64bit enums, upgrade clang and pahole");
 }
 
+/**
+ * scx_bpf_create_dsq - Create a custom DSQ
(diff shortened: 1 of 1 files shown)


arighi (commented):
I think it makes sense from an scx perspective, but then in the kernel we need to maintain the same doc in two different places. We can move the doc form ext.c to common.bpf.h, but then we mess up the kernel docs and it's nice to have the doc close to the code, as people are more willing to keep that in sync when something changes.

Hm, I don't know exactly how to solve this.... @htejun opinions?

likewhatevs:
regex'ing this into an already existing vmlinux wouldn't be too hard but I will ask bpf folks also because that would be a very useful thing if possible.

PR #1656, opened 2025-04-10
chaos: add --ppid-required flag

JakeHillion:
A feature to only affect tasks in the process under test has been requested in
chaos. The implementation chosen here is filtering based on having a specific
parent somewhere in the process's lineage, with the option to include that parent
in the chaos or not. This new setting defaults to on when you provide a process
for the scheduler to manage or a PID to track, but can be disabled.

The implementation walks the `real_parent` path to root for the process, marking
it as `CHAOS_MATCH_HAS_PARENT` if it finds the target PPID or another process with
`CHAOS_MATCH_HAS_PARENT`. The implementation of `chaos_match` uses a bitset so
we can test more parameters in future.

Test plan:
- `-p` argument isn't implemented, this is tech debt for later!
- Added some logging: `scx_chaos -vv -- sh -c 'for i in {1..120}; do ( sleep 1 & ); sleep 1; done; wait; exit 0'` matches the shell and children but not chaos itself.
- `scx_chaos --ppid-targeting=false -vv -- sh -c 'sleep 120'` matches everything on the box.
- `scx_chaos` matches everything on the box.


scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -124,12 +251,6 @@ __weak s32 enqueue_chaotic(struct task_struct *p __arg_trusted, u64 enq_flags,
 	return out;
 }
 
-enum dsq_check_result {
-	CHAOS_DSQ_CHECK_OKAY,
-	CHAOS_DSQ_CHECK_DISPATCH_NOW,
-	CHAOS_DSQ_CHECK_DISPATCH_SLOW,
-};
-
```
JakeHillion, line 132: misc cleanup

hodgesds (approved):
LGTM, one thought I had is if there are potentially going to be multiple PIDs targeted at some point maybe storing targeted pids in a map would make sense.

JakeHillion:
> LGTM, one thought I had is if there are potentially going to be multiple PIDs targeted at some point maybe storing targeted pids in a map would make sense.

Yeah, this is probably a good idea. I think it's more likely that we'll have one targeted & a list of exclusions. Going to hold off for now because I'm waiting for p2dq/chaos to get arenas before doing anything with runtime data - I think I do need to migrate from the const volatiles before writing autopilot and arenas seem like the right way to go.

PR #1658, opened 2025-04-11
scx_utils: Correctly test is_migration_disabled().

multics69:
Testing p->migration_disabled in a BPF code is tricky because the migration is _always_ disabled while running the BPF code. The prolog (__bpf_prog_enter) and epilog (__bpf_prog_exit) for BPF code execution disable and re-enable the migration of the current task, respectively. So, the _current_ task of the sched_ext ops is always migration-disabled. Moreover, p->migration_disabled could be two or greater when a sched_ext ops BPF code (e.g., ops.tick) is executed in the middle of the other BPF code execution.

Therefore, we should decide that the _current_ task is migration-disabled only when its migration_disabled count is greater than one. In other words, when  p->migration_disabled == 1, there is an ambiguity, so we should check if @p is the current task or not.

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -442,8 +442,27 @@ static __always_inline const struct cpumask *cast_mask(struct bpf_cpumask *mask)
  */
 static inline bool is_migration_disabled(const struct task_struct *p)
 {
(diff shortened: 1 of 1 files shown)


arighi (approved):
This is included in #1659 so this should be merged before.

PR #1659, opened 2025-04-11
scx_lavd: Do not extend the overflow set for migration-disabled tasks.

multics69:
It is necessary to pinned tasks (nr_allowed_cpus == 1) and migration-disabled tasks (migration_disabled == ture) differently. The pinned tasks are pinned by the program so they are very likely to run on the pinned CPU for a while. However, disabling migration is temporary for the sake of kernel implementation (e.g., running a BPF code). 

Therefore, let's extend the overflow set only for the pinned tasks.

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -442,8 +442,27 @@ static __always_inline const struct cpumask *cast_mask(struct bpf_cpumask *mask)
  */
 static inline bool is_migration_disabled(const struct task_struct *p)
 {
-	if (bpf_core_field_exists(p->migration_disabled))
-		return p->migration_disabled;
+	/*
+	 * Testing p->migration_disabled in a BPF code is tricky because the
+	 * migration is _always_ disabled while running the BPF code.
+	 * The prolog (__bpf_prog_enter) and epilog (__bpf_prog_exit) for BPF
+	 * code execution disable and re-enable the migration of the current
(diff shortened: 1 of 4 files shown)


scheds/include/scx/common.bpf.h
```diff
@@ -442,8 +442,27 @@ static __always_inline const struct cpumask *cast_mask(struct bpf_cpumask *mask)
...
+	 * task, respectively. So, the _current_ task of the sched_ext ops is
+	 * always migration-disabled. Moreover, p->migration_disabled could be
+	 * two or greater when a sched_ext ops BPF code (e.g., ops.tick) is
+	 * executed in the middle of the other BPF code execution.
+	 *
+	 * Therefore, we should decide that the _current_ task is
+	 * migration-disabled only when its migration_disabled count is greater
+	 * than one. In other words, when  p->migration_disabled == 1, there is
+	 * an ambiguity, so we should check if @p is the current task or not.
+	 */
```
arighi, line 459: Ohhh thank you for figuring this out! I was wondering why so many tasks had `migration_disabled` > 0, that's the reason!

At some point I just ended up ignoring `migration_disabled` and relying only on `p->nr_cpus_allowed`.

arighi (approved):

scheds/include/scx/common.bpf.h
multics69: Yes, I had the same curiosity for a while and took quite a time to figure this out.

PR #1662, opened 2025-04-11
scx_bpfland: Normalize task weight

arighi:
The task weight range is quite large: [1 .. 10000] (default 100) and it can occasionally lead to stalls when tasks with low nice values run alongside those with high nice values. This issue arises from the way vruntime, and consequently task's deadline, is scaled.

For example, in Fedora with dnf-automatic enabled, the dnf process may run with a niceness of 19. If there are tasks running with nice values of -10 or -20, their relative vruntime scaling factor would be 1000 or even 10000, which can easily increase the risk of triggering stalls.

To mitigate the influence of task weight on deadline computation, normalize the weight into a smaller range: [1 .. 128] (default 64).

This normalization ensures that the relative vruntime scaling factor in the previous example would be around 20 or 100 (at most), instead of 1000 and 10000.

This ensures a more balanced scheduling, influenced more by the task's behavior rather than its priority, and it can also help to avoid potential stall conditions in presence of large priority gaps.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -323,6 +323,28 @@ static u64 nr_tasks_waiting(int node)
 }
 
 /*
+ * Scale a value inversely proportional to the task's normalized weight.
+ */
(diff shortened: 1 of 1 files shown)


multics69 (approved):
This makes sense to me.

PR #1663, opened 2025-04-11
wd40: fixups for getting single unit compilation working

etsal:
Compilation times for WD40 have become unreasonable, often reaching above a minute. One way to reduce them is compiling the scheduler as a single compilation unit using the "include all *.c files in main.bpf.c" hack. To get this working with WD40:

1) Add #pragma once to headers from which it is missing so that we can safely include them multiple times
2) Remove the GPL license string from the allocator library. The license is still present in the beginning of the file, and the scheduler code has its own copy of the license string.

### DIFF
--- a/lib/sdt_alloc.bpf.c
+++ b/lib/sdt_alloc.bpf.c
@@ -8,8 +8,6 @@
 #include <scx/common.bpf.h>
 #include <lib/sdt_task.h>
 
-char _license[] SEC("license") = "GPL";
-
 struct {
 	__uint(type, BPF_MAP_TYPE_ARENA);
 	__uint(map_flags, BPF_F_MMAPABLE);
--- a/scheds/include/scx/ravg_impl.bpf.h
+++ b/scheds/include/scx/ravg_impl.bpf.h
@@ -1,3 +1,5 @@
+#pragma once
+
 /* to be included in the main bpf.c file */
 #include "ravg.bpf.h"
 
--- a/scheds/rust/scx_wd40/src/bpf/lb_domain.h
+++ b/scheds/rust/scx_wd40/src/bpf/lb_domain.h
@@ -1,3 +1,4 @@
+#pragma once
 
 #include <lib/sdt_task.h>
 


htejun:
I'm surprised bpf build time can be so long? Where is that coming from? It'd be sad if we fall back to the inclusion method due to build time.

etsal:
I agree and I am not planning on changing it back, but just wanted to have a stopgap. I am looking into why there is a difference, directly using bpftool to link and compile the files is very fast so I am wondering if this is due to some misconfiguration or due to libbpf-rs.

htejun (approved):

likewhatevs:
something's wrong w/ something btw -- on my setup, after editing main.bpf.c, rebuild times are 16s.

likewhatevs (approved):

likewhatevs:
maybe split the compilation steps for individual files, add things like the following for each:
```
    println!("cargo::rerun-if-changed=src/hello.c");
```

then add one that if any has changed, re-link?

likewhatevs:
> maybe split the compilation steps for individual files, add things like the following for each:
> 
> ```
>     println!("cargo::rerun-if-changed=src/hello.c");
> ```
> 
> then add one that if any has changed, re-link?

Hmm you know, I think the API for this might not be a thing (checked and didn't see it), and idk if the smart thing (don't recompile if no changes) is being done by libbpf-rs.

etsal:
> > maybe split the compilation steps for individual files, add things like the following for each:
> > ```
> >     println!("cargo::rerun-if-changed=src/hello.c");
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > then add one that if any has changed, re-link?
> 
> Hmm you know, I think the API for this might not be a thing (checked and didn't see it), and idk if the smart thing (don't recompile if no changes) is being done by libbpf-rs.

Yeah if there was a per-target rerun-if-changed that would be useful, though if any *.c file has changed we must regenerate the skeleton and that in turn requires recompiling the Rust side. Ideally we would be able to check whether the skeleton is identical and skip that part, but I assume Rust uses modification times just like make.

PR #1667, opened 2025-04-13
scx_bpfland: Avoid unnecessary cpumask updates

arighi:
Prevent updating task's local cpumasks when it's not needed. Moreover, fix a bug where the L2 cpumask was incorrectly ignored when L3 cpumask is defined.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -244,6 +244,12 @@ struct task_ctx {
 	 * for latency-sensitive tasks.
 	 */
 	u64 deadline;
+
+	/*
+	 * Task's recently used CPU: used to determine whether we need to
+	 * refresh the task's cpumasks.
+	 */
+	s32 recent_used_cpu;
 };
 
 /* Map that contains task-local storage. */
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -495,7 +502,8 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
 	/*
 	 * Refresh task domain based on the previously used cpu.
 	 */
-	task_update_domain(p, tctx, prev_cpu, p->cpus_ptr);
+	if (tctx->recent_used_cpu != prev_cpu)
```
multics69, line 505: I am a bit confused by `prev_cpu`. In what cases is `prev_cpu` different from `recent_used_cpu`? I somehow presume that both should be the same.

arighi: @multics69 the idea here is to use `recent_used_cpu` to check whether we need to refresh the cpumasks in tctx or not. If we keep selecting the same cpu from `pick_idle_cpu()` we don't have to refresh the cpumasks, if we pick a different cpu, the next time `prev_cpu` will be different than `tctx->recent_used_cpu` and we'll have to update the cpumasks. Keep in mind that `tctx->recent_used_cpu` is updated only when the cpumasks are updated via `task_update_domain()`.

I'll add a comment to clarify this.

arighi: PR updated adding some comments, hope it's more clear now. Thanks!

multics69 (approved):

PR #1670, opened 2025-04-14
scxtop: clean up unused imports

JakeHillion:
scxtop has `unused_imports = allow`, I'm guessing because it was copy/pasted
from somewhere. Remove this and cleanup the unused imports. Will be checked in
the CI under the clippy step in future.

Test plan:
- CI

### DIFF
--- a/tools/scxtop/src/perfetto_trace.rs
+++ b/tools/scxtop/src/perfetto_trace.rs
@@ -3,7 +3,7 @@
 // This software may be used and distributed according to the terms of the
 // GNU General Public License version 2.
 
-use anyhow::{Error, Result};
+use anyhow::Result;
 use protobuf::Message;
 use rand::rngs::StdRng;
 use rand::RngCore;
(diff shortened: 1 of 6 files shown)


hodgesds (approved):
LGTM

tools/scxtop/src/perfetto_trace.rs
```diff
@@ -14,12 +14,10 @@ use std::collections::BTreeMap;
 use std::fs;
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use crate::bpf_skel::types::bpf_event;
-use crate::edm::{ActionHandler, BpfEventHandler};
+use crate::edm::ActionHandler;
 use crate::{
     Action, CpuhpEnterAction, CpuhpExitAction, ExecAction, ExitAction, ForkAction, GpuMemAction,
-    HwPressureAction, IPIAction, SchedSwitchAction, SchedWakeupAction, SchedWakingAction,
```
hodgesds, line 21: Should probably add some action handlers for some of these events, since they could be added to the traces. I think some of them are unhandled.

JakeHillion: Downside of these all being on one line, I think it's only `HwPressureAction` that's unused.

PR #1672, opened 2025-04-14
scx_p2dq: Refactor dispatch load balancing

hodgesds:
Add a slack factor that prevents load balancing if load is within a certain percentage.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -58,12 +58,14 @@ const volatile bool max_dsq_pick2 = false;
 const volatile bool select_idle_in_enqueue = true;
 
 const volatile u64 wakeup_lb_busy = 90;
+const volatile u64 lb_slack_factor = 5;
 
 const volatile bool smt_enabled = true;
 const volatile bool has_little_cores = false;
 const volatile u32 debug = 2;
 
 const u32 zero_u32 = 0;
+
 const u64 lb_timer_intvl_ns = 250LLU * NSEC_PER_MSEC;
 const u64 lb_backoff_ns = 5LLU * NSEC_PER_MSEC;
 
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -334,6 +337,9 @@ static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *cur_llcx, struct llc_ctx
 	right_load = right->load;
 
 	// If the current LLCs has more load don't try to pick2.
+	u64 slack_factor = lb_slack_factor > 0 ? (lb_slack_factor * cur_load) / 100 : 0;
```
JakeHillion, line 340: nit: pointless ternary, the left hand case will be `* 0` anyway.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -57,13 +57,16 @@ const volatile bool kthreads_local = true;
 const volatile bool max_dsq_pick2 = false;
 const volatile bool select_idle_in_enqueue = true;
 
+const volatile u64 lb_slack_factor = 5;
```
JakeHillion, line 60: Is it worth making this a 128th instead of a percentage so we divide by a power of 2? Now the load balancing path is in dispatch it's not that hot and it likely doesn't make a big difference, but this is a pattern we've used elsewhere.

JakeHillion (approved):

etsal (commented):
I think I am missing how the slack factor prevents lb, does the increment operation prevent load balancing somehow or should there be a check for the slack factor that short-circuits the pick 2 functions somewhere?

EDIT: Ah nvm this makes sense

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -334,6 +337,9 @@ static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *cur_llcx, struct llc_ctx
 	right_load = right->load;
 
 	// If the current LLCs has more load don't try to pick2.
+	u64 slack_factor = (lb_slack_factor * cur_load) / 100;
+	if (slack_factor > 0)
```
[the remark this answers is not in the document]
hodgesds: Yeah, good catch I guess it can be a one liner.

etsal (approved):

PR #1674, opened 2025-04-14
scxtop: add process and thread descriptors

tyroguru:
This change establishes relationships between a thread group leader (the main thread in a process) and its children. This therefore enables us to aggregate various attributes to the process which is very necessary for analysis of workloads.

At a perfetto UI level we now get what appear to be correct groupings of threads within their process groups (the attached image shows this). We still don't label all processes with their comm as we rely on having seen the main thread during capture for that. This is annoying and I'll see if there is a way to fix that.

Screenshot 2025-04-14 at 16 25 13

### DIFF
--- a/tools/scxtop/src/bpf/intf.h
+++ b/tools/scxtop/src/bpf/intf.h
@@ -74,6 +74,7 @@ struct sched_switch_event {
 
 struct wakeup_event {
 	u32		pid;
+	u32		tgid;
 	int		prio;
 	char		comm[MAX_COMM];
 };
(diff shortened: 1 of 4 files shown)


likewhatevs:
dumb question -- does this answer the question "which pid is on which layer"?

tyroguru:
> dumb question -- does this answer the question "which pid is on which layer"?

I'm too dumb to answer :-) .  I'm using perfetto's python APIs to develop a notebook from which you can have pre-canned queries to dive down on scheduler behaviour and from there you can expand those queries as you see fit in an ad-hoc way. That's the plan anyway. This is just adding the pid<->tid groupings that we need for that. So, this change may not specifically address your need (I'm actually not sure TBH) but that question is one that needs to be addressed (pid/thread grouping by layer).

JakeHillion (approved):
I think Clippy's going to ding you for some style, especially around the map `.contain_key` calls, but otherwise this looks good to me! Will be really nice to have process relations better represented.

tyroguru:
> I think Clippy's going to ding you for some style, especially around the map `.contain_key` calls, but otherwise this looks good to me! Will be really nice to have process relations better represented.

Thanks @JakeHillion. Yep, as we're discussing offline, I've got some issues running clippy in my repo - hopefully you can fix it for me tomorrow :-) .

PR #1678, opened 2025-04-15
scx_p2dq: Fix handling of affinitized tasks

hodgesds:
Currently affinitized tasks can cause stalls. For now select a random CPU in the affinitized mask and direct dispatch. On `main` p2dq can stall when handling affinitized  tasks:

```
[sudo] password for :
22:03:56 [INFO] Running scx_p2dq (build ID: 1.0.13-g6483f5bf-dirty x86_64-unknown-linux-gnu)
22:03:56 [INFO] DSQ[0] slice_ns 100000
22:03:56 [INFO] DSQ[1] slice_ns 3200000
22:03:56 [INFO] DSQ[2] slice_ns 6400000
22:03:56 [INFO] Kernel does not support queued wakeup optimization.
22:03:56 [INFO] P2DQ scheduler started! Run `scx_p2dq --monitor` for metrics.

DEBUG DUMP
================================================================================

kworker/u97:3[16913] triggered exit kind 1026:
  runnable task stall (plasmashel:cs0[25797] failed to run for 24.000s)

Backtrace:
  scx_watchdog_workfn+0x119/0x1c0
  process_scheduled_works+0x249/0x5a0
  worker_thread+0x108/0x2d0
  kthread+0x26a/0x290
  ret_from_fork.cold+0x9/0x19
  ret_from_fork_asm+0x1a/0x30

CPU states
----------

CPU 1   : nr_run=2 flags=0x1 cpu_rel=0 ops_qseq=24571 pnt_seq=754353
          curr=swapper/1[0] class=idle_sched_class

  R plasmashel:cs0[25797] -24000ms
      scx_state/flags=3/0x9 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x8
      dsq_vtime=1631398897 slice=100000 weight=100
      cpus=03f03f

    futex_wait+0x7e/0xd0
```

Reported by CachyOs user Nick

PR #1679, opened 2025-04-15
ci: Add affinity stress tests for CI jobs

hodgesds:
Add stress tests for testing affinity to CI jobs and increase the timeout to detect stalls in handling affinitized tasks.


likewhatevs (approved):
Ahh seeing https://github.com/sched-ext/scx/pull/1678 , if the idea wrt/ this is just catch `outright fails w/ affinitized tasks` vs another situation I had in mind..., yeah this LGTM.

hodgesds:
> to add, but why these starter numbers (and why them for all)?
> 
> I ask because I think these starter numbers need to be specific to each scheduler (or CI is noisier or a touch misleading wrt/ what it covers).
> 
> Not sure what good numbers are for layered, but guessing, this test being meaningful for w/e we want covered means these numbers need to be high enough to overload the system until it fail

Yeah, the numbers are intentionally there to go just beyond the max stall timeout.

meson-scripts/stress_tests.ini
```diff
@@ -1,54 +1,54 @@
 [scx_lavd]
 sched: scx_lavd
 sched_args: -v
-stress_cmd: stress-ng -t 14 --aggressive -M -c `nproc` -f `nproc`
-timeout_sec: 15
+stress_cmd: stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-sleep 1 --affinity-pin
```
arighi, line 4: If we want to stress the scheduler even more we could use `--affinity-delay 1`, since it'd spin on `sched_yield()` instead of using `nanosleep()`.

Basically with such a short sleep time, `nanosleep()` is likely handled in the VDSO, so the task remains in user-space and once it gets the CPU it'd run for the entire time slice. With `sched_yield()` as soon the task gets the CPU it gives it back and gets queued again in the scheduler, so we should have more `ops.enqueue()` invocations and more chances to expose stall conditions.

arighi (approved):
LGTM.

I left a comment with a suggestion if we want to be more aggressive at stressing the scheduler, but the current approach works as well. Let me know if it makes sense.

JakeHillion (approved):
LGTM. It makes me wonder whether we should have a kconfig for the 30s max timeout so we can create CI kernels with a shorter one, but this is good for now.

PR #1680, opened 2025-04-15
scx_lavd: Use 1024 instead 1000 for fixed point arithmetic.

multics69:
1000 has been used for fixed point arithmetic since it is easier to understand for humans. Change to 1024 because 1) it is more efficient in the machine and 2) it is compatible with the kernel conventions.

LAVD_SCALE and LAVD_SHIFT are defined and used throughout the code. p2scale() and s2percent() are provided and used for conversion between scale (1024) and percent (100).

There is no functional change.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
+++ b/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
@@ -7,6 +7,17 @@
 #define __LAVD_H
 
 /*
+ * common macros
+ */
+#define U32_MAX		((u32)~0U)
+#define S32_MAX		((s32)(U32_MAX >> 1))
+
+#define LAVD_SHIFT			10
+#define LAVD_SCALE			(1L << LAVD_SHIFT)
+#define p2s(percent)			(((percent) << LAVD_SHIFT) / 100)
+#define s2p(scale)			(((scale) * 100) >> LAVD_SHIFT)
+
+/*
  * common constants
  */
 enum consts_internal  {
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -6,6 +6,17 @@
...
+/*
+ * common macros
+ */
+#define U32_MAX		((u32)~0U)
+#define S32_MAX		((s32)(U32_MAX >> 1))
+
+#define LAVD_SHIFT			10
+#define LAVD_SCALE			(1L << LAVD_SHIFT)
+#define p2scale(percent)		(((percent) << LAVD_SHIFT) / 100)
+#define s2percent(scale)		(((scale) * 100) >> LAVD_SHIFT)
```
arighi, line 18: Would it be too rough to approximate even 100 to 128? In this way we could just divide or multiply by 8.

About the naming how about p2s() and s2p()?

arighi (approved):
Left a comment, but LGTM.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
multics69: Thank you for the review @arighi !

> Would it be too rough to approximate even 100 to 128? In this way we could just divide or multiply by 8.

Percent is for human developers; Instead of manually calculating, for example, 70% of 1024, p2scale() is supposed to use. Similartly, instead of converting to 1024 to % for logging, s2percent() is supposed to use. For p2scale(), so far, all use cases takes an constant argument, so any reasonable compiler should calculate the value of p2scale() at the compilation time (constant folding). Hence, there should be no runtime overhead in using p2scale().

> About the naming how about p2s() and s2p()?

That sounds better. I will change it.

PR #1681, opened 2025-04-15
ci: add nixed script to cover rust part of ci

JakeHillion:
Nix'ify the build process for Rust tests and move them to the large machine when in org. Remove the test matrix, as the overhead here is mostly contained in starting the VM and it's inefficient to start many/repeat the build.

The new CI script, which only coveres Rust tests for now, doesn't require Nix to run. Nix enables it to run more reproducibly, and is required for the CI as that's how the self hosted runner handles dependencies, but is not required to run the script locally.

Specific changes:
- Move `rust-tests` to the big self-hosted machine when in the `sched-ext` org.
- Nix'ify the dependencies. This uses a fixed version of `cargo`, `rustc` and `clippy` rather than the moving target of `rust-toolchain.toml`, but it is still `stable`. We can pull more speciifc versions using a tool like fenix[0] in the future, but this is fine for now.
- Add a new `package.metadata.scx` section to describe whether to run clippy, moving this out of the GitHub actions spec and into each package's spec.
- Run for the entire workspace instead of a set of named packages. This means we can require `rust-tests` in the CI instead of requiring each individual package's matrix entry, reducing errors. It also significantly reduces duplicated work.

This new rust-tests step takes approximately the same amount of time as `rust-tests (scxtop)` on the hosted runners, but instead of using 14 runners it uses one slot on the self-hosted runner. Out of repo it takes longer, but still takes less machine time which seems like a fair compromise - forks are lower priority here.

Test plan:
- This PR's CI.
- Ran in my fork with the usual GitHub Actions runners. It works.

```
$ mkdir -p linux/arch/x86/boot
$ ln -s $(nix build --no-link --print-out-paths ./.github/include#kernels.sched_ext/for-next)/bzImage linux/arch/x86/boot/bzImage
$ nix run ./.github/include#ci all
```

Follow ups:
- The actual `vng` step is slower on the NixOS self hosted runner than the Ubuntu GitHub hosted runner (the Rust builds are faster). It is exactly the same binary with the same arguments, both using the slow bash script for now, so I'm guessing there's something missing to make it go faster. It's still faster than not having this change for now so can address later.

[0] https://github.com/nix-community/fenix

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -1,0 +1,215 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import os
+import subprocess
+import sys
+from typing import List, Optional
+import random
+
+
+def run_command(
+    cmd: List[str], check: bool = True, env: Optional[dict] = None
+) -> subprocess.CompletedProcess:
+    """Run a command and return the result."""
+    print(f"Running: {' '.join(cmd)}", flush=True)
+    merged_env = os.environ.copy()
+    if env:
+        merged_env.update(env)
+    return subprocess.run(cmd, check=check, env=merged_env)
+
+
+def get_clippy_packages() -> List[str]:
+    """Get list of packages that should be linted with clippy."""
+    result = subprocess.run(
+        ["cargo", "metadata", "--format-version", "1"],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+    metadata = json.loads(result.stdout)
+
+    clippy_packages = []
+    for pkg in metadata.get("packages", []):
+        pkg_metadata = pkg.get("metadata")
+        if pkg_metadata:
+            scx_metadata = pkg_metadata.get("scx")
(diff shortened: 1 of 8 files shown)


.github/include/ci.py
```diff
@@ -0,0 +1,215 @@
...
+    cargo_results = [
+        json.loads(line) for line in result.stdout.splitlines() if line.strip()
+    ]
+    test_binaries = [
+        x["executable"]
+        for x in cargo_results
+        if x["reason"] == "compiler-artifact"
+        if x["profile"].get("test", False)
+        if x["executable"] is not None
+    ]
```
JakeHillion, line 104: Cargo doesn't have the equivalent of `ctest` that runs all tests in the build directory without building them, so this is the recommended alternative. It's a bit grim but won't break unless we update Cargo, and Cargo is pinned now so should be fine.

hodgesds (approved):

PR #1683, opened 2025-04-16
scx_lavd: Calculate scaled and invariant CPU utilization.

multics69:
Previously, neither CPU capacity nor frequency were considered when calculating CPU utilization. CPU utilization merely means the compute time ratio to a certain time duration. So, it is not possible to compare the utilization value of one CPU directly with that of another CPU.

The new scaled utilization (sys_stat.avg_sc_util and sys_stat.cur_sc_util) is normalized value to the highest frequency of the fastest CPU, so the scaled utilization values are safely compared across CPU types with different frequencies.

After this PR, the scaled, invariant utilization will be used for core compaction.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -674,6 +674,12 @@ static void update_stat_for_stopping(struct task_struct *p,
 		WRITE_ONCE(cur_svc_time, taskc->svc_time);
 
 	/*
+	 * Increase total scaled CPU time of this CPU,
+	 * whcih is capacity and frequency invariant.
+	 */
+	cpuc->tot_sc_time += scale_cap_freq(task_runtime, cpuc->cpu_id);
+
+	/*
 	 * Reset task's lock and futex boost count
 	 * for a lock holder to be boosted only once.
 	 */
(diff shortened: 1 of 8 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1317,20 +1323,52 @@ void BPF_STRUCT_OPS(lavd_set_cpumask, struct task_struct *p,
...
+	 * used -- i.e., [lavd_cpu_release, lavd_cpu_acquire]. This will be
+	 * used to calculate capacity-invariant and frequency-invariant CPU
+	 * utilization.
+	 */
+	dur = time_delta(scx_bpf_now(), cpuc->cpu_release_clk);
+	scaled_dur = scale_cap_freq(dur, cpu);
+	cpuc->tot_sc_time += scaled_dur;
+
+	/*
+	 * The higher-priority scheduler class could change the CPU frequency,
```
hodgesds, line 1350: This is pretty neat!

hodgesds (approved):

htejun (approved):

arighi (approved):

multics69:
Thank you @hodgesds @arighi and @htejun for the review!

PR #1686, opened 2025-04-16
scxtop: better naming for kernel workqueue threads

tyroguru:
Kernel workqueue threads have a comm field that always begins with "kworker/" and this isn't helpful at all for identifying the actual function of the thread which is super useful when analysing workloads. This change identifies workqueue threads and replaces the comm field we stash with the description of the thread that was provided when it was created. This idea and the bpf code was shamelessly stolen from the [systing](https://github.com/josefbacik/systing) project.

Just a note on the lib.rs changes (which @JakeHillion knows all about as he basically came up with them :-) ): the current lib.rs implementation pulls 16 bytes from the passed in event struct and uses `std::str::from_utf8()`  to construct a string which asserts if any of the bytes are not utf8. Using `bpf_probe_read_kernel_str()` to copy the comm value from a kernel worker struct exposes the fact that the target event struct may contain random garbage sometimes as it only copies up to and including the NULL terminator. We never previously hit this if we memcpy from a processes comm area as it is 0 filled in creation and therefore we always copy valid characters in the 16 bytes comm area. The fix is to switch to a CStr based solution that allows us to deal with invalid unicode bytes.

Workqueue threads before showed up like:

Screenshot 2025-04-16 at 19 09 14

The now look like:

Screenshot 2025-04-16 at 18 43 47

### DIFF
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -405,6 +405,23 @@ int BPF_KPROBE(scx_dispatch_from_dsq_set_vtime, struct bpf_iter_scx_dsq *it__ite
 	return on_move_set_vtime(NULL, vtime);
 }
 
+static void record_real_comm(char *comm, struct task_struct *task)
+{
+	if (task->flags & PF_WQ_WORKER) {
+		/*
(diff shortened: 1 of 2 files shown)


hodgesds (approved):
Nice!

PR #1687, opened 2025-04-16
scx_layered: add timeout ms flag

likewhatevs:
Add timeout-ms flag to layered and pass timeout ms when turning on scheduler. Set this to the default (30000). This enables setting timeout-ms via layered.

image


htejun:
@JakeHillion wanted to add something similar but across all schedulers. That probably is better?

likewhatevs:
SGTM -- will re-open if we need this before that's available. So long as this doesn't add any problematic dependencies, however this is added should work fine due to the constraints layered being published on crates.io imposes, I think.

PR #1689, opened 2025-04-17
scx_layered: add antistall_aggressiveness

likewhatevs:
Add flag `--antistall-aggressiveness`, which takes a uint.

This flag sets a multiplier on the number of CPUs anti-stall dedicates to clearing stuck DSQs each pass.

By default, anti-stall selects 1 CPU per stuck DSQ each pass, and each pass happens every second. This is fine for smaller systems with more well-behaved workloads (i.e. workloads w/ a few affined tasks), but falls over on huge systems with less well-behaved workloads (i.e. workloads with enough affined tasks that much, much more CPU time than lo frac provides *may* be needed to prevent any task from triggering a stall watchdog).

This is kind-of a half-way measure vs increasing lo frac. 

Not sure about the name (it is kind of long).

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -60,6 +60,7 @@ const volatile bool enable_gpu_support = false;
 const volatile u64 antistall_sec = 3;
 /* Multiplier on CPUs antistall dedicates to clearing stalls per run. */
 const volatile u32 antistall_aggressiveness = 1;
+const volatile u64 antistall_slice_override = 0;
 const u32 zero_u32 = 0;
 
 private(unprotected_cpumask) struct bpf_cpumask __kptr *unprotected_cpumask;
@@ -1587,7 +1588,8 @@ bool antistall_consume(struct cpu_ctx *cpuc)
 {
 	u64 *antistall_dsq, jiffies_now, cur_delay;
 	bool consumed;
-	struct task_struct *p;
+	struct task_struct *p = NULL;
+	struct layer *layer;
 
 	cur_delay = 0;
 	consumed = false;
@@ -1604,6 +1606,19 @@ bool antistall_consume(struct cpu_ctx *cpuc)
 
 	if (*antistall_dsq == SCX_DSQ_INVALID)
 		return false;
+
+	if (antistall_slice_override != 0) {
+		if (!(layer = lookup_layer(cpuc->layer_id))) {
+			scx_bpf_error("cant happen");
+			return consumed;
+		}
+		// override slice size while in antistall.
+		bpf_for_each(scx_dsq, p, *antistall_dsq, 0) {
+			if (p->scx.slice == antistall_slice_override)
+				break;
+			p->scx.slice = antistall_slice_override;
+		}
+	}
 
 	consumed = scx_bpf_dsq_move_to_local(*antistall_dsq);
 
@@ -1625,6 +1640,12 @@ bool antistall_consume(struct cpu_ctx *cpuc)
 	}
 
 reset:
+	if (antistall_slice_override != 0) {
+		bpf_for_each(scx_dsq, p, *antistall_dsq, 0) {
+			p->scx.slice = layer->slice_ns;	
+		}
+	}
+	
 	trace("antistall reset DSQ[%llu] SELECTED_CPU[%llu] DELAY[%llu]",
 	      *antistall_dsq, cpuc->cpu, cur_delay);
 	*antistall_dsq = SCX_DSQ_INVALID;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -58,6 +58,8 @@ const volatile bool enable_antistall = true;
 const volatile bool enable_gpu_support = false;
 /* Delay permitted, in seconds, before antistall activates */
 const volatile u64 antistall_sec = 3;
+/* Multiplier on CPUs antistall dedicates to clearing stalls per run. */
+const volatile u32 antistall_aggressiveness = 1;
 const u32 zero_u32 = 0;
 
 private(unprotected_cpumask) struct bpf_cpumask __kptr *unprotected_cpumask;
@@ -2976,6 +2978,7 @@ static bool antistall_scan(void)
 	s32 llc;
 	u64 layer_id;
 	u64 jiffies_now;
+	u32 antistall_multiplier;
 
 	if (!enable_antistall)
 		return true;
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2924,36 +2927,39 @@ u64 antistall_set(u64 dsq_id, u64 jiffies_now)
...
+						scx_bpf_error("cant happen");
+						goto unlock;
+					}
+
+					if ((pass == 0 && *antistall_dsq == SCX_DSQ_INVALID) ||
+			    		(pass != 0 && *delay < cur_delay)) {
+						trace("antistall set DSQ[%llu] SELECTED_CPU[%llu] DELAY[%llu]", dsq_id, cpu, cur_delay);
+						*delay = cur_delay;
+						*antistall_dsq = dsq_id;
+						if (antistall_aggressiveness <= 0)
```
JakeHillion, line 2957: This is querying the constant which doesn't seem to be correct. I think the idea is that if the loop variable is on the last iteration terminate, but I'm still not sure that's right. We likely want to break out if the innermost `bpf_for` but still have another go, which I think means this unlock case needs to go.

JakeHillion (commented):
A few comments on the code, but I'm not sure this works fundamentally. Could we see a test plan for it?

Although we now repeat the loop `antistall_aggressiveness` number of times, we still only set a single antistall DSQ for each CPU. I don't think this actually enables any more DSQs to be set, unless we get really lucky and race with dispatch and this repeating. It seems like it would enable the same behaviour with fewer races to reduce the timer instead. Please correct me if I've misunderstood.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: ahh yeah that should be multiplier

likewhatevs:
> Although we now repeat the loop antistall_aggressiveness number of times, we still only set a single antistall DSQ for each CPU. I don't think this actually enables any more DSQs to be set, unless we get really lucky and race with dispatch and this repeating. It seems like it would enable the same behaviour with fewer races to reduce the timer instead. Please correct me if I've misunderstood.

Rewrote this making it easier to read. It's pretty much the same thing as reducing the timer but with less interim steps/ w/o the timer.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2976,17 +2978,19 @@ static bool antistall_scan(void)
...
 	if (!enable_antistall)
 		return true;
 
 	jiffies_now = bpf_jiffies64();
 
 	bpf_for(layer_id, 0, nr_layers)
-		bpf_for(llc, 0, nr_llcs)
-			antistall_set(layer_dsq_id(layer_id, llc), jiffies_now);
+		bpf_for(llc, 0, nr_llcs) 
+			for(antistall_multiplier = antistall_aggressiveness; antistall_multiplier < 0; --antistall_multiplier)
```
etsal: (Possible repost not sure the comment got submitted) The condition should be flipped otherwise it never runs (antistall_aggressiveness is unsigned and so > 0). AFAICT it is also equivalent to the more conventional

```
for(antistall_multiplier 0; antistall_multiplier < antistall_aggressiveness; antistall_multiplier++)
```

likewhatevs: thank you, I made that <= so it'll decrement to 0 which would do the right thing I think

JakeHillion (commented):
I still don't understand the motivation for this change. The refactor makes it clearer, but I don't think it changed the outcomes. Here's my understanding of the behaviour:

`antistall_set` now runs in a tight loop for each DSQ. I don't see how this helps. `antistall_set` may or may not set every CPU a task can run on's antistall_dsq to the value of the DSQ it's passed. The next dispatch on those CPUs will leave that queue in place as long as the foremost task is stalling, so adding it 2/3 times makes no difference.

In contrast leaving `antistall_sec` the same and increasing the frequency of the timer would allow a removed DSQ to come back more quickly: it does better than a tight loop, but I still don't think it solves the problem.

I think the correct fix here, other than obviously not placing tasks into queues they can't run in in the first place, is to somehow randomise the DSQs each time the timer runs to ensure each gets a fair chance to become an antistall_dsq. As far as I can tell the current behaviour prioritises the last DSQ (modulo race conditions) which would suggest a lot of affinitised tasks still have a chance to stall. What do you think?

likewhatevs:
> antistall_set now runs in a tight loop for each DSQ. I don't see how this helps. antistall_set may or may not set every CPU a task can run on's antistall_dsq to the value of the DSQ it's passed. The next dispatch on those CPUs will leave that queue in place as long as the foremost task is stalling, so adding it 2/3 times makes no difference.

Antistall set flags CPUs to do antistall work each time it's run. This is done in an accumulating manner. By multiplying the number of runs of antistall_set, we multiply the number of eligible CPUs which are set to do antistall work each time the timer runs.

> I think the correct fix here, other than obviously not placing tasks into queues they can't run in in the first place, is to somehow randomise the DSQs each time the timer runs to ensure each gets a fair chance to become an antistall_dsq. As far as I can tell the current behaviour prioritises the last DSQ (modulo race conditions) which would suggest a lot of affinitised tasks still have a chance to stall. What do you think?

In antistall scan, every DSQ is processed every run of the timer. The thing to randomize (if needed, because this would probably be detrimental for perf), would be the order in which CPUs are flagged to do antistall work during the first pass of CPU selection (i.e. where CPU selection is linear and based upon eligibility instead of current delay).

JakeHillion:
> Antistall set flags CPUs to do antistall work each time it's run. This is done in an accumulating manner. By multiplying the number of runs of antistall_set, we multiply the number of eligible CPUs which are set to do antistall work each time the timer runs.

Good, I agree, thanks! This means that there is a higher chance more CPUs will be selected in each pass.

> In antistall scan, every DSQ is processed every run of the timer. The thing to randomize (if needed, because this would probably be detrimental for perf), would be the order in which CPUs are flagged to do antistall work during the first pass of CPU selection (i.e. where CPU selection is linear and based upon eligibility instead of current delay).

It's good, there is already the check for greatest delay which I think sorts that out. As long as we are choosing the highest delay we should eventually get everything. I'm still a bit concerned about this change because if a task has N-1 affinity and is the most delayed it will be placed as the antistall_dsq for min(N-1, antistall_aggressiveness) CPUs and potentially cause some starvation if that number is too big, but I think it's mostly fine.

What I still don't understand is the goal here. Is it to eliminate 30s stalls that get the scheduler kicked out, or something else? I think it now bumps the speed at which a stalling DSQ is cleared, but only after that task has already been delayed for 3s - is this important? I don't see how this reduces the chance of 30s stalls at all, because antistall_consume already deals with the entire stalling part of the queue before removing it from a CPU.

likewhatevs:
> It's good, there is already the check for greatest delay which I think sorts that out. As long as we are choosing the highest delay we should eventually get everything. I'm still a bit concerned about this change because if a task has N-1 affinity and is the most delayed it will be placed as the antistall_dsq for min(N-1, antistall_aggressiveness) CPUs and potentially cause some starvation if that number is too big, but I think it's mostly fine.

I don't follow that exact example, but if you're saying "if the multiplier is too big, and antistall fully drains a layer of CPUs, then antistall would ping-pong on most delayed DSQs/eligible CPUs", yes but that's also kind-of intended (until the overload passes/policy and workload work together again).

> What I still don't understand is the goal here. Is it to eliminate 30s stalls that get the scheduler kicked out, or something else? I think it now bumps the speed at which a stalling DSQ is cleared, but only after that task has already been delayed for 3s - is this important? I don't see how this reduces the chance of 30s stalls at all, because antistall_consume already deals with the entire stalling part of the queue before removing it from a CPU.

To avoid sched kicks/stalls when a layer has 1 CPU and 80000 tasks on it.

htejun:
I may be wrong but am a bit doubtful how useful this would be. There may be cases that this can help with but the conditions have to be just right where the system isn't too overloaded for this to not matter but overloaded enough for a short enough duration that cycling through tasks faster can avoid the scheduler getting kicked out.

Taking a step back. I don't think it's generally a good idea to aim for avoiding the scheduler getting kicked out. The watchdog is already porous and the system can accidentally be rendered unusable while avoiding triggering the watchdog. If we implement features and tune parameters to avoid watchdog, that will increase the chance hitting that window where the system is almost completely useless but the scx watchdog can't rescue it. Watchdog triggering in an indication that there is a larger underlying problem and it's fortunate that the underlying problem triggered the watchdog instead of stalling out the system for prolonged duration.

In general, if we can't spend a reasonable amount of time slice per runnable task in a few seconds, something is already really wrong. We should solve that instead of avoiding the watchdog.

PR #1690, opened 2025-04-17
scx_layered: Implement sticky modulation optimization

kkdwvd:


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -871,6 +877,47 @@ s32 pick_idle_big_little(struct layer *layer, struct task_ctx *taskc,
...
+	u64 time = bpf_ktime_get_ns();
+	const struct cpumask *cpumask;
+	struct cpu_ctx *cpu_ctx;
+	s32 cpu = -1;
+	int i;
+
+	if (!active_sticky_mod)
+		return cpu;
+
+	cpu_ctx = lookup_cpu_ctx(prev_cpu);
```
etsal, line 892: Nit: Factoring out into a small inline function can avoid all the gotos.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1195,6 +1245,55 @@ static void layer_kick_idle_cpu(struct layer *layer)
 	scx_bpf_put_idle_cpumask(idle_smtmask);
 }
 
+SEC("tp_btf/sched_switch")
```
etsal, line 1248: What is the overhead of the extra probe? Both in terms of having an extra probe and in terms of the code itself. Can we roll this into our starting/stopping methods?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -871,6 +877,47 @@ s32 pick_idle_big_little(struct layer *layer, struct task_ctx *taskc,
...
+		goto llc;
+	if (cpu_ctx->sticky_mod_pred_pct < layer->sticky_mod_pred_pct)
+		goto llc;
+	if (cpu_ctx->sticky_mod_end_time_ns - time > layer->sticky_mod_min_ns)
+		goto llc;
+	return prev_cpu;
+llc:
+	if (!(cpumask = cast_mask(llc->cpumask)))
+		goto out;
+	bpf_for(i, 0, nr_possible_cpus) {
```
etsal, line 903: (Comment: I'm surprised we don't have an FFS based foreach cpu in cpumask primitive)

etsal:
Looks good overall, my main nit would be rolling the stat update logic into our own methods instead of a sched_switch-based probe, it is both cleaner (doesn't grab tasks that belong to other schedulers) and more consistent

scheds/rust/scx_layered/src/bpf/main.bpf.c
kkdwivedi: The probe itself should be pretty fast, but I think longer term hooking into starting/stopping methods is the better way, so I will do that.

scheds/rust/scx_layered/src/bpf/main.bpf.c
kkdwivedi: Yeah, if there's a better way to do this, I'm all ears. It's suboptimal, especially on machines with a lot of CPUs. Even on Bergamo, we're iterating over 88 CPUs for every 8 we want to test.

kkdwivedi: One thing that comes to mind is storing start and end CPUs in llc_ctx, but that requires the assignment to be sequential.

likewhatevs:
IIUC the default is what is done currently/who this flag, right? If so this is good.

Could you update the documentation to suggest like, what a reasonable default is, or set the default to disabled and add an enable flag which when passed defaults to a reasonable value?

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: layered stores iteration indices in per-cpu/llc arrays. It's a bit more setup work but overall not that bad. But yeah, bit-wise iterators would be great.

htejun:

PR #1692, opened 2025-04-17
scx_p2dq: Add configuration for dispatch load balancing

hodgesds:
Make load balancing configurable on the dispatch path by setting the dispatch_lb_busy value. If the load proportion is under the busy value then allow the CPU to migrate tasks across LLCs.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -57,6 +57,7 @@ const volatile bool kthreads_local = true;
 const volatile bool max_dsq_pick2 = false;
 const volatile bool select_idle_in_enqueue = true;
 
+const volatile u64 dispatch_lb_busy = 75;
 const volatile u64 wakeup_lb_busy = 90;
 const volatile u64 lb_slack_factor = 5;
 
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -330,16 +332,15 @@ static struct llc_ctx *pick_two_llc_ctx(struct llc_ctx *cur_llcx, struct llc_ctx
...
+	if (min_nr_queued_pick2 > 0 && cur_queued < min_nr_queued_pick2)
 		return NULL;
 
 	left_load = left->load;
 	right_load = right->load;
 
 	// If the current LLCs has more load don't try to pick2.
 	cur_load += (lb_slack_factor * cur_load) / 100;
-	if ((nr_llcs > 2 && cur_load > left_load && cur_load > right_load))
+	if ((nr_llcs > 2 && (cur_load > left_load || cur_load > right_load)))
```
JakeHillion, line 343: This is an interesting change. Previously we'd have picked from the more heavily loaded option of the two, now we pick from neither if we're in the middle. Seems to make sense and reduce migrations.

scheds/rust/scx_p2dq/src/lib.rs
```diff
@@ -53,6 +53,11 @@ pub struct SchedulerOpts {
     #[clap(short = 'd', long, action = clap::ArgAction::SetTrue)]
     pub dispatch_pick2_disable: bool,
 
+    /// Enables pick2 load balancing on the dispatch path when LLC utilization is under the
+    /// specified utilization.
+    #[clap(long, default_value = "75", value_parser = clap::value_parser!(u64).range(0..99))]
```
JakeHillion, line 58: Is it worth allowing 100 to make this path always true?

JakeHillion (approved):

scheds/rust/scx_p2dq/src/lib.rs
hodgesds: Good call

PR #1693, opened 2025-04-17
scx_bpfland: improve wakeups

arighi:
Allow direct dispatch on sync wakeup only within the same LLC (to reduce LLC migrations that can be costly on certain architectures - e.g., AMD).

Based on tests and benchmarks from the CachyOS community, these changes don't appear to bring significant improvements, but the overall trend seems positive, with small performance gains.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -442,11 +442,39 @@ static void task_update_domain(struct task_struct *p, struct task_ctx *tctx,
 		bpf_cpumask_and(l3_mask, cast_mask(p_mask), cast_mask(l3_domain));
 }
 
-static bool is_wake_sync(const struct task_struct *p,
-			 const struct task_struct *current,
-			 s32 prev_cpu, s32 cpu, u64 wake_flags)
+/*
+ * Return true if all the CPUs in the LLC of @cpu are busy, false
+ * otherwise.
+ */
+static bool is_llc_busy(const struct cpumask *idle_cpumask, s32 cpu)
 {
-	if (wake_flags & SCX_WAKE_SYNC)
+	const struct cpumask *primary, *l3_mask;
+	struct cpu_ctx *cctx;
+
+	primary = cast_mask(primary_cpumask);
+	if (!primary)
+		return false;
+
+	cctx = try_lookup_cpu_ctx(cpu);
+	if (!cctx)
+		return false;
+
+	l3_mask = cast_mask(cctx->l3_cpumask);
+	if (!l3_mask)
+		l3_mask = primary;
+
+	return !bpf_cpumask_intersects(l3_mask, idle_cpumask);
+}
+
+/*
+ * Return true if the waker commits to release the CPU after waking up @p,
+ * false otherwise.
+ */
+static bool is_wake_sync(s32 prev_cpu, s32 this_cpu, u64 wake_flags)
+{
+	const struct task_struct *current = (void *)bpf_get_current_task_btf();
+
+	if ((wake_flags & SCX_WAKE_SYNC) && !(current->flags & PF_EXITING))
 		return true;
 
 	/*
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
LGTM

arighi:
Actually... let's wait a bit before merging this, I'm seeing regressions in kernel builds with the 2nd commit applied (I feel like the wakeup from IRQ optimization brings more overhead than benefits).

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -527,50 +592,57 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
...
-		share_llc = bpf_cpumask_test_cpu(prev_cpu, curr_l3_domain);
-		if (share_llc &&
-		    scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
-			cpu = prev_cpu;
+		if ((!smt_enabled || bpf_cpumask_test_cpu(this_cpu, idle_smtmask)) &&
+		    scx_bpf_test_and_clear_cpu_idle(this_cpu)) {
+			cpu = this_cpu;
 			*is_idle = true;
 			goto out_put_cpumask;
 		}
```
hodgesds, line 634: Might be worth trying here if nothing is found to return a CPU in the LLC anyways? This is the part that I tried testing and couldn't get good results on.

arighi: Yeah that's the culprit, it's more or less what fair.c is doing, but in our case it seems to introduce more overhead than benefits. Maybe it's good in some special cases, but it seems a bit too situational.

I think it makes sense to return any CPU in the LLC, but that means removing this chunk of code and fallback to the regular path of finding an idle CPU. So... basically we can drop the last commit and we get better performance in general.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -455,15 +483,43 @@ static bool is_wake_sync(const struct task_struct *p,
...
+	 */
+	if ((this_cpu != prev_cpu) && scx_bpf_test_and_clear_cpu_idle(prev_cpu))
+		return prev_cpu;
+
+	/*
+	 * If waker and wakee are on the same CPU and no other tasks are
+	 * queued, consider the waker's CPU as idle.
+	 */
+	if (!scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | this_cpu))
+		return this_cpu;
```
multics69, line 527: Testing the waker's idleness this way is because the waker's CPU is not idle always. Am I right?

multics69 (approved):

multics69:
> Actually... let's wait a bit before merging this, I'm seeing regressions in kernel builds with the 2nd commit applied (I feel like the wakeup from IRQ optimization brings more overhead than benefits).

Regarding the second commit, the logic looks reasonable to me. If it causes some regressions, would it be because a single waker wakes up multiple wakees, hampering cache hotness and load balancing?

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Exactly, the waker is still running, but it made a promise to release the CPU right after the wakeup (because SCX_WAKE_SYNC is set), so the CPU is not idle, but it'll be immediately after the wakeup, therefore this CPU is basically idle for the wakee. It's like a "function call" between two tasks.

arighi:
> > Actually... let's wait a bit before merging this, I'm seeing regressions in kernel builds with the 2nd commit applied (I feel like the wakeup from IRQ optimization brings more overhead than benefits).
> 
> Regarding the second commit, the logic looks reasonable to me. If it causes some regressions, would it be because a single waker wakes up multiple wakees, hampering cache hotness and load balancing?

For what I see this might help in some rare cases, but often there's not a strong relationship between the IRQ CPU and the wakee CPU (in terms of caching) and in general prev_cpu has better hot cache data.

I think fair.c does this for historical reasons (looking at the git log this logic to move a task to the IRQ CPU seems pretty old) and it was actually mitigated at some point to limit the migration within the LLC, see this commit by Peter: https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=7332dec055f2457c386032f7e9b2991eb05c2a0a

Just looking at the beginning of this commit message:
```
If waking from an idle CPU due to an interrupt then it's possible that
the waker task will be pulled to wake on the current CPU. Unfortunately,
depending on the type of interrupt and IRQ configuration, there may not
be a strong relationship between the CPU an interrupt was delivered on
and the CPU a task was running on. For example, the interrupts could all
be delivered to CPUs on one particular node due to the machine topology
or IRQ affinity configuration. Another example is an interrupt for an IO
completion which can be delivered to any CPU where there is no guarantee
the data is either cache hot or even local.
```

It confirms that the benefits of this optimization are very situational and in this case the LLC limitation helped, because doing a cross-node migration for uncertain benefits is not that great.

What I see in my tests is that if the system is pretty much idle, it makes no measurable difference (probably in modern CPUs once you're within the same LLC, picking a CPU or another doesn't make much difference). However, when the system is busy, the extra logic just adds more overhead, it's tiny, but consistent and measurable, in a parallel kernel build for example it's around 0.6% (in the system that I've used).

Therefore, I'll update this PR to drop the second commit.

PS it'd be interesting to try to drop this logic also from fair.c and see what happens, will do some tests with this as well later.

arighi:
Alright, dropped the 2nd commit and updated PR description.

This one seems good according to my tests. Limiting WAKE_SYNC migrations within the LLC seems beneficial: we can avoid costly cross-LLC / cross-node migrations and for single-LLC / single-node systems there's no extra overhead.

PR #1694, opened 2025-04-17
scxtop: fixing missing process names

tyroguru:
We now have process -> thread relationships setup so the grouping are good. However, for longer running processes it is quite common for the main thread to not run at all while we are tracing and this means we never get to capture its comm. This makes it hard to identify the process to which a bunch of threads belong as any `ProcessDescriptors` objects which have the `process_name` Option unset causes the process thread (the thread group leader) to be labelled as "Process" in the resulting perfetto trace.

The solution here is to simply go lookup the `comm` value in `/proc` if it exists when scxtop exits. I'm not worried about pid reuse as trace times are too short for it to be a problem.

An example of before the changes shows the current state with lots of processes being labelled with the default label:

Screenshot 2025-04-17 at 18 42 08

Now virtually every process is labeled correctly (at least on my development machine).

Screenshot 2025-04-17 at 18 44 20

### DIFF
--- a/tools/scxtop/src/perfetto_trace.rs
+++ b/tools/scxtop/src/perfetto_trace.rs
@@ -248,6 +248,12 @@ impl PerfettoTraceManager {
         }
     }
 
+    fn get_comm(pid: i32) -> Result<String> {
+        let path = format!("/proc/{}/comm", pid);
+        let content = fs::read_to_string(path)?;
+        Ok(content.trim_end().to_string())
+    }
+
     /// Stops the trace and writes to configured output file.
     pub fn stop(
         &mut self,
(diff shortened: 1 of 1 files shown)


tools/scxtop/src/perfetto_trace.rs
```diff
@@ -248,6 +249,12 @@ impl PerfettoTraceManager {
         }
     }
 
+    fn get_comm(pid: i32) -> Result<String> {
```
hodgesds, line 252: nit: there's already a [helper function](https://github.com/sched-ext/scx/blob/fd88ee611130433371c37474f449985abb975d35/tools/scxtop/src/util.rs#L11)

hodgesds (approved):

tools/scxtop/src/perfetto_trace.rs
tyroguru: Thanks @hodgesds for pointing this out! I'll try and refactor some time to use that interface.

PR #1699, opened 2025-04-18
scx_lavd: Use scaled CPU utilization for calculating # of active CPUs.

multics69:
Calculate the number of active CPUs using the scaled CPU utilization. The scaled CPU utilization is capacity-invariant and frequency-invariant, so this approach works better with heterogeneous processor architecture. By this, ad-hoc heuristics (e.g., nr_violation) were dropped.

In addition, two changes were added to quickly change the # of active CPUs upon load spikes without much fluctuation.


hodgesds (approved):
LGTM

scheds/rust/scx_lavd/src/bpf/power.bpf.c
```diff
@@ -99,6 +80,49 @@ static const volatile u16 *get_cpu_order(void)
...
+static int calc_nr_active_cpus(void)
+{
+	const volatile u16 *cpu_order;
+	u64 req_cap, cap_cpu, cap_sum = 0;
+	u16 cpu_id, i;
+
+	/*
+	 * Calculate the required compute capacity:
+	 *
+	 * Scaled utilization assumes all the CPUs are the fastest ones
```
hodgesds, line 92: I think this is probably a good enough assumption to make for now.

multics69: Thanks @hodgesds for the review!

PR #1701, opened 2025-04-18
scx_bpfland: Introduce --no-wake-sync

arighi:
Certain workloads (especially games) seem to benefit from the synchronous wakeup optimization, which tries to dispatch a wakee on the waker’s CPU when both share the same LLC. This can improve cache locality and responsiveness in typical producer/consumer pipelines, where the same two tasks communicate frequently.

However, in other workloads, this behavior can lead to performance regressions and excessive CPU migrations. In particular, it can be detrimental in:

 - Mixed workloads with latency-sensitive and background pipe-intensive tasks (e.g., telemetry or logging daemons), where the latter are chatty but not performance-critical. Prioritizing them via direct dispatch skews fairness and may hurt important foreground tasks.

 - Many-to-one communication patterns, such as a centralized logger, being woken by many different producers. With the synchronous wakeup optimization, the consumer may bounce across CPUs trying to follow each producer, causing unnecessary migrations and disrupting CPU locality.

 - One-to-many patterns, such as a producer waking multiple consumers (e.g., multi-threaded services or a poll loop across many readers). Although each wakeup targets a single task, the aggregate behavior leads to similar instability and loss of locality when the waker is always favored.

 - Power-saving oriented cpufreq governors (e.g., balance_power with intel_pstate): disabling synchronous wakeups can promote a more balanced load across CPUs, helping to avoid overly aggressive frequency downscaling. This can lead to more consistent performance in interactive or graphics-intensive workloads.

To accommodate all these cases, introduce the --no-wake-sync option that allows to disable direct dispatch on synchronous wakeups. This can improve load distribution and overall performance in all these cases where prioritizing synchronous wakeups is counterproductive.

By default, the optimization remains enabled, as the majority of gaming and desktop workloads see clear gains from it, while the advantages of disabling it tend to be more rare and situational.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -43,6 +43,11 @@ const volatile s64 slice_lag = 20ULL * NSEC_PER_MSEC;
 const volatile bool no_preempt;
 
 /*
+ * Ignore synchronous wakeup events.
+ */
+const volatile bool no_wake_sync;
+
+/*
  * When enabled always dispatch per-CPU kthreads directly.
  *
  * This allows to prioritize critical kernel threads that may potentially slow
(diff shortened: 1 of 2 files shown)


hodgesds (approved):

multics69 (approved):

multics69:
Thanks @arighi for the detailed explanation. It helps a lot devising a better policy.

PR #1703, opened 2025-04-18
scx_lavd: Improve preemption at ops.tick() and ops.qneueue().

multics69:
The preemption at ops.tick() is cheap because it does not require expensive IPI. However, the quality of the preemption decision is suboptimal using only local information. It is insufficient to check that the current task is less latency-critical than the one in a DSQ. Even if it is true, there is no guarantee that the current task is the least latency-critical among currently running tasks. Moreover, the preemption at ops.tick() is rather redundant with the preemption at ops.qneueue(). The poor, redundant preemption decision pressures ops.enqueue() and ops.dispatch() operations. So, drop the preemption at ops.tick().

On the other hand, the preemption at the ops.enqueue() is more expensive. However, the quality of the decision is better because it tries to find the least latency-critical tasks among the running tasks using the Power of Two Random Choices technique. To avoid unnecessary preemption at ops.enqueue(), make the preemption decision more judiciously by tightening the preemption condition.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -1010,25 +1010,6 @@ consume_out:
 		prev->scx.slice = calc_time_slice(taskc);
 }
 
-void BPF_STRUCT_OPS(lavd_tick, struct task_struct *p)
-{
-	struct cpu_ctx *cpuc;
-	struct task_ctx *taskc;
-	u64 now;
-
-	/*
-	 * For a non-latency-critical task, try to yield the current CPU if
-	 * there is a higher priority task in the run queue.
-	 */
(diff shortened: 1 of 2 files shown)


hodgesds:
>  To avoid unnecessary preemption at ops.enqueue(), make the preemption decision more judiciously by tightening the preemption condition.

This is similar to what I observed with trying to load balancing in the select/enqueue paths. The decisions need to be made fast or scheduling latency becomes an issue.

hodgesds (approved):

multics69:
> > To avoid unnecessary preemption at ops.enqueue(), make the preemption decision more judiciously by tightening the preemption condition.
> 
> This is similar to what I observed with trying to load balancing in the select/enqueue paths. The decisions need to be made fast or scheduling latency becomes an issue.

Hehe. Your that change was my motivation to take this look at again.  :-) Thanks!

PR #1704, opened 2025-04-19
scx_lavd: Skip load balancing when there is only one compute domain.

multics69:
When there is only one compute domain, there is only one DSQ so load balancing is unnecessary. So skip the load balancing when nr_cpdoms == 1.


hodgesds (approved):
Makes sense

PR #1708, opened 2025-04-19
scx_utils: Add mangolog module

hodgesds:
Add module for reading from mangoapp via gamescope. This module can be used to gather real time data for gaming.

Output from running the example:
```
~/.local/share/Steam/steamapps/common/Counter-Strike Global Offensive $  ~/git/scx/target/release/examples/mangolog
Received MangoApp data:
  Visible Frametime: 4753886
  FSR Upscale: 0
  FSR Sharpness: 2
  App Frametime: 4753886
  Latency: 18446744073709551615
  PID: 27004
  Output Width: 1920
  Output Height: 1177
  App Wants HDR: false
  Steam Focused: false
  Engine Name:
Received MangoApp data:
  Visible Frametime: 5128282
  FSR Upscale: 0
  FSR Sharpness: 2
  App Frametime: 5128282
  Latency: 18446744073709551615
  PID: 27004
  Output Width: 1920
  Output Height: 1177
  App Wants HDR: false
  Steam Focused: false
  Engine Name:
  ```
  I may integrate this library into `scxtop` for collecting gaming/scheduling stats.

### DIFF
--- a/rust/scx_utils/examples/mangolog.rs
+++ b/rust/scx_utils/examples/mangolog.rs
@@ -1,0 +1,91 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+use scx_utils::mangoapp::mangoapp_msg_v1;
+
+use anyhow::bail;
+use anyhow::Result;
+use libc::{msgget, msgrcv, IPC_NOWAIT};
+
+use std::mem;
+use std::path::Path;
+use std::thread;
+use std::time::Duration;
+use std::{ffi::CString, io};
+
+fn main() -> Result<()> {
+    // NOTE: the mangoapp file has to be present when gamescope launches in order to read from the
+    // message queue. For example when launching Counter Strike 2 the mangoapp file needs to be in
+    // $HOME/.local/share/Steam/steamapps/common/Counter-String Global Offensive/mangoapp and the
+    // program needs to run in the same directory as well.
+    let file_path = Path::new("mangoapp");
+    if !file_path.exists() {
+        bail!("mangoapp file does not exists");
+    }
+
+    // Create the key for msgget reads using the mangoapp file
+    let path = CString::new("mangoapp").unwrap();
+    let key = unsafe { libc::ftok(path.as_ptr(), 65) };
+    if key == -1 {
+        bail!("failed to ftok: {}", io::Error::last_os_error());
+    }
+
+    // Get the key from the queue
+    let msgid = unsafe { msgget(key.try_into().unwrap(), 0) };
+    if msgid == -1 {
+        bail!("msgget failed: {}", std::io::Error::last_os_error());
+    }
+
+    let mut raw_msg: mangoapp_msg_v1 = unsafe { mem::zeroed() };
+
+    loop {
+        let msg_size = unsafe {
+            msgrcv(
+                msgid,
+                &mut raw_msg as *mut _ as *mut libc::c_void,
+                mem::size_of::<mangoapp_msg_v1>() - mem::size_of::<i64>(),
+                0,
+                IPC_NOWAIT, // XXX: this should probably use MSG_COPY as it pulls messages off the
+                            // queue and may mess with mangohud or other mangoapp uses.
+            )
+        };
+
+        if msg_size as isize != -1 {
+            let frametime = raw_msg.visible_frametime_ns;
+            let upscale = raw_msg.fsr_upscale;
+            let sharpness = raw_msg.fsr_sharpness;
+            let app_frametime_ns = raw_msg.app_frametime_ns;
+            let pid = raw_msg.pid;
+            let latency_ns = raw_msg.latency_ns;
+            let output_width = raw_msg.output_width;
+            let output_height = raw_msg.output_height;
+            let wants_hdr = raw_msg.wants_hdr();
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: rust/scx_utils/examples/mangolog.rs [hunk 1]
```diff
@@ -0,0 +1,91 @@
...
+    let mut raw_msg: mangoapp_msg_v1 = unsafe { mem::zeroed() };
+
+    loop {
+        let msg_size = unsafe {
+            msgrcv(
+                msgid,
+                &mut raw_msg as *mut _ as *mut libc::c_void,
+                mem::size_of::<mangoapp_msg_v1>() - mem::size_of::<i64>(), // Exclude msg_type
+                0,
+                IPC_NOWAIT,
```
likewhatevs, line 48: I think you want MSG_COPY or'ed with IPC_NOWAIT kinda like here btw: 

https://github.com/likewhatevs/scx_tsatw/blob/c84f0b0b8fcdc61f1e484d3d97ef2bfcb7ef0ea0/util.c#L56

Without that this could mess w/ mangohud output/logs by pulling msgs off of the queue instead of just snooping on them.

Also a wrapper like https://github.com/likewhatevs/scx_tsatw/blob/c84f0b0b8fcdc61f1e484d3d97ef2bfcb7ef0ea0/util.c#L59 might make sense to help with polling (it's kinda wild how fast those IPC msgs are pulled from the queue by mangohud tbh).

rust/scx_utils/examples/mangolog.rs
```diff
@@ -0,0 +1,91 @@
...
+use libc::{msgget, msgrcv, IPC_NOWAIT};
+
+use std::mem;
+use std::path::Path;
+use std::thread;
+use std::time::Duration;
+use std::{ffi::CString, io};
+
+fn main() -> Result<()> {
+    // NOTE: the mangoapp file has to be present when gamescope launches in order to read from the
```
likewhatevs, line 18: nice trick, although I only tested this with mangohud enabled. IDK if that data is emitted from gamescope when the overlay is off, that might be worth testing idk.

likewhatevs (commented):
LGTM -- cool making this available to rust scheds, just maybe make it more clear this could mess w/ mangohud output/logs or avoid doing that via MSG_COPY.

rust/scx_utils/examples/mangolog.rs
likewhatevs: * Ahh IIRC getting passing MSG_COPY to msgrcv working depends on `-D_GNU_SOURCE` and IDK how one enables that in rust (i.e. maybe a comment noting that this pulls msgs off the queue instead of snooping is sufficient etc.).

rust/scx_utils/examples/mangolog.rs
hodgesds: It does seem to work without `mangohud` with
```
gamescope -e -W 1920 -H 1080 -r 60 --mangoapp -- %command%
```
with `mangohud` via:
```
gamescope -e -W 1920 -H 1080 -r 60 --mangoapp -- mangohud %command%
```
It was super annoying to figure that out

rust/scx_utils/examples/mangolog.rs
hodgesds: I think that's right, but for some reason it fails to parse messages with it set:
```
$  ~/git/scx/target/release/examples/mangolog
Error: mangoapp: msgrcv returned -1 with error Function not implemented (os error 38)
```
I'll leave a note in, since this is just example code.

multics69 (commented):
As you know, the data presented in the MangoHud can be configured using the config file. I wonder if the message format remains the same regardless the configuration.

hodgesds:
> As you know, the data presented in the MangoHud can be configured using the config file. I wonder if the message format remains the same regardless the configuration.

I believe it does, the structs were converted from [mangohud](https://github.com/flightlessmango/MangoHud/blob/master/src/app/mangoapp_proto.h) and [gamescope](https://github.com/ValveSoftware/gamescope/blob/1ab8009d5dc5faaff5d890ef896483ef14363536/src/mangoapp.cpp#L15). I believe mangohud handles the control messages, would have to dig through the code a bit to find how they are handled.

rust/scx_utils/examples/mangolog.rs
likewhatevs: Ahh when I say mangohud I mean mangoapp/the gamescope kinda-sorta vendored version of mangohud. I didn't intentionally test w/o that, but in my C version of this, things would hang when I ran this on machines w/o gamescope accidentally. 

That being said, you mentioned this is example code in which case, that doesn't matter so much.

likewhatevs (approved):

hodgesds:
> > As you know, the data presented in the MangoHud can be configured using the config file. I wonder if the message format remains the same regardless the configuration.
> 
> I believe it does, the structs were converted from [mangohud](https://github.com/flightlessmango/MangoHud/blob/master/src/app/mangoapp_proto.h) and [gamescope](https://github.com/ValveSoftware/gamescope/blob/1ab8009d5dc5faaff5d890ef896483ef14363536/src/mangoapp.cpp#L15). I believe mangohud handles the control messages, would have to dig through the code a bit to find how they are handled.

Here is the code for the [thread](https://github.com/flightlessmango/MangoHud/blob/master/src/app/main.cpp#L76) that handles control messages. So in theory the handling of control messages shouldn't be too different.

PR #1710, opened 2025-04-20
scx_flash: Avoid direct local dispatch of per-CPU tasks on SCX_ENQ_REENQ

arighi:
When a higher scheduling class steals a CPU, we trigger a re-enqueue of all the tasks waiting on that CPU, to give them a chance to be dispatched elsewhere, preventing potential stalls.

Per-CPU tasks are excluded by this logic and they are just re-enqueued directly to the same CPU's local DSQ, as they can't be migrated somewhere else.

However, this behavior can unintentionally favor per-CPU tasks if a higher class task periodically steals the CPU, since these tasks get continuously re-enqueued locally and may run ahead of other tasks without affinity constraints, potentially causing stalls.

To address this, treat per-CPU tasks like any other: enqueue them to the shared per-node DSQ during SCX_ENQ_REENQ to ensure a fairer scheduling and prevent starvation.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -222,37 +222,15 @@ void BPF_STRUCT_OPS(flash_enqueue, struct task_struct *p, u64 enq_flags)
 	struct task_ctx *tctx;
 
 	/*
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
I think this is a good direction, less special cases seems like a more balanced approach.

PR #1712, opened 2025-04-20
scx_p2dq: introduce arenas

etsal:
Introduce arenas usage to scx_p2dq by using the arena allocator to manage task contexts. This change keeps CPU masks out of arenas to ensure the code runs in all reasonably up-to-date kernels.


etsal:
I assume that this error for scx_chaos is expected and caused by the method override, right?

```
warning: scx_chaos@1.0.11: src/bpf/main.bpf.c:507:22: warning: initializer overrides prior initialization of this subobject [-Winitializer-overrides]
warning: scx_chaos@1.0.11:   507 |                .running                 = (void *)p2dq_running,
warning: scx_chaos@1.0.11:       |                                           ^~~~~~~~~~~~~~~~~~~~
warning: scx_chaos@1.0.11: /home/etsal/scx/target/debug/build/scx_chaos-17041fd05553c871/out/scx_utils-bpf_h/scx/compat.bpf.h:235:3: note: expanded from macro 'SCX_OPS_DEFINE'
```

hodgesds:
Slight verifier error in CI, otherwise LGTM:
```
 if (!llcx) { @ main.bpf.c:174
31: (55) if r0 != 0x0 goto pc+26 58: R0=map_value(map=llc_ctxs,ks=4,vs=264) R6=scalar(smin=umin=umin32=0x80000000,smax=umax=0xffffffff,smax32=-1,var_off=(0x80000000; 0x7fffffff)) R7=trusted_ptr_task_struct() R8=scalar(id=2,umin=1) R9=map_value(map=cpu_ctxs,ks=4,vs=112) R10=fp0 fp-8=mmmm????
; if (taskc->llc_id != cpuc->llc_id) { @ main.bpf.c:846
58: (61) r1 = *(u32 *)(r9 +4)         ; R1_w=scalar(smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) R9=map_value(map=cpu_ctxs,ks=4,vs=112)
59: (61) r2 = *(u32 *)(r8 +16)
R8 invalid mem access 'scalar'
```

etsal:
(NOTE: There will be some force-push spam as I try to fiddle the code to pass with the CI, since the code otherwise works fine on all of our machines)

JakeHillion:
> (NOTE: There will be some force-push spam as I try to fiddle the code to pass with the CI, since the code otherwise works fine on all of our machines)

Is it a kernel thing? You can get very nearly the same kernel (will be the same at some point in the future) with `$(nix build ./.github/include#kernels.sched_ext/for-next --no-link --print-out-paths)/bzImage`, then use vng to run it. Fine to churn the CI though :)

scheds/rust/scx_p2dq/build.rs
```diff
@@ -8,6 +8,8 @@ fn main() {
         .unwrap()
         .enable_intf("src/bpf/intf.h", "bpf_intf.rs")
         .enable_skel("src/bpf/main.bpf.c", "bpf")
+        .add_source("../../../lib/sdt_task.bpf.c")
+        .add_source("../../../lib/sdt_alloc.bpf.c")
```
JakeHillion, line 12: I think depending on things above Cargo.toml for that specific crate breaks `cargo publish`. This is already an issue for chaos, will take a look today and see if I can find a nice way to fix it.

JakeHillion: Okay, I solved it for chaos by exposing the files from `scx_p2dq`, seems like this is different. I think you need to add lib to this tar: https://github.com/sched-ext/scx/blob/main/rust/scx_utils/src/builder.rs, then it will be available in `target/scx_p2dq` somewhere. Can give you a hand if needed, I think there will be some trial and error.

etsal: Sounds good, I might make a second tar that we also link into the target since the header tar is named bpf_h and I'd rather not mix them up.

etsal:
For future reference: The problem was that CI was using Clang 18, which requires explicit typecasting for arena pointers. Since the main arena library lacks this kind of casting it, and we do not particularly care about supporting Clang 18, we are moving the CI to Clang 19.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -195,22 +201,20 @@ static struct node_ctx *lookup_node_ctx(u32 node_id)
 	return nodec;
 }
 
+struct lock_wrapper {
```
JakeHillion, line 204: nit: is this the right name?

JakeHillion (approved):

hodgesds (approved):
lets go!

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
etsal: Definitely not  :) This is an error due to muscle memory because we used to have wrappers for locks on scx_rusty I'll fix it up

PR #1713, opened 2025-04-21
scxtop: Add mangoapp TUI view

hodgesds:
Add a view to the TUI to render data from mangoapp. Show scheduler metrics and perf events for the application running with mangoapp enabled.
    
This is a continuation of #1708. An example showing filtered perf events and scheduler stats for CS2:

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -26,8 +26,9 @@ use crate::LICENSE;
 use crate::SCHED_NAME_PATH;
 use crate::{
     Action, CpuhpEnterAction, CpuhpExitAction, ExecAction, ExitAction, ForkAction, GpuMemAction,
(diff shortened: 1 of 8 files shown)


tools/scxtop/src/app.rs
```diff
@@ -2101,11 +2120,75 @@ impl<'a> App<'a> {
...
+
+        let left_constraints = vec![
+            Constraint::Percentage(2),
+            Constraint::Percentage(49),
+            Constraint::Percentage(49),
+        ];
+        let left_areas = Layout::vertical(left_constraints).split(left);
+        let theme = self.theme();
+
+        let mut pid_comm = if self.process_id > 0 {
```
JakeHillion, line 2149: nit

```suggestion
        let mut comm = if self.process_id > 0 {
```

PR #1715, opened 2025-04-22
scx_utils, scx_lavd: Use per-CPU cache size for CPU preference order.

multics69:
There are processors where CPUs with the same capacity have different cache sizes. To address such processors in determining CPU preference order, made the following three changes in topology:

- Add per-CPU cache size to Topology::Cpu (cache_size)
- Add the SMT level of a CPU to Topology::Cpu (smt_level)
- Assign core_id in a natural order

With these changes, scx_lavd was changed to take the per-CPU cache size into account -- always prefer a CPU with a bigger cache.

### DIFF
--- a/rust/scx_utils/src/misc.rs
+++ b/rust/scx_utils/src/misc.rs
@@ -117,6 +117,28 @@ pub fn read_file_usize_vec(path: &Path, separator: char) -> Result<Vec<usize>> {
         .collect::<Result<Vec<usize>>>()
 }
 
+pub fn read_file_byte(path: &Path) -> Result<usize> {
+    let val = std::fs::read_to_string(path)?;
+    let val = val.trim();
+
+    // E.g., 10K, 10M, 10G, 10
+    if val.ends_with("K") {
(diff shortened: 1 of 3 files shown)


hodgesds (approved):
LGTM

arighi (commented):
@multics69 this seems to introduce some breakage in multi-sockets / NUMA machines.

Reproducer:
```
$ vng --ssh -r --cpu 16,sockets=4,cores=2,threads=2  -m 4G --numa 1G,cpus=0-3 --numa 1G,cpus=4-7 --numa 1G,cpus=8-11 --numa 1G,cpus=12-15 -- ./build/release/scx_lavd
07:07:38 [INFO] Autopilot mode is enabled by default.

thread 'main' panicked at scheds/rust/scx_lavd/src/main.rs:284:40:
Failed to build host topology: Duplicate CPU ID 0
```

This happens with all Rust schedulers.

multics69:
Thanks @arighi ! I will fix it soon.

multics69:
This should fix the problem: https://github.com/sched-ext/scx/pull/1717

PR #1717, opened 2025-04-22
scx_utils: Fix incorrect handling of CPU IDs on multi-NUMA machines.

multics69:
The previous handling of CPU IDs for natural ordering ignored the NUMA boundaries, causing an error on multi-NUMA machines. So, pursue the numerical ordering of CPU IDs within a NUMA domain.

Fixes: 98f24eaa ("scx_utils: Create Cpus in a numerical order.")
Reported-by: Andrea Righi <arighi@nvidia.com>

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -785,13 +785,27 @@ fn create_numa_nodes(
             }
         }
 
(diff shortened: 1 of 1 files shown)


arighi (commented):
Looks good now, thanks for the fix!

PR #1720, opened 2025-04-22
p2dq/chaos: expose bpf file contents from p2dq crate

JakeHillion:
scx_chaos can't be published to crates.io currently because it depends on being built in the monorepo. Expose the BPF sources as bytes in the p2dq crate, which also solves the issue of needing to rebuild when p2dq changes.

Adding the custom include dir is really hacky. The BpfBuilder defines the cflags really early, and it's not really possible to neatly change them after, for example with a `.add_include_path()` method. We would need to break this API pretty significantly to add this cleanly. For now, hack the environment variable in the build of chaos to add an extra flag "after" (with) the includes.

In the future we may need to do something more generic like the TAR_H feature in scx_utils, but for 2 files I think this is simpler to follow.

Test plan:
- `cargo build`
- CI

### DIFF
--- a/scheds/rust/scx_chaos/Cargo.toml
+++ b/scheds/rust/scx_chaos/Cargo.toml
@@ -21,3 +21,4 @@ simplelog = "0.12"
 
 [build-dependencies]
 scx_utils = { path = "../../../rust/scx_utils", version = "1.0.13" }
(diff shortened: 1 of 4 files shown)


hodgesds (approved):
LGTM

JakeHillion:
Rustland runnable task stall https://github.com/sched-ext/scx/actions/runs/14597803757/job/40948107262 - putting it back in the queue.

PR #1723, opened 2025-04-22
scx_layered: add cpuset based growth algos

likewhatevs:
add cpuset support to layered like how we added numa support.

this is roughly the first half of this.

my thinking is we (a) add growth algos to ensure we have enough flexibility wrt/ core selection to write configs expressing what we want then we (b) do what allow_node_aligned does but using cpumasks of cpuset's instead of numa nodes.

cpuset spread:
```
16:31:19 [DEBUG] layer: hhvmworkers algo: CpuSetSpread core order: [0, 4, 8, 12, 1, 5, 9, 13, 2, 6, 10, 14, 3, 7, 11, 15]
```

cpuset spread reverse:
```
16:32:03 [DEBUG] layer: hhvmworkers algo: CpuSetSpreadReverse core order: [15, 11, 7, 3, 14, 10, 6, 2, 13, 9, 5, 1, 12, 8, 4, 0]
```

cpuset spread random:
```
16:32:21 [DEBUG] layer: hhvmworkers algo: CpuSetSpreadRandom core order: [0, 7, 8, 15, 3, 6, 10, 14, 1, 5, 9, 13, 2, 4, 11, 12]
16:32:29 [DEBUG] layer: hhvmworkers algo: CpuSetSpreadRandom core order: [2, 7, 9, 13, 3, 5, 8, 15, 0, 4, 10, 14, 1, 6, 11, 12]
16:32:43 [DEBUG] layer: hhvmworkers algo: CpuSetSpreadRandom core order: [2, 5, 8, 15, 3, 6, 11, 12, 1, 4, 9, 14, 0, 7, 10, 13]
```

from docker compose file specifying cpusets:
```
cpuset: "0-3,16-19"   #  0  1  2  3
cpuset: "4-7,20-23"   #  4  5  6  7
cpuset: "8-11,24-27"  #  8  9  10 11
cpuset: "12-15,28-31" #  12 13 14 15
```


htejun (commented):
Hmmm... I'm not sure how this is supposed to work. The node spread one assumes that each node has the same number of cpus and the layer ordering matches the node assignment. However, the same approach probably wouldn't work - e.g. the number of cpus per cpuset might not be a nice round number, it's difficult to define / predict ordering among layers and so on. Wouldn't it make more sense for each layer to specify the cgroup that the cpuset should be read from?

likewhatevs:
> Hmmm... I'm not sure how this is supposed to work. The node spread one assumes that each node has the same number of cpus and the layer ordering matches the node assignment. However, the same approach probably wouldn't work - e.g. the number of cpus per cpuset might not be a nice round number, it's difficult to define / predict ordering among layers and so on. 

How allow node aligned works with NodeSpread* doesn't depend upon layer ordering IIUC, it depends upon layer sizing (i.e. it would fall over if a layer had 1 CPU). 

My thinking is that, in addition to what this PR has so far, I can set the minimum number of CPUs per layer to the number of distinct cpusets in the system (ensuring all layers have at least 1 CPU). I think this paired with (maybe) resizing layers in num_cpusets sized steps (and maybe scaling layer size based on DSQ max delay in addition to util), we could get layered supporting homogenous containerized workloads w/ minimal deviation from how we do NUMA support ATM.


> Wouldn't it make more sense for each layer to specify the cgroup that the cpuset should be read from?

Dynamic naming of cpusets would be a problem wrt/ that I think. We could do something like generating layers from those passed in and making those generated layers be cpuset specific (based upon present cpusets), but I think that deviates enough from how layered is battle-tested that it may not work well (i.e. this would effectively make a 3 layer config a 12 layer config if there were 4 homogenous containers, which might be more problematic of a change than making a 1 NUMA node config a 4 kinda-sorta NUMA node config, which is the direction of this PR).

Does this check out?

htejun:
Right, I was confused. The layer rotation isn't being used. We're using the same layer across nodes and you're saying that the plan is to do the same across cpusets and the only thing we need for that is for the layer to grow in a way that populates each cpuset equally. It is janky but yeah I don't see a reason why it wouldn't work.

htejun (approved):

likewhatevs:
cleaned up, still works:

Screenshot from 2025-04-24 09-17-08
Screenshot from 2025-04-24 09-16-46

Also added handling of overlapping cpusets to this. ATM I think all we want is "only consider subsets when supersets are present" and "toss one if overlap is present".

etsal (approved):
Looks reasonable, non-disjoint cpusets are the only edge case I see but for these kinds of configurations we will just not use this growth algorithm.

likewhatevs:
> Looks reasonable, non-disjoint cpusets are the only edge case I see but for these kinds of configurations we will just not use this growth algorithm.

At that point the answer might be "don't use cpuset to isolate your containers if you don't want isolation" lol, idk.

PR #1724, opened 2025-04-22
[WIP] scx_rusty: add AMD IBS performance monitoring

etsal:
scx_rusty is currently using execution time as the main metric with which is load balances tasks between domains. However, there are other forms of resource contention that we can avoid through load balancing, e.g., L3 footprint and memory bandwidth. We can track these metrics accurately, per-process using hardware monitoring extensions like AMD IBS. Add initial support for reading data from these extensions into the scheduler.

STATUS: 
- The number of L1 hits and number of DRAM hits is suspiciously high and low, respectively.
- There is currently a large number of valid samples that hold seemingly invalid physical and virtual addresses. These are exclusively L1 hits, but it remains to be seen whether they are actually valid or somehow erroneous and should be filtered out.


multics69:
@etsal -- The idea of leveraging the perf counter in making scheduling decisions is super cool! I wonder how much overhead it imposes.

etsal:
> @etsal -- The idea of leveraging the perf counter in making scheduling decisions is super cool! I wonder how much overhead it imposes.

One possible issue wrt overhead is that right now we trigger for every single perf event, which is not really necessary. This is because we need fine granularity for samples but do not really care about receiving them immediately. If that turns out ot be a problem we should should either find a way to batch the delivery of these events, or we should write one :)

multics69:
> > @etsal -- The idea of leveraging the perf counter in making scheduling decisions is super cool! I wonder how much overhead it imposes.
> 
> One possible issue wrt overhead is that right now we trigger for every single perf event, which is not really necessary. This is because we need fine granularity for samples but do not really care about receiving them immediately. If that turns out ot be a problem we should should either find a way to batch the delivery of these events, or we should write one :)

I think you can change the sampling rate of AMD IBS  (like Intel PEBS). If so,  you can introduce a logic to autotune the sampling rate to control the overhead (e.g., 1000 samples / second).

etsal:
> > > @etsal -- The idea of leveraging the perf counter in making scheduling decisions is super cool! I wonder how much overhead it imposes.
> > 
> > 
> > One possible issue wrt overhead is that right now we trigger for every single perf event, which is not really necessary. This is because we need fine granularity for samples but do not really care about receiving them immediately. If that turns out ot be a problem we should should either find a way to batch the delivery of these events, or we should write one :)
> 
> I think you can change the sampling rate of AMD IBS (like Intel PEBS). If so, you can introduce a logic to autotune the sampling rate to control the overhead (e.g., 1000 samples / second).

It is, but ideally we would still get a very high amount of samples in batches instead of one at a time to avoid running the BPF callback as often, which I'm not sure is currently possible.

multics69:
> > 
> > I think you can change the sampling rate of AMD IBS (like Intel PEBS). If so, you can introduce a logic to autotune the sampling rate to control the overhead (e.g., 1000 samples / second).
> 
> It is, but ideally we would still get a very high amount of samples in batches instead of one at a time to avoid running the BPF callback as often, which I'm not sure is currently possible.

Sure, the batching will be essential to put aside processing IBS samples from the critical path.

htejun:

PR #1725, opened 2025-04-23
scx_utils, scx_lavd: Group CPUs in the same performance domain.

multics69:
When deciding the CPU preference order, group the CPUs in the same performance domain (i.e., CPU frequency domain) to minimize the number of active performance domains.

### DIFF
--- a/rust/scx_utils/src/energy_model.rs
+++ b/rust/scx_utils/src/energy_model.rs
@@ -65,6 +65,15 @@ impl EnergyModel {
 
         Ok(EnergyModel { perf_doms })
     }
+
+    pub fn get_pd(&self, cpu_id: usize) -> Option<&PerfDomain> {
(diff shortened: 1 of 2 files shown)


rust/scx_utils/src/energy_model.rs
```diff
@@ -65,6 +65,15 @@ impl EnergyModel {
...
         Ok(EnergyModel { perf_doms })
     }
+
+    pub fn get_pd(&self, cpu_id: usize) -> Result<&PerfDomain> {
+        for (_, pd) in self.perf_doms.iter() {
+            if pd.span.test_cpu(cpu_id) {
+                return Ok(&pd);
+            }
+        }
+        bail!("Fail to find a performance domain for cpu {:#}", cpu_id);
```
hodgesds, line 75: How about returning an option instead? `bail` seems rather aggressive.

PR #1728, opened 2025-04-23
scx_lavd: get cpu context of scx_bpf_task_cpu in running

JakeHillion:
Recent changes in the kernel added new restrictions to rq locks, aiming to prevent deadlocks. lavd triggers this new case in running when calling `scx_bpf_cpuperf_set`. This is because running has already locked an rq before entry and cpuperf_set attempted to operate on and lock a different rq, potentially causing a deadlock.

Get the `cpuc` in lavd_running identified by `scx_bpf_task_cpu` instead of the current CPU. I look forward to being told why this works, but the failure was completely reliable before and goes away with this change.

Relevant kernel commits:
- https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git/commit/?h=for-next&id=18853ba782bef65fc81ef2b3370382e5b479c5eb
- https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git/commit/?h=for-next&id=a11d6784d7316a6c77ca9f14fb1a698ebbb3c1fb

Test plan:
- This fix & CI with updated kernel: https://github.com/sched-ext/scx/actions/runs/14617926571/job/41011059539
- This fix & CI with previous kernel on this PR.
- No fix & failed CI with updated kernel: https://github.com/sched-ext/scx/actions/runs/14614177179/job/41003568052?pr=1727

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -1075,7 +1075,7 @@ void BPF_STRUCT_OPS(lavd_running, struct task_struct *p)
 	/*
 	 * Update task statistics
 	 */
-	cpuc = get_cpu_ctx();
+	cpuc = get_cpu_ctx_id(scx_bpf_task_cpu(p));
 	taskc = get_task_ctx(p);
 	if (!cpuc || !taskc)
 		return;


arighi:
The fix makes sense to me, but I don't see in the kernel where `ops.running()` is called with `rq != this_rq()`, I'll investigate a bit more... good thing the kernel change exposed this behavior, I was ignoring this particular case.

multics69:
> The fix makes sense to me, but I don't see in the kernel where `ops.running()` is called with `rq != this_rq()`, I'll investigate a bit more... good thing the kernel change exposed this behavior, I was ignoring this particular case.

I assumed `rq == this_rq()`.

arighi:
> > The fix makes sense to me, but I don't see in the kernel where `ops.running()` is called with `rq != this_rq()`, I'll investigate a bit more... good thing the kernel change exposed this behavior, I was ignoring this particular case.
> 
> I assumed `rq == this_rq()`. Yeah, we need to take a look such case is quire okay.

Same, but looking at the error condition that we hit, we have a case where `rq != this_rq()`, I think I understand now.

Looking at the trace that we got before @JakeHillion's fix:
```
[    8.816843] sched_ext: lavd: Invalid target CPU 0
[    8.817819]    bpf_prog_50c71f4cf236e424_lavd_running+0x42f/0x7a9
[    8.824081]    bpf__sched_ext_ops_running+0x47/0xa7
[    8.825081]    set_next_task_scx+0x170/0x240
[    8.833011]    __set_cpus_allowed_ptr_locked+0x14a/0x1e0
[    8.834052]    __set_cpus_allowed_ptr+0x63/0xb0
[    8.835048]    __sched_setaffinity+0x74/0x100
[    8.842999]    sched_setaffinity+0x17e/0x2a0
[    8.843898]    __x64_sys_sched_setaffinity+0x50/0x90
[    8.844927]    do_syscall_64+0xbb/0x1e0
[    8.845684]    entry_SYSCALL_64_after_hwframe+0x77/0x7f
```

When a task is changing affinity it triggers `scx_next_task_scx()` that calls `ops.running()` on a different CPU different than the target one. In this case `this_rq()` can be different than `rq`.

So, when we consider the target CPU in `ops.running()` we should always rely on `scx_bpf_task_cpu(p)` rather than `bpf_get_smp_processor_id()`.

I think we an easily stress test this case running a `stress-ng --race-sched 0` and add a check like `scx_bpf_task_cpu(p) != bpf_get_smp_processor_id()`.

In any case, the fix makes sense, now I think we understand that relying on the current CPU from `ops.running()` is not always correct.

arighi:
And it would be nice do document this information somewhere, ideally in `struct sched_ext_ops` where we document 
 `ops.running()`. Moreover, the same should apply to `ops.stopping()`, because we can have a similar pattern with `dequeue_task_scx()`.

If this makes sense to you all I'll send a patch to clarify this in the doc.

PR #1729, opened 2025-04-23
scx_lavd: Improve handling of synchronous wake-up.

multics69:
Previously, the synchronous wake-up is used to mitigate cross-domain migrations. However, waker's CPU is good for cache locality too.

If the waker CPU is idle, stay on the waker CPU when there is no other idle CPU in the sticky domain. It is good for cache locality because the waker task hands over the CPU to the wakee task for the further processing after finishing its job.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
@@ -36,7 +36,7 @@ struct pick_ctx {
 	/*
 	 * Additional input arguments for find_sticky_cpu_and_cpdom().
 	 */
-	s32 waker_cpu;
+	s32 sync_waker_cpu;
 	bool is_task_big;
 	/*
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
Should help a bit with stickiness, it could be useful to add a flag to allow LLC migrations on wakeups for some workloads that have waker/wakee cache locality.

multics69:
> Should help a bit with stickiness, it could be useful to add a flag to allow LLC migrations on wakeups for some workloads that have waker/wakee cache locality.

Thanks for the suggestion. I will address this in a separate PR.

multics69:
@hodgesds -- I think we can use some heuristics to guestimate whether a wakee migration to its waker is beneficial. For example, if a wakee is a long-running/non-interactive/non-latency-critical task, we can safely migrate to the waker's domain because the migration cost is less visible but two tasks likely share the cache (long running time). What do you think?

PR #1730, opened 2025-04-23
ci: update ubuntu installations to clang 19

JakeHillion:
Update llvm to 19 in the Ubuntu runners. This is needed for proper arena support and was leading to verifier errors.

Test plan:
- CI

### DIFF
--- a/.github/actions/install-deps-action/action.yml
+++ b/.github/actions/install-deps-action/action.yml
@@ -23,8 +23,10 @@ runs:
         linux-headers-generic linux-tools-common linux-tools-generic make \
         ninja-build pahole pkg-config python3-dev python3-pip python3-requests \
         qemu-kvm rsync stress-ng udev zstd libseccomp-dev libcap-ng-dev \
-        llvm clang python3-full curl meson bpftrace dwarves rustup \
+        llvm-19 clang-19 python3-full curl meson bpftrace dwarves rustup \
         protobuf-compiler
+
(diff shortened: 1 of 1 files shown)


arighi:
`clang-19` should be available in ubuntu 24.04 (noble), is there a reason to use the one from llvmg.org?

```
$ rmadison clang-19 | grep noble
 clang-19 | 1:19.1.1-1ubuntu1~24.04.2 | noble-updates/universe | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
```

JakeHillion:
> `clang-19` should be available in ubuntu 24.04 (noble), is there a reason to use the one from llvmg.org?
> 
> ```
> $ rmadison clang-19 | grep noble
>  clang-19 | 1:19.1.1-1ubuntu1~24.04.2 | noble-updates/universe | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
> ```

Was not aware of that, thanks! Also cleaned up that symlink nonsense by appending the right directory to $GITHUB_PATH.

arighi (approved):
Looks great, thanks!

etsal:
Should we look into the rustland stall? It has become more frequent recently.

JakeHillion:
> Should we look into the rustland stall? It has become more frequent recently so it may not be a fluke but sth worth checking more closely.

Yeah, created #1731 to track this. Let's discuss there.

PR #1732, opened 2025-04-23
scx_lavd: Get the right CPU context from ops.stopping()

arighi:
Similarly to commit e81fa0d3 ("lavd: get cpu context of scx_bpf_task_cpu in running") apply the same change to ops.stopping() as well.

Both ops.running() and ops.stopping() can be called from a CPU other than the one the task is assigned to (i.e., a task changing a property of another task), therefore always rely on scx_bpf_task_cpu() to get the right CPU context.


hodgesds (approved):

multics69:
Thanks @arighi for the fix!

PR #1733, opened 2025-04-24
scx_lavd: Refactor the use of get_cpu_ctx().

multics69:
Add a new util function, get_cpu_ctx_task(@p), to get a cpu_ctx of task @p. This function is intended to use at ops.running(), ops.stopping(), and ops.quiescent() where current task's rq could be different from task @p's rq.


JakeHillion (approved):
Nice! Thanks for finishing this off now that we understand it better.

PR #1736, opened 2025-04-24
Revamp scx_rustland

### LINKED ISSUE #1731: scx_rustland stalls in CI quite often
scx_rustland has been struggling in the CI recently and regularly stalls when the workflow is otherwise fine. This change has been noticeable since https://github.com/sched-ext/scx/pull/1679 when the stress-ng arguments got a little more aggressive.

Let's discuss how to fix this. It might be that we need more lightweight stress arguments for the userspace scheduling, but this hypothesis seems to be disproved by rlfifo being fine. Alternatively we might be able to reduce stalls by increasing the stall timer for rustland from 5000ms to 10000ms using the new environment variable SCX_TIMEOUT_MS.

An example is https://github.com/sched-ext/scx/pull/1730 where we had to re-add to the merge queue 3 times, which means rustland stalled 3 times in a row in 3 separate job instances: https://github.com/sched-ext/scx/actions/runs/14624011790/job/41031030390

cc: @arighi @etsal

arighi:
A set of fixes/improvements for scx_rustland_core:
 - avoid refilling task's time slice in ops.running() that may cause starvation
 - expose task's CPU time since last sleep to user-space
 - allow to toggle bulilt-in idle CPU policy

And a redesign of scx_rustland:
 - use scheduling policy similar to bpfland and flash
 - enable the built-in idle CPU selection policy to get optimal performance when the system is not fully saturated

This should fix #1731.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf.rs
+++ b/rust/scx_rustland_core/assets/bpf.rs
@@ -195,6 +195,7 @@ impl<'cb> BpfScheduler<'cb> {
         exit_dump_len: u32,
         partial: bool,
         debug: bool,
+        builtin_idle: bool,
     ) -> Result<Self> {
         let shutdown = Arc::new(AtomicBool::new(false));
         set_ctrlc_handler(shutdown.clone()).context("Error setting Ctrl-C handler")?;
@@ -255,6 +256,7 @@ impl<'cb> BpfScheduler<'cb> {
         skel.struct_ops.rustland_mut().exit_dump_len = exit_dump_len;
 
         skel.maps.bss_data.usersched_pid = std::process::id();
+        skel.maps.rodata_data.builtin_idle = builtin_idle;
         skel.maps.rodata_data.debug = debug;
 
         // Attach BPF scheduler.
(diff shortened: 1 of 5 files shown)


arighi:
Well, it doesn't fix CI... this needs more work.

hodgesds:
> And a redesign of scx_rustland:
> 
> * use scheduling policy similar to bpfland and flash
> * enable the built-in idle CPU selection policy to get

We can make the CI less aggressive for now if wish.

arighi:
> > And a redesign of scx_rustland:
> > 
> > * use scheduling policy similar to bpfland and flash
> > * enable the built-in idle CPU selection policy to get
> 
> We can make the CI less aggressive for now if wish.

It seems a bit more stable now, the stress tests are quite stressful for user-space scheduling, but let's keep it as they are for now. If it becomes too painful and CI keeps failing we can relax the tests for rustland. Thanks!

arighi:
Alright, I did more tests and this one looks more solid now. We can probably do better, especially in terms of CPU utilization, but at least with these changes we should see less CI failures, so I'm going to merge this one.

PR #1737, opened 2025-04-24
scx_utils, scxtop: Bindgen perf event

frelon:
This PR adds a perf module to scx_utils that uses the bindgen crate to generate the bindings for perf_event_open syscall.

This replaces the use of perf-event-open-sys2 crate in scxtop in order to be more portable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1859,16 +1859,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 
 [[package]]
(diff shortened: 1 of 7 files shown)


hodgesds (approved):

etsal (commented):
@frelon Can you elaborate on what you mean by "more portable"? AFAICT this just exposes the system call directly instead of going through the crate. If the crate is a liability this looks good, since it extracts the syscall definition out of it and throws out the rest.

frelon:
@etsal Sure, the crate generates the bindings for arm64, x86_64 and riscv64 and to build for other platforms would need updates in the crate.

This approach generates the bindings during build for the target architecture, which hopefully means it works for ppc, s390 etc without changes.

PR #1740, opened 2025-04-24
scx_layered: allow protected CPUs to pull tasks from the low fallback queues

etsal:
The protected flag currently prevents any tasks from executing in CPUs owned by layers that set it. This ensures QoS for the tasks in the protected layer,  but can lead to stalls if there are not enough unprotected CPUs in the system to drain the low fallback queues in time. Relax the guarantees the flag provides by letting protected CPUs run low fallback tasks. Also clean up some of the logic related to the protected flag.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1828,7 +1828,7 @@ void BPF_STRUCT_OPS(layered_dispatch, s32 cpu, struct task_struct *prev)
 	 * Low fallback DSQ execution is forced upto lo_fb_share_ppk fraction
 	 * after the DSQ had tasks queued for longer than lo_fb_wait_ns.
 	 */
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -404,8 +399,7 @@ static void refresh_cpumasks(u32 layer_id)
 
 		if ((u8_ptr = MEMBER_VPTR(layers, [layer_id].cpus[cpu / 8]))) {
 			if (*u8_ptr & (1 << (cpu % 8))) {
-				if (refresh_layer_cpuc(cpuc, layer))
-					protected_changed = true;
+				protected_changed = refresh_layer_cpuc(cpuc, layer);
```
htejun, line 402: This should be cumulative. You can do `|=` but `=` means that we're only accounting for the latest result.

PR #1746, opened 2025-04-25
Bump rust to 2024

likewhatevs:
Bump rust to 2024 and remove pin to rust versions from before then + fixes across upgrade.

This this shaves an extra 5 sec off cold build time and 1 sec off warm for me when used w/ https://github.com/sched-ext/scx/pull/1745 .

Also the word gen is now a keyword, so I renamed problematic instances of that to genn.


JakeHillion:
Will bump the Nix deps for you. This is pretty early though, I think only the latest release of Cargo at this time supports edition2024 and it might cause issues for distros trying to package.

likewhatevs:
> Will bump the Nix deps for you. This is pretty early though, I think only the latest release of Cargo at this time supports edition2024 and it might cause issues for distros trying to package.

I thought so too when I looked at it but was kind of surprised to see that this seems to be everywhere/non-problematic already tbh. All for not doing this if I missed something.

lib/sdt_alloc.bpf.c
```diff
@@ -444,7 +444,7 @@ int scx_alloc_free_idx(struct scx_allocator *alloc, __u64 idx)
 	if (likely(!data)) {
 
 		data[pos] = (struct sdt_data) {
-			.tid.gen = data->tid.gen + 1,
+			.tid.genn = data->tid.genn + 1,
```
htejun, line 447: What's with `s/gen/genn/`? Did `gen` become a reserved token?

htejun (approved):

lib/sdt_alloc.bpf.c
likewhatevs: Yep. Everything w/ codegen based ffi and garbage collection is gonna have fun w/ gen becoming a keyword.

arighi (commented):
Hm... this breaks everything in Ubuntu and most likely in many other distros that don't ship the ultra latest version of cargo. Do we really need to rush this change? Aside from the minor build time improvement, is there any real benefit?

PR #1747, opened 2025-04-25
layered cpuset support

likewhatevs:
I need to debug this because it is still has all tasks going to lo fallback when I run: 

https://github.com/likewhatevs/perfstuff/blob/main/noisy-workload.compose.yml

That being said, this does run/pass the verifier and has all the idk components needed to make this work so lmk thoughts on the approach etc.


etsal (commented):
So AFAICT The changes are:
1) Machinery for forwarding cpuset information to the BPF side
2) The corresponding BPF code
3) Logic that replicates allow_node_aligned but for cpusets.

If this fixes cpuset related problems I think it is reasonable, but I'm not sure about the naming - contianer enable is a bit confusing because containers aren't really a thing at this level of abstraction. Maybe replace "container"  with "cpuset-based workloads"? This way it's clear what the code does concretely.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1333,8 +1350,10 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
 	 * with open layers on non-saturated machines to avoid possible stalls.
 	 */
 	if ((!taskc->all_cpus_allowed &&
-	     !(layer->allow_node_aligned && taskc->cpus_node_aligned)) ||
-	    !layer->nr_cpus) {
+		!(layer->allow_node_aligned && taskc->cpus_node_aligned)) ||
+		!(enable_container && taskc->cpus_cpuset_aligned) ||
```
etsal, line 1354: Wrt the bug, maybe wrap lines 1354-1355 in parentheses? && has a higher precedence than || so rn all tasks without cpus_cpuset_aligned are getting put into the fallback queue regardless of the value of tasks->all_cpus_allowed.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -30,6 +30,7 @@ enum consts {
 	MAX_TASKS		= 131072,
 	MAX_PATH		= 4096,
 	MAX_NUMA_NODES		= 64,
+	MAX_CONTAINERS		= 64,
```
htejun, line 33: Can we use 'cpuset` / `CPUSET` consistently instead of using both `container` and `cpuset` and document the implemented behavior in a comment?

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: think that was it, thanks 1000x for that.

likewhatevs:
got this working right (finally, lol), I think. 
image

most of the changes are gymnastics around getting bitmasks from rust to bpf cpumasks in a way that keeps the verifier happy w/o messing up the bitmask.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3377,6 +3414,57 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
...
+
+	if (enable_cpuset) {
+		bpf_for(i, 0, nr_cpusets) {
+			cpumask = bpf_cpumask_create();
+			
+			if (!cpumask)
+				return -ENOMEM;
+
+			bpf_for(j, 0, MAX_CPUS/64) {
+				bpf_for(cpu, 0, 63) {
```
etsal, line 3427: This effectively skips the last CPU.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3377,6 +3414,57 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
...
+					if (cpu < 0 || cpu >= 64 || j < 0 || j >= (MAX_CPUS/64) || i < 0 || i >= MAX_CPUSETS) {
+						bpf_cpumask_release(cpumask);
+						return -1;
+					}
+					if (cpuset_fakemasks[i][j] & (1LLU << cpu)) {
+					 	bpf_cpumask_set_cpu((MAX_CPUS/64 - j - 1) * 64 + cpu, cpumask);
```
etsal, line 3433: So AFAICT the cpumask should fit all node cpumasks? This looks like it works because we clobber vmlinux.h to hold 128 64-bit numbers, which is fine to bypass verifier behavior but I'm not sure is great to depend on.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3377,6 +3414,57 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
...
+					}
+
+				}
+			}
+
+			// pay init cost once for faster lookups later.
+			bpf_for(cpu, 0, nr_possible_cpus) {
+				cpumask_box = bpf_map_lookup_percpu_elem(&cpuset_cpumask, &i, cpu);
```
etsal, line 3441: Minor nit: Could you maybe rename to cpumask_wrapper to go with the lock_wrapper we use for our spinlocks across the schedulers?

likewhatevs: misc btw -- is that a thing for the same reason as this (i.e. verifier pointer tracking)?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3377,6 +3414,57 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
...
+	if (enable_cpuset) {
+		bpf_for(i, 0, nr_cpusets) {
+			cpumask = bpf_cpumask_create();
+			
+			if (!cpumask)
+				return -ENOMEM;
+
+			bpf_for(j, 0, MAX_CPUS/64) {
+				bpf_for(cpu, 0, 63) {
+					if (cpu < 0 || cpu >= 64 || j < 0 || j >= (MAX_CPUS/64) || i < 0 || i >= MAX_CPUSETS) {
```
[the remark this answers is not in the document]
likewhatevs: yeah it only needs i checks atm

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: Yeah I guess wrapping the type around makes it possible to reason about what type the pointer is? Not sure though.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: ideally we'll rip out all unprintable trusted ptr cpumasks w/ printable arena cpumasks eventually I think, maybe. I really wish these were printable lol...

likewhatevs:
still works w/ the indexes fixed/renames etc.:

cpuset workload only:
Screenshot From 2025-05-12 10-27-57

cpuset workload + random affinity workload:
Screenshot From 2025-05-12 10-27-44

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1381,9 +1399,11 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
 	 * without making the whole scheduler node aware and should only be used
 	 * with open layers on non-saturated machines to avoid possible stalls.
 	 */
```
htejun, line 1401: Please update the comment to explain `cpus_cpuset_aligned`.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2658,7 +2678,7 @@ static void refresh_cpus_flags(struct task_ctx *taskc,
 
 		if (!(nodec = lookup_node_ctx(node_id)) ||
 		    !(node_cpumask = cast_mask(nodec->cpumask)))
-			return;
+			break;
```
htejun, line 2681: This is `scx_bpf_error()` condition. There's no point in continuing.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2667,6 +2687,21 @@ static void refresh_cpus_flags(struct task_ctx *taskc,
 			break;
 		}
 	}
+	if (enable_cpuset) {
```
htejun, line 2690: Maybe a blank line above?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2667,6 +2687,21 @@ static void refresh_cpus_flags(struct task_ctx *taskc,
...
+		bpf_for(cpuset_id, 0, nr_cpusets) {
+			struct cpumask_wrapper* wrapper;
```
htejun, line 2692: Blank line.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2667,6 +2687,21 @@ static void refresh_cpus_flags(struct task_ctx *taskc,
...
+			wrapper = bpf_map_lookup_elem(&cpuset_cpumask, &cpuset_id);
+			if (!wrapper || !wrapper->mask) {
+				scx_bpf_error("error marking tasks as cpuset aligned");
+				return;
+			}
+			if (bpf_cpumask_equal(cast_mask(wrapper->mask), cpumask)) {
+				taskc->cpus_cpuset_aligned = true;
+				return;
```
htejun, line 2700: `break;` here so that it's consistent with the node aligned block and the function can be extended in the future? Note that this would require moving the `false` setting. BTW, why not use the same partial overlapping test used by node alignment test instead of equality test? Is that not sufficient for forward progress guarantee? If not, it'd probably be worthwhile to explain why.

htejun (commented):
I still have trouble understanding how this is supposed to work. Node aligned is easier because DSQs are LLC aligned and LLCs are node aligned. We don't have the guarantee that cpusets are LLC aligned. Is that okay? If so, why? Can you please document the theory of operation?

likewhatevs:
> We don't have the guarantee that cpusets are LLC aligned. Is that okay? If so, why? Can you please document the theory of operation?

Will add documentation w/ the other updates. The TL;DR wrt/ theory of operation is that, if cpusets are being used, the onus is on whoever is setting those to ensure they are LLC aligned or perf will be affected and the scheduler can't (or perhaps shouldn't) really fix that.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3377,6 +3414,57 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
...
+						return -1;
+					}
+					if (cpuset_fakemasks[i][j] & (1LLU << cpu)) {
+					 	bpf_cpumask_set_cpu((MAX_CPUS/64 - j - 1) * 64 + cpu, cpumask);
+					}
+
+				}
+			}
+
+			// pay init cost once for faster lookups later.
```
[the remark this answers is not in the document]
likewhatevs: > Why do we need per-cpu copies?

This is called via set_cpumask. IIUC, this can be called from any CPU, for any task, and would be called for all tasks on workloads utilizing cpuset. This is per cpu to reduce lookup cost on high core count machines (where cpuset would more likely be used).

> Can't this be a part of cpu_ctx? 

I don't think so. cpu_ctx is a map with one entry. This map has an entry per cpuset on the system (i.e. the per-cpu duplication is solely to make runtime lookups faster).

> Note that percpu maps have a limited number of hot-caches per task and has a performance cliff beyond.

I can run an AB vs EEVDF w/ this (which isn't the highest bar), sort of like how the regression detector does that, but I'm not sure how to tell if I've crossed that threshold tbh?

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: made this an explicit error instead of just a return, thanks.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: Ahh, correct me if I am wrong, that partial test is wrong/the node code should do what the cpuset code does?

For example:
```
1000 task_mask
1100 system_mask

system_mask intersect task_mask == true
system_mask subset task_mask == false
```

IIUC, allow node aligned is only supposed to avoid lo fb when task cpumask matches node cpumasks to ensure lo fb is used for tasks with single cpu affinities (i.e. which are stall prone)?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3377,6 +3414,57 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(layered_init)
 	if (tmp_unprotected_cpumask)
 		bpf_cpumask_release(tmp_unprotected_cpumask);
 
+
+
+	if (enable_cpuset) {
+		bpf_for(i, 0, nr_cpusets) {
+			cpumask = bpf_cpumask_create();
+
```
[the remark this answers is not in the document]
likewhatevs: Ahh LMK if I got that right. I follow do that wrt/ simple assignments/checks, but am not sure if it should be done w/ more complex ones (i.e. I see more places I could do that, not sure if I should though).

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: I think I figured out the whitespace convention (before if, before for, sometimes after for), but LMK if there's still places it's wrong.

likewhatevs:
I think I covered everything, LMK if not! I noted the guidance I didn't quite follow in responses and ran the test cases to confirm things still work.

scheds/rust/scx_layered/src/bpf/main.bpf.c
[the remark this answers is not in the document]
htejun: > > Why do we need per-cpu copies?
> 
> This is called via set_cpumask. IIUC, this can be called from any CPU, for any task, and would be called for all tasks on workloads utilizing cpuset. This is per cpu to reduce lookup cost on high core count machines (where cpuset would more likely be used).

I'm not sure how that matters. This is mostly read-only data. There's no synchronization involved in looking up array map element. It doesn't matter how many cpus are looking these up. The cachelines would just be shared across the caches.

> > Note that percpu maps have a limited number of hot-caches per task and has a performance cliff beyond.
> 
> I can run an AB vs EEVDF w/ this (which isn't the highest bar), sort of like how the regression detector does that, but I'm not sure how to tell if I've crossed that threshold tbh?

All percpu map users share the same cache whether that's sched_ext or something else. The current cache entry limit is 16 and if the number of active percpu maps go over that, the lookups will fall back to slow path which can be significantly slower. This is a system-wide limit and the higher number of per-cpu maps you use, the more likely this will become a problem, so it's best to reduce the number as much as possible. Besides, I just don't understand why per-cpu map is used in the first place.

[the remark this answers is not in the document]
likewhatevs: Thank you for the explanation!

> Besides, I just don't understand why per-cpu map is used in the first place.

I was just trying to do everything per cpu for multi-node machines (which I now see was misguided and counterproductive because of that cache entry limit).

htejun:
> > We don't have the guarantee that cpusets are LLC aligned. Is that okay? If so, why? Can you please document the theory of operation?
> 
> Will add documentation w/ the other updates. The TL;DR wrt/ theory of operation is that, if cpusets are being used, the onus is on whoever is setting those to ensure they are LLC aligned or perf will be affected and the scheduler can't (or perhaps shouldn't) really fix that.

This isn't going to be true in a lot of cases. I'm not sure how this PR would behave when cpusets aren't necessarily LLC aligned. Maybe it's okay but we really should think it through.

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: Hmmm... yeah, on machines with single LLC nodes, this is fine, but if there are multiple LLCs per node, the current `allow_node_aligned` condition can lead to unexpected behaviors as tasks can end up in a DSQ that a layer doesn't have CPUs on. But going back to this PR, if cpusets aren't aligned with LLCs, wouldn't we have a similar problem?

likewhatevs:
> > > We don't have the guarantee that cpusets are LLC aligned. Is that okay? If so, why? Can you please document the theory of operation?
> > 
> > 
> > Will add documentation w/ the other updates. The TL;DR wrt/ theory of operation is that, if cpusets are being used, the onus is on whoever is setting those to ensure they are LLC aligned or perf will be affected and the scheduler can't (or perhaps shouldn't) really fix that.
> 
> This isn't going to be true in a lot of cases. I'm not sure how this PR would behave when cpusets aren't necessarily LLC aligned. Maybe it's okay but we really should think it through.

I don't see anything immediately off when running the test program I've been using on a system where all the cpusets are intentionally misaligned, and I've had that test running for a few minutes so I don't think misalignment would cause stalls (although, in testing we're definitely gonna find some, probably around layer growth/shrinkage, I think):

image

Also I replaced percpu array use with regular array use.

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: > Hmmm... yeah, on machines with single LLC nodes, this is fine, but if there are multiple LLCs per node, the current allow_node_aligned condition can lead to unexpected behaviors as tasks can end up in a DSQ that a layer doesn't have CPUs on.

Will fix that in a separate PR, thx. 

> But going back to this PR, if cpusets aren't aligned with LLCs, wouldn't we have a similar problem?

Yeah, I was just able to produce that kind of behavior w/ my test setup, gonna have to add something to deal with that.

likewhatevs:
My prior test w/ misaligned LLCs was bad (either I forgot to docker down/up to update cgroups or restart the scheduler).

I think this confirms things work as expected.

Htop shows funny cpusets are processing load:
Screenshot From 2025-05-19 19-51-15

Layered stats show cpuset affinitized work going to lo fallbacks:
Screenshot From 2025-05-19 19-51-32

The cpusets specified match cpus used in htop above:
Screenshot From 2025-05-19 19-52-06

I think this means that perf will suffer if cpuset's aren't LLC aligned, but this is OK/best that can be done in that case.

PR #1748, opened 2025-04-25
ci: refactor integration-tests part of workflows to reduce duplication

JakeHillion:
Refactor the `integration-test` and `layered-matrix` steps into a reusable workflow named `integration-tests`. This ensures any changes are correctly propagated between `caching-build` (the PR gate) and the various timer based workflows. For example, we weren't running layered-matrix on bpf-next or stable and we probably should. Also makes it easier to track more kernels in the future.

Left the Slack notification in the individual workflows for now. We could put it in the shared one with an argument to gate it, but it's unlikely to change.

Test plan:
- CI
- Pushed a version where `stable` and `bpf-next-test` triggered on push. They worked fine.


JakeHillion:
This renames a check because of the new layering. Worth it for the abstraction. People behind will have to rebase before merging.

PR #1749, opened 2025-04-25
scheds/c: Validate CPU count from libbpf

EricccTaiwan:
Ensure the return value of `libbpf_num_possible_cpus()` is positive before use. This avoids undefined behavior when the function fails and returns zero or a negative error code.

Ref : https://docs.ebpf.io/ebpf-library/libbpf/userspace/libbpf_num_possible_cpus/

I believe the Rust-based schedulers might also be affected by the same issue. If this patch is helpful, I’d be happy to send a follow-up patch to add a similar check in the Rust version as well.

### DIFF
--- a/scheds/c/scx_central.c
+++ b/scheds/c/scx_central.c
@@ -61,6 +61,7 @@ restart:
 	skel->rodata->nr_cpu_ids = libbpf_num_possible_cpus();
 	skel->rodata->slice_ns = __COMPAT_ENUM_OR_ZERO("scx_public_consts", "SCX_SLICE_DFL");
 
+	assert(skel->rodata->nr_cpu_ids > 0);
 	assert(skel->rodata->nr_cpu_ids <= INT32_MAX);
 
 	while ((opt = getopt(argc, argv, "s:c:pvh")) != -1) {
--- a/scheds/c/scx_flatcg.c
+++ b/scheds/c/scx_flatcg.c
@@ -6,6 +6,7 @@
  */
 #include <stdio.h>
 #include <signal.h>
+#include <assert.h>
 #include <unistd.h>
 #include <libgen.h>
 #include <limits.h>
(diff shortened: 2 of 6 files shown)


JakeHillion:
Thanks for pointing this out! It looks from the implementation that this function can only fail the first time you call it in a single threaded program, and it's not entirely clear what conditions it fails in.

Have you seen it fail at all? My thinking is we should probably hard error if it's possible for it to fail in the wild rather than assert.

EricccTaiwan:
I haven’t observed it fail during actual execution. That said, while tracing the code paths, I found this potential risk.

I agree that enforcing a hard error would be safer and more appropriate than relying on an assert.

htejun (approved):

EricccTaiwan:
Thanks for the review and approval! I'll leave it as-is unless further changes are needed.

htejun (approved):

EricccTaiwan:
Hi @htejun ,
Thanks for merging!
Now that this PR is merged, would it make sense to also add a `nr_cpus > 0` check on the Rust side?
Or is it fine to leave it as is?

JakeHillion:
> Hi @htejun , Thanks for merging! Now that this PR is merged, would it make sense to also add a `nr_cpus > 0` check on the Rust side? Or is it fine to leave it as is?

Should already be checked as long as we're using `libbpf_rs::num_possible_cpus`. See https://docs.rs/libbpf-rs/latest/src/libbpf_rs/util.rs.html#63-66.

EricccTaiwan:
> > Hi @htejun , Thanks for merging! Now that this PR is merged, would it make sense to also add a `nr_cpus > 0` check on the Rust side? Or is it fine to leave it as is?
> 
> Should already be checked as long as we're using `libbpf_rs::num_possible_cpus`. See https://docs.rs/libbpf-rs/latest/src/libbpf_rs/util.rs.html#63-66.

Thanks for the reply! I'll go check it out.

PR #1751, opened 2025-04-25
ci: track 6.12 lts kernel and test it 6-hourly

JakeHillion:
6.12 is LTS and has several incompatibilities with 6.13+. Add it to the CI tests, 6-hourly for now.

Test plan:
- CI
- Ran separately with `push:` enabled, it worked.


likewhatevs (approved):
nice

PR #1753, opened 2025-04-25
scx_layered: add runtime-based task placement hints

etsal:
Add a degree of support for runtime-based task placement in scx_layered. This includes two additions. First, add a match rule based on the average runtime of a task. Second, add a new configuration option for layers that dictates how sticky their task placement logic is. For now there are two options apart from the standard: Always re-enqueueing in the local cpu, and completely ignoring locality, instead picking the first available CPU in the layer.


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2066,6 +2074,21 @@ static __noinline bool match_one(struct layer_match *match,
...
+			if (!taskc) {
+				scx_bpf_error("could not find task");
+				return false;
+			}
+
+			/* To match, we must get min <= time < max. */
+			if (match->min_avg_runtime_us > avg_runtime_us)
+				return false;
+
+			return avg_runtime_us < match->max_avg_runtime_us;
```
htejun, line 2090: Maybe combine the two tests?
```
return avg >= min && avg < max;
```

htejun (approved):
I wonder whether it'd be more useful if this were implemented as in-layer property rather than a match for a separate layer. What are the expected use cases?

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: > Maybe combine the two tests?
> 
> ```
> return avg >= min && avg < max;
> ```

Will do, the line was kinda long so I split it up

etsal:
> I wonder whether it'd be more useful if this were implemented as in-layer property rather than a match for a separate layer. What are the expected use cases?

I agree, the reason these are layers is because I want to run tasks with similar runtimes on the same CPUs, or rather prevent short-running tasks from running in the same CPUs as long-running tasks. 

I think the problem is that we currently can't have layer subspecifications for different kinds of tasks within them. Ideally we could decide on the attributes (e.g. time slice) of tasks within the layer the same way we currently match tasks into layers, but right now we need to make a layer for each different set of attributes we want to apply. This makes it hard to size layers, e.g., if I have a workload with a bunch of short-running and long-running tasks, how do I decide how many each one needs?

If we had this kind of subspecifications we can use the matching rules and policies in this diff more flexibly.

htejun:
I think this approach is fine for experimenting. However, when do you trigger layer matching? Shouldn't there be something which updates layer membership when avg changes?

etsal:
> I think this approach is fine for experimenting. However, when do you trigger layer matching? Shouldn't there be something which updates layer membership when avg changes?

That is true, there is the implicit assumptions that tasks have a pretty stable average so after we match them we do not have to worry about their average changing enough to warrant moving them and our periodic refreshes keep layers reasonably up-to-date. If we want to be more proactive we can set the refresh_layer flag by keeping the previous average and setting the refresh_layer flag.

htejun:
So, the problem is that the average simply won't be available when the scheduler starts and when a new task is forked. If we just run the matcher once, a lot of, even most, tasks can be stuck in the wrong layer indefinitely.

etsal:
> So, the problem is that the average simply won't be available when the scheduler starts and when a new task is forked. If we just run the matcher once, a lot of, even most, tasks can be stuck in the wrong layer indefinitely.

Ah I see, yes this is a bug - I conflated refresh_cpumasks with refresh_layers. Would making layers that employ this flag to periodically (e.g, 2s) force the refresh_layers flag of their members work? This would have the expected behavior and be inexpensive for layers that are not heavily loaded. Alternatively we can "remember" the matched rule in task_ctx and set the flag for the task if the runtime average goes out of these bounds.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -296,6 +299,12 @@ enum layer_growth_algo {
 	GROWTH_ALGO_STICKY_DYNAMIC,
 };
 
+enum layer_task_place {
```
hodgesds, line 302: What are your thoughts on future values? For example you may want a task that is sticky, but if being woken by a cross layer dependency to migrate. When I first saw the name I thought maybe a separate struct would be nice for having more fine grain controls, but I guess there aren't that many combinations.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1296,6 +1300,11 @@ impl<'a> Scheduler<'a> {
                             mt.kind = bpf_intf::layer_match_kind_MATCH_USED_GPU_PID as i32;
                             mt.used_gpu_pid.write(*polarity);
                         }
+                        LayerMatch::AvgRuntime(min, max) => {
```
hodgesds, line 1303: It might be worth asserting that the `slice_us` for the layer is greater than the `min_avg_runtime_us` .

htejun:
> > So, the problem is that the average simply won't be available when the scheduler starts and when a new task is forked. If we just run the matcher once, a lot of, even most, tasks can be stuck in the wrong layer indefinitely.
> 
> Ah I see, yes this is a bug - I conflated refresh_cpumasks with refresh_layers. Would making layers that employ this flag to periodically (e.g, 2s) force the refresh_layers flag of their members work? This would have the expected behavior and be inexpensive for layers that are not heavily loaded. Alternatively we can "remember" the matched rule in task_ctx and set the flag for the task if the runtime average goes out of these bounds.

Doing matches periodically seems like a reasonble direction; however, I'm not how this can be limited to a subset of tasks on the system. To do that, you'd first have to be able to say "these tasks are candidates for some layers depending on their average runtime". We can construct such conditions but it'd be a bit involved.

scheds/rust/scx_layered/src/bpf/intf.h
etsal: I think that'd be pretty nice, maybe we can turn this into a bitmap in the future if it ends up being a useful match rule. That way we can combine options nicely.

etsal:
Trace produced with the example config with a layer refresh rate of 2s while running stress-ng malloc and CPU stressors.

Screenshot 2025-04-28 at 15-06-23 trace pb (21 MB) - Perfetto UI

scheds/rust/scx_layered/src/main.rs
```diff
@@ -618,6 +625,11 @@ struct Opts {
 
     /// Layer specification. See --help.
     specs: Vec<String>,
+
+    /// Periodically tasks to reevaluate which layer they belong to. Default period of 0
+    /// turns this off.
+    #[clap(long, default_value = "0")]
+    force_layer_refresh_ms: u64,
 }
```
htejun, line 633: Why not tie this to the runtime matches being used? This is not useful for other matches but is critical for the runtime matches. It's unnecessarily error-prone to make this an independent option.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -52,6 +52,7 @@ const volatile u64 min_open_layer_disallow_preempt_after_ns;
 const volatile u64 lo_fb_wait_ns = 5000000;	/* !0 for veristat */
 const volatile u32 lo_fb_share_ppk = 128;	/* !0 for veristat */
 const volatile bool percpu_kthread_preempt = true;
+volatile u64 layer_refresh_seq;
```
htejun, line 55: Maybe `force_layer_refresh_seq` to keep consistent with the option?

scheds/rust/scx_layered/src/main.rs
etsal: Sounds good, I will make it so that this only triggers on layers with AvgRuntime rules and adjust the name accordingly. I initially had that and removed, my rationale was it was that silently ignoring the flag for some layers would be unexpected.

htejun: Maybe it can be named in a way that clarifies where it applies. Also, it'd be better if the users aren't required to set this option to use the runtime matches. ie. If you limit its scope to the ones that need it, you can also pick a default value that makes sense for them instead of 0.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -2571,6 +2592,13 @@ impl<'a> Scheduler<'a> {
                 }
             }
 
+            if enable_layer_refresh && now >= next_layer_refresh_at {
```
hodgesds, line 2595: Could also be a bpf timer, but I suppose this works as well.

etsal:
Updated the refresh to only be triggered for layers that include an AvgRuntime rule. This introduces an imo acceptable edge case where a config that only tracks AvgRuntime in some layers risks a task going into a layer w/o an AvgRuntime rule, at which point it will not rejoin layers with AvgRuntime even if it is eligible.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -880,6 +882,22 @@ s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
 	if (layer_id >= MAX_LAYERS || !(layer_cpumask = lookup_layer_cpumask(layer_id)))
 		return -1;
 
+	if (layer->task_place == PLACEMENT_FLOAT) {
```
htejun, line 885: Can we restructure it a bit to reduce duplications?
```
bool is_float = layer->task_place == PLACEMENT_FLOAT;

if (!is_float && (p->nr_cpus_allowed == 1 || is_migration_disabled(p)) {
   ...
}

cpus_seq = ...;
maybe_refresh...();
if (!(layered_cpumask ...))
    return -1;
if (!(idle_smtmask = ..))
    return -1;

if (is_float)
    goto no_locality;
```

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1079,7 +1098,14 @@ s32 BPF_STRUCT_OPS(layered_select_cpu, struct task_struct *p, s32 prev_cpu, u64
 	if (taskc->layer_id == MAX_LAYERS || !(layer = lookup_layer(taskc->layer_id)))
 		return prev_cpu;
 
-	cpu = pick_idle_cpu(p, prev_cpu, cpuc, taskc, layer, true);
+	if (layer->periodically_refresh && taskc->layer_refresh_seq < layer_refresh_seq_avgruntime)
+		taskc->refresh_layer = true;
```
htejun, line 1102: It probably would be better to do this in `pick_idle_cpu()` as select_cpu() may be skipped in some cases.

htejun (approved):
Generally looks good to me. Left some minor comments.

etsal:
Addressed feedback and resolved merge conflict with cgroups suffix feature.

PR #1755, opened 2025-04-25
scxtop: Add default perf event configuration

hodgesds:
Add configuration for setting the default perf event from either the CLI or config file.

```
$ ./target/release/scxtop --default-perf-event sw:task-clock
```


hodgesds:
I think another fix is needed after #1737 landed.

JakeHillion (approved):

PR #1761, opened 2025-04-26
add cargo fmt git hook

likewhatevs:
Add a git hook to format files like how ci does before committing (i.e. to remove fmt from the list of things everyone needs to care about etc.).

This is enable-able for folks's clones/forks via the following command:
```
git config core.hooksPath .githooks
```

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+echo "Running cargo fmt..."
+cargo fmt --all
+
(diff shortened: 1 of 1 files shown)


JakeHillion:
This isn't guaranteed to be identical because it gets cargo from your path, is it hard to bypass if it disagrees with the CI? It's opt in though so not a big deal.

htejun:
As long as it's not enabled by default, I think it's fine.

htejun (approved):

PR #1765, opened 2025-04-27
scx_rlfifo: Fix CPU assignment logic

EricccTaiwan:
When an idle CPU is selected via `select_cpu()`, assign it directly instead of falling back to `RL_CPU_ANY`. If no idle CPU is available, keep the task running on its previous CPU.

This change makes the CPU assignment logic consistent with the comment description and ensures more deterministic scheduling behavior.


arighi (approved):
Ohh yes I noticed that, but always forgot to fix it, thanks for taking a look!

EricccTaiwan:
No problem! 
I also noticed that the corresponding README should be updated to reflect the corrected CPU assignment logic. 
I'll submit a patch for that as well.

PR #1768, opened 2025-04-28
scx_utils, scx_lavd: Properly handle task's niceness.

multics69:
Incorporate the task's nice weight in calculating the task's service time and weight factor in calculating latency criticality. Essentially, give more CPU time and latency criticality to a less nice task.


hodgesds (approved):
This looks nice, thanks for pulling this out into utils.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -318,7 +318,7 @@ static u64 calc_weight_factor(struct task_struct *p, struct task_ctx *taskc)
 	/*
 	 * Respect nice priority.
 	 */
-	weight_ft = p->scx.weight * weight_boost;
+	weight_ft = p->scx.weight * nice_to_weight(p) * weight_boost;
```
htejun, line 321: I don't think this makes sense. `p->scx.weight` is `nice_to_weight(p) * 10000 / 1024`, so the above is essentially doing `p->scx.weight * p->scx.weight * 1024 / 10000`, which doesn't really make much sense.

multics69: Ah, Thanks @htejun ! I was confused! Will revert the changes!

PR #1772, opened 2025-04-28
scx_rustland_core: Periodically wakeup user-space scheduler

arighi:
Keep track if the user-space scheduler is not running for too long (USERSCHED_TIMER_NS) and in that case explicitly wakeup an idle CPU and force the user-space scheduler to run.

This allows to automatically recover from potential stall conditions when the BPF <-> user-space scheduling pipeline gets stuck.


hodgesds (approved):
Nice!

PR #1773, opened 2025-04-28
scx_rustland: Update doc and default settings

arighi:
With rustland now aligned to a scheduling policy similar to bpfland, update the default time slice settings and the scheduler's documentation to reflect the new design.


hodgesds (approved):
LGTM

PR #1774, opened 2025-04-28
scx_rlfifo: Clarify Round-Robin scheduling

EricccTaiwan:
Update both README and source file to clarify that scx_rlfifo implements a simple Round-Robin scheduler rather than pure FIFO.

Tasks are dequeued in FIFO order but assigned dynamic time slices, enabling preemption and re-enqueueing to achieve Round-Robin execution.


arighi (approved):
LGTM, thanks!

PR #1778, opened 2025-04-29
scx_rustland_core: Move mandatory scheduler options to BPF

arighi:
SCX_OPS_KEEP_BUILTIN_IDLE and SCX_OPS_ENQ_LAST are strictly required for scx_rustland_core to function correctly. Without them, the framework may behave incorrectly, potentially leading to runtime errors or stalls.

To ensure these options are always defined, move them to BPF, instead of relying on the compat APIs. In this way, any missing definitions will result in a build-time error rather than runtime failures.


JakeHillion (approved):
Nice!

PR #1779, opened 2025-04-29
partially revert "migrate to rust 2024, minimal fixes to build all"

JakeHillion:
Revert the edition upgrade but keep the patches to improve compatibility next time this is attempted. Ubuntu and NixOS stable are both behind on Cargo, and this makes development a pain for a small win in performance.

Test plan:
- Built on NixOS with stable. Didn't work before, does work now.
- CI.


likewhatevs (approved):

arighi (commented):
Thanks!

PR #1781, opened 2025-04-29
layered: add suffix matcher

likewhatevs:
add a cgroup matcher to match suffix of cgroup also.

I used suffix to partition layers docker containers tasks were ran on here:
image

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/util.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/util.bpf.c
@@ -124,10 +124,12 @@ bool __noinline match_prefix_suffix(const char *prefix, const char *str, bool ma
 	if (match_suffix)
 		offset = str_len - match_str_len;
 	
-	bpf_for(c, offset, str_len) {
+	// use MAX_PATH instead of str_len for when 
+	// prefix/suffix == string.
+	bpf_for(c, offset, MAX_PATH) {
 		i = c - offset;
 
-		if ((c > str_len) || (i >= str_len) || (i < 0))
+		if ((c > MAX_PATH) || (i >= MAX_PATH) || (i < 0))
 			return false;
 
 		if (match_buf[i] == '\0')
--- a/scheds/rust/scx_layered/src/bpf/util.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/util.bpf.c
@@ -31,7 +31,7 @@ struct {
 	__uint(key_size, sizeof(u32));
 	__uint(value_size, MAX_PATH);
 	__uint(max_entries, 1);
-} prefix_bufs SEC(".maps");
+} match_bufs SEC(".maps");
 
 struct {
 	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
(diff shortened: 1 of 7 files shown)


etsal:
For the logic, instead of calculating the offsets can you just traverse both the match and the string in reverse if we specify the suffix?

etsal:
Also this solves the immediate issue but we don't really care about the suffix, we rather care about a substring of the input that comes after the delimiter. We may need to handle other kinds of matching, e.g.,  instead of a/b/* we may need a/*/c. In that case maybe we could preemptively expand this diff to handle these cases so we don't have to add another matching rule later.

likewhatevs:
> Also this solves the immediate issue but we don't really care about the suffix, we rather care about a substring of the input that comes after the delimiter. We may need to handle other kinds of matching, e.g., instead of a/b/* we may need a/*/c. In that case maybe we could preemptively expand this diff to handle these cases so we don't have to add another matching rule later.

My thinking is we push this into config/this does deal with `a/*/c` in the example:
```
	[
		{ "CgroupPrefix": "system.slice/docker" },
		{ "CgroupSuffix": "11df.scope/" }
	]
```

I was thinking of adding a `CgroupContains` matcher for completeness but I kind of doubt that situation will arise (atm).

likewhatevs:
> For the logic, instead of calculating the offsets can you just traverse both the match and the string in reverse if we specify the suffix?

I think that would be faster for long matches, will do.

etsal:
> > Also this solves the immediate issue but we don't really care about the suffix, we rather care about a substring of the input that comes after the delimiter. We may need to handle other kinds of matching, e.g., instead of a/b/* we may need a/*/c. In that case maybe we could preemptively expand this diff to handle these cases so we don't have to add another matching rule later.
> 
> My thinking is we push this into config/this does deal with `a/*/c` in the example:
> 
> ```
> 	[
> 		{ "CgroupPrefix": "system.slice/docker" },
> 		{ "CgroupSuffix": "11df.scope/" }
> 	]
> ```
> 
> I was thinking of adding a `CgroupContains` matcher for completeness but I kind of doubt that situation will arise (atm).

Fair, personally I'd always take a more complex config than more layered code :)

likewhatevs:
>> For the logic, instead of calculating the offsets can you just traverse both the match and the string in reverse if we specify the suffix?

> I think that would be faster for long matches, will do.

Actually, this would be the same wrt/ work done and (I think) offset calculation/that complexity would still need to be there because the strings are of different length and aren't anchored to an absolute start point like 0, right?

etsal:
> > > For the logic, instead of calculating the offsets can you just traverse both the match and the string in reverse if we specify the suffix?
> 
> > I think that would be faster for long matches, will do.
> 
> Actually, this would be the same wrt/ work done and (I think) offset calculation/that complexity would still need to be there because the strings are of different length and aren't anchored to an absolute start point like 0, right?

Not sure I get wym, if you traverse both the match and the string in reverse why would it matter? 

> > > For the logic, instead of calculating the offsets can you just traverse both the match and the string in reverse if we specify the suffix?
> 
> > I think that would be faster for long matches, will do.
> 
> Actually, this would be the same wrt/ work done and (I think) offset calculation/that complexity would still need to be there because the strings are of different length and aren't anchored to an absolute start point like 0, right?

Not sure I get wym, in the example config the rule 11d.scope will match layers like a/b/211.scope and asdf/333ab11d.scope, right? So if you start reading the string in reverse you can start matching characters like e, p, o, and so on.

scheds/rust/scx_layered/src/bpf/util.bpf.c
```diff
@@ -86,45 +86,62 @@ __hidden char *format_cgrp_path(struct cgroup *cgrp)
...
 		scx_bpf_error("failed to read str");
 		return false;
 	}
 
-	bpf_for(c, 0, max_len) {
+	if (match_str_len > str_len)
+		return false;
+	
+	offset = str_len - match_str_len;
+	// max_path is just for the verifier.
```
htejun, line 123: `MAX_PATH`? but I don't think the comment is necessary.

scheds/rust/scx_layered/src/bpf/util.bpf.c
```diff
@@ -86,45 +86,62 @@ __hidden char *format_cgrp_path(struct cgroup *cgrp)
...
+	bpf_for(c, 0, MAX_PATH) {
 		c &= 0xfff;
-		if (c > len) {
+		
+		if (offset >= MAX_PATH || c >= MAX_PATH || (offset + c) >= MAX_PATH)
```
htejun, line 127: `offset >= MAX_PATH` doesn't need to be tested inside the loop and doesn't `offset +c >= MAX_PATH` imply `c >= MAX_PATH`?

scheds/rust/scx_layered/src/bpf/util.bpf.c
```diff
@@ -86,45 +86,62 @@ __hidden char *format_cgrp_path(struct cgroup *cgrp)
...
 	}
 
-	bpf_for(c, 0, max_len) {
+	if (match_str_len > str_len)
+		return false;
+	
+	offset = str_len - match_str_len;
+	// max_path is just for the verifier.
+	bpf_for(c, 0, MAX_PATH) {
 		c &= 0xfff;
```
htejun, line 125: Can you see whether we can drop this?

scheds/rust/scx_layered/src/bpf/util.bpf.c
```diff
@@ -86,45 +86,62 @@ __hidden char *format_cgrp_path(struct cgroup *cgrp)
...
-		if (str_buf[c] != pre_buf[c])
+
+		if (match_suffix && str_buf[offset + c] == '\0')
+			return true;
+		
+		if (match_suffix && (str_buf[offset + c] != match_buf[c]))
+			return false;	
+
+		if (!match_suffix && (str_buf[c] != match_buf[c]))
 			return false;
```
htejun, line 145: Can't we just move the start of the string according to `match_prefix` and use the same matching body? If `match_prefix` is false, match from the beginning of the string. If true, start at `str_len - match_len`. I don't follow why the match body needs to be different.

scheds/rust/scx_layered/src/bpf/util.bpf.c
likewhatevs: I made both paths the same but am not sure about the thing to drop.

likewhatevs:
I think this addresses everything and am reasonably confident it still works:
Screenshot From 2025-04-30 19-33-05

htejun (approved):

PR #1785, opened 2025-04-30
scx_rustland_core: Reuse task vtime during direct dispatch

arighi:
During direct dispatch, reuse the task's existing vtime instead of using the minimum vtime possible (0ULL).

This can prevent directly dispatched tasks from potentially starving tasks dispatched through the regular scheduling path.


multics69 (approved):
This makes sense.

PR #1791, opened 2025-05-01
scx_rustland_core: Remove unnecessary CPU wakeup from rs_select_cpu()

arighi:
When a user-space scheduler selects an idle CPU via rs_select_cpu(), there's no need to immediately wake it up. At this point, the task still needs to be dispatched through the BPF code, and its exact dispatch time is unknown.

The BPF side already handles CPU wakeups proactively when a task is enqueued in a DSQ, which is more accurate and efficient than waking the CPU prematurely during selection, therefore drop the proactive wakeup from rs_select_cpu().

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -663,15 +663,6 @@ int rs_select_cpu(struct task_cpu_arg *input)
 
 	bpf_task_release(p);
 
-	/*
-	 * Wake-up the CPU if idle. Use SCX_KICK_IDLE to prevent unecessary
-	 * rescheduling events in case the CPU is already awake (since we don't
-	 * know exactly what the user-space scheduler is doing we can't
-	 * implicitly assume that the target CPU is idle here).
-	 */
-	if (cpu >= 0)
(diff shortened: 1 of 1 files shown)


hodgesds (approved):

arighi:
> Maybe a silly question could `task_cpu_arg` be extended so that `SCX_KICK_PREEMPT` could be used? I think that would be the only reason to maybe keep something like this around. Otherwise this seems good.

Possibly, but if the goal is to preempt a CPU, it makes more sense to pass `SCX_ENQ_PREEMPT` through `DispatchedTask::flags`, allowing the CPU to be preempted as soon as the task is enqueued (later via `scx_bpf_dsq_insert_vtime()`), so it can run immediately.

Triggering a reschedule from within `rs_select_cpu()` would simply cause a context switch without a having a task ready to be dispatched. In practice, I can't find any useful use case for preempting a CPU from `rs_select_cpu()`.

hodgesds:
> Triggering a reschedule from within `rs_select_cpu()` would simply cause a context switch without a having a task ready to be dispatched. In practice, I can't find any useful use case for preempting a CPU from `rs_select_cpu()`.

Ahh, that makes sense. Thanks for clarifying that!

PR #1792, opened 2025-05-01
scx_layered: Support task hint map

kkdwivedi:
Add support for creating and pinning a 'scx_layered_task_hint_map' in bpffs so that tasks on the system can open it and update their values to add task hints. Mark the map as 666 so that all tasks are allowed to open it and write to it by default.

As a follow up, we would want to enforce access control such that a bpf_map_update_elem on this map can only happen from a task when the pidfd for the key is the same as its thread ID. This needs to be done using BPF LSM or something similar. Currently anyone who can obtain a pidfd for a task can write its hint. However, this is not something we can fix now as LSM or equivalent support in kernel is missing.

TODO: 

- [ ] Access control so tasks cannot overwrite each others value (low priority)?
- [ ] Consume the hint in `struct task_hint` to perform weighted vruntime progression.

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -630,6 +630,10 @@ struct Opts {
     /// turns this off.
     #[clap(long, default_value = "2000")]
     layer_refresh_ms_avgruntime: u64,
+
+    /// Set the path for pinning the task hint map.
+    #[clap(long, default_value = "")]
+    task_hint_map: String,
 }
 
 fn read_total_cpu(reader: &procfs::ProcReader) -> Result<procfs::CpuStat> {
@@ -1919,6 +1923,17 @@ impl<'a> Scheduler<'a> {
         Self::init_layers(&mut skel, &layer_specs, &topo)?;
         Self::init_nodes(&mut skel, opts, &topo);
 
+        // We set the pin path before loading the skeleton. This will ensure
+        // libbpf creates and pins the map, or reuses the pinned map fd for us,
+        // so that we can keep reusing the older map already pinned on scheduler
+        // restarts.
+        let layered_task_hint_map_path = &opts.task_hint_map;
+        let hint_map = &mut skel.maps.scx_layered_task_hint_map;
+        // Only set pin path if a path is provided.
+        if layered_task_hint_map_path.is_empty() == false {
+            hint_map.set_pin_path(layered_task_hint_map_path).unwrap();
+        }
+
         let mut skel = scx_ops_load!(skel, layered, uei)?;
 
         let mut layers = vec![];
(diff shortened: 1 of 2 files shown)


htejun (approved):

scheds/rust/scx_layered/src/main.rs
```diff
@@ -2614,6 +2630,13 @@ impl<'a> Scheduler<'a> {
 
 impl Drop for Scheduler<'_> {
     fn drop(&mut self) {
+        // Remove the pinned task hint map
+        let _ = self
```
hodgesds, line 2634: Maybe a naive question, but what happens if a task opens the map and then the scheduler is unloaded/reloaded?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -85,6 +85,8 @@ const NR_LLC_LSTATS: usize = bpf_intf::llc_layer_stat_id_NR_LLC_LSTATS as usize;
 
 const NR_LAYER_MATCH_KINDS: usize = bpf_intf::layer_match_kind_NR_LAYER_MATCH_KINDS as usize;
 
+const LAYERED_TASK_HINT_MAP_PATH: &str = "/sys/fs/bpf/scx_layered_task_hint_map";
```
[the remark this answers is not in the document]
kkdwivedi: Will update.

scheds/rust/scx_layered/src/main.rs
kkdwivedi: Not naive, it's probably a bug. The original map disappears and the scheduler stops using it, meanwhile user space tasks have it open and keep updating it. Next time it pins a new map which won't be picked up by the service until it restarts.

One way to address this is to make user space always reopen the map fd when doing the update, but this is a bit ugly, since it now requires two syscalls for every update, and in the future we also want to have __uptr for the same task storage map where we'll register it once during the lifetime of the application.

The other is to leave the map open, and then avoid pinning if we have an existing map. The we can perhaps replace the map fd with the one we get from bpf_obj_get of the existing pinned map, since libbpf will prepare the task storage map when the scheduler restarts, but it requires the maps to match in terms of value size, which should be true.

I think second approach is the way to go.

@htejun WDYT?

etsal (approved):
Makes sense to me.

kkdwivedi:
Updated. I talked to @anakryiko and he suggested usage of `bpf_map__set_pin_path` API to do what we were discussing. This should reuse the pinned map fd instead of recreating a new one.

First time layered runs:
```
bpf(BPF_OBJ_GET, {pathname="/sys/fs/bpf/scx_layered_task_hint_map", bpf_fd=0, file_flags=0, path_fd=0}, 20) = -1 ENOENT (No such file or directory)
bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_TASK_STORAGE, key_size=4, value_size=32, max_entries=0, map_flags=BPF_F_NO_PREALLOC, inner_map_fd=0, map_name="scx_layered_tas", map_ifindex=0, btf_fd=80, btf_key_type_id=2, btf_value_type_id=25, btf_vmlinux_value_type_id=0, map_extra=0}, 80) = 81
bpf(BPF_OBJ_PIN, {pathname="/sys/fs/bpf/scx_layered_task_hint_map", bpf_fd=9, file_flags=0, path_fd=0}, 20) = 0
```

Second time with the pan already there:
```
bpf(BPF_OBJ_GET, {pathname="/sys/fs/bpf/scx_layered_task_hint_map", bpf_fd=0, file_flags=0, path_fd=0}, 20) = 81
```

kkdwivedi:
So CI failed because in some of the cases, there is no bpffs where we attempt to put the task storage map by default.

We can disable pinning by default and only enable when --task-hint-map= is provided.

etsal (approved):
I assume there will be a part 2 where the BPF side will read from the hint map if passed the right option?

kkdwivedi:
> I assume there will be a part 2 where the BPF side will read from the hint map if passed the right option?

Yes.

Issue #1588, opened 2025-03-31
scxtop: Don't crash when scheduler does not use scx_stats


hodgesds:
Currently `scxtop` will crash when a scheduler doesn't use `scx_stats`. On the [`App`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/app.rs#L75) struct the stats client should be handled in a way that doesn't cause `scxtop` to crash.

PR #1796, opened 2025-05-01
scxtop: Make stats client optional to avoid panic

EricccTaiwan:
`scxtop` currently panics if `StatsClient` fails to connect, which happens with schedulers that do not expose `scx_stats` over the default socket.

This patch wraps `stats_client` in `Option<>` and conditionally performs stats fetching. This avoids crashing when stats are unavailable.

Closes: #1588


hodgesds (approved):
Awesome, thanks for implementing this!

PR #1797, opened 2025-05-01
Revert "Merge pull request #1781 from likewhatevs/layered-more-cgroup…

likewhatevs:
Something outside of the body of the match_prefix_suffix is problematic for older verifiers, so revert the suffix matcher to avoid this issue for now.

This reverts commit 1a77ba56a2307e8fcacb0a2ac5db97fb1b6c30ec, reversing changes made to f40a355a41b30b4e18f9985b16e41d9dc7854a25.


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2045,8 +2042,8 @@ static __noinline bool match_one(struct layer_match *match,
 		if (!taskc->join_layer[0])
 			return false;
 
-		return match_prefix_suffix(match->comm_prefix, taskc->join_layer,
-			false);
+		return match_prefix(match->comm_prefix, taskc->join_layer,
+			SCXCMD_COMLEN);
```
etsal, line 2046: Should this be MAX_COMM?

etsal: Nevermind looking at the context this makes sense, but I am ont sure if match_prefix_suffix shoud be passed the max length as an argument because it is different for SCXCMD_JOIN and everything else.

etsal (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: Yeah tbh think I am going to try to fix this one fwd in part because the code around this is nicer than it was before this change.

PR #1802, opened 2025-05-02
scx_p2dq: Make interactive sticky work again

hodgesds:
Make the `--interactive-sticky` no longer work as a noop and fast path the `pick_idle_cpu` path for interactive tasks.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -470,6 +470,12 @@ static s32 pick_idle_cpu(struct task_struct *p, task_ctx *taskc,
 	if (!idle_cpumask || !idle_smtmask)
 		goto found_cpu;
 
+	if (interactive_sticky && interactive) {
+		cpu = prev_cpu;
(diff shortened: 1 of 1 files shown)


hodgesds:
> Makes sense.
> 
> Maybe we can add a comment to clarify that interactive tasks won't be migrated on wakeup when interactive_sticky is enabled, but they can still migrate depending on CPUs availability: if their CPU is busy, they'll be added to the LLC DSQ and in `ops.dispatch()` the first CPU that consumes the task can potentially trigger a migration.

From doing some random server benchmarks I'm not sure if this works that great compared to doing more work and trying to find an idle CPU. I'll do some more testing this weekend on gaming workloads where I think CPU locality + wakeup latency seem to be the most critical factors.

multics69 (commented):
I am curious if this implies an interactive task is *never* considered for load balancing. I think there should be in order to use all available CPUs.

hodgesds:
> I am curious if this implies an interactive task is _never_ considered for load balancing. I think there should be in order to use all available CPUs.

By default they are skipped for load balancing, only if `--dispatch-lb-interactive` is configured then load balancing is done.

Issue #1803, opened 2025-05-02
scxtop: Error compiling for aarch64/riscv64


frelon:
Building the project for aarch64/riscv64 fails with the following error:

```
$ meson compile -C build-aarch64
...
FAILED: Cargo.toml.__PHONY__
/usr/bin/meson --internal exe --unpickle /home/frelon/src/scx/build-aarch64/meson-private/meson_exe_cargo_b941442f0071e0429b019d53682d6c3956f0c011.dat
while executing ['/home/frelon/.cargo/bin/cargo', 'build', '--manifest-path=../Cargo.toml', '--target-dir=.', '--quiet', '--target', 'aarch64-unknown-linux-gnu']
--- stdout ---

--- stderr ---
error[E0308]: mismatched types
   --> tools/scxtop/src/lib.rs:393:52
    |
393 |                 let comm = unsafe { CStr::from_ptr(event.event.wakeup.comm.as_ptr()) }
    |                                     -------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `*const u8`, found `*const i8`
    |                                     |
    |                                     arguments to this function are incorrect
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`
note: associated function defined here
   --> /home/frelon/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ffi/c_str.rs:264:25
    |
264 |     pub const unsafe fn from_ptr<'a>(ptr: *const c_char) -> &'a CStr {
    |                         ^^^^^^^^
...
```

It seems that using `char` in the [scxtop intf.h](https://github.com/sched-ext/scx/blob/c6604d8581f6adcf78f8f7cf057f873e0a04f1e8/tools/scxtop/src/bpf/intf.h#L57) generates an array of `i8`, but what `CStr` accepts depends on the platform (`char` is signed by default on x86, but not on the other platforms)

PR #1804, opened 2025-05-02
scxtop: Switch from char to u8

frelon:
char is signed by default on x86 but not on aarch64/riscv64, this produces a build-error when trying to use CStr to load a pointer to the generated skel.

Switching to u8 instead of char and modifying the rust code to use String::from_utf8_lossy compiles and runs on the specified platforms.

Fixes: #1803


hodgesds (approved):
Nice, thanks for making this work!

PR #1807, opened 2025-05-03
scx_lavd: Do not deprioritize a new task.

multics69:
scx_lavd has deprioritized a new task to avoid a fork-bomb attack and to make a judicious scheduling decision after knowing its behavior by giving a hefty greedy penalty (5x). However, this hampers workloads that create many short-lived tasks. Hence, the deprioritization should be dropped, and a new task should be treated as an average task whose service time is on average so as not to give a penalty on the deadline.


arighi (approved):
Makes sense. I think there are better ways to prevent fork bombs at the system level and from a scheduler perspective it's actually good to not de-prioritize new short-lived tasks, or it may impact on system responsiveness. For example, it's way more useful to be able to run 'kill' quickly from the command line when the system is too much overloaded.

multics69:
> Makes sense. I think there are better ways to prevent fork bombs at the system level and from a scheduler perspective it's actually good to not de-prioritize new short-lived tasks, or it may impact on system responsiveness. For example, it's way more useful to be able to run 'kill' quickly from the command line when the system is too much overloaded.

Yes, I agree.

PR #1809, opened 2025-05-03
scx_rustland_core: fix nr_queued not decremented after dequeue

EricccTaiwan:
The `dequeue_task()` function called `saturating_sub(1)` on `nr_queued` but did not assign the result back to the field, leaving the counter unchanged. Assign the decremented value back to ensure `nr_queued` is updated correctly.


arighi (approved):
Looks good, thanks for catching this.

PR #1819, opened 2025-05-06
scx_userspace_arena: add api to allocate from arenas in userspace

JakeHillion:
Arenas and our c-side allocators allow us to store more complex data
structures. However some of these data structures need to be managed
from Rust side, like mutable configuration. Create an allocator usable
from Rust backed by the arena memory.

This allocator does not directly interoperate with the BPF side
allocator. Instead both we and the bpf side allocator treat
`bpf_arena_alloc_pages` as `mmap`, and allocate separately on top of
this. We'll get some external fragmentation here but it's fine. All
memory allocated in BPF must be free'd in BPF, and all memory allocated
in Rust must be free'd in Rust. We'll work out the semantics of this,
and whether we need a better system for this, with experience.

Implementation is roughly:
- Add syscall progs to the shared `sdt_alloc.c` which expose
  `bpf_arena_{,de}alloc_pages` to userspace.
- Write an Allocator trait based on the unstable Rust allocator trait
  (we should switch to that eventually).
- Use a crate for a buddy allocator on top of our BPF side allocators to
  manage smaller units of memory.
- Implement Allocator for chaos's skel and store this allocator.

This allows us to allocate large amounts of memory in chaos dynamically,
and share pointers between kernel and userspace. It's not yet used to do
anything useful, though has been tested with trivial examples.

Next step is writing some data structures like an equivalent of `Box`
that support our allocator. Once the Allocator interface is out of Rust
nightly we can switch to that and use upstream datastructures, but I'm
not hoping that'll be soon.

Test plan:
- Allocated an arena pointer and copied a string into it. Stored this in
  a global in chaos and bpf_printk'd it. It worked!


JakeHillion:
Hmm, this is getting intermittent lockdep errors followed by a kernel panic ([this path](https://github.com/torvalds/linux/blob/0d8d44db295ccad20052d6301ef49ff01fb8ae2d/kernel/locking/lockdep.c#L979-L988)). I'm not sure if this is a kernel thing or due to this code change. Will need some further debugging to be confident, though I'm not sure why this change would have caused it given there's no change to the BPF except for two new syscall progs.

htejun (approved):
Left minor comments. I just scanned through it and it seemed okay to me.

JakeHillion:
Prepended/renamed to `scx_userspace_*` to avoid confusion

rust/scx_userspace_arena/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "scx_userspace_arena"
+version = "1.0.11"
+edition = "2024"
```
arighi, line 4: @JakeHillion this breaks the build on Ubuntu, similar to https://github.com/sched-ext/scx/commit/b4170148639b7b2ea1669514cfd786328bfd8ad2.

Do we really need 2024 here?

PR #1821, opened 2025-05-07
layered: only preempt scx tasks

likewhatevs:
I *think* this should get the job done wrt/ avoiding preempting non sched_ext tasks. This uses tick to store the sched policy of the task a CPU is running to cpu_ctx. 

I think this should work well so long as slice sizes are greater than 1/CONFIG_HZ, and might work OK when there are a mix of SCHED_RT tasks and SCHED_SCX tasks w/ preempt_first and yield_ignore (i.e. once tick happened while the RT task was running, bouncing between tasks would end, I think).


JakeHillion:
I don't think this works, reading the docs for tick at https://github.com/torvalds/linux/blob/0d8d44db295ccad20052d6301ef49ff01fb8ae2d/kernel/sched/ext.c#L330-L338:

```
	/**
	 * @tick: Periodic tick
	 * @p: task running currently
	 *
	 * This operation is called every 1/HZ seconds on CPUs which are
	 * executing an SCX task. Setting @p->scx.slice to 0 will trigger an
	 * immediate dispatch cycle on the CPU.
	 */
	void (*tick)(struct task_struct *p);
```

Which would imply the new field is always set to true. Have you seen an example where this works?

likewhatevs:
> 	 * This operation is called every 1/HZ seconds on CPUs which are
>	 * executing an SCX task. Setting @p->scx.slice to 0 will trigger an

If that's the case (that tick isn't on all CPUs) I don't think this is doable w/o changing something kernel side, but I'd kind of think it would be tbh?

PR #1822, opened 2025-05-07
scx_tickless: Improve scheduling timer

arighi:
Avoid pinning the scheduling timer to a single primary CPU, instead, allow it to run on any available CPU. While this may introduce some minor system noise, it improves the overall throughput, especially when the system is fully saturated.

Moreover, rather than cycling through all CPUs to identify the idle ones, directly iterate over the idle CPUs using the idle cpumasks. This reduces the overhead of finding which CPUs are idle, especially on large systems.


multics69 (approved):
LGTM

PR #1826, opened 2025-05-07
scx_central: Suppress potential unused param warning

EricccTaiwan:
Cast the unused parameter to void to prevent potential -Wunused-parameter warnings under stricter compiler flags. No functional changes.


htejun:
Let's not do this. That warning flag isn't widely used in general and in kernel tree.

PR #1827, opened 2025-05-07
scx_central: Remove unused -p option from getopt()

EricccTaiwan:
Drop the -p option from the getopt option string since it was not handled in the switch statement.


etsal:
In general, scx_central and most other C schedulers only serve as simple examples. However, they are also part of the kernel tree code and have to be manually synced through this repo. Changes to them are churn for no tangible benefit.

EricccTaiwan:
Got it!

PR #1828, opened 2025-05-07
layered: dont preempt if policy is not sched_ext

likewhatevs:
Don't preempt tasks with sched policy that is not sched_ext.


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1137,6 +1138,20 @@ static bool try_preempt_cpu(s32 cand, struct task_struct *p, struct task_ctx *ta
...
+	if (!rq) {
+		scx_bpf_error("failed to get rq to determine preempt eligibility for cpu %d", cand);
+		return false;
+	}
+	
+	/*
+	 * Don't preempt if sched policy is not sched_ext.
+	 * 
+	 */
+	if (rq->curr->policy != SCHED_EXT)
```
htejun, line 1152: This would exclude preempting an idle CPU. Also, SCHED_NORMAL tasks are on SCX when switching all tasks. `p->sched_class` points to the actual scheduling class in use. Comparing that against `ext_sched_class` and `idle_sched_class` is probably better.

likewhatevs: Ahh thanks -- yeah that makes sense wrt/ IDLE CPUs. Will do wrt/ the rest, thank you!

likewhatevs: I used policy previously to avoid having to sort this out:

```
❯ sudo ./target/release/scx_layered --run-example --stats 2
...
01:33:59 [WARN] libbpf: extern (var ksym) 'ext_sched_class': not found in kernel BTF
...

❯ bpftool btf dump file /sys/kernel/btf/vmlinux | grep ext_sched_class

❯ cat /proc/kallsyms | grep ext_sched_class
0000000000000000 D ext_sched_class
```

IIUC -- I think there needs to be an `EXPORT_SYMBOL_GPL` here to enable using `ext_sched_class` etc., in bpf programs: https://elixir.bootlin.com/linux/v6.14.5/source/kernel/sched/sched.h#L2532

The comment above that macro (and how it is used) make me think the usual "copy-paste the definition into common.bpf.h" approach might not work for this, but I'll try that and if it doesn't work, go back to using sched_policy (which I think would work, if I get the policies to select for right).

likewhatevs:
hmm, this example:

```
stress-ng --perf --cpu 4 --cpu-load 60 --cpu-method fft
sudo taskset -c 14,15,30,31 stress-ng --sched fifo --perf --cpu 4 --cpu-load 60 --cpu-method fft
```

w/ this config:
```
[
        {
                "name": "second",
                "matches": [
                        [{ "CommPrefix": "stress" }, {"UIDEquals": 1000}]
                ],
                "kind": {
                        "Confined": {
                                "cpus_range":  [6, 6],
                                "util_range": [0.4, 0.85],
                                "preempt": true,
                                "preempt_first": true,
                                "growth_algo": "Reverse",
                                "slice_us": 1000,
                                "yield_ignore": 0.50
                        }
                }
        },
        {
                "name": "third",
                "matches": [
                        []
                ],
                "kind": {
                        "Open": {
                                "growth_algo": "Linear"
                        }
                }
        }
]
```

causes stalls w/ kthreads reliably on both main and this PR.

a stall log is here: https://gist.github.com/likewhatevs/a0336e9169fdad0a4c6c225e97ea0ed5

A thing I noticed while debugging this is that , with or without these changes to not preempt rt tasks, voluntary switches in the rt task seemed to determine involuntary switches in the scx task (i.e. scx wasn't preempting rt):
Screenshot From 2025-05-07 23-40-44

and I can't seem to get nivcsw of the rt task to go up on main or this PR. 

Reading https://man7.org/linux/man-pages/man7/sched.7.html I think that makes sense (i.e. if rt is greater priority than scx, rt would preempt), but then idk how scx would deal w/ that?

likewhatevs:
think this works now (i.e. skip preempt moves like I'd expect).

when running the following two cmds w/ the test config on PR (to create contention w/ preemption enabled w/o tasks w/ a different sched class):
```
sudo taskset -c 14,15,30,31 stress-ng --cpu 4 --cpu-load 50
stress-ng --cpu 4 --cpu-load 50
```
Screenshot From 2025-05-08 21-11-29

when running the following two cmds w/ the test config on PR (to create contention w/ preemption enabled w/ tasks w/ a different sched class):
```
sudo taskset -c 14,15,30,31 chrt -f 50 stress-ng --cpu 4 --cpu-load 50
stress-ng --cpu 4 --cpu-load 50
```
Screenshot From 2025-05-08 21-11-09

etsal (approved):
Looks good, dumping the class addresses from kallsyms and comparing against is perfectly valid imo.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1824,6 +1833,9 @@ impl<'a> Scheduler<'a> {
             }
         }
 
+        skel.maps.rodata_data.ext_sched_class_addr = get_kallsyms_addr("ext_sched_class")?;
+        skel.maps.rodata_data.idle_sched_class_addr = get_kallsyms_addr("idle_sched_class")?;
```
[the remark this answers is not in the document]
likewhatevs: thx, opened https://github.com/sched-ext/scx/pull/1843 for this.

PR #1829, opened 2025-05-08
scx_p2dq: Refactor slice handling

hodgesds:
Add slice_ns to the task_ctx field so that slice times can be dynamically adjusted per task. Refactor various lookups to use the new slice_ns field on the task_ctx.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -710,8 +710,7 @@ static __always_inline s32 p2dq_select_cpu_impl(struct task_struct *p, s32 prev_
 	cpu = pick_idle_cpu(p, taskc, prev_cpu, wake_flags, &is_idle);
 	if (is_idle) {
 		stat_inc(P2DQ_STAT_IDLE);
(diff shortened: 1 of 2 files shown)


multics69 (approved):
If I understand the changes correctly, there is no functional change in this commit. If so, it would be good to mention that in the commit message. LGTM.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -955,6 +953,7 @@ void BPF_STRUCT_OPS(p2dq_stopping, struct task_struct *p, bool runnable)
 	} else {
 		stat_inc(P2DQ_STAT_DSQ_SAME);
 	}
+	taskc->slice_ns = task_slice_ns(p, taskc->dsq_index);
```
arighi, line 956: Wouldn't be more "accurate" to update the task's time slice in ops.enqueue(), instead of ops.stopping()? Potentially the task may go to sleep for a while and when it's resumed dsq time slices may have been updated.

hodgesds: Yeah, that's a good call. I'm planning on reworking all the time slice handling soon, but let me fix that.

PR #1831, opened 2025-05-08
scx_rustland: replace `or_insert_with_key` with `or_insert`

EricccTaiwan:
The use of `or_insert_with_key()` is unnecessary here since the inserted value does not depend on the key. 
`or_insert()` provides the same behavior more simply.

This minor change improves clarity and readability, and I believe it's a reasonable cleanup rather than churn.


arighi (approved):
LGTM, thanks.

PR #1832, opened 2025-05-08
layered: use passed in p instead of bpf_task_from_pid

likewhatevs:
layered: use passed in p instead of bpf_task_from_pid.

This fixes the situation where a task disappears (i.e. ends or it's pid is changed) wrt/ `bpf_task_from_pid` inside matcher code, causing layered to report unmatched tasks.

This looks like it was compat code around verifier issues previously (i.e. pid lookups are a nice, but somewhat problematic, way to make the verifier trust pointers).

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2120,20 +2120,20 @@ static __noinline bool match_one(struct layer_match *match,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


htejun:
Maybe explain what this fixes?

htejun (approved):

etsal (approved):
Makes sense, rn it seems to be hitting a verification failure but marking as accepted so we can merge as soon as it passes.

PR #1833, opened 2025-05-08
scx_p2dq: Refactor vtime handling of migrating tasks

hodgesds:
When updating the vtime of a task moving across LLCs give tasks with a lower weight a boost in vtime. Refactor vtime updating into a helper method and update the enqueue method to properly update vtime after selecting a CPU.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -114,6 +114,11 @@ static __always_inline u64 dsq_time_slice(int dsq_index)
 	return dsq_time_slices[dsq_index];
 }
 
+static __always_inline u64 max_dsq_time_slice(void)
+{
+	return dsq_time_slices[nr_dsqs_per_llc - 1];
+}
+
 static __always_inline u64 task_slice_ns(struct task_struct *p, int dsq_index)
 {
 	return p->scx.weight * dsq_time_slice(dsq_index) / 100;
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -731,89 +767,67 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
...
 	/*
-	 * Push per-cpu kthreads at the head of local dsq's and preempt the
-	 * corresponding CPU. This ensures that e.g. ksoftirqd isn't blocked
-	 * behind other threads which is necessary for forward progress
-	 * guarantee as we depend on the BPF timer which may run from ksoftirqd.
+	 * Per-cpu kthreads are considered interactive and dispatched directly
+	 * into the local DSQ.
 	 */
-	if ((p->flags & PF_KTHREAD) && !taskc->all_cpus &&
+	if ((p->flags & PF_KTHREAD) && p->cpus_ptr == &p->cpus_mask && p->nr_cpus_allowed == nr_cpus &&
```
hodgesds, line 780: This is purely to avoid looking up the `task_ctx` for affinitized kthreads.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -279,6 +284,37 @@ static bool can_pick2(task_ctx *taskc)
...
+	 */
+	if (taskc->llc_id == cpuc->llc_id) {
+		u64 max_slice = max_dsq_time_slice();
+		u64 vtime_min = vtime_now - max_slice;
+
+		p->scx.dsq_vtime = max(p->scx.dsq_vtime, vtime_min);
+		return;
+	}
+
+	// Give a slight boost to high priority tasks migrating across LLCs.
```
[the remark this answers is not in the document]
hodgesds: Let me do some more testing on this. I was trying to avoid extra lookups of the previous `llc_ctx` and use some heuristic to give the right tasks enough of a bump. Perhaps in this case is is worth doing the check though.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -279,6 +284,37 @@ static bool can_pick2(task_ctx *taskc)
...
+	if (p->scx.weight < 100) {
```
multics69, line 309: If I recall correctly, higher `scx.weight` means higher priority. If so, it should be `p->scx.weight > 100`.

multics69: Also, it will be a good idea to document the idea behind this `if-else` block more in detail.

EricccTaiwan: > If I recall correctly, higher `scx.weight` means higher priority. If so, it should be `p->scx.weight > 100`.

Minor note: $NICE < 0$ maps to `scx.weight > 100` in terms of scheduling weight.

hodgesds: Going to go with the simple approach and just set the set the vtime to the current llc vtime.

PR #1834, opened 2025-05-08
scx_rustland: fallback to original CPU if no idle CPU is available

EricccTaiwan:
If no idle CPU is found, fall back to the task's original CPU instead of `RL_CPU_ANY` to avoid unnecessary migrations.

Ref: #1765


hodgesds:
Yeah, this seems like it should be a better choice.

arighi:
NACK, let's not do this. If the system is fully saturated we want tasks to be consumed by the first CPU available to improve work conservation. If we don't do this, we lose the key advantage of this scheduler, particularly its ability to keep the system responsiveness under (almost) all conditions.

EricccTaiwan:
That makes much more sense after your explanation, thanks!
I’ll do some experiments to see how `RL_CPU_ANY` helps under load.

PR #1843, opened 2025-05-09
layered: cleanup skip-preempt

likewhatevs:
Have layered only attempt to skip preemption's that it could not accomplish (i.e. due to tasks with RT sched class) when kallsyms addresses for idle sched class and ext sched class are present.

At the present, layered would exit if those addresses were unobtainable.

still works when kallsyms present for both the skip and not-skip cases:
Screenshot From 2025-05-09 13-37-40
Screenshot From 2025-05-09 13-37-30

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1135,7 +1135,7 @@ static bool try_preempt_cpu(s32 cand, struct task_struct *p, struct task_ctx *ta
 	struct cpu_ctx *cpuc, *cand_cpuc, *sib_cpuc = NULL;
 	struct rq *rq = NULL;
 	s32 sib;
-	struct sched_class *ext_sched_class = NULL, *idle_sched_class = NULL;
+	struct sched_class *ext_sched_class, *idle_sched_class;
 
 	if (cand >= nr_possible_cpus || !bpf_cpumask_test_cpu(cand, p->cpus_ptr))
 		return false;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -60,6 +60,7 @@ const volatile bool enable_gpu_support = false;
 /* Delay permitted, in seconds, before antistall activates */
 const volatile u64 antistall_sec = 3;
 const u32 zero_u32 = 0;
+const volatile bool enable_skip_preempt = false;
 
 /*
  * XXX sched classes should be exported kernel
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1148,15 +1149,17 @@ static bool try_preempt_cpu(s32 cand, struct task_struct *p, struct task_ctx *ta
...
-	ext_sched_class = (struct sched_class *)(unsigned long long)ext_sched_class_addr;
-	idle_sched_class = (struct sched_class *)(unsigned long long)idle_sched_class_addr;
+	if (enable_skip_preempt) {
+		ext_sched_class = (struct sched_class *)(unsigned long long)ext_sched_class_addr;
+		idle_sched_class = (struct sched_class *)(unsigned long long)idle_sched_class_addr;
 
-	if (rq && (rq->curr->sched_class != ext_sched_class) &&
-		(rq->curr->sched_class != idle_sched_class)) {
-		if (!(cpuc = lookup_cpu_ctx(-1)))
+		if (rq && (rq->curr->sched_class != ext_sched_class) &&
```
etsal, line 1156: Can skip the enable_skip_preempt variable and roll the outer loop into the inner one by writing (ext_sched_class_addr != NULL), but it's just for terseness.

likewhatevs: ahh that makes sense, thx.

likewhatevs: Still works wrt/ doing the right thing when kallsyms is present, does the C look about right (I think it is/that one can cast null like that fine)?

etsal (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: The C does look right, I assume the verifier has no issue with using the addr as a scalar right? Other than that yeah, the address is in the BSS so should be init'ed to 0, and will always be non-zero when both symbols are read.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1135,7 +1135,7 @@ static bool try_preempt_cpu(s32 cand, struct task_struct *p, struct task_ctx *ta
 	struct cpu_ctx *cpuc, *cand_cpuc, *sib_cpuc = NULL;
 	struct rq *rq = NULL;
 	s32 sib;
-	struct sched_class *ext_sched_class, *idle_sched_class;
+	struct sched_class *ext_sched_class = NULL, *idle_sched_class = NULL;
```
htejun, line 1138: I don't think you need this. The variables are unconditionally set below, right?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1833,8 +1833,15 @@ impl<'a> Scheduler<'a> {
...
-        skel.maps.rodata_data.ext_sched_class_addr = get_kallsyms_addr("ext_sched_class")?;
-        skel.maps.rodata_data.idle_sched_class_addr = get_kallsyms_addr("idle_sched_class")?;
+        let ext_sched_class_addr = get_kallsyms_addr("ext_sched_class");
+        let idle_sched_class_addr = get_kallsyms_addr("idle_sched_class");
+
+        if ext_sched_class_addr.is_ok() && idle_sched_class_addr.is_ok() {
+            skel.maps.rodata_data.ext_sched_class_addr = ext_sched_class_addr.unwrap();
+            skel.maps.rodata_data.idle_sched_class_addr = idle_sched_class_addr.unwrap();
+        } else {
+            warn!("skip_preempt is not supported, ignoring");
```
htejun, line 1843: Maybe mention that the addresses couldn't be read from /proc/kallsyms so that users can hunt down what's not working more easily?

likewhatevs: I'll add this, but also like, do we think these symbols could be exported kernel side such that we don't need to do this in the future or no?

htejun: It will probably remain this way.

likewhatevs: will make this into a helper in common then, thanks!

htejun: Oh, let's keep it in layered for now. We can move it to utils when there are other users.

etsal:
Followup to this: If we can check the current running task for the rq instead of the rq itself, #1844 solves this problem a bit more idiomatically.

likewhatevs:
> Followup to this: If we can check the current running task for the rq instead of the rq itself, https://github.com/sched-ext/scx/pull/1844 solves this problem a bit more idiomatically.

Just to make sure I follow -- more/less, store acquired/released status on cpuc, toggle it with those two callbacks, and use that instead of checking the rq?

etsal:
> > Followup to this: If we can check the current running task for the rq instead of the rq itself, #1844 solves this problem a bit more idiomatically.
> 
> Just to make sure I follow -- more/less, store acquired/released status on cpuc, toggle it with those two callbacks, and use that instead of checking the rq?

Exactly, it requires the use of the two extra scx methods. As long as there is a kfunc for grabbing the current running task for a cpu, that should work (not sure whether that exists given that it sounds inherently racy, i.e. the function would be returning a task that may be switched out immediately after being returned).

PR #1844, opened 2025-05-09
scx_p1dq: Acccount for non scx usage

hodgesds:
Add accounting for non scx usage for LLC utilization. This should make load balancing decisions more accurate.


etsal:
This assumes the CPU was running full throttle when not in scx, but IIUC that includes time spent in the idle scheduling class, so isn't this going to skew accounting?

hodgesds:
> This assumes the CPU was running full throttle when not in scx, but IIUC that includes time spent in the idle scheduling class, so isn't this going to skew accounting?

Hmmm good observation... let me think about this some more may be able to test if `TASK_IDLE` for that case.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -36,6 +36,7 @@ char _license[] SEC("license") = "GPL";
 
 UEI_DEFINE(uei);
 
+
```
etsal, line 39: Nit: Stray line?

etsal (approved):
Assuming PF_IDLE covers it, looks good.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -1594,6 +1600,38 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(p2dq_init)
...
+void BPF_STRUCT_OPS(p2dq_cpu_release, s32 cpu,
+		    struct scx_cpu_release_args *args)
+{
+	struct cpu_ctx *cpuc;
+	u64 now = scx_bpf_now();
+
+	if (!(cpuc = lookup_cpu_ctx(cpu)))
+	    return;
+
+	if (is_idle_task(args->task)) {
```
arighi, line 1627: Hm... I don't think `ops.cpu_release()` is called if we're switching to a scheduling class below SCHED_EXT. See: https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/sched/ext.c#n3171

So this condition should be always false, unless I'm missing something...

multics69, line 1627: A better approach would be to use `ops.update_idle()` to track the idle transition.

hodgesds: I completely missed that, let me try that!

EricccTaiwan:
> scx_p2dq: Acccount for non scx usage

Nit: s/Acccount/Account

PR #1852, opened 2025-05-10
scx_rustland: Refactor TaskTree pop to remove redundant else

EricccTaiwan:
The else branch in `TaskTree::pop` was returning `None` explicitly, even though that is already the default behavior when using `Option::map`. This refactor simplifies the logic and improves readability by adopting a more idiomatic Rust approach.

No change in functionality; this is a cleanup for clarity.


arighi (approved):
Looks good, it's more idiomatic and maintains the same behavior. Thanks!

PR #1854, opened 2025-05-11
scx_tickless: Misc improvements

arighi:
A sync wakeup change that improves performance for pipe-intensive workloads, along with some refactoring of timer initialization and task dispatch to eliminate unnecessary overhead.

### DIFF
--- a/scheds/rust/scx_tickless/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_tickless/src/bpf/main.bpf.c
@@ -198,11 +198,19 @@ static inline u64 tick_interval_ns(void)
 
 s32 BPF_STRUCT_OPS(tickless_select_cpu, struct task_struct *p, s32 prev_cpu, u64 wake_flags)
 {
+	s32 cpu;
+
 	/*
 	 * Always route wakeups to a CPU from the primary group to minimize
 	 * noise on the other CPUs.
+	 *
+	 * Also make sure to lock the selected CPU, so that it's not picked
+	 * when distributing tasks in dispatch_all_cpus().
 	 */
-	return pick_primary_cpu();
+	cpu = pick_primary_cpu();
+	scx_bpf_test_and_clear_cpu_idle(cpu);
+
+	return cpu;
 }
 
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_tickless/src/bpf/main.bpf.c
```diff
@@ -371,46 +402,36 @@ static int sched_timerfn(void *map, int *key, struct bpf_timer *timer)
...
- * Start the BPF timer on a target CPU (if not already started).
+ * Start the BPF timer on a target CPU.
  */
-static void fire_timer(s32 cpu)
+static void init_timer(s32 cpu)
 {
 	struct cpu_ctx *cctx;
 	int ret;
 
+	if (!is_primary_cpu(cpu))
```
hodgesds, line 417: I had to look through the timer docs again, it's really annoying that `bpf_timer_start` only can pin with `BPF_F_TIMER_CPU_PIN` and that only works with the caller CPU.

scheds/rust/scx_tickless/src/bpf/main.bpf.c
```diff
@@ -643,7 +664,13 @@ int enable_primary_cpu(struct cpu_arg *input)
...
 s32 BPF_STRUCT_OPS_SLEEPABLE(tickless_init)
 {
-	return scx_bpf_create_dsq(SHARED_DSQ, -1);
+	int ret;
+
+	ret = scx_bpf_create_dsq(SHARED_DSQ, -1);
+	if (ret < 0)
+		return ret;
+
+	init_timer(bpf_get_smp_processor_id());
```
hodgesds, line 673: I guess the other way to init the timer would be use another prog timer and init from userspace. Then you could manually set the affinity and have a way of directly placing a timer on a particular CPU.

hodgesds (approved):
WAKE_SYNC stuff makes sense, LGTM

scheds/rust/scx_tickless/src/bpf/main.bpf.c
arighi: Yeah, apparently there's not a way to pin the timer to CPU N, where N != current CPU.

multics69 (approved):
LGTM

PR #1857, opened 2025-05-11
scx_rustland: Refactor dispatch_tasks to simplify control flow

EricccTaiwan:
1. Replaces a match statement with an if-let binding in dispatch_tasks, as the None arm was unused.

2. Also replaces a match on Result with an is_err() check, since only the Err case was handled and the Ok arm was empty.


arighi (approved):
LGTM, thanks!

PR #1860, opened 2025-05-11
scx_utils: cast ioctl opcodes to libc::Ioctl

vimproved:
On glibc, the ioctl() function takes an unsigned 32-bit integer. However, on other platforms such as musl, it instead takes a signed 32-bit integer. When perf_wrapper.h was being processed through bindgen, it was internally representing the enum variants as u32, which caused issues on musl where ioctl expects an i32. This commit does away with the perf_event_ioctls enum entirely, instead using constant uint16_t values to ensure they can be passed to ioctl on both glibc and musl systems.

### DIFF
--- a/rust/scx_utils/src/perf.rs
+++ b/rust/scx_utils/src/perf.rs
@@ -52,11 +52,11 @@ pub mod ioctls {
 
     #[allow(clippy::missing_safety_doc)]
     pub unsafe fn enable(fd: c_int, arg: c_uint) -> c_int {
-        unsafe { libc::ioctl(fd, perf::bindings::ENABLE.into(), arg) }
+        unsafe { libc::ioctl(fd, perf::bindings::ENABLE as libc::Ioctl, arg) }
     }
 
     #[allow(clippy::missing_safety_doc)]
     pub unsafe fn reset(fd: c_int, arg: c_uint) -> c_int {
-        unsafe { libc::ioctl(fd, perf::bindings::RESET.into(), arg) }
+        unsafe { libc::ioctl(fd, perf::bindings::RESET as libc::Ioctl, arg) }
     }
 }


vimproved:
it seems the value PERF_EVENT_IOC_MODIFY_ATTRIBUTES is `0x4008240b`, so this solution will not work actually. I'll update this PR with one that actually does.

hodgesds:
Can maybe use `#ifdef __MUSL__` in `perf_wrapper.h` and a [macro](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env) on the Rust side I think. Thanks for taking a look!

vimproved:
> Can maybe use `#ifdef __MUSL__` in `perf_wrapper.h` and a [macro](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env) on the Rust side I think. Thanks for taking a look!

musl deliberately does not define a `__MUSL__` macro: https://wiki.musl-libc.org/faq . Would this cast cause any issues on other platforms that would necessitate guarding it like that?

Additionally, while musl is the most high-profile libc that has ioctl accept a signed integer, as per ioctl(2) it is presumably not the only one.

JakeHillion:
Took a look at the docs [here](https://docs.rs/libc/0.2.172/libc/fn.ioctl.html), this seems to take a `c_ulong` in every relevant version. It seems like this cast is going to basically be a no-op. Does this definitely change the behaviour under musl?

vimproved:
> Took a look at the docs [here](https://docs.rs/libc/0.2.172/libc/fn.ioctl.html), this seems to take a `c_ulong` in every relevant version. It seems like this cast is going to basically be a no-op. Does this definitely change the behaviour under musl?

This documentation is seemingly incorrect. If you look at the source, ioctl really takes `libc::Ioctl`, which is defined per-platform. On musl, it's defined to c_int: https://github.com/rust-lang/libc/blob/main/src/unix/linux_like/linux/musl/mod.rs#L33 .

JakeHillion (approved):
Thank you for the explanation, that doc thing is absurd. Confirmed this builds now on musl.

PR #1861, opened 2025-05-12
layered: add tgid tid count matcher

likewhatevs:
Add tgid tid matcher and supporting tgid/tid-count tracking.

This is to be used to enable writing policies that partition spammy PIDs with spammy threadpools (i.e. 10k threadpools) from those with not-spammy threadpools (i.e. threadpools w/ say, 10 threads).

This is imprecise in how it works (i.e. it tracks TIDs counted using a bloom filter), but this is for confining strongly misbehaving processes (i.e. 1k+ threadpools), so this might be fine (and preferable to more accurately tracking counts, tbh).

~I need to debug why counters aren't updating right, but the general idea is add a matcher that will run once per minute and count the number of task structs with the same TGID.~

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -56,6 +56,8 @@ volatile u64 layer_refresh_seq_avgruntime;
 
 /* Flag to enable or disable antistall feature */
 const volatile bool enable_antistall = true;
+/* Flag to enable or disable task count feature */
+const volatile bool enable_task_count = true;
 const volatile bool enable_gpu_support = false;
 /* Delay permitted, in seconds, before antistall activates */
 const volatile u64 antistall_sec = 3;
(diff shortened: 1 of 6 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3078,13 +3092,82 @@ static bool antistall_scan(void)
...
+			if (!counter) {
+				bpf_map_update_elem(&tgid_tid_count, &pidk, &zero_u32, 0);
+			} else {
+				*counter = 0;
+			}
+		} else {
+			if (!counter) {
+				bpf_map_update_elem(&tgid_tid_count, &pidk, &one_u32, 0);
+			} else {
+				(*counter)++;
```
htejun, line 3119: This is racy. What is this count for? is it for stats? Note that the count can be tracked dynamically from task state transitions - `ops.runnable/running/stopping/quiescent()`.

likewhatevs: Will set this there instead of via a timer, thanks!

> What is this count for? 

To be able to have policy place spammy tasks into unpriv layer (i.e. for generic things where CommPrefix doesn't work).

likewhatevs:
This seems to work alright using the example config on PR and `sysbench cpu --threads=2 --time=100 run` w/ threads set to 2, 3 and 4 (i.e. sysbench was correctly assigned to le, ge, and unpriv based on thread count):
Screenshot From 2025-05-16 02-49-44

htejun:
I have trouble understanding what this is trying to achieve. It looks like it's counting the number of threads ever seen per tgid. Why is that meaningful? What properties do a process which keeps forking a short lived thread repeatedly and a process which has a big thread pool share? If it's just trying to count the number of threads a process has, why not `p->signal->nr_threads`?

likewhatevs:
> why not p->signal->nr_threads?

That's a much better way to do this, thanks!

Still works w/ the switch to that:
image

likewhatevs:
> That's a much better way to do this, thanks!

Ahh, actually, I spoke w/ @etsal, this needs to track running/runnable/quescient (and maybe init/exit) instead of using `p->signal->nr_threads` to keep matches accurate (because matching only happens once).

PR #1863, opened 2025-05-12
scx_rustland_core: Fix CPU wakeup with RL_CPU_ANY

arighi:
When specifying RL_CPU_ANY as the target CPU for task dispatch, attempt to wake the task's previously used CPU, instead of trying to wakeup a CPU close to RL_CPU_ANY, which is an invalid CPU ID.

Moreover, always use SCX_KICK_IDLE to prevent unnecessary wakeup events when the target CPU is not idle (which can occur due to concurrent wakeup attempts).

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -567,7 +567,8 @@ static void dispatch_task(const struct dispatched_task_ctx *task)
 	 */
 	if (task->cpu == RL_CPU_ANY) {
 		scx_bpf_dsq_insert_vtime(p, SHARED_DSQ, task->slice_ns, task->vtime, task->flags);
-		goto out_kick_idle_cpu;
+		scx_bpf_kick_cpu(scx_bpf_task_cpu(p), SCX_KICK_IDLE);
+		goto out_release;
 	}
 
 	/* Read current cpumask generation counter */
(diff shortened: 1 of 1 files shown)


rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -577,7 +578,8 @@ static void dispatch_task(const struct dispatched_task_ctx *task)
 	if (!bpf_cpumask_test_cpu(task->cpu, p->cpus_ptr)) {
 		scx_bpf_dsq_insert_vtime(p, SHARED_DSQ, task->slice_ns, task->vtime, task->flags);
 		__sync_fetch_and_add(&nr_bounce_dispatches, 1);
-		goto out_kick_idle_cpu;
+		scx_bpf_kick_cpu(scx_bpf_task_cpu(p), SCX_KICK_IDLE);
```
hodgesds, line 581: nit: can `task->cpu` be used here since it's already being used on the check in the outer `if` statement?

arighi: We could, but the task won't be able to run on task->cpu because it's not in its allowed cpumask, so kicking that CPU isn't really useful

hodgesds: I'm blind sorry about that

PR #1858, opened 2025-05-11
scx_layered: fix verification failure on Clang 20

etsal:
scx_layered is failing verification on Clang 20 for kernels >= 6.14. Verification logs point to the root cause being that the bounds checking for the prefix/suffix matching function in util.bpf.c are being optimized out. This causes subsequent indexing into the prefix array to cause the aforementioned verification failure. Clamp the index into the array within [0, MAX_PATH) instead of bounds checking to avoid this problem.

PR #1868, opened 2025-05-12
scx_layered: prevent offset calculation during prefix matching from being optimized out

etsal:
Despite our best efforts, Clang keeps optimizing out the bounds checking we do in scx_layered for our string indices. Prevent the code introduced in #1858 from being inlined to avoid this.


hodgesds (approved):
LGTM

igs-raiderler:
The documentation in this repository is excellent! Very clear and comprehensive.

PR #1871, opened 2025-05-13
scx_lavd: Make the load balancing core compaction- and capacity-aware.

multics69:
Previously, load balancing assumed that all CPUs were in active sets and had
the same capacity. Also, it aims to equalize the DSQ length of each domain.
However, it does not work well if a system has heterogeneous processors
(big/LITTLE) or is under-loaded (i.e., DSQ lengths are all zero).

Thus, let's make the load-balancing core compaction and CPU capacity aware.
The load balancing now considers the active CPUs and each CPU's capacity.
The load balancing aims for two goals:

1) The *non-scaled* CPU utilizations of all active CPUs should be the same or
   similar. This helps to maintain low latency when the system is underloaded.

2) The *scaled* queue lengths of active compute domains should be the same or
   similar. Using scaled queue length allows putting more tasks to the powerful
   compute domains. This helps to maintain high throughput when the system is
   overloaded.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -7,58 +7,141 @@
 /*
  * To be included to the main.bpf.c
  */
+
+u64 __attribute__ ((noinline)) calc_mig_delta(u64 avg_sc_load, int nz_qlen)
+{
+	/*
+	 * Note that added "noinline" to make the verifier happy.
+	 */
+	if (nz_qlen >= sys_stat.nr_active_cpdoms)
+		return avg_sc_load >> LAVD_CPDOM_MIG_SHIFT_OL;
+	if (nz_qlen == 0)
+		return avg_sc_load >> LAVD_CPDOM_MIG_SHIFT_UL;
+	return avg_sc_load >> LAVD_CPDOM_MIG_SHIFT;
+}
+
 static void plan_x_cpdom_migration(void)
 {
 	struct cpdom_ctx *cpdomc;
 	u64 dsq_id;
-	u32 avg_nr_q_tasks_per_cpu = 0, x_mig_delta;
 	u32 stealer_threshold, stealee_threshold, nr_stealee = 0;
+	u64 avg_sc_load = 0, min_sc_load = U64_MAX, max_sc_load = 0;
+	u64 x_mig_delta, util, qlen, sc_qlen;
+	bool overflow_running = false;
+	int nz_qlen = 0;
 
 	/*
-	 * Calcualte average queued tasks per CPU per compute domain.
+	 * The load balancing aims for two goals:
+	 *
+	 * 1) The *non-scaled* CPU utilizations of all active CPUs should be
+	 * the same or similar. This helps to maintain low latency
+	 * when the system is underloaded.
+	 *
+	 * 2) The *scaled* queue lengths of active compute domains should be
+	 * the same or similar. Using scaled queue length allows putting more
+	 * tasks to the powerful compute domains. This helps to maintain high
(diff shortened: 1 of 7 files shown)


scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -8,6 +8,143 @@
...
+			break;
+
+		cpdomc = MEMBER_VPTR(cpdom_ctxs, [dsq_id]);
+
+		/*
+		 * Under-loaded active domains become a stealer.
+		 */
+		if (cpdomc->nr_active_cpus &&
+		    cpdomc->sc_load <= stealer_threshold) {
+			WRITE_ONCE(cpdomc->is_stealer, true);
```
hodgesds, line 122: I like this abstraction of having the load balancing update the `cpdomc` state, it nicely separates the load balancing logic from the actual load balancing.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -8,6 +8,143 @@
...
+	 * when the system is underloaded.
+	 *
+	 * 2) The *scaled* queue lengths of active compute domains should be
+	 * the same or similar. Using scaled queue length allows putting more
+	 * tasks to the powerful compute domains. This helps to maintain high
+	 * throughput when the system is overloaded.
+	 */
+
+	/*
+	 * Calcualte scaled load for each active compute domain.
```
hodgesds, line 47: nit: Calculate

marioroy:
Building scx with pr commits.
```text
_backports=(
  ef13ab4aab43c5b67b2fd34c16e78edbfd2a1644
  a28de627eeee401e4b4cefb3798c49d36b3fd536
)

==> Starting prepare()...
ef13ab4a scx_lavd: Refactor the load balancing.
a28de627 scx_lavd: Make the load balancing core compaction- and capacity-aware.
```

No improvement for running lesser NCPUs than full. Checking for new LAVD options.
`scx_lavd --performance --slice-max-us=5000 --slice-min-us=500`
The primesieve.pl (using C-API) is more cpu-bound (lesser number of chunks).
```text
$ scx_lavd --version
scx_lavd 1.0.12-g1709af21-dirty x86_64-unknown-linux-gnu

23.501  ./algorithm3.pl 1e12 --threads=20 --procbind
28.656  ./algorithm3.pl 1e12 --threads=20

17.997  ./primesieve.pl 2e12 --threads=20 --procbind
20.346  ./primesieve.pl 2e12 --threads=20
```

Rustland for comparison, without CPU affinity.
```text
$ ./algorithm3.pl 1e12 --threads=20
Primes found: 37607912018
Seconds: 23.839

$ ./primesieve.pl 2e12 --threads=20
Primes found: 73301896139
Seconds: 18.086
```

EEVDF-BORE for comparison. Kernel built with sched/fair: Prefer full-idle SMT cores
https://web.git.kernel.org/pub/scm/linux/kernel/git/arighi/linux.git/commit/?h=sched-smt-idle
```text
$ ./algorithm3.pl 1e12 --threads=20
Primes found: 37607912018
Seconds: 25.395

$ ./primesieve.pl 2e12 --threads=20
Primes found: 73301896139
Seconds: 18.282
```

No new regressions from older LAVD 🙂 Similar FPS and smoothness 🙂

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
multics69: I found marking stealer and stealee and performing load balancing later is pretty handy. Especially, it is easy to implement to gradual load balancing to avoid the thundering herd problem.

PR #1876, opened 2025-05-14
scx_rustland_core: skip L2/L3 search when domain is empty

charliechiou:
Handle the return value of bpf_cpumask_and() when computing L2 and L3 domain masks to avoid propagation of invalid masks and potential verifier rejection or runtime errors.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -433,7 +433,8 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
 	 * CPU (ignore if this cpumask completely overlaps with the task's
 	 * cpumask).
 	 */
-	bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain));
+	if (!bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain)))
+		l2_mask = NULL;
 
 	/*
 	 * Determine the L3 cache domain as the intersection of the task's
@@ -441,7 +442,8 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
 	 * CPU (ignore if this cpumask completely overlaps with the task's
 	 * cpumask).
 	 */
-	bpf_cpumask_and(l3_mask, p->cpus_ptr, cast_mask(l3_domain));
+	if (!bpf_cpumask_and(l3_mask, p->cpus_ptr, cast_mask(l3_domain)))
+		l3_mask = NULL;
 
 	/*
 	 * Find the best idle CPU, prioritizing full idle cores in SMT systems.
@@ -462,19 +464,23 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
 		 * Search for any full-idle CPU in the task domain that shares
 		 * the same L2 cache.
 		 */
-		cpu = bpf_cpumask_any_and_distribute(cast_mask(l2_mask), idle_smtmask);
-		if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
-		    scx_bpf_test_and_clear_cpu_idle(cpu))
-			goto out_put_cpumask;
+		if (l2_mask) {
+			cpu = bpf_cpumask_any_and_distribute(cast_mask(l2_mask), idle_smtmask);
+			if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
+			    scx_bpf_test_and_clear_cpu_idle(cpu))
+				goto out_put_cpumask;
+		}
 
 		/*
 		 * Search for any full-idle CPU in the task domain that shares
 		 * the same L3 cache.
 		 */
-		cpu = bpf_cpumask_any_and_distribute(cast_mask(l3_mask), idle_smtmask);
-		if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
-		    scx_bpf_test_and_clear_cpu_idle(cpu))
-			goto out_put_cpumask;
+		if (l3_mask) {
+			cpu = bpf_cpumask_any_and_distribute(cast_mask(l3_mask), idle_smtmask);
+			if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
+			    scx_bpf_test_and_clear_cpu_idle(cpu))
+				goto out_put_cpumask;
+		}
 
 		/*
 		 * Otherwise, search for another usable full-idle core.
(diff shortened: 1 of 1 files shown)


rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -433,15 +433,23 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
 	 * CPU (ignore if this cpumask completely overlaps with the task's
 	 * cpumask).
 	 */
-	bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain));
+	if (bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain)) < 0) {
+		scx_bpf_error("failed to compute l2_mask");
```
arighi, line 437: This doesn't seem correct, `l2_mask` can be empty and in that case we don't want to error., instead we should ignore picking a CPU in the same L2 cache.

Maybe a better solution could be to set l2_mask to NULL in this case, so that we can skip the L2 step quicker without searching for a bit in the empty cpumask.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -433,15 +433,23 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
...
 
 	/*
 	 * Determine the L3 cache domain as the intersection of the task's
 	 * primary cpumask and the L3 cache domain mask of the previously used
 	 * CPU (ignore if this cpumask completely overlaps with the task's
 	 * cpumask).
 	 */
-	bpf_cpumask_and(l3_mask, p->cpus_ptr, cast_mask(l3_domain));
+	if (bpf_cpumask_and(l3_mask, p->cpus_ptr, cast_mask(l3_domain)) < 0) {
+		scx_bpf_error("failed to compute l3_mask");
```
arighi, line 449: Ditto for the L3 cache.

arighi (changes requested):

rust/scx_rustland_core/assets/bpf/main.bpf.c
charliechiou: Thanks for the suggestion! I’ve added a check using `bpf_cpumask_empty()` to handle the case where `l2_mask` is empty after bpf_cpumask_and(). In such cases, l2_mask is now set to NULL to allow skipping the L2 step.

charliechiou:
Apologies for the multiple pushes. 
I've updated the logic for both `l2_mask` and `l3_mask`: if bpf_cpumask_and() fails (returns < 0), the corresponding mask is now set to NULL instead of raising an error. 

I initially attempted to use `bpf_cpumask_empty()` to detect whether the `l2_mask` and `l3_mask` were empty, but it caused the BPF program to fail to load. Therefore, I opted to directly set the masks to NULL as a fallback.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -433,15 +433,23 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
 	 * CPU (ignore if this cpumask completely overlaps with the task's
 	 * cpumask).
 	 */
-	bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain));
+	if (bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain)) < 0){
+		l2_mask = NULL;
+		cpu = -ENOENT;
+		goto out_put_cpumask;
```
arighi, line 439: Even if l2_mask is empty we still want to try with the l3_mask and all the other stuff below, we don't want to give up immediately.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -433,15 +433,23 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
...
 	 * Determine the L3 cache domain as the intersection of the task's
 	 * primary cpumask and the L3 cache domain mask of the previously used
 	 * CPU (ignore if this cpumask completely overlaps with the task's
 	 * cpumask).
 	 */
-	bpf_cpumask_and(l3_mask, p->cpus_ptr, cast_mask(l3_domain));
+if (bpf_cpumask_and(l3_mask, p->cpus_ptr, cast_mask(l3_domain)) < 0){
+		l3_mask = NULL;
+		cpu = -ENOENT;
+		goto out_put_cpumask;
```
arighi, line 451: Same here.

Something we can do is set the mask to NULL, keep going and explicitly check if the mask != NULL when it's used. Similarly to what we do in scx_bpfland.

arighi (changes requested):

charliechiou:
I removed the early exit and instead check whether `l2_mask` and `l3_mask` are non-NULL where they are used.

arighi:
> I removed the early exit and instead check whether `l2_mask` and `l3_mask` are non-NULL where they are used.

The change looks good now, but the commit message doesn't match what the code is doing. Can you update the commit message?

Basically with this change, we skip searching for an idle CPU in the L2/L3 cache when the task can't use any of the CPUs in that cache. So, instead of searching in an empty cpumask, we just skip the search, which should be a bit faster.

charliechiou:
PR title and commit message have been updated as suggested. Thanks!

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -433,15 +433,19 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
 	 * CPU (ignore if this cpumask completely overlaps with the task's
 	 * cpumask).
 	 */
-	bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain));
+	if (bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain)) < 0) {
+		l2_mask = NULL;
+	}
```
EricccTaiwan, line 438:
```
warning: scx_rlfifo@1.0.12: main.bpf.c:436:66: warning: result of comparison of constant 0 with expression of type 'bool' is always false [-Wtautological-constant-compare]
warning: scx_rlfifo@1.0.12:   436 |         if (bpf_cpumask_and(l2_mask, p->cpus_ptr, cast_mask(l2_domain)) < 0) {
warning: scx_rlfifo@1.0.12:       |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~
warning: scx_rlfifo@1.0.12: main.bpf.c:446:66: warning: result of comparison of constant 0 with expression of type 'bool' is always false [-Wtautological-constant-compare]
warning: scx_rlfifo@1.0.12:   446 |         if (bpf_cpumask_and(l3_mask, p->cpus_ptr, cast_mask(l3_domain)) < 0) {
warning: scx_rlfifo@1.0.12:       |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~
```

Since [`bpf_cpumask_and`](https://elixir.bootlin.com/linux/v6.14.6/source/kernel/bpf/cpumask.c#L251) is type of bool, this `if` statement is always false.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -462,19 +466,23 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
...
 		 * the same L2 cache.
 		 */
-		cpu = bpf_cpumask_any_and_distribute(cast_mask(l2_mask), idle_smtmask);
-		if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
-		    scx_bpf_test_and_clear_cpu_idle(cpu))
-			goto out_put_cpumask;
+		if (l2_mask) {
+			cpu = bpf_cpumask_any_and_distribute(cast_mask(l2_mask), idle_smtmask);
+			if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
+				scx_bpf_test_and_clear_cpu_idle(cpu))
```
arighi, line 472: Can you align this to bpf_cpumask_test_cpu() since it's part of the condition? Otherwise it looks like the first instruction of the if block.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -499,19 +507,23 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu)
...
 	 * L3 cache.
 	 */
-	cpu = bpf_cpumask_any_and_distribute(cast_mask(l3_mask), idle_cpumask);
-	if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
-	    scx_bpf_test_and_clear_cpu_idle(cpu))
-		goto out_put_cpumask;
+	if (l3_mask) {
+		cpu = bpf_cpumask_any_and_distribute(cast_mask(l3_mask), idle_cpumask);
+		if (bpf_cpumask_test_cpu(cpu, online_cpumask) &&
+			scx_bpf_test_and_clear_cpu_idle(cpu))
```
arighi, line 524: And here as well.

arighi (commented):
Left some comments, only minor coding-style fixes. Apart than that,everything looks good. Thanks.

EricccTaiwan (approved):

charliechiou:
Thanks for the approval!

PR #1880, opened 2025-05-14
scx_lavd: Don't print unnecessary CPU preference orders.

multics69:
When starting the scheduler, CPU preference orders for performance and power-save modes were printed. Printing unnecessary information (e.g., the power-save order when the performance option is specified) would confuse the user. So let's print only relevant information.


arighi (approved):
Makes sense, maybe we can print both in verbose mode, but it's up to you (it may still confuse some users).

Issue #1524, opened 2025-03-18
Rust minimum version


anisse:
Hello,

It seems that the rust minimum version is now 1.81:

```
$ meson compile -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /usr/bin/ninja -C /home/an.astier/scx/build
ninja: Entering directory `/home/an.astier/scx/build'
[47/47] Generating rust_all with a custom command (wrapped by meson to set env)
FAILED: Cargo.toml.__PHONY__ 
/home/an.astier/scx/env3.12/bin/meson --internal exe --unpickle /home/an.astier/scx/build/meson-private/meson_exe_cargo_aa9ee77049b90130b859abff6af5250bdf5ccd06.dat
while executing ['/usr/bin/cargo', 'build', '--manifest-path=../Cargo.toml', '--target-dir=.', '--quiet']
--- stdout ---

--- stderr ---
error: rustc 1.79.0 is not supported by the following packages:
  colored@3.0.0 requires rustc 1.80
  home@0.5.11 requires rustc 1.81
  zbus_names@4.1.0 requires rustc 1.80
  zbus_names@4.1.0 requires rustc 1.80
  zvariant@5.1.0 requires rustc 1.80
  zvariant@5.1.0 requires rustc 1.80
  zvariant_derive@5.1.0 requires rustc 1.80
Either upgrade rustc or select compatible dependency versions with
`cargo update <name>@<current-ver> --precise <compatible-ver>`
where `<compatible-ver>` is the latest version supporting rustc 1.79.0


ninja: build stopped: subcommand failed.
```

Shall it be bumped in the README?
https://github.com/sched-ext/scx/blob/59ed5d8cb7570261f8fda401515c2278eed6f043/README.md?plain=1#L127

Issue #1419, opened 2025-02-25
Update README


torvic9:
Hi,

the README lists the kernel config options which have to be enabled to use scx.

There is however one missing option: `CONFIG_KPROBES=y`.

While it is not a hard dependency, some scx schedulers need this option to work (scx_rustland and likely others as well).

So I suggest to include it in the README.

Cheers,

Tor

PR #1881, opened 2025-05-14
doc : require Rust 1.86 and add recommended Kconfig

EricccTaiwan:
- Bump minimum Rust version from 1.82 to [1.86](https://github.com/sched-ext/scx/blob/main/tools/scxctl/Cargo.toml#L15).
- Add recommended Kconfig options for consistency with the [upstream kernel README](https://elixir.bootlin.com/linux/v6.14.6/source/tools/sched_ext/README.md#L49).

close: #1419
close: #1524


README.md
```diff
@@ -124,7 +124,7 @@ repo](https://mesonbuild.com/Quick-guide.html#installation-from-source) and call
 - `clang`: >=16 required, >=17 recommended
 - `libbpf`: >=1.2.2 required, >=1.3 recommended (`RESIZE_ARRAY` support is
   new in 1.3). It's preferred to link statically against the source from the libbpf git repo, which is cloned during setup.
-- Rust toolchain: >=1.82
+- Rust toolchain: >=1.86
```
arighi, line 127: Apparently the reason is:
```
  scxctl@1.0.12 requires rustc 1.86
```

Can we just get rid of this requirement in scxctl (that doesn't really need rustc 1.86)?

cc: @frap129

README.md
```diff
@@ -137,6 +137,14 @@ The kernel has to be built with the following configuration:
...
 - `CONFIG_SCHED_CLASS_EXT=y`
 
+It's also recommended that you also include the following Kconfig options:
+- `CONFIG_BPF_JIT_ALWAYS_ON=y`
+- `CONFIG_BPF_JIT_DEFAULT_ON=y`
+- `CONFIG_KPROBES=y`
+- `CONFIG_PAHOLE_HAS_SPLIT_BTF=y`
+- `CONFIG_PAHOLE_HAS_BTF_TAG=y`
+
+There is a `Kconfig` file in this directory whose contents you can append to your local `.config` file, as long as there are no conflicts with any existing options in the file.
```
arighi, line 147: The configs look good, but I wouldn't mention about the file. Since it's part of the github CI, it might be not available in certain source packages shipped by distros. If we want to add a reference to a kconfig file we can either add a link here, or move the file outside of the `.github` dir and add an explicit reference to that.

arighi (changes requested):

README.md
arighi: I like the idea of moving the configs to a more public location, either `tools/` or even the top dir. I'm not sure about the naming. Maybe something like `sched_ext.config` or just `kernel.config`? Personally I like the `.config` extension, because it's also used in the kernel.

README.md
frap129: Yeah, no reason it needs to be 1.86

EricccTaiwan: Just to confirm, do we still need to declare a minimum Rust version in both the `README` and `scxctl`?

README.md
EricccTaiwan: Also, I noticed that there's no `Kconfig` or config fragment in the upstream kernel under [`linux/tools/sched_ext`](https://elixir.bootlin.com/linux/v6.15-rc6/source/tools/sched_ext).
If adding `sched_ext.config` here is a good idea, I’d be happy to send a patch upstream as well.

EricccTaiwan: > Also, I noticed that there's no `Kconfig` or config fragment in the upstream kernel under [`linux/tools/sched_ext`](https://elixir.bootlin.com/linux/v6.15-rc6/source/tools/sched_ext). If adding `sched_ext.config` here is a good idea, I’d be happy to send a patch upstream as well.

I did find a config file in [linux/tools/testing/selftests/sched_ext/config](https://elixir.bootlin.com/linux/v6.14.6/source/tools/testing/selftests/sched_ext/config), but it differs from the description listed in [linux/tools/sched_ext/README.md](https://elixir.bootlin.com/linux/v6.15-rc6/source/tools/sched_ext/README.md#L49), and also from the config recommendations in the [scx/README.md](https://github.com/sched-ext/scx?tab=readme-ov-file#build--install).

Would it make sense to keep these three config references in sync?

EricccTaiwan:
Hi @arighi, I’ve pushed some updates. When you have a moment, I’d appreciate a review!

README.md
```diff
@@ -137,6 +137,23 @@ The kernel has to be built with the following configuration:
 - `CONFIG_FTRACE=y`
 - `CONFIG_SCHED_CLASS_EXT=y`
 
+It's also recommended that you also include the following Kconfig options:
```
arighi, line 140: Drop one "also".

README.md
```diff
@@ -137,6 +137,23 @@ The kernel has to be built with the following configuration:
...
+- `CONFIG_BPF_JIT_DEFAULT_ON=y`
+- `CONFIG_KPROBES=y`
+- `CONFIG_PAHOLE_HAS_SPLIT_BTF=y`
+- `CONFIG_PAHOLE_HAS_BTF_TAG=y`
+
+A recommended config fragment is available in [`scx/sched_ext.config`](./sched_ext.config).
+It includes all required and recommended options for using `sched_ext`.
+You can append its contents to your kernel `.config` file to enable the necessary features.
+
+To apply it:
```
arighi, line 151: I would drop this "To apply it" section, people that are willing to recompile the kernel likely know how to modify their config.

arighi (commented):
Left a couple of comments, but overall I think it's a good idea to provide a file with the required and recommended kernel configs.

About the naming, what do you think about `kernel.config`? I think it makes it clearer that the file is related to the kernel config.

EricccTaiwan:
> About the naming, what do you think about `kernel.config`? I think it makes it clearer that the file is related to the kernel config.

Agreed, `kernel.config` is much clearer.

JakeHillion:
> LGTM, thanks!
> 
> @JakeHillion @likewhatevs what do you think about including this `kernel.config` in the CI in addition to `.github/include/sched-ext.config`?

Do you mean doubling the matrix? It's not super easy to do that at the minute, but we could do it 12-hourly for sched_ext/for-next for example. Is there a significant change in behaviour here that's worth testing?

likewhatevs:
> @JakeHillion @likewhatevs what do you think about including this kernel.config in the CI in addition to .github/include/sched-ext.config?

I'd say yeah for when we accidentally leverage some capability only enabled via a debug flag not default in fedora/ubuntu etc., defer to @JakeHillion though.

arighi:
> > LGTM, thanks!
> > 
> > @JakeHillion @likewhatevs what do you think about including this `kernel.config` in the CI in addition to `.github/include/sched-ext.config`?
> 
> Do you mean doubling the matrix? It's not super easy to do that at the minute, but we could do it 12-hourly for sched_ext/for-next for example. Is there a significant change in behaviour here that's worth testing?

Oh no, I was just suggesting to merge the two configs, where this new one should be a subset of the old one.

EricccTaiwan:
> Oh no, I was just suggesting to merge the two configs, where this new one should be a subset of the old one.
> Edit: something like this basically (not sure about the nix part), where do you get `./sched-ext.config`?):

I also noticed that there's a more detailed config in `scx/.github/include/sched-ext.config`,
so maybe we can just reference that in the `README` and remove `scx/kernel.config` to avoid duplication?

EricccTaiwan:
> > If we want to add a reference to a kconfig file we can either add a link here, or move the file outside of the .github dir and add an explicit reference to that.
> 
> Or maybe just move `sched-ext.config` outside of `.github` and reference it?

I relocated `sched-ext.config` from `scx/.github/include` to `scx/` and renamed it to `kernel.config`, but it looks like this change caused the [build-and-test / build-kernel](https://github.com/sched-ext/scx/actions/runs/15141818893/job/42567621982?pr=1881) job to fail. 
Does that mean we can't move the file like this? Or should we update something in the CI (which I'm note very familiar with)? 

Thanks.

JakeHillion:
> > > If we want to add a reference to a kconfig file we can either add a link here, or move the file outside of the .github dir and add an explicit reference to that.
> > 
> > 
> > Or maybe just move `sched-ext.config` outside of `.github` and reference it?
> 
> I relocated `sched-ext.config` from `scx/.github/include` to `scx/` and renamed it to `kernel.config`, but it looks like this change caused the [build-and-test / build-kernel](https://github.com/sched-ext/scx/actions/runs/15141818893/job/42567621982?pr=1881) job to fail. Does that mean we can't move the file like this? Or should we update something in the CI (which I'm not very familiar with)?
> 
> Thanks.

You'll need to update this path:

    https://github.com/sched-ext/scx/blob/d247349fc28c8eedd18a25e92cd91796e260e4cd/.github/include/build-kernel.nix#L32

Keep it relative and should work fine. You can test locally with `nix build --no-link ./.github/include#kernels.sched_ext/for-next`, or push and the CI will test for you.

EricccTaiwan:
> > I relocated `sched-ext.config` from `scx/.github/include` to `scx/` and renamed it to `kernel.config`, but it looks like this change caused the [build-and-test / build-kernel](https://github.com/sched-ext/scx/actions/runs/15141818893/job/42567621982?pr=1881) job to fail. Does that mean we can't move the file like this? Or should we update something in the CI (which I'm not very familiar with)?
> > Thanks.
> 
> You'll need to update this path:
> 
> https://github.com/sched-ext/scx/blob/d247349fc28c8eedd18a25e92cd91796e260e4cd/.github/include/build-kernel.nix#L32
> 
> Keep it relative and should work fine. You can test locally with `nix build --no-link ./.github/include#kernels.sched_ext/for-next`, or push and the CI will test for you.

It’s working now. Thanks!
Though I’m wondering if there’s a cleaner or simpler way to write this in Nix?

.github/include/build-kernel.nix
```diff
@@ -29,7 +30,8 @@ let
     nativeBuildInputs = linuxPackages_latest.kernel.nativeBuildInputs;
 
     buildPhase = ''
-      ${virtme-ng}/bin/virtme-ng -v --kconfig --config ${./sched-ext.config}
+      cp ${kernelConfig} ./kernel.config
+      ${virtme-ng}/bin/virtme-ng -v --kconfig --config ./kernel.config
```
JakeHillion, line 34:
```suggestion
      ${virtme-ng}/bin/virtme-ng -v --kconfig --config ${../../kernel.config}
```

.github/include/build-kernel.nix
```diff
@@ -10,6 +10,7 @@
 , commitHash
 , narHash
 , version
+, kernelConfig ? ../../kernel.config
```
JakeHillion, line 13:
```suggestion
```

EricccTaiwan:
> Yeah, the reason we had the `${}` is to encourage some Nix magic. For future reference, we're referencing a file the flake knows about but converting it from the `../../kernel.config` literal into a valid Nix store path at runtime. The `${}` is a string substitution from the Nix variable, which is why it didn't work when you swapped it to a pure string. This is basically what's happening with your `kernelConfig` input, but I think for now we should avoid abstracting it given we only use one config.
> 
> I tested the suggested changes on my machine and they do work.

Thanks for the clarification and suggestion!
However, since I updated this PR, all schedulers have been getting stuck at the `build-kernel` stage for a while and eventually fail. I haven't been able to figure out why.

image

JakeHillion:
> > Yeah, the reason we had the `${}` is to encourage some Nix magic. For future reference, we're referencing a file the flake knows about but converting it from the `../../kernel.config` literal into a valid Nix store path at runtime. The `${}` is a string substitution from the Nix variable, which is why it didn't work when you swapped it to a pure string. This is basically what's happening with your `kernelConfig` input, but I think for now we should avoid abstracting it given we only use one config.
> > I tested the suggested changes on my machine and they do work.
> 
> Thanks for the clarification and suggestion! However, since I updated this PR, all schedulers have been getting stuck at the `build-kernel` stage for a while and eventually fail. I haven't been able to figure out why.
> 
> image

Sorry about that, I changed the kernel build process a little while ago and you got caught out by an edge case. Pushed a commit to fix it for you.

JakeHillion (approved):
Thanks for the changes!

EricccTaiwan:
> Sorry about that, I changed the kernel build process a little while ago and you got caught out by an edge case. Pushed a commit to fix it for you.

No worries at all. I'm just trying to understand how things work. Really APPRECIATE the fix!
Just curious: once this patch is merged, will the CI still take this long to build kernel? Or was the delay mainly due to this edge case?

JakeHillion:
> > Sorry about that, I changed the kernel build process a little while ago and you got caught out by an edge case. Pushed a commit to fix it for you.
> 
> No worries at all. I'm just trying to understand how things work. Really APPRECIATE the fix! Just curious: once this patch is merged, will the CI still take this long to build kernel? Or was the delay mainly due to this edge case?

It'll be fine after the current run. There are a few fun interactions here, as you're interested:

- Your initial builds didn't have our repository secrets because they're on a fork, so couldn't upload the cache entry. This is basically intended behaviour.
- The GitHub merge queue `push` workflow does have our secrets, but hit the second edge case. Our NixOS builder that runs the `build-kernel` step has a local cache, and that got a hit. Because the local cache hit, it didn't upload it to the remote cache as it wasn't freshly built. This is suboptimal behaviour.
- Now the GitHub Runners have built your kernel once with secrets, all future flows will have access to it from the cached entry - fork or not.

Usually any changes to the kernels are generated from a branch on this repo (generated by another action) which always has the secrets, which is why we don't often see this weirdness. The timeouts unblock forked PRs, and I'll put up a PR shortly to clean up the middle case too (where it should have been cached but wasn't).

PR #1866, opened 2025-05-12
scx_bpfland: simplify error handling for sibling CPU setup

EricccTaiwan:
Replaces a match on Result with an is_err() check, since only the Err case was meaningful and the Ok arm was empty.

No functional behavior is changed.

PR #1883, opened 2025-05-14
scx_rustland_core: propagate errors from enable_sibling_cpu_fn

EricccTaiwan:
Previously, errors during sibling CPU setup were silently ignored, making failures hard to debug. Now they are returned with context as `io::Error,` allowing the caller to handle them as needed.

Ref: #1866


arighi (approved):
Looks good, thanks.

PR #1886, opened 2025-05-14
chaos/p2dq: symlink lib to fix publishing

JakeHillion:
scx_chaos currently can't be published. It turns out this is because the p2dq
version that is published is broken - it can't be installed. This is because it
depended on `../../../lib/...` which is valid on the local system and let's it
be published, but isn't valid when someone installs it from the crates.io sources.

Symlink `lib` into `chaos` and `p2dq` so it gets uploaded to crates.io.

Test plan:
- Cargo is really bad at this. CI for the change, publishing is going to be largely
  luck.


JakeHillion:
This commit is based on the v1.0.12 tag to ensure we haven't included any API changes in scx_p2dq or scx_chaos. This is to avoid having to publish scx_utils to fix these versions. When it merges the commit will still exist, so at least it's vaguely reproducible.

hodgesds (approved):

PR #1892, opened 2025-05-15
scx_lavd:  Find an alternative domain beyond the current node/LLC.

multics69:
Each compute domain has two IDs: its unique ID and the alternative compute domain's ID (e.g., LITTLE compute domain's ID for a big compute domain). The alternative compute domain is used to quickly find a compute domain matching with task's performance criticality type.

Previously, an alternative domain was considered only within the same node/LLC domain. However, it does not work in such an architecture where big and LITTLE cores do not share the same LLC.

Hence, choose the closest alternative domain if big and LITTLE cores do not share the same LLC.

### DIFF
--- a/scheds/rust/scx_lavd/src/main.rs
+++ b/scheds/rust/scx_lavd/src/main.rs
@@ -463,6 +463,7 @@ impl FlatTopology {
         // are under the same node and LLC and have the same core type.
         let mut cpdom_id = 0;
         let mut cpdom_map: BTreeMap<ComputeDomainKey, ComputeDomainValue> = BTreeMap::new();
+        let mut cpdom_types: BTreeMap<usize, bool> = BTreeMap::new();
         for cpu_fid in cpu_fids.iter() {
             let key = ComputeDomainKey {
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_lavd/src/main.rs
```diff
@@ -519,6 +510,38 @@ impl FlatTopology {
...
+                    for ncpdom_id in ncpdoms.borrow().iter() {
+                        match cpdom_types.get(ncpdom_id) {
+                            Some(is_big) => {
+                                if *is_big == key.is_big {
+                                    v.cpdom_alt_id.set(*ncpdom_id);
+                                    break 'outer;
+                                }
+                            }
+                            _ => {}
+                        }
```
EricccTaiwan, line 539: Would it be more idiomatic to simplify this using `if let`?
```rust
if let Some(is_big) = cpdom_types.get(ncpdom_id) {
    if *is_big == key.is_big {
        v.cpdom_alt_id.set(*ncpdom_id);
        break 'outer;
    }
}
```

hodgesds (approved):

PR #1894, opened 2025-05-16
scx_rustland: Introduce a congestion threshold

arighi:
If too many tasks are piling up in the user-space scheduler we may risk to hit stall conditions.

To prevent this, introduce a congestion threshold: when the number of waiting tasks exceeds this threshold, the scheduler will proactively flush the queue to bring the task count back below the critical level.

This helps handle heavy stress tests that might flood the system with a high volume of tasks.

### DIFF
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -82,6 +82,16 @@ struct Opts {
     #[clap(short = 'S', long, default_value = "1000")]
     slice_us_min: u64,
 
+    /// Specifies the maximum number of tasks that can be queued in the user-space scheduler.
+    /// If this limit is exceeded, the scheduler will attempt to flush tasks until the count falls
+    /// below the threshold.
+    ///
+    /// Lowering this value reduces the effectiveness of the scheduler under heavy load, while
+    /// raising it can improve scheduling efficiency. However, a higher limit may also reduce
+    /// robustness and increase the risk of stalls when the system is overloaded.
+    #[clap(short = 'w', long, default_value = "100")]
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_rustland/src/main.rs
```diff
@@ -117,6 +117,9 @@ struct Opts {
 // Time constants.
 const NSEC_PER_USEC: u64 = 1_000;
 
+// Congestion threshold.
+const NR_WAITING_MAX: u64 = 128;
```
hodgesds, line 121: Could be interesting to try scaling this by the number of cores.

arighi: > Could be interesting to try scaling this by the number of cores.

I was thinking about that initially, but it's not trivial to model this effectively. In systems that are really big we may allow thousands of tasks to queue up before triggering any flush, leading to burstiness and stuttering behavior.

If tasks are queuing up and the length of the queue keeps growing over a certain threshold, it doesn't matter much if we have 1 CPU or 1000 CPUs, the fact is that the system doesn't have enough capacity to consume the amount of requests, so, in that case, we may want to operate in a more synchronous way, flushing tasks to prevent too long wait time (that may lead to stalls).

BTW, I may also update this PR, I'm not really happy how I've implemented the flush. I'm currently running more tests with a slightly different approach. :)

arighi: Oh yes, also replacing the ring buffers used to bounce tasks from/to BPF with arenas would be interesting. Something that I'm planning to do at once arenas become a bit more stable.

PR #1898, opened 2025-05-16
scx_rustland_core: Refactor memory lock/unlock error handling

EricccTaiwan:
Replaces match statements with if-let bindings, as the OK arms were empty.


arighi (approved):
Small typo in the subject (unloc -> unlock), but apart than that looks good.

PR #1899, opened 2025-05-16
scx_chaos: Refactor thread join error handling

EricccTaiwan:
Replace match statement with an if-let binding, as the Ok arm is empty.


JakeHillion (approved):
thanks!

PR #1900, opened 2025-05-16
ci: switch to nix built kernel

JakeHillion:
Continue addressing the reproducibility issues in the CI by switching to the Nix built kernel. This kernel has been building and working for some time, but we need a specific kernel header for `process_runqlat.bt`. Add a derivation for the kernel headers and build that too.

Use cachix to store the Nix built artifacts. This requires a little bit of setup and was a tad fiddly on the self-hosted runners, but it appears to be working well now. Forks will not be able to upload to our cache but will be able to restore from it, so as long as they're hitting one of our recently locked kernels they'll get hits. In the future we should employ an alternative like magic-nix-cache too if that ever gets revived. Requires adding the cache to our self-hosted runner as it doesn't have root[0].

The kernel build in CI will then be reproducible and accessible to anyone on their local machine. Anyone running the kernel `nix build` command with `--accept-flake-config` will pull the cachix cached version, or alternatively can build it themselves reproducibly.

Test plan:
- CI, twice. Once to build the cache, once to test the restore.

[0] https://github.com/JakeHillion/scx-nixos-deployments/pull/18


hodgesds (approved):
LGTM

PR #1904, opened 2025-05-17
scx_rustland_core: Get rid of the custom ops.update_idle()

arighi:
Move the check for pending tasks directly into ops.dispatch(). In theory, there is no functional change, but in practice we should save a little bit of overhead, avoiding to jump from ops.dispatch() to ops.update_idle() when a CPU is about to go idle and there are still pending tasks for the user-space scheduler.


arighi:
@JakeHillion FYI this should fix the CI issues with kernel 6.12.

hodgesds (approved):

Issue #1893, opened 2025-05-16
scx_lavd detects wrong cpu configuration


tr1xem:
```
May 16 11:13:15 am am cachyos scx_loader[342185]: 05:43:15 am am [INFO] CPU pref order in performance mode: [12, 13, 14, 15, 16, 17, 18, 19, 0, 2, 4, 6, 8, 10, 1, 3, 5, 7, 9, 11]
May 16 11:13:15 am am cachyos scx_loader[342185]: 05:43:15 am am [INFO] CPU pref order in powersave mode: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
```
Using scx_lavd I have i7-13650hx 
According my cpu config 0-12 are performance core aka high power core and powersave use them is it not weird??
while performance mode use 12 - 19 which are low power threads aka efficient cores

I have i7-13650hx 

```
CPU NODE SOCKET CORE L1d:L1i:L2:L3 ONLINE    MAXMHZ   MINMHZ       MHZ
  0    0      0    0 0:0:0:0          yes 2600.0000 800.0000 2600.0000
  1    0      0    0 0:0:0:0          yes 2600.0000 800.0000 2600.0259
  2    0      0    1 4:4:1:0          yes 2600.0000 800.0000 2600.0000
  3    0      0    1 4:4:1:0          yes 2600.0000 800.0000 2600.0000
  4    0      0    2 8:8:2:0          yes 2600.0000 800.0000 2136.8860
  5    0      0    2 8:8:2:0          yes 2600.0000 800.0000 1128.5850
  6    0      0    3 12:12:3:0        yes 2600.0000 800.0000 1599.2610
  7    0      0    3 12:12:3:0        yes 2600.0000 800.0000  800.0000
  8    0      0    4 16:16:4:0        yes 2600.0000 800.0000  948.2410
  9    0      0    4 16:16:4:0        yes 2600.0000 800.0000  800.0000
 10    0      0    5 20:20:5:0        yes 2600.0000 800.0000 2482.2520
 11    0      0    5 20:20:5:0        yes 2600.0000 800.0000  800.0000
 12    0      0    6 24:24:6:0        yes 1900.0000 800.0000 1900.6110
 13    0      0    7 25:25:6:0        yes 1900.0000 800.0000  800.0000
 14    0      0    8 26:26:6:0        yes 1900.0000 800.0000 1899.9611
[...]

PR #1905, opened 2025-05-17
scx_utils: Remove unused assignment warning

EricccTaiwan:
Initialize `raw_capacity` only when needed to avoid warning.

Ref: #1893

Log:
```shell
warning: value assigned to `raw_capacity` is never read
   --> rust/scx_utils/src/topology.rs:606:13
    |
606 |     let mut raw_capacity = 0;
    |             ^^^^^^^^^^^^
    |
    = help: maybe it is overwritten before being read?
    = note: `#[warn(unused_assignments)]` on by default

warning: `scx_utils` (lib) generated 1 warning
```


multics69 (approved):
Good catch!

PR #1906, opened 2025-05-17
docs: clarify QueuedTask time field comments

charliechiou:
This patch updates the comments for `exec_runtime` and `sum_exec_runtime` in the `QueuedTask` struct to better reflect their distinct meanings.

### DIFF
--- a/rust/scx_rustland_core/README.md
+++ b/rust/scx_rustland_core/README.md
@@ -79,8 +79,8 @@ struct QueuedTask {
     pub pid: i32,              // pid that uniquely identifies a task
     pub cpu: i32,              // CPU previously used by the task
     pub flags: u64,            // task's enqueue flags
-    pub exec_runtime: u64,     // Total cpu time in nanoseconds
-    pub sum_exec_runtime: u64, // Total cpu time in nanoseconds
+    pub exec_runtime: u64,     // Total cpu time since last sleep (in ns)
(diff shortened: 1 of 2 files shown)


rust/scx_rustland_core/README.md
```diff
@@ -79,8 +79,8 @@ struct QueuedTask {
     pub pid: i32,              // pid that uniquely identifies a task
     pub cpu: i32,              // CPU previously used by the task
     pub flags: u64,            // task's enqueue flags
-    pub exec_runtime: u64,     // Total cpu time in nanoseconds
-    pub sum_exec_runtime: u64, // Total cpu time in nanoseconds
+    pub exec_runtime: u64,     // CPU time used during the current execution interval (not yet accumulated), in nanoseconds
```
arighi, line 82: I'm currently changing this struct a bit, see https://github.com/sched-ext/scx/tree/rustland-next.

I think a better description for exec_runtime is "Total cpu time since last sleep (in ns)", that's the same description I'm using in rustland-next. If it makes sense to you, can you use mine as well, so we're aligned?

Then we can merge your commit and I'll rebase my patch set.

charliechiou: Sure, I have changed the description for `exec_runtime` and also `sum_exec_runtime`. That makes sense to me.

arighi (approved):

PR #1907, opened 2025-05-17
Cargo: Clean up libbpf-rs version syntax

EricccTaiwan:
Removed redundant `=` from the libbpf-rs version spec in Cargo.toml. This has no effect on behavior but improves readability.


EricccTaiwan:
Hi @etsal  , would you mind taking a look at this PR when you have time? Thanks!

etsal:
Hi Eric, could you explain how the two expressions are equal? The current configs pin libbpf-rs to the specified version, I believe this change would unpin it, no?

EricccTaiwan:
My intention was to clean up the version spec to match the style used elsewhere in the same `Cargo.toml`.
For example:
```toml
plain = "0.2.3"
seccomp = "0.1"
libbpf-rs = "=0.25.0-beta.1"  <-- only one with explicit `=`
```
Since the version is already fully specified, I considered the `=` to be redundant for consistency and readability.
But you're right — I wasn't aware that the leading `=` enforces strict pinning. Thanks for pointing that out!

If the strict pinning is intentional, we can go ahead and close this PR. Thanks.

PR #1908, opened 2025-05-17
scx_utils: gpu topologyu add GPU's VRAM (max) clock frequency data.

devnexen:
So to gives a hint about the memory bandwith to do a better workload placement. Can be useful later if NVML leverages Multi Instances GPU data.


hodgesds (approved):
LGTM

PR #1909, opened 2025-05-17
docs: Add systing to developer guide

hodgesds:
Add `systing` to the developer guide. `systing` can be used to generate detailed perfetto traces for debugging performance and scheduling issues.


arighi (approved):
Awesome! I didn't know about systing, thanks!

PR #1911, opened 2025-05-17
load_balance: Refactor `find_first_candidate`

EricccTaiwan:
Replace redundant `match` expression with a direct call to `next()`, which already returns an `Option`.

Edit: Change title prefix to `load_balance`


htejun (approved):

htejun (approved):

EricccTaiwan:
This PR only affects `scx_rusty` and `scx_wd40`, but it fails again in `scx_rustland`, which it shouldn't.
@htejun, feel free to add it to the merge queue when convenient. Thanks!

PR #1918, opened 2025-05-19
scx_layered: Make exclusive work better

htejun:
Exclusive flag was really leaky - e.g. it wasn't excluding idle CPU picking so it was really easy for other tasks to occupy the preempted sibling. This PR improves exclusive handling so that it's a lot less leaky (the sibling is a lot more likely to stay idle) and exclusive tasks can share siblings when there are no CPUs left.


likewhatevs:
sibling sharing seems a bit iffy but idk i say try it

likewhatevs (approved):

PR #1920, opened 2025-05-20
arenas: update aarch64 compat ifdef

likewhatevs:
Update aarch64 compat ifdef -- the current ifdef doesn't work on all aarch64 machines. This change makes it work on more.


etsal (approved):
Looks good!

PR #1922, opened 2025-05-20
scx_lavd: Improving the performance stability.

multics69:
This PR improves the performance stability of the scheduler under computationally heavy, dynamic workloads, like kernel build.

It includes:
1) More slowly changing the moving average for stability.
2) Counting task runtime only once in latency calculation for simplicity.


marioroy:
Confirming the LAVD regression with 6.15 is resolved. The WebGL fps remain steady and no longer decrease to low 20's running concurrently, `primesieve.pl 2e12`.

arighi (approved):

PR #1923, opened 2025-05-20
rust: Handle return value from `Option::take`

EricccTaiwan:
Use `let _ = self.struct_ops.take();` to explicitly discard the return value and suppress potential warnings. This makes the intent clearer, indicating that the call is made purely for its side effect.

Ref : https://doc.rust-lang.org/core/option/enum.Option.html#method.take


multics69 (approved):

EricccTaiwan:
Unexpected failure in `scx_chaos`, followed by build errors in `scx_layered` and `scx_bpfland`.

multics69 (approved):

PR #1927, opened 2025-05-20
scx_chaos: Add degradation chaos event

hodgesds:


scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -452,6 +463,21 @@ void BPF_STRUCT_OPS(chaos_enqueue, struct task_struct *p __arg_trusted, u64 enq_
 	    enqueue_chaotic(p, enq_flags, taskc))
 		return;
 
+	// NOTE: this may not work for affinitized tasks because p2dq does
+	// direct dispatch in some situations.
+	if (taskc->next_trait == CHAOS_TRAIT_DEGRADATION && degradation_percent > 0) {
```
JakeHillion, line 468: nit: the `> 0` shouldn't be necessary here either as it'll just be `* 100 / 100`.

scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -452,6 +463,21 @@ void BPF_STRUCT_OPS(chaos_enqueue, struct task_struct *p __arg_trusted, u64 enq_
...
+		if (promise.kind == P2DQ_ENQUEUE_PROMISE_FIFO) {
+			promise.fifo.slice_ns = ((100 - degradation_percent) * promise.fifo.slice_ns) / 100;
```
JakeHillion, line 470: could we switch these to `/ 128` so it's a bitshift? Would change the var name to `_frac7` but it's basically a percentage.

JakeHillion (approved):
This looks good to me. I'm a little worried that it'll have little impact when the machine isn't at saturation though - there will be an overhead from more frequent context switches but the task will still run again soon. I'm still happy with the interface for this and we can modify the implementation later on as needed.

PR #1931, opened 2025-05-20
chaos: refactor trait selection

JakeHillion:
Chaos trait selection is currently handwritten for each trait. Refactor this to have a uniform selection process based on the `chaos_trait` value. This makes adding types easier in the future.

This also implies that CHAOS_TRAIT_NONE is the most unlikely, and this is now explicitly selected as the first random value. Also uses an unrolled loop in the hope that the `const volatile` can be JIT'ted nicely, but this isn't overly important.

Test plan:
- CI
- Manually, tested a few and read the debug output which looks good.

### DIFF
--- a/scheds/rust/scx_chaos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_chaos/src/bpf/main.bpf.c
@@ -39,11 +39,11 @@ const volatile u64 chaos_timer_check_queues_min_ns = 500000;
 const volatile u64 chaos_timer_check_queues_max_ns = 2000000;
 const volatile u64 chaos_timer_check_queues_slack_ns = 2500000;
 
-const volatile u32 random_delays_freq_frac32 = 1; /* for veristat */
+const volatile u32 trait_delay_freq_frac32[CHAOS_TRAIT_MAX];
+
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -79,13 +79,16 @@ static __always_inline enum chaos_trait_kind choose_chaos(struct chaos_task_ctx
...
 
-	if (bpf_get_prandom_u32() < random_delays_freq_frac32)
-		return CHAOS_TRAIT_RANDOM_DELAYS;
+	u32 roll = bpf_get_prandom_u32();
 
-	if (bpf_get_prandom_u32() < cpu_freq_frac32)
-		return CHAOS_TRAIT_CPU_FREQ;
+	#pragma unroll
+	for (int i = 0; i < CHAOS_TRAIT_MAX; ++i) {
+		if (roll <= trait_delay_freq_frac32[i])
```
hodgesds, line 86: Long term it may be interesting to think about enabling multiple chaos traits at a time. If multiple traits have high frequencies it will be biased towards the trait that comes earlier in the iteration.

hodgesds (approved):

PR #1932, opened 2025-05-20
scx_mitosis: Strip down for static cpu pinning

dschatzberg:
This more or less completely reworks mitosis to allocate cells based on static CPU pinning. If it sees a cgroup with a cpuset configured, it will dynamically allocate a new cell and otherwise schedule as before. Eventually I expect to add back in the dynamic cell creation from userspace, but for now it's not needed.


likewhatevs:
you should enable tests for this btw imo

likewhatevs (approved):
CI failure looks unrelated

PR #1936, opened 2025-05-20
scx_p2dq: Call print_topology after attaching scheduler

EricccTaiwan:
Call the topology print test after attaching to verify arena setup and suppress dead_code warning.

<details>
<summary> Edit : Warning log</summary>

```shell
warning: method `print_topology` is never used
   --> scheds/rust/scx_p2dq/src/main.rs:266:8
    |
95  | impl<'a> Scheduler<'a> {
    | ---------------------- method in this implementation
...
266 |     fn print_topology(&mut self) -> Result<()> {
    |        ^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: `scx_p2dq` (bin "scx_p2dq") generated 1 warning
```

### DIFF
--- a/scheds/rust/scx_p2dq/src/main.rs
+++ b/scheds/rust/scx_p2dq/src/main.rs
@@ -88,6 +88,7 @@ struct CliOpts {
 struct Scheduler<'a> {
     skel: BpfSkel<'a>,
     struct_ops: Option<libbpf_rs::Link>,
+    verbose: u8,
 
     stats_server: StatsServer<(), Metrics>,
 }
@@ -124,6 +125,7 @@ impl<'a> Scheduler<'a> {
         Ok(Self {
             skel,
             struct_ops: None,
+            verbose,
             stats_server,
         })
     }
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/main.rs
```diff
@@ -285,6 +285,8 @@ impl<'a> Scheduler<'a> {
 
         self.struct_ops = Some(scx_ops_attach!(self.skel, p2dq)?);
 
+        self.print_topology()?;
```
hodgesds, line 288: Hmmm, maybe it's better to check the value of `opts.verbose` and print only if > 1?

EricccTaiwan: Agreed, it’s a bit annoying for regular users. I’ll work on it later.

EricccTaiwan:
The CI failure is unrelated to `scx_p2dq`.

PR #1938, opened 2025-05-20
arena: compat fix

likewhatevs:
This improves compatibility by avoiding conflicting definitions of mask_size (i.e. when pulled in via p2dq).


etsal:
I think we should keep the variable in this file, because otherwise other schedulers pulling the library in the future would be getting an error until they define the variable. Right now p2dq has a version of the variable because we did not factor out the test_run() based API we use to initialize the arenas, but that was only for the sake of expediency.

PR #1943, opened 2025-05-21
ci: fix CACHIX_AUTH_TOKEN access in integration-tests

JakeHillion:
Reusable workflows can't read secrets by default. This meant that jobs under `integration-test.yml` couldn't upload artifacts to the Nix cache, causing some cache misses. As we own and trust this workflow, inherit all secrets.

Do a little cleanup in `install-nix` to make the use of "self-hosted" uniform.

Test plan:
- CI
- Previously "Post Install Nix" reported that it couldn't upload artifacts due to a missing key. Now it works and veristat isn't always a cache miss.

### DIFF
--- a/.github/actions/install-nix/action.yml
+++ b/.github/actions/install-nix/action.yml
@@ -14,14 +14,14 @@ runs:
         nix_path: nixpkgs=channel:nixos-unstable
 
     - name: Load dependencies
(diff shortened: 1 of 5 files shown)


hodgesds (approved):
LGTM

PR #1944, opened 2025-05-21
scx_rustland_core: Minor optimizations

arighi:
Some small optimizations, cleanups and refactoring aimed at making the code more understandable, especially the BPF -> user-space scheduler activation workflow.

In particular, the user-space scheduler is now dispatched using a separate DSQ (reducing the overhead a bit on the shared DSQ),  SCX_LOCAL_DSQ / SCX_LOCAL_DSQ_ON are used when possible to reduce the overhead on the per-CPU shared DSQs and per-CPU kthreads are not allowed to preempt other tasks (which is overkill and it only introduced unnecessary context switches).

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -872,7 +872,7 @@ void BPF_STRUCT_OPS(rustland_dispatch, s32 cpu, struct task_struct *prev)
 	 * Consume all tasks from the @dispatched list and immediately dispatch
 	 * them on the target CPU decided by the user-space scheduler.
 	 */
-	bpf_user_ringbuf_drain(&dispatched, handle_dispatched_task, NULL, 0);
+	bpf_user_ringbuf_drain(&dispatched, handle_dispatched_task, NULL, BPF_RB_NO_WAKEUP);
 
        /*
 	* Always dispatch the user-space scheduler every time that a CPU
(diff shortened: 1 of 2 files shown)


rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -807,35 +832,26 @@ void BPF_STRUCT_OPS(rustland_enqueue, struct task_struct *p, u64 enq_flags)
...
 	}
 
 	/*
-	 * Always append the user-space scheduler at the end of the shared
-	 * DSQ, so that it'll run after all the tasks currently dispatched
-	 * have used their assigned time slice on their target CPU.
-	 *
-	 * At the same time assign an infinite time slice, so that it can
-	 * completely drain all the pending tasks.
+	 * Assign an infinite time slice to the user-space scheduler, so
```
hodgesds, line 846: I don't quite understand this bit, but in the future would it make sense to allow the userspace side to set the slice as well?

arighi: So, `BpfScheduler` provides a method called `notify_complete()`, which allows the user-space scheduler to signal the BPF side that it has finished its work and is voluntarily yielding the CPU. This mechanism effectively lets the user-space scheduler define its own time slice, not through time-based preemption, but via a voluntary preemption model.

The core idea behind the voluntary `notify_complete()` is to let the scheduler define discrete "bursts" of task dispatching in a deterministic way. This allows to dispatch one task at a time, a batch of N tasks, etc., with the guarantee that the dispatching process will complete without interruption.

But we could definitely extend this by offering an API to configure also a time-based time slice for the user-space scheduler. In that case, we would assign the specified time slice here instead of the infinite time slice. So, yes, looks like a nice thing to add, thanks!

hodgesds: Ahh that makes sense, thanks for the explaination!

PR #1945, opened 2025-05-21
scxtop: Update TUI defaults to slice-consumed

yaakov-stein:
Originally, the default CPU, LLC, and NUMA views showed `dsq_vtime_delta` pane (bottom left). Ex.:
Screenshot 2025-05-20 at 5 55 59 PM

However, the `dsq_slice_consumed` is more useful, so the default has been updated to show that instead (bottom left). Ex.:
Screenshot 2025-05-21 at 10 12 52 AM

Once the default `dsq_vtime_delta` was no longer the default, we can just use the generic `render_scheduler` method to display the `dsq_vtime_delta` when needed. `dsq_vtime_delta` can still be found by using the `s` keybinding.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -899,7 +899,7 @@ impl<'a> App<'a> {
         }
 
         self.render_scheduler("dsq_lat_us", frame, top_left, true, true)?;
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
LGTM, thanks for cleaning up the redundant code

JakeHillion (approved):
Nice!

PR #1948, opened 2025-05-21
scx_layered: Flip the return polarity of sib_keep_idle()

htejun:
It was returning false when the CPU should go idle. Flip it.


etsal (approved):

etsal (commented):
This should be a no-op, right? The only non-flipped return value should be an error condition or only kick in for new tasks AFAICT.

htejun:
Yeah, no behavior changes expected.

PR #1949, opened 2025-05-21
scx_rustland: Read page faults via procfs crate

EricccTaiwan:
Replace manual /proc parsing with procfs for safety and clarity.


scheds/rust/scx_rustland/Cargo.toml
```diff
@@ -16,6 +16,7 @@ libbpf-rs = "=0.25.0-beta.1"
 libc = "0.2.137"
 log = "0.4.17"
 ordered-float = "3.4.0"
+procfs = "0.17"
```
hodgesds, line 19: Why add more dependencies when the [`fb_procfs`](https://docs.rs/fb_procfs/latest/fb_procfs/struct.PidStat.html) crate already provides the same functionality?

EricccTaiwan: Ahh, I didn't realize that. I'll update this PR accordingly.

arighi (commented):
I'm a bit conflicted about this....

On one hand, I like to use something more standard like procfs to get these metrics. On the other hand I'm not sure how much overhead it introduces (it seem to parse every single attribute from /proc/pid/stat), if this was done in a separate thread that would be totally fine, but in this case it's in the main loop, so the extra overhead can potentially affect the scheduler's performance.

However, it's executed only when we're collecting/printing the stats, so if we don't print anything we should be good.

Do you notice any performance regression running for example `scx_rustland --stats 0.1` with vs without this patch?

EricccTaiwan:
> On the other hand I'm not sure how much overhead it introduces (it seem to parse every single attribute from /proc/pid/stat), if this was done in a separate thread that would be totally fine, but in this case it's in the main loop, so the extra overhead can potentially affect the scheduler's performance.

Totally got it — I didn’t consider the overhead at first.

> Do you notice any performance regression running for example `scx_rustland --stats 0.1` with vs without this patch?

I’ll try running some experiments to observe the impact of using procfs. 
Would running something like `scx_rustland --stats 0.1` alongside a workload generator such as `schbench` be a good way to evaluate it?

hodgesds:
For small userspace components a regular rust benchmark would probably be fine as well.

EricccTaiwan:
> For small userspace components a regular rust benchmark would probably be fine as well.

I’ve mostly used `schbench `and `perf` for performance testing. Would you mind sharing which Rust benchmarking tool you had in mind? 
Thanks.

EricccTaiwan:
> If we can use that instead of procfs it'd be nicer, so we don't add another dependency. In terms of perf overhead I'm sure it's the same.

Hi @arighi , @hodgesds ,
I tried to switch to `fb_procfs` so we can drop the extra `procfs` dependency, but I’m hitting a runtime failure I can't nail down.

Here's my code:
```rust
 // Get total page faults from /proc/self/stat using fb_procfs.
  fn get_page_faults() -> Result<u64, io::Error> {
      let pid = std::process::id();             
      info!("Get page faults for PID {}", pid);
      let reader = ProcReader::new();      // <--- Trigger EPERM
      info!("get reader");
      let stat = reader.read_pid_stat(pid).map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
      info!("get stat");
      Ok(stat.minflt.unwrap_or(0) + stat.majflt.unwrap_or(0))
  }
```

<details> <summary>error log</summary>

```shell
➜  ~ sudo ./bin/scx_rustland --stats 0.1                 
16:09:41 [INFO] RustLand version 1.0.12-g69aaab93-dirty x86_64-unknown-linux-gnu/debug - scx_rustland_core 2.3.1
16:09:42 [INFO] Get page faults for PID 465731

thread 'main' panicked at /home/eric/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/threadpool-1.8.1/src/lib.rs:777:10:
called `Result::unwrap()` on an `Err` value: Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }
16:09:42 [INFO] Unregister RustLand scheduler
➜  ~ 
```
</details>

Any hint would be appreciated. Thanks!

arighi:
@EricccTaiwan Oh I know what that is, somehow fb_procfs is using mmap() directy, which is bad in the main thread of the user-space scheduler, because it can trigger stalls.  You can do that in a separate thread, that can be blocked, but not in the main one.

So, I added a seccomp rule to return EPERM when mmap() is called, in this way we can catch hidden mmap() invocations in advance, instead of triggering stalls hard to debug at runtime.

Take a look here:
https://github.com/sched-ext/scx/blob/c23fc8f8357261d10dc02d374f8c3a35904d0ada/rust/scx_rustland_core/src/alloc.rs#L462

And:
https://github.com/sched-ext/scx/blob/c23fc8f8357261d10dc02d374f8c3a35904d0ada/rust/scx_rustland_core/assets/bpf.rs#L282

If fb_procfs is doing crazy things with mmap() it's a big no and we should definitely replace it with procfs.

hodgesds:
Sorry about suggesting that, yeah it makes sense that parsing procfs can easily allocate underneath.

arighi:
> Sorry about suggesting that, yeah it makes sense that parsing procfs can easily allocate underneath.

No worries, it actually turned out to be a useful test case for the seccomp rule!

EricccTaiwan:
Seperated the changes into two commits and included the benchmark reults.

arighi (approved):
Good job, thanks!

PR #1951, opened 2025-05-21
scx_layered: add templates for layer generation 

etsal:
Add a template field to scx_layered configurations for automatically generating layers at initialization time. We must currently hardcode all layers in the config before starting the scheduler. However, this makes it difficult to handle scenarios where we do not already know how exactly what matching we should add to the config. An example would be wanting to create a layer for each container running in a system, without knowing the exact cgroup names when writing the config.

With templates, we can write an approximate match and have it expand to a set of matches at initialization time, depending on the state of the running system. For each match, we then create a layer by copying the original layer's config and appending the generated rule.


etsal:
Updated version of #1937, more robust because it generates the new matches at config parsing time.

likewhatevs (approved):
seems reasonable

htejun:
Maybe add some documentation?

PR #1952, opened 2025-05-21
scx_layered: Fix race condition between picking an idle CPU and the CPU going idle

htejun:
select_cpu/enqueue() looks for an idle CPU and if there is none, the task is queued to the layer's DSQ. However, a CPU can go idle between select_cpu/enqueue() checking idle cpumaks and the task being queued on DSQ:
```
          CPU A                      CPU B
  -------------------------------------------------
     Waking task P
     Look for idle CPU
     No CPU idle
                                  Going idle
                                  Set idle bit
     Enqueue P on layer DSQ
                                  Enters idle
```
This leads to task P waiting in a DSQ while CPU B is sitting idle. This patch updates enqueue() and update_idle() so that:

- After enqueue() puts a task in a DSQ, it looks for idle CPUs one more time and if any is idle, kick it.

- update_idle() scans eligible DSQs. If any is occupied, kicks itself.

Combined with kernel update to update built-in idle mask before calling ops.update_idle(), this guarantees that either enqueue() sees idle bit or update_idle() sees occupied DSQ.

This behavior is pronounced when the machine is near or exactly at saturation - the above races can only happen and lead to scheduling bubbles when there aren't enough idle CPUs but the DSQs are also frequently empty. The following is `stress-ng -c 16` running on a 16 CPU qemu VM where the race seems more pronounced.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1263,18 +1263,27 @@ static void task_uncharge_qrt(struct task_ctx *taskc)
 	taskc->qrt_llc_id = MAX_LLCS;
 }
 
-static void layer_kick_idle_cpu(struct layer *layer)
+static void kick_idle_cpu(struct task_struct *p, struct layer *layer)
 {
-	const struct cpumask *layer_cpumask, *idle_smtmask;;
+	const struct cpumask *idle_smtmask;
 	s32 cpu;
 
-	if (!(layer_cpumask = lookup_layer_cpumask(layer->id)) ||
-	    !(idle_smtmask = scx_bpf_get_idle_smtmask()))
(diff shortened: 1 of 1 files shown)


likewhatevs (approved):
nice

arighi (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2798,12 +2812,29 @@ void BPF_STRUCT_OPS(layered_set_cpumask, struct task_struct *p,
...
+	unsigned layer_id;
+	u32 llc_id;
 
 	if (!idle || !(cpuc = lookup_cpu_ctx(cpu)))
 		return;
 
 	cpuc->protect_owned = false;
 	cpuc->usage_at_idle = cpuc->usage;
+
+	if (!idle)
```
hodgesds, line 2824: nit: I think idle is already checked at 2818

PR #1953, opened 2025-05-22
scx_rustland_core:  Expose nr_cpus_allowed to user-space

arighi:
Drop the internal cpumask generation counter from the task context shared with user-space (it's just easier to check the validity of the target CPU using `p->cpus_ptr`) and expose `p->nr_cpus_allowed` to the user-space scheduler, which can use this information to implement better scheduling policies.

After these changes the size of `QueuedTask / queued_task_ctx` is still 64 bytes, that still fits in a cacheline (typically) and it corresponds to the minimal allocation size in the custom user-space memory allocator, and the size of `DispatchedTask / dispatched_task_ctx` is reduced from 40 bytes to 32 bytes.


hodgesds (approved):
LGTM

PR #1954, opened 2025-05-22
p2dq/chaos: add trait for skeletons that can have their arenas setup

JakeHillion:
c5cb9b4 recently started using the arena to store data for the topology in scx_p2dq. This wasn't fully implemented for chaos so chaos lost visibility of the topology abstraction, causing certain failures. Implement a trait on the p2dq/chaos skeletons to expose this functionality, and call it in the shared `init_skel` macro rather than in each scheduler. The trait approach was chosen as a compromise between code duplication and being stuck in macros for this sort of work.

Test plan:
- CI


hodgesds (approved):
LGTM

PR #1957, opened 2025-05-22
chaos: implement pid targeting

JakeHillion:
Implement PID targeting for scx_chaos. Turns out this was easier than expected, I simply poll `kill(pid, 0)` to check when it drops.

Test plan:
- Pointed it at a shell instance. That one fell apart, everything else worked (testing the combo with ppid targeting).
- Exiting the shell/program under test successfully tells the scheduler to exit.

### DIFF
--- a/scheds/rust/scx_chaos/src/main.rs
+++ b/scheds/rust/scx_chaos/src/main.rs
@@ -9,7 +9,7 @@ use scx_chaos::Trait;
 
 use scx_p2dq::SchedulerOpts as P2dqOpts;
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use clap::Parser;
 use log::info;
 use nix::unistd::Pid;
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_chaos/src/main.rs
```diff
@@ -241,6 +237,30 @@ fn main() -> Result<()> {
         }
     });
 
+
+    if let Some(pid) = args.pid {
+        info!("Monitoring process with PID: {}", pid);
+
+        let is_process_running = |pid: libc::pid_t| -> bool {
+            unsafe {
+                // SAFETY: kill with signal 0 only runs validity checks. There's noc hance of
```
hodgesds, line 246: s/noc hance/no chance

hodgesds: Why kill vs using something like `waitpid`?

JakeHillion: > Why kill vs using something like `waitpid`?

By memory you can only `waitpid` a child and afaik Linux doesn't have a call to do anything smarter for non-child processes.

PR #1959, opened 2025-05-22
scx_p2dq: Cleanup pick_idle_cpu

hodgesds:
Rename cur_llcx to waker_llcx which is a more accurate description of what the variable represents. Move the fast path for checking the previous cpu to the start of pick_idle_cpu. This should improve throughput during mostly idle situations.


multics69 (approved):
LGTM.

arighi (approved):

PR #1960, opened 2025-05-22
scx_rustland_core: Disable THP for user-space scheduler

arighi:
Unmapping/remapping pages in the user-space scheduler's address space can lead to stall conditions: completing the operation may requires a kthread to run, but the kthread cannot run because the user-space scheduler is blocked => deadlock.

To prevent this stall condition, disable transparent huge pages for the user-space scheduler task. This ensures that khugepaged won't touch those pages, thereby avoiding the stall.

### DIFF
--- a/rust/scx_rustland_core/src/alloc.rs
+++ b/rust/scx_rustland_core/src/alloc.rs
@@ -497,7 +497,13 @@ impl UserAllocator {
             if res != 0 {
                 panic!("mlockall failed with error code: {}", res);
             }
-        };
+
+            // Prevent khugpaged from unmapping pages underneath.
+            let res = libc::prctl(libc::PR_SET_THP_DISABLE, 1, 0, 0, 0);
+            if res != 0 {
+                panic!("Failed to disable THP: {}", res);
+            }
+        }
     }
 
     #[allow(static_mut_refs)]


hodgesds (approved):

hodgesds:
I tried to look if `userfaultfd` would help but it still has the problem that the allocing thread still gets put to sleep. I think this works, it's just tricky to find all the edge cases.

arighi:
> I tried to look if `userfaultfd` would help but it still has the problem that the allocing thread still gets put to sleep. I think this works, it's just tricky to find all the edge cases.

I've been playing a bit with userfaultfd in the past, but it wasn't quite working for some reasons and I didn't investigate further. Maybe I can resume this idea and try again now. In theory everything should work if we run the page fault thread as SCHED_FIFO, in this way the fault thread can run even if the user-space scheduler is blocked, complete the mapping and unblock the user-space sched.

And even with userfaultfd we still need to disable THP, because a page remap from khugepaged won't trigger the userfaultfd thread, it still needs to be completed by khugepaged (and potentiall other kthreads are also involved), so maybe that was the reason why my initial implementation wasn't reliably working...

arighi:
@hodgesds now I remember why userfaultfd wasn't working... it's because it's designed to operate only on a specific memory area allocated via mmap(), you can't catch *all* page faults. For instance, it doesn't work with the stack, .text, .data, .bss. So, it could be used for the internal custom allocator arena, but at this point it's just easier (and more efficient) to simply preallocate the whole memory area and mlock() everything.

Unfortunately mlock() is not as powerful as we think... The kernel can still take the initiative to unmap pages (e.g., khugepaged, kcompactd, probably something else...) or force random tasks to reclaim memory, forcing them to flush dirty pages, etc. and if the user-space scheduler is selected in the process, we can potentially trigger a stall.

So far I haven't been able to find a generic way to prevent all these scenarios, so it seems that addressing each one of them separately is the only way for now...

PR #1961, opened 2025-05-22
scxtop: aggregate GLOBAL and LOCAL DSQs

yaakov-stein:
scxtop displays the data for each local DSQ, which can become unusable given the large amount of local DSQs. Ex.:

Screenshot 2025-05-22 at 4 38 28 PM

In order, to make it more useful, we'll now group GLOBAL, LOCAL, and LOCAL_ON DSQs by using their respective [bitmasks](https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/sched/ext.h#n47). Ex.:

Screenshot 2025-05-22 at 4 42 11 PM

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -2331,7 +2331,10 @@ impl<'a> App<'a> {
             self.max_cpu_events,
         ));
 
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
LGTM!

PR #1963, opened 2025-05-22
scx_p2dq: Improve accuracy of runtime accounting

hodgesds:
When tasks stop in p2dq only update the DSQ if the task is no longer runnable. This improves accounting of task utilization by not penalizing tasks that run for different intervals.

Suggested-by: Emil Tsalapatis <aimilios.tsalapatis@gmail.com>

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -909,7 +909,12 @@ static __always_inline int p2dq_running_impl(struct task_struct *p)
 	if (freq_control && taskc->dsq_index == nr_dsqs_per_llc-1) {
 		scx_bpf_cpuperf_set(task_cpu, SCX_CPUPERF_ONE);
 	}
-	taskc->last_run_at = bpf_ktime_get_ns();
+
+	u64 now = bpf_ktime_get_ns();
+	if (taskc->last_run_started == 0)
+		taskc->last_run_started = now;
+
+	taskc->last_run_at = now;
 
 	return 0;
 }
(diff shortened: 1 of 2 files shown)


multics69 (approved):
LGTM

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -909,7 +909,12 @@ static __always_inline int p2dq_running_impl(struct task_struct *p)
 	if (freq_control && taskc->dsq_index == nr_dsqs_per_llc-1) {
 		scx_bpf_cpuperf_set(task_cpu, SCX_CPUPERF_ONE);
 	}
-	taskc->last_run_at = bpf_ktime_get_ns();
+
+	u64 now = bpf_ktime_get_ns();
+	if (taskc->last_run_started == 0)
```
arighi, line 914: I'm wondering, can you just set last_run_started in ops.runnable() instead of doing this check in ops.running()?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -950,31 +955,35 @@ void BPF_STRUCT_OPS(p2dq_stopping, struct task_struct *p, bool runnable)
...
-			stat_inc(P2DQ_STAT_DSQ_SAME);
-		}
-	// If under half the slice was consumed move the task back down.
-	} else if (used < last_dsq_slice_ns / 2) {
-		if (taskc->dsq_index > 0) {
-			taskc->dsq_index -= 1;
-			stat_inc(P2DQ_STAT_DSQ_CHANGE);
-			trace("%s[%p]: DSQ dec %llu -> %u", p->comm, p,
-			      taskc->last_dsq_index, taskc->dsq_index);
+	if (!runnable) {
```
arighi, line 958: And move this logic to ops.quiescent()? In this case it should be equivalent IIUC, but maybe moving the logic to separate callbacks make the code easier to follow?

arighi (approved):
I left a couple of comments, they can be addressed also in a separate PR. This one LGTM.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: I tried that as well, it didn't seem to make much of a difference either way.

PR #1969, opened 2025-05-23
scx_rustland_core: Add support to scx_bpf_select_cpu_and()

arighi:
Newer kernels introduce the scx_bpf_select_cpu_and() API, enabling the built-in idle CPU selection policy to be used consistently across all contexts, rather than being limited to just ops.select_cpu() [1][2].

Implement a compatible wrapper for scx_bpf_select_cpu_and() in scx_rustland_core: kernels that support this kfunc will use it, older kernels will fallback to the custom idle CPU selection policy.

Moreover, refactor the custom idle CPU selection policy, re-aligning with scx_bpfland, and make sure to use the same policy consistently from all contexts, instead of using the built-in policy from ops.select_cpu() and the custom policy from all the other contexts.

Once scx_bpf_select_cpu_and() is available in all the supported kernels, the compatibility wrapper and the custom policy can be removed and we can rely completely on the core sched_ext idle CPU selection policy.

[1] https://lore.kernel.org/all/20250405134041.13778-1-arighi@nvidia.com/
[2] https://lore.kernel.org/all/20250515191716.327518-1-arighi@nvidia.com/


arighi:
Ignore this one for now, it's triggering failures and people from the CachyOS community are reporting regressions. I'm investigating.

hodgesds (approved):
LGTM

arighi:
Bah... CI failure is unrelated to this change I'll address that in a separate PR.

arighi:
oof, what a pain... looks like these CI failures are all caused by memory pressure conditions. How do we convince kcompactd & friends to not touch scx_rustland when the system is short on memory?

arighi:
Alright, I give up. I don't understand why this change makes the CI unhappy. I'll just temporarily drop this and move on with other changes.

multics69:
> Alright, I give up. I don't understand why this change makes the CI unhappy. I'll just temporarily drop this and move on with other changes.

Wow, it is quite surprising. The changes seem quite sane. Hmm...

PR #1971, opened 2025-05-23
scxtop: add search module with fuzzy matching

yaakov-stein:
In order to enable filtering perf events in scxtop, we'll need fuzzy matching functionality. This module will provide the initial fuzzy search. At the moment, it's relatively naive but it will fuzzy match on missed characters and substrings (I'll probably add some manual levenshtein calculations as well to enable it to catch typos and out of order characters). Here are some current examples of what it will match on and what it won't:

Substring - "gettid" == "syscalls:sys_exit_gettid"
Missing - "ssall" == "syscalls:sys_exit_gettid"
Capitalization + substring + missing - "alrMcacEL" == "alarmtimer:alarmtimer_cancel"
Typo - "albrm" != "alarmtimer:alarmtimer_cancel"

The fuzzy matching is largely done using the [fuzzy-matching](https://docs.rs/fuzzy-matcher/latest/fuzzy_matcher/index.html) crate, which is based on [skim](https://github.com/skim-rs/skim).

Here are some performance benchmarks, comparing fuzzy_search to a substring_search:

Screenshot 2025-05-23 at 5 53 13 PM

Given that the median is consistently less than 350 microseconds, we should be fine performance-wise. Benchmarking was done using [Criterion](https://docs.rs/criterion/latest/criterion/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -73,6 +73,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "anstream"
 version = "0.6.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


yaakov-stein:
Going to update with a new fuzzy matcher, will post a new PR next week.

PR #1972, opened 2025-05-23
scx_layered: associate each layer with a cpumask

etsal:
Associate each layer in the scheduler with the cpuset of the tasks running in the layer. This will be used in the future to improve support for containers that have associated cpusets.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -351,6 +351,7 @@ struct layer {
 	char			name[MAX_LAYER_NAME];
 	bool			is_protected;
 	bool			periodically_refresh;
+	u8			cpuset[MAX_CPUS_U8];
 };
 
 struct scx_cmd {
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -351,6 +351,7 @@ struct layer {
 	char			name[MAX_LAYER_NAME];
 	bool			is_protected;
 	bool			periodically_refresh;
+	u8			cpuset[MAX_CPUS_U8];
```
htejun, line 354: While it's following the existing usage, it'd probably be better to switch to u64 for bpf masks. All registers are 64bits anyway, so using u8 would just add unnecessasry masking operations.

htejun (approved):

likewhatevs:
Misc thing for later, this should probably error if the last layer has a cpuset on it.

etsal:
> Misc thing for later, this should probably error if the last layer has a cpuset on it.
Makes sense, though if we want to always have a default open layer maybe we shouldn't have to always specify it and hardcode it somehow.

PR #1973, opened 2025-05-24
scx_p2dq: Refactor pick_idle_cpu for affinitized tasks

hodgesds:
Refactor pick_idle_cpu to split out affinitized task handling to pick_idle_affinitized_cpu. This makes the code for handling affinitized tasks easier to follow. Remove the `cpu` field from the `taskc_ctx` as this can easily be queried with `scx_bpf_task_cpu`.


arighi (approved):
Good idea to split affinitized and non-affinitized tasks, code is much easier to follow.

PR #1977, opened 2025-05-24
scx_layered: limit layer growth to within its cpumask

etsal:
Now that layers have asssociated cpusets, confine layer growth to only pick cpus in its allowed cpuset.


likewhatevs (approved):
solid, we need this part

etsal:
Superseded by #1979

PR #1980, opened 2025-05-25
scx_layered: Replace division with bitwise shift

EricccTaiwan:
Replaces division by 1024 with a right shift by 10 when scaling runtime with task_hint.
No functional change.


htejun:
This is not a meaningful optimization. Compiler can figure this out trivially. Code should be written to maximize readability.

EricccTaiwan:
Got it!

PR #1986, opened 2025-05-25
bitmap: Use 0ULL in scx_bitmap_clear()

EricccTaiwan:
Assign 0ULL instead of 0 to match the u64 type of mask->bits[].


htejun:
Compiler can zero extend 0 completely fine. I don't think these change help.

PR #1988, opened 2025-05-25
lib/topo: Remove redundant scx_bitmask_copy() and properly comment topo_node() behavior

EricccTaiwan:
Allocate memory for topo->mask using scx_static_alloc() to avoid aliasing the input mask. This ensurs that each topo node owns its own bitmap.


etsal:
This is by design, the topo_init() call gets passed a bitmap that we deliberately "consume" by attaching it to the topology node. The scx_bitmap_copy call, however, is extraneous. The better solution would be to remove the scx_bitmap_copy and add a comment about this in the code. I can do this tomorrow, ofc feel free to amend this diff accordingly if you want to :)

And thank you for the code reviews! They have been very helpful.

etsal (approved):

PR #1989, opened 2025-05-25
scxtop: add search module

yaakov-stein:
(Sorry for the length, it's mostly tests and benchmarking, the logic is the top lines in search.rs)

In order to enable filtering perf events in scxtop, we'll need fuzzy matching functionality. This module will provide the initial fuzzy search. The fuzzy matching works in two stages:
1) Initially, the method will check if the input is a substring for each entries, and the entry will receive a score of 100 if it is. If it is not a substring, we will check if the entry contains the input spread out over the string. If it does, it will lose one point per additional character breaking up the search input. This is extremely performant even for large strings (~200-300 microseconds, similar to just calling contains).
2) If no results matched from step (1), then we will check if there was a typo. This is more expensive, and for long strings, can take 2-3 milliseconds. 

Example from step (1) (see tests for more examples): 
input = "alcncl" (len 6)
entry = "alarmtimer:alarmtimer_cancel" (len 28)
score = 100 - 22 = 78

[skim](https://github.com/skim-rs/skim) and other standard [fuzzy-matchers](https://github.com/skim-rs/fuzzy-matcher/tree/master) generally do a more complex scoring method for step (1), but do not have step (2). Given the use-case, the slightly simpler scoring method for step (1) plus the convenience of having step (2) should be fine.

Here are some performance benchmarks, comparing fuzzy_search to a basic substring_search:

Screenshot 2025-05-25 at 3 28 21 PM

We can see that for non-typo complex inputs, we have a median of ~250 microseconds and for complex inputs with a typo (the last case - Fuzzy Long Complex String), we have a median of ~2.5 milliseconds, both of which should be fine performance-wise. Benchmarking was done using [Criterion](https://docs.rs/criterion/latest/criterion/).


tools/scxtop/benches/search_benchmark.rs
```diff
@@ -0,0 +1,93 @@
...
+fn get_search() -> Search {
+    let available_perf_events_list: Vec<String> = available_perf_events()
+        .unwrap()
+        .iter()
+        .flat_map(|(subsystem, events)| {
+            events
+                .iter()
+                .map(|event| format!("{}:{}", subsystem.clone(), event.clone()))
+        })
+        .collect();
```
JakeHillion, line 14: nit: you don't need the clones here
```suggestion
    let available_perf_events_list: Vec<String> = available_perf_events()
        .unwrap()
        .into_iter()
        .flat_map(|(subsystem, events)| {
            events
                .into_iter()
                .map(move |event| format!("{}:{}", &subsystem, &event))
        })
        .collect();
```

JakeHillion (approved):
nice! Thanks for committing the benchmarks, really interesting to see your thought process there.

tools/scxtop/Cargo.toml
```diff
@@ -56,6 +56,11 @@ log-panics = { version = "2", features = ["with-backtrace"]}
 hashbrown = "0.15.2"
 smartstring = { version = "1.0.1", features = ["serde"] }
 nix = { version = "0.29", features = ["time"] }
+criterion = "0.6.0"
```
hodgesds, line 59: nit: I think you can put this into [`dev-dependencies`](https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html) since the binary shouldn't need criterion for runtime.

tools/scxtop/src/lib.rs
```diff
@@ -57,7 +59,7 @@ pub const TRACE_FILE_PREFIX: &str = "scxtop_trace";
 pub const STATS_SOCKET_PATH: &str = "/var/run/scx/root/stats";
 pub const LICENSE: &str = "Copyright (c) Meta Platforms, Inc. and affiliates.
 
-This software may be used and distributed according to the terms of the 
+This software may be used and distributed according to the terms of the
```
hodgesds, line 62: Double check this, I think the extra space is there on purpose. It could also be that the string formatting for the license in the help section could use some better formatting.
image

hodgesds (approved):

PR #1996, opened 2025-05-26
lib/sdt_alloc: Switch to new memory block on alloc overflow

EricccTaiwan:
When the static allocator runs out of space it now allocates a fresh arena block, resets the offset to zero, and recalculates the base address before resuming. 
This prevents overlapping allocations and possible memory corruption that occurred when continuing to use the exhausted block. The existing spin-lock check still guards against concurrent replacements, so thread-safety semantics are preserved.

### DIFF
--- a/lib/sdt_alloc.bpf.c
+++ b/lib/sdt_alloc.bpf.c
@@ -679,6 +679,13 @@ void __arena *scx_static_alloc(size_t bytes, size_t alignment)
 			scx_bpf_error("concurrent static memory allocations unsupported");
 			return NULL;
 		}
+
+		/* Switch to new memory block, reset offset,
+		 * and recalculate base address.
+		 */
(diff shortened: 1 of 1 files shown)


etsal:
The static allocator is only supposed to be used in very limited contexts for small amounts of data, so it returning -ENOMEM is a deliberate choice. Users should be using the buddy allocator in that case instead.

EricccTaiwan:
> The static allocator is only supposed to be used in very limited contexts for small amounts of data, so it returning -ENOMEM is a deliberate choice. Users should be using the buddy allocator in that case instead.

Got it. So there should not be any fallback for this, correct?

etsal:
So the principle of the change is correct, but there is an outstanding issue that we need to recompute the padding. I will merge this change and send out another patch.

etsal (approved):

PR #1999, opened 2025-05-27
bpfland: Revert commits incorrectly applied

arighi:
Revert three commits incorrectly pushed directly to main without a proper pull request.


arighi:
> The throttle looks really interesting!

It works on my new laptop (with a new keyboard.. hence the fat fingering), but it probably requires a bit more testing, CI wasn't happy with that. I should have throttled myself. :smile:

arighi:
> > The throttle looks really interesting!
> 
> It works on my new laptop (with a new keyboard.. hence the fat fingering), but it probably requires a bit more testing, CI wasn't happy with that. I should have throttled myself. 😄

...and I see the bug now, oh well, I'll create proper PRs to re-apply these new changes. Sorry for the noise.

Issue #1221, opened 2025-01-18
scxtop: Implement searchable events


hodgesds:
Once #1220 is merged a search function should be implemented for perf events. Pressing the `e` key renders the event selection interface. This will be a tricky feature to implement due to the fact that keybindings will need to be aware of the [`AppState`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/lib.rs#L50). In order to make this work it might require a new `AppState`, something like `TextInput`. When events are [handled](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/main.rs#L191) they are handled based on the key map. However, this should be made aware of the `AppState` so that in the `TextInput` state any keys are passed directly to [`handle_action`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/app.rs#L1720) function. From there it will need to keep a buffer of the input on the [`App`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/app.rs#L51) struct and use that buffer for filtering events in [`render_event_list`](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/app.rs#L1556). Ideally fuzzy matching would be supported similar to [skim](https://github.com/skim-rs/skim) but a substring match is fine for a first implementation.

PR #2008, opened 2025-05-27
scxtop: enabling filtering perf events via input

yaakov-stein:
See #1221 for initial issue. This update enables filtering perf events in TUI with fuzzy matching (#1989). 

The main challenge was determining how to differentiate standard keybindings from input. To do this, the code checks if the current state is AppState::Event, and if so, places characters into an input buffer as opposed to using the standard character keybindings. The escape button can be used to exit the event selection page (and can now also be used throughout the entire tool as an alternative to 'q'). Here's a quick visual view for how this works:

Screenshot 2025-05-27 at 10 25 56 PM

Additionally, the scrollbar for the event selection menu has been fixed and now functions as expected.

Example usage:

https://github.com/user-attachments/assets/84038890-43e6-4b3d-86b4-394822843a2c

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -19,6 +19,7 @@ use crate::LlcData;
 use crate::NodeData;
 use crate::PerfEvent;
 use crate::PerfettoTraceManager;
+use crate::Search;
 use crate::VecStats;
 use crate::ViewState;
 use crate::APP;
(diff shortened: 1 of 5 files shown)


yaakov-stein:
Thanks @JakeHillion - nice catch! I was handling the return/escape by going to the previous screen, which is the same way that help handles escape/quit (hence the endless bouncing back and forth if the prev_state is help). 

I think the best way to handle this would be to keep the initial prev_state when switching between the help menu and event selection page, so whenever one would enter or return, they'd then go back to whatever "main" screen they were on initially (can be default, LLC, NUMA view etc.). This seems intuitive to me as one who clicks on either help or event selection would want to return to the initial screen they were on after making their selection.

tools/scxtop/src/app.rs
```diff
@@ -1899,15 +1904,30 @@ impl<'a> App<'a> {
     fn render_event_list(&mut self, frame: &mut Frame) -> Result<()> {
         let area = frame.area();
         let default_style = Style::default().fg(self.theme().text_color());
-        let chunks = Layout::vertical([Constraint::Min(1), Constraint::Percentage(99)]).split(area);
+        let chunks = Layout::vertical([
+            Constraint::Min(1),
```
hodgesds, line 1908: silly question, is this more than 100%

hodgesds: Nice, thanks for the clarification!

hodgesds (approved):

PR #2010, opened 2025-05-28
layered: add cgrp contains matcher

likewhatevs:
Add a matcher to layered to match cgroups which contain a specified substring.

seems to work w/ test json:
image


likewhatevs:
Copy pasta of match_prefix_suffix is because I don't think there's a way to pack more into that fn w/o making the verifier sad.

scheds/rust/scx_layered/src/bpf/util.bpf.h
```diff
@@ -20,7 +20,13 @@ extern const volatile u32 debug;
 #define dbg(fmt, args...)	do { if (debug) bpf_printk(fmt, ##args); } while (0)
 #define trace(fmt, args...)	do { if (debug > 1) bpf_printk(fmt, ##args); } while (0)
 
-bool match_prefix_suffix(const char *prefix, const char *str, bool match_suffix);
+enum MatchType {
```
hodgesds, line 23: nice!

PR #2013, opened 2025-05-28
layered: cpumask match constraints

likewhatevs:
Add cpumask based matcher constraints to layer matching to nudge tasks onto the best suited layer.

This solves the problem of lo fb being overutilized due to tasks matching into their first eligible layer.

It is unclear if matching in this manner is excessive, but test results indicate this is better than what we were previously doing.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2367,7 +2367,7 @@ static bool get_aligned_layer(struct task_struct *p __arg_trusted, u64 layer_id,
 		}
 	}
 
-	if (is_cpumask) {
+	if (!is_cpumask) {
 		if (is_subset) {
 			if ((cpumask = cast_mask(lookup_layer_cpuset(layer_id))) &&
 				!bpf_cpumask_subset(p->cpus_ptr, cpumask))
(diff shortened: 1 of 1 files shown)


htejun:
This feels rather round-about. With the template, we're associating a cgroup hierarchy with a layer and getting cpuset mask from it, right? Wouldn't it make more sense to make that match cgroup prefix then? ie. Such layer should only match tasks that are in that cgroup sub-hierarchy.

etsal:
It seems like this addresses the problem of cgroup suffix/prefix matching only being up to 64 bytes long. Can we instead bump the limit to 512 bytes to handle more complex cgroup hierarchies? This has 512 * MAX_LAYERS = 4K bytes overhead so it should be essentially free.

likewhatevs:
This addresses the problem of the first matched layer by a template matching all tasks (with most of them going to lo fb because that layer can only run a subset of tasks).

> It seems like this addresses the problem of cgroup suffix/prefix matching only being up to 64 bytes long. Can we instead bump the limit to 512 bytes to handle more complex cgroup hierarchies? This has 512 * MAX_LAYERS = 4K bytes overhead so it should be essentially free.

I'm not certain tbh. Template generation doesn't mess with matchers/scope tasks matches to layers. If we inserted a cgroup prefix matcher into the spec along with template generation and raising the limit on matcher length, that would do it also, but like, I kind of wonder if there's a benefit to the tiering here, idk.

> This feels rather round-about. With the template, we're associating a cgroup hierarchy with a layer and getting cpuset mask from it, right? Wouldn't it make more sense to make that match cgroup prefix then? ie. Such layer should only match tasks that are in that cgroup sub-hierarchy.

Ahh yeah that makes sense and sounds a lot like inserting a prefix matcher while generating layers + bumping the prefix length so will do that.

likewhatevs:
ahh PEBCAK

PR #2015, opened 2025-05-28
lib, scx_p2dq: factor out arena interface programs

etsal:
Move the BPF arena API into a separate file, where it can be reused by multiple schedulers. Add argument passing to the functions to parameterize the initialization process.


hodgesds:
I wonder if it's easier/more maintainable to have a separate compilation unit for the arena setup. I don't have a good idea of which way is better though.

PR #2016, opened 2025-05-28
scx_bpfland: Support balance_power and balance_performance power modes

arighi:
CPU frequency scaling drivers, such as intel_pstate or amd_pstate, expose intermediate "balanced" performance scaling preferences: `balance_power` and `balance_performance`.

Expose them via `scx_utils::autopower, so that scx schedulers can use this information to apply more fine-grained performance / power-saving optimizations.

Adjust the behavior of  `scx_bpfland -m auto` to use the new power modes `balance_power` and `balance_performance`.

@multics69 FYI, this includes a small change to scx_lavd as well to support the new `PowerProfile::Balanced { power: true | false }` states, which are both mapped to `LAVD_PM_BALANCED`, so there's no functional change.


hodgesds (approved):
Seems legit

PR #2017, opened 2025-05-28
scx_rusty: arena library backports

EricccTaiwan:
Here are the changes I made:
- Handle allocation failure.
- Fix incorrect NULL check.
- Unlock std_lock after allocation.
- Switch to new mem block. 
- ~[RFC] Recompute padding on allocation $\to$ should this align with `*scx_static_alloc(size_t bytes, size_t alignment)` ?~

Edit:
- Fix tab usage.


etsal:
Thank you! I'll take another look.

PR #2018, opened 2025-05-28
Layered idle enq numa

likewhatevs:
scope idle cpu selected to the idle cpus of a tasks numa node


htejun:
Let's hold this for now. I'm not sure this would help much with our internal use case given that the tasks in the layers that matter are all node affined already, and it may cause problems with non-numa-aligned layers (it might not but I'm not sure).

PR #2028, opened 2025-05-29
ci.py: add --locked to cargo to detect bad locks

JakeHillion:
Occasionally we push changes that forget to lock the Cargo.lock file correctly after updating dependencies or changing versions for a release. Add `--locked` to the build step in CI to validate this.

Test plan:
- Reverted f932754 and validated that this failed.


JakeHillion:
@EricccTaiwan this should address the issue in #2025 for the future, thanks!

PR #2029, opened 2025-05-29
scx_p2dq: update load balancing

yaakov-stein:
This adds a bit of load-balancing by assigning a load-balancing LLC to each LLC in a round robin fashion (if there is a load imbalance) whenever the load-balancing timer is triggered. Then, when we are looking for an idle CPU, we'll use that load-balancing LLC to migrate the current task from the over worked LLC to the under worked LLC. 

Additionally, while testing, changing the slack_factor argument to 20 showed a slight performance boost for large machiines.


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -75,7 +75,7 @@ const volatile bool dispatch_lb_interactive = false;
 const volatile u64 dispatch_lb_busy = 75;
 const volatile u64 wakeup_lb_busy = 90;
 const volatile bool wakeup_llc_migrations = false;
-const volatile u64 lb_slack_factor = 5;
+const volatile u64 lb_slack_factor = 20;
```
hodgesds, line 78: nit: use the new const:
```
const volatile u64 lb_slack_factor = LOAD_BALANCE_SLACK;
```

hodgesds (approved):
LGTM

PR #2031, opened 2025-05-29
scx_layered: Fix vtime corruption when tasks are migrating across LLCs

htejun:
layered_enqueue() calls maybe_update_task_llc() to move the task to the LLC that it's being queued on. This involves calculating the vtime delta in the old LLC and applying that against the new LLC's vtime_now. Unfortunately, the function was reading p->scx.dsq_vtime into a local variable before calling maybe_update_task_llc(). maybe_update_task_llc() updates p->scx.dsq_vtime but layered_enqueue() keeps using the vtime read from before when determining the vtime to use to enqueue the task on the new LLC's DSQ.

This leads to completely corrupt and widely divergent vtimes, that messes up prioritization and can lead to starvations under the right conditions.

- Fix it by reading p->scx.dsq_vtime after maybe_update_task_llc().

- Check whether vtime is too far in the future, record the event in a stat counter if so and fix it up. This shouldn't happen. The new counter should give us a way to detect if something is going wrong with vtime management.


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1459,9 +1458,32 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
 	if (llc_id >= MAX_LLCS || !(llcc = lookup_llc_ctx(llc_id)))
 		return;
 
+	/*
+	 * If necessary, migrate @p to new LLC and constrain its vtime. As
+	 * maybe_update_task_llc() can update @p->scx.dsq_vtime, it can only be
+	 * read afterwards.
+	 *
+	 * A task is allowed to carray at most a slice worth of vtime budget
```
hodgesds, line 1466: nit: carry

PR #2033, opened 2025-05-30
scxtop: fix minor rendering bug

yaakov-stein:
There is a minor rendering bug when the screen is large. At the top left of the screen, the panes are a bit misaligned. Ex.:

Screenshot 2025-05-29 at 6 23 08 PM

With the fix, the panes stay aligned even at larger sizes:

Screenshot 2025-05-29 at 6 24 50 PM


JakeHillion (approved):
Nice! This has bugged me for ages.

PR #2044, opened 2025-05-30
scxtop/search: Return early if pattern is too long

EricccTaiwan:
Avoid unnecessary scan in contains_spread when the pattern is longer than the word.
Also reduce repeated `.len()` calls by caching lengths.


yaakov-stein (approved):
lgtm

PR #2046, opened 2025-05-30
scx_layered: Better prioritize highpri per-cpu kthreads

htejun:
For scheduling latency, scx_layered usually depends on the fact that a task
can be pulled by multiple CPUs and an eligible CPU is likely to open up
soon. However, if a task is a high-priority per-cpu kthread, it has to run
soon and can only run on one particular CPU. Preempt whatever is running on
that CPU.

Preempt whatever is running on the CPU for them and don't let them be
preempted by others.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1248,6 +1248,10 @@ static bool try_preempt_cpu(s32 cand, struct task_struct *p, struct task_ctx *ta
 		return false;
 	}
 
+	/* don't preempt highpri kthreads */
+	if ((curr->flags & PF_KTHREAD) && curr->scx.weight > 100)
+		return false;
+
 	/*
 	 * Don't preempt if protection against is in effect. However, open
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1235,19 +1235,23 @@ static bool try_preempt_cpu(s32 cand, struct task_struct *p, struct task_ctx *ta
...
+		return false;
+	curr = rq->curr;
 
-	ext_sched_class = (struct sched_class *)(unsigned long long)ext_sched_class_addr;
-	idle_sched_class = (struct sched_class *)(unsigned long long)idle_sched_class_addr;
-
-	if (rq && ext_sched_class_addr && idle_sched_class_addr &&
-		(rq->curr->sched_class != ext_sched_class) &&
-		(rq->curr->sched_class != idle_sched_class)) {
+	if (ext_sched_class_addr && idle_sched_class_addr &&
```
likewhatevs, line 1242: this is a nicer way to write this thx for fixing

likewhatevs (approved):

PR #2048, opened 2025-05-31
scx_lavd: simplify conditional return in can_direct_dispatch()

charliechiou:
Replace verbose if-statement with a single-line conditional return to improve readability. 

The logic remains unchanged.


multics69 (approved):

multics69:
Thanks!

PR #2052, opened 2025-05-31
scx_lavd: fix null check order

charliechiou:
Ensure the task_struct pointer is checked for NULL before being passed to `bpf_task_storage_get()`. The original code performed the null check only after it was already used.


multics69 (approved):
This makes sense.

PR #2055, opened 2025-05-31
rust: Remove redundant references in log statements

EricccTaiwan:
Avoid unnecessary use of `&` in formatting macros like `info!` and `warn!` for cleaner ~and more idiomatic~ Rust code.

Edit: Since what is considered idiomatic can be subjective.


htejun:
I'm not sure but isn't passing in reference more idiomatic? What if we want to use the variable after printing information about them?

htejun (approved):

PR #2057, opened 2025-06-01
scx_rustland_core/alloc: Remove redundant type

EricccTaiwan:
The explicit type `: *mut Node` is redundant since it can be inferred from the assignment.


arighi (approved):
LGTM, thanks.

PR #2063, opened 2025-06-02
scx_p2dq: Deprecate wakeup pick two load balancing

hodgesds:
Doing a full pick two load balance on the wakeup path is rather expensive and hurts performance. Deprecate the existing wakeup pick two load balancing in favor of the precomputed load timer based wakeup balancing.


arighi (approved):
Seems reasonable to do only a timer-based load balancing.

PR #2064, opened 2025-06-02
scx_utils: add the current GPU performance state data.

devnexen:
hints on the performance level actually set on the device.


hodgesds (approved):
LGTM

PR #2065, opened 2025-06-02
scx_wd40: Fix incorrect null check

EricccTaiwan:
`lb_domain_alloc()` was checking `cpumask` twice instead of checking `direct_greedy_cpumask`.

### DIFF
--- a/scheds/rust/scx_wd40/src/bpf/lb_domain.bpf.c
+++ b/scheds/rust/scx_wd40/src/bpf/lb_domain.bpf.c
@@ -80,7 +80,7 @@ dom_ptr lb_domain_alloc(u32 dom_id)
 	}
 
 	domc->direct_greedy_cpumask = scx_bitmap_alloc();
-	if (!domc->cpumask) {
+	if (!domc->direct_greedy_cpumask) {
 		scx_bitmap_free(domc->cpumask);
 		lb_domain_free(domc);
 		return NULL;


scheds/rust/scx_wd40/src/bpf/lb_domain.bpf.c
```diff
@@ -80,7 +80,7 @@ dom_ptr lb_domain_alloc(u32 dom_id)
 	}
 
 	domc->direct_greedy_cpumask = scx_bitmap_alloc();
-	if (!domc->cpumask) {
+	if (!domc->direct_greedy_cpumask) {
```
hodgesds, line 83: I think this should be at 90 instead no?

etsal: I'm not sure, isn't the condition at 90 about the nodemask?

hodgesds: Yeah, but line 91 is freeing the direct greedy mask, which is what this is checking for. Unless there is some ordering requirement.

etsal: I don't think I follow, in line 83 we check if the allocation in line 82 failed and if so free all allocations made until that point, same in line 90. So in line 90 we are checking if we successfully allocated the node mask, and if not we free the direct greedy mask, the regular mask, and the domain itself, because we are undoing the work done by the function up until this point.  Is there an edge case to this?

hodgesds: Yeah, that's right. Sorry I had to reread he function.

PR #2067, opened 2025-06-03
lib/arena: expand argument passing from userspace

etsal:
Refactor and expand the function signatures of the BPF test programs used to initialize scheduler state on startup. First, refactor the code to not require bouncing allocated mask pointers through a temporary global variable. Also add a currently unused argument for allocating per-topology node state at node creation time.


etsal:
TODO: Fixups for scx_p2dq and scx_chaos that use the modified API.

PR #2070, opened 2025-06-03
scx_bpfland: Rework sync wakeup logic (again)

arighi:
In case of a sync wakeup the waker releases the CPU immediately after the wakeup.

Currently the logic is to consider the waker's CPU as idle (assuming no other tasks are queued to the same CPU already) and attempt to migrate the wakee to the waker's CPU, but only if both CPUs are sharing the same LLC, which doesn't make much sense, because if both CPUs are sharing the same LLC, there's not much benefit to perform the migration (the cache-hot data is already in the LLC).

Instead, change the logic to keep the wakee running on the same CPU if it's idle and both waker and wakee share the same LLC. Otherwise, attempt to migrate the wakee to the waker's CPU to take advantage of potential cache locality.

This seems to improve responsiveness in multi-LLC systems.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -492,10 +492,12 @@ static void task_update_domain(struct task_struct *p, struct task_ctx *tctx,
  * Return true if all the CPUs in the LLC of @cpu are busy, false
  * otherwise.
  */
-static bool is_llc_busy(const struct cpumask *idle_cpumask, s32 cpu)
+static bool is_llc_busy(s32 cpu)
 {
-	const struct cpumask *primary, *l3_mask;
+	const struct cpumask *primary, *l3_mask, *idle_cpumask;
 	struct cpu_ctx *cctx;
+	int node;
+	bool ret;
 
 	primary = cast_mask(primary_cpumask);
 	if (!primary)
@@ -509,7 +511,14 @@ static bool is_llc_busy(const struct cpumask *idle_cpumask, s32 cpu)
 	if (!l3_mask)
 		l3_mask = primary;
 
-	return !bpf_cpumask_intersects(l3_mask, idle_cpumask);
+	node = __COMPAT_scx_bpf_cpu_node(cpu);
+	idle_cpumask = __COMPAT_scx_bpf_get_idle_cpumask_node(node);
+
+	ret = !bpf_cpumask_intersects(l3_mask, idle_cpumask);
+
+	scx_bpf_put_cpumask(idle_cpumask);
+
+	return ret;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


arighi:
Let's put this on hold for now, there are some regressions with this change reported by the CachyOS community. I'll investigate.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -543,30 +552,23 @@ static bool is_wake_sync(s32 prev_cpu, s32 this_cpu, u64 wake_flags)
...
- * Return the target CPU for @p in case of a sync wakeup.
- *
- * During a sync wakeup, the waker commits to releasing the CPU immediately
- * after the wakeup event, so we should consider a sync wakeup almost like
- * a direct function call between a waker and a wakee.
+ * Return true if @this_cpu and @that_cpu shares the same LLC, false
+ * otherwise.
  */
-static s32 try_sync_wakeup(const struct task_struct *p, s32 prev_cpu, s32 this_cpu)
+static bool cpus_share_llc(s32 this_cpu, s32 that_cpu)
```
hodgesds, line 558: heh, I like these variable names 😆

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -543,30 +552,23 @@ static bool is_wake_sync(s32 prev_cpu, s32 this_cpu, u64 wake_flags)
...
-	 */
-	if (!scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | this_cpu))
-		return this_cpu;
+	cctx = try_lookup_cpu_ctx(that_cpu);
+	if (!cctx)
+		return false;
+
+	llc_mask = cast_mask(cctx->l3_cpumask);
+	if (!llc_mask)
+		return true;
```
hodgesds, line 578: should this be false if the mask can't be cast?

arighi: If l3_cpumask is NULL it means that all the CPUs share the same L3 cache, or the scheduler is running with `--disable-l3`. Either way, the two CPUs are sharing the same LLC (the `--disable-l3` case is interpreted as squashing all the L3 caches into a single virtual LLC). We could return false in the `--disable-l3` case, but I'm not sure if it's really useful, we would always ignore `prev_cpu` on sync wakeups in this way.

Anyway, I'll add a comment to clarify this logic, thanks!

multics69 (approved):
The change makes sense to me. Thanks!

arighi:
I've decided to drop this PR since it still seems to regress certain workloads (reported by the CachyOS community).

Since it has become really challenging to do any change in bpfland without regressing someone's workload, I've decided to revamp scx_flash (that hasn't been touched in a while), port all the recent changes/features from scx_bpfland to scx_flash, re-align the two schedulers and then start working at the new experimental changes in scx_flash.

This should give us a solid baseline (scx_bpfland) that can be used to easily compare the more experimental changes done in scx_flash.

Therefore, I'm closing this one for now. I'll send a separate PR for the scx_flash revamp.

PR #2071, opened 2025-06-03
lib/cpumask: move idle CPU selection helpers out of scx_wd40

etsal:


etsal:
Yeah, it doesn't do exactly what we need but the more I look at this the more I think we need intermediate steps:

1) Moving from BPF cpumasks to arena cpumasks, this is necessary for performance but we can temporarily live without it
2) Changing the selection logic - this is independent of whether we actually use the topology and we can do it by having a pile of cpumasks and going through them in a specific order
3) Describing existing idle selection algorithms as topology traversal operations

Basically these helpers enable 2) and help us prepare to do 1) as a temporary measure. I will send another patch I think helps with 3).

PR #2072, opened 2025-06-03
lib/topology: add user-provided ids to topology nodes

etsal:
Topology nodes do not currently have identifiers, but existing code assumes they do. We do not fundamentally need identifiers because nodes are arena-based, so we can just hold a pointer to them. However, all existing schedulers add ids to LLCs/nodes/CPUs because they retrieve them from BPF arrays and maps. This mismatch makes integrating topologies with existing schedulers more complicated.

As a temporary measure, introduce a global indexing structure for accessing topology nodes. These are functionally identical to the arrays of structs used by existing schedulers, and should make it easier to adapt existing code to use topology nodes.

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -35,6 +35,10 @@
 #define PF_EXITING			0x00000004
 #define CLOCK_MONOTONIC			1
 
+#ifndef NR_CPUS
+#define NR_CPUS 1024
+#endif
+
 #ifndef NUMA_NO_NODE
 #define	NUMA_NO_NODE	(-1)
 #endif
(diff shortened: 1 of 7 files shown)


etsal:
The patch that we can use to handle issue 3 that I mentioned in #2071. This should make it pretty easy to write code we can adapt later to just traverse the topology instead.

scheds/include/scx/common.bpf.h
```diff
@@ -35,6 +35,10 @@
 #define PF_EXITING			0x00000004
 #define CLOCK_MONOTONIC			1
 
+#ifndef NR_CPUS
+#define NR_CPUS 512
```
hodgesds: Current gen epyc already have almost [400 cpus](https://www.amd.com/en/products/processors/server/epyc/9005-series/amd-epyc-9965.html), maybe bump this to 1024?

hodgesds (approved):

PR #2073, opened 2025-06-04
scx_wd40: use topology IDs

etsal:
Adjust scx_wd40 to use the node indices we added for the topology library, replacing scx_wd40's own arrays used for indexing. This also serves as a case study on how to use the topology indices on other schedulers.

### DIFF
--- a/scheds/rust/scx_wd40/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_wd40/src/bpf/main.bpf.c
@@ -875,8 +875,9 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(wd40_init)
 		if (ret)
 			return ret;
 
-		scx_bitmap_or(all_cpumask, all_cpumask, dom_ctxs[i]->cpumask);
 	}
+
+	scx_bitmap_or(all_cpumask, all_cpumask, topo_all->mask);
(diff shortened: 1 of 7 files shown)


scheds/rust/scx_wd40/src/bpf/main.bpf.c
```diff
@@ -564,6 +576,8 @@ void BPF_STRUCT_OPS(wd40_dispatch, s32 cpu, struct task_struct *prev)
 	u32 curr_dom = cpu_to_dom_id(cpu);
 	struct pcpu_ctx *pcpuc;
 
+	scx_arena_subprog_init();
```
hodgesds, line 579: silly question, why is this in dispatch?

hodgesds (approved):

scheds/rust/scx_wd40/src/bpf/main.bpf.c
etsal: Not silly at all, every program that uses arena pointers has to have instructions that access the arena map. This normally happens when we allocate/deallocate, so for most program's it's fine. However, dispatch directly grabs the pointers from our arrays instead of interacting with the arena map, so we need this dummy call.

PR #2080, opened 2025-06-04
scx_utils: netdev: Account for disabled NIC queues

daidavid:
netdev_irq_balance currently includes irqs for all queues which may not be enabled, filter out the irqs for the disabled queues.

Test plan:
- cargo build
- run scx_layered with --netdev-irq-balance


htejun (approved):

daidavid:
Thanks for reviewing Tejun! It looks like the "workflow/CI" also requires approvals? Could you approve that as well?

PR #2082, opened 2025-06-04
scx_p2dq: Misc cleanup

hodgesds:
Cleanup unused global CPU mask and remove counter that was used for debugging in dispatch.


yaakov-stein (approved):
LGTM, nice find

PR #2084, opened 2025-06-04
scx_layered: Implement LayerKind::Grouped::util_includes_open_cputime

htejun:
52e3ace ("scx_layered: Use only owned execution time when sizing
grouped layers") made grouped layers exclude open execution time when
determining the target number of CPUs. This was primarily to avoid
osciallting the sizes of grouped layers when confined layers' owned
protection statuses fluctuate; however, bb600e6 ("scx_layered: Simplify
owned execution protection") removed the source of most of such fluctuations
and the exclusion of open cputime made configuring grouped layers more
difficult as how much a layer open executes isn't something the user has
much control over.

Add a layer option to include open cputime in grouped layer sizing. We
should transition to make this the default behavior and eventually drop the
option.

Note that the upper bound of 1.0 for util_range is removed as a grouped
layer can legimitately target >1.0 util when including open cputime.

### DIFF
--- a/scheds/rust/scx_layered/src/config.rs
+++ b/scheds/rust/scx_layered/src/config.rs
@@ -163,6 +163,8 @@ pub enum LayerKind {
     Grouped {
         util_range: (f64, f64),
         #[serde(default)]
+        util_includes_open_cputime: bool,
+        #[serde(default)]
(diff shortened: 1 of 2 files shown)


likewhatevs (approved):
makes sense

PR #2088, opened 2025-06-05
scx_layered: Idle rnode cleanup

likewhatevs:
update idle interlocking paths to account for skip-rnode constraints.

we probably want (need?) this to account for cpuset also, but, for now just deal with node constraints.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1340,7 +1340,10 @@ static void kick_idle_cpu(struct task_struct *p, struct layer *layer)
 
 	if (!(idle_smtmask = scx_bpf_get_idle_smtmask()))
 		return;
-
+	
+	/*
+	 * XXX: this should account for cpuset also.
+	 */
 	if (!(cpuc = lookup_cpu_ctx(-1)) || !(cand_cpumask = bpf_cpumask_create())) {
 		scx_bpf_put_idle_cpumask(idle_smtmask);
 		return;
@@ -2976,7 +2979,7 @@ void BPF_STRUCT_OPS(layered_update_idle, s32 cpu, bool idle)
 
 	bpf_for(layer_id, 0, nr_layers) {
 		/*
-		* XXX this should account for cpuset also.
+		* XXX: this should account for cpuset also.
 		*/
 		if((layer = lookup_layer(layer_id)) && 
 			layer->skip_remote_node && 
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1333,25 +1333,50 @@ static void task_uncharge_qrt(struct task_ctx *taskc)
 static void kick_idle_cpu(struct task_struct *p, struct layer *layer)
 {
 	const struct cpumask *idle_smtmask;
+	struct bpf_cpumask *cand_cpumask;
+	struct task_ctx *taskc;
+	struct cpu_ctx *cpuc;
 	s32 cpu;
 
 	if (!(idle_smtmask = scx_bpf_get_idle_smtmask()))
 		return;
 
+	if (!(cand_cpumask = bpf_cpumask_create()) || !(cpuc = lookup_cpu_ctx(-1))) {
+		scx_bpf_put_idle_cpumask(idle_smtmask);
+		return;
+	}
+	
 	if (layer->kind == LAYER_KIND_CONFINED) {
 		const struct cpumask *layer_cpumask;
 
-		if (!(layer_cpumask = lookup_layer_cpumask(layer->id)))
-			goto out;
-		cpu = pick_idle_cpu_from(layer_cpumask, 0, idle_smtmask, layer);
+		if (!(layer_cpumask = lookup_layer_cpumask(layer->id))) {
+			bpf_cpumask_release(cand_cpumask);
+			scx_bpf_put_idle_cpumask(idle_smtmask);
+			return;
+		}
+
+		if (layer->skip_remote_node && (taskc = lookup_task_ctx(p)) && taskc->layered_node_mask) {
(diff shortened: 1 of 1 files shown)


likewhatevs:
~I think adding a check of skip remote node in pick_idle_from is the other one?~

ahh, layers don't have node masks, too low level for this sort of scoping I think, needs to be done higher up.

likewhatevs:
I *think* that should do it once I get that to verify (unreleased reference because of adding a bpf cpumask create to kick idle...).

likewhatevs:
I don't recall how hard this was (need to test older binaries), but I think this as it is might be too hard (i.e. some spill over when one node is saturated is good I think, maybe).

Edit: looking at https://github.com/sched-ext/scx/pull/1698 , this should be as hard as it is.

Edit Edit: I forgot that the NodeSpread growth algos are key to how this works, retesting w/ this.

likewhatevs:
Hmm, with these changes skip_rnode went down a lot (peak 30k to peak 1k).

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2945,7 +2973,18 @@ void BPF_STRUCT_OPS(layered_update_idle, s32 cpu, bool idle)
 	 * see idle set or we see the task in one of the DSQs.
 	 */
 	llc_id = cpu_to_llc_id(cpu);
+
 	bpf_for(layer_id, 0, nr_layers) {
+		if((layer = lookup_layer(layer_id)) && 
+			(layer_cpumask = lookup_layer_cpumask(layer_id)) &&
+			!bpf_cpumask_test_cpu(cpu, layer_cpumask) &&
```
likewhatevs, line 2985: @etsal correct me if I am wrong but this is all that's needed to have a similar effect on cpuset'ed workloads as the one wrt/ skip-rnode right (i.e. don't bother checking dsq's which cannot be processed)?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1333,24 +1333,49 @@ static void task_uncharge_qrt(struct task_ctx *taskc)
...
+			cpu = pick_idle_cpu_from(layer_cpumask, 0, idle_smtmask, layer);
+		}
 	} else {
-		cpu = pick_idle_cpu_from(p->cpus_ptr, 0, idle_smtmask, layer);
+		if (layer->skip_remote_node && (taskc = lookup_task_ctx(p)) && taskc->layered_node_mask) {
+			bpf_cpumask_and(cand_cpumask, p->cpus_ptr, cast_mask(taskc->layered_node_mask));
+			lstat_inc(LSTAT_SKIP_REMOTE_NODE, layer, cpuc);
+			cpu = pick_idle_cpu_from(cast_mask(cand_cpumask), 0, idle_smtmask, layer);
+		} else {
+			cpu = pick_idle_cpu_from(p->cpus_ptr, 0, idle_smtmask, layer);
```
likewhatevs, line 1377: @etsal this right here deals w/ cpuset scoping and in the other paths, layer_cpumask (which is a subset of/effectively anded with layer_cpuset) would deal w/ that scoping right?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1333,24 +1333,49 @@ static void task_uncharge_qrt(struct task_ctx *taskc)
...
+	struct bpf_cpumask *cand_cpumask;
+	struct task_ctx *taskc;
+	struct cpu_ctx *cpuc;
 	s32 cpu;
 
 	if (!(idle_smtmask = scx_bpf_get_idle_smtmask()))
 		return;
-
+	
+	if (!(cpuc = lookup_cpu_ctx(-1)) || !(cand_cpumask = bpf_cpumask_create())) {
```
hodgesds: This creates a new cpumask every enqueue, is that expensive?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1333,24 +1333,49 @@ static void task_uncharge_qrt(struct task_ctx *taskc)
...
-		if (!(layer_cpumask = lookup_layer_cpumask(layer->id)))
-			goto out;
-		cpu = pick_idle_cpu_from(layer_cpumask, 0, idle_smtmask, layer);
+		if (!(layer_cpumask = lookup_layer_cpumask(layer->id))) {
+			bpf_cpumask_release(cand_cpumask);
+			scx_bpf_put_idle_cpumask(idle_smtmask);
+			return;
+		}
+
+		if (layer->skip_remote_node && (taskc = lookup_task_ctx(p)) && taskc->layered_node_mask) {
```
hodgesds: `kick_idle_cpu` is only called from `enqueue` where `taskc` is already available. Why not pass it as a arg instead of doing a double map lookup every enqueue?

scheds/rust/scx_layered/src/bpf/main.bpf.c
likewhatevs: little idea but probably cheaper than the alternative (skipping later) when we need it. scoped creation to when we need it.

likewhatevs:
good finds @hodgesds , no perf left on floor etc.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2945,7 +2978,18 @@ void BPF_STRUCT_OPS(layered_update_idle, s32 cpu, bool idle)
 	 * see idle set or we see the task in one of the DSQs.
 	 */
 	llc_id = cpu_to_llc_id(cpu);
+
 	bpf_for(layer_id, 0, nr_layers) {
+		if((layer = lookup_layer(layer_id)) &&
```
likewhatevs, line 2983: @kkdwivedi @etsal do you think this is needed for when workloads are cpusetted?

likewhatevs: nah

PR #2089, opened 2025-06-05
scx_layered: Unbreak layered_dispatch

kkdwivedi:
Since the fixed commit, prev_taskc and prev_layer are always NULL in layered_dispatch and all checks conditional on them (keep_running() heuristics and slice assignment to the previous task) do not occur.

Move the hunk initializing them back to layered_dispatch. Before this, all keep/max/busy counters for layers are 0, after this they are populated again.

Restore the comment before slice refresh again for clarity.

Fixes: e2b6cbe7bee7 ("layered: improve compat of excl fixes on older kernels")


hodgesds (approved):
LGTM

etsal (approved):

PR #2093, opened 2025-06-06
scx_mitosis: log global + per-cell queue counts.

tommy-u:
Adds functionality to print:

  • total queue placement decisions and share of global work
  • per-queue percentages (Local, Default, Hi, Lo)
  • affinity-violation percentage

Output is width-aligned and skips empty cells.

Output example:
Total Decisions:     6872 100.0% | L:92.7% D: 5.9% hi: 1.4% lo: 0.0% | V: 0.0%
Cell  0 Decisions:   6872 100.0% | L:92.7% D: 5.9% hi: 1.4% lo: 0.0% | V: 0.0%

Modifies BPF code to add three counters.
Modifies intf.h to extend cell_stat_idx enum.
Modifies main.rs to read maps, compute cell stat diffs, compute percentages, and log output.

### DIFF
--- a/scheds/rust/scx_mitosis/src/main.rs
+++ b/scheds/rust/scx_mitosis/src/main.rs
@@ -6,6 +6,7 @@ mod bpf_skel;
 pub use bpf_skel::*;
 pub mod bpf_intf;
 
+use std::cmp::max;
 use std::collections::HashMap;
 use std::mem::MaybeUninit;
 use std::sync::atomic::AtomicBool;
@@ -13,6 +14,7 @@ use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::Duration;
 
+use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
 use clap::Parser;
@@ -35,6 +37,7 @@ use scx_utils::NR_CPUS_POSSIBLE;
 use scx_utils::NR_CPU_IDS;
 
 const MAX_CELLS: usize = bpf_intf::consts_MAX_CELLS as usize;
+const NR_CSTATS: usize = bpf_intf::cell_stat_idx_NR_CSTATS as usize;
 
 /// scx_mitosis: A dynamic affinity scheduler
 ///
@@ -42,7 +45,7 @@ const MAX_CELLS: usize = bpf_intf::consts_MAX_CELLS as usize;
 /// dynamic set of CPUs. The BPF part does simple vtime scheduling for each cell.
 ///
 /// Userspace makes the dynamic decisions of which Cells should be merged or
-/// split and which cpus they should be assigned to.
+/// split and which CPUs they should be assigned to.
 #[derive(Debug, Parser)]
 struct Opts {
     /// Enable verbose output, including libbpf details. Specify multiple
@@ -67,6 +70,17 @@ struct Opts {
     monitor_interval_s: u64,
 }
 
+// The subset of cstats we care about.
+// Local + Default + Hi + Lo = Total Decisions
+// Affinity violations are not queue decisions, but
+// will be calculated separately and reported as a percent of the total
+const QUEUE_STATS_IDX: [bpf_intf::cell_stat_idx; 4] = [
+    bpf_intf::cell_stat_idx_CSTAT_LOCAL,
+    bpf_intf::cell_stat_idx_CSTAT_DEFAULT_Q,
+    bpf_intf::cell_stat_idx_CSTAT_HI_FALLBACK_Q,
+    bpf_intf::cell_stat_idx_CSTAT_LO_FALLBACK_Q,
+];
+
 unsafe fn any_as_u8_slice<T: Sized>(p: &T) -> &[u8] {
     unsafe {
         ::std::slice::from_raw_parts((p as *const T) as *const u8, ::std::mem::size_of::<T>())
@@ -84,6 +98,9 @@ struct Scheduler<'a> {
     prev_percpu_cell_cycles: Vec<[u64; MAX_CELLS]>,
     monitor_interval: Duration,
     cells: HashMap<u32, Cell>,
+    // These are the per-cell cstats.
+    // Note these are accumulated across all CPUs.
+    prev_cell_stats: [[u64; NR_CSTATS]; MAX_CELLS],
 }
 
 impl<'a> Scheduler<'a> {
@@ -111,6 +128,7 @@ impl<'a> Scheduler<'a> {
             prev_percpu_cell_cycles: vec![[0; MAX_CELLS]; *NR_CPU_IDS],
             monitor_interval: Duration::from_secs(opts.monitor_interval_s),
             cells: HashMap::new(),
+            prev_cell_stats: [[0; NR_CSTATS]; MAX_CELLS],
         })
     }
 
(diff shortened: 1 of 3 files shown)


tommy-u:
Please review @dschatzberg and @dforsyth 🙏

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+            bpf_intf::cell_stat_idx_CSTAT_HI_FALLBACK_Q,
+            bpf_intf::cell_stat_idx_CSTAT_LO_FALLBACK_Q,
+        ];
+        // Get total decisions
+        let global_queue_decisions: u64 = self.cell_stats_delta
+            .iter()
+            .flat_map(|cell| queue_stats_idx.iter().map(|&idx| cell[idx as usize]))
+            .sum();
+
+        // We don't want to divide by zero later, but this is never expected.
```
dforsyth, line 294: > never expected

Should we return an error, then?

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
         uei_report!(&self.skel, uei)
     }
 
-    /// Output various debugging data like per cell stats, per-cpu stats, etc.
-    fn debug(&mut self) -> Result<()> {
+    fn load_cpu_contexts(&self) -> Result<Vec<Vec<u8>>> {
```
dforsyth: We can use https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_mitosis/src/main.rs#L196 to go straight to cpu_ctx

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
-                    .zip(cpu_ctx.cell_cycles.iter())
-                    .map(|(a, b)| (b - a) as i64)
-                    .collect();
-                self.prev_percpu_cell_cycles[cpu] = cpu_ctx.cell_cycles;
-                trace!("CPU {}: {:?}", cpu, diff_cycles);
+            Ok(Some(v)) => v,
+            Ok(None) => return Err(anyhow::anyhow!("Found no values for cpu_ctxs map")),
+            Err(e) => return Err(anyhow::anyhow!("Error looking up cpu_ctxs map: {:?}", e)),
+        };
+        Ok(v)
```
dschatzberg: You're trying to go from `Result<Option<T>>` to `Result<T>` - I'd say the more idiomatic way to do this would be.

`return lookup_percpu(...).with_context(|| "Error look up cpu_ctxs map")?.ok_or(Err(anyhow!("Found no values for cpu_ctxs map"));`

The first part up until the `?` converts the `Result<Option<T>>` into `Option<T>` and then the `ok_or()` part converts `Option<T>` to `Result<T>`

This may all be moot given David's comment though

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -107,9 +112,10 @@ impl<'a> Scheduler<'a> {
 
         Ok(Self {
             skel,
-            prev_percpu_cell_cycles: vec![[0; MAX_CELLS]; *NR_CPU_IDS],
             monitor_interval: std::time::Duration::from_secs(opts.monitor_interval_s),
             cells: HashMap::new(),
+            prev_cell_stats: [[0; NR_CSTATS]; MAX_CELLS],
+            cell_stats_delta: [[0; NR_CSTATS]; MAX_CELLS],
```
dschatzberg: Do we need to store `cell_stats_delta` in the `Scheduler` object itself or could it be returned from `calculate_cell_stat_delta`? I think this data is only needed for each invocation of `debug()` so we can reduce its scope.

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+    fn calculate_cell_stat_delta(&mut self) -> Result<()> {
+        self.cell_stats_delta = [[0; NR_CSTATS]; MAX_CELLS];
+
+        // Sum each CPU's stats into the diff array
+        let v = self.load_cpu_contexts()?;
+        for (cpu, ctx) in v.iter().enumerate() {
+            let cpu_ctx = unsafe {
+                let ptr = ctx.as_slice().as_ptr() as *const bpf_intf::cpu_ctx;
+                &*ptr
+            };
```
dschatzberg: I think using `read_cpu_ctxs` that David added here makes sense. I have some thoughts on the implementation of that function, but we can discuss that separately.

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+            // Each CPU has a u64 cstats[MAX_CELLS][NR_CSTATS];
+            for cell in 0..MAX_CELLS {
```
dschatzberg: This is probably fine for now - but we are likely to see cases of cells dynamically becoming in_use or not - It might make sense to handle that correctly here. Can also just drop a TODO and get to it later

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+
+        Ok(())
+    }
+
+    fn calculate_cell_stat_delta(&mut self) -> Result<()> {
+        self.cell_stats_delta = [[0; NR_CSTATS]; MAX_CELLS];
+
+        // Sum each CPU's stats into the diff array
+        let v = self.load_cpu_contexts()?;
+        for (cpu, ctx) in v.iter().enumerate() {
```
dschatzberg: Are you using the `cpu` index here at all? Probably don't even need an `enumerate()` if not.

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+                }
+            }
+        }
+        // Right now cell_stats_delta holds the total counts
+        // So subtracting prev will give us the diff, what we're after.
+        // We can also update prev_cell_stats here by adding the diff
+        for cell in 0..MAX_CELLS {
+            for stat in 0..NR_CSTATS {
+                self.cell_stats_delta[cell][stat] -= self.prev_cell_stats[cell][stat];
+                self.prev_cell_stats[cell][stat] += self.cell_stats_delta[cell][stat];
```
dschatzberg: I think I would structure all these loops as follows:

```
for cell in cells {
  for stat in stats {
    let prev_cell_stat = self.prev_cell_stats[cell][stat];
    let mut cur_cell_stat = 0;
    for cpu_ctx in cpu_ctxs {
      cur_cell_stat += ...
    }
    cell_stats_delta[cell][stat] = cur_cell_stat - prev_cell_stat;
    prev_cell_stat[cell][stat] = cur_cell_stat;
  }
}
```

It's a bit more intuitive to me that way and also reduces number of iterations

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+        // The subset of cstats we care about.
+        // Local + Default + Hi + Lo = Total Decisions
+        // Affinity violations are not queue decisions, but
+        // will be calculated separately and reported as a percent of the total
+        let queue_stats_idx = [
+            bpf_intf::cell_stat_idx_CSTAT_LOCAL,
+            bpf_intf::cell_stat_idx_CSTAT_DEFAULT_Q,
+            bpf_intf::cell_stat_idx_CSTAT_HI_FALLBACK_Q,
+            bpf_intf::cell_stat_idx_CSTAT_LO_FALLBACK_Q,
+        ];
```
dschatzberg: I think I'd make this a global const. At the very least I'd pass it by reference to the other methods but I think global const is even better

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+            share_of_global,
+            queue_pct[0], queue_pct[1], queue_pct[2], queue_pct[3],
+            affinity_violations_percent,
+            width = decisions_format_width
+        );
+        Ok(())
+    }
+
+    // Queue stats for the whole node
+    fn log_global_queue_stats<const N: usize>(
```
dschatzberg: I think we should be able to avoid parameterizing this method by making queue_stats_idx a global variable

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+    fn calculate_distribution_and_log<const N: usize>(
+        &self,
+        queue_counts:           &[u64; N],
+        global_queue_decisions:  u64,
+        scope_queue_decisions:         u64,
+        scope_affn_viols:        u64,
+        prefix:                 &str)
+        -> Result<()> {
+
+        debug_assert!(global_queue_decisions > 0 && scope_queue_decisions > 0);
```
dschatzberg: I think I would probably just handle these all being 0 by setting share_of_global and queue_pct to 0 even though in practice we probably wouldn't see this.

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,31 +131,200 @@ impl<'a> Scheduler<'a> {
...
+        for (i, total) in queue_counts.iter().enumerate() {
+            queue_pct[i] = 100.0 * (*total as f64) / (scope_queue_decisions as f64);
+        }
+
+        // These are summed differently for the global and per-cell totals.
+        let affinity_violations_percent = 100.0 * (scope_affn_viols as f64) / (scope_queue_decisions as f64);
+
+        // Decisions width. We know global_queue_decisions is non-zero.
+        const MIN_DECISIONS_WIDTH: usize = 5;
+        let decisions_format_width: usize = max(MIN_DECISIONS_WIDTH, (global_queue_decisions as f64).log10().ceil() as usize);
```
dschatzberg: Why do you need to calculate the decision width or set a minimum?

scheds/rust/scx_mitosis/src/main.rs
dforsyth: Does it make sense to start collecting this in a per-cell stat structs?

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Switched this to returning an error.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Nice, switched to using `read_cpu_ctxs()`

scheds/rust/scx_mitosis/src/main.rs
tommy-u: I applied David's approach. Will use this pattern in the future.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Nice, I reworked things to return `cell_stats_delta` from `calculate_cell_stat_delta` as you suggested. Minimizing the scope of `cell_stats_delta`

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Eliminated by using `read_cpu_ctxs `.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Added a TODO.

Perhaps we should provide helper functions for iterating the in_use cells and the in_use CPUs.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Good eye. Removed cpu.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: This is way clearer. I made the change.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Made QUEUE_STATS_IDX global.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Done

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Modified this to set these percentages to zero if `scope_queue_decisions` is 0, and to produce debug output.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: TLDR: This isn't necessary, but it makes the printout easier for humans to read when the width of total counts is different from the width of cell counts, and between logging intervals.

Contrived example without this formatting (total decisions >> cell decisions):
```
Total Decisions:  7988 100.0% | L:94.7% D: 3.5% hi: 1.8% lo: 0.0% | V: 0.0%
        Cell  0:  988  12.3% | L:94.7% D: 3.5% hi: 1.8% lo: 0.0% | V: 0.0%
```
Contrived example with this formatting:

```
Total Decisions:  7988 100.0% | L:94.7% D: 3.5% hi: 1.8% lo: 0.0% | V: 0.0%
        Cell  0:   988  12.3% | L:94.7% D: 3.5% hi: 1.8% lo: 0.0% | V: 0.0%
```

Actual example without this formatting (total decisions in iteration i << total decisions in iteration i+1)
```
Total Decisions: 7420 100.0% | L:91.3% D: 7.0% hi: 1.8% lo: 0.0% | V: 0.0%
        Cell  0: 7420 100.0% | L:91.3% D: 7.0% hi: 1.8% lo: 0.0% | V: 0.0%
CELL[0]: ffffffff
Total Decisions: 23540 100.0% | L:95.8% D: 2.6% hi: 1.6% lo: 0.0% | V: 0.0%
        Cell  0: 23540 100.0% | L:95.8% D: 2.6% hi: 1.6% lo: 0.0% | V: 0.0%
```

A few alternatives: 1) remove it and make the log harder to read; 2) Remove the random constant '5' by tracking the widest value we've seen so far and using that; 3) Move this number to the end of the log line (harder to intuitively understand I think).

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Dan clarified this with his restructured loops. Good idea going forward

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,32 +139,185 @@ impl<'a> Scheduler<'a> {
...
+                let prev_cell_stat = self.prev_cell_stats[cell][stat];
+                let mut cur_cell_stat = 0;
+
+                // Accumulate stats from all CPUs
+                for cpu_ctx in cpu_ctxs.iter() {
+                    cur_cell_stat += cpu_ctx.cstats[cell][stat];
+                }
+
+                // Calculate delta and update previous stat
+                cell_stats_delta[cell][stat] = cur_cell_stat - prev_cell_stat;
```
dschatzberg: nit: don't need to create a local `prev_cell_stat` variable, just use `self.prev_cell_stats[cell][stat]`

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,32 +139,185 @@ impl<'a> Scheduler<'a> {
...
+    ) -> Result<()> {
+        // Get total decisions
+        let global_queue_decisions: u64 = cell_stats_delta
+            .iter()
+            .flat_map(|cell| QUEUE_STATS_IDX.iter().map(|&idx| cell[idx as usize]))
+            .sum();
+
+        // We don't want to divide by zero later, but this is never expected.
+        if global_queue_decisions == 0 {
+            return Err(anyhow::anyhow!("Error: No queueing decisions made globally"));
```
dschatzberg: return `Err(anyhow!(...` can be shortened to `bail!(...)`

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,32 +139,185 @@ impl<'a> Scheduler<'a> {
...
-                };
-                let diff_cycles: Vec<i64> = self.prev_percpu_cell_cycles[cpu]
-                    .iter()
-                    .zip(cpu_ctx.cell_cycles.iter())
-                    .map(|(a, b)| (b - a) as i64)
-                    .collect();
-                self.prev_percpu_cell_cycles[cpu] = cpu_ctx.cell_cycles;
-                trace!("CPU {}: {:?}", cpu, diff_cycles);
+
+    fn calculate_distribution_and_log<const N: usize>(
```
dschatzberg: I'm interested in if you can get rid of this parameterization. N should always be 4, so it shouldn't be needed. I don't think you can just replace N with `QUEUE_STATS_IDX.len()` since that's not a const expression. I think it might be cleaner to just declare `const QUEUE_STATS_LEN: usize = 4` and use that everywhere rather than parameterizing

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -125,32 +139,185 @@ impl<'a> Scheduler<'a> {
...
+        // We know global_queue_decisions is non-zero.
+        let share_of_global = 100.0 * (scope_queue_decisions as f64) / (global_queue_decisions as f64);
+
+        // Each queue's % of the scope total
+        let mut queue_pct = if scope_queue_decisions == 0 {
+            debug!("No queue decisions in scope, zeroing out queue distribution");
+            [0.0; N]
+        } else {
+            queue_counts.iter().map(|&q_count| 100.0 * (q_count as f64) / (scope_queue_decisions as f64))
+                .collect::<Vec<f64>>().try_into().expect("Failed to convert Vec to array")
```
dschatzberg: I *think* you can use https://doc.rust-lang.org/std/array/fn.from_fn.html to do this - not sure if we have that in our Rust version just yet.

If that doesn't work, I think I'd write this as:

```
let mut out = [0; N];
for i in 0..queue_counts.len() {
  out[i] = 100.0 * (queue_counts[i] as f64) ...
}
```
Which is how you'd write it if this were C or something, but also just avoids the extra collection through a Vec, the failure case, etc.

scheds/rust/scx_mitosis/src/main.rs
dschatzberg: Thanks, maybe just drop a comment explaining this inline.

dschatzberg (changes requested):
This is pretty much all set - just a couple small things and then this is good to merge

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Done

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Done

scheds/rust/scx_mitosis/src/main.rs
tommy-u: By changing `QUEUE_STATS_IDX` to const, `QUEUE_STATS_IDX.len()` actually becomes a compile time constant. I used this to de-parameterize the function.

The logging step of this function assumes 4 elements in `queue_pct.len()`. I changed the `debug_assert()` to a conditional check with a `bail!()` if there aren't exactly 4 elements. This seems hacky, but it's probably better than the complexity of generating the format string and arguments dynamically.

scheds/rust/scx_mitosis/src/main.rs
tommy-u: `from_fn()` works great, way clearer!

scheds/rust/scx_mitosis/src/main.rs
tommy-u: Done

dschatzberg (approved):

PR #2097, opened 2025-06-06
scx_p2dq: removing redundant check

yaakov-stein:
It appears that we're doing this check and "and" twice, even though the stage in between should not have an affect on this. This removes the redundant check.


JakeHillion (approved):
LGTM

hodgesds (approved):

yaakov-stein:
As an update - did some checking and ensured that the mask never changes between the two checks (did a bpf_cpumask_copy and then checked for equality before the second check).

PR #2098, opened 2025-06-06
ci: add CI-Test-Kernel trailers

JakeHillion:
Currently our CI gates PRs/the merge queue based on whether schedulers succeed under the sched_ext/for-next tracked kernel. This is good for most code changes, but doesn't do what we need when making a change for a specific kernel or updating the kernel locks themselves. An update to the sched_ext/for-next kernel that breaks will stop at the PR, but an update to bpf/bpf-next will go in and fail at the next CI run.

Add support for trailers in commit messages requesting a specific kernel be tested. Expand our testing matrix to include this kernel for each scheduler as well as the kernel being tested in that run. Include this trailer in the automatically generated update commits so the kernel(s) that are being updated get tested at PR time.

Test plan:
- This PR tests with the bpf/bpf-next kernel as well as the default sched_ext/for-next.
- It's unclear how this interacts with the merge queue/already merged commits, we may need some extra variable to track the merge base properly. Unfortunately merge queue stuff is nearly impossible to test, so will follow up with fixes if needed. Worst case we test too many kernels until this is patched.

CI-Test-Kernel: bpf/bpf-next

### DIFF
--- a/.github/include/list-integration-tests.py
+++ b/.github/include/list-integration-tests.py
@@ -3,19 +3,75 @@
 import itertools
 import json
 import os
+import subprocess
 import sys
+
+
+def get_kernel_trailers_from_commits():
+    """Get CI-Test-Kernel trailers from commits between current HEAD and base branch."""
+    # In GitHub Actions, GITHUB_BASE_REF contains the target branch name
+    # Default to main branch if not in a PR context
+    base_ref = os.environ.get("GITHUB_BASE_REF", "main")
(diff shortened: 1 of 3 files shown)


JakeHillion:
As soon as I opened the PR I realised this doesn't actually stop the PR going in because GitHub is dumb and doesn't let you define in code whether a workflow matters, and it's differently named for every matrix entry. Will follow up afterwards with some code changes to make the checks required.

PR #2101, opened 2025-06-06
scx_mitosis: Use local time for log timestamps

EricccTaiwan:
Log timestamps in local time rather than UTC.

Ref: #2081 (I forgot to update `scx_mitosis` :sweat_smile:)

### DIFF
--- a/scheds/rust/scx_bpfland/src/main.rs
+++ b/scheds/rust/scx_bpfland/src/main.rs
@@ -700,7 +700,7 @@ fn main() -> Result<()> {
 
     let mut lcfg = simplelog::ConfigBuilder::new();
     lcfg.set_time_offset_to_local()
-        .unwrap()
+        .expect("Failed to set local time offset")
         .set_time_level(simplelog::LevelFilter::Error)
         .set_location_level(simplelog::LevelFilter::Off)
         .set_target_level(simplelog::LevelFilter::Off)
--- a/scheds/rust/scx_chaos/src/main.rs
+++ b/scheds/rust/scx_chaos/src/main.rs
@@ -20,7 +20,7 @@ fn main() -> anyhow::Result<()> {
         llv,
         simplelog::ConfigBuilder::new()
             .set_time_offset_to_local()
-            .unwrap()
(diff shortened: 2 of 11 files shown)


dschatzberg:
Could you switch the `unwrap()` to an `expect()` which documents the failure case?

EricccTaiwan:
> Could you switch the `unwrap()` to an `expect()` which documents the failure case?

Sure, would it make sense to change it to `.expect("Failed to set local time offset")`?
Also, would it make sense to align this with the other schedulers, where I consistently use `.err()` to handle errors?
Thanks.

dschatzberg:
> Sure, would it make sense to change it to .expect("Failed to set local time offset")?

Yeah, that `expect` line is fine or changing it to return an error (instead of panic) works too.

> Also, would it make sense to align this with the other schedulers, where I consistently use .err() to handle errors?

Not sure what you mean by this - I see consistent use of `unwrap()` in https://github.com/sched-ext/scx/commit/76e72998a36475a42686a45fcf96184cf7e06145 - stylistically, I think unwrap is generally poor form and I try to avoid it in lieu of expect (since it self-documents better) for logic errors that cannot be appropriately handled.

In most of these cases, since its being invoked from main() you could just return the error but panic is pretty reasonable here.

EricccTaiwan:
> Not sure what you mean by this - I see consistent use of `unwrap()` in [76e7299](https://github.com/sched-ext/scx/commit/76e72998a36475a42686a45fcf96184cf7e06145) - stylistically, I think unwrap is generally poor form and I try to avoid it in lieu of expect (since it self-documents better) for logic errors that cannot be appropriately handled.
> 
> In most of these cases, since its being invoked from main() you could just return the error but panic is pretty reasonable here.

Oh, I misremembered -- I had used `unwrap()` before, not `err()`.
Thanks for pointing that out. I'll switch to using `expect()` instead, as you suggested, to better document the failure case.

EricccTaiwan:
Hi @dschatzberg , PTAL thanks!

dschatzberg (approved):

PR #2102, opened 2025-06-06
lib/cpumask: Add helpers for idle cpu selection

etsal:
Add helpers for idle cpu selection on arena-based cpumasks.


lib/cpumask.bpf.c
```diff
@@ -8,6 +8,77 @@ extern const volatile u32 nr_cpu_ids;
...
+
+	if (unlikely(mask_size < SCXMASK_NLONG))
+		return -EINVAL;
+
+	for (i = 0; i < mask_size && can_loop; i++) {
+		ind = (*start + i) % mask_size;
+
+		old = mask->bits[ind];
+		if (old)
+			continue;
```
EricccTaiwan: Shouldn't this be `if (!old)`? When old == 0 there are no available bits, so we should skip.

lib/cpumask.bpf.c
```diff
@@ -8,6 +8,77 @@ extern const volatile u32 nr_cpu_ids;
...
+
+		cpu = scx_ffs(old);
+		new = old & ~(1ULL << cpu);
+		if (__sync_val_compare_and_swap(&mask->bits[ind], old, new) == new)
+			return -EAGAIN;
```
EricccTaiwan: I think this should check `!= old`. CAS returns the previous value, so a mismatch (`!=old`) indicates the CAS failed.

lib/cpumask.bpf.c
```diff
@@ -8,6 +8,77 @@ extern const volatile u32 nr_cpu_ids;
...
+}
+
+__weak s32
+scx_bitmap_pick_any_cpu_from(scx_bitmap_t __arg_arena mask, u64 __arg_arena *start)
+{
+	s32 cpu;
+
+	do {
+		cpu = scx_bitmap_pick_any_cpu_once(mask, start);
+	} while (cpu == -EAGAIN && can_loop);
```
yaakov-stein, line 51: Probably stupid question, but where is can_loop defined?

EricccTaiwan: > The can_loop macro was introduced in kernel commit ba39486...

You can check PR #1161 !

etsal: It's somewhere in the arena-related headers, it is basically open-coded BPF assembly that embeds a counter in the loop so that the verifier is convinced the loop can terminate.

lib/cpumask.bpf.c
etsal: Unless I'm misreading this or the documentation is wrong, this specific intrinsic returns the desired value on error: https://docs.ebpf.io/linux/concepts/concurrency/. I don't have a machine available to test it right now but will have one later tonight.

yaakov-stein (approved):
I'll do some testing with it once it's merged, but LGTM

lib/cpumask.bpf.c
EricccTaiwan: I ran a quick test using this minimal example [try.c](https://gist.github.com/EricccTaiwan/5ea5db523d93bf8d8d85e11eb5f186de)
From the results, it seems that checking `ret != old` is a more robust way to determine whether the CAS succeeded or failed.
Would love to hear your thoughts on this.

etsal: Fair enough, assuming the intrinsic behaves consistently across architectures (in this case BPF),  this seems to be an error in the documentation. I will test to be sure, but GCC's documentation is consistent with your results https://gcc.gnu.org/onlinedocs/gcc-4.5.2/gcc/Atomic-Builtins.html so all signs point to this needing to be changed.

hodgesds: Since we are compiling using LLVM it's probably better to look at the [tests](https://github.com/llvm/llvm-project/blob/b7ef3e7a8595fd5e80ed3d1ef2fe9cd7e3b63530/llvm/test/CodeGen/BPF/atomics_2.ll#L139). I don't know if a standalone C program will replicate the behavior.

etsal: Can confirm the documentation was wrong, updated the diff.

PR #2104, opened 2025-06-07
lib/cpumask: Fix cmpxchg failure detection

EricccTaiwan:
cmpxchg returns the previous value, not the new one. To check for failure, the return value should not equal the expected old value.

Ref: PR #2102


etsal:
This must have slipped through... LGTM

PR #2105, opened 2025-06-07
lib: Separate out cpumask helpers that depend on bpf_cpumask_populate

hodgesds:
On older kernels `bpf_cpumask_populate` is not available. Split out the arena cpumask helpers into a separate file for schedulers that need to convert arena cpumasks into bpf cpumasks (wd40). This provides better compatibility for schedulers that only need arena cpumask operations without the need of converting arena cpumasks into bpf cpumasks.

### DIFF
--- a/lib/cpumask.bpf.c
+++ b/lib/cpumask.bpf.c
@@ -87,23 +87,6 @@ scx_bitmap_vacate_cpu(scx_bitmap_t __arg_arena mask, s32 cpu)
 	return 0;
 }
 
-__weak int
-scx_bitmap_to_bpf(struct bpf_cpumask __kptr *bpfmask __arg_trusted,
-		   scx_bitmap_t __arg_arena scx_bitmap)
-{
-	struct scx_bitmap *tmp;
-	int ret;
-
-	tmp = scx_percpu_scx_bitmap_stack();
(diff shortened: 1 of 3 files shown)


hodgesds:
This fixes runtime issues with p2dq on older kernels where it fails to load:
```
08:36:18 [WARN] libbpf: btf: type [163714]: referenced type [163712] is not FUNC_PROTO

08:36:18 [WARN] libbpf: failed to load module [amdxcp]'s BTF object #2: -22

08:36:18 [WARN] libbpf: extern (func ksym) 'bpf_cpumask_populate': not found in kernel or module BTFs

08:36:18 [WARN] libbpf: failed to load object 'bpf_bpf'

08:36:18 [WARN] libbpf: failed to load BPF skeleton 'bpf_bpf': -22

Caused by:
```

etsal:
I don't think we need to split the file apart, we only need to conditionally check whether the symbol exists and return EOPNOTSUPP if not.

etsal:
#2108 uses a compat macro to solve this without code movement, I will add an older kernel to confirm this passes verification on older kernels tonight.

hodgesds:
I think we can close this with the new compat helpers.

PR #2110, opened 2025-06-08
scx_lavd/main: Use WRITE_ONCE to update shared state safely

EricccTaiwan:
Replace plain assignments with `WRITE_ONCE` to ensure cross-CPU visibility and prevent reordering in critical sections.

```c
a = READ_ONCE(b) // original
WRITE_ONCE(a, b) // rewrite
```

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -463,8 +463,8 @@ static void advance_cur_logical_clk(struct task_struct *p)
 	u64 nr_queued, delta, new_clk;
 	int i;
 
-	vlc = READ_ONCE(p->scx.dsq_vtime);
-	clc = READ_ONCE(cur_logical_clk);
+	WRITE_ONCE(vlc, p->scx.dsq_vtime);
+	WRITE_ONCE(clc, cur_logical_clk);
 
 	bpf_for(i, 0, LAVD_MAX_RETRY) {
 		/*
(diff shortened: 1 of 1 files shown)


kkdwvd:
I don't see how this change prevents reorder (maybe compiler reorderings, sure, but then barrier() would be more clearer). What race or reordering does this prevent?

EricccTaiwan:
> I don't see how this change prevents reorder (maybe compiler reorderings, sure, but then barrier() would be more clearer). What race or reordering does this prevent?

I found `WRITE_ONCE(a, b)` more intuitive than `a = READ_ONCE(b)`, since it makes the write intent explicit. But you're right -- it doesn’t prevent any meaningful reordering. 

Happy to revise the commit message or drop this change if preferred.

kkdwvd:
It's not the same thing. `READ_ONCE(b)` implies that you're doing a volatile read of b. When you write `WRITE_ONCE(a, b)` this means you're loading the value of b without a `READ_ONCE`. Even then, in all four cases you change, the value is being written non-concurrently, so there is no data race. Instead, the values being read _are_ accessed concurrently.

So the change is incorrect. The compiler now things loaded values are going to be unchanged, so it can perform additional optimizations and break the code. The `READ_ONCE` is necessary to suppress such assumptions.

EricccTaiwan:
> It's not the same thing. `READ_ONCE(b)` implies that you're doing a volatile read of b. When you write `WRITE_ONCE(a, b)` this means you're loading the value of b without a `READ_ONCE`. Even then, in all four cases you change, the value is being written non-concurrently, so there is no data race. Instead, the values being read _are_ accessed concurrently.
> 
> So the change is incorrect. The compiler now things loaded values are going to be unchanged, so it can perform additional optimizations and break the code. The `READ_ONCE` is necessary to suppress such assumptions.

Thank you for the explanation.
I overlooked that `WRITE_ONCE(a, b)` doesn't protect the read side. As you pointed out, in these cases it's actually the read of `b` that may be concurrent, and removing `READ_ONCE` can lead to incorrect compiler assumptions and subtle bugs, especially in a BPF context.

I'll go ahead and close this PR. Appreciate your insight!

PR #2113, opened 2025-06-09
macros: add ifndef to all likely/unlikely definitions

JakeHillion:
We're providing the standard definitions for likely/unlikely/__maybe_unused in certain headers but they're not gated on already existing. This can cause a compiler error depending on the order of includes. Add an `ifndef` to ensure these macros are only defined once.

Test plan:
- CI


htejun:
Can we rearrange stuff so that we have only one definition of the macros?

JakeHillion:
> Can we rearrange stuff so that we have only one definition of the macros?

The time I got a collision it was with `bpf-helpers.h` which isn't ours so a bit of a lost cause.

PR #2120, opened 2025-06-10
Fixes for layered regressions

kkdwivedi:
We noticed regressions for production workloads due to recent changes. This pull request reverts one (which we will revisit by reintroducing a modified version of the patch after testing internally), and a fix for protected layers that causes idle selections to drop to 0 under layer saturation / load.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -3490,7 +3490,8 @@ static s32 init_layer(int layer_id)
  */
 static s32 init_cpu(s32 cpu, int *nr_online_cpus,
 		    struct bpf_cpumask *cpumask,
-		    struct bpf_cpumask *tmp_big_cpumask)
+		    struct bpf_cpumask *tmp_big_cpumask,
+		    struct bpf_cpumask *tmp_unprotected_cpumask)
 {
 	const volatile u8 *u8_ptr;
 	struct cpu_ctx *cpuc;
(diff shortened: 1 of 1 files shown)


likewhatevs:
~I think the interlocking change should be a flag btw, not a revert.~

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1330,27 +1330,18 @@ static void task_uncharge_qrt(struct task_ctx *taskc)
...
-static void kick_idle_cpu(struct task_struct *p, struct layer *layer)
+static void layer_kick_idle_cpu(struct layer *layer)
 {
-	const struct cpumask *idle_smtmask;
+	const struct cpumask *layer_cpumask, *idle_smtmask;;
 	s32 cpu;
 
-	if (!(idle_smtmask = scx_bpf_get_idle_smtmask()))
+	if (!(layer_cpumask = lookup_layer_cpumask(layer->id)) ||
+	    !(idle_smtmask = scx_bpf_get_idle_smtmask()))
```
hodgesds, line 1339: I wonder if `scx_bpf_get_idle_cpumask` would help, especially when a machine is mostly saturated. So maybe have a param to control whether idle smt or idle cpu is used. Or try some clever heuristic based on load.

kkdwvd: I will give it a try!

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1694,7 +1680,7 @@ static bool keep_running(struct cpu_ctx *cpuc, struct task_struct *p,
 		 * have tasks waiting, keep running it. If there are multiple
 		 * competing preempting layers, this won't work well.
 		 */
-		u64 dsq_id = layer_dsq_id(layer->id, cpuc->llc_id);
+		u32 dsq_id = layer_dsq_id(layer->id, cpuc->llc_id);
```
htejun: Ah, this was a stray fix in the commit. Can you revive this as a separate commit? It's a hidden bug which can bite us in the future.

PR #2121, opened 2025-06-10
scx_flash: Rework direct dispatch

arighi:
We take a really conservative approach to direct dispatching and disable it whenever there’s at least one task waiting in the per-node DSQ. This is meant to prevent scenarios where tasks are repeatedly re-dispatched in a tight loop, potentially starving other tasks queued in the per-node DSQ.

However, this strategy sacrifices efficiency under normal conditions, as it bypasses many valid opportunities for direct dispatch.

In practice, the only tasks at real risk of starvation are those restricted to a single CPU. If such a task ends up in a per-node DSQ and its only eligible CPU is occupied by a direct dispatch loop, it may be indefinitely delayed. In contrast, tasks without strict affinity constraints are more likely to be picked up by any CPU, significantly reducing the starvation risk.

In reality, the tasks most vulnerable to starvation in this scenario are per-CPU kthreads.

To improve efficiency, change the default behavior to always perform direct dispatch when an idle CPU is available and always dispatch per-CPU kthreads directly, bypassing the per-node DSQ.

While this makes the scheduler slightly less fair, giving per-CPU kthreads a huge priority boost, it is generally beneficial for the system: delays to per-CPU kthreads can cascade into broader system delays.

There is still a small risk of starving user-space per-CPU tasks, but in most real-world workloads this risk should be negligible. For workloads that require the previous behavior, the --local-pcpu option can now be used to restore it. The option name and description remain coherent with the behavior, as it still results in a form of prioritization for per-CPU tasks when enabled.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -873,13 +873,17 @@ out_put_cpumask:
  * Return true if we can perform a direct dispatch on @node, false
  * otherwise.
  */
-static bool can_direct_dispatch(int node)
+static inline bool can_direct_dispatch(s32 cpu)
 {
 	/*
-	 * Allow direct dispatch when @local_pcpu is enabled, or when there
-	 * are no tasks queued in the node DSQ.
+	 * If @local_pcpu is enabled allow direct dispatch only if there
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
This seems like a decent trade off and can be disabled as needed. LGTM

PR #2126, opened 2025-06-10
lib/bitmap: adding test_and_clear helper method

yaakov-stein:
Since we call the test and clear methods in quick succession generally, this adds a helper method (similar to bpf_cpumask_test_and_clear_cpu) that makes it one call (cleaner to call) with slightly less operations. Logic stays exactly the same as the original test and clear methods.


lib/bitmap.bpf.c
```diff
@@ -89,6 +89,16 @@ bool scx_bitmap_test_cpu(u32 cpu, scx_bitmap_t __arg_arena mask)
 	return mask->bits[cpu / 64] & (1ULL << (cpu % 64));
 }
 
+__weak
+bool scx_bitmap_test_and_clear_cpu(u32 cpu, scx_bitmap_t __arg_arena mask)
```
htejun, line 93: This may be a bit misleading as people would expect `test_and_*` operations to be atomic. FWIW, in kernel, we use `__` prefix for operations that are explicitly non-atomic.

yaakov-stein: Agreed, thanks for the catch - I'll update it to be atomic.

yaakov-stein: It looks like arenas causes the verifier to have issues with __sync_fetch_and_and, so for now, I'll just update the name and I can discuss with @etsal to figure out what the best option here is.

lib/bitmap.bpf.c
```diff
@@ -90,13 +90,19 @@ bool scx_bitmap_test_cpu(u32 cpu, scx_bitmap_t __arg_arena mask)
...
 __weak
-bool __scx_bitmap_test_and_clear_cpu(u32 cpu, scx_bitmap_t __arg_arena mask)
+bool scx_bitmap_test_and_clear_cpu(u32 cpu, scx_bitmap_t __arg_arena mask)
 {
 	u64 bit = 1ULL << (cpu % 64);
 	u32 idx = cpu / 64;
-	bool was_set = mask->bits[idx] & bit;
-	mask->bits[idx] &= ~bit;
-	return was_set;
+	u64 old = mask->bits[idx];
```
etsal: Can we use a loop here? AFAICT there is no reason to expose the fact the call can fail to the user.

PR #2129, opened 2025-06-11
scx_layered: Collect layer LLC DSQ enqueues statistics

kkdwivedi:
This is useful to understand the ratio of idle selections and runqueue insertions, and to understand the effect of optimizations which free up more idle CPUs for higher priority tasks from the statistics.


hodgesds (approved):
Makes sense

PR #2133, opened 2025-06-11
scx_flash: Always update task domain before proactive wakeup

arighi:
After a task is enqueued to the per-node DSQ we proactively wakeup an idle CPU usable by the task, potentially giving a quicker opportunity for the task to be consumed.

Like any idle CPU selection, we take into account the CPU previously used by the task as a preference. However, in this case, the task's domain might be not updated if the task has bounced across different CPUs without going through ops.select_cpu(), that is the only place where we refresh the preferred domains.

Therefore, make sure to refresh the task's preferred domains before applying the proactive wakeup logic.


multics69 (approved):
LGTM.

PR #2134, opened 2025-06-11
scx_utils, scx_lavd: Improve detection of heterogeneous architectures (big/LITTLE).

multics69:
Previously, if all CPU capacities are the same, the processor architecture is considered homogeneous. If they vary, a core type is classified as big or LITTLE if the CPU's capacity is above or below the average capacity.

However, such an approach could be incorrect because CPU capacities can still vary in a homogeneous architecture—for instance, due to chip binning or when only a subset of CPUs supports turbo boost.

Thus, we consider a system to have a heterogeneous CPU architecture only when there is a significant capacity gap (e.g., 1.5x). 

This PR first improves the big/LITTLE detection in Topology, then changes the `scx_lavd` to rely on it.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -1499,10 +1499,9 @@ static s32 init_per_cpu_ctx(u64 now)
 	struct bpf_cpumask *turbo, *big, *little, *active, *ovrflw, *cd_cpumask;
 	const struct cpumask *online_cpumask;
 	struct cpdom_ctx *cpdomc;
-	int cpu, i, j, err = 0, nr_cpus_non_zero = 0;
+	int cpu, i, j, err = 0;
 	u64 cpdom_id;
-	u32 sum_capacity = 0, avg_capacity, big_capacity = 0;
-	u16 turbo_cap;
+	u32 sum_capacity = 0, big_capacity = 0;
 
 	bpf_rcu_read_lock();
 	online_cpumask = scx_bpf_get_online_cpumask();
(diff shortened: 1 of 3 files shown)


rust/scx_utils/src/topology.rs
```diff
@@ -580,7 +579,18 @@ fn read_cpu_ids() -> Result<Vec<usize>> {
...
-fn cpu_capacity_source() -> Option<(String, usize, usize)> {
+struct CapacitySource {
+    /// Path suffix after /sys/devices/system/cpu/cpuX
+    suffix: String,
+    /// Average raw capacity value
+    avg_rcap: usize,
+    /// Maximum raw capacity value
+    max_rcap: usize,
+    /// Does a system have little cores?
+    het_arch: bool,
```
arighi: What do you think about renaming this to something like "has_hybrid" or "has_biglittle"? I struggled a bit to realize that 'het' stands for heterogeneous. :)

multics69: That's a good point. I will change it to `has_biglittle`.

multics69:
> I wish there was a better systematic way to determine the presence of hybrid cores

I agree with that. 

I also expected and wanted a more systematic approach, but I found it to be challenging. I checked various things, including cluster_id, CPU model, and the minimum and maximum CPU frequencies. However, I failed to find more direct info that is compatible across Intel, AMD, and ARM SoCs. It appears that 1.5x is reasonable for at least all my tested processors, including Intel, AMD, and ARM SoCs, allowing it to identify big/LITTLE configurations correctly and also disregard the capacity difference due to chip binning.

rust/scx_utils/src/topology.rs
```diff
@@ -617,27 +627,43 @@ fn cpu_capacity_source() -> Option<(String, usize, usize)> {
...
+        suffix = "";
+        avg_rcap = 1024;
+        max_rcap = 1024;
+    } else {
+        avg_rcap /= nr_cpus;
+        // We consider a system to have a heterogeneous CPU architecture only
+        // when there is a significant capacity gap (e.g., 1.5x). CPU capacities
+        // can still vary in a homogeneous architecture—for instance, due to
+        // chip binning or when only a subset of CPUs supports turbo boost.
+        has_biglittle = max_rcap as f32 >= (1.5 * min_rcap as f32);
```
hodgesds, line 658: My only big little CPU would fail this check https://www.intel.com/content/www/us/en/products/sku/226257/intel-core-i31220p-processor-12m-cache-up-to-4-40-ghz/specifications.html 

3.3 * 1.5 = 4.95 > 4.4 (CPU max)

multics69: Thanks, @hodgesds, for the info. I will change it to 1.3, which covers your case and all my cases.

PR #2137, opened 2025-06-11
scx_utils: Tune the capacity gap threshold for correct detection of big/LITTLE.

multics69:
Some big/LITTLE architectures have less capacity gap [1], so tune thedifference threshold from 1.5x to 1.3x to cover such case.

[1] https://www.intel.com/content/www/us/en/products/sku/226257/intel-core-i31220p-processor-12m-cache-up-to-4-40-ghz/specifications.html


hodgesds (approved):
LGTM

PR #2139, opened 2025-06-11
scx_flash: Prevent potential ops.enqueue() stall

arighi:
When dispatching directly to an idle CPU using SCX_DSQ_LOCAL[_ON] the sched_ext core should take care of waking up the target CPU automatically.

However, this assumption is only valid when the dispatch is happening in ops.select_cpu(). Tasks directly dispatched from ops.enqueue() can trigger stalls like the following:
```
CPU states
----------

CPU 3   : nr_run=1 flags=0x1 cpu_rel=0 ops_qseq=1726682 pnt_seq=51006
          curr=swapper/3[0] class=idle_sched_class

  R scx_flash[35365] -5056ms
      sticky/holding_cpu=-1/-1 dsq_id=0x8000000000000002
      dsq_vtime=1302214 slice=20000000 weight=100
      cpus=ffffff

    do_syscall_64+0x7b/0x830
```
From the trace we can see a task dispatched to the local DSQ (dsq_id=0x8000000000000002), but the corresponding CPU was idle (class=idle_sched_class), meaning that the target CPU needed to be kicked.

Prevent this by always calling scx_bpf_kick_cpu() when dispatching directly from ops.enqueue().


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1052,25 +1051,27 @@ static bool try_direct_dispatch(struct task_struct *p, struct task_ctx *tctx,
...
-	 * the task we want it to either be re-added to the idle pool or
-	 * consume another task from the per-node DSQ.
+	 * Kick the CPU even if we didn't directly dispatch, so that it can
+	 * be re-added to the idle pool or consume another task from the
+	 * per-node DSQ.
 	 */
+	is_idle = false;
+
+out_kick:
 	scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
```
hodgesds, line 1069: should this check:
```
if (is_idle):
```

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1012,19 +1013,16 @@ static bool try_direct_dispatch(struct task_struct *p, struct task_ctx *tctx,
...
-			return true;
+			is_idle = true;
 		}
 
 		/*
 		 * No need to check for other CPUs if the task can only run
 		 * on a single one.
 		 */
-		cpu = prev_cpu;
 		goto out_kick;
```
hodgesds, line 1026: See comment below, I think this might kick if the affinitized CPU is not idle which maybe is ok.

hodgesds (approved):

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: Yeah this part is a bit tricky to follow, maybe I'll rework the code a bit to make it more clear.

Basically is_idle is used at the end to determine if the direct dispatch happened or not. But even if the direct dispatch didn't happen we still want to kick the CPU, because it's idle-locked now and we want it to transition from idle->awake->idle to clear its idle state  (it's explained a bit in the comment above).

arighi:
@hodgesds code updated, it should easier to understand now. Thanks!

PR #2140, opened 2025-06-11
lib/cpumask: use cmpxchg in scx_bitmap_vacate_cpu

etsal:
Atomics are not allowed on arena memory. Replace the sole instance with cmpxchg to pass verification.

### DIFF
--- a/lib/cpumask.bpf.c
+++ b/lib/cpumask.bpf.c
@@ -71,6 +71,7 @@ scx_bitmap_vacate_cpu(scx_bitmap_t __arg_arena mask, s32 cpu)
 {
 	int off = (u32)cpu / 64;
 	int ind = (u32)cpu % 64;
+	u64 old, new;
 
 	if (cpu < 0 || cpu >= nr_cpu_ids) {
 		bpf_printk("freeing invalid cpu");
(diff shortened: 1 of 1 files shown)


etsal:
This is dependent on the other commit that removes scx_bpf_error so we will need to merge that first.

hodgesds (approved):
One thing that might be nice is like a config struct to control the printk path and also having a prefix on the printk's for filtering:
```
		bpf_printk("ARENA scx_task_init failed with %d", ret);
```
That way schedulers running in debug mode could also print arena debugging when needed, can be done later though.

yaakov-stein (commented):
MIssing context on this, but shouldn't vacate clear the bit? Also, what additional functionality does this add beyond scx_bitmap_clear_cpu/scx_bitmap_set_cpu?

etsal:
> MIssing context on this, but shouldn't vacate clear the bit? Also, what additional functionality does this add beyond scx_bitmap_clear_cpu/scx_bitmap_set_cpu?

The method is called vacate because it leaves the cpu free to run something else, the reason it uses or is because unless I'm mistaken the idle cpu mask marks with a set bit the CPUs that are available for running tasks (EDIT: The cpumask methods we do here definitely do, the "unless I'm mistaken" part is for the scx idle mask. If that mask behaves differently we should change our code to have the same behavior). When we leave a cpu idle we set its bit back to 1 in the cpumask. 

The difference with the existing bitmap operations is basically that this one is thread safe, because we use cmpxchg to set the bit. This way there is no potential race between two threads setting different bits in the same word simultaneously and causing one of the updates to be lost.

etsal:
> One thing that might be nice is like a config struct to control the printk path and also having a prefix on the printk's for filtering:
> 
> ```
> 		bpf_printk("ARENA scx_task_init failed with %d", ret);
> ```
> 
> That way schedulers running in debug mode could also print arena debugging when needed, can be done later though.

Will do, since we cannot use scx_bpf_error that gives us a trace I will wrap the printk's in a macro to make them easier to track down.

yaakov-stein (approved):

PR #2151, opened 2025-06-12
scx_p2dq: Optimize sync wakeup path

hodgesds:
Remove previously done checks on the wakeup path and avoid extra map lookups for interactive tasks.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -437,8 +437,7 @@ static s32 pick_idle_cpu(struct task_struct *p, task_ctx *taskc,
 	}
 
 	// First check if last CPU is idle
-	if (taskc->all_cpus &&
-	    bpf_cpumask_test_cpu(prev_cpu, (smt_enabled && !interactive) ?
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -460,22 +459,8 @@ static s32 pick_idle_cpu(struct task_struct *p, task_ctx *taskc,
...
-		// Shouldn't happen, but makes code easier to follow
-		if (!waker_taskc) {
-			cpu = prev_cpu;
-			goto found_cpu;
-		}
-
 		// Interactive tasks aren't worth migrating across LLCs.
 		if (interactive) {
-			cpu = prev_cpu;
-			if (scx_bpf_test_and_clear_cpu_idle(cpu)) {
```
yaakov-stein, line 474: Wouldn't this check be worth keeping? If possible, keeping the interactive task sticky to the CPU is most ideal, no?

yaakov-stein (commented):
LGTM, added a question about one check removal

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: It's already covered in 439, so this check is redundant.

PR #2152, opened 2025-06-12
scx_p2dq: Special case dual socket wakeups

hodgesds:
On systems that have two LLCs and two NUMA nodes keep wakeups sticky to their LLC. This provides better bounds on wakeup latency.


JakeHillion (approved):
LGTM.

PR #2154, opened 2025-06-12
scx_chaos: Refactor stats_thread initialization using map()

EricccTaiwan:
Replaced the if-let-else construct with Option::map() to simplify the creation of the stats_thread.

No functional changes.


JakeHillion (approved):
thanks!

PR #2160, opened 2025-06-13
scx_flash: Introduce "turbo" mode

arighi:
Allow to specify "turbo" with the option --primary-domain to prioritize the CPUs with the highest max frequency.

This differs from the "performance" mode, which prioritizes the fast cores in systems with hybrid cores.

The "turbo" option can be beneficial with certain workloads in systems with turbo-boosted cores.


hodgesds (approved):
LGTM

PR #2162, opened 2025-06-13
scx_mitosis: Remove unused field and helper to suppress warnings

EricccTaiwan:
- Remove the unused `prev_percpu_cell_cycles` field from the `Scheduler` struct, which was never read and triggered a dead code warning.
- Delete the unused unsafe helper function `any_as_u8_slice()`, as it was never called and also produced a dead_code warning.
- Replace `return bail!(...)` with `bail!(...)` to avoid an unreachable code warning since the macro already expands to a return.

Log
```sh
warning: unreachable expression
   --> scheds/rust/scx_mitosis/src/main.rs:296:13
    |
296 |             return bail!("Error: No queueing decisions made globally");
    |             ^^^^^^^---------------------------------------------------
    |             |      |
    |             |      any code following this expression is unreachable
    |             unreachable expression
    |
    = note: `#[warn(unreachable_code)]` on by default

warning: function `any_as_u8_slice` is never used
  --> scheds/rust/scx_mitosis/src/main.rs:84:11
   |
84 | unsafe fn any_as_u8_slice<T: Sized>(p: &T) -> &[u8] {
   |           ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field `prev_percpu_cell_cycles` is never read
  --> scheds/rust/scx_mitosis/src/main.rs:98:5
   |
96 | struct Scheduler<'a> {
   |        --------- field in this struct
97 |     skel: BpfSkel<'a>,
98 |     prev_percpu_cell_cycles: Vec<[u64; MAX_CELLS]>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^

warning: `scx_mitosis` (bin "scx_mitosis") generated 3 warnings
```

### DIFF
--- a/scheds/rust/scx_mitosis/src/main.rs
+++ b/scheds/rust/scx_mitosis/src/main.rs
@@ -34,7 +34,6 @@ use scx_utils::Cpumask;
 use scx_utils::Topology;
 use scx_utils::UserExitInfo;
 use scx_utils::NR_CPUS_POSSIBLE;
-use scx_utils::NR_CPU_IDS;
 
 const MAX_CELLS: usize = bpf_intf::consts_MAX_CELLS as usize;
 const NR_CSTATS: usize = bpf_intf::cell_stat_idx_NR_CSTATS as usize;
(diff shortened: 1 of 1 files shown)


dforsyth:
lftm

EricccTaiwan:
@dschatzberg PTAL, thanks!

dschatzberg:
Thanks, all these look good to me. Is there some way we can make CI flag warnings on PRs? That might be the easiest way to avoid you having to clean up after others.

EricccTaiwan:
> Thanks, all these look good to me. Is there some way we can make CI flag warnings on PRs? That might be the easiest way to avoid you having to clean up after others.

Thanks, I’ll keep it in mind. I’ll try to look into it in about three weeks (got a thesis oral coming up) .
Not too familiar with the CI setup yet, but I’ll give it a shot!

PR #2164, opened 2025-06-13
lib/sdt_alloc: Fix inverted null check in free path

EricccTaiwan:
Fix a logic bug where a null pointer could be dereferenced during freeing due to an inverted condition.


EricccTaiwan:
I think I misunderstood smth. Close it now.

PR #2165, opened 2025-06-13
scx_layered: Synchronize among competing preemptors

htejun:
Multiple preemptors can race and pick the same victim CPU and end up queueing multiple tasks on the same local DSQ severely impacting scheduling latencies.

While preemption, as implemented in scx_layered, has always been racy, c66e2c374904 ("scx_layered: Reimplement preemption using direct dispatch to foreign local DSQs") made it significantly worse. Before, the race would just fail to kick enough CPUs but the preempting task would still be picked up relatively quickly as CPUs enter dispatch path after finishing their current tasks. After, the preempting tasks end up queued on a specific local DSQ.

Fix it by using cpu_ctx->preempting_task to synchronize preemptors. The field is claimed in try_preempt_cpu() and cleared from running() when the task goes on the CPU. There are cases that the task may not go on the victim CPU even after being enqueued on its local DSQ - e.g. when the task gets reenqueued due to cpu_release(). Clear cpu_ctx->preempting_task from any task to avoid blocking preemption on the CPU for too long.

Direct dispatch, regular dispatch and preemption can still race each other after this and we likely need kernel side update to remove races completely. For now, add opportunistic tests to significantly reduce incidence rate.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1229,6 +1229,7 @@ static bool try_preempt_cpu(s32 cand, struct task_struct *p, struct task_ctx *ta
 	struct rq *rq;
 	struct task_struct *curr;
 	const struct cpumask *idle_cpumask;
+	bool cand_idle;
 	s32 sib;
 
 	if (cand >= nr_possible_cpus || !bpf_cpumask_test_cpu(cand, p->cpus_ptr))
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1237,7 +1238,19 @@ static bool try_preempt_cpu(s32 cand, struct task_struct *p, struct task_ctx *ta
...
+	cand_idle = scx_bpf_test_and_clear_cpu_idle(cand);
+	if (cand_idle)
+		goto preempt;
+
+	/*
+	 * This racy check helps with reducing the incidence rate of preempting
+	 * a CPU which already has tasks queued in its local DSQ. See the
+	 * comment above preempting_task check.
+	 */
+	if (scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | cand))
```
likewhatevs, line 1253: this might need to account for skip rnode I think

htejun: Yeah, selcpu/enq paths need to be updated in general. I'm not sure whether expanding skip_rnode is a good idea tho. We need to split internal data structure per node and implement load balancing across them.

PR #2166, opened 2025-06-14
scx_p2dq: Add optimization for pick2

hodgesds:
When there are only two LLCs there is no need to randomly pick two. This improves overall LLC selection and improves performance.


htejun (approved):

likewhatevs (approved):
make cpu go melt

PR #2169, opened 2025-06-14
scx_p2dq: Update handling of affinitized kthreads

hodgesds:
Only do direct dispatch of affinitized kthreads if the kthread is bound to a single CPU and ensure to kick the cpu if the cpu is idle.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -637,12 +637,12 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
 	 * into the local DSQ.
 	 */
 	if ((p->flags & PF_KTHREAD) &&
-	    p->cpus_ptr == &p->cpus_mask &&
-	    p->nr_cpus_allowed != nr_cpus &&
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -637,12 +637,13 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
 	 * into the local DSQ.
 	 */
 	if ((p->flags & PF_KTHREAD) &&
-	    p->cpus_ptr == &p->cpus_mask &&
-	    p->nr_cpus_allowed != nr_cpus &&
+	    p->nr_cpus_allowed == 1 &&
 	    bpf_cpumask_test_cpu(cpu, p->cpus_ptr) &&
 	    kthreads_local) {
 		stat_inc(P2DQ_STAT_DIRECT);
 		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON|cpu, dsq_time_slices[0], enq_flags);
```
arighi: Considering that per-cpu kthreads are basically never migrating, I was wondering if we could just use SCX_LOCAL_DSQ and get rid of the bpf_cpumask_test_cpu().

PR #2176, opened 2025-06-16
scx_lavd: Refactor FlatTopology.

multics69:
As a first step toward energy model-aware CPU ordering, let's refactor FlatTopology by separating it into a new module, cpu_order.rs. As the name 'FlatTopology' no longer makes sense, let's rename it to 'CpuOrder'. Additionally, clean up struct names for consistency (e.g., CpuFlatId -> CpuId, etc.).

Then, let's separate the implementation details from CpuOrder, so that struct CpuOrderCtx can guiltlessly carry implementation-dependent fields.

There are no functional changes.


hodgesds (approved):
Makes sense

PR #2179, opened 2025-06-16
lib: add minheap data structure

etsal:


lib/minheap.bpf.c
```diff
@@ -0,0 +1,82 @@
...
+int scx_minheap_pop(scx_minheap_t *heap, struct scx_minheap_elem *helem)
+{
+	if (heap->size == 0)
+		return -EINVAL;
+
+	*helem = heap->helems[0];
+	heap->helems[0] = heap->helems[heap->size - 1];
+	heap->size -= 1;
+
+	scx_minheap_balance(heap, 0);
```
yaakov-stein: When the index here is set to 0, doesn't that break the balance code as parent will be heap->helems[-1/2] and it will just check against itself? Wouldn't we want to "swim down" in this case?

etsal: Yeah something is borked here, I need to fix this and another bug I saw. Taking a step back, we also need selftests for the lib/ code because bugs are super easy to slip in right now since we write it before actually using it in the schedulers.

JakeHillion: The only BPF specific in here is `scx_static_alloc` right, and it gets statically linked? Should be really easy to stub out and run in an ordinary userspace C testing library.

etsal: We could do that for this struct, but for a lot of other arena code we use kfuncs/BPF-specific code so I think we can keep testing similar to the kernel BPF selftests for now. We can always add userspace testing/fuzzing later.

hodgesds: I think using a test prog should be sufficient for testing this and should have enough instructions.

lib/minheap.bpf.c
```diff
@@ -0,0 +1,83 @@
...
+
+	while (ind >= 0 && can_loop) {
+		parent = (ind - 1) / 2;
+
+		if (heap->helems[parent].weight <= heap->helems[ind].weight)
+			break;
+
+		htmp = heap->helems[parent];
+		heap->helems[parent] = heap->helems[ind];
+		heap->helems[ind] = htmp;
```
EricccTaiwan: Missing `ind = parent` here?

lib/selftests/selftest.h
```diff
@@ -1,3 +1,13 @@
 #pragma once
 
+#define SCX_SELFTEST(func, ...)		\
```
hodgesds, line 3: nice!

hodgesds (approved):

lib/minheap.bpf.c
```diff
@@ -0,0 +1,117 @@
...
+static inline
+int scx_minheap_balance_bottom_up(void __arena *heap_ptr __arg_arena)
+{
+	scx_minheap_t *heap = (scx_minheap_t *)heap_ptr;
+	u64 elem, weight;
+	int parent;
+	int ind;
+
+	for (ind = heap->size - 1; ind > 0 && can_loop; ind = parent) {
+		parent = ind % 2 ? ind / 2 : ind / 2 - 1;
```
tavianator, line 82: `(ind - 1) / 2`?

lib/minheap.bpf.c
```diff
@@ -0,0 +1,117 @@
...
+		if (next == ind)
+			break;
+
+		elem = heap->elems[next];
+		weight = heap->weights[next];
+
+		heap->elems[next] = heap->elems[ind];
+		heap->weights[next] = heap->weights[ind];
+
+		heap->elems[ind] = elem;
```
tavianator, line 66: I think you can just do this once at the end of the loop

lib/minheap.bpf.c
etsal: Should be the same when rounding down an even, right?

lib/minheap.bpf.c
etsal: Don't we need it for every level? We pick the lowest child in the inner loop, so there is at most one swap per level

tavianator: What I mean is instead of doing this:

```
   1           7           2           2
 2   3  -->  2   3  -->  7   3  -->  4   3
4 5 6 7     4 5 6       4 5 6       7 5 6
```

you can do this

```
  [1]            2              2            2
 2   3  -->  [2]   3   -->    4   3   -->  4   3
4 5 6 7     4   5 6 7      [4] 5 6 7      7 5 6
```

etsal: I think I see, basically avoid placing the last element at the top then tumbling it down, instead  creating a gap and bubbling the parent up. This avoids having to swap instead of just copying over AFAICT, correct?

PR #2180, opened 2025-06-16
scx_flash: Introduce --cpu-runqueue

arighi:
Introduce the new option --cpu-runqueue to use distinct per-CPU DSQs. This makes the scheduler less aggressive with migrations and it can help improve certain cache-sensitive workloads.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -5,6 +5,8 @@
 #include <scx/common.bpf.h>
 #include "intf.h"
 
+#define MAX_VTIME	(~0ULL)
+
 const volatile u64 __COMPAT_SCX_PICK_IDLE_IN_NODE;
 
 char _license[] SEC("license") = "GPL";
@@ -20,6 +22,9 @@ const volatile bool debug;
 
 /* Enable round-robin mode */
 const volatile bool rr_sched;
+
+/* Use per-CPU DSQs */
+const volatile bool pcpu_dsq;
 
 /*
  * Default task time slice.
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1216,9 +1230,39 @@ void BPF_STRUCT_OPS(flash_dispatch, s32 cpu, struct task_struct *prev)
...
+		s32 other_cpu;
+
+		bpf_for(other_cpu, 0, nr_cpu_ids) {
+			const struct task_struct *p;
+
+			if (cpu == other_cpu || !cpus_share_llc(cpu, other_cpu))
+				continue;
+
+			dsq_id = cpu_to_dsq(other_cpu);
+			bpf_for_each(scx_dsq, p, dsq_id, 0) {
```
hodgesds, line 1252: Does this perform well? I've tried similar code for p2dq but it seemed to hurt performance.

arighi: I've tried to implement this using both the BPF iterator and by storing min_vtime in the cpu_ctx. In terms of performance it was almost the same, so I decided to go with the iterators, since it's easier and I can get rid of the logic to update the min_vtime on each CPU.

hodgesds: I just did some similar testing and I think it's pretty similar as well. Will update p2dq, thanks for testing!

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1216,9 +1230,39 @@ void BPF_STRUCT_OPS(flash_dispatch, s32 cpu, struct task_struct *prev)
...
 	 * local CPU DSQ.
 	 */
-	if (scx_bpf_dsq_move_to_local(node))
+	if (scx_bpf_dsq_move_to_local(dsq_id))
 		return;
 
+	/*
+	 * If per-CPU DSQs are enabled attempt to consume the task with the
+	 * minimum vruntime within the same LLC.
+	 */
```
multics69, line 1239: @arighi -- Do you see noticible benefit using per-CPU DSQ compared to  a wise select_cpu policy (e.g., preferring prev CPU(?

arighi: @multics69 with certain workloads I see significant benefits, `ops.select_cpu() `isn't enough because it's involved only during task wakeup. For tasks that are always running only `ops.enqueue()` gets involved and using a global DSQ (or per-node DSQ) is great for work conservation, but can also introduce many migrations, since tasks are consumed by the first CPU that becomes available.

On the other hand, the downside of the per-CPU approach is that on saturation it's more difficult to do an effective load balancing and the system becomes less responsive compared to using a global DSQ. The ideal solution would be to use per-CPU DSQs and have a good load balancer capable of distributing the load effectively when the system is saturated.

multics69: Thanks for the clarification. That makes sense.

PR #2182, opened 2025-06-16
scx_p2dq: Update dispatch DSQ selection to use DSQ iterators

hodgesds:
Use a DSQ iterator to find the min vtime across DSQs. This makes DSQ selection more fair during dispatch and removes extra DSQ tracking code. This uses a similar idea as #2180

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -808,29 +808,6 @@ static __always_inline int p2dq_running_impl(struct task_struct *p)
 		__sync_val_compare_and_swap(&llcx->vtime, llcx->vtime, p->scx.dsq_vtime);
 	}
 
-	// For non affinitized tasks update the vtime if it is larger than the
-	// current LLC vtime. Affinitized tasks are direct dispatched and don't
-	// strictly follow vtime.
-	if ((taskc->dsq_index >= 0 && taskc->dsq_index < nr_dsqs_per_llc) &&
-	    taskc->all_cpus &&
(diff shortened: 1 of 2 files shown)


JakeHillion (approved):
LGTM.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -1105,18 +1076,26 @@ static __always_inline void p2dq_dispatch_impl(s32 cpu, struct task_struct *prev
...
-		if (llcx->dsq_max_vtime[i] < min_vtime) {
-			min_vtime = llcx->dsq_max_vtime[i];
-			dsq_id = llcx->dsqs[i];
+		cur_dsq_id = llcx->dsqs[i];
+		bpf_for_each(scx_dsq, p, cur_dsq_id, 0) {
+			if (p->scx.dsq_vtime < min_vtime) {
+				min_vtime = p->scx.dsq_vtime;
+				dsq_id = cur_dsq_id;
+			}
+			break;
```
htejun, line 1085: This seems to be a common-ish usage. We probably should add `scx_bpf_dsq_peek()` so that we can avoid the two locking cycles with the iterator. It should be possible to do this locklessly, I think.

htejun (approved):

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: I started poking around at that this weekend, I'll take a look at it more this week. The other approach I've discussed with @etsal is trying an arena based DSQ.

htejun: Created https://github.com/sched-ext/scx/issues/2184.

PR #2185, opened 2025-06-16
scxtop: add generic bpf kprobe handler

yaakov-stein:
This is the first part of a multi-stage process that will eventually enable a scxtop user to select an arbitrary kprobe and view usage statistics/data on it in the terminal UI or in a perfetto trace.

This code adds the BPF generic handler that stores some basic data (more can be added as we see what is helpful) and also adds the Action in lib.rs so we can easily use the data in a trace or in the UI.

### DIFF
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -204,6 +204,7 @@ int generic_kprobe(struct pt_regs *ctx)
 	event->cpu = bpf_get_smp_processor_id();
 	event->ts = bpf_ktime_get_ns();
 	event->event.kprobe.pid = bpf_get_current_pid_tgid() & 0xffffffff;
+	event->event.kprobe.instruction_pointer = PT_REGS_IP(ctx);
 
 	bpf_ringbuf_submit(event, 0);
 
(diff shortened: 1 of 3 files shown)


tools/scxtop/src/bpf/main.bpf.c
```diff
@@ -189,6 +189,27 @@ static int update_task_ctx(struct task_struct *p, u64 dsq, u64 vtime, u64 slice_
...
+	if (!enable_bpf_events)
+		return 0;
+
+	if (!(event = try_reserve_event()))
+		return -ENOMEM;
+
+	event->type = GENERIC_KPROBE;
+	event->cpu = bpf_get_smp_processor_id();
+	event->ts = bpf_ktime_get_ns();
+	event->event.kprobe.pid = bpf_get_current_pid_tgid() & 0xffffffff;
```
hodgesds, line 206: One thing to maybe add here is the address of the probe, see [`bpf_get_func_ip`](https://man7.org/linux/man-pages/man7/bpf-helpers.7.html) docs.

etsal (approved):
This looks good, small nit: maybe drop the GENERIC_/generic_ prefix from the name? Sure we may use kprobes for other event types too, but I think it would be terser.

hodgesds (approved):

PR #2188, opened 2025-06-16
scxtop: fix number overflow errors

yaakov-stein:
Previously, scxtop's dsq_slice_consumed overflowed and the logic needed to be slightly updated to accurately get the data for both the beginning of a slice and the end of a slice. One of the bugs was that we randomly sampled events, often missing the tail-end of a slice. Another bug included incorrect operand ordering for a subtraction.

To ensure we don't miss the tail-end of a slice, we always check if the previous task is the tail end of a slice and perform all necessary calculations if so, regardless of whether we are currently sampling. After that, we determine if we should sample the incoming task based on the sample rate.

Before:
Screenshot 2025-06-16 at 5 13 41 PM

After:
Screenshot 2025-06-16 at 5 11 36 PM

(Note: There is still a bug with dsq 0 for dsq_lat_us.)

### DIFF
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -518,89 +518,89 @@ int BPF_PROG(on_sched_switch, bool preempt, struct task_struct *prev,
 	struct task_ctx *next_tctx, *prev_tctx;
 	struct bpf_event *event;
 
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
Thanks for cleaning this up

PR #2190, opened 2025-06-17
scx_p2dq: Use SCX_DSQ_LOCAL for select_cpu

hodgesds:
Use SCX_DSQ_LOCAL when dispatching to local CPU in select_cpu.


yaakov-stein:
Just wondering, based on [scx_bpf_dsq_insert](https://docs.ebpf.io/linux/kfuncs/scx_bpf_dsq_insert/) docs, aren't these the same thing (because we are selecting this cpu in select_cpu either way)?

yaakov-stein (approved):

PR #1992, opened 2025-05-26
scx_lavd: Fix CPU topology sorting for ARM big.LITTLE systems

CosmicBulb:
## Description
Fix CPU topology sorting in performance mode for ARM big.LITTLE systems with 3 cluster types.

## Problem
On ARM big.LITTLE systems with shared LLC (like Google Pixel 6), the original sorting logic prioritized physical CPU position (`cpu_pos`) over performance capacity (`cpu_cap`), causing latency-critical tasks to be scheduled on low-performance cores.

### Current Behavior (Incorrect)
```
CPU pref order in performance mode: [2, 3, 0, 1, 4, 5, 6, 7]
```
- Little cores (160 capacity) get priority over prime cores (1024 capacity)

### Expected Behavior (Fixed)
```
CPU pref order in performance mode: [6, 7, 4, 5, 0, 1, 2, 3]  
```
- Prime cores (1024) → Big cores (498) → Little cores (160)

## Solution
Reorder the sorting criteria in performance mode to prioritize:
1. NUMA node locality (`node_id`)
2. LLC locality (`llc_pos`) 
3. **CPU performance capacity (`cpu_cap`)** ← Key fix
4. Physical position (`cpu_pos`) as tie-breaker

## Testing
Verified on Google Pixel 6 (Google Tensor SoC) with TikTok performance testing using scx_lavd autopilot mode:
### Performance Comparison
| Metric | Before Fix | After Fix | Improvement |
|--------|------------|-----------|-------------|
| **Average FPS** | **18.25 FPS** | **18.65 FPS** | **+2.2%** |
| Jank | 43.5-57.1 | 4.9-31.6 | **~40-85% reduction** |

## Impact
- Fixes latency-critical task scheduling on ARM big.LITTLE systems
- Improves interactive performance on mobile devices

Fixes #1991

PR #2192, opened 2025-06-17
scx_utils: Add an environment variable, SCX_SYSFS_PREFIX, for ease of debugging.

multics69:
Currently, Topology and  EnergyModel are created by reading information
under sysfs. The sysfs path names, like "/sys/devices/system/cpu/online"
or "sys/kernel/debug/energy_model" are hard-coded in the code. While there
is nothing wrong with this, it is inconvenient to debug and test the related
code, especially when access to the environment is not possible [1].

To mitigate such a problem for ease of debugging, an environment variable,
SCX_SYSFS_PREFIX, is added, denoting the prefix of the root path, which is
an empty string, "", when SCX_SYSFS_PREFIX is not defined. topology.rs and
energy_model.rs are updated to read relevant data from
$SCX_SYSFS_PREFIX + "/sys/...".

For example, if the SCX_SYSFS_PREFIX is modified as follows and the relevant
sysfs files are placed under "/pixel6", Topology and EnergyModel can be
tested easily.

    sudo  SCX_SYSFS_PREFIX="/pixel6" scx_lavd

[1] https://github.com/sched-ext/scx/pull/1992

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -7,6 +7,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use libbpf_rs::libbpf_sys::*;
 use libbpf_rs::{AsRawLibbpf, OpenProgramImpl};
 use log::warn;
+use std::env;
 use std::ffi::c_void;
 use std::ffi::CStr;
 use std::ffi::CString;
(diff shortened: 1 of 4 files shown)


rust/scx_utils/src/compat.rs
```diff
@@ -19,6 +19,7 @@ use std::slice::from_raw_parts;
 const PROCFS_MOUNTS: &str = "/proc/mounts";
 const TRACEFS: &str = "tracefs";
 const DEBUGFS: &str = "debugfs";
+pub const ROOT_PREFIX: &str = "";
```
htejun: Maybe make this read from an env var?

htejun: And name the env var something like `SCX_SYSFS_PREFIX`?

multics69: Sure, I will change it to use the env variable.

multics69: The fix was force-pushed.

PR #2194, opened 2025-06-17
scx_p2dq: Optimize dispatch_cpu

hodgesds:
When dispatching to a CPU check if the DSQ has any queued tasks before iterating. This saves the overhead of creating the DSQ iterator.


etsal (approved):
Minor nit, can we just return 0? No functional change, just making it explicit this is the "the queue is empty" case.

PR #2196, opened 2025-06-17
scxtop: adding sched_migrate_task tracepoint.

devnexen:
process being migrated from cpu to a new one with a given scheduler priority level.


tools/scxtop/src/bpf/intf.h
```diff
@@ -175,6 +183,7 @@ struct bpf_event {
 		struct	softirq_event softirq;
 		struct	wakeup_event wakeup;
 		struct	wakeup_event waking;
+        struct  migrate_event migrate;
```
hodgesds: nit: formatting

PR #2200, opened 2025-06-18
scxtop: fix tracing race bug

yaakov-stein:
Previously, the tracer attached several programs only after the trace had completed (even though the trace_async method contained a timer), likely due to some race condition between attaching the programs to events and the timer. This caused all events that are attached in trace_async to not show up in perfetto/the trace. Now, we attach the programs earlier on and keep an explicit timer in the run_trace method (for clarity mainly).

Before:
Screenshot 2025-06-17 at 8 46 57 PM

After (notice the additional statistics on the bottom - softirq_exit, etc.):
Screenshot 2025-06-17 at 8 42 28 PM


hodgesds (approved):
LGTM

PR #2203, opened 2025-06-18
scx_rusty: Propagate format() error

charliechiou:
The monitor callback previously ignored a fallible operation's result. Add a `?` to propagate any I/O errors handling.

### DIFF
--- a/scheds/rust/scx_rusty/src/stats.rs
+++ b/scheds/rust/scx_rusty/src/stats.rs
@@ -269,7 +269,8 @@ pub fn monitor(intv: Duration, shutdown: Arc<AtomicBool>) -> Result<()> {
                 dt.to_rfc2822(),
                 (cst.lb_at_us as i64 - cst.at_us as i64) as f64 / 1000.0
             );
-            cst.format(&mut std::io::stdout())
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_rusty/src/stats.rs
```diff
@@ -269,7 +269,8 @@ pub fn monitor(intv: Duration, shutdown: Arc<AtomicBool>) -> Result<()> {
                 dt.to_rfc2822(),
                 (cst.lb_at_us as i64 - cst.at_us as i64) as f64 / 1000.0
             );
-            cst.format(&mut std::io::stdout())
+            cst.format(&mut std::io::stdout())?;
+            Ok(())
```
htejun, line 273: The before and after are effectively the same, right? `Before, cst.format()` returns `Result` which is casted to the return `Result` type and returned. After, `?` forwards the error if the returned result is an error or the last `Ok(())` is returned. The code is more verbose after but the end results are the same.

charliechiou: Got it, thanks for the clarification!

PR #2206, opened 2025-06-18
scx_flash: Improve task re-enqueue

arighi:
If a task is re-enqueued because a higher scheduling class stole the CPU, allow it to migrate to a different CPU via direct dispatch in ops.enqueue() and skip updating its deadline, since it didn't actually ran.

In all other cases, migration should only be attempted on task wakeup.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -988,15 +988,6 @@ static bool try_direct_dispatch(struct task_struct *p, struct task_ctx *tctx,
 	s32 cpu = prev_cpu;
 
 	/*
-	 * If a task has been re-enqueued because its assigned CPU has been
-	 * taken by a higher priority scheduling class, force it to follow
-	 * the regular scheduling path and give it a chance to run on a
-	 * different CPU.
-	 */
-	if (enq_flags & SCX_ENQ_REENQ)
-		return false;
-
-	/*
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1020,13 +1011,28 @@ static bool try_direct_dispatch(struct task_struct *p, struct task_ctx *tctx,
 		return false;
 
 	/*
-	 * Try direct dispatch only if ops.select_cpu() has been skipped,
-	 * since the task already had a chance to be directly dispatched
-	 * from ops.select_cpu().
+	 * Skip direct dispatch if ops.select_cpu() was already called, as
```
hodgesds, line 1014: A CPU could have gone idle from the time between the `select_cpu` and `enqueue` call. It's probably not significant though.

hodgesds (approved):

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: True, but I think most of the times doing another attempt here would produce the same result, so it'd be just overhead. It'd be like calling `ops.select_cpu()` twice, which might work sometimes, but only if we're lucky. :)

Also, if we take the enqueue path we still do a proactive wakeup, so if a CPU has gone idle, we would kick that and it can immediately consume the task.

Anyway, thanks for pointing this out, it made me think a bit if I was doing the right thing. :)

PR #2208, opened 2025-06-18
scxtop: adding kprobe events to trace

yaakov-stein:
These additions add a kprobe_event module that will be used to simplify adding kprobe event data into traces and the terminal UI. With this, an arbitrary number of kprobe_event's can be added into the trace and can be entered at the command line like this:

`sudo ./target/release/scxtop trace --kprobes do_sys_open alloc_pages -v`

After validation, the trace will now show the # of times the provided kprobes were hit:
Screenshot 2025-06-18 at 1 33 43 PM


yaakov-stein:
Just realized that these tests will likely fail on systems that aren't exactly like mine. I'll update to make them more general.

yaakov-stein:
Re: above comment - testing issue should be fixed now

tools/scxtop/src/kprobe_event.rs
```diff
@@ -0,0 +1,104 @@
...
+
+use anyhow::Result;
+use scx_utils::compat::tracefs_mount;
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+
+use crate::Search;
+
+pub struct AllKprobeEvents {
+    entries: Search,
```
yaakov-stein: After thinking about it, I don't think this is necessary. These methods can exist on the Search struct as `exists` etc. I'll remove this struct and just keep the available_kprobe_events method.

hodgesds: Yeah, that seems cleaner.

PR #2215, opened 2025-06-19
scx_p2dq: Add migration DSQs

hodgesds:
Add a migration DSQ for tasks that are eligible to be load balanced. Previously when iterating across all DSQs the time complexity is O(N) across all N tasks in the worst case. The migration eligible DSQ allows for O(logN) lookups when doing LLC migrations, where N is the number of eligible tasks for migration.

Refactor p2dq_dispatch_impl to only lookup the llc_ctx context when needed in order to reduce map lookups in the dispatch path.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -280,6 +280,13 @@ static bool is_interactive(task_ctx *taskc)
 	return taskc->dsq_index == 0;
 }
 
+static __always_inline bool can_migrate(task_ctx *taskc)
+{
+	return (nr_llcs > 1 &&
+		taskc->all_cpus &&
+		taskc->llc_runs > min_llc_runs_pick2);
+}
+
 static void set_deadline_slice(task_ctx *taskc, struct llc_ctx *llcx)
 {
 	u64 dsq_id, nr_queued = 0;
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -1125,18 +1058,30 @@ static __always_inline void p2dq_dispatch_impl(s32 cpu, struct task_struct *prev
...
 				min_vtime = p->scx.dsq_vtime;
 				dsq_id = cpuc->affn_dsq;
 			}
 			break;
 		}
 	}
 
-	trace("DISPATCH cpu[%d] cur vtime %llu min_vtime %llu dsq_id %llu",
-	      cpu, llcx->vtime, min_vtime, dsq_id);
+	// Migration eligible
```
arighi, line 1073: IIUC affinitized tasks always get a chance to run before the migration-eligible tasks. Are we potentially starving non-affinitized tasks? Have you tried to run a bunch of per-CPU hogs and see if they can starve other tasks?

hodgesds: Affinitized tasks should use the same vtime timeline so their shouldn't be a problem as the min vtime should be used. If the affinitized dsq on the cpu has the lowest vtime then it's consumed otherwise all other DSQs are consumed from first and the `affn_dsq` is consumed after. Long term I may try to get rid of the multiple LLC DSQs, but we'll see.


I tested with `stress-ng` and saw no stalls:
```
$ stress-ng -c $(($(nproc) * 2)) --affinity $(($(nproc) * 2)) -t 60 --affinity-rand --affinity-sleep 1 --affinity-pin
stress-ng: info:  [25086] setting to a 40 secs run per stressor
stress-ng: info:  [25086] dispatching hogs: 64 cpu, 64 affinity
stress-ng: info:  [25086] skipped: 0
stress-ng: info:  [25086] passed: 128: cpu (64) affinity (64)
stress-ng: info:  [25086] failed: 0
stress-ng: info:  [25086] metrics untrustworthy: 0
stress-ng: info:  [25086] successful run completed in 40.43 secs
```

arighi: Ack, thanks for clarifying and checking it!

hodgesds: I switched up the order to check the affinitized dsq first as after doing some more tests I would sometimes see stalls.

PR #2221, opened 2025-06-20
scxtop: fix event clearing bug

yaakov-stein:
Currently, pressing 'x' to clear events does not work as expected. It clears the data for the event, but leaves the event visible and present in the available_events array. This fix causes 'x' to fully clear whatever events have been added.

Also removed unused variable non_hw_event_active.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -303,6 +303,19 @@ impl<'a> App<'a> {
 
     /// Stop all active perf events.
     fn stop_perf_events(&mut self) {
+        let default_perf_event = self.config.default_perf_event();
+        let default_perf_event_parts: Vec<&str> = default_perf_event.split(':').collect();
+
+        // We have already checked that this was a valid perf event in new(), no
+        // need to check here.
+        let subsystem = default_perf_event_parts[0].to_string();
+        let event = default_perf_event_parts[1].to_string();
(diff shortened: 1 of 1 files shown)


yaakov-stein:
Needs to be fixed, but needs a bit more than just this for a proper fix.
Edit: resolved, this should be good now.

tools/scxtop/src/app.rs
```diff
@@ -302,11 +300,19 @@ impl<'a> App<'a> {
...
 
     /// Stop all active perf events.
-    fn stop_perf_events(&mut self) {
+    fn stop_perf_events(&mut self) -> Result<()> {
         for cpu_data in self.cpu_data.values_mut() {
             cpu_data.data.clear();
         }
         self.active_perf_events.clear();
+
+        self.available_events = PerfEvent::default_events();
```
hodgesds, line 313: Maybe this could be moved into a helper `reset_perf_events` as stopping seems like it should be logically separated from resetting.

hodgesds (approved):
LGTM

tools/scxtop/src/app.rs
yaakov-stein: Agreed - I'll have `stop_perf_events` clear the current data but leave it in `available_events` and `reset_perf_events` will reset everything to default.

PR #2225, opened 2025-06-22
scxtop: decouple filtering and rendering event list

yaakov-stein:
It is unnecessary and inefficient to filter the events every time we render the event list. Filtering should occur when we enter the event list and whenever input is updated. [This performance boost is necessary for when we will be filtering kprobe events (there are ~60k on my current machine and even a simple substring search takes >6ms)]. This decoupling also leads to a cleaner and more correct design.

To do this, I've added a FilteredEventState struct to the App struct. Since an input update and rendering an event list can occur simultaneously, this is stored behind a mutex.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -57,14 +57,14 @@ use scx_utils::scx_enums;
 use scx_utils::Topology;
 use serde_json::Value as JsonValue;
 use tokio::sync::mpsc::UnboundedSender;
-use tokio::sync::Mutex;
+use tokio::sync::Mutex as TokioMutex;
 
 use std::collections::BTreeMap;
 use std::os::fd::{AsFd, AsRawFd};
 use std::path::Path;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
-use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


hodgesds (approved):
LGTM, the mutex makes sense

tools/scxtop/src/app.rs
```diff
@@ -90,16 +91,14 @@ pub struct App<'a> {
...
-
     active_event: PerfEvent,
     active_hw_event_id: usize,
     active_perf_events: BTreeMap<usize, PerfEvent>,
     available_events: Vec<PerfEvent>,
     event_input_buffer: String,
     event_search: Search,
 
-    filtered_perf_events_list: Vec<String>,
+    filtered_events_state: Arc<std::sync::Mutex<FilteredEventState>>,
```
hodgesds: could maybe be a [`RwLock`](https://doc.rust-lang.org/std/sync/struct.RwLock.html) as well, but mutex seems fine.

yaakov-stein: Given we only read once (when rendering), I feel like this keeps it simpler for now and prevents the possible write starvation that can come with RwLock. But if we end up using it for more reads in the future, I'll update it to a RwLock.

PR #2226, opened 2025-06-22
scx_p2dq: Refactor DSQs

hodgesds:
Refactor DSQs so that there are no longer DSQs per slice interval. This Reduces the total number of DSQs that a CPU will check during dispatch to three. The layout of the DSQs is as follows:

LLC
DSQ - DSQ for LLC runnable tasks
MIG_DSQ- DSQ for migration eligible tasks

CPU
AFFN_DSQ- DSQ for affinitized tasks

When a CPU is dispatching it first checks the affinitized DSQ, then LLC DSQ and finally the migration eligible DSQ. All DSQs on a LLC share the same vtime priotization.


arighi (approved):
LGTM

PR #2227, opened 2025-06-22
Add cpufreq/amd_pstate_prefcore_ranking to topology

ferrreo:
Adding this give basic AMD prefcore support on sched startup. 
Doesn't regress none prefcore CPUs.


arighi (approved):
LGTM, we can refine this in tree and add the extra logic to detect frequency / cache mode variations later. But at least this should already improve things on these AMD CPUs with prefcore ranking.

Thanks!

multics69:
@ferrreo -- Thanks for the PR! As you know, the capacity code in the Topology does not re-read the capacity source when updated. I wonder when `amd_pstate_prefcore_ranking` is updated, and see if it is *critical*, not considering the update.

PR #2231, opened 2025-06-23
scx_flash: Introduce tickless mode

arighi:
Add a new `--tickless-us N` option to enable tickless mode: tasks receive an infinite time slice, and the scheduler performs periodic checks (every N microseconds) to detect CPU contention and trigger preemption only when necessary.

This mode can improve performance isolation and reduce task migrations when the system is not saturated.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -140,6 +140,11 @@ static inline void set_throttled(bool state)
 }
 
 /*
+ * Tickless CPU contention check frequency.
+ */
+const volatile u64 tickless_ns;
+
+/*
  * Exit information.
  */
 UEI_DEFINE(uei);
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1754,6 +1782,37 @@ static int throttle_timerfn(void *map, int *key, struct bpf_timer *timer)
...
+		int node = __COMPAT_scx_bpf_cpu_node(cpu);
+
+		p = scx_bpf_cpu_rq(cpu)->curr;
+		if (p->scx.slice != SCX_SLICE_INF)
+			continue;
+
+		if (scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | cpu) ||
+		    (pcpu_dsq && scx_bpf_dsq_nr_queued(cpu_to_dsq(cpu))) ||
+		    (node_dsq && scx_bpf_dsq_nr_queued(node_to_dsq(node))))
+			p->scx.slice = task_slice(cpu);
```
multics69, line 1807: Overall the PR makes sense to me. Instead of using BPF timer, I think you can check if the preemption is necessary *when* a task is enqueued.

multics69 (approved):

hodgesds (approved):

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: Yeah, I should try that too. One thing that I like about the timer is that it doesn't introduce any extra overhead in the hot path (`ops.enqueue()`). It might respond a bit more slowly since the timer callback has to iterate over all CPUs, but that shouldn’t be a major issue.

Anyway, will do some tests moving this logic to ops.enqueue(), thanks for the suggestion @multics69 !

PR #2232, opened 2025-06-23
scx_utils, scx_lavd: Introduce EnergyModelOptimizer.

multics69:
EnergyModelOptimizer generates a table of sets of CPUs to achieve a specific
performance capacity with minimal energy consumption, given the system's CPU
topology and energy model.

It mainly consists of three steps:
  1. First, generate all possible combinations of CPUs (e.g., two CPUs in
     performance domain 0 and three CPUs in performance domain 1) to achieve
     the possible performance capacities with minimal energy consumption. We
     assume a reasonable load balancer, so the utilization of the used CPUs is
     similar.

  2. Then, from all the possible combinations of performance versus CPU sets,
     select a list of combinations that minimize the number of active
     performance domains and reduce the number of performance domain switches
     when changing performance levels.

  3. Finally, assign CPUs (@cpu_adx) to the virtual CPU ID (@cpu_vid) of
     a performance domain.

The table consists of the following information (PerfCpuOrder):
 - PerfCpuOrder::perf_cap: The upper bound of the performance capacity covered
   by this tuple.

 - PerfCpuOrder::cpus_perf: Primary CPUs to be used is ordered by preference.

 - PerfCpuOrder::cpus_ovrflw: When the system load goes beyond @perf_cap, the
   list of CPUs to be used is ordered by preference.

One exemplary result is as follows:
```
    PerfCpuOrder {
        perf_cap: 300,
        cpus_perf: [0, 1]
        cpus_ovflw: [2, 3, 4, 5, 6, 7]
    },
    PerfCpuOrder {
        perf_cap: 1138,
        cpus_perf: [2, 3, 0, 1]
        cpus_ovflw: [4, 5, 6, 7]
    },
    PerfCpuOrder {
        perf_cap: 3386,
        cpus_perf: [2, 3, 4, 5, 6]
        cpus_ovflw: [ 0, 1, 7]
    },
    PerfCpuOrder {
        perf_cap: 3977,
        cpus_perf: [2, 3, 4, 5, 6, 0]
        cpus_ovflw: [1, 7]
    },
    PerfCpuOrder {
        perf_cap: 4508,
        cpus_perf: [2, 3, 4, 5, 6, 0, 1]
        cpus_ovflw: [7]
    },
     PerfCpuOrder {
        perf_cap: 5627,
        cpus_perf: [7, 2, 3, 4, 5, 6, 0, 1]
        cpus_ovflw: []
    }
 ```

The previous code for generating CPU preference orders based on the system's CPU topology is now handled as a special case when the energy model is unavailable.

This PR does not include the proper changes to the BPF code, so only the first and last performance levels are used as the powersave and performance modes, respectively.

Note that when the energy model is not available in the system, no functional and behavioral changes are expected.

### CODE REVIEW

File: rust/scx_utils/src/energy_model.rs [hunk 1]
```diff
@@ -95,6 +96,21 @@ impl PerfDomain {
             perf_table,
         })
     }
+
+    /// Lookup a performance state by a given CPU utilization.
+    /// @util is in %, ranging [0, 100].
+    pub fn get_ps_by_util(&self, util: f32) -> Option<&Arc<PerfState>> {
```
arighi: This is just personal preference, but I tend to use get/put prefix only for methods that are acquiring an object (like incrementing/decrementing a refcount or similar). Also "ps" reminds me a bit "process".

What do you think about renaming this function `select_perf_state()` or something similar?

rust/scx_utils/src/energy_model.rs
```diff
@@ -95,6 +106,32 @@ impl PerfDomain {
...
+        let req_perf = (perf_max as f32 * (util / 100.0)) as usize;
+        for (perf, ps) in self.perf_table.iter() {
+            if *perf >= req_perf {
+                return Some(ps);
+            }
+        }
+        None
+    }
+
+    pub fn get_total_performance(&self) -> usize {
```
arighi: And maybe rename this `perf_total()`?

rust/scx_utils/src/energy_model.rs
```diff
@@ -67,14 +68,24 @@ impl EnergyModel {
...
+    pub fn get_pd_by_cpu_id(&self, cpu_id: usize) -> Option<&PerfDomain> {
         for (_, pd) in self.perf_doms.iter() {
             if pd.span.test_cpu(cpu_id) {
                 return Some(&pd);
             }
         }
         None
     }
+
+    pub fn get_total_performance(&self) -> usize {
```
arighi: Maybe rename to `perf_total()` (see other comments)?

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -390,7 +390,7 @@ impl<'a> Scheduler<'a> {
         debug!("{:#?}", order);
 
         // Initialize CPU capacity
-        for (_, cpu) in order.cpus_pf.iter().enumerate() {
+        for (_, cpu) in order.cpuids.iter().enumerate() {
```
hodgesds: nit: can probably drop the `enumerate()` since it's not being used
```suggestion
        for cpu in order.cpuids.iter() {
```

scheds/rust/scx_lavd/src/cpu_order.rs
```diff
@@ -344,3 +352,720 @@ impl CpuOrderCtx {
         d
     }
 }
+
+#[derive(Debug)]
+struct EnergyModelOptimizer<'a> {
```
hodgesds, line 357: This is really nice!

rust/scx_utils/src/energy_model.rs
multics69: Thank you for the suggestions! That makes sense. It would be better to be aligned with the kernel convention. I will change it as suggested!

scheds/rust/scx_lavd/src/main.rs
multics69: Thanks! Will fix it!

scheds/rust/scx_lavd/src/cpu_order.rs
multics69: Thank you!

multics69:
> Really interesting! I need to integrate this in scx_flash as well.

Thanks a lot! I plan to promote the CPU ordering related to the scx_utils lib. once tested a bit more. :-)

PR #2236, opened 2025-06-23
scxtop: add kprobe events to TUI

yaakov-stein:
With PR's #2225, #2229, #2230, and #2234, we can now add the ability to monitor an arbitrary number of kprobes from the TUI. The kprobe event list is separate from perf events list and can be rendered by pressing 'K'. Performance is more important with kprobe events than perf events as (on my machine) there are about 60k kfuncs, so even a basic substring search can take >6ms.

Test Plan: 
1) Attached a kprobe to `schedule` kfunc, ensured using `sudo bpftool link show` that a single kprobe was properly attached and then properly de-attached when pressing 'x'.
2) Tested multiple kprobes on different functions while perf events were also selected (once again, using `sudo bpftool link show` to ensure the kprobes were properly attached). Ensured data was only being counted for the proper kfunc (the kfunc instruction pointer address that matches the bpf event's instruction pointer address).

Demo (selecting `schedule` kfunc):

https://github.com/user-attachments/assets/598215e3-10ae-4f40-b792-75ad8bb8cc2e

Part 2 (also selecting `__alloc_pages` kfunc and moving between that and `schedule` kprobe):

https://github.com/user-attachments/assets/fe8ea0a3-5d0f-4218-be50-c270d640a123

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -2263,7 +2263,7 @@ impl<'a> App<'a> {
                 let prev_state = self.prev_state.clone();
                 self.prev_state = self.state.clone();
                 self.state = prev_state;
-                self.available_events.push(prof_event.clone());
+                self.available_events.push(prof_event);
             }
         }
 
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -3,6 +3,7 @@
 // This software may be used and distributed according to the terms of the
 // GNU General Public License version 2.
 
+use crate::available_kprobe_events;
 use crate::available_perf_events;
 use crate::bpf_intf;
 use crate::bpf_skel::BpfSkel;
@@ -16,6 +17,7 @@ use crate::AppTheme;
 use crate::CpuData;
 use crate::EventData;
 use crate::FilteredEventState;
+use crate::KprobeEvent;
 use crate::LlcData;
 use crate::NodeData;
 use crate::PerfEvent;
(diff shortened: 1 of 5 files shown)


tools/scxtop/src/kprobe_event.rs
```diff
@@ -8,6 +8,53 @@ use scx_utils::compat::tracefs_mount;
...
+pub struct KprobeEvent {
+    pub cpu: usize,
+    pub event_name: String,
+    pub count: u64,
+    pub instruction_pointer: Option<u64>,
+}
+
+impl KprobeEvent {
+    pub fn new(event_name: String, cpu: usize) -> Self {
+        let instruction_pointer = resolve_kfunc_address(&event_name);
```
yaakov-stein, line 21: At the moment, I'm just finding the address each time we make a new kprobe and this worked fine performance-wise (even though we're searching through everything in /proc/kallsyms, which was about 170k items for me). A more time-efficient but slightly more complex/annoying and less space-efficient approach would be to map all kfuncs to their address when scxtop is starting up and then just do a map lookup for each new kprobe. I thought the current approach makes sense for the current stage of scxtop, but interested to hear if anyone thinks otherwise.

tools/scxtop/src/app.rs
```diff
@@ -92,18 +94,22 @@ pub struct App<'a> {
...
+
+    // Event related
     active_event: ProfilingEvent,
     active_hw_event_id: usize,
     active_prof_events: BTreeMap<usize, ProfilingEvent>,
     available_events: Vec<ProfilingEvent>,
     event_input_buffer: String,
     perf_event_search: Search,
-
+    kprobe_event_search: Search,
```
hodgesds, line 110: It's nice that this is just a separate instance of the `Search` struct.

tools/scxtop/src/kprobe_event.rs
hodgesds: I think you could probably add a map on the `App` of cached addresses, but for now this seems good enough. Events aren't going to be going through attach/detach cycles that quickly, so that can be added as a future optimization.

tools/scxtop/src/kprobe_event.rs
```diff
@@ -8,6 +8,53 @@ use scx_utils::compat::tracefs_mount;
...
+        let count = self.count;
+        if reset {
+            self.count = 0;
+        }
+        Ok(count)
+    }
+}
+
+fn resolve_kfunc_address(name: &str) -> Option<u64> {
+    let file = File::open("/proc/kallsyms").unwrap();
```
hodgesds: It might be nice if this fails to include a message that says `CONFIG_KALLSYMS` is needed in the kernel config.

PR #2242, opened 2025-06-24
Nvml numa bind

likewhatevs:
This is a start (in need of much debugging) at getting layered to do the equivalent of setting numa affinities, more or less, when nvidia gpus are used.

It does the following:
1) Uses a template to generate layers for each numa node which are constrained to each numa node.
2) Uses nvml to create a map of gpu devices to ideal numa nodes.
3) Uses nvml to create a map of gpu devices to pids on gpus and those pids descendants.
4) It uses a new matcher to get tasks placed on layers matching the tasks ideal numa nodes (per nvml).

This does not work atm (some issue w/ template handling/types), but I think the approach is maybe straightforward enough it could work.


likewhatevs:
It didn't occur to me but this requires touching flaky/barely verifies parts of layered wrt/ older kernels, so I'm gonna table this approach for now in favor of affinitizing tasks user space side.

scheds/rust/scx_layered/src/bpf/intf.h
```diff
@@ -285,6 +286,7 @@ struct layer_match {
 	bool		exclude;
 	u64		min_avg_runtime_us;
 	u64		max_avg_runtime_us;
+	u64     node;
```
xerothermic: nit: spacing

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+    gpu_devs_to_node_info: HashMap<u32, NodeInfo>,
+    gpu_pids_to_devs: HashMap<Pid, u32>,
+    last_process_time: Option<Instant>,
+    sys: System,
+    pid_map: HashMap<Pid, Vec<Pid>>,
+    poll_interval: Duration,
+    enable: bool,
+}
+
+impl GpuTaskHandler {
```
xerothermic: If it's too much to ask, then we can defer it. I think GpuTaskHandler may belong to scx/rust/scx_utils better, wdyt?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+        for &mask in affinity.iter() {
+            let mut inner_offset: u64 = 1;
+            for _ in 0..64 {
+                if (mask & inner_offset) != 0 {
+                    return Ok(cpu);
+                }
+                inner_offset = inner_offset << 1;
+                cpu += 1;
+            }
+        }
```
xerothermic: Have we considered using cpumask.rs?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -620,6 +633,15 @@ struct Opts {
     #[clap(long, default_value = "false")]
     disable_antistall: bool,
 
+    /// Enable numa topology based gpu task affinitization.
+    #[clap(long, default_value = "false")]
+    enable_gpu_affinitize: bool,
+
+    /// Interval at which to reaffinitize gpu tasks to numa nodes.
+    /// Defaults to 30s
+    #[clap(long, default_value = "30")]
```
xerothermic: I feel we maybe even afford to do this one shot. Could we please allow 0 to just perform it once?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -688,18 +710,18 @@ struct Opts {
     print_and_exit: bool,
 }
 
-fn read_total_cpu(reader: &procfs::ProcReader) -> Result<procfs::CpuStat> {
+fn read_total_cpu(reader: &fb_procfs::ProcReader) -> Result<fb_procfs::CpuStat> {
```
xerothermic, line 712: Curious why do we change to explicitly specify the namespace? Is it for readability improvement?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+        }
+        Ok(cpuset)
+    }
+    fn init_dev_node_map(&mut self, topo: Arc<Topology>) -> Result<(), anyhow::Error> {
+        let nvml = nvml()?;
+        let device_count = nvml.device_count()?;
+
+        for idx in 0..device_count {
+            let dev = nvml.device_by_index(idx)?;
+            let cpu = dev.cpu_affinity(16)?;
```
xerothermic, line 1346: Does the 16 here means cpuSetSize = 16? If so, since we only need to find 1 CPU, perhaps we can set this to 1?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+    }
+    pub fn init(&mut self, topo: Arc<Topology>) {
+        if !self.enable || self.last_process_time.is_some() {
+            return;
+        }
+
+        match self.init_dev_node_map(topo) {
+            Ok(_) => {}
+            Err(e) => {
+                error!("Error initializing gpu map: {}", e);
```
xerothermic: Maybe "Error initializing NVML..."?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+            }
+        };
+        match self.init_process_info() {
+            Ok(_) => {}
+            Err(e) => {
+                error!("Error initializing gpu map: {}", e);
```
xerothermic: Maybe Error initializing GPU process info?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+        if !self.enable {
+            return Ok(());
+        }
+        for (pid, dev) in &self.gpu_pids_to_devs {
+            let node_info = self
+                .gpu_devs_to_node_info
+                .get(&dev)
+                .expect("Unable to get gpu pid node mask");
+            for child in self.get_child_pids_and_tids(*pid) {
+                match nix::sched::sched_setaffinity(
```
xerothermic, line 1422: Not sure if we track all the syscall we used in a guide. If not, perhaps we can start documenting what cap we need to allow to make this work in a systemd environment

scheds/rust/scx_layered/src/bpf/intf.h
likewhatevs: ahh so you want to open PR/the overall to see things to comment on (i.e. this was from the first approach/doesn't apply to latest). So like, the files changed page.

scheds/rust/scx_layered/src/main.rs
likewhatevs: Eventually perhaps, but atm this is a kinda-nasty hack so I'd leave it local to layered until then. That being said, it is written such that a change like that would be easy in the future.

scheds/rust/scx_layered/src/main.rs
likewhatevs: This makes a nix cpumask because the set affinity fn exposed by that uses that.

scheds/rust/scx_layered/src/main.rs
likewhatevs: I would say, in that case, set time to 1 hour.

scheds/rust/scx_layered/src/main.rs
likewhatevs: Misc cleanup w/ the other changes (i.e. the namespace rename never should have been done). W/o this change, we use fb_procfs here and rename it as procfs (and the two are similar enough that like... not a good thing to do), despite the fact that procfs has features fb_procfs doesn't have, which we'll probably want to use when (if) ever we go back to migrate memory manually when making gpu using pids use their ideal numa nodes.

scheds/rust/scx_layered/src/main.rs
likewhatevs: that means this works for machines with up to 16 (size of cpumasks returned) * 64 (number of cpus a u64 can represent when it's a bitmask) cpus. so this will work for machines w/ up to 1024 cpus.

scheds/rust/scx_layered/src/main.rs
likewhatevs: that would be in the {} part of that string/the thing that returns is the device to node map.

scheds/rust/scx_layered/src/main.rs
likewhatevs: ahh yeah two same errors are useless, thx

scheds/rust/scx_layered/src/main.rs
likewhatevs: the presumption w/ a lot of this is this needs to be ran as root w/ more/less no constraints. iiuc. constraints etc. from things like systemd units are all not-default (so if encountering issues, the way around it would be to remove them till it works).

scheds/rust/scx_layered/src/main.rs
likewhatevs: Err, 1 hour or 1 day or 1 week etc. Having this just be ran once is probably unwise because the system it is helping is not like, a static thing. Tasks running on a system change, so if this process is heavy, the interval at which this is ran should be adjusted to accommodate that (but it should be an interval, not a one off thing).

xerothermic: Actually, thinking about it some more. Not sure we can / want to enable this continuously fleet-wide. For GPU workloads with proper NUMA binding settings, this has no added benefit and likely causing regressions. If it's only done once, then the overhead can be discarded.

I agree for no NUMA binding jobs, it could be slightly safer to refresh it (though in practice, we haven't seen the need to refresh because the bash script also only set it once.) 

As for when the system changes, we will relaunch the scheduler, so the setting gets to apply again.

scheds/rust/scx_layered/src/main.rs
xerothermic: A comment here would be nice.

xerothermic:
Thanks a lot for this PR! Definitely helped with performance of jobs I tested today. 

You will probably agree this is a stop-gap solution until we get proper NUMA binding support from within the application. For this purpose, do you think we can have a monotonic increased counters to show how many tasks we have corrected the "wrong CPU affinity"? 

I think exposing this information will give us the visibility which workflows still have application bug that needs attention, and we can go after them to fix it (once the PyTorch PR is ready.) 

The reason for requesting a monotonic increased counter is because it's likely the tasks stay alive after we set it the first time, so periodic counters will not be able to capture this precisely.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -88,6 +94,13 @@ const NR_LLC_LSTATS: usize = bpf_intf::llc_layer_stat_id_NR_LLC_LSTATS as usize;
 
 const NR_LAYER_MATCH_KINDS: usize = bpf_intf::layer_match_kind_NR_LAYER_MATCH_KINDS as usize;
 
+static NVML: OnceCell<Nvml> = OnceCell::new();
+
+#[inline]
```
htejun: Let's let the compiler decide this.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
         Ok(nr_new_cpus)
     }
 }
+#[derive(Debug, Clone)]
+struct NodeInfo {
+    node_mask: nix::sched::CpuSet,
+    _node_id: usize,
+}
```
htejun, line 1286: Can we add a blank line between blocks? Here and other places too.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+        let mut cpu: u32 = 0;
+        for &mask in affinity.iter() {
+            let mut inner_offset: u64 = 1;
+            for _ in 0..64 {
+                if (mask & inner_offset) != 0 {
+                    return Ok(cpu);
+                }
+                inner_offset = inner_offset << 1;
+                cpu += 1;
+            }
```
htejun, line 1326: Use `u64::trailing_zeros()`?

scheds/rust/scx_layered/src/main.rs
htejun: It'd probably be better to add a conversion function to cpumask.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+        }
+        anyhow::bail!("unable to get CPU from NVML bitmask");
+    }
+    fn node_to_cpuset(&self, node: &scx_utils::Node) -> Result<CpuSet, anyhow::Error> {
+        let mut cpuset = CpuSet::new();
+        for (cpu_id, _cpu) in &node.all_cpus {
+            cpuset.set(cpu_id.clone())?;
```
htejun: Isn't `cpu_id` an integer? `*cpu_id`?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+    fn init_dev_node_map(&mut self, topo: Arc<Topology>) -> Result<(), anyhow::Error> {
+        let nvml = nvml()?;
+        let device_count = nvml.device_count()?;
+
+        for idx in 0..device_count {
+            let dev = nvml.device_by_index(idx)?;
+            let cpu = dev.cpu_affinity(16)?;
+            let ideal_cpu = self.find_one_cpu(cpu)?;
+            for (node_id, node) in topo.nodes.iter() {
+                if node.all_cpus.contains_key(&(ideal_cpu as usize)) {
```
htejun: Isn't this loop just doing `topo.all_cpus[ideal_cpu].node_id`?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+                    self.gpu_devs_to_node_info.insert(
+                        idx,
+                        NodeInfo {
+                            node_mask: self.node_to_cpuset(node)?,
+                            _node_id: *node_id,
+                        },
```
htejun: If we add a helper to return `CpuSet` to `Cpumask`, the above would just need to store gpu -> node_id, right?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+            gpu_devs_to_node_info: HashMap::new(),
+            gpu_pids_to_devs: HashMap::new(),
+            last_process_time: None,
+            sys: System::default(),
+            pid_map: HashMap::new(),
+            poll_interval: Duration::from_secs(poll_interval),
+            enable,
+        }
+    }
+    fn find_one_cpu(&self, affinity: Vec<u64>) -> Result<u32, anyhow::Error> {
```
htejun: Isn't `anyhow::Error` redundant? Here and other places too.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+                .into_iter()
+                .chain(device.running_graphics_processes()?.into_iter())
+            {
+                self.gpu_pids_to_devs.insert(Pid::from_u32(proc.pid), i);
+            }
+        }
+
+        Ok(())
+    }
+    fn init_process_info(&mut self) -> Result<(), anyhow::Error> {
```
htejun: Is this a useful function?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
         Ok(nr_new_cpus)
     }
 }
+#[derive(Debug, Clone)]
+struct NodeInfo {
+    node_mask: nix::sched::CpuSet,
+    _node_id: usize,
+}
+#[derive(Debug)]
+struct GpuTaskHandler {
```
htejun: Overall comment explaining what the struct does would be great. Maybe the name is too generic. `GpuTaskAffinitizer`?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1268,211 @@ impl Layer {
...
+                        );
+                    }
+                };
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn process(&mut self) {
```
htejun: `affinitize()`?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -2699,6 +2930,7 @@ impl<'a> Scheduler<'a> {
         )?;
         self.refresh_cpumasks()?;
         self.refresh_idle_qos()?;
+        self.gpu_task_handler.process();
```
htejun: We prioritize userspace scheduler thread quite a bit and listing all PIDs in the system can take a while. Given that this isn't latency sensitive, it likely would be a good idea to spawn this into a separate thread which is kicked off from `.init()` and update `is_scheduler_task()` to exclude the thread. Given that what this does doesn't have much to do with the scheduling loop, this can also be moved into its own process. ie. Make it a subcommand of `scx_layered`.

scheds/rust/scx_layered/src/main.rs
likewhatevs: We can, but chat change would require adding the nix crate as a dependency to utils, which everyone depending upon that would then need to deal with.

scheds/rust/scx_layered/src/main.rs
likewhatevs: Will try and make this change -- this is boilerplate to make nvml only be initialized once because that can be heavy.

scheds/rust/scx_layered/src/main.rs
likewhatevs: That would be nice but that would add the dependency on the crate we're using for this to all scheds (which we could deal w/ w/ a feature flag which kinda helps but not that much tbh).

scheds/rust/scx_layered/src/main.rs
likewhatevs: > Given that what this does doesn't have much to do with the scheduling loop, this can also be moved into its own process. ie. Make it a subcommand of scx_layered.

Running on a separate thread periodically we can do. Making this a subcommand wouldn't solve much though.

scheds/rust/scx_layered/src/main.rs
htejun: Yeah, I don't think depending on `nix` would be a problem. It's not huge or anything, right?

scheds/rust/scx_layered/src/main.rs
htejun: Oh, I was just talking about dropping `inline`.

scheds/rust/scx_layered/src/main.rs
likewhatevs: True but the concern wrt/ that is more that it's gonna require downstreams all use the same version (and opens the possibility of conflicts wrt/ versions needed). Will give this a try, could be fine if that crate is pretty stable/unchanging.

likewhatevs:
Closing due to concerns with potential overhead of running this in a periodic manner.

likewhatevs:
Added stats to enable us to see if additional perf optimizations are needed wrt/ this + misc cleanup. I *think* running it every X mins is probably fine and stats should confirm/deny this.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1246,6 +1279,218 @@ impl Layer {
...
+                        // Increment the global counter for successful affinitization
+                        self.tasks_affinitized += 1;
+                    }
+                    Err(_) => {
+                        warn!(
+                            "Error affinitizing gpu pid {} to node {:#?}",
+                            child.as_u32(),
+                            node_info
+                        );
+                    }
```
xerothermic, line 1436: I'd suggest we need the following counters.
1. number of task that has different CPU affinity than the desired affinity. For this, we probably need to read the affinity first to compare. I know it's going to be more expensive, but I think this is the only way to know if we are doing useful work. (e.g. correct the affinity settings)
2. number of task successfully affinitized, you already added this.
3. number of task failed to affinitized, I think it could be interesting to know if we have any task that failed to affinitize, although I can't think of a reason now.

likewhatevs: (1) runs the risk of requiring this be refactored to run in a separate thread. I advise doing this in a subsequent PR or waiting for when time permits me to update this PR (guessing, at least month or so from now).
(2) no -- I am logging the count of tasks which blindly have their affinity set en-masse to what we think they should be.
(3) Not sure that failing is a thing tbh. To avoid excess stats (as may be the case w/ layer stats), a warning log is probably best for this.

xerothermic: Thanks, @likewhatevs. Let's merge this and I'll submit a follow up PR to track the number of tasks that had incorrect CPU affinity.

likewhatevs: Alright -- I think that might require moving this to a thread, but maybe do that after confirming w/ the processing delay stat that that's needed (it *might* not be).

PR #2246, opened 2025-06-24
ci: enable scx_wd40

### LINKED ISSUE #1517: scx_wd40: No CI job
@likewhatevs I am opening a separate issue out of the pull request so we can discuss adding a CI job for WD40. The scheduler should only fail when it gets added features that have not yet been merged into upstream trees, though that happens from time to time. I assume we can add a CI job that fails silently to avoid excessive noise, right? Basically same way we were treating the CI jobs against bpf-next in the past.

JakeHillion:
Add `scx_wd40` to our CI runs. This needs a few new features to work properly. Namely:
- Add `metadata.scx.ci.kernel.default` to specify a different default kernel for pull requests when the usual default of `sched_ext/for-next` is inappropriate.
- Add `metadata.scx.ci.kernel.allowlist` to specify an acceptable set of kernels when many are inappropriate.
- Add similar `metadata.scx.ci.kernel.blocklist` to avoid testing on a kernel when its known to fail.

The new metadata is managed in each scheduler's Cargo.toml rather than having to change a central list in the CI spec and removes some edge cases.

Closes #1517

Test plan:
- Ran `.github/include/list-integration-tests.py` with various kernels. It worked.
- Checked the new `scx_wd40` entry on this PR. It selects the correct kernel in the "Load kernel" step.

### DIFF
--- a/.github/include/list-integration-tests.py
+++ b/.github/include/list-integration-tests.py
@@ -132,6 +132,7 @@ def main():
             "scx_rustland",
             "scx_rusty",
             "scx_tickless",
+            "scx_wd40",
         ]:
             reqs = kernel_reqs.get(scheduler, {})
             allowlist = reqs.get("allowlist", [])
--- a/meson-scripts/stress_tests.ini
+++ b/meson-scripts/stress_tests.ini
@@ -10,6 +10,12 @@ sched_args: -v
 stress_cmd: stress-ng -t 31 --aggressive -M -c `nproc` -f `nproc` --affinity 1 --affinity-delay 1 --affinity-pin
 timeout_sec: 45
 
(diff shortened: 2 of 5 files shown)


JakeHillion:
Draft as we need to fix the new CI failure, I think it's legitimate but not sure yet.

etsal:
This looks reasonable, the build is failing because wd40 directly uses BPF cpumasks and so depends on the definition in vmlinux.h being correct. However, our vmlinux.h has a hardcoded value for cpumask length that is causing this failure. I will look into either generating a vmlinux.h on the fly for wd40, or packaging it with its own private copy.

htejun:

PR #2247, opened 2025-06-24
lib/minheap: error out on invalid capacity sizes

etsal:
Add an additional check during minheap creation to prevent callers from requesting overly large minheaps. We are currently hardcoding the size of all minheaps due to verifier limitations, but this means we are unable to handle requests to create a minheap larger than our current hardcoded maximum size.


etsal:
Note: The testbench is currently failling with the current addition, but i am not currently on a machine where I can debug it. I will do so later tonight.

lib/selftests/st_minheap.bpf.c
```diff
@@ -202,6 +202,18 @@ int scx_selftest_minheap_random(scx_minheap_t *heap)
 	return 0;
 }
 
+static
+int scx_selftest_minheap_toolarge(scx_minheap_t *heap)
+{
+	void __arena *heap2;
+
+	heap2 = scx_minheap_alloc(SCX_MINHEAP_MAX_CAPACITY + 1);
+	if (heap2)
```
hodgesds, line 211: nit: could assert the error value as well

Issue #2209, opened 2025-06-18
scxtop 1.0.14 tui broken


dongdigua:
```
OS: Arch
Kernel: 6.15.2-zen1-1-zen
scxtop 1.0.14
```
behavior: run `scxtop`, flashes a tui for a short moment, then tui closed and show nothing

scxtop 1.0.13 from cargo isn't broken

hodgesds:
@dongdigua quick question, what hardware are you running it on? I've seen a similar bug on a Ryzen 7940HX.

terencode:
I'm affected too and I have a ryzen 3900x.

mati865:
I've been trying to run it for a few months and it always had this problem. If I overcommit my CPU really hard then TUI is visible for a few seconds.
Ryzen 5950X here.

yaakov-stein:
We'll look into this. Can you add RUST_BACKTRACE=1 and let me know what you get? Thanks

dongdigua:
> [@dongdigua](https://github.com/dongdigua) quick question, what hardware are you running it on? I've seen a similar bug on a Ryzen 7940HX.

11th Gen Intel i5-1135G7

dongdigua:
> RUST_BACKTRACE=1

it don't crash, so no stack trace

mati865:
Actually there is a crash but ratatui hides the stacktrace. It can be obtained with this trick:
```
❯ sudo RUST_BACKTRACE=1 ../../target/debug/scxtop 2> ~/a.txt

❯ bat ~/a.txt
```
Then among the mangled output you can see:
```
thread 'main' panicked at /home/mateusz/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/str/mod.rs:816:21:
byte index 4 is not a char boundary; it is inside '\u{202f}' (bytes 2..5) of `79 817 881`
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/panicking.rs:75:14
   2: core::str::slice_error_fail_rt
   3: core::str::slice_error_fail
             at /rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/str/mod.rs:68:5
[...]

PR #2248, opened 2025-06-24
scxtop: temporary fix for locale crash

yaakov-stein:
Several users from the community noted that scxtop crashed right after starting it (see #2209). With a helpful backtrace from @mati865, it became clear the root of the problem was in combining num_format::SystemLocale's separator with the way ratatui splits strings. The num_format library uses the short 3 byte whitespace character `\u{202f}` as a separator for certain regions (France, several European countries, and wherever they write one million as 1 000 000). However, ratatui assumes the whitespace characters are in 4 byte increments and was therefore panicking when it attempted to split_at() the whitespaces.

For a temporary fix, I'm going to fixe the locale to use Locale::en, which uses a `,` as a separator.

Test Plan:
- I put myself in the French locale using num_format::Locale::fr and reproduced the crash. The crash does not occur in regions with other separators.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -40,7 +40,7 @@ use anyhow::{bail, Result};
 use glob::glob;
 use libbpf_rs::Link;
 use libbpf_rs::ProgramInput;
-use num_format::{SystemLocale, ToFormattedString};
+use num_format::{Locale, ToFormattedString};
 use ratatui::prelude::Constraint;
 use ratatui::{
     layout::{Alignment, Direction, Layout, Rect},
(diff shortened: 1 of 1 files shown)


tools/scxtop/src/app.rs
```diff
@@ -236,7 +236,7 @@ impl<'a> App<'a> {
             config,
             localize: true,
             hw_pressure,
-            locale: SystemLocale::default()?,
+            locale: Locale::en,
```
yaakov-stein, line 239: I think a better solution would be to sanitize the input (convert the 3 byte whitespace to a standard 4 byte whitespace) and continue using SystemLocale so people can view the data in their preferred way. This feels a bit hacky to me though, so I'm interested to hear what others think.

hodgesds: Yeah, I feel like that is a better user experience. Not sure how hard that would be.

hodgesds (approved):
This should work for now, hopefully we can get a fix upstream in ratatui.

tools/scxtop/src/app.rs
yaakov-stein: I don't think it should be complicated. Will push this for now (to prevent the crash) and I'll plan on adding the more permanent fix tomorrow.

PR #2252, opened 2025-06-25
lib/minheap: Simplify parent index calculation

EricccTaiwan:
No functional change.


etsal:
LGTM, but let's wait until the selftests are on the CI to ensure there is no regression (there won't be, but just as a precaution).

EricccTaiwan:
The CI seems to be stuck. Should I just rebase and push?
image

Edit: rebased, and `push --force`

EricccTaiwan:
@etsal PTAL, thanks.

PR #2256, opened 2025-06-25
ci: run scx_lib_selftests and disable it publishing

JakeHillion:
`scx_lib_selftests` is a crate provided to run selftests written in BPF for the lib code. Run this in the CI so it doesn't regress. We already build all of the Rust crates in the workspace and have a VM (selftests need root) so simply throw running the binary in the VM inner logic.

This only runs on our sched_ext/for-next kernel for now. It's pretty easy to extend, but we currently only spin up the for-next VM in this step. Easy to extend later on.

Drive-by: disable publishing on `scx_lib_selftests` crate. I don't think we need to publish this, it's effectively an internal crate for testing.

Test plan:
- CI

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -165,6 +165,8 @@ def run_tests_in_vm():
         ]
     )
 
+    run_command(["target/debug/scx_lib_selftests"])
+
 
 def run_all():
     """Run all CI steps in the correct order."""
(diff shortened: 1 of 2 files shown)


etsal:
Looks good! And yeah, the tests will at some point be factored out together with the library into a separate crate, but at no point will the tests themselves be standalone.

etsal (approved):

PR #2257, opened 2025-06-25
scx_p2dq: Fix interactive load balancing

hodgesds:
The --dispatch-lb-interactive flag is used to configure the load balancing of interactive tasks on the dispatch path. However, this functionality was removed in recent refactoring. Add handling for the flag in the can_migrate helper.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -278,6 +278,10 @@ static bool is_interactive(task_ctx *taskc)
 
 static __always_inline bool can_migrate(task_ctx *taskc)
 {
+	if (!dispatch_lb_interactive && taskc->interactive)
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -728,7 +730,7 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
 		if (deadline_scheduling)
 			set_deadline_slice(taskc, llcx);
 
-		if (interactive_dsq && is_interactive(taskc) && !can_migrate(taskc)) {
+		if (interactive_dsq && tackc->interactive && !can_migrate(taskc)) {
```
hodgesds: drive by fixes to use the stored value on the task_ctx

etsal (approved):

PR #2258, opened 2025-06-25
scxtop: bump release to 1.0.15

JakeHillion:


yaakov-stein (commented):
Do we want to wait until we get rid of a few more known bugs before bumping the release? Have a few that I should be able to get rid of before EOW.

JakeHillion:
> Do we want to wait until we get rid of a few more known bugs before bumping the release? Have a few that I should be able to get rid of before EOW.

I think little and often is the way to go here, we didn't have to bump any dependency versions for this release so it's not much effort. We can do another after the bug fixes or wait until the monthly sched_ext release for the rest.

PR #2259, opened 2025-06-25
scx_flash: Misc improvements

arighi:
A series of minor adjustments and optimizations that do not alter the scheduler's functional behavior.

These changes allowed to improve performance across several benchmarks (notably MySQL and Redis) while maintaining responsiveness for typical interactive workloads (gaming).


hodgesds (approved):
Makes sense

PR #2260, opened 2025-06-25
scx: add more support code for unit tests

josefbacik:
Add a basic bpf map implementation and then rework the existing p2dq unit tests to be in their own file, and then add a basic unit test for a function that uses the map code to validate everything works.


scheds/rust/scx_p2dq/src/bpf/main.test.bpf.c
```diff
@@ -0,0 +1,44 @@
...
+
+#include <string.h>
+
+#include "main.bpf.c"
+
+static struct scx_percpu_test_map *cpu_ctxs_map = NULL;
+
+static void test_lookup_cpu_ctx(void)
+{
+	u32 index = 0;
```
rrnewton, line 14: Minor: Since it's a data structure, wouldn't it be only a little more code to put a loop around this and test a collection of random indices (with constant seed) rather than index 0?

(Since it's pointwise operations only for this basic map no good way to check the size / set contents of the map... but can at least do N (read;write) pairs or (N writes; N reads) with an optional outer loop.)

PR #2263, opened 2025-06-25
scx_chaos: add kprobe delays

yaakov-stein:
This adds the ability to introduce delays whenever a provided kprobe is hit. kprobes and a desired delay frequency can be passed in via the command line.

Testing: 
1) Ensured that only using kprobe-frequency arg causes stops scheduler from starting.
2) Ensured multiple kprobes can be attached at different frequencies. Used `bpftool link show` to verify all kprobes were attaching properly. Used dbg to ensure BPF handler was properly executing.
3) Ensured that program correctly exits immediately if kprobes fail validation.
4) Works with only --kprobes set (uses default frequency of .1)

### DIFF
--- a/scheds/rust/scx_chaos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_chaos/src/bpf/main.bpf.c
@@ -91,6 +91,24 @@ static __always_inline void chaos_stat_inc(enum chaos_stat_idx stat)
 		(*cnt_p)++;
 }
 
+SEC("kprobe/generic")
+int generic(struct pt_regs *ctx)
+{
+	struct task_struct *p;
+	struct chaos_task_ctx *taskc;
+
+	p = (struct task_struct *)bpf_get_current_task_btf();
+	if (!p)
+		return -EINVAL;
+
+	if (!(taskc = lookup_create_chaos_task_ctx(p)))
+		return -EINVAL;
+
+	taskc->next_trait = CHAOS_TRAIT_RANDOM_DELAYS;
+
+	return 0;
+}
+
 static __always_inline enum chaos_trait_kind choose_chaos(struct chaos_task_ctx *taskc)
 {
 	if (taskc->match & CHAOS_MATCH_EXCLUDED) {
--- a/scheds/rust/scx_chaos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_chaos/src/bpf/main.bpf.c
@@ -599,7 +599,6 @@ int generic(struct pt_regs *ctx)
 	u32 roll = bpf_get_prandom_u32();
 	if (roll < kprobe_delays_freq_frac32) {
 		taskc->next_trait = CHAOS_TRAIT_RANDOM_DELAYS;
-		p->scx.slice = 0;
 		dbg("GENERIC: setting next_trait to RANDOM_DELAYS - task[%d]", p->pid);
 	}
 
--- a/scheds/rust/scx_chaos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_chaos/src/bpf/main.bpf.c
@@ -589,7 +589,7 @@ SCX_OPS_DEFINE(chaos,
 	       .init_task		= (void *)chaos_init_task,
 	       .runnable		= (void *)chaos_runnable,
 	       .select_cpu		= (void *)chaos_select_cpu,
-		.tick 			= (void *)chaos_tick,
+	       .tick 		    	= (void *)chaos_tick,
 
 	       .exit_task		= (void *)p2dq_exit_task,
 	       .exit			= (void *)p2dq_exit,
--- a/scheds/rust/scx_chaos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_chaos/src/bpf/main.bpf.c
@@ -50,6 +50,8 @@ const volatile u32 cpu_freq_max = SCX_CPUPERF_ONE;
 const volatile u32 degradation_freq_frac32 = 1;
 const volatile u64 degradation_frac7 = 0;
 
+const volatile u32 kprobe_delays_freq_frac32 = 1;
+
 #define MIN(x, y) ((x) < (y) ? (x) : (y))
 #define MAX(x, y) ((x) > (y) ? (x) : (y))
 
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -584,3 +586,25 @@ SCX_OPS_DEFINE(chaos,
...
+	struct task_struct *p;
+	struct chaos_task_ctx *taskc;
+
+	p = (struct task_struct *)bpf_get_current_task_btf();
+	if (!p)
+		return -EINVAL;
+
+	if (!(taskc = lookup_create_chaos_task_ctx(p)))
+		return -EINVAL;
+
```
yaakov-stein, line 615: Ideally, we should be calling [scx_bpf_task_cpu](https://docs.ebpf.io/linux/kfuncs/scx_bpf_task_cpu/) and [scx_bpf_kick_cpu](https://docs.ebpf.io/linux/kfuncs/scx_bpf_kick_cpu/), but those cannot be used in a kprobe BPF handler (see Program Types under those functions).

scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -584,3 +586,25 @@ SCX_OPS_DEFINE(chaos,
...
+	u32 roll = bpf_get_prandom_u32();
+	if (roll < kprobe_delays_freq_frac32) {
+		taskc->next_trait = CHAOS_TRAIT_RANDOM_DELAYS;
```
yaakov-stein: If I understand the way scx schedulers work correctly, this should not be overwritten until we add a delay. Given we can't call `scx_bpf_kick_cpu`, the task will finish running and then get re-entered into `select_cpu`. In `chaos_select_cpu`, we call `chaos_trait_skips_select_cpu`, which will be true if we just set the trait.

scheds/rust/scx_chaos/src/bpf/main.bpf.c
hodgesds: I believe you can use this instead:
```suggestion
p->scx.slice = 0
```
That should set the slice to 0 and on the next tick it should go through the `stopping` path.

yaakov-stein: It looks like that causes a verifier error:

```
; p->scx.slice = 0; @ main.bpf.c:606
21: (7b) *(u64 *)(r7 +904) = r1
only read is supported
```

My guess is the verifier won't let me write when I'm in a kprobe handler, but I'll look into it a bit more and see if I can find a way to get around the verifier.

hodgesds: could try something like [this](https://gist.github.com/hodgesds/d797a165eb091a15c41d26f1324fb9c6)

scheds/rust/scx_chaos/src/lib.rs
```diff
@@ -546,6 +598,18 @@ pub struct PerfDegradationArgs {
     pub degradation_frac7: u64,
 }
 
+/// Delay a process when a kprobe is hit.
+#[derive(Debug, Parser)]
+pub struct KprobeArgs {
+    /// Introduce random delays in the scheduler whenever a provided kfunc is hit.
```
JakeHillion:
```suggestion
    /// Introduce random delays in the scheduler whenever a provided kprobe is hit.
```

nit: I don't think kfunc is the write word here

scheds/rust/scx_chaos/src/bpf/main.bpf.c
JakeHillion: Well that's super annoying, none of the workarounds I tried work either, and it seems this is intentional. Would be happy for you to add the tick function, this is behaviour we need more generally.

scheds/rust/scx_chaos/src/bpf/main.bpf.c
JakeHillion: I think we need to add a check in chaos_runnable and change the state machine slightly. chaos_runnable calls `choose_chaos(wakee_ctx);` every time and sets `next_trait` to that value, so I think we'll be overwriting this every time. What do you reckon?

scheds/rust/scx_chaos/src/lib.rs
```diff
@@ -546,6 +598,18 @@ pub struct PerfDegradationArgs {
...
+    #[clap(long, num_args = 1.., value_parser)]
+    pub kprobes: Vec<String>,
```
JakeHillion: nit: could we be super explicit at the minute and name this `kprobes_for_random_delays` and `kprobe_random_delay_frequency`? I've been sticking to longer names so far because they get confusing quickly, and I'm thinking we'll likely want to complicate the kprobe interface more at some point in the future.

scheds/rust/scx_chaos/src/bpf/main.bpf.c
yaakov-stein: Ah, ya - I incorrectly assumed that since select_cpu checks for CHAOS_TRAIT_RANDOM_DELAYS, it would prevent runnable from running (not sure why I assumed that). I'll add a check in runnable to stop it from overwriting.

scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -580,3 +599,25 @@ SCX_OPS_DEFINE(chaos,
...
+
+	p = (struct task_struct *)bpf_get_current_task_btf();
+	if (!p)
+		return -EINVAL;
+
+	if (!(taskc = lookup_create_chaos_task_ctx(p)))
+		return -EINVAL;
+
+	u32 roll = bpf_get_prandom_u32();
+	if (roll < kprobe_delays_freq_frac32) {
```
JakeHillion: nit: I think this should be <= else setting it to 100% won't be strictly correct

scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -571,6 +589,7 @@ SCX_OPS_DEFINE(chaos,
 	       .init_task		= (void *)chaos_init_task,
 	       .runnable		= (void *)chaos_runnable,
 	       .select_cpu		= (void *)chaos_select_cpu,
+		   .tick 			= (void *)chaos_tick,
```
JakeHillion: nit: indentation

scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -550,6 +558,16 @@ s32 BPF_STRUCT_OPS(chaos_select_cpu, struct task_struct *p, s32 prev_cpu, u64 wa
...
 }
 
+void BPF_STRUCT_OPS(chaos_tick, struct task_struct *p)
+{
+	struct chaos_task_ctx *taskc;
+	if (!(taskc = lookup_create_chaos_task_ctx(p)))
+		return;
+
+	if (taskc->pending_trait)
+		p->scx.slice = 0;
```
JakeHillion: As mentioned offline, could we add a function to determine whether the task needs to be preempted? For example the futex method I'm adding won't require preemption in tick.

yaakov-stein:
> Nice! One major thing I noticed is that we don't seem to place this task in the delay DSQs. What we have here at the minute is a preemption soon after the task hits the kprobe, and not a random delay after it hits the kprobe.
> 
> I'm fine with that as the behaviour if that's intentional, but I think in that case the comments/help message would need updating. Alternatively you can throw it into the delay DSQ when it next hits enqueue after being preempted.

In runnable, we set next_trait to pending_trait, and since pending_trait was set to CHAOS_TRAIT_RANDOM_DELAYS in the kprobe handler, this will trigger the standard random delay process.

JakeHillion (approved):
Fantastic, I look forward to playing with this!

PR #2264, opened 2025-06-26
lib: add initial arena DSQ abstraction

etsal:
Add an arena-based DSQ abstraction built on top of a minheap. This abstraction lets us experiment with alternative (non-local) DSQ designs by providing a BPF-based implementation. The abstraction initially supports simple enqueue/dequeue operations and will be expanded as needed to support all operations offered by the scx_bpf kfunc API.

NOTE: DSQs are currently allocated statically and not freed. They also have a maximum size. Both of these limitations are temporary and will be removed in subsequent patches.


htejun:
I don't think it's a good idea to use DSQ in the name. It's too confusing. Let's pick some other name. ARQ?

etsal:
> I don't think it's a good idea to use DSQ in the name. It's too confusing. Let's pick some other name. ARQ?

Makes sense, how about ATQ (Arena task queue)?

lib/atq.bpf.c
```diff
@@ -0,0 +1,97 @@
...
+#include <lib/sdt_task.h>
+
+#include <lib/atq.h>
+
+/*
+ * Arena task queue implementation.
+ */
+
+__weak
+u64 scx_arena_atq_create_internal(bool fifo)
```
htejun: Can we just use`scx_atq_` as the prefix. I kinda hate long names and `atq` by itself should be identifying enough.

lib/atq.bpf.c
```diff
@@ -0,0 +1,97 @@
...
+	int ret;
+
+	if (!atq->fifo)
+		return -EINVAL;
+
+	ret = arena_spin_lock(&atq->lock);
+	if (ret)
+		return ret;
+
+	ret = scx_minheap_insert(atq->heap, taskc_ptr, atq->seq++);
```
yaakov-stein, line 41: I'd imagine a u64 is just so large that we're not concerned for overflow here, even if a long-running scheduler uses an atq? (Because theoretically, if it did overflow, it would starve the tasks that got queued before the overflow occured.)

etsal: I'd say so, though we can add an if (unlikely(seq == U64_MAX)) statement to make extra sure.

PR #2267, opened 2025-06-26
docs: Update gentoo instructions

hodgesds:


hodgesds:
> Dumb question, does this mean the `sys-kernel/scx` ebuild is wrong? At least libseccomp must be needed to compile rustland_core. https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-kernel/scx/scx-1.0.13.ebuild

Yeah, it's probably broken. The other annoying thing is different schedulers may require different sets of kconfigs.

JakeHillion (approved):

vimproved:
> > Dumb question, does this mean the `sys-kernel/scx` ebuild is wrong? At least libseccomp must be needed to compile rustland_core. https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-kernel/scx/scx-1.0.13.ebuild
> 
> Yeah, it's probably broken. The other annoying thing is different schedulers may require different sets of kconfigs.

Gentoo maintainer here. This is my bad, I'll send in a fix for the missing dependencies along with the 1.0.14 update. Sorry for any confusion or trouble. If you ever notice any other issues that you suspect are caused by Gentoo's packaging, feel free to just tag me in the issue.

As for kernel config, Gentoo has been moving more and more towards distribution kernels as opposed to manually configured kernels, so I wouldn't worry too much about kernel config issues.

PR #2268, opened 2025-06-26
scx_mitosis: Eliminate fallback dsqs

dschatzberg:
mitosis has the same logic from layered to assign pinned tasks to fallback dsqs. Cells are determined by having an explicit cpuset so pretty much all tasks in a cell were getting sent to the lo_fallback dsq. The first commit fixes this by only assigning tasks to the fallback dsqs if their pinned cpus aren't a superset of the cell cpus.

In doing so, I noticed this could cause starvation of the lo_fallback dsq. The same behavior was noticed with scx_layered and the fix was to periodically service the lo_fallback dsq. I decided to go a slightly different direction and eliminate the strict prioritization of fallback dsqs in favor of vtime scheduling.

The second commit adds per-cpu dsqs for pinned tasks. The logic ensures that the vtime of per-cpu dsqs and per-cell dsqs can be compared. Each cpu pulls from whichever has the lower vtime.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -120,13 +120,14 @@ static inline struct cgroup *task_cgroup(struct task_struct *p)
 
 /*
  * task_ctx is the per-task information kept by scx_mitosis
-*/
+ */
 struct task_ctx {
 	/* cpumask is the set of valid cpus this task can schedule on */
 	/* (tasks cpumask anded with its cell cpumask) */
 	struct bpf_cpumask __kptr *cpumask;
 	/* started_running_at for recording runtime */
 	u64 started_running_at;
+	u64 basis_vtime;
 	/* For the sake of monitoring, each task is owned by a cell */
 	u32 cell;
 	/* For the sake of scheduling, a task is exclusively owned by either a cell
@@ -200,7 +201,7 @@ static inline struct cell *lookup_cell(int idx)
 /*
  * Cells are allocated concurrently in some cases (e.g. cgroup_init).
  * allocate_cell and free_cell enable these allocations to be done safely
-*/
+ */
 static inline int allocate_cell()
 {
 	int cell_idx;
@@ -239,8 +240,10 @@ static inline int free_cell(int cell_idx)
  */
 struct cell_cpumask_wrapper {
 	struct bpf_cpumask __kptr *cpumask;
-	/* To avoid allocation on the reconfiguration path, have a second cpumask we
-	   can just do an xchg on. */
+	/*
+	 * To avoid allocation on the reconfiguration path, have a second cpumask we
+	 * can just do an xchg on.
+	 */
 	struct bpf_cpumask __kptr *tmp_cpumask;
 };
 
@@ -417,7 +420,7 @@ static inline int update_task_cpumask(struct task_struct *p,
 /*
  * Figure out the task's cell, dsq and store the corresponding cpumask in the
  * task_ctx.
-*/
+ */
 static inline int update_task_cell(struct task_struct *p, struct task_ctx *tctx,
 				   struct cgroup *cg)
 {
@@ -641,13 +644,15 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
 		basis_vtime = READ_ONCE(cctx->vtime_now);
 	}
 
+	tctx->basis_vtime = basis_vtime;
+
 	if (time_after(vtime, basis_vtime + 8192 * slice_ns)) {
 		scx_bpf_error("vtime is too far in the future for %d", p->pid);
 		goto out;
 	}
 	/*
-     * Limit the amount of budget that an idling task can accumulate
-     * to one slice.
+	 * Limit the amount of budget that an idling task can accumulate
+	 * to one slice.
 	 */
 	if (time_before(vtime, basis_vtime - slice_ns))
 		vtime = basis_vtime - slice_ns;
@@ -696,17 +701,17 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
 	}
 
 	/*
-     * The move_to_local can fail if we raced with some other cpu in the cell
-     * and now the cell is empty. We have to ensure to try the cpu_dsq or else
-     * we might never wakeup.
-     */
+	 * The move_to_local can fail if we raced with some other cpu in the cell
+	 * and now the cell is empty. We have to ensure to try the cpu_dsq or else
+	 * we might never wakeup.
+	 */
 	if (!scx_bpf_dsq_move_to_local(min_vtime_dsq))
 		scx_bpf_dsq_move_to_local(dsq);
 }
 
 /*
  * On tick, we apply CPU assignment
-*/
+ */
 void BPF_STRUCT_OPS(mitosis_tick, struct task_struct *p_run)
 {
 	if (bpf_get_smp_processor_id())
@@ -716,7 +721,7 @@ void BPF_STRUCT_OPS(mitosis_tick, struct task_struct *p_run)
 	if (local_configuration_seq == READ_ONCE(applied_configuration_seq))
 		return;
 
-	// Get the root cell (cell 0) and its cpumask
+	/* Get the root cell (cell 0) and its cpumask */
 	struct cell_cpumask_wrapper *root_cell_cpumaskw;
 	int zero = 0;
 	if (!(root_cell_cpumaskw =
@@ -743,9 +748,11 @@ void BPF_STRUCT_OPS(mitosis_tick, struct task_struct *p_run)
 	}
 	bpf_cpumask_copy(root_bpf_cpumask, (const struct cpumask *)all_cpumask);
 
-	// Iterate through the rest of the cells and (if in_use), clear their cpus
-	// from the root cell and assign them to the correct core
-	// TODO: Handle freed cells by giving their cores back to the root cell
+	/*
+	 * Iterate through the rest of the cells and (if in_use), clear their cpus
+	 * from the root cell and assign them to the correct core
+	 * TODO: Handle freed cells by giving their cores back to the root cell
+	 */
 	int cell_idx;
 	bpf_for(cell_idx, 1, MAX_CELLS)
 	{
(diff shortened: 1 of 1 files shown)


dschatzberg:
@htejun - wouldn't mind a scan of particularly the 2nd commit - it ended up being easier than I expected which makes me think I might have missed something significant

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -563,43 +559,75 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
...
-	if (p->flags & PF_KTHREAD && p->nr_cpus_allowed == 1) {
-		scx_bpf_dsq_insert(p, HI_FALLBACK_DSQ, slice_ns, 0);
-		cstat_inc(CSTAT_HI_FALLBACK_Q, tctx->cell, cctx);
-	} else if (!tctx->all_cpus_allowed) {
-		// FIXME: With cpusets, most schedules will fall into this section and
-		// not actually get distributed to the correct cell. We need to loosen
-		// the check on tctx->all_cpus_allowed
-		scx_bpf_dsq_insert(p, LO_FALLBACK_DSQ, slice_ns, 0);
-		cstat_inc(CSTAT_LO_FALLBACK_Q, tctx->cell, cctx);
+	if (!wakeup) {
```
htejun: This doesn't cover all cases. Maybe use `__COMPAT_is_enq_cpu_selected()`?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -612,13 +640,37 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+	{
+		min_vtime = p->scx.dsq_vtime;
+		min_vtime_dsq = cell;
+		found = true;
+		break;
+	}
 
-	scx_bpf_dsq_move_to_local(LO_FALLBACK_DSQ);
+	u64 cpu_dsq = pcpu_dsq_id(cpu);
+	bpf_for_each(scx_dsq, p, cpu_dsq, 0)
```
htejun: Maybe skip the iteration if `found` is `true`? The iteration is relatively expensive involving grabbling and releasing the DSQ lock. Hopefully, in the future, we should have cheaper `scx_bpf_dsq_peek()`.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -712,15 +764,24 @@ void BPF_STRUCT_OPS(mitosis_tick, struct task_struct *p_run)
...
 
+	/*
+        * Update both the CPU's cell and the cpu's vtime so the vtime's are
+        * comparable at dispatch time.
+        */
 	if (time_before(cell->vtime_now, p->scx.dsq_vtime))
 		cell->vtime_now = p->scx.dsq_vtime;
 
+	if (time_before(cctx->vtime_now, p->scx.dsq_vtime))
+		cctx->vtime_now = p->scx.dsq_vtime;
```
htejun: Also, I'm not sure this is enough. This would work while the cell CPU allocation stays stable but when the assignment changes, there can be a large delta which can lead to long stalls.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -563,43 +559,75 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
...
 
 	/*
-	 * If we aren't in the wakeup path, layered_select_cpu() hasn't run and thus
-	 * we haven't looked for and kicked an idle CPU. Let's do it now.
+     * Limit the amount of budget that an idling task can accumulate
+     * to one slice.
 	 */
-	if (!(enq_flags & SCX_ENQ_WAKEUP))
-		pick_idle_cpu_and_kick(p, task_cpu, cctx, tctx);
+	if (time_before(vtime, basis_vtime - slice_ns))
```
htejun, line 657: I'm not sure this works. The pcpu's vtimeline is kept close with the vtimeline of the cell that the CPU belongs to; however, any cell can insert to the pcpu queue and their vtime would have nothing to do with the pcpu's vtimeline - ie. it could easily be far enough into the future from pcpu vtime's POV and if that CPU stays busy from tasks in the owning cell, that task can easily stall. You can limit how far into the future the vtime could be in and that wouldn't lead to stalls but again the vtime itself wouldn't really mean anything. You can calculate the budget or deficit that the task has and try to translate to the pcpu's vtimeline, but again, the budget/deficit numbers can only be translated across vtimeslines if they have some mapping relationship - ie. one vtimeline is twice as high priority as the other vtimeline or somesuch. Otherwise, the mapping of the number wouldn't carry any meaning.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Hmm, if `found` is true then I do need to see if the first element of the pcpu dsq has lower vtime so I think the iteration is unavoidable

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Correct, right now CPU allocation is stable - so not an issue but it will need to adjust when CPU allocation changes.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Hmm, so I'm aware of the fact that any cell can insert to the pcpu queue but I don't see a case where the task's cell's vtime is compared to the cpu's vtime. It's true that I would need to handle this properly where there's a taskset or where cpu <=> cell assignments change (as that could make a pinned task into an unpinned task, or vice versa).

Basically, once a task is sent to a pcpu queue - it should always get sent to some pcpu queue unless some explicit event happens (taskset, cpu <=> cell assignment changes) so the only thing to concern ourselves with is pcpu vtime comparison.

I think the problem you're highlighting might still be an issue with a task pinned to e.g. 2 cpus since it could be scheduled to either CPU and the vtime between the two would need to be comparable (it isn't today).

dschatzberg: > You can calculate the budget or deficit that the task has and try to translate to the pcpu's vtimeline, but again, the budget/deficit numbers can only be translated across vtimeslines if they have some mapping relationship - ie. one vtimeline is twice as high priority as the other vtimeline or somesuch. Otherwise, the mapping of the number wouldn't carry any meaning.

I didn't quite follow this point - why doesn't a naive translation still hold up?

htejun: > Hmm, so I'm aware of the fact that any cell can insert to the pcpu queue but I don't see a case where the task's cell's vtime is compared to the cpu's vtime. It's true that I would need to handle this properly where there's a taskset or where cpu <=> cell assignments change (as that could make a pinned task into an unpinned task, or vice versa).
>
> Basically, once a task is sent to a pcpu queue - it should always get sent to some pcpu queue unless some explicit event happens (taskset, cpu <=> cell assignment changes) so the only thing to concern ourselves with is pcpu vtime comparison.

Yeah, tasks' affinities and/or cell CPU assignments can change and these things can bounce around between different vtimelines.
 
> I think the problem you're highlighting might still be an issue with a task pinned to e.g. 2 cpus since it could be scheduled to either CPU and the vtime between the two would need to be comparable (it isn't today).

Yes, this too.

htejun: > > You can calculate the budget or deficit that the task has and try to translate to the pcpu's vtimeline, but again, the budget/deficit numbers can only be translated across vtimeslines if they have some mapping relationship - ie. one vtimeline is twice as high priority as the other vtimeline or somesuch. Otherwise, the mapping of the number wouldn't carry any meaning.
> 
> I didn't quite follow this point - why doesn't a naive translation still hold up?

Absolute vtime values can diverge significantly - e.g. a vtimeline which contains significant portion of low priority tasks would progress way faster than a vtimeline with high pri tasks. The relative position in the timeline is more translatable. However, this may not be one to one. For example if a vtimeline is higher priority than the other vtimeline, the budget translation may need to factor in the weight difference. If all vtimelines are more or less homogeneous, yeah, naive budget translation should work (that's also what layered does when tasks are moved across llc boundaries).

dschatzberg:
I'm going to rework things a bit so each task is exclusively owned by a pcpu dsq or a per-cell dsq and then just adjust vtimes when migrating. I think that coupled with the existing logic to ensure per cpu and per cell dsqs advance vtime together should work

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -349,17 +386,43 @@ static inline int update_task_cpumask(struct task_struct *p,
...
+			bpf_cpumask_subset(cell_cpumask, p->cpus_ptr);
+
+	/*
+	 * XXX - To be correct, we'd need to calculate the vtime
+	 * delta in the previous dsq, scale it by the load
+	 * fraction difference and then offset from the new
+	 * dsq's vtime_now. For now, just do the simple thing
+	 * and assume the offset to be zero.
+	 *
+	 * Revisit if high frequency dynamic cell switching
```
tommy-u, line 401: If I'm reading right, this is about dropping the lag/lead a task had in its prior DSQ. Maybe this will become relevant if we implement work stealing (between L3 DSQs) which could induce higher frequency switches.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -712,15 +789,24 @@ void BPF_STRUCT_OPS(mitosis_tick, struct task_struct *p_run)
...
 
+	/*
+	 * Update both the CPU's cell and the cpu's vtime so the vtime's are
+	 * comparable at dispatch time.
+	 */
 	if (time_before(cell->vtime_now, p->scx.dsq_vtime))
 		cell->vtime_now = p->scx.dsq_vtime;
 
+	if (time_before(cctx->vtime_now, p->scx.dsq_vtime))
+		cctx->vtime_now = p->scx.dsq_vtime;
```
tommy-u: Were we just missing this cctx update before? Or did changes in this PR make it relevant to update cctx vtime when it wasn't before?

tommy-u (approved):
Eliminates fallback queues. Makes explicit that a task is exclusively owned by a cell or cpu by adding `dsq` to the `task_ctx`. Cpumask updates are a bit more robust now that task vtimes are updated to cell/cpu dsq vtime.

Looks good to me

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Correct - for true vtime fairness we'd want to maintain the lag/lead across these switches

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: We didn't have percpu DSQs so there was no cctx vtime before this PR. It's just once we add it, it's necessary to keep it up to date with the cell

PR #2271, opened 2025-06-26
ci: move chaos to linux-rolling-stable

JakeHillion:
For some reason chaos is failing with the latest sched_ext/for-next kernel, see
https://github.com/sched-ext/scx/actions/runs/15892717558/job/44818708334?pr=2253 for an example. This is fully reproducible it seems and it's unclear why it's happening.

As I don't have time to debug this at the minute, drop chaos to linux-rolling-stable (6.15.3 at the minute) where it does pass CI so we can deal with it later. Currently chaos is blocking the for-next import.

Changes to the CI to enable this:
- Add `metadata.scx.ci.kernel.default` to specify a different default kernel for pull requests when the usual default of `sched_ext/for-next` is inappropriate.
- Add `metadata.scx.ci.kernel.allowlist` to specify an acceptable set of kernels when many are inappropriate.
- Add similar `metadata.scx.ci.kernel.blocklist` to avoid testing on a kernel when it's known to fail.

The new metadata is managed in each scheduler's Cargo.toml rather than having to change a central list in the CI spec and removes some edge cases.

Most of this change was tested in #2246 but we can't land that yet due to vmlinux issues. Keep all the new behaviour, even though we don't need `allowlist` except for wd40.

Test plan:
- Ran `.github/include/list-integration-tests.py` with various kernels. It worked.
- Checked the new `scx_chaos` entry on this PR. It selects the correct kernel in the "Load kernel" step.

### DIFF
--- a/.github/include/flake.nix
+++ b/.github/include/flake.nix
@@ -124,6 +124,8 @@
               pyproject = false;
               dontUnpack = true;
 
+              propagatedBuildInputs = with pkgs; [ cargo ];
+
               installPhase = "install -Dm755 ${./list-integration-tests.py} $out/bin/list-integration-tests";
             };
 
--- a/.github/include/list-integration-tests.py
+++ b/.github/include/list-integration-tests.py
@@ -7,6 +7,35 @@ import subprocess
 import sys
 
 
+def get_package_kernel_requirements():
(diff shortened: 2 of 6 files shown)


hodgesds (approved):
LGTM

PR #2273, opened 2025-06-26
ci: deduplicate kernel specific matrix entries properly

JakeHillion:
A recent change made it possible to specify different default kernels for the standard PR CI job for each scheduler. This broke the deduplication of the "default" jobs, meaning commits with trailers were triggering way more tests than necessary. Clean this up by removing the kernel argument from the matrix if it matches the scheduler's default choice.

Test plan:
- This CI job. It has a tonne of test kernels. With the old behaviour we'd see 3x the normal jobs, the new behaviour should be that chaos doesn't get a for-next job and that we only have 2x.

CI-Test-Kernel: sched_ext/for-next
CI-Test-Kernel: stable/linux-rolling-stable

### DIFF
--- a/.github/include/list-integration-tests.py
+++ b/.github/include/list-integration-tests.py
@@ -118,11 +118,8 @@ def main():
     kernels_to_test = {default_kernel}
     kernels_to_test.update(trailer_kernels)
 
-    matrix = []
+    matrix = set()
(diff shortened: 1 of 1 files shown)


etsal (approved):
Nice!

PR #2274, opened 2025-06-26
scxtop: sanitize no break space

yaakov-stein:
Following up from #2209 and #2248, this fix allows us to return to using SystemLocale to better suit each user's preferences. A more permanent fix will come with the next ratatui version (see [issue](https://github.com/ratatui/ratatui/issues/1928), [PR](https://github.com/ratatui/ratatui/pull/1934)), but it's unclear exactly when that might happen.

Testing:

Used Locale::fr to ensure that narrow no-break whitespaces (\u{202F}) were being properly replaced with standard whitespaces. Ex.:
Screenshot 2025-06-26 at 3 04 15 PM

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -12,6 +12,7 @@ use crate::config::get_config_path;
 use crate::config::Config;
 use crate::format_hz;
 use crate::read_file_string;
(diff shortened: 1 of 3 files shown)


etsal (approved):
LGTM!

PR #2275, opened 2025-06-26
lib: dynamically allocate minheap and add minheap/ATQ fixups

etsal:
Minheaps are currently statically allocated because the verifier does not accept data structures with a variable array at the end. Bite the bullet and make the minheap's element array into a separate allocation referred to with a pointer. This allows us to bump the maximum task capacity of an ATQ significantly. Also add a method for dumping the minheap's contents to help with debugging, and fix a minheap bug assigning elem to the weight field of an helem.

### DIFF
--- a/lib/atq.bpf.c
+++ b/lib/atq.bpf.c
@@ -17,7 +17,7 @@ u64 scx_atq_create_internal(bool fifo)
 	if (!atq)
 		return (u64)NULL;
 
-	atq->heap = scx_minheap_alloc(SCX_MINHEAP_MAX_CAPACITY);
+	atq->heap = scx_minheap_alloc(SCX_ATQ_MAX_CAPACITY);
 	if (!atq->heap)
 		return (u64)NULL;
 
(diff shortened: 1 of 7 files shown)


yaakov-stein (approved):
Minheap updates LGTM

lib/atq.bpf.c
```diff
@@ -0,0 +1,97 @@
...
+	return (u64)atq;
+}
+
+__hidden
+int scx_atq_insert(scx_atq_t *atq, u64 taskc_ptr)
+{
+	int ret;
+
+	if (!atq->fifo)
+		return -EINVAL;
```
hodgesds: I was wondering if this should `scx_bpf_error` because at that point the scheduling logic is just wrong with respect to the DSQs. It's probably fine for now, but I'd rather have things blow up when you forget the type of DSQ and do something silly. Will hopefully save someone some debugging pain if return codes aren't checked on insert.

etsal:
> LGTM, no `scx_atq_peek`?

Will do!

lib/atq.bpf.c
etsal: I agree, but for now we are unable to run scx_bpf_error from the libraries because the code may be called from a non-scheduling context. Of course in practice ATQs are a scheduling-only abstraction, but I'd rather fix the issue of error handling all at once across the libraries once we get BPF Streams. When we do, we can have a thread waiting on the stderr stream from userspace and force a scheduler exit/abort when something is written to it.

PR #2277, opened 2025-06-27
scx_lavd: Use the PCO (Performance vs. CPU Order) table when calculating the number of active CPUs.

multics69:
When the energy model is available, use the upper bounds of the lowest
performance state and the lower bounds of the highest performance state
as the thresholds for the power save and performance modes, respectively.
Use the PCO table so that we can also utilize the middle states in
the PCO table.

When switching to a different performance mode (e.g., power save -> balanced),
use the required compute capacity, which is compatible with the PCO table,
instead of relying on CPU utilization based on wall-clock time. Clean up the
calculation of avg_sc_util to use more time_delta() and remove redundant reads
of volatile variables, etc.

The old, unnecessary code is removed as now the code compiles successfully.
Now the scheduler uses the first PCO entry for powersave mode and the last
PCO entry for performance mode. It does not use the states in the middle yet.

Add a CLI option, "--no-use-em". When the option is given, do not use the
energy model but solely rely on the CPU topology or user-provided CPU
preference order ("--cpu-pref-order"). When "--cpu-pref-order" is given,
it implies "--no-use-em".

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/power.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/power.bpf.c
@@ -9,22 +9,11 @@
  */
 
 /*
- * CPU order
- */
-static u64		LAVD_AP_LOW_UTIL;
-
-/* CPU preference order for performance and balanced mode */
-const volatile u16	cpu_order_performance[LAVD_CPU_ID_MAX];
-
-/* CPU preference order for powersave mode */
-const volatile u16	cpu_order_powersave[LAVD_CPU_ID_MAX];
-
-
-/*
  * System-wide properties of CPUs
  */
 static bool		have_turbo_core;
 static bool		have_little_core;
+const volatile bool	is_smt_active;
 
 
 /*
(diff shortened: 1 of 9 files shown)


scheds/rust/scx_lavd/src/bpf/power.bpf.c
```diff
@@ -99,26 +139,82 @@ static int calc_nr_active_cpus(void)
...
+static u64 get_human_readable_avg_sc_util(u64 avg_sc_util)
+{
 	/*
-	 * Fill the required compute capacity in the CPU preference order,
-	 * utilizing each CPU in a certain % (LAVD_CC_PER_CORE_UTIL or
-	 * LAVD_CC_PER_CORE_SHIFT).
+	 * avg_sc_util is the utilization assuming all the CPUs are the
+	 * fastest ones (i.e., capacity = 1024) at the highest frequency.
+	 * Hence, when all CPUs are 100% utilized, the avg_sc_util is not
+	 * 1024, it is the sum of capacities of all CPUs. Printing
```
hodgesds: should this say online CPUs
```suggestion
	 * 1024, it is the sum of capacities of all online CPUs. Printing
```

scheds/rust/scx_lavd/src/bpf/power.bpf.c
```diff
@@ -591,28 +708,52 @@ static u64 scale_cap_freq(u64 dur, s32 cpu)
...
+		do_update_autopilot_high_cap();
+}
+
+static void init_autopilot_caps(void)
+{
+	if (no_use_em) {
 		/*
-		 * When there are little cores, we move up to the balanced mode
-		 * if one little core is fully utilized.
+		 * When the energy model is not available, rely on the heuritics.
```
hodgesds: nit
```suggestion
		 * When the energy model is not available, rely on the heuristics.
```

scheds/rust/scx_lavd/src/bpf/util.bpf.c
```diff
@@ -142,36 +141,50 @@ static struct cpu_ctx *get_cpu_ctx_task(const struct task_struct *p)
 	return get_cpu_ctx_id(scx_bpf_task_cpu(p));
 }
 
-static u32 calc_avg32(u32 old_val, u32 new_val)
+#define __calc_avg(old, new, decay) ({						\
```
hodgesds, line 144: nice cleanup!

hodgesds (approved):

PR #2278, opened 2025-06-27
scx_flash: Remove the tickless timer

arighi:
Replace the periodic timer used to check for contended CPUs with a direct check in ops.enqueue(): when enqueuing a task, if its previously used CPU is running a tickless task (a task with an infinite time slice), assign it a regular time slice. This resumes the tick on that CPU, effectively re-enabling preemption for contended CPUs.

With this, the global tickless timer is no longer needed. Consequently, the --tickless_ns option can be simplified to a --tickless boolean flag.

Suggested-by: Changwoo Min <changwoo@igalia.com>

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -99,6 +99,11 @@ const volatile bool local_pcpu;
 const volatile bool native_priority;
 
 /*
+ * Enable tickless mode.
+ */
+const volatile bool tickless_sched;
+
+/*
  * The CPU frequency performance level: a negative value will not affect the
  * performance level and will be ignored.
  */
@@ -141,11 +146,6 @@ static inline void set_throttled(bool state)
 {
 	WRITE_ONCE(cpus_throttled, state);
 }
-
-/*
- * Tickless CPU contention check frequency.
- */
-const volatile u64 tickless_ns;
 
 /*
  * Exit information.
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1131,6 +1117,25 @@ static bool is_interactive(const struct task_struct *p, const struct task_ctx *t
...
+{
+	struct task_struct *curr;
+
+	if (!tickless_sched)
+		return;
+
+	bpf_rcu_read_lock();
+	curr = scx_bpf_cpu_rq(cpu)->curr;
+	if (curr->scx.slice == SCX_SLICE_INF)
+		curr->scx.slice = task_slice(cpu);
```
JakeHillion, line 1135: It doesn't really matter, but is it worth:
```suggestion
        __sync_val_compare_and_swap(&curr->scx.slice, SCX_SLICE_INF, task_slice(cpu));
```
To avoid confusion if slice gets updated elsewhere in future and this races?

arighi: Hey @JakeHillion  thanks for taking a look at this!

I’ve considered making this more atomic, but in the end, I don’t think it’s strictly necessary. If the running task has an infinite slice, the tick is disabled, so the time slice doesn’t change. We might have concurrent calls to `ops.enqueue()` from different CPUs, but in the worst case, multiple CPUs just end up writing the same value concurrently. That leads to unnecessary write operations, but the time slice itself should remain consistent.

Using an atomic operation could avoid the redundant writes, but it might also be more expensive in general. In the end, I think either approach is fine and the non-atomic version is probably easier to understand. What do you think?

JakeHillion: Sounds good, I agree that the race is irrelevant in the current code, as it only ever gets updated from a single value to a single value.

PR #2279, opened 2025-06-27
scx_flash: Fix no-SMT sync wakeup logic

arighi:
We don't want to completely skip reusing the previous CPU when SMT is disabled. In this case, just check whether the previous CPU is idle, ignoring the SMT cpumask.

This implements a consistent sync wakeup logic also with SMT disabled.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -735,7 +735,7 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
 		 * than cache hot data in the wakee's CPU.
 		 */
 		if (share_llc &&
-		    (smt_enabled && bpf_cpumask_test_cpu(prev_cpu, idle_smtmask)) &&
+		    (!smt_enabled || bpf_cpumask_test_cpu(prev_cpu, idle_smtmask)) &&
 		    scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
 			cpu = prev_cpu;
 			*is_idle = true;


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -735,7 +735,7 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
 		 * than cache hot data in the wakee's CPU.
 		 */
 		if (share_llc &&
-		    (smt_enabled && bpf_cpumask_test_cpu(prev_cpu, idle_smtmask)) &&
+		    (!smt_enabled || bpf_cpumask_test_cpu(prev_cpu, idle_smtmask)) &&
```
multics69, line 738: Hmm... I am not able to follow why `!smt_enabled ||` is required. Can you explain me more why SMT does matter here? I may miss something. :-)

arighi: The logic is the following:
 - if SMT is enabled we want to check if `prev_cpu` is a full-idle SMT core
 - if SMT is disabled we  just want to check if `prev_cpu` is idle (basically with `smt_enabled == false` we skip the `bpf_cpumask_test_cpu()` check

multics69: Now it is clear to me. It would be good to add comments on this.

multics69 (approved):

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: @multics69 this PR was already merged so I sent another one on top https://github.com/sched-ext/scx/pull/2288

PR #2283, opened 2025-06-27
scxtop: ensure loud failure

yaakov-stein:
Currently, we can mask bugs by using `.or_insert()` when searching maps. The node, llc, and cpu BTreeMaps are all fully populated in `App::new()` and it is a bug or unexpected behavior if we are searching for a node, llc, or cpu outside of the topology. To ensure we fail loudly, all `.or_insert()` 's were replaced by just getting the value and `.expect()` with an error message.


etsal (approved):
Nice!

PR #2287, opened 2025-06-28
scx_lavd: Handling the CPU utilization spikes in autopilot mode.

multics69:
When a CPU is almost fully utilized (say 90%) during a period, we may underestimate the utilization. For example, when the measured CPU utilization is 100%, there is a possibility that the actual utilization is actually higher, such as 130%.

To handle such utilization spike cases, we give a 50% premium for the scaled CPU time where the CPU is almost fully utilized. This overestimates the scaled CPU utilization and the required compute capacity. Then, this finally allocates more active CPUs. The over-allocated CPUs become the breathing room.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
@@ -30,6 +30,7 @@ struct sys_stat_ctx {
 	u64		compute_total;
 	u64		tot_svc_time;
 	u64		tot_sc_time;
+	u64		tsct_spike;
 	u64		nr_queued_task;
 	s32		max_lat_cri;
 	s32		avg_lat_cri;
@@ -198,6 +199,12 @@ static void collect_sys_stat(struct sys_stat_ctx *c)
 		 */
 		c->idle_total += cpuc->idle_total;
 		cpuc->idle_total = 0;
+
+		/*
+		 * Track the scaled time when the utilization spikes happened.
+		 */
+		if (cpuc->cur_util > LAVD_CC_UTIL_SPIKE)
+			c->tsct_spike += cpuc_tot_sc_time;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
```diff
@@ -213,17 +220,41 @@ static u64 clamp_time_slice_ns(u64 slice)
...
 
+	/*
+	 * When calculating the scaled utilization, which determines the
+	 * required compute capacity and the number of active CPUs, we add the
+	 * half-capacities of CPUs where utilization spikes occurred.
+	 *
+	 * Hence, when a CPU is almost fully utilized (say 90%) during a
+	 * period, we may underestimate the utilization. For example, when the
+	 * measured CPU utilization is 100%, there is a possibility that the
+	 * actual utilization is actually higher, such as 130%.
```
arighi: It's not very clear to me how the utilization can be > 100%, maybe you can clarify this a bit more in this comment?

multics69: Suppose that a CPU can provide the compute capacity upto 100 and task A running on the CPU A consumed the compute capacity 100. Then the measured CPU utilization is of course 100%.

However, what if task A is a CPU-bound, consuming a lot more CPU cycles? In that case, if task A is scheduled on a more powerful CPU B whose capacity is, say 200. Task A may consume 130 out of 200 on CPU B. In that case, the true capacity for task A should be 130, not 100. This is what I want to measure at a given moment to eventually calaucate the require capacity.

Does it make sense to you?

arighi: Oh I see, that's because of capacity comes into play, it's clear now, thanks!

arighi (approved):

PR #2289, opened 2025-06-29
scx_lavd: Be conservative about migrating to a waker's CPU.

multics69:
Previously, when selecting an idle CPU in the event of a synchronous wake-up, we treated the previous CPU and the waker's CPU equally, so triggering more task migrations.

Now, we treat them differently; we prefer to reuse the previous CPU and its domain. If that's not available, we will consider using waker's domain (not the CPU).

Additionally, when there is no fully idle CPU in the system, we won't chase a partially idle CPU in neighboring domains.


hodgesds (approved):
Makes sense!

multics69:
> In general, for what I've seen, the pure "signalling wakeup" seems to be more common in gaming, therefore this change totally makes sense for lavd.

I agree. Such a signaling wakeup seems to be the majority in general. I am not sure in the other domain. I *guess* the other workloads would be similar, like MySQL, Redis, etc.

PR #477, opened 2024-08-08
scx_rusty: Make layer matching a global function

hodgesds:
Layer matching currently takes a large number of bpf instructions. Moving layer matching to a global function will reduce the overall instruction count and allow for other layer matching methods such as glob.

main:
```
$ sudo veristat ./target/release/build/scx_layered-ed7ff1cb72366b2f/out/bpf.bpf.o 
Processing 'bpf.bpf.o'...
File       Program                Verdict  Duration (us)   Insns  States  Peak states
---------  ---------------------  -------  -------------  ------  ------  -----------
bpf.bpf.o  layered_cpu_release    success             24       2       0            0
bpf.bpf.o  layered_dispatch       success           4888    6472     651          274
bpf.bpf.o  layered_dump           success           9651   17363     465          192
bpf.bpf.o  layered_enqueue        success          27073    7539     629          343
bpf.bpf.o  layered_exit           success             36      25       2            2
bpf.bpf.o  layered_exit_task      success             65      46       4            4
bpf.bpf.o  layered_init           success           2503    4514     407          252
bpf.bpf.o  layered_init_task      success           1358      58       5            5
bpf.bpf.o  layered_quiescent      success            314     376      29           29
bpf.bpf.o  layered_runnable       success         946655  333037   13118         2304
bpf.bpf.o  layered_running        success            217     194      17           17
bpf.bpf.o  layered_select_cpu     success           4766     697      65           65
bpf.bpf.o  layered_set_cpumask    success             59      35       3            3
bpf.bpf.o  layered_set_weight     success             34      18       1            1
bpf.bpf.o  layered_stopping       success            139     226      16           16
bpf.bpf.o  layered_yield          success             88     126      10           10
bpf.bpf.o  sched_tick_fentry      failure              0       0       0            0
bpf.bpf.o  tp_cgroup_attach_task  success           7456     206      18           18
bpf.bpf.o  tp_task_rename         success             47      11       1            1
---------  ---------------------  -------  -------------  ------  ------  -----------
Done. Processed 1 files, 0 programs. Skipped 19 files, 0 programs.
```

this PR:
```
$ sudo veristat ./target/release/build/scx_layered-ed7ff1cb72366b2f/out/bpf.bpf.o 
Processing 'bpf.bpf.o'...
File       Program                Verdict  Duration (us)  Insns  States  Peak states
---------  ---------------------  -------  -------------  -----  ------  -----------
bpf.bpf.o  layered_cpu_release    success             21      2       0            0
[...]

PR #2294, opened 2025-06-30
Scx spark

esot0:


scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/* 
+ * Copyright (c) 2024 Andrea Righi <andrea.righi@linux.dev>
```
hodgesds, line 3: feel free to add your name as an author

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+	__type(key, u32);
+	__type(value, struct cpu_ctx);
+	__uint(max_entries, 1);
+} cpu_ctx_stor SEC(".maps");
+
+
+/*
+ * Detect workload type based on process name and behavior patterns.
+ */
+static void detect_workload_type(const struct task_struct *p, u32 pid, u32 tid) {
```
hodgesds, line 259: you may eventually want this function to be a bpf global function, see #477 for reference

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+	u64 last_cpu_access;
+};
+
+/*
+ * GPU detection kprobes.
+ */
+
+SEC("kprobe/nvidia_poll")
+int kprobe_nvidia_poll() {
+	// bpf_printk("nvidia_poll detected, saving pid/tid\n");
```
hodgesds:
```suggestion
	dbg_msg("nvidia_poll detected, saving pid/tid\n");
```

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+	return save_gpu_tgid_pid();
+}
+
+SEC("kprobe/nvidia_open")
+int kprobe_nvidia_open() {
+	// bpf_printk("nvidia_open detected, saving pid/tid\n");
```
hodgesds:
```suggestion
	dbg_msg("nvidia_open detected, saving pid/tid\n");
```

scheds/rust/scx_spark/test_workload_detection.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Test script for workload detection features in scx_spark
+# This script demonstrates how the scheduler can detect and optimize for different ML workload types
+
+set -e
+
+echo "=== SCX Spark Workload Detection Test ==="
+echo
+
+LHOME=localhome/local-emilys
```
etsal, line 11: Can we turn this into a configurable argument?

scheds/rust/scx_spark/test_workload_detection.sh
```diff
@@ -0,0 +1,50 @@
...
+    --workload-aware-scheduling \
+    --debug \
+    --verbose \
+    & SCHED_PID=$!
+
+# Wait for scheduler to start
+sleep 3
+
+echo "Scheduler started with PID: $SCHED_PID"
+/${LHOME}/llama.cpp/build/bin/llama-bench -m /${LHOME}/llama.cpp/models/Llama-3.2-1B-Instruct-Q6_K.gguf & LLAMA_PID=$!
```
etsal, line 43: Can dynamically pull in the dependency if it is not there?

scheds/rust/scx_spark/src/main.rs
```diff
@@ -0,0 +1,941 @@
...
+    // Convert the byte vector to a hexadecimal string.
+    let hex_str: String = bitmask.iter().rev().fold(String::new(), |mut f, byte| {
+        let _ = write!(&mut f, "{:02x}", byte);
+        f
+    });
+
+    format!("0x{}", hex_str)
+}
+
+/// scx_bpfland: a vruntime-based sched_ext scheduler that prioritizes interactive workloads.
```
etsal, line 169: I assume this was copied from scx_bpfland, maybe update the comment to reflect the new scheduler's functionality and explain the differences from bpfland? Imo a short version of the description from the README would do.

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+/*
+ * The CPU frequency performance level: a negative value will not affect the
+ * performance level and will be ignored.
+ */
+volatile s64 cpufreq_perf_lvl;
+
+/*
+ * Enable GPU support for task detection and prioritization.
+ */
+const volatile bool enable_gpu_support = false;
```
arighi, line 78: Minor style nit: no need to explicitly set it to false, it's implicitly initialized to false in the BSS.

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+
+/*
+ * Mask of Turbo (performance) CPUs.
+ */
+private(BPFLAND) struct bpf_cpumask __kptr *turbo_cpumask;
+
+/*
+ * DSQ dispatch mode.
+ */
+const volatile u32 dsq_mode = 0;
```
arighi, line 210: Ditto for the 0.

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+
+/*
+ * Detect workload type based on process name and behavior patterns.
+ */
+static void detect_workload_type(const struct task_struct *p, u32 pid, u32 tid) {
+	if (!enable_gpu_support)
+		return;
+		
+	struct workload_info info = {0};
+	u64 now = bpf_ktime_get_ns();
```
arighi, line 264: Any reason to not use `scx_bpf_now()` here?

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+	if (!p)
+		return;
+	
+	info.detection_time = now;
+	info.workload_type = WORKLOAD_TYPE_UNKNOWN;
+	
+	/* Use the comm field from the task_struct instead of bpf_get_current_comm */
+	char *comm = p->comm;
+	
+	/* Inference workload detection */
```
arighi, line 275: Clarified privately that these strcmp's are just placeholder code.

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+ */
+struct cpu_ctx *try_lookup_cpu_ctx(s32 cpu)
+{
+	const u32 idx = 0;
+	return bpf_map_lookup_percpu_elem(&cpu_ctx_stor, &idx, cpu);
+}
+
+/*
+ * Save current task's PID/TID to GPU maps when GPU operations are detected.
+ * Both entire processes using the GPU and single threads using the GPU will be tracked.
```
arighi, line 344: How do you plan to use the map of GPU PIDs/TIDs? Isn't enough to save the "GPU task" property in the task context?

Maybe clarify in the comment here.

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+static u32 get_workload_type(const struct task_struct *p) {
+	if (!enable_gpu_support)
+		return WORKLOAD_TYPE_UNKNOWN;
+		
+	u32 tid = p->pid;
+	u32 tgid = p->tgid;
+	struct workload_info *info;
+	
+	/* Check thread-specific workload info first */
+	info = bpf_map_lookup_elem(&workload_tid, &tid);
```
arighi, line 382: The workload_info can be added to the task context, we don't necessarily need a map here, right?

The reason is that most of the time we already have a  `struct task_ctx` pointer (`tctx`), so checking a flag there is much faster than doing another lookup in a BPF map.

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+
+/*Try looking for idle turbo CPUs, even if it's not a cache sibling. NOTE: This will probably happen dynamically based on cache sensitivity when workload detection improves.*/
+if(turbo_mask){
+    //Check if the previous CPU is idle and in the turbo mask
+	if (bpf_cpumask_test_cpu(prev_cpu, turbo_mask) && scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
+		*is_idle = true;
+		return prev_cpu;
+	}
+//Select any turbo CPU :
+	cpu = find_idle_cpu_in_mask(turbo_mask, 0);
```
arighi, line 915: Can we have turbo CPUs across multiple LLCs? A cross-LLC migration can be expensive.

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+	 * Try to re-use the same CPU if it's a big CPU.
+	 */
+	if (big_mask && bpf_cpumask_test_cpu(prev_cpu, big_mask) && scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
+		*is_idle = true;
+		return prev_cpu;
+	}
+
+	//All big CPUs are L3 cache siblings
+	if (big_mask) {
+		cpu = find_idle_cpu_in_mask(big_mask, 0);
```
arighi, line 931: Ditto for the big mask, it'd be nice (somehow) to prevent cross-LLC migrations (if they can happen).

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+void BPF_STRUCT_OPS(bpfland_running, struct task_struct *p)
+{
+	struct task_ctx *tctx;
+	u64 now = scx_bpf_now();
+	s32 cpu = scx_bpf_task_cpu(p);
+
+	__sync_fetch_and_add(&nr_running, 1);
+
+	/* If stay_with_kthread is enabled, track kthread activity */
+	if (stay_with_kthread && is_kthread(p)) {
```
arighi, line 1474: You may want to check `is_kthread(p) && p->nr_cpus_allowed == 1` here, because unbound kthreads can be pretty CPU-intensive.

scheds/rust/scx_spark/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1961 @@
...
+{
+	u64 now = scx_bpf_now(), slice, delta_runtime;
+	s32 cpu = scx_bpf_task_cpu(p);
+	struct cpu_ctx *cctx;
+	struct task_ctx *tctx;
+
+	__sync_fetch_and_sub(&nr_running, 1);
+
+	/* If stay_with_kthread is enabled, reset kthread activity flag */
+	if (stay_with_kthread && is_kthread(p)) {
```
arighi, line 1520: Ditto, Restrict this to per-CPU kthreads.

scheds/rust/scx_spark/src/bpf/main.bpf.c
esot0: It's some leftover code--when I was first experimenting with GPU support, I tried to see if I could take more from scx_layered's approach. For now, I've decided I just needed the simple tctx flag. I'll probably get rid of it in my next commit.

scheds/rust/scx_spark/src/bpf/main.bpf.c
esot0: No, this function just came to mind faster. I'll switch it out for scx_bpf_now().

scheds/rust/scx_spark/src/bpf/main.bpf.c
esot0: Makes sense, I'll add that in.

scheds/rust/scx_spark/src/bpf/main.bpf.c
esot0: There are only 5 turbo CPUs, all sharing the same LLC.

scheds/rust/scx_spark/src/bpf/main.bpf.c
esot0: Yes--the comment here isn't right. There are 5 big CPUS for each LLC. I'll try and add some logic to prioritize those that share an LLC first.

scheds/rust/scx_spark/src/main.rs
esot0: Will do, thank you for the suggestions!

scheds/rust/scx_spark/src/bpf/main.bpf.c
esot0: I'll look into it--thanks for the review!

PR #2296, opened 2025-06-30
scxtop: switch to use perfetto_protos crate

yaakov-stein:
Cleaning up the perfetto usage. This has three main benefits:
1) We no longer need to have the proto file in the repo and we can use the standard rust [documentation](https://docs.rs/perfetto_protos/latest/perfetto_protos/index.html) when updating the trace code.
2) This aligns the code more closely with the way [systing](https://github.com/josefbacik/systing/tree/main#) performs traces, making it easier to duplicate features between the two.
3) Build times should be much faster now.

These were largely non-functional changes (there is a place or two where I needed to make very slight changes, I'll comment in the code to highlight those).

Testing:
1) Ran a trace before and after. All data seems to have been retained (there is also no reason anything should have switched).


tools/scxtop/src/perfetto_trace.rs
```diff
@@ -270,10 +278,12 @@ impl PerfettoTraceManager {
         let trace_dsqs: Vec<u64> = self.dsq_nr_queued_events.keys().cloned().collect();
 
         fn timestamp_absolute_us(e: &TrackEvent) -> i64 {
-            use crate::protos::protos_gen::perfetto_scx::track_event::Timestamp;
+            use perfetto_protos::track_event::track_event::Timestamp;
             match e.timestamp {
                 Some(Timestamp::TimestampAbsoluteUs(t)) => t,
                 None | Some(Timestamp::TimestampDeltaUs(_)) => 0,
+                // e.timestamp is #[non-exhaustive] so we need this extra case.
+                Some(_) => 0,
```
yaakov-stein, line 286: As the comment notes, this was needed due to the fact that e.timestamp is #[non-exhaustive]. See [here](https://docs.rs/perfetto_protos/latest/perfetto_protos/track_event/track_event/enum.Timestamp.html).

tools/scxtop/src/perfetto_trace.rs
```diff
@@ -388,26 +391,24 @@ impl PerfettoTraceManager {
...
+                        })
+                        .unwrap_or_default(),
+                    ..FtraceEventBundle::default()
+                })),
                 ..TracePacket::default()
             });
         }
 
-        let out_bytes: Vec<u8> = self.trace.encode_to_vec();
+        let out_bytes: Vec<u8> = self.trace.write_to_bytes()?;
```
yaakov-stein, line 411: `encode_to_vec()` was not available in the perfetto_protos library, but [`write_to_bytes`](https://docs.rs/protobuf/latest/protobuf/trait.Message.html#method.write_to_bytes) accomplishes the same thing.

yaakov-stein:
> LGTM, hopefully this makes build times even faster

Just did some quick testing (not thorough at all), but it looks like most build times are down >50% when I make basic changes.

PR #2297, opened 2025-06-30
lib/minheap: Add helper to swap heap elements safely

EricccTaiwan:
Replace field-wise swaps with `scx_swap_minheap_arena()` for clarity and `__arena` safety.


EricccTaiwan:
Hmmm, let me fix this.

PR #2300, opened 2025-07-01
scx_utils, scx_lavd: Use ftrace for futex tracing whenever possible.

multics69:
Since ftrace has a lower overhead than tracepoint, use ftrace for futex tracing whenever possible. If any functions for futex tracing are not available (e.g., due to function inlining), fall back to using tracepoints.


multics69:
> Do you have some numbers? I thought ftrace was more expensive than tracepoints, due to the ftrace handler calls...
> 
> If you have numbers and this improves performance, then LGTM.

Tracepoint is slower since it copies the arguments to the struct.

PR #2301, opened 2025-07-01
lib/minheap: Add scx_minheap_swap_elems helper

EricccTaiwan:
Deduplicate elem/weight swapping in the two balance helpers.

I originally wanted to put the swap helper in `lib/minheap.h`, but the Rust tests kept failing and I’m not sure why.

### DIFF
--- a/lib/minheap.bpf.c
+++ b/lib/minheap.bpf.c
@@ -10,6 +10,20 @@
 
 #include <lib/minheap.h>
 
+static __always_inline void
+scx_minheap_swap_elems(struct scx_minheap_elem __arena *a __arg_arena,
+                       struct scx_minheap_elem __arena *b __arg_arena)
+{
+        u64 tmp_elem = a->elem;
+        u64 tmp_weight = a->weight;
+
+        a->elem = b->elem;
+        a->weight = b->weight;
+
+        b->elem = tmp_elem;
+        b->weight = tmp_weight;
+}
+
 __weak
 u64 scx_minheap_alloc_internal(size_t capacity)
 {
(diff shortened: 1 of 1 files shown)


etsal:
Seems reasonable, will test and merge. It's better to hold actual code in the C files so no need to try and put it in the header.

EricccTaiwan:
> Seems reasonable, will test and merge. It's better to hold actual code in the C files so no need to try and put it in the header.

Ok! Thanks.

etsal:
One thing, can you not tab align the assignments in scx_swap_elems? This is not really idiomatic in C.

etsal:
Also, can you rename the function to scx_minheap_swap_elems? You are correct that it is generic code, but the only  current user is the minheap code and I don't think it's a good idea to make a full-fledged library API call for such a simple operation. For this reason I think it should be clearly denoted as part of the API call.

EricccTaiwan:
Agreed. I will fix both in the morning. Thanks for pointing it out.

etsal:
> Agreed. I will fix both in the morning. Thanks for pointing it out.

No worries, and thank you for taking the time to look into this code!

etsal:
Looks good, can you also remove the tabs from inside scx_minheap_swap_elems?

EricccTaiwan:
> Looks good, can you also remove the tabs from inside scx_minheap_swap_elems?

~Are you referring to the swap part? If so, I have updated it!~
Edit: Oh, I see. You meant `u64 tmp_elem    = a->elem;` I have updated it!

Also, is there a specific criterion for when to use tabs versus spaces? 
For example, in `scx_minheap_alloc_internal`, there are only tabs used.

etsal:
> > Looks good, can you also remove the tabs from inside scx_minheap_swap_elems?
> 
> ~Are you referring to the swap part? If so, I have updated it!~ Edit: Oh, I see. You meant `u64 tmp_elem = a->elem;` I have updated it!
> 
> Also, is there a specific criterion for when to use tabs versus spaces? For example, in `scx_minheap_alloc_internal`, there are only tabs used.

As long as the style is consistent at least within a file it's fine.

etsal (approved):

Issue #2184, opened 2025-06-16
kernel: Implement lockless `scx_bpf_dsq_peek()`


htejun:
One common-ish use of `bpf_for(scx_dsq, ...)` is peeking the first element; however, using the DSQ iterator for peeking is not the most efficient as it involves locking the DSQ, finding the first task and then inserting the cursor to just re-lock the DSQ and remove it right after the first iteration. It should be possible to implement `scx_bpf_dsq_peek()` locklessly either by making the DSQ task list RCU safe or maintaining a first task pointer.

multics69:
As it is implemented, let's close it.

PR #2302, opened 2025-07-01
lib/atq: add peek operation

etsal:
Following up on the discussion from #2275, add a peek operation to the ATQ abstraction.


hodgesds (approved):
LGTM

lib/atq.bpf.c
```diff
@@ -89,6 +89,28 @@ u64 scx_atq_pop(scx_atq_t *atq)
 	return helem.elem;
 }
 
+__weak
+u64 scx_atq_peek(scx_atq_t *atq)
+{
+	u64 elem;
+	int ret;
+
+	ret = arena_spin_lock(&atq->lock);
```
yaakov-stein, line 98: Based on #2184, should this be lockless? Both for performance and to match the kernel's peek operation (whenever it is added)?

etsal: I can add a peek-and-lock call to return with the lock taken to prevent racing with other pop() operations. I do not think we have to make this lockless yet, because:

a) ATQ peek is more lightweight than DSQ peek in terms of work done, since it is essentially an atomic + a minheap operation,
b) it is not very likely this locking will be a significant overhead during the scheduler operations.

yaakov-stein: Sounds good! Makes sense to me to keep it as is for now

PR #2303, opened 2025-07-01
scx_flash: CPU stickiness

arighi:
A set of changes to make tasks more sticky to their CPU, improving cache locality, without regressing work conservation or responsiveness under high load.

The key idea is to favor per-CPU DSQs over per-node DSQs, as long as the CPU isn't saturated. This helps avoid unnecessary migrations. Example: when a short per-CPU kthread interrupts a task for a really short amount of time, the task is queued to a per-node DSQ and potentially picked up by another CPU, triggering a migration.

Instead, when the CPU becomes saturated, the scheduler switches to using per-node DSQs, increasing the chances of task migration. This boosts work conservation and maintains responsiveness under high load conditions.

To determine when a CPU becomes saturated the scheduler dynamically adjusts the saturation threshold using this heuristic: a CPU is considered saturated based on the formula `100% - user utilization %`; as a result, CPUs spending more time idle or in system mode will favor task stickiness, while those spending more time in user mode will promote migration.

User can also set a fixed saturation threshold using the option `--cpu-busy-thresh N` (with `N` being the saturation percentage).

Results:
  - before:
```
> ./schbench -L -m 4 -M auto -t 64 -n 0
auto pinning message and worker threads
Pinning to message thread index 1 cpu 1
Pinning to message thread index 2 cpu 2
Pinning to message thread index 0 cpu 0
Pinning to message thread index 3 cpu 3
Wakeup Latencies percentiles (usec) runtime 10 (s) (4642619 total samples)
	  50.0th: 55         (1384351 samples)
	  90.0th: 138        (1845799 samples)
	* 99.0th: 793        (415669 samples)
	  99.9th: 2356       (41705 samples)
	  min=1, max=15438
Request Latencies percentiles (usec) runtime 10 (s) (4649089 total samples)
	  50.0th: 339        (1399920 samples)
	  90.0th: 464        (1852548 samples)
	* 99.0th: 861        (415980 samples)
	  99.9th: 2468       (41626 samples)
	  min=102, max=11639
RPS percentiles (requests) runtime 10 (s) (11 total samples)
	  20.0th: 246528     (3 samples)
	* 50.0th: 518656     (3 samples)
	  90.0th: 543744     (4 samples)
	  min=246025, max=545217
sched delay: message 167 (usec) worker 87 (usec)
current rps: 529253.71
```

 - after:
```
> ./schbench -L -m 4 -M auto -t 64 -n 0
auto pinning message and worker threads
Pinning to message thread index 0 cpu 0
Pinning to message thread index 1 cpu 1
Pinning to message thread index 2 cpu 2
Pinning to message thread index 3 cpu 3
Wakeup Latencies percentiles (usec) runtime 10 (s) (13924693 total samples)
	  50.0th: 13         (3996277 samples)
	  90.0th: 42         (5561462 samples)
	* 99.0th: 123        (1240745 samples)
	  99.9th: 162        (120024 samples)
	  min=1, max=5372
Request Latencies percentiles (usec) runtime 10 (s) (14079726 total samples)
	  50.0th: 150        (4426557 samples)
	  90.0th: 167        (5386080 samples)
	* 99.0th: 193        (1261018 samples)
	  99.9th: 252        (117497 samples)
	  min=103, max=5432
RPS percentiles (requests) runtime 10 (s) (11 total samples)
	  20.0th: 996352     (3 samples)
	* 50.0th: 1538048    (4 samples)
	  90.0th: 1542144    (3 samples)
	  min=989399, max=1544828
sched delay: message 3 (usec) worker 5 (usec)
current rps: 1543326.27
```

In this case, with schbench, it's highly beneficial for tasks to remain on the same CPU, as the workload involves mostly inter-task synchronization (mostly CPU system time).

At the same time, the typical “gaming while recompiling the kernel” scenario still performs well, maintaining stable FPS without regressions, because when the CPUs get saturated with user CPU activity, the scheduler automatically falls back to using per-node DSQs.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -81,11 +81,14 @@ const volatile bool local_kthreads;
 /*
  * Prioritize per-CPU tasks (tasks that can only run on a single CPU).
  *
- * This allows to prioritize per-CPU tasks that usually tend to be
- * de-prioritized (since they can't be migrated when their only usable CPU
- * is busy). Enabling this option can introduce unfairness and potentially
- * trigger stalls, but it can improve performance of server-type workloads
- * (such as large parallel builds).
+ * Enabling this option allows to prioritize per-CPU tasks that usually
+ * tend to be de-prioritized, since they can't be migrated when their only
+ * usable CPU is busy.
+ *
+ * This is implemented by disabling direct dispatch when there are tasks
+ * queued to the per-CPU DSQ or the per-node DSQ. In this way, per-CPU
+ * tasks waiting in those queues are scheduled based solely on their
+ * deadline, avoiding further delays caused by direct dispatches.
  */
 const volatile bool local_pcpu;
 
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -38,12 +38,6 @@ const volatile bool rr_sched;
 /* Primary domain includes all CPU */
 const volatile bool primary_all = true;
 
-/* Enable per-CPU DSQs */
-const volatile bool pcpu_dsq = true;
-
-/* Enable per-node DSQs */
-const volatile bool node_dsq = true;
-
 /*
  * Default task time slice.
  */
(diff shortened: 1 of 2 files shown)


arighi:
Thanks @sirlucjan! Added you tested-by line.

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1111,15 +1145,41 @@ static bool try_direct_dispatch(struct task_struct *p, struct task_ctx *tctx,
 }
 
 /*
- * Return true if @tctx is an interactive task, false otherwise.
+ * Return true if a CPU is busy (based on its utilization), false
```
hodgesds, line 1148: I wonder if the per cpu queue depth is something to consider as well.

arighi: I've tried to add the queue depth to the equation as well, but it didn't perform well in the tests I ran.

What would be nice is to estimate the total wait time of the queue: for example, if there are many short-lived tasks, it might make sense to stay in the queue. On the other hand, if there are only a few tasks but long running, it’s probably better to fall back to the global queue instead.

To predict the total wait time we can probably take a look at the exec_runtime of each time (accumulated runtime since last sleep), but we need to iterate the queue using iterators and that might be expensive... so at the end what seemed to work best was just looking at the CPU load.

But it might be interesting to figure out a better way to estimate the total wait time and use this metric to determine if we want to stay on the same CPU or jump somewhere else.

PR #2306, opened 2025-07-01
scxtop: add cpu util events

yaakov-stein:
Following on #2290, #2293, (and somewhat #2282, more on that later), this adds `CpuUtilEvent`'s to allow checking cpu utilization at the CPU, LLC, and Node level within the event pane. The main challenge with this addition was ensuring that the LLC and Node cpu utilization statistics were correct and properly normalized (i.e., that the summed utilization was accurately divided by the relevant number of CPUs). Previously, `VecStats` was only used for aggregating raw totals across LLC/Node's (such as total cycles), and not for computing percentage-based metrics like CPU utilization.

To enable this, a `num_cpus` field was added to the `LlcData` and `NodeData` structs. This allows us to normalize utilization values by dividing them by the number of CPUs that belong to a given LLC or node, prior to calculating aggregate statistics. #2282 explored handling the normalization inside `VecStats`, but it led to a more complex and less readable design.

Overall look:

https://github.com/user-attachments/assets/2a9a9f09-e5cc-4c78-8152-5643c093efde

Running `schbench -t 60`:

https://github.com/user-attachments/assets/d419a8a7-4c94-4b23-89cf-f529b79e90ba

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -338,10 +338,10 @@ impl<'a> App<'a> {
         self.stop_prof_events();
         self.kprobe_links.clear();
 
-        let mut default_events = get_default_events(self.cpu_stat_tracker.clone());
+        self.available_events = get_default_events(self.cpu_stat_tracker.clone());
         let config_perf_events = PerfEvent::from_config(&self.config)?;
 
-        default_events.extend(
+        self.available_events.extend(
             config_perf_events
                 .iter()
                 .cloned()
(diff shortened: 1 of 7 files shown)


tools/scxtop/src/app.rs
```diff
@@ -785,16 +795,21 @@ impl<'a> App<'a> {
...
+            .expect("NodeData should have been present");
+        let divisor = match self.active_event {
+            ProfilingEvent::CpuUtil(_) => node_data.num_cpus,
+            _ => 1,
         };
-        let stats = VecStats::new(&data, 1, None);
+        let data = node_data
+            .event_data_immut(self.active_event.event_name())
+            .iter()
+            .map(|x| x / divisor as u64)
```
hodgesds, line 809: that's pretty clever

hodgesds (approved):

PR #2309, opened 2025-07-01
scxtop: fix scheduler load crash

yaakov-stein:
Previously, we used a `.blocking_lock()` in an async context (indirectly, from main.rs). This was causing the TUI to crash whenever we reloaded the stats client. Ex:

https://github.com/user-attachments/assets/63adadce-5f7e-40b3-86ab-bed26435bc68

By placing it inside `tokio::task::block_in_place`, we can now use `blocking_lock()` without causing a crash. This also adds a check in `on_tick()' to fully clear the schedule data when a schedule is unloaded. Ex:

https://github.com/user-attachments/assets/5d81699b-9ec7-4838-9ab1-3da9693293fd

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -516,8 +516,8 @@ impl<'a> App<'a> {
     fn reload_stats_client(&mut self) -> Result<()> {
         let stats_socket_path = self.config.stats_socket_path();
         let mut new_client = StatsClient::new();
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
Nice!

yaakov-stein:
A nice-to-have would be if a special message pops up and says something along the lines of “To view scheduler data, make sure a scx scheduler is loaded.”

PR #2316, opened 2025-07-02
scx_flash: Prevent potential stall with --rr-sched

arighi:
In round-robin mode we are always re-enqueueing tasks to the local DSQ. However, if a CPU is stolen by a higher scheduling class (i.e., real-time), we want to give the task a chance to move to a different CPU, to reduce the risk of stalls.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -1218,21 +1218,33 @@ void BPF_STRUCT_OPS(flash_enqueue, struct task_struct *p, u64 enq_flags)
 	preempt_curr(prev_cpu);
 
 	/*
-	 * Keep reusing the same CPU in round-robin mode.
-	 */
-	if (rr_sched) {
-		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | prev_cpu, slice_max, enq_flags);
-		__sync_fetch_and_add(&nr_direct_dispatches, 1);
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1217,22 +1217,34 @@ void BPF_STRUCT_OPS(flash_enqueue, struct task_struct *p, u64 enq_flags)
...
 	 */
 	if (rr_sched) {
+		/*
+		 * In case of a re-enqueue due to a higher scheduling class
+		 * stealing the CPU, allow the task to move to a different
+		 * CPU.
+		 */
+		if (enq_flags & SCX_ENQ_REENQ) {
+			scx_bpf_dsq_insert(p, SCX_DSQ_GLOBAL, slice_max, enq_flags);
+			__sync_fetch_and_add(&nr_direct_dispatches, 1);
```
hodgesds, line 1238: is this the right stat?

hodgesds (approved):
LGTM

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: More or less :), in round-robin mode I want to make sure none of the per-CPU / per-node DSQs are used, so I'm using the direct dispatch stat. Maybe I should add a nr_global_dispatches... but also --rr-sched should be rarely used and it's mostly to run experiments, so I'm not sure if it's worth adding another stat.

PR #2317, opened 2025-07-02
scx_p2dq: minor select_cpu path cleanup

yaakov-stein:
`cpu` is set to `prev_cpu` in the beginning of both `select_cpu` methods, so unless it is possible it has changed, we don't need to reset it to `prev_cpu`.


hodgesds (approved):
LGTM

PR #2318, opened 2025-07-02
scx_flash: Disable NUMA optimizations on non-NUMA systems

arighi:
NUMA optimizations are only introducing overhead if systems that don't have multiple nodes.

Therefore, automatically disable such optimizations on non-NUMA systems or when --disable-numa is specified.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -370,6 +370,40 @@ static bool is_queued(const struct task_struct *p)
 }
 
 /*
+ * Return the cpumask of full-idle SMT CPUs associated to @node.
+ *
+ * If NUMA support is disabled, @node is ignored.
+ */
+static const struct cpumask *get_idle_smtmask_node(int node)
+{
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -369,6 +369,41 @@ static bool is_queued(const struct task_struct *p)
...
+
+/*
+ * Return an idle CPU within the @cpus_allowed mask and @node.
+ *
+ * If NUMA support is disabled, @node is ignored.
+ */
+static s32 pick_idle_cpu_node(const struct cpumask *cpus_allowed, int node, u64 flags)
+{
+	return numa_disabled ?
+		// scx_bpf_pick_idle_cpu(cpus_allowed, flags & ~__COMPAT_SCX_PICK_IDLE_IN_NODE) :
```
hodgesds: delete?

arighi: oh good catch, I'll delete that, thanks!

hodgesds (approved):
LGTM

PR #2323, opened 2025-07-03
scx_bpfland: Refactor code to use calloc_cpumask

palashahuja:
This update replaces the manual creation and exchange of a CPU-mask with a single call to the helper function `calloc_cpumask`.


palashahuja:
Sure, will send the PR :)

palashahuja:
I couldn't find any instances where I could replace the invocations by `calloc_cpumask`. Could you help me point out where you intend to do the refactor?

PR #2325, opened 2025-07-03
scx_flash: Handle empty NUMA nodes

arighi:
Certain systems may have NUMA nodes that don't have any CPU assigned to them. These empty nodes shouldn't be counted when determining the total number of NUMA nodes in the system.

### DIFF
--- a/scheds/rust/scx_flash/src/main.rs
+++ b/scheds/rust/scx_flash/src/main.rs
@@ -381,7 +381,12 @@ impl<'a> Scheduler<'a> {
             }
         }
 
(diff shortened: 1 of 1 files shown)


arighi:
> Makes sense, there's always some weird hardware combination.

Yep, empty NUMA nodes are actually quite common in the NVIDIA platforms (like Grace Hopper), the GPU memory is represented using NUMA abstraction, even though the nodes don't have CPUs or system RAM assigned. So, from a scheduler perspective we need to ignore those nodes.

PR #2299, opened 2025-07-01
scx_p2dq: Fix weight handling

hodgesds:
Currently the task weight is handled inversely where low priority (weight) tasks are given priority. Tested with `stress-ng` with a background workload using all CPUs.

`eevdf`:
```
$ nice -n 10 stress-ng -c 8 -t 10 -M
stress-ng: info:  [32163] setting to a 10 secs run per stressor
stress-ng: info:  [32163] dispatching hogs: 8 cpu
stress-ng: metrc: [32163] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s CPU used per       RSS Max
stress-ng: metrc: [32163]                           (secs)    (secs)    (secs)   (real time) (usr+sys time) instance (%)          (KB)
stress-ng: metrc: [32163] cpu               17790     10.03     11.53      0.01      1774.10        1542.06        14.38          6600
stress-ng: info:  [32163] skipped: 0
stress-ng: info:  [32163] passed: 8: cpu (8)
stress-ng: info:  [32163] failed: 0
stress-ng: info:  [32163] metrics untrustworthy: 0
stress-ng: info:  [32163] successful run completed in 10.13 secs
$ nice -n -10 stress-ng -c 8 -t 10 -M
stress-ng: info:  [32201] setting to a 10 secs run per stressor
stress-ng: info:  [32201] dispatching hogs: 8 cpu
stress-ng: metrc: [32201] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s CPU used per       RSS Max
stress-ng: metrc: [32201]                           (secs)    (secs)    (secs)   (real time) (usr+sys time) instance (%)          (KB)
stress-ng: metrc: [32201] cpu              123683     10.00     79.67      0.02     12367.61        1552.05        99.61          6696
stress-ng: info:  [32201] skipped: 0
stress-ng: info:  [32201] passed: 8: cpu (8)
stress-ng: info:  [32201] failed: 0
stress-ng: info:  [32201] metrics untrustworthy: 0
stress-ng: info:  [32201] successful run completed in 10.01 secs
```

`p2dq`:
```
$ nice -n 10 stress-ng -c 8 -t 10 -M
stress-ng: info:  [32300] setting to a 10 secs run per stressor
stress-ng: info:  [32300] dispatching hogs: 8 cpu
stress-ng: metrc: [32300] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s CPU used per       RSS Max
[...]

PR #2326, opened 2025-07-04
scx_chaos: fix random delay stalling

yaakov-stein:
Random delays use a different vtime than p2dq that, in many cases, can be a much smaller number. In order to ensure this doesn't starve a task, the `p2dq_vtime` is saved and then reset in `chaos_dispatch`. However, when using `async_p2dq_enqueue` in `chaos_dispatch`, there were times when `scx_bpf_dsq_insert` would not properly move the task to a new dispatch queue (any ideas why this would fail??). When this occured, we had a task with a very large vtime (`p2dq_vtime`) on the delay queue that uses wall time, which caused the scheduler to stall.

This avoids the need to call `scx_bpf_dsq_insert` and instead moves the task directly to the local dsq. It also adds a lockless mark-and-sweep variable for good measure (I thought it made sense to avoid races, but that's not what prevented the stalls).

Even with p2dq vtime calculations updated in #2299, chaos was still stalling without this fix.

<details>
<summary>If you're interested in what I was seeing that made me conclude this was the problem:</summary>

When adding the following debugging lines in `chaos_dispatch`:

```
async_p2dq_enqueue_weak(&promise, p, taskc->enq_flags);
if (promise.kind == P2DQ_ENQUEUE_PROMISE_COMPLETE) {
	bpf_printk("DELAY_DSQ-ENQUEUE-PROMISE: kind=complete, dsq=0x%llx task: %s-%llu-%llu",
		p->scx.dsq->id,
		p->comm,
		p->pid,
		p->tgid);
}
complete_p2dq_enqueue_move(&promise, BPF_FOR_EACH_ITER, p);
bpf_task_release(p);
```

And the following lines in `async_p2dq_dispatch` (line [810](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_p2dq/src/bpf/main.bpf.c#L810)):

```
update_vtime(p, cpuc, taskc, llcx->vtime);
if (scx_bpf_test_and_clear_cpu_idle(cpu)) {
        u64 pre_dsq_id = p->scx.dsq->id;
        scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON|cpu, taskc->slice_ns, enq_flags);
        stat_inc(P2DQ_STAT_IDLE);
        bpf_printk("DELAY_DSQ: [p2dq-enqueue-6] pre-dsq: %llx, post-dsq: %llx, vtime: %llu, task: %s-%llu-%llu",
	        pre_dsq_id,
	        p->scx.dsq->id,
	        p->scx.dsq_vtime,
	        p->comm,
	        p->pid,
	        p->tgid);
        ret->kind = P2DQ_ENQUEUE_PROMISE_COMPLETE;
        return;
}
```

If you then find the first occasion of when the stalled task failed to be removed from the delay queue, you'll see something like this (with some other helpful logs):

```
sptp-1688427 [008] d.s.1 83586.441663: bpf_trace_printk: DELAY_DSQ-ENQUEUE-START-1: dsq: 10008, p->scx.dsq_vtime: 19578630807392904, taskc->p2dq_vtime: 225872528926995 task: sptp-1688431-1688422
sptp-1688427 [008] d.s.1 83586.441662: bpf_trace_printk: DELAY_DSQ-ENQUEUE-START-2: dsq: 10008, vtime: 83581052434161, task: sptp-1688431-1688422
sptp-1688427 [008] d...1 83586.441709: bpf_trace_printk: DELAY_DSQ-ENQUEUE-RESET: dsq: 10008, p->scx.dsq_vtime: 83581052434161, taskc->p2dq_vtime: 19578630807392904 task: sptp-1688431-1688422
sptp-1688427 [008] d...1 83586.441712: bpf_trace_printk: DELAY_DSQ: [p2dq-enqueue-6] pre-dsq: 10008, post-dsq: 10008, vtime: 36868751750957273, task: sptp-1688431-1688422
sptp-1688427 [008] d...1 83586.441713: bpf_trace_printk: DELAY_DSQ-ENQUEUE-PROMISE: kind=complete, dsq=0x10008 task: sptp-1688431-1688422
sptp-1688427 [008] d...1 83586.441715: bpf_trace_printk: DELAY_DSQ-VTIME: [dispatch] dsq: 10008, vtime: 36868751750957273, task: sptp-1688431-1688422
```

What we're seeing here is that enqueue starts, saving the `p2dq_vtime` of `19578630807392904` in `taskc->p2dq_vtime`. Then, we see the vtimethat is placed the delay queue (wall time) of `83581052434161`. The random delay is over, and we reset the vtime to `19578630807392904` (the value stored in `taskc->p2dq_vtime`).  Now, when we call `async_p2dq_enqueue`, the dsq should be moved. However, we see that pre-dsq and post-dsq are the same, with a very large vtime that was updated via `update_vtime` in p2dq. Since it is on the delay queue still, we see that we get a log from chaos_dispatch, but this time with an extremely high vtime that the delay queue will never kick out (because the delay queue uses wall time).

I'm not sure why `scx_bpf_dsq_insert` isn't properly putting the task into the new dispatch queue. Any ideas? 

</details>


yaakov-stein:
> Nice, thanks for root causing this! Looking at the code, I wonder whether we could deal with it slightly differently though?

Yes, agreed that the current approach is awkward and would lead to unexpected results (ie adding a random delay can theoretically speed up a task by moving it local).

> I'm thinking if `complete_` fails (check the return val is correct), we enqueue locally and increment a stat. IMO this failing is a bug and it would be good to keep track of how many tasks get delayed enqueued "properly" and how many get direct dispatched.

The problem with this is that `complete_` will always succeed. The failing method of `scx_bpf_dsq_insert` is inside p2dq and I don't believe it even registers as a failure, as it returns void per the [docs](https://docs.ebpf.io/linux/kfuncs/scx_bpf_dsq_insert/). I'll add a new commit that checks if the current dispatch queue is still the delay queue, and only then uses `scx_bpf_dsq_move_to_local`. I'm interested to hear what you think about that approach.

> The reason for the delayed enqueued is to attempt to delay things by a reasonably distributed amount. That is, if we delay for 100ms then direct dispatch, the task is delayed for 100ms enqueue->running (practically). If we delay for 100ms then enqueue the task is delayed 100ms from enqueue->enqueue, which _should_ provide a cleaner distribution.

This makes total sense and 100% agree.

scheds/rust/scx_chaos/src/bpf/intf.h
```diff
@@ -50,6 +51,11 @@ enum chaos_stat_idx {
 	CHAOS_STAT_CHAOS_EXCLUDED,
 	CHAOS_STAT_CHAOS_SKIPPED,
 	CHAOS_STAT_TIMER_KICKS,
+	CHAOS_STAT_DISPATCH_MOVE_SUCCESS,
+	CHAOS_STAT_DISPATCH_MOVE_FAILURE,
+	CHAOS_STAT_DISPATCH_COMPLETE,
+	CHAOS_STAT_DISPATCH_FIFO,
+	CHAOS_STAT_DISPATCH_VTIME,
```
yaakov-stein, line 58: These are mostly for testing and to show the strange behavior, I'd only keep the `CHAOS_STAT_DISPATCH_MOVE_FAILURE`.

scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -454,7 +466,18 @@ void BPF_STRUCT_OPS(chaos_dispatch, s32 cpu, struct task_struct *prev)
...
+
+		// In case p2dq_enqueue didn't properly move the task off the delay
+		// queue, we'll do that manually now.
+		if (p->scx.dsq->id == cpu_delay_queue) {
+			chaos_stat_inc(CHAOS_STAT_DISPATCH_MOVE_FAILURE);
+			scx_bpf_dsq_move_to_local(cpu_delay_queue);
+		} else {
+			chaos_stat_inc(CHAOS_STAT_DISPATCH_MOVE_SUCCESS);
+		}
+
```
yaakov-stein, line 478: Interestingly, you can see that every `P2DQ_ENQUEUE_PROMISE_COMPLETE` we get fails this check here (failures # always lines up with enqueue complete #):
Screenshot 2025-07-06 at 5 01 48 PM

yaakov-stein: I looked a bit more at the kernel code and noticed that `scx_bpf_dsq_insert` works pretty differently than `scx_bpf_dsq_move_to_local` and goes into a dispatch buffer, which is later flushed. Need to dig into that a bit and see how that is causing this bug. Wondering if there's some race condition while it's waiting in the buffer, will dig into it more tomorrow.

yaakov-stein: Something along the lines of this fix probably works for a temporary fix though.

JakeHillion:
Awesome debugging! Thank you for adding all the stats and clearly explaining the problem.

In short: this makes total sense and you've exposed the bug. The `P2DQ_ENQUEUE_PROMISE_COMPLETE` state makes no sense for the `complete_p2dq_enqueue_move` case. `_COMPLETE` contains the cases where p2dq couldn't ship a promise that made sense, normally because it's already directly dispatched the task. But we can't directly dispatch the task when it's already in a DSQ, hence the failure.

There are a few solutions here. One is to do the direct dispatch but in `complete_p2dq_enqueue_move`. I think this could lead to stalls in the future though, as we'd have to be careful to check the cpumask etc. This all seems a bit noisy.

The better solution is to properly encapsulate the work p2dq is doing in the promise. This will require adding a new state for direct dispatch & adding a destructor (`destroy_p2dq_enqueue_promise` or similar) to clean up the idle mask if we don't end up running the dispatch. I think this is the better way to go, but slightly more work.

I'm a little worried about merging this as is because p2dq might have broken its idle state tracking in the _COMPLETE state. Let's chat about this offline and figure out the best way forward.

yaakov-stein:
A temporary fix is being shipped in #2343. @JakeHillion and I discussed offline and to ensure chaos works to some extent for the upcoming release, this fix will suffice for now, even though it somewhat distorts the random delays. 

For a more permanent fix, there are two options that we discussed (@JakeHillion will be working on the permanent fix):
1) Find a way to flush the dispatch buffer without needing to call `scx_bpf_dsq_move_to_local`. This would be a simple solution but doesn't seem to be supported at the moment.
2) Re-work how `p2dq_enqueue` and `chaos_dispatch` work together, updating the way promises currently work to be more compatible with chaos.

PR #2334, opened 2025-07-05
scx_lavd: Correctly measure wake frequency. 

multics69:
When SCX_OPS_ALLOW_QUEUED_WAKEUP is set, wake-up operations are batch-processed in the kernel. Hence, @current is no longer a waker task, so we cannot correctly track the task's wake frequency. So we disable SCX_OPS_ALLOW_QUEUED_WAKEUP for now.

Also, we limit the wake frequency tracking within userspace tasks under the same parent. The kernel tasks are too broadly related.


hodgesds (approved):
This makes sense, thanks for the good description.

PR #2335, opened 2025-07-05
lib/minheap: Guard zero-capacity and trivial heap cases

EricccTaiwan:
Add early returns for:
- `scx_minheap_alloc_internal()` when capacity is 0.
- `scx_minheap_balance_{top,bottom}_down()` when heap size < 2.

These avoid unnecessary work in edge cases.


etsal:
I agree with @JakeHillion on this, this seems to have no measurable benefit. The NULL allocation case is taken care of in the existing code because the internal allocation will return NULL after requesting 0 bytes. The balance case will also avoids  a couple instructions at most because the loop never runs.

Issue #2310, opened 2025-07-02
scxtop: add "missing scheduler" message


yaakov-stein:
If the scheduler is unloaded and one selects the scheduler view, it would be nice to get a message that tells you why the screen is empty instead of a bunch of empty panes. See #2309 for a bit of context.

yaakov-stein:
As mentioned in #2379, it would also be nice to have a banner or pane message appear if one of the events fails to attach.

PR #2336, opened 2025-07-05
scxtop: Add "missing scheduler" message

EricccTaiwan:
https://github.com/user-attachments/assets/fe6a50ad-03e7-48cb-9385-c6fc907f775d

Close: #2310

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -47,12 +47,12 @@ use num_format::{SystemLocale, ToFormattedString};
 use ratatui::prelude::Constraint;
 use ratatui::{
     layout::{Alignment, Direction, Layout, Rect},
-    style::{Modifier, Style, Stylize},
+    style::{Color, Modifier, Style, Stylize},
     symbols::bar::{NINE_LEVELS, THREE_LEVELS},
     text::{Line, Span},
     widgets::{
-        Bar, BarChart, BarGroup, Block, BorderType, Borders, Gauge, Paragraph, RenderDirection,
-        Scrollbar, ScrollbarOrientation, ScrollbarState, Sparkline,
+        Bar, BarChart, BarGroup, Block, BorderType, Borders, Clear, Gauge, Paragraph,
+        RenderDirection, Scrollbar, ScrollbarOrientation, ScrollbarState, Sparkline,
     },
     Frame,
 };
@@ -1449,6 +1449,33 @@ impl<'a> App<'a> {
         Ok(())
     }
 
+    /// Draw the "Missing Scheduler" message.
+    fn missing_scheduler_msg(&self, frame: &mut Frame, area: Rect) {
+        frame.render_widget(Clear, area);
+
+        let top_pad = area.height.saturating_sub(1) / 2;
+
+        let mut lines: Vec<Line> = Vec::with_capacity(top_pad as usize + 1);
+        for _ in 0..top_pad {
+            lines.push(Line::raw(""));
+        }
+        lines.push(Line::from(Span::styled(
+            "Missing Scheduler",
+            Style::default().fg(Color::Red).add_modifier(Modifier::BOLD),
+        )));
+
+        let block = Block::default()
+            .borders(Borders::ALL)
+            .border_type(BorderType::Rounded)
+            .style(Style::default().fg(Color::Red));
+
+        let para = Paragraph::new(lines)
+            .alignment(Alignment::Center)
+            .block(block);
+
+        frame.render_widget(para, area);
+    }
+
     /// Renders the scheduler application state.
     fn render_scheduler(
         &mut self,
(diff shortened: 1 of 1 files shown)


tools/scxtop/src/app.rs
```diff
@@ -1322,6 +1322,37 @@ impl<'a> App<'a> {
...
 
+    /// Draw a top banner saying "missing scheduler" when no scx scheduler
+    /// is loaded.
+    fn render_missing_scheduler_banner(&self, frame: &mut Frame) {
+        use ratatui::{
+            layout::{Constraint, Layout},
+            style::Color,
+            text::Span,
+            widgets::{Clear, Paragraph},
+        };
```
yaakov-stein, line 1333: These should already be imported at the top of the file. I don't think we need them here

yaakov-stein: Most of these should already be imported at the top of the file. If there are additional imports that we need, probably best to add them at the top of the file.

tools/scxtop/src/app.rs
```diff
@@ -1488,11 +1519,18 @@ impl<'a> App<'a> {
...
+            )?,
             ViewState::BarChart => {
-                self.render_scheduler_barchart(event, frame, area, render_sample_rate)
+                self.render_scheduler_barchart(event, frame, area, render_sample_rate)?
             }
         }
+
+        // Overlay the banner if no scheduler loaded.
+        if self.scheduler.is_empty() {
+            self.render_missing_scheduler_banner(frame);
```
yaakov-stein: While this method probably needs a rename (`render_scheduler` is a bit misleading as it render's a "scheduler" event), at the moment, this code is re-rendering the banner multiple times for each `render_scheduler` call. See [here](https://github.com/sched-ext/scx/blob/main/tools/scxtop/src/app.rs#L2230) for an example of where we call `render_scheduler` multiple times for a single frame.

tools/scxtop/src/app.rs
```diff
@@ -1322,6 +1322,37 @@ impl<'a> App<'a> {
...
+            layout::{Constraint, Layout},
+            style::Color,
+            text::Span,
+            widgets::{Clear, Paragraph},
+        };
+
+        let [banner, _] =
+            Layout::vertical([Constraint::Length(3), Constraint::Min(0)]).areas(frame.area());
+
+        frame.render_widget(Clear, banner);
```
yaakov-stein: This overlays whatever data is at the top of the screen, which we don't want in case someone would still like to see the event data or any other non-sched_ext data.

yaakov-stein (commented):
Thanks - I think overall this idea makes sense. I'd imagine doing this at the `AppState` level in `render` would be the cleanest approach here (ie checking if we're in `AppState::Scheduler` and splitting the frame to add an area for a banner while leaving the rest of the panes the same). It probably makes the most sense to start with `AppState::Scheduler` because that data is irrelevant when there isn't a sched_ext scheduler running.

EricccTaiwan:
> I'd imagine doing this at the `AppState` level in `render` would be the cleanest approach here (ie checking if we're in `AppState::Scheduler` and splitting the frame to add an area for a banner while leaving the rest of the panes the same). It probably makes the most sense to start with `AppState::Scheduler` because that data is irrelevant when there isn't a sched_ext scheduler running.

Agreed. I’ll be away on a trip and will return to fix this by the weekend. Thanks for the review! :+1:

tools/scxtop/src/app.rs
hodgesds: Yeah, keep all the imports at the top of the file please.

tools/scxtop/src/app.rs
hodgesds: Or have a bool option passed into to optionally clear and then depending on the render context it could be set when needed (ex: scheduler view).

hodgesds:
I think this is close to good, maybe in the default pane put the warning/error message in the center of the panels that render scheduler specific stats if that makes sense? Then the layout doesn't need to change at all, just the content in the scheduler specific panes.

tools/scxtop/src/app.rs
EricccTaiwan: Would you prefer I rename `render_scheduler` to something like `render_scheduler_event` (since it’s really drawing one scheduler metric), or did you have a different name in mind?
https://github.com/sched-ext/scx/blob/505a0b6267811523d4e57bb28a1041bb6645de89/tools/scxtop/src/app.rs#L1453

EricccTaiwan:
I previously used a top‑of‑screen banner, but that ended up covering part of scxtop’s data.
With the latest commit, I now render “missing scheduler” centered inside each scheduler‑specific pane instead. The layout size stays the same, and no other content is overlapped. 

[Screencast from 2025-07-20 15-42-28.webm](https://github.com/user-attachments/assets/fc6400fc-3809-4fa5-bd4f-60e60632a05e)

>I think this is close to good, maybe in the default pane put the warning/error message in the center of the panels that render scheduler specific stats if that makes sense? Then the layout doesn't need to change at all, just the content in the scheduler specific panes.

I am concerned that adding a separate frame for the banner could overlap the status panes on the right side.
Right now the code draws the centered "Missing Scheduler" message inside each scheduler‑specific pane so the overall layout and borders stay the same.
Is this approach acceptable, or would you prefer another way to display the warning?

hodgesds:
LGTM, please publish first :+1:

yaakov-stein:
Thanks for adding this!

PR #2338, opened 2025-07-06
scx_flash: Improve CPU locality

arighi:
A set of changes aimed to improve CPU locality without regressing work conservation and system responsiveness when the system becomes saturated or overcommitted.

These changes are still following the idea of keeping tasks running on the same CPU (as long as the CPU is sufficiently idle), using per-CPU DSQs, and move tasks to the per-node DSQs when their CPU becomes busy.

A set of new options have been added to better adapt the scheduler to different workloads:
- `--slice-lag-scaling`: Dynamically adjusts the maximum sleep-time budget based on CPU load. When CPUs are mostly idle, the budget shrinks, making the scheduler behave more like FIFO. When the system is busy, the budget increases to better differentiate between interactive and CPU-intensive tasks.
- `--direct-dispatch`: Enables immediate dispatching to idle CPUs. By default, direct dispatch only occurs after draining both the per-CPU and per-node DSQs to avoid starvation. Enabling this option allows to always dispatch directly to idle CPUs.
- `--no-builtin-idle`: Now, by default, the scheduler uses the in-kernel built-in idle CPU selection policy (if `scx_bpf_select_cpu_and()` is available). This option disables that behavior and falls back to the custom selection policy.

Round-robin mode has also been refined, yielding interesting performance results on interactive workloads, as long as the system is not overcommitted.

Finally, the `scx_loader` performance profiles have been updated to reflect the new options and behaviors.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -124,6 +124,11 @@ const volatile bool local_pcpu;
 const volatile bool direct_dispatch;
 
 /*
+ * Enable built-in idle CPU selection policy.
+ */
+const volatile bool builtin_idle;
+
+/*
  * Native tasks priorities.
  *
  * By default, the scheduler normalizes task priorities to avoid large gaps
@@ -764,6 +769,38 @@ static bool cpus_share_llc(s32 this_cpu, s32 that_cpu)
 }
 
 /*
+ * Compatibility helper to transparently use the built-in idle CPU
+ * selection policy (if scx_bpf_select_cpu_and() is available) or fallback
+ * to the custom idle selection policy.
+ */
+static s32 pick_idle_cpu_builtin(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bool *is_idle)
+{
+	const struct cpumask *primary;
+	s32 cpu;
+
+	if (!builtin_idle || !bpf_ksym_exists(scx_bpf_select_cpu_and))
+		return -ENOENT;
+
+	primary = cast_mask(primary_cpumask);
+	if (!primary)
+		return -EINVAL;
+
+	if (no_wake_sync)
+		wake_flags &= ~SCX_WAKE_SYNC;
+
+	cpu = primary_all ? -ENOENT :
+			scx_bpf_select_cpu_and(p, prev_cpu, wake_flags, primary, 0);
+	if (cpu < 0) {
+		cpu = scx_bpf_select_cpu_and(p, prev_cpu, wake_flags, p->cpus_ptr, 0);
+		if (cpu < 0)
+			return prev_cpu;
+	}
+	*is_idle = true;
+
+	return cpu;
+}
+
+/*
  * Find an idle CPU in the system.
  *
  * NOTE: the idle CPU selection doesn't need to be formally perfect, it is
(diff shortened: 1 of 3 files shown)


sirlucjan (approved):
I've been using it for a few days. flash -D -L in particular achieved very good results.

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -553,6 +559,14 @@ static void update_task_deadline(struct task_struct *p, struct task_ctx *tctx)
...
 	lag_scale = max_avg_nvcsw ? log2_u64(MAX(tctx->avg_nvcsw, 2)) : 1;
 
+	/*
+	 * Adjust the budget in function of the average CPU utilization:
+	 * increase the allowed spread when CPUs are more utilized and
+	 * reduce it when they are more idle.
+	 */
+	if (slice_lag_scaling)
+		lag_scale = lag_scale * cpu_util / SCX_CPUPERF_ONE;
+
```
multics69: It would be useful to add why dynamic lag scaling can increase the throughput of message passing workloads and reduce the system responsiveness.

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -1228,6 +1329,40 @@ static void preempt_curr(s32 cpu)
...
+ */
+static void rr_enqueue(struct task_struct *p, struct task_ctx *tctx,
+		       s32 prev_cpu, u64 enq_flags)
+{
+	/*
+	 * Try to bounce the task on another CPU if it has been re-enqueued
+	 * due to a higher scheduling class stealing the CPU.
+	 */
+	if (enq_flags & SCX_ENQ_REENQ) {
+		scx_bpf_dsq_insert(p, SCX_DSQ_GLOBAL, task_slice(prev_cpu), enq_flags);
```
hodgesds, line 1355: It might be nice to try to keep it sticky to the LLC if possible

hodgesds (approved):

scheds/rust/scx_flash/src/bpf/main.bpf.c
```diff
@@ -495,16 +495,12 @@ static inline u64 nr_tasks_waiting(s32 cpu)
...
- * Return the maximum time slice that can be assigned to a task, or an
- * infinite time slice if tickless mode is enabled.
- */
-static inline u64 task_slice_max(void)
-{
-	return tickless_sched ? SCX_SLICE_INF : slice_max;
+	if (!nr_wait)
+		return tickless_sched ? SCX_SLICE_INF : slice_max;
+
+	return MAX(slice_max / nr_wait, slice_min);
```
multics69, line 503: Would this make the time slice too short under pressure since slice_max is 4096 usec?

multics69 (approved):
Left two questions. Overall looks good to me.

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: Good point, thanks @multics69 . I've added a comment to better explain the idea behind the `slice_.lag_scaling` and I've also updated the help to document this concept.

All the details are in commit https://github.com/sched-ext/scx/pull/2338/commits/19b97c28b6de43d7f7f1697877ff4e92c245f0c8.

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: Absolutely, I've used  `SCX_DSQ_GLOBAL` for simplicity, but this part of the round-robin mode can be definitely improved. I'll  add this "locality" improvement in another PR. Thanks @hodgesds!

scheds/rust/scx_flash/src/bpf/main.bpf.c
arighi: The min time slice is still capped by `slice_min`, so it won't go below that limit. The slice_max at 4096 us seems to work pretty well for interactivity, maybe we should bump up the default slice_min a bit (now it's 128  us, which is probably a bit too small). Will do some tests trying to increase the lower limit.

PR #2341, opened 2025-07-07
scx_p2dq: Code clean and add __pick_idle_cpu helper

hodgesds:
Add a new helper __pick_idle_cpu to remove a large number of duplicate cpumask casts and refactor code to be more readable. No functional changes, only prep work for arena implementations.


arighi (approved):
LGTM

PR #2343, opened 2025-07-07
scx_chaos: temp fix for random delay stalls

yaakov-stein:
This is a temporary fix for the random delay stalls that were occuring. See #2326 for details. This temporary solution somewhat distorts the idea of random delays due to the fact that when a random delay is up, the task is immediately placed into a local dispatch queue.

Testing:
- It runs without stalling, even with an aggresive random delay frequency

Notes:
I'm leaving `complete_p2dq_enqueue_move` and `async_p2dq_enqueue_weak` in the code because they will be used for the permanent fix. As of now though, they'll cause a warning for unused code.


JakeHillion (approved):
Can you drop the now unused `promise` variable and add `__maybe_unused` to the two unused functions? It's a bit of formatting churn but will stop us annoying everyone else with warnings.

PR #2347, opened 2025-07-07
scxtop: fix config save

yaakov-stein:
Previously, we suppressed any errors that `Config::load()` caused. This led to the introduction of bugs in the development of `MangoApp` + `KprobeEvent`'s. So, whenever the `Config::load().unwrap_or(Config::default())` ran, we always used the default config.

This takes a few steps to ensure that doesn't occur again:
1) We panic with a helpful message on any errors.
2) Comprehensive tests (#2350) will be introduced for the Config struct (credit to @hodgesds for a good chunk of these test cases).

Testing:
* Ran without a scxtop.toml file and saved, ensured the file persisted and was used in next run.
* Updated a scxtop.toml file and ensured changes persisted.

### DIFF
--- a/tools/scxtop/src/cli.rs
+++ b/tools/scxtop/src/cli.rs
@@ -62,7 +62,7 @@ pub struct TuiArgs {
     #[arg(long, default_value_t = -1)]
     pub process_id: i32,
     /// Custom perf events colon delimited (ex: "<event_name>:<event and umask ex: 0x023>:<event_type ex: 4>")
-    #[arg(long, num_args = 1.., value_parser)]
+    #[arg(long)]
     pub perf_events: Vec<String>,
     /// Default perf event colon delimited (ex: "<event_name>:<event and umask ex: 0x023>:<event_type ex: 4>")
     #[arg(long, default_value = "hw:cycles")]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3503,9 +3503,9 @@ dependencies = [
 
 [[package]]
 name = "tempfile"
-version = "3.20.0"
+version = "3.19.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e8a64e3985349f2441a1a9ef0b853f869006c3855f2cda6862a94d26ebb9d6a1"
+checksum = "7437ac7763b9b123ccf33c338a5cc1bac6f69b45a136c19bdd8a65e3916435bf"
 dependencies = [
  "fastrand",
  "getrandom 0.3.2",


tools/scxtop/src/cli.rs
```diff
@@ -62,7 +62,7 @@ pub struct TuiArgs {
     #[arg(long, default_value_t = -1)]
     pub process_id: i32,
     /// Custom perf events colon delimited (ex: "<event_name>:<event and umask ex: 0x023>:<event_type ex: 4>")
-    #[arg(long)]
+    #[arg(long, num_args = 1.., value_parser)]
```
yaakov-stein: Side fix: this makes testing a bit easier and is easier for the user as well

tools/scxtop/src/config.rs
```diff
@@ -210,10 +210,7 @@ impl Config {
 
     /// Duration of trace in ns.
     pub fn trace_duration_ns(&self) -> u64 {
-        self.trace_duration_ms
-            .or(self.trace_ticks.map(|t| (t * self.tick_rate_ms()) as u64))
-            .unwrap_or(1_250)
-            * 1_000_000
```
yaakov-stein: Side fix: trace_ticks and trace_tick_warmup are deprecated, we shouldn't be using them.

tools/scxtop/src/config.rs
```diff
@@ -286,24 +278,34 @@ impl Config {
...
         let contents = fs::read_to_string(config_path)?;
         let mut config: Config = toml::from_str(&contents)?;
 
-        if let Some(keymap_config) = &config.keymap {
+        config.resolve_keymap()?;
+
+        Ok(config)
+    }
+
+    fn resolve_keymap(&mut self) -> Result<()> {
```
yaakov-stein, line 304: Separating these makes things cleaner and also makes testing a bit easier since we can't use `get_config_path`.

tools/scxtop/src/cli.rs
yaakov-stein: moved to #2348

tools/scxtop/src/config.rs
yaakov-stein: moved to #2348

tools/scxtop/src/cli.rs
yaakov-stein: moved to #2349

tools/scxtop/src/config.rs
```diff
@@ -286,24 +286,34 @@ impl Config {
     }
 
     /// Loads the config from XDG configuration.
-    pub fn load() -> Result<Config> {
+    pub fn load_or_default() -> Result<Config> {
```
hodgesds, line 289: Yeah, this seems like a slightly better solution.

PR #2349, opened 2025-07-07
scxtop: make entering perf events easier

yaakov-stein:
This makes it so you can enter perf events without comma-delimiting and aligns the way we accept these args with krpobe args.


hodgesds (approved):
LGTM

PR #2353, opened 2025-07-08
scx_p2dq: Fix load balancing

hodgesds:
Fix multiple issues with the current load balancing implementation. The slack factor was refactored and no longer being applied as a factor which meant any non zero value would cause load balancing not to work properly. When determining if a task is eligible to be placed on the migration DSQ check the relative load for the DSQ index rather than relying on heuristics. Using DSQ index relative load does a better job of spreading work evenly across the machine.


hodgesds:
Stats show load balancing is working properly under saturation as previously no `pick2 dispatch` events would ever occur.
```
direct/idle/keep 385/32283/0
        dsq same/migrate 20647/28081
        enq llc/intr/mig 8171/12378/330
        wake prev/llc/mig 42/11/0
        pick2 select/dispatch 0/16
        migrations llc/node: 301/0
 ```

File: scheds/rust/scx_p2dq/src/bpf/main.bpf.c [hunk 1]
```diff
@@ -302,37 +302,53 @@ static bool is_interactive(task_ctx *taskc)
...
+		if (mig_nr_queued >= intr_nr_queued)
 			return false;
 	}
 
-	return (nr_llcs > 1 &&
-		taskc->all_cpus &&
-		taskc->llc_runs > min_llc_runs_pick2);
+	// The least interactive can always migrate
+	if (taskc->dsq_index == MAX_DSQS_PER_LLC - 1)
+		return true;
```
JakeHillion, line 331: Does this mean "is it the migration DSQ?" now?

JakeHillion (approved):
I'm not totally on top of this logic, particularly the bits where the migration eligible tasks can't exceed the non-eligible, but the changes make sense.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: I guess so, there's still the check that the task has run sufficiently in the LLC prior to this to prevent the least interactive tasks from ping-ponging.

hodgesds:
> I'm not totally on top of this logic, particularly the bits where the migration eligible tasks can't exceed the non-eligible, but the changes make sense.

Yeah, it still needs some work but this should be good to go for the release.

PR #2354, opened 2025-07-08
p2dq/chaos: switch more enqueue cases to be async

JakeHillion:
p2dq's enqueue has diverged from the implementation when chaos was initially implemented. Then, we had P2DQ_ENQUEUE_COMPLETE but only based on a fixed task condition. Now, _COMPLETE gets hit based on a variety of conditions, many of which include the idle state of the machine. As this state can change, we started hitting complete in dispatch (delayed enqueue) even though we didn't hit it in enqueue the first time around. This meant that tasks weren't being enqueued properly the second time, and tasks were stalling.

Implement `kick_if_idle` for the enqueue promise and cover many of the remaining p2dq cases. I left the per-CPU kthread for now in the hope it includes BPF timers, as delaying BPF timers would prevent forward progress in chaos and lead to stalls. This also required the destruction of a promise if it's not executed, as p2dq might have already set the CPU to not idle. After discussion with @arighi, this is undone by kicking the CPU, which resets the idle state.

Test plan:
- CI
- `sudo target/release/scx_chaos --random-delay-frequency 0.5 --random-delay-min-us 500000 --random-delay-max-us 750000` - ran for about half an hour and all was fine on my machine.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -723,10 +723,13 @@ static __always_inline s32 p2dq_select_cpu_impl(struct task_struct *p,
  * is primarily used so that scx_chaos can decide to enqueue a task either
  * immediately in `enqueue` or later in `dispatch`. This returns a tagged union
  * with three states:
- * - P2DQ_ENQUEUE_PROMISE_COMPLETE: Either the task has been enqueued, or there
- *     is nothing to do (enqueue failed).
+ * - P2DQ_ENQUEUE_PROMISE_COMPLETE: The enqueue has been completed. Note that
+ *     this case _must_ be determinstic, or else scx_chaos will stall. That is,
+ *     if the same task and enq_flags arrive twice, it must have returned
+ *     _COMPLETE the first time to return it again.
  * - P2DQ_ENQUEUE_PROMISE_FIFO: The completer should enqueue this task on a fifo dsq.
  * - P2DQ_ENQUEUE_PROMISE_VTIME: The completer should enqueue this task on a vtime dsq.
+ * - P2DQ_ENQUEUE_PROMISE_FAILED: The enqueue failed.
  */
 static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
 					       struct task_struct *p,
@@ -736,6 +739,10 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
 	struct llc_ctx *llcx;
 	task_ctx *taskc;
 	s32 cpu = scx_bpf_task_cpu(p);
+
+	// Default to 0 and set to failed.
+	__builtin_memset(ret, 0, sizeof(*ret));
+	ret->kind = P2DQ_ENQUEUE_PROMISE_FAILED;
 
 	/*
 	 * Per-cpu kthreads are considered interactive and dispatched directly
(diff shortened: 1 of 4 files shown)


hodgesds (approved):
The code is much more easy to reason about now, thanks for cleaning enqueue up!

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -791,45 +795,44 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
...
-				scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
-			}
-			ret->kind = P2DQ_ENQUEUE_PROMISE_COMPLETE;
+		if (ret->has_cleared_idle || cpuc->nice_task) {
+			ret->kind = P2DQ_ENQUEUE_PROMISE_FIFO;
+			ret->fifo.dsq_id = SCX_DSQ_LOCAL_ON|cpu;
+			ret->fifo.slice_ns = taskc->slice_ns;
+			ret->fifo.enq_flags = enq_flags;
+
+			ret->kick_if_idle = true;
```
hodgesds: I think this might need to be the following
```suggestion
			ret->kick_if_idle = ret->has_cleared_idle;
```

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -840,15 +843,13 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
...
```
hodgesds: Same here
```suggestion
			ret->kick_if_idle = ret->has_cleared_idle;
```

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -896,15 +898,14 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
...
-		}
-		ret->kind = P2DQ_ENQUEUE_PROMISE_COMPLETE;
+	ret->has_cleared_idle = scx_bpf_test_and_clear_cpu_idle(cpu);
+	if (ret->has_cleared_idle || cpuc->nice_task) {
+		ret->kind = P2DQ_ENQUEUE_PROMISE_FIFO;
+		ret->fifo.dsq_id = SCX_DSQ_LOCAL_ON|cpu;
+		ret->fifo.slice_ns = taskc->slice_ns;
+		ret->fifo.enq_flags = enq_flags;
+
+		ret->kick_if_idle = true;
```
hodgesds: Same here
```suggestion
		ret->kick_if_idle = ret->has_cleared_idle;
```

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
JakeHillion: Yeah, this is a bit tricky. I did it like this because I was a bit worried about the race condition. The old flow was:

- Check if machine is idle.
- Enqueue task.
- Kick it.

The new flow is (as is):

- Check if machine is idle.
- Promise task, some delay while other code is executed.
- Enqueue task.
- Kick it if it was idle initially, else check if it's now idle and kick it.

I was worried that the slightly longer race if anybody decided to keep a promise around/do extra checks would end up placing a task onto a queue of a newly idle CPU and not kicking it. What do you reckon?

hodgesds: I think I got slightly better `schbench` results by using the `ret->has_cleared_idle` approach (in the fraction of a percent), but that might be hardware dependent.

PR #2357, opened 2025-07-08
versions: Bump versions for v1.0.14 release

JakeHillion:


ptr1337:
Was this missed to put it on the tag? It appears to be still qeueud, but tag is already available

JakeHillion:
> Was this missed to put it on the tag? It appears to be still qeueud, but tag is already available

Tag should be the commit that was merged from this PR, but missing the merge commit of it into main. It's because `cargo publish` can't be reliably tested, so I keep things on the branch until I can cut the tag. Also means the commit in the tag is exactly what I published from, even if the merge races.

JakeHillion:
@likewhatevs vmlinux_docify didn't publish, would you mind sorting this or sticking `publish = false` in the Cargo.toml to have it skipped?

PR #2361, opened 2025-07-08
scx_p2dq: Set CPU on promise during enqueue

hodgesds:
When running the promise for enqueue ensure that the CPU is properly set when handling cases where the CPU needs to be selected such as affinitization. This improves wakeup latency as CPU kicks are properly routed to the correct CPU.


JakeHillion (approved):
Good catch.

PR #2364, opened 2025-07-09
scxtop: switch cpu to procfs

yaakov-stein:
This uses the cleaner `fb_procfs` wrapper instead of manually parsing /proc/stats. Given that we are planning on adding more proc data, a "global" `proc_reader` was also added to the `App` struct.


hodgesds (approved):
Makes sense, nice to have less code to maintain.

PR #2365, opened 2025-07-09
scx_bpf_unittests: add crate to drive BPF unittests

JakeHillion:
Currently scxtest tests are driven by a main function handwritten in C and ran by `meson test`. This is fine, but let's change it for a few reasons:
- I want fewer dependencies on meson, not more.
- We already have a richer test runner, `cargo test`, that is used in the repo.
- There's no autodetection of tests available and every new test must be added to main.

This change builds the BPF unittests in a new crate, `scx_bpf_unittests`, that doesn't get published. It uses codegen based on this build and a new macro, `SCX_TEST`, to find the tests and automatically shim them into Rust. `cargo test` then runs this userspace code perfectly normally. As we already have `cargo test` in the CI, this runs the tests in the CI, with full multithreading.

Changes made to the existing test framework:
- Add `SCX_TEST` macro with a `.scxtest` section to automatically locate all the built tests.
- Use a longjmp for `scx_test_assert` to avoid `abort()` as the Rust test runner doesn't like this and we still want to run other tests after one fails.
- Switch every test global I could find to `__thread`. We may need to be smarter about this in the future when we test scheduler methods that touch scheduler owned globals, but it should be fine for now.
- Runs `init_p2dq_test` in every test instead of once in `main` to make sure every thread is initialised properly.
- Rip out the meson logic. We should only have one way of running these tests and it should be the version we run in the CI to avoid diverging/confusion.

In future we should make this test builder a library and expose it to the crates so these tests run as part of `scx_p2dq` instead of in a separate crate (keeping a separate crate for the cscheds). For now, let's run them in their own crate to let them bed in for a while.

Test plan:
- CI


JakeHillion:
Naturally broke Meson. Will iterate on the cleanup.

hodgesds (approved):
Good stuff

lib/scxtest/scx_test_map.c
```diff
@@ -26,14 +26,14 @@ struct scx_percpu_map_entry {
 	struct scx_percpu_test_map *map;
 };
 
-static struct scx_map_entry *scx_map_entries = NULL;
-static int scx_map_entries_count = 0;
+static __thread struct scx_map_entry *scx_map_entries = NULL;
```
hodgesds, line 29: Maybe a silly question but do all of these need to be thread local?

JakeHillion: I'm not super sure and haven't reviewed this code well. My theory is that it's better to play it safe and start with everything thread local then relax it later or for new additions, given none of this code was written with multithreading in mind.

PR #2367, opened 2025-07-09
scxtop: add cpu_frequency tracking

yaakov-stein:
This PR does a few things:
1) Adds cpu frequency data to the `CpuStatTracker`, giving an easy way for other structs to access this data.
2) Adds a `CpuUtilEvent` so users can see statistics on cpu frequency as well as frequency data at the LLC and NUMA level.
3) Adds testing for the new data and methods.

TODO: 
Now that we have this data in `CpuStatTracker` we can use this instead of calling `record_cpu_freq` and manually parsing the sys files.


tools/scxtop/src/app.rs
```diff
@@ -63,6 +63,7 @@ use scx_utils::misc::read_from_file;
 use scx_utils::scx_enums;
 use scx_utils::Topology;
 use serde_json::Value as JsonValue;
+use sysinfo::System;
```
hodgesds, line 66: Why a new crate? Doesn't `fb_procfs` also have a [`Cpu`](https://docs.rs/sysinfo/latest/sysinfo/struct.Cpu.html) struct? It should be available on the [`Stat`](https://docs.rs/fb_procfs/0.9.0/fb_procfs/struct.Stat.html#structfield.cpus_map).

JakeHillion (approved):
LGTM.

tools/scxtop/src/app.rs
yaakov-stein: I couldn't find any of that data in `fb_procfs`. `fb_procfs` seems to have alot of the data from the `proc` folder, but not from the `sys` folder. They're both pretty helpful crates that can be used for alot of the future process/thread and system info work (I'll default to `fb_procfs` as it doesn't maintain state and is a bit lighter, and move to `sysinfo` if it doesn't have the info).

yaakov-stein: The `Stat` struct in `fb_procfs` is just a parsed view of `/proc/stat`, but I think that's all it has.

hodgesds:
LGTM

PR #2368, opened 2025-07-10
scx_lavd: Collect more information for '--monitor-sched-samples'

multics69:
The following additional information is collected when monitoring scheduling samples:

- Previous CPU ID where the task last ran.
- Suggested CPU ID when the task is enqueued at ops.enqueue() or ops.select_cpu().
- The time interval since a task's last run.
- Waker's PID and comm.


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -457,9 +467,12 @@ s32 BPF_STRUCT_OPS(lavd_select_cpu, struct task_struct *p, s32 prev_cpu,
 
 	/*
 	 * Find an idle cpu and reserve it since the task @p will run
-	 * on the idle cpu.
+	 * on the idle cpu. Even if there is no idle cpu, still repect
```
hodgesds:
```suggestion
	 * on the idle cpu. Even if there is no idle cpu, still respect
```

PR #2372, opened 2025-07-11
scx_lavd: Add 'rerunnable_interval' for '--monitor-sched-samples'

multics69:
Let's collect the time duration from when a task becomes quiescent to when it becomes runnable again.


hodgesds (approved):
Makes sense

PR #2360, opened 2025-07-08
/etc/default/scx: set scx_flash as default scheduler

sirlucjan:
After the latest changes made before the release of version 1.0.14, scx_flash began to outperform scx_bpfland in terms of performance. The start of work on the new 1.0.15 release is a good time to try to make scx_flash the new suggested scheduler.

PR #2374, opened 2025-07-11
scx_loader: cleaning and adjusting settings

sirlucjan:
With PR #2360, we changed the default scheduler from bpfland to flash in the /etc/default/scx file. To be consistent between settings, let's apply similar suggested settings in scx_loader.conf.

### DIFF
--- a/rust/scx_loader/configuration.md
+++ b/rust/scx_loader/configuration.md
@@ -16,7 +16,7 @@ If no configuration file is found at any of these paths, `scx_loader` will use t
 The configuration file has the following structure:
 
 ```toml
-default_sched = "scx_bpfland"
+default_sched = "scx_flash"
 default_mode = "Auto"
 
 [scheds.scx_bpfland]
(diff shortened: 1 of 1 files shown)


sirlucjan:
I also noticed that the configuration.md file was skipped when committing https://github.com/sched-ext/scx/commit/5f50c6a22999cc9ca72bf3f0bffc3ccba85346e2 so let's unify the flags to avoid confusion.

rust/scx_loader/configuration.md
```diff
@@ -90,16 +97,25 @@ The example configuration above shows how to set custom flags for different sche
 
 * For `scx_bpfland`:
     * Gaming mode: `-m performance`
-    * Low Latency mode: `-s 5000 -S 500 -l 5000`
+    * Low Latency mode: `"-s 5000 -S 500 -l 5000 -m performance`
```
arighi: Stray double quote?

arighi (approved):
Left a comment about a stray double quote, but overall looks good to me. And thanks for the cleanups.

About replacing scx_bpfland with scx_flash as the default choice, I think it makes sense considering the recent feedback from the users and all our testing. So it can be a reasonable default scheduler, as it’s generic enough despite being tuned for latency.

rust/scx_loader/configuration.md
sirlucjan: I'll fix it soon.

PR #2375, opened 2025-07-11
scxtop: fix trace

yaakov-stein:
Previously, the ring buffer was not cleared before stopping the trace and the receiver was not given time to process all events, leading to event being missed. The warmup period was also unnecessary.

Now, we `consume` the ring buffer and wait for the receiver to return `None`, signaling all channels have been closed and no more messages are waiting to be received.


hodgesds (approved):
LGTM

PR #2376, opened 2025-07-11
scx_p2dq: Refactor configuration

hodgesds:
Add configuration structs in the bpf side to make managing different types of scheduler configs scalable.


yaakov-stein (approved):
LGTM, thanks for the refactor! We probably should figure out why the structs cause us to need `MaybeUninit::new(...)` on the bools (in case it causes some bug in the future), but should be good for now.

PR #2377, opened 2025-07-11
scx_p2dq: Fix initialization

hodgesds:
Fix missing initialization of `interactive_dsq` config.


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -950,6 +950,7 @@ static __always_inline void async_p2dq_enqueue(struct enqueue_promise *ret,
 		scx_bpf_error("invalid lookup");
 		return;
 	}
+	ret->cpu = cpuc->id;
```
hodgesds, line 953: drive by fix to make sure this is always initialized

yaakov-stein (approved):
👍

PR #2378, opened 2025-07-11
scxtop: add cpu stats to trace

yaakov-stein:
Now that scxtop keeps all cpu stats in `CpuStatTracker`, we can add that data to the trace. 

Cpu Frequency:
Screenshot 2025-07-11 at 3 56 48 PM

Cpu Utilization (example is idle ns):
Screenshot 2025-07-11 at 3 57 05 PM

### DIFF
--- a/tools/scxtop/src/cpu_stats.rs
+++ b/tools/scxtop/src/cpu_stats.rs
@@ -37,7 +37,7 @@ impl CpuUtilData {
 #[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct CpuStatSnapshot {
     pub cpu_util_data: CpuUtilData,
-    pub freq: u64, // in kHz
+    pub freq_khz: u64,
 }
 
 #[derive(Clone, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
@@ -62,7 +62,7 @@ impl CpuStatTracker {
                 );
                 let snapshot = CpuStatSnapshot {
                     cpu_util_data,
-                    freq: cpu.frequency(),
+                    freq_khz: cpu.frequency(),
                 };
                 self.current.insert(i, snapshot);
             }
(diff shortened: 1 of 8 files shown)


tools/scxtop/src/perfetto_trace.rs
```diff
@@ -741,6 +751,57 @@ impl PerfettoTraceManager {
...
+            }
+
+            let current_data = &data.cpu_util_data;
+            let prev_data = &cpu_data_prev
+                .get(cpu)
+                .expect("Should have cpu data")
+                .cpu_util_data;
+
+            let cpu_time = CpuTimes {
+                cpu_id: Some(*cpu as u32),
```
yaakov-stein, line 783: For some reason, the `CpuTimes` cpu_id doesn't show up in Perfetto. It seems like it may be a Perfetto bug (or intentional behavior). Going to check into it a bit more, but I think this should be fine for the moment.

hodgesds: It might need a sequence id is the only thing I can think of.

tools/scxtop/src/util.rs
```diff
@@ -25,6 +25,15 @@ pub fn format_hz(hz: u64) -> String {
     }
 }
 
+/// Returns the current clock_id time in nanoseconds.
+pub fn get_clock_value(clock_id: libc::c_int) -> u64 {
+    let mut ts: libc::timespec = unsafe { std::mem::zeroed() };
+    if unsafe { libc::clock_gettime(clock_id, &mut ts) } != 0 {
```
hodgesds: since the `nix` crate is already a dependency may be able to use the safe version of [`clock_gettime`](https://docs.rs/nix/latest/nix/time/fn.clock_gettime.html).

tools/scxtop/src/cpu_stats.rs
```diff
@@ -35,28 +34,27 @@ impl CpuUtilData {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct CpuStatSnapshot {
     pub cpu_util_data: CpuUtilData,
-    pub freq: u64,
+    pub freq: u64, // in kHz
```
hodgesds: nit:
```suggestion
    pub freq_khz: u64,
```

hodgesds (approved):
few nits, seems legit

PR #2379, opened 2025-07-11
scxtop: fix perf event attach

yaakov-stein:
Here, we remove the code that suppressed the error coming from attempting to attach `hw:stalled-cycles-backend` and `hw:stalled-cycles-frontend`. Once removing that, the error from attempting to attach these events crashes the TUI, so we'll remove those for now.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -2728,14 +2728,10 @@ impl<'a> App<'a> {
                 }
             }
             Action::NextEvent => {
-                if self.next_event().is_err() {
-                    // XXX handle error
-                }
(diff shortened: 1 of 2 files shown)


hodgesds:
This might be a bad experience on some architectures where those perf events aren't present. That's why those errors are discarded.

yaakov-stein:
> This might be a bad experience on some architectures where those perf events aren't present. That's why those errors are discarded.

Ah, got it, that makes sense. As of now, the screen just show zero for everything, which can be a bit confusing. I guess ideally it would display a message that says the perf event failed to attach. Let's add that to #2310 , I'll add a comment there.

PR #2382, opened 2025-07-13
scx_p2dq: Add load balance mode

hodgesds:
Add different modes for load balancing. When determining the load for pick2 load balancing either use the total load, which is the current defaul or the number of tasks queued. The different load calculations give tradeoffs between latency and throughput.

`--lb-mode nr-queued`:
```
Wakeup Latencies percentiles (usec) runtime 30 (s) (132726 total samples)
          50.0th: 11         (45465 samples)
          90.0th: 20         (40756 samples)
        * 99.0th: 541        (11015 samples)
          99.9th: 941        (1191 samples)
          min=1, max=1761
Request Latencies percentiles (usec) runtime 30 (s) (133041 total samples)
          50.0th: 7032       (41174 samples)
          90.0th: 7432       (49979 samples)
        * 99.0th: 11120      (11836 samples)
          99.9th: 18080      (1193 samples)
          min=3520, max=28600
          20.0th: 4440       (14 samples)
        * 50.0th: 4456       (8 samples)
          90.0th: 4488       (7 samples)
          min=3777, max=4503
average rps: 4434.70
```
` --lb-mode load`
```
Wakeup Latencies percentiles (usec) runtime 30 (s) (131612 total samples)
          50.0th: 10         (36082 samples)
          90.0th: 30         (52052 samples)
        * 99.0th: 255        (11068 samples)
          99.9th: 967        (1184 samples)
          min=1, max=2607
Request Latencies percentiles (usec) runtime 30 (s) (131930 total samples)
          50.0th: 7192       (43839 samples)
          90.0th: 7336       (47176 samples)
        * 99.0th: 9264       (11005 samples)
          99.9th: 16112      (1181 samples)
          min=3634, max=30372
          20.0th: 4408       (16 samples)
        * 50.0th: 4408       (0 samples)
          90.0th: 4440       (15 samples)
          min=3785, max=4443
average rps: 4397.67
```


arighi (approved):
An idea, maybe for another PR.

Have you tried to use a combination of load and nr_queued? Ideally what we would like to know is the estimated wait time. If there are many tasks queued, but they're all short-lived the estimated wait time is probably smaller than a short queue with long-lived tasks. Maybe the load can give an idea if the tasks waiting on that CPU are more short-lived or long-lived.

So, I think it'd be interesting to have these options for lb-mode:
 - load
 - nr_queued
 - load * nr_queued

hodgesds:
> Have you tried to use a combination of load and nr_queued? Ideally what we would like to know is the estimated wait time. If there are many tasks queued, but they're all short-lived the estimated wait time is probably smaller than a short queue with long-lived tasks. Maybe the load can give an idea if the tasks waiting on that CPU are more short-lived or long-lived.

The load and nr_queued are considered when tasks are enqueued into the migration DSQ ([code](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_p2dq/src/bpf/main.bpf.c#L385)). In `can_migrate` it also tries to keep the migration DSQ smaller than the LLC DSQs ([code](https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_p2dq/src/bpf/main.bpf.c#L372)). So I think the optimal solution requires considering both load and nr_queued on both the enqueue and dispatch sides. It's not quite there yet, but I think this gets it closer.

PR #2383, opened 2025-07-13
scxtop: update default event

yaakov-stein:
Majority of this is tests (for config and the event structs). The main functional change here is to generalize the default event so it can be any of the current event types (krpobe, perf, or cpu) or any other future event type. This also allows us to clean up the `app.rs` file a bit. To do this, the config will now require a <source> prefix and a `from_str` method was added to the `ProfilingEvent` enum. For example, this is what default events can now look like:
* `perf:hw:cycles`
* `cpu:cpu_total_util_percent`
* `kprobe:vfs_read`

Now that it is generalized, we will also switch to using `cpu:cpu_total_util_percent` to show cpu utilization, at least until this gets it's own dedicated pane.

### DIFF
--- a/tools/scxtop/src/profiling_events/cpu_util.rs
+++ b/tools/scxtop/src/profiling_events/cpu_util.rs
@@ -7,7 +7,7 @@ use crate::CpuStatTracker;
 use anyhow::{bail, Result};
 use std::sync::{Arc, RwLock};
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub enum CpuUtilMetric {
     Total,
     User,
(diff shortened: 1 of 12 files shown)


tools/scxtop/src/profiling_events/cpu_util.rs
```diff
@@ -58,6 +74,9 @@ impl CpuUtilEvent {
 
     pub fn value(&self) -> Result<u64> {
         let tracker = self.tracker.read().unwrap();
+        if tracker.prev.is_empty() || tracker.current.is_empty() {
+            return Ok(0);
```
hodgesds, line 78: I think this makes sense.

yaakov-stein: Now that the cpu event was the default, it was calling value in `render` before the tracker had been updated twice (from `on_tick`), and so `prev` and/or `current` were empty. So added this check as a fallback for that case or similar cases.

PR #2388, opened 2025-07-14
nix: fix veristat build

JakeHillion:
Veristat build broke with the last flake update because it's not compatible with the nixpkgs libbpf. Fix the CI to fail in this case (it should have before but was getting lost in a subshell), and temporarily revert veristat so it builds again.

Test plan:
- Made the CI change first, pushed it, and everything went red.
- CI with the fix.


hodgesds (approved):
LGTM

PR #2389, opened 2025-07-14
scheds: unify a bit more the rust schedulers at exit time.

devnexen:


scheds/rust/scx_flash/src/main.rs
```diff
@@ -907,6 +907,10 @@ impl Drop for Scheduler<'_> {
                 }
             }
         }
+
+        if let Some(struct_ops) = self.struct_ops.take() {
+            drop(struct_ops);
+        }
```
JakeHillion: I don't see any value in this code. Without good reason, let's not adopt it in more places.

Firstly, calling `drop` here is very strange. A more appropriate piece of Rust code would be:
```
self.struct_ops = None;
```

Which will implicitly call drop - you should very rarely need to manually call drop.

Secondly, this is the last line of `impl Drop for Scheduler<'_>`. That means the fields in Scheduler are going to be dropped immediately anyway. The only reason you would need to do this is if there's an ordering constraint on which order the fields are dropped, and I don't believe there is. Plus, there are better ways to deal with that.

PR #2390, opened 2025-07-14
scxtop: add mem stats

yaakov-stein:
This adds a `MemStatSnapshot` struct to collect memory data that can be used for both the TUI and the trace. For now, it will just be added into the trace, with `free` and`swap_free`, as well as with the ratios for those, `mem_ratio` and `swap_ratio`:

Screenshot 2025-07-14 at 5 26 40 PM

### DIFF
--- a/tools/scxtop/src/perfetto_trace.rs
+++ b/tools/scxtop/src/perfetto_trace.rs
@@ -16,9 +16,9 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use crate::edm::ActionHandler;
 use crate::get_clock_value;
 use crate::{
-    Action, CpuStatAction, CpuhpEnterAction, CpuhpExitAction, ExecAction, ExitAction, ForkAction,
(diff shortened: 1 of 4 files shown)


tools/scxtop/src/perfetto_trace.rs
```diff
@@ -793,13 +848,59 @@ impl PerfettoTraceManager {
...
         self.sys_stats.entry(*ts).or_default().push({
             SysStats {
                 cpu_stat,
                 cpufreq_khz,
+                meminfo: mem_data,
                 ..SysStats::default()
             }
         });
+
+        self.mem_events
```
yaakov-stein, line 861: If we decide to add more of these in the future, it probably would make sense to have an iterator and "skeleton" method (where we can just pass in name + value and get the track event) due to their similarity. But with just two memory stats, this is probably simpler and avoids overcomplicating it for now. Happy to switch it now though if anyone feels differently.

hodgesds (approved):
LGTM

PR #2392, opened 2025-07-15
scx_lavd: Refactoring struct task_ctx.

multics69:
`task_ctx` has bloated over time with many boolean flags and a computable field. Let's clean this up. Replace all the boolean flags with bit flags (`LAVD_FLAG_*`). Also, remove dsq_id and instead use prev_cpu_id to account for cross-domain migrations.

There are no functional changes.


scheds/rust/scx_lavd/src/bpf/intf.h
```diff
@@ -118,7 +118,7 @@ struct task_ctx {
 	u64	wait_freq;		/* waiting frequency in a second */
 	u64	wake_freq;		/* waking-up frequency in a second */
 	u64	svc_time;		/* total CPU time consumed for this task scaled by task's weight */
-	u64	dsq_id;			/* DSQ id where a task run for statistics */
+	u32	prev_cpu_id;		/* where a task ran last time */
```
hodgesds, line 121: it might make sense to document the difference between `prev_cpu_id` and `cpu_id`.

PR #2395, opened 2025-07-15
scxtop: add process cmdline

yaakov-stein:
This adds the process command line arguments to `scxtop trace` by adding the command line args to a `Process` that is inserted in a `ProcessTree` and having perfetto link those `ProcessTree`'s to the proper `ProcessDescriptor`. Examples:

Screenshot 2025-07-15 at 10 24 35 AM

Screenshot 2025-07-15 at 10 24 47 AM

### DIFF
--- a/tools/scxtop/src/perfetto_trace.rs
+++ b/tools/scxtop/src/perfetto_trace.rs
@@ -34,6 +34,7 @@ use perfetto_protos::{
     ipi::IpiRaiseFtraceEvent,
     irq::{SoftirqEntryFtraceEvent, SoftirqExitFtraceEvent},
     process_descriptor::ProcessDescriptor,
+    process_tree::{process_tree::Process, ProcessTree},
     sched::{
         SchedMigrateTaskFtraceEvent, SchedProcessExecFtraceEvent, SchedProcessExitFtraceEvent,
         SchedProcessForkFtraceEvent, SchedSwitchFtraceEvent, SchedWakeupFtraceEvent,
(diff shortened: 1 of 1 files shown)


tools/scxtop/src/perfetto_trace.rs
```diff
@@ -207,48 +213,68 @@ impl PerfettoTraceManager {
...
+                Some(comm)
+            } else {
+                self.get_comm(pid)
+            };
+            let cmdline = self.get_cmdline(pid);
+            self.process_descriptors.insert(
+                parent_key,
+                ProcessDescriptor {
                     pid: Some(pid as i32),
+                    cmdline: cmdline.clone(),
```
yaakov-stein, line 278: For whatever reason, the Perfetto UI doesn't display this cmdline and only displays the `Process` cmdline (that gets put in `ProcessTree`) for some reason. But figured I'd stick it in once we have it here.

yaakov-stein:
> Nice. It makes sense to me to fill in the comm/cmdline immediately. I think this was previously done at the end to ensure the inner loop was fast, but we don't seem to have the big issue with dropping events anymore. Maybe an improvement for the future is to figure out how to delay these IO heavy tasks until later.

Ya, the main reason I changed it is because it can be called at most [# total pid's] times, so it shouldn't really have any performance hit. On the other hand, if a process is killed before the trace ends and we're calling `get_comm` only at the end, IIUC, the `get_comm` call would fail.

PR #2396, opened 2025-07-15
ci: reduce number of runs

JakeHillion:
Currently a commit gets 3/4 CI runs before being merged (assuming no amends of course):
- 1 on the original push, if you push to the main repo.
- 1 on the PR update.
- 1 when the merge queue bot creates a temporary branch for the merge queue state.
- 1 when merge queue checks occur.

The 2 merge queue related events happen very close together which means there's often a little queueing. This isn't painful, just inefficient.

Add the GitHub managed merge queue branches to `branches-ignore` of push and rely on the `merge_group` event instead.

Test plan:
- CI. If this doesn't merge then it doesn't work.

### DIFF
--- a/.github/workflows/caching-build.yml
+++ b/.github/workflows/caching-build.yml
@@ -1,9 +1,11 @@
 name: build-and-test
 
 on:
-  push:
   pull_request:
   merge_group:
+  push:
+    branches-ignore:
+      - gh-readonly-queue/main/**
 
 jobs:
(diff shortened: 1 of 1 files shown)


JakeHillion:
I miscounted in the commit message, it's actually +1 from that as it does an extra CI run when it pushes to main too. Honestly, that one is redundant too with the merge queue, but let's keep it for now. This change worked.

PR #2397, opened 2025-07-15
deps: update libbpf-rs/libbpf-cargo to 0.25.0

JakeHillion:
This update brings with it some API pain. Major points:
- All skel fields became optional[1]. I believe this only affects kernels way before sched_ext, so we're using .as_ref().unwrap() everywhere. Wherever the diff made it obvious that there were many in one scope I added a variable, but we'll likely end up refactoring this more in the future.
- New tracing API for clang warnings. Implemented as suggested this doesn't work, so I wrapped it in another layer to make sure we only get the clang stdout/stderr.

Test plan:
- CI

[1] https://github.com/libbpf/libbpf-rs/issues/1151


arighi (approved):
It's probably going to be a bit painful, but let's do this!

Also, once this is merged I can finally remove an ugly hack in rustland-core (something that I started to do in https://github.com/sched-ext/scx/commit/62a48c3ca90e8762b364ea0cc0f508bf59476756 but never landed, because I was waiting for a more 'official" libbpf-rs update).

JakeHillion:
> It's probably going to be a bit painful, but let's do this!
> 
> Also, once this is merged I can finally remove an ugly hack in rustland-core (something that I started to do in [62a48c3](https://github.com/sched-ext/scx/commit/62a48c3ca90e8762b364ea0cc0f508bf59476756) but never landed, because I was waiting for a more 'official" libbpf-rs update).

Nice! Are you planning to land cosmos today? I can wait and land this after.

JakeHillion:
Updated for scx_cosmos

PR #2400, opened 2025-07-16
Add scx_events: an io_uring-based event reactor

silvanshade:
This PR adds a new crate, `scx_events`, which defines an `io_uring`-based event reactor.

# Motivation

This work was originally motivated by wanting to help complete:

- https://github.com/sched-ext/scx/pull/1207

I mentioned in a [comment](https://github.com/sched-ext/scx/pull/1207#issuecomment-2945619499) that I intended to implement support for Netlink events associated with the AMD P-State Preferred Core Ranking updates.

After several iterations, it became clear that a more general reactor design would be useful.

The logic required to handle Netlink messaging cleanly was already substantial, so generalizing the approach and extracting it into its own crate made sense.

This also created an opportunity to focus on performance while providing a safe interface, where the unsafe implementation details of `io_uring` could be kept abstract. 

# Design

### Core Architecture

- There is a struct [`EventReactor`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor.rs#L42-L48) which launches a run-loop from [`EventReactor::new`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor.rs#L58-L123).

- The run-loop in [`EventReactorThread::run`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor/thread.rs#L41-L96) processes `io_uring` task completions in a loop.

- The `io_uring` tasks are managed by a [`TaskContext`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor/task/context.rs#L113-L118).

- The `TaskContext` [installs](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor/task/context.rs#L167-L335) tasks to `io_uring` and [performs](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor/task/context.rs#L338-L493) dispatching for task completion.

### Memory and Task Management

- The `TaskContext` attaches and pins task-related user data to an [`ArenaSlab`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/arena.rs#L32-L56).

- The `ArenaSlab` uses pinned addresses to [retrieve](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/arena.rs#L267-L303) user data from `io_uring` cqes to avoid lookups.

- The `ArenaSlab` minimizes allocation churn by [bump-allocating slots](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/arena.rs#L138-L184) which get [recycled](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/arena.rs#L351-L361) on task completion.

### Netlink Messaging Support

- The `EventReactor` supports [receiving](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor/task/context.rs#L266-L302) and [zero-copy parsing](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/events/netlink/genl/acpi.rs#L102-L134) Generic Netlink ACPI messages.

- The `EventReactor` supports [receiving](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor/task/context.rs#L304-L335) and [zero-copy parsing](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/events/netlink/kobject_uevent.rs#L77-L150) Kobject Uevent messages.

- [`neli`](https://docs.rs/neli/latest/neli/) is used for handling the framing for generic netlink messages.

- [`winnow`](https://docs.rs/winnow/latest/winnow/) and [`zerocopy`](https://docs.rs/zerocopy/latest/zerocopy/) are used for defining the zero-copy parsers.

- The Generic Netlink ACPI parsing includes support for [`ACPI_PROCESSOR_NOTIFY_HIGHEST_PERF_CHANGED`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/events/netlink/genl/acpi.rs#L235-L255).

### Resource Management

- The `EventReactor` minimizes syscalls through multi-shot ops, bundling, and groups for buffers and files.

- The `EventReactor` manages buffer rings through [`BufferRingRegion`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor/buffer.rs#L19-L34) and [`BufferRing`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/reactor/buffer.rs#L191-L232).

- The `EventReactor` [data](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/data.rs) can be accessed through a shared interface or client notification.

### Data Access and Client Notification

- Notifications use [`chute`](https://github.com/tower120/chute) MPMC channels, which provide fast, low-overhead, lock-free broadcasting.

- The `EventReactor` data provides fast cached, lock-free, [`RCU`](https://github.com/sched-ext/scx/blob/b037bffe8f703999393a21d1cb93fd032efc940d/rust/scx_events/src/data/amd_pstate_prefcore_rankings.rs#L9-L15) access via [arc-swap](https://docs.rs/arc-swap/latest/arc_swap/) where possible.

# TODO

Most of the basic design is in place and described above.

I've documented as much as I can and have added some tests and benchmarks for some parts but more is planned.

I've also ensured existing tests and code compile without issue with [cargo careful](https://github.com/RalfJung/cargo-careful) and [cargo miri](https://github.com/rust-lang/miri).

Remaining work:

- [ ] File IO (open/close/read) for sysfs paths (e.g., pref core data)
- [ ] Refactor the interface to pref core data for `Topology`
- [ ] Support for handling CPU hotplug events from Kobject Uevent messages
- [ ] Notification support
- [ ] Task cancellation
- [ ] Mock IO for better test coverage


hodgesds:
I'm not sure this approach fits well with most of the existing schedulers. I think what maybe makes more sense is reading the value from BPF (ex: #2441). Then a kprobe can be attached to [`sched_set_itmt_core_prio`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/x86/kernel/itmt.c#n164) to update userspace via a message over a BPF ringbuffer (if the scheduler wants topology updates in userspace). The kprobe can easily be found with `bpftrace`:
```
$ bpftrace -l | grep sched_set_itmt_core_prio
fentry:vmlinux:sched_set_itmt_core_prio
kprobe:sched_set_itmt_core_prio
```

silvanshade:
> I'm not sure this approach fits well with most of the existing schedulers.

Is there anything specific about the design that can you elaborate on in this regard?

As far as I can tell it should be fairly straight forward to integrate into most of the schedulers at least for the existing functionality it would provide (pref-core data).

> I think what maybe makes more sense is reading the value from BPF (ex: #2441). Then a kprobe can be attached to [`sched_set_itmt_core_prio`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/x86/kernel/itmt.c#n164) to update userspace via a message over a BPF ringbuffer (if the scheduler wants topology updates in userspace).

I agree your suggestion definitely sounds like a much more precisely targeted solution.

I guess the question then is whether there is utility in having an io_uring event reactor like this available for schedulers to use in a more general sense.

I think it could be useful as a way to more methodically organize some of the random IO and syscalls and to be more efficient about it but I also understand what I'm proposing is a fairly substantial new system to integrate.

If this is unlikely to be adopted, that's okay. Should I discontinue the work on this then?

hodgesds:
> Is there anything specific about the design that can you elaborate on in this regard?

This adds a lot of dependencies in both user space and kernel space. What happens if a scheduler runs on a kernel with `CONFIG_IO_URING` not set? Is netlink always available for ACPI events in every kernel config? Adding something to a scheduler means it has to run on a variety of devices and kernel configs.

> I guess the question then is whether there is utility in having an io_uring event reactor like this available for schedulers to use in a more general sense.

For what use cases? I don't think it is primarily a problem that is trying to be solved. Maybe something for rustland that does userspace scheduling, but there are a bunch of complexities there about not having userspace page fault in the hotpath.

> I think it could be useful as a way to more methodically organize some of the random IO and syscalls and to be more efficient about it but I also understand what I'm proposing is a fairly substantial new system to integrate.

Is that a problem? Does your solution work on any kernel config?

> If this is unlikely to be adopted, that's okay. Should I discontinue the work on this then?

It maybe makes sense as a separate crate, but it's not clear what problem is being solved. I think your solution is well written and tested, but I think the amount of dependencies that it adds are rather high.

silvanshade:
Thanks for the more detailed response.

> > Is there anything specific about the design that can you elaborate on in this regard?
> 
> This adds a lot of dependencies in both user space and kernel space. What happens if a scheduler runs on a kernel with `CONFIG_IO_URING` not set?

A kernel configured with `CONFIG_IO_URING` completely disabled is going to present a difficulty.

EDIT: Actually, I don't think it would be that difficult to provide a fallback implementation for this interface for kernels where io_uring is disabled. It would require implementing the operations directly and would lose out on efficiency compared to the io_uring version but would still provide a clean interface.

Is there in fact a known set of minimal features that all schedulers expect the kernel to support? Is there a test configured for that on CI that I could reference?

> Is netlink always available for ACPI events in every kernel config? Adding something to a scheduler means it has to run on a variety of devices and kernel configs.

I don't immediately know the answer to that.

But this approach wasn't designed under the assumption that all kernels would need to support ACPI Netlink.

The intention was that the Netlink ACPI events logic would only engage for hosts where the AMD 3D V-Cache Performance Optimizer driver was detected. Maybe some other subset could be engaged for more general pref-core support.

It's relatively easy to enable or disable specific event handlers from being installed depending on the host configuration.

> > I guess the question then is whether there is utility in having an io_uring event reactor like this available for schedulers to use in a more general sense.
> 
> For what use cases? I don't think it is primarily a problem that is trying to be solved. Maybe something for rustland that does userspace scheduling, but there are a bunch of complexities there about not having userspace page fault in the hotpath.

I don't know. I began implementing this under the impression that the desired approach was to listen for netlink events because that was what was suggested in the original PR thread.

I don't yet have a lot of detailed knowledge of the rest of this project so I don't think I can give a good answer to this.

> > I think it could be useful as a way to more methodically organize some of the random IO and syscalls and to be more efficient about it but I also understand what I'm proposing is a fairly substantial new system to integrate.
> 
> Is that a problem?

Just as an example, the original approach relied on individual file read calls to the sysfs entries. Using batched reads with io_uring is at least an improvement over that. But getting the data through BPF as you suggested is a better solution for this particular case.

Anything else that might end up doing similar things involving IO could benefit from having this interface.

I don't know if there is any general utility in having support for netlink or not because I don't know enough about the needs of the schedulers to be able to provide good examples.

> Does your solution work on any kernel config?

It won't work everywhere, no. It may not be appropriate for very minimal systems without io_uring support.

Does it need to work with every scheduler? Will every scheduler actually support pref-core data?

> > If this is unlikely to be adopted, that's okay. Should I discontinue the work on this then?
> 
> It maybe makes sense as a separate crate, but it's not clear what problem is being solved. I think your solution is well written and tested, but I think the amount of dependencies that it adds are rather high.

One reason I implemented it as a separate crate was to avoid the need for every scheduler to integrate it.

Regarding dependencies, if you look at the `Cargo.lock`, there aren't actually that many completely new dependencies.

At a glance, its mostly `arc-swap`, `fixedbitset`, `neli`, `rapidhash`, `rustix-uring`, `triomphe`, `zerocopy`.

Half of those could be dropped at the cost of implementing more stuff manually and maybe less efficiency.

Anyway, I appreciate the feedback. I'm happy to continue working on this if there's interest.

If not, it's okay to close this and I will focus on other things.

PR #2403, opened 2025-07-16
atq: Add sized ATQ helper

hodgesds:
Add a helper method to create a sized ATQ. This is used in cases where an absolute size of the ATQ is useful, such as ATQs for load balancing.

### DIFF
--- a/scheds/include/lib/atq.h
+++ b/scheds/include/lib/atq.h
@@ -17,8 +17,9 @@ struct scx_atq {
 
 typedef __arena scx_atq, scx_atq_t;
 
-u64 scx_atq_create_internal(bool fifo);
-#define scx_atq_create(fifo) (scx_atq_create_internal((fifo)))
+u64 scx_atq_create_internal(bool fifo, size_t capacity);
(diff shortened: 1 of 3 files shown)


scheds/include/lib/atq.h
```diff
@@ -17,8 +17,9 @@ struct scx_atq {
 
 typedef __arena scx_atq, scx_atq_t;
 
-u64 scx_atq_create_internal(bool fifo);
-#define scx_atq_create(fifo) (scx_atq_create_internal((fifo)))
+u64 scx_atq_create_internal(bool fifo, size_t capacity);
+#define scx_atq_create(fifo) scx_atq_create_internal((fifo), SCX_ATQ_MAX_CAPACITY)
+#define scx_atq_create_size(fifo, capacity) scx_atq_create_internal((fifo), (capacity))
```
JakeHillion, line 22: How come we're using macros for this sort of thing instead of `__always_inline` functions?

hodgesds: Just following the style of @etsal

PR #2405, opened 2025-07-16
scx_loader: add config search path for distributions

vnepogodin:
that allows distributions to ship with default config at:
- /usr/share/scx_loader/config.toml
- /usr/share/scx_loader.toml

user can overwrite it with own configuration at existing paths:
- /etc/scx_loader/config.toml
- /etc/scx_loader.toml


arighi (approved):
LGTM

PR #2407, opened 2025-07-16
scx_loader: Support build-time configuration of vendordir

ReillyBrogan:
Some Linux distributions have specific directories for package-shipped default config files. For example Solus, AerynOS, and Clear all use `/usr/share/defaults`, while SUSE uses `/usr/etc`. Rather than hardcoding `/usr/share/scx_loader/config.toml` as the config file location instead allow it to be set at build-time.


sirlucjan:
@ReillyBrogan Thanks for your contribution! Could you rebase your PR against: https://github.com/sched-ext/scx/pull/2406 ? Thanks.

ReillyBrogan:
> @ReillyBrogan Thanks for your contribution! Could you rebase your PR against: #2406 ? Thanks.

Done

PR #2410, opened 2025-07-17
rust utils: addressing clippy warnings

devnexen:
mostly string formats and variables/const, useless cast after var shadowing first cast and more rust idioms usage overall.


JakeHillion (approved):
Looks cleaner to me, thanks.

PR #2411, opened 2025-07-17
scx_cosmos: Introduce flat idle CPU scan

arighi:
On systems with simple topologies, such as those without SMT and with all the CPUs on a single last-level cache (LLC), sequentially scanning CPUs can be more efficient than using cpumasks to locate idle ones.

Therefore, introduces the new option --flat-idle-scan / -i to enable such lightweight idle CPU scanning.

This mode can introduce inaccuracy and unbalanced CPU usage, but it can provide consistent improvements with certain benchmarks (i.e., heavy message passing workloads).

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -36,6 +36,11 @@ private(COSMOS) struct bpf_cpumask __kptr *primary_cpumask;
 const volatile bool primary_all = true;
 
 /*
+ * Enable flat iteration to find idle CPUs (fast but inaccurate).
+ */
+const volatile bool flat_idle_scan = true;
+
+/*
  * Enable cpufreq integration.
  */
 const volatile bool cpufreq_enabled = true;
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -234,6 +239,69 @@ static inline u64 shared_dsq(s32 cpu)
...
+	s32 i;
+
+	if (is_cpu_idle(prev_cpu))
+		return prev_cpu;
+
+	if (is_pcpu_task(p))
+		return -EBUSY;
+
+	bpf_for(i, 0, nr_cpu_ids) {
+		s32 cpu = (last_cpu + i) % nr_cpu_ids;
```
JakeHillion: Is it intended to use the moving `last_cpu` here? It seems like the loop intends to cover every CPU, but that won't happen if this method races. It seems redundant to still loop over all the CPUs even though other racers have already checked they're idle.

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -234,6 +239,69 @@ static inline u64 shared_dsq(s32 cpu)
...
+
+		if (cpu == prev_cpu)
+			continue;
+
+		if (bpf_cpumask_test_cpu(cpu, p->cpus_ptr) &&
+		    is_cpu_idle(cpu)) {
+			last_cpu = (cpu + 1) % nr_cpu_ids;
```
JakeHillion: Could we switch `last_cpu` to a `u32` with defined overflow and drop the mod? Every use case of `last_cpu` mods it again anyway.

JakeHillion (approved):

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
arighi: > Is it intended to use the moving `last_cpu` here? It seems like the loop intends to cover every CPU, but that won't happen if this method races. It seems redundant to still loop over all the CPUs even though other racers have already checked they're idle.

Oh yes, good catch! We should read `last_cpu` in a local var to get the starting CPU and then use the local variable to loop. It's still racy, but a bit less. I'll apply this change and re-run some tests, thanks!

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
arighi: Yes, I like that. Will change last_cpu to u32 and drop the mod.

JakeHillion (approved):

arighi:
Re-ran all the benchmark with the last change, no significant change in the results, but the code is cleaner now and less racy. Thanks @JakeHillion for the review!

PR #2412, opened 2025-07-17
nix: switch to fenix to better pin rust dependencies

JakeHillion:
Currently CI and anyone using the Nix flake gets their rust dependencies from nixpkgs. This is generally fine as they are locked, but we have less control over which version is selected than with tools like rustup.

Switch to fenix[1] which gives equivalent control to rustup. This enables testing on nightly toolchains with a simple change. It's mostly an ergonomic benefit for now. Required some minor clippy fixes for the updated clippy version.

Test plan:
- CI

[1] https://github.com/nix-community/fenix

### DIFF
--- a/.github/include/flake.lock
+++ b/.github/include/flake.lock
@@ -1,5 +1,26 @@
 {
   "nodes": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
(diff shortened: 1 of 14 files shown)


hodgesds:
I'm guessing folks will need to update their clippy versions.

JakeHillion:
> I'm guessing folks will need to update their clippy versions.

Yep, CI remains the source of truth for crates with it enabled, and you can still use `nix develop --command cargo clippy -p scxtop --no-deps` to exactly replicate it.

PR #2414, opened 2025-07-17
scx_p2dq: Add ATQ for migrations

hodgesds:
Add the option to use a migration ATQ. When ATQs are enabled migrations will use a fixed size migration ATQ rather than a DSQ.


JakeHillion:
Hmm, this actually has me a bit stumped for chaos. The build caught this which is good news, but it's expecting to be able to perform a move from a built in DSQ (the delay DSQs) to an ATQ. As I understand it that hasn't been done at any point, and I'm not sure it fits the sched_ext model...

Unless you've got any good ideas, would you mind preventing chaos from starting if this p2dq flag is set & `scx_bpf_error` in the named cases in that failing switch statement?

JakeHillion (approved):
ack for chaos too

PR #2416, opened 2025-07-17
scxtop: fix trace

yaakov-stein:
#2398 breaks `scxtop trace` with:
```
libbpf: prog 'on_sched_wait': can't attach BPF program without FD (was it loaded?)
Error: failed to attach BPF program

Caused by:
```
(I believe due to the `?` in the BPF method and not [autoloading](https://github.com/sched-ext/scx/blob/d4dd7b1bc73028af5e9a65ee409633cd00b5c8f2/rust/scx_utils/src/compat.rs#L233).) This fixes that by removing the cond (I don't see a reason for this to be cond) and satisfying the verifier.

We also should add the `scxtop trace` command to the CI to ensure it runs without errors (I incorrectly assumed that we did..).

### DIFF
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -921,7 +921,7 @@ int BPF_PROG(on_sched_exec, struct task_struct *p, u32 old_pid, struct linux_bin
 	return 0;
 }
 
-SEC("?tp_btf/sched_process_wait")
+SEC("tp_btf/sched_process_wait")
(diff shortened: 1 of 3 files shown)


hodgesds:
It still might be worth doing setting autoloading to false if the symbol doesn't exist.

yaakov-stein:
> It still might be worth doing setting autoloading to false if the symbol doesn't exist.

For all of the functions we currently attach (`sched_process_exec|fork|exit` etc.)? Are there modern kernels that don't have these (I assumed not)?

devnexen (approved):
thx

PR #2418, opened 2025-07-18
scx_lavd: Mostly refactoring and clean up, and one preemption optimization.

multics69:
This PR includes the refactoring and clean-up patches, which do not cause any functional & behavioral changes, as follows:
- scx_lavd: Add const to no_wake_sync. 
- scx_lavd: Extend test_task_flag() for multi-bit testing. 
- scx_lavd: Keep running task's flag on cpu_ctx.
- scx_lavd: Rename cpuc->stopping_tm_est_ns to cpuc->est_stopping_clk. 
- scx_lavd: Keep track of running_clk at cpu_ctx. 
- scx_lavd: Refactoring greedy penalty calculation.
- scx_lavd: Refactoring preemption-related code.
- scx_lavd: Remove redundant time slice calculation. 

Additionally, it includes one optimization for preemption: when a particular CPU is chosen for an enqueuing task, consider that CPU as a preemption candidate. This way, we can increase the chance that the enqueued task runs on the chosen CPU.
- scx_lavd: Prioritize the chosen CPU for preemption.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -904,6 +904,7 @@ void BPF_STRUCT_OPS(lavd_running, struct task_struct *p)
 	 */
 	cpuc->flags = taskc->flags;
 	cpuc->lat_cri = taskc->lat_cri;
+	cpuc->running_clk = now;
 	cpuc->est_stopping_clk = get_est_stopping_clk(taskc, now);
 
 	/*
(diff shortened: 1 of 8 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -372,6 +353,14 @@ static void update_stat_for_running(struct task_struct *p,
 		cpuc->sum_perf_cri += taskc->perf_cri;
 	}
 
+	/*
+	 * Update running task's information for preemption
+	 */
+	cpuc->flags = taskc->lat_cri;
```
hodgesds: should this be using the tasks flags?
```suggestion
	cpuc->flags = taskc->flags;
```

multics69: Thanks! @hodgesds

PR #2424, opened 2025-07-18
scx_cosmos: Use flat idle scan only when the system is sufficiently idle

arighi:
On systems with a many CPUs sharing a single LLC, enabling flat idle scanning (-i) can be really beneficial. However, as system load increases, scanning all CPUs to find an idle one becomes more costly than using the built-in cpumask-based method.

To find the optimal balance, use the global CPU load to decide whether to use the flat idle scanning or the built-in idle CPU selection policy: when the system is mostly idle (load below the busy threshold), use flat scanning for faster placement; otherwise, fall back to the default idle CPU selection policy.

This adaptive approach provides the advantages of both methods, especially on large systems with a flat topology.

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -318,7 +318,7 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
 	 * Use the lightweight idle CPU scanning if flat idle scan is
 	 * enabled and all the CPUs are included in the primary domain.
 	 */
-	if (flat_idle_scan && primary_all)
+	if (flat_idle_scan && primary_all && !is_system_busy())
 		return pick_idle_cpu_flat(p, prev_cpu);
 
 	/*


multics69 (approved):
The change makes sense. Do you know how expensive the built-in cpumask based approach is? And, when the system is mostly idle, how much performance benefit is expected using loop-based idle CPU scanning?

arighi:
> The change makes sense. Do you know how expensive the built-in cpumask based approach is? And, when the system is mostly idle, how much performance benefit is expected using loop-based idle CPU scanning?

Obviously it really depends on the particular topology, I've been running some I/O-bound benchmarks (MySQL / Redis) to stress more the sleep/wakeup patterns (and consequently `ops.select_cpu()`) on a 72-cores arm64 system with all the CPUs on the same LLC, and I've seen a consistent speedup that can go from 2.5% up to 5% in some cases.

However, when the system is close to saturation I see an equivalent performance drop (that's why I added the extra check to use the flat scan based on the load with this PR). Probably this has a less impact on systems with less CPUs, but even in this case it's probably safer to use the built-in idle policy when the system is saturated.

PR #2427, opened 2025-07-18
selftests: Add topology selftest

hodgesds:
Add BPF selftests for BPF arena topology and helper functions.


arighi (approved):
Looks good, thanks!

PR #2428, opened 2025-07-18
scx_lavd: Sample dsq_consume latencies

daidavid:
Sample time for scx_bpf_dsq_move_to_local() latencies to determine how much contention is happening at each DSQ.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -149,12 +149,22 @@ static void plan_x_cpdom_migration(void)
 	sys_stat.nr_stealee = nr_stealee;
 }
 
-static bool consume_dsq(u64 dsq_id)
+static bool consume_dsq(struct cpdom_ctx *cpdomc)
 {
+	bool ret;
+	u64 before = 0;
+
+	if (is_monitored)
+		before = bpf_ktime_get_ns();
 	/*
 	 * Try to consume a task on the associated DSQ.
 	 */
-	return scx_bpf_dsq_move_to_local(dsq_id);
+	ret = scx_bpf_dsq_move_to_local(cpdomc->id);
+
+	if (is_monitored)
+		cpdomc->dsq_consume_lat = time_delta(bpf_ktime_get_ns(), before);
+
+	return ret;
 }
 
 static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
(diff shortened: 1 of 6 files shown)


scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -149,12 +149,17 @@ static void plan_x_cpdom_migration(void)
...
+static bool consume_dsq(struct cpdom_ctx *cpdomc)
 {
+	bool ret;
+	u64 before = scx_bpf_now();
 	/*
 	 * Try to consume a task on the associated DSQ.
 	 */
-	return scx_bpf_dsq_move_to_local(dsq_id);
+	ret = scx_bpf_dsq_move_to_local(cpdomc->id);
+	cpdomc->dsq_consume_lat = time_delta(scx_bpf_now(), before);
```
multics69: Since `consume_dsq()` is one of the hottest paths and getting the current timestamp is not free, it will be better to put this under `is_monitored`. In this way, we can avoid timestamping overhead at the critical path when the scheduler is not monitored (`--monitor-sched-samples`).

Additionally, `scx_bpf_now()` is slightly less precise, as it essentially piggybacks on the rq clock to avoid repetitive `rdtsc`. Here we are measuring the duration of a single function; we will need a more precise one: `bpf_ktime_get_ns()`.

scheds/rust/scx_lavd/src/bpf/introspec.bpf.c
```diff
@@ -48,6 +53,9 @@ int submit_task_ctx(struct task_struct *p, struct task_ctx *taskc, u32 cpu_id)
 	m->taskc_x.thr_perf_cri = sys_stat.thr_perf_cri;
 	m->taskc_x.nr_active = sys_stat.nr_active;
 	m->taskc_x.cpuperf_cur = cpuc->cpuperf_cur;
+	//Refactor this when per-cpu DSQs are added
```
multics69: Nits: Let's just use a C-style comment for consistency. ` /* ... /* `

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -101,6 +101,7 @@ struct cpdom_ctx {
 	u32	cur_util_sum;			    /* the sum of CPU utilization in the current interval */
 	u32	cap_sum_active_cpus;		    /* the sum of capacities of active CPUs in this domain */
 	u32	cap_sum_temp;			    /* temp for cap_sum_active_cpus */
+	u32	dsq_consume_lat;		    /* latency to consume from dsq, shows how contended the dsq is*/
```
multics69: `is*/` ==> `is */`

multics69 (changes requested):
Thanks @daidavid ! I added a few minor comments.

hodgesds:
fyi you can get these metrics out of band with `scxtop`, but it's a neat idea to implement this directly in the scheduler.

multics69 (approved):
Thanks!

PR #2431, opened 2025-07-19
scx_tickless: Add support for preferred cores

arighi:
Since the scheduler already dispatches tasks by iterating over all CPUs, implementing preferred core prioritization is straightforward.

This change introduces a simple mechanism to prefer higher-capacity CPUs (based on capacity information provided by the topology crate). More exactly, the scheduler prioritizes dispatching to the fastest full-idle SMT cores first, followed by the fastest idle CPUs.

CPU capacity is currently evaluated only at scheduler startup, so runtime capacity changes are not handled unless the scheduler is restarted.

### DIFF
--- a/scheds/rust/scx_tickless/src/main.rs
+++ b/scheds/rust/scx_tickless/src/main.rs
@@ -31,6 +31,7 @@ use log::warn;
 use log::{debug, info};
 use scx_stats::prelude::*;
 use scx_utils::build_id;
+use scx_utils::compat;
 use scx_utils::scx_ops_attach;
 use scx_utils::scx_ops_load;
 use scx_utils::scx_ops_open;
@@ -53,7 +54,9 @@ struct Opts {
 
     /// Define the set of CPUs, represented as a bitmask in hex (e.g., 0xff), dedicated to process
     /// scheduling events.
-    #[clap(short = 'm', long, default_value = "0x1")]
+    ///
+    /// 0x0 = autodetect the slowest CPU to process scheduling events.
+    #[clap(short = 'm', long, default_value = "0")]
     primary_domain: String,
 
     /// Maximum scheduling slice duration in microseconds (applied only when multiple tasks are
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_tickless/src/main.rs
```diff
@@ -135,24 +131,46 @@ impl<'a> Scheduler<'a> {
             warn!("nohz_full is not enabled in the kernel");
         }
 
+        // Generate the list of available CPUs sorted by capacity in descendind order.
+        let mut cpus: Vec<_> = topo.all_cpus.values().collect();
```
hodgesds, line 135: On AMD CPUs with X3D cores are they clocked lower than regular cores (I don't have any X3D CPUs)? Otherwise, this seems correct.

hodgesds (approved):

scheds/rust/scx_tickless/src/main.rs
arighi: Yeah I also don't have any AMD with X3D vcache. For what I understand the only difference is that the CPUs capacity can change depending if the user selects cache mode vs frequency mode. When it happens we need to restart the scheduler to re-generate the sorted cpus array. It'd nice to have a mechanism that does that automatically, I'll take a look at this next.

ptr1337: Zen4 X3D CCD0 has a lower frequency then CCD1

arighi: > Zen4 X3D CCD0 has a lower frequency then CCD1

Just to make sure I understand, that means the CPUs on CCD0 have lower frequency then the CPUs on CCD1?

In that case everything should work because the frequency determines the CPU capacity and CPUs are sorted by capacity in descending order, so the scheduler will prioritize those in CCD1.

silvanshade: > It'd nice to have a mechanism that does that automatically, I'll take a look at this next.

@arighi I am working on something that should provide RCU-style access to pref-core data as it is updated by listening to netlink events in this PR:

- https://github.com/sched-ext/scx/pull/2400

silvanshade: > > Zen4 X3D CCD0 has a lower frequency then CCD1
> 
> Just to make sure I understand, that means the CPUs on CCD0 have lower frequency then the CPUs on CCD1?

Generally, yes, but it's not a significant difference like you might expect with big/little or performance/efficiency cores or something else along those lines. Most of the time the clocks on both CCDs are pretty close to the same range if all cores are loaded.

It's just that generally the frequency CCD tends to run the cores at slightly higher clocks and will likely boost higher, particularly under medium loads. But it's not guaranteed, and the user can even override that behavior with, for example, a custom voltage curve that pushes the CCD core.

PR #2434, opened 2025-07-20
scxtop: add v0 process view

yaakov-stein:
This adds an initial process view that is sorted by number of threads. The idea here is to get this out and quickly iterate with new features (more on that below). The process view will be focused on information that would be helpful for scheduler development and debugging as opposed to a general process view like top/btop. Following that idea, the v0 contains the last CPU/LLC/NUMA and DSQ that the process ran on.

This also changes the default view to be more of an overview, pushing the DSQ statistics to the scheduler view (`s`).

Example:
Screenshot 2025-07-20 at 9 48 53 AM

Upcoming Features:
* Turn into stateful widget - ability to scroll and filter by process name
* Sorted by Cpu Utilization
* Thread view (allows one to dive into a specific process)
* Prettier UI
* Additional fields (reach out to me here or in the discord if there are any that you think would be helpful)


tools/scxtop/src/lib.rs
```diff
@@ -66,7 +66,7 @@ pub const TRACE_FILE_PREFIX: &str = "scxtop_trace";
 pub const STATS_SOCKET_PATH: &str = "/var/run/scx/root/stats";
 pub const LICENSE: &str = "Copyright (c) Meta Platforms, Inc. and affiliates.
 
-This software may be used and distributed according to the terms of the 
+This software may be used and distributed according to the terms of the
```
hodgesds: Double check the help widget, IIRC this is there for a reason.

yaakov-stein: Ya, happens every time, `cargo fmt` gets rid of the extra space 🤦‍♂️ . Going to update it so it's not a problem going forward

yaakov-stein: #2436

hodgesds:
LGTM

hodgesds (approved):

tools/scxtop/src/proc_data.rs
```diff
@@ -0,0 +1,83 @@
...
+            threads,
+            data: EventData::new(max_data_size),
+            max_data_size,
+        };
+
+        Ok(proc_data)
+    }
+
+    /// Creates a new ProcData from a given tgid.
+    pub fn from_tgid(tgid: i32, max_data_size: usize) -> Result<ProcData> {
```
hodgesds, line 62: should be easy to generate tests for this

PR #2435, opened 2025-07-21
scx_chaos: separate kprobe delays

yaakov-stein:
Previously, kprobe random delays where tied to the standard random delays. This also included a bug where one would be required to set the random delay frequency argument, even after setting the kprobe delay frequency argument.

This PR decouples kprobe delays from random delays by creating a new "trait" (it still is triggered differently than the other traits and is not placed into the trait array) and adding independent args for kprobes. This also allows one to use random delays and kprobe delays with different settings.

Testing:
* Ran `scx_chaos --stats 1 --kprobes-for-random-delays schedule --kprobe-random-delay-min-us 100 --kprobe-random-delay-max-us 10000` for > 30 minutes, once with default frequency and once with `--kprobe-random-delay-frequency .5`. Statistics show that kprobe delays occur regularly and the numbers line up with the different frequencies.


JakeHillion (approved):
Looks great, thanks!

PR #2438, opened 2025-07-21
scx_p2dq: Refactor dispatch

hodgesds:
Refactor dispatch to add additional checks to avoid consuming from the same ATQ multiple times in the dispatch path. Add a check to always consume from the migration DSQ/ATQ before trying to load balance to reduce cross LLC migrations. Update load balancing mode to use a switch statement for better performance/readability.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -1307,8 +1307,10 @@ static __always_inline bool consume_llc(struct llc_ctx *llcx)
 	if (!llcx)
 		return false;
 
-	if (p2dq_config.atq_enabled &&
-	    scx_atq_nr_queued(llcx->mig_atq) > 0) {
+	if (p2dq_config.atq_enabled) {
+		if (scx_atq_nr_queued(llcx->mig_atq) == 0)
+			return false;
+
 		pid = scx_atq_pop(llcx->mig_atq);
 		p = bpf_task_from_pid((s32)pid);
 		if (!p) {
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -1630,6 +1635,29 @@ static __always_inline void p2dq_dispatch_impl(s32 cpu, struct task_struct *prev
 	    scx_bpf_dsq_move_to_local(cpuc->llc_dsq))
 		return;
 
+	if (topo_config.nr_llcs > 1) {
+		if (p2dq_config.atq_enabled) {
+			pid = scx_atq_pop(cpuc->mig_atq);
```
yaakov-stein, line 1640: If this atq is empty, this will return `(u64)NULL`. IIUC, on the next line, I believe this will get the idle task (pid 0), which could cause problems. Am I understanding that correctly?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -1307,8 +1307,10 @@ static __always_inline bool consume_llc(struct llc_ctx *llcx)
 	if (!llcx)
 		return false;
 
-	if (p2dq_config.atq_enabled &&
-	    scx_atq_nr_queued(llcx->mig_atq) > 0) {
+	if (p2dq_config.atq_enabled) {
+		if (scx_atq_nr_queued(llcx->mig_atq) == 0)
+			return false;
+
 		pid = scx_atq_pop(llcx->mig_atq);
```
arighi, line 1314: Similar to what @yaakov-stein pointed out, `scx_atq_nr_queued() + scx_atq_pop()` are not atomic, so potentially the first one can return a value > 0, but then when we do the pop the queue could be empty. Is there any benefit checking the amount of tasks in the queue? Can we just do the pop and check if pid != NULL?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: Yeah, that's correct.

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: I'm going to do another PR on top of #2492 as that seems to fix some of the other issues. Will close this one.

hodgesds:
closing this as some of these cases are handled.

PR #2440, opened 2025-07-21
scx_lavd: Boost the time slice of long-running tasks.

multics69:
A comput-intensive task can run longer than a regular time slice. Such a long-running task will be scheduled in and out (i.e., involuntary context switch) to finish its job. Involuntary context switch incurs higher cache misses and scheduling overhead. Moreover, if consumer tasks are waiting for the outcome of long-running tasks, it also incurs cascading delays in the consumer tasks.

We assign a longer timeslice for long-running tasks (i.e., slice boosting up to 1 sec) to address the problem of involuntary context switch and its side effect (e.g., cascading delay). Note that if a task is pinned to a particular CPU, we should not increase its time slice to avoid the pinned tasks being delayed indefinitely.

In addition to slice boosting, we extend the preemption mechanism so that the time slice of slice-boosted tasks is reduced to the regular slice to avoid latency increases caused by slice-boosted tasks.


scheds/rust/scx_lavd/src/bpf/preempt.bpf.c
```diff
@@ -207,37 +213,142 @@ static void ask_cpu_yield(struct cpu_ctx *victim_cpuc)
...
+	/*
+	 * It is okay to set p->scx.slice to zero since this is supposed to
+	 * be called by ops.tick(), which is the scheduling point.
+	 */
+	p->scx.slice = new_slice;
+
+	cpuc->nr_preempt++;
+}
+
+static void give_up_cpu_now(struct task_struct *p, struct cpu_ctx *cpuc)
```
daidavid: Is this being used anywhere?

scheds/rust/scx_lavd/src/bpf/intf.h
```diff
@@ -126,12 +127,12 @@ struct task_ctx {
 	u32	lat_cri;		/* final context-aware latency criticality */
 	u32	lat_cri_waker;		/* waker's latency criticality */
 	u32	perf_cri;		/* performance criticality of a task */
-	u32	slice_ns;		/* time slice */
+	u64	slice;			/* time slice */
```
hodgesds, line 130: I get this matches the kernel, but I like units on variables 😄

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -912,6 +976,35 @@ void BPF_STRUCT_OPS(lavd_running, struct task_struct *p)
...
+	if (!cpuc || !taskc) {
+		scx_bpf_error("Failed to lookup context for task %d", p->pid);
+		return;
+	}
+
+	now = scx_bpf_now();
+	account_task_runtime(p, taskc, cpuc, now);
+
+	/*
+	 * If this task is slice-boosted and there is a pinned task that
```
hodgesds, line 1021: One other idea is maybe if the task is a nice take it could just yield, but I guess preemption handling is done elsewhere.

scheds/rust/scx_lavd/src/bpf/preempt.bpf.c
```diff
@@ -207,37 +213,142 @@ static void ask_cpu_yield(struct cpu_ctx *victim_cpuc)
...
+		if (dur >= sys_stat.slice) {
+			init_prm_by_task(&prm_t, taskc, now);
+			if (can_x_kick_cpu2(&prm_t, &prm_c, cpuc_victim)) {
+				reset_cpu_flag(cpuc_victim, LAVD_FLAG_SLICE_BOOST);
+				goto kick_out;
+			}
+		}
+
+		/*
+		 * If the enqueuing task is affinitized, it should run on
```
hodgesds, line 331: I see now, this makes sense!

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -256,13 +256,79 @@ static void advance_cur_logical_clk(struct task_struct *p)
...
+	 * (i.e., taskc->avg_runtime > sys_stat.slice), that means the task
+	 * could be scheduled out due to a shorter time slice than required.
+	 * In this case, let's consider boosting task's time slice.
+	 *
+	 * However, if there are pinned tasks waiting to run on this CPU,
+	 * we do not boost the task's time slice to avoid delaying the pinned
+	 * task that cannot be run on another CPU.
+	 */
+	if (!no_slice_boost && !cpuc->nr_pinned_tasks &&
+	    (taskc->avg_runtime >= sys_stat.slice)) {
```
daidavid, line 278: n00b question, how often does taskc->avg_runtime >= sys_stat.slice occur in a very loaded system before slice_boost kicks in? IIUC, scx balance/pick_task seems to only select previously ran task/refill slices if it can't find a task to run.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -256,13 +256,79 @@ static void advance_cur_logical_clk(struct task_struct *p)
...
+		}
+
+		/*
+		 * When the system is under high load, we will boost the time
+		 * slice of only latency-critical tasks, which are likely in
+		 * the middle of a task chain. Also, increase the time slice
+		 * proportionally to the latency criticality up to 2x the
+		 * regular time slice.
+		 */
+		if (taskc->lat_cri > sys_stat.avg_lat_cri) {
```
daidavid, line 312: Is it possible that wake_freq/wait_freq can be skewed with the slice_boost? And can that lead to poor behavior when determining lat_cri?

scheds/rust/scx_lavd/src/bpf/preempt.bpf.c
multics69: Hehe, no. It is a leftover after a few rounds of refactoring & optimizations. I will remove it.

scheds/rust/scx_lavd/src/bpf/intf.h
multics69: Initially, I put `_ns` everywhere. Later, I realized that ns is the only unit used in the scheduler. :-)

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Basically, `avg_runtime` is the time duration that a task runs until going to sleep, accounting for its accumulated runtime (`acc_runtime) since it becomes runnable. So `avg_runtime` should reflect how much time a task needs to finish the job without interruption.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: It won't affect a lot because how frequently a task wakes up others or how often a task waits for an event is the property of a task itself.

Of course, it will be affected a bit depending on the exact deadline calculation and the system load. However, that won't affect much because the high system load will penalize all the tasks equally. Moreover, the log-scaled value of frequency is resilient to small fluctuations in frequency.

PR #2441, opened 2025-07-21
scx_p2dq: Add CPU priority

hodgesds:
On some systems CPUs can have different performance priorities. With kernels that have itmt support use the per CPU value for the relative performance.

See the following files for implementations/definitions:
	arch/x86/include/asm/topology.h
	arch/x86/kernel/itmt.c
	drivers/cpufreq/intel_pstate.c
	drivers/cpufreq/amd-pstate.c
	
Add idle CPU tracking that uses a minheap to track idle CPUs by priority. When cpu_priority scheduling is enabled idle CPUs are ranked by their relative CPU performance and considered for idle CPU selection. This can improve latencies on some combination of workload/hardware support.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -140,8 +140,8 @@ const volatile struct {
 	.select_idle_in_enqueue = true,
 };
 
-extern int sched_core_priority __ksym;
-extern bool sched_itmt_capable __ksym;
+extern int sched_core_priority __ksym __weak;
+extern bool sched_itmt_capable __ksym __weak;
 
 const volatile u32 debug = 2;
 const u32 zero_u32 = 0;


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -300,6 +300,43 @@ static struct cpu_ctx *lookup_cpu_ctx(int cpu)
...
+		cpuc = bpf_map_lookup_elem(&cpu_ctxs, &zero_u32);
+	} else {
+		cpuc = bpf_map_lookup_percpu_elem(&cpu_ctxs,
+						  &zero_u32, cpu);
+	}
+
+	return cpuc;
+}
+
+SEC("kprobe/sched_set_itmt_core_prio")
```
multics69: It would be nice to attach the kprobe conditionally not to break non-intel/amd systems.

multics69 (approved):
LGTM. Do you have a concrete idea how to use the information in making scheduling decisions?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -297,6 +300,43 @@ static struct cpu_ctx *lookup_cpu_ctx(int cpu)
...
+int BPF_KPROBE(set_itmt_core_prio)
+{
+	struct cpu_ctx *cpuc;
+	int cpu, *prio;
+
+	if (!(bool)&sched_itmt_capable)
+		return 0;
+
+	bpf_for(cpu, 0, topo_config.nr_cpus) {
```
arighi: Can we get `prio` and `cpu` from the arguments of the kprobe and update only the affected CPU instead of refreshing all CPUs?

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: Good call!

scheds/rust/scx_p2dq/src/bpf/main.bpf.c
hodgesds: Sounds good!

hodgesds:
> s wondering why don't we read the priority directly from `sched_core_priority` when we need it, providing a simple helper like this:

Yeah, that seems better let me add that.

hodgesds:
> LGTM. Do you have a concrete idea how to use the information in making scheduling decisions?

I was talking to @etsal about maybe storing these in an arena (btree?) and then during idle cpu selection iterating over the preferred cores. Still need to think about the implementation a bit.

scheds/rust/scx_p2dq/src/main.rs
```diff
@@ -118,6 +118,7 @@ impl<'a> Scheduler<'a> {
         if opts.queued_wakeup {
             open_skel.struct_ops.p2dq_mut().flags |= *compat::SCX_OPS_ALLOW_QUEUED_WAKEUP;
         }
+        open_skel.struct_ops.p2dq_mut().flags |= *compat::SCX_OPS_KEEP_BUILTIN_IDLE;
```
JakeHillion, line 121: could you put this in `scx_p2dq::init_open_skel!` as it's common to both? also the `opts.queued_wakeup` even though it's not new in this PR

hodgesds: I might clean that up in a separate diff as `p2dq_mut` and `chaos_mut` are the methods to access the flags based on the struct_ops name. Rather annoying.

PR #2442, opened 2025-07-21
scxtop: fix sched hang prog attach

yaakov-stein:
On some systems, `sched_process_hang` is not enabled. Here, we were failing loudly unnecesarily. This switches the behavior to print a warning and continue.

### DIFF
--- a/tools/scxtop/src/main.rs
+++ b/tools/scxtop/src/main.rs
@@ -38,6 +38,7 @@ use libbpf_rs::RingBufferBuilder;
 use libbpf_rs::UprobeOpts;
 use log::debug;
 use log::info;
+use log::warn;
 use ratatui::crossterm::event::{KeyCode::Char, KeyEvent};
 use simplelog::{
     ColorChoice, Config as SimplelogConfig, LevelFilter, TermLogger, TerminalMode, WriteLogger,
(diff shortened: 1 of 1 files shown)


tools/scxtop/src/main.rs
```diff
@@ -139,6 +140,16 @@ fn attach_progs(skel: &mut BpfSkel) -> Result<Vec<Link>> {
     if let Ok(link) = skel.progs.on_pstate_sample.attach() {
         links.push(link);
     }
+    if let Ok(link) = skel.progs.on_sched_wait.attach() {
+        links.push(link);
+    } else {
+        warn!("tradepoint on_sched_wait is missing, tracepoint not loaded");
```
devnexen: nit: typo with tradepoint

tools/scxtop/src/main.rs
```diff
@@ -139,6 +140,16 @@ fn attach_progs(skel: &mut BpfSkel) -> Result<Vec<Link>> {
...
+    }
+    if let Ok(link) = skel.progs.on_sched_hang.attach() {
+        links.push(link);
+    } else {
+        warn!("tradepoint on_sched_hang is missing, tracepoint not loaded");
```
devnexen: ditto

hodgesds (approved):
LGTM, thanks for fixing

PR #2450, opened 2025-07-22
scxtop: add fields to process view

yaakov-stein:
This adds both an average slice duration (in ns) and maximum latency (in us) to the process view. Example:

Screenshot 2025-07-22 at 12 32 39 PM

On my machine, with ~1000 processes running, I timed `render_process_table` and found that the entire method takes ~3.5 ms without these and 3.5-4 ms with it. As we'll probably continue to add data, we should keep an eye on this.


hodgesds (approved):
nice!

PR #2456, opened 2025-07-22
scxtop: add pause to help

yaakov-stein:
Forgot to add the pause/unpause to the help page. Just having an empty character space is pretty confusing for a newcomer (it shows up as empty space), so if it is set as a lone empty character, we'll switch it to use the string "Space". If it's set to something else, or if it is set to multiple things (via config), it should be a bit clearer and we don't need to change it to "Space".


hodgesds (approved):
LGTM minus the new clippy updates.

PR #2461, opened 2025-07-23
scxtop: prevent trace crash

yaakov-stein:
Previously, one was allowed to switch views while tracing. This led to strange behavior - when the trace completes, it returns to the previous state, which in the case of one who switched away from the trace means they are returned to the trace screen even though it has completed (and they are stuck there until they click a new key). 

This restricts switching the view while the 1 second trace occurs.


JakeHillion (approved):
Makes sense

PR #2462, opened 2025-07-23
scx_cosmos: Introduce preferred idle CPU scan

arighi:
Commit a937a247 ("scx_cosmos: Introduce flat idle CPU scan") added support for a lightweight flat idle scanning.

This commit extends that logic by changing the iteration order: instead of scanning CPUs sequentially, it is also possible to use CPU ranking, prioritizing faster CPUs first.

This new mode can be enabled using the option --preferred-idle-scan.

Additionally, both flat and preferred idle CPU scan modes are now updated to support CPU prioritization based on SMT, NUMA, and primary scheduling domains (if enabled).

### DIFF
--- a/scheds/rust/scx_cosmos/src/main.rs
+++ b/scheds/rust/scx_cosmos/src/main.rs
@@ -311,7 +311,7 @@ impl<'a> Scheduler<'a> {
         // Normalize CPU busy threshold in the range [0 .. 1024].
         rodata.busy_threshold = opts.cpu_busy_thresh * 1024 / 100;
 
-        // Generate the list of available CPUs sorted by capacity in descendind order.
+        // Generate the list of available CPUs sorted by capacity in descending order.
         if opts.preferred_idle_scan {
             let mut cpus: Vec<_> = topo.all_cpus.values().collect();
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_cosmos/src/main.rs
```diff
@@ -285,12 +304,27 @@ impl<'a> Scheduler<'a> {
...
         rodata.cpufreq_enabled = !opts.disable_cpufreq;
         rodata.flat_idle_scan = opts.flat_idle_scan;
+        rodata.smt_enabled = smt_enabled;
         rodata.numa_enabled = opts.enable_numa;
         rodata.no_wake_sync = opts.no_wake_sync;
 
         // Normalize CPU busy threshold in the range [0 .. 1024].
         rodata.busy_threshold = opts.cpu_busy_thresh * 1024 / 100;
 
+        // Generate the list of available CPUs sorted by capacity in descendind order.
```
hodgesds:
```suggestion
        // Generate the list of available CPUs sorted by capacity in descending order.
```

hodgesds (approved):
LGTM

PR #2463, opened 2025-07-24
scx_tickless: Immprove per-CPU tasks handling

arighi:
Rather than short-cutting per-CPU tasks in ops.enqueue(), always route them through the shared DSQ, but still prioritize them by allowing to be dispatched even when their only usable CPU is busy.

This approach seems to improve both performance and fairness across multiple benchmarks.


scheds/rust/scx_tickless/src/bpf/main.bpf.c
```diff
@@ -294,6 +290,16 @@ static bool dispatch_cpu(s32 cpu, bool from_dispatch)
 			continue;
 		}
 
+		/*
+		 * If the CPU is busy and the first task in the SHARED_DSQ
+		 * can run somewhere else, stop immediately and move to the
+		 * next CPU.
+		 */
+		if (!scx_bpf_test_and_clear_cpu_idle(cpu) && !is_pcpu_task(p)) {
```
JakeHillion, line 312: I'm a bit stumped on this section, so I'll share some thoughts. Mainly, does this mutation of the idle state make sense? `scx_bpf_test_and_clear_cpu_idle` sets the idle state to false, but if we hit this `if` case we deliberately dispatch no task. I don't know at which point it'd get reset, particularly in a tickless context (would love to know more about the built in idle tracking).

Beyond losing track of the idle state, I have some concerns here. Mainly, I think we need to do this check exactly once before entering the loop. We know that as long as we're in the loop the idle state is unchanged, but the `scx_bpf_test_and_clear_cpu_idle` mutates it. This means that every iteration after the first will treat this CPU as not idle. I think it's plausible that this loop never continues after this point but that is not guaranteed due to the `continue` on move failure. Though, reading the comment, perhaps this makes sense? I'm not sure I understand the motivation of this, and perhaps the code would be clearer by using a named variable rather than the idle tracking state.

arighi: @JakeHillion the idea here is to try to mark the CPU as busy (clearing its idle state) and, if successful, pull the first eligible task from the shared DSQ and dispatch it to that CPU. If the CPU is busy (we fail to clear the bit), we don't want to dispatch a task and we ignore that CPU, since the goal is to avoid CPU contention, so that tasks can keep using their infinite time slice as long as possible.

The CPU will return to the idle state either when the task voluntarily yields (i.e., switches to the idle thread - this is when the idle state is reset) or when a new task is queued to the local DSQ. In the latter case, we detect the contention from a BPF timer, changing the slice from `SCX_SLICE_INF` to a finite `slice_ns`, making the current task preemptible.

The other part of the condition, `is_pcpu_task(p)`, ensures that the first eligible per-CPU task is dispatched immediately on its only usable CPU, even if the CPU is busy. This will trigger the same `SCX_SLICE_INF` -> `slice_ns` transition from the BPF timer, allowing the per-CPU task to preempt the current one and avoid starvation.

Does it make sense? Do you think it'd help to rephrase this concept in the comment?

JakeHillion: I think I'm still unclear on why the first task is special. The logic of "if we're already busy and the highest priority task needs to be dispatched somewhere else don't keep searching" makes sense. Why is that not still the case even if this is the second time we checked?

arighi: @JakeHillion it's not about the first task in the queue, it's about the distinction between per-CPU tasks (`is_pcpu_task(p) == true`) and tasks that can run also on other CPUs. Per-CPU tasks are allowed to be dispatched even if their target CPU is busy, because they have no alternative, avoiding contention isn't possible without risking starvation. So it’s not that the first task is special in general, it’s the first eligible per-CPU task that is treated specially due to its affinity constraint.

JakeHillion (approved):
Nice to see better per-CPU task handling, and I'm 90% sure I understand the behaviour now :)

arighi:
@JakeHillion I rephrase the comment adding more details, now the logic should be a lot more clear (hopefully). :)

hodgesds (approved):
Thanks for adding the comment, this makes sense!

PR #2468, opened 2025-07-24
ci: build kernel directly in ci script

JakeHillion:
Currently we build the kernels into an env variable in the CI workflow and use that where available. This is not ideal as it makes the CI flow harder to reproduce locally & constrains us to a single kernel. Add a helper function to build a named kernel and use in the script (still with a single hardcoded kernel for now). We can add a command line argument if this becomes useful but it isn't needed yet as this is only for the cargo tests that don't require a specific kernel.

Test plan:
- CI


hodgesds (approved):
LGTM

PR #2469, opened 2025-07-24
scxtop: make process columns extensible

yaakov-stein:
The previous way to add a column to the process table required updating the code in a few different places to create the header, the constraint, and the function we'd apply to get the row. It wasn't immediately clear how we'd easily use this approach to add features like toggling on/off a column, having different process table views, or being able to configure your view from a config. With the new Column struct, we can now easily toggle visibility on/off, it's easier to add a new column, and the render method is cleaner.

Side note: Right now, we can access the column positionally. We may want to revisit whether to use some map as well in the future to make toggling easier (it's a bit complex because we need to maintain order, so going to leave that for if/when we need it).


tools/scxtop/src/app.rs
```diff
@@ -2206,28 +2209,106 @@ impl<'a> App<'a> {
...
+                header: "Slice ns",
+                constraint: Constraint::Length(8),
+                visible: true,
+                value_fn: Box::new(|_, data| {
+                    let stats = VecStats::new(&data.event_data_immut("slice_consumed"), None);
+                    stats.avg.to_string()
+                }),
+            },
+            Column {
+                header: "Avg/Max Lat us",
```
yaakov-stein, line 2250: Side fix: added average latency as well

hodgesds (approved):
LGTM

PR #2470, opened 2025-07-24
lib: introduce arena btrees

etsal:
Add an initial implementation of arena-based btrees. This btree data structure will be useful for general indexing tasks, and may be useful as a backend for ATQs. This patch introduces an insert path for the structure, with subsequent patches introducing removal and merging.


lib/selftests/st_btree.bpf.c
```diff
@@ -0,0 +1,376 @@
+#include <scx/common.bpf.h>
+#include <lib/sdt_task.h>
+
+#include <lib/btree.h>
+
+#include "selftest.h"
+
+u64 keys[] = { 51, 43,  37, 3, 301,  46, 383, 990, 776, 729, 871, 96, 189, 213,
```
EricccTaiwan: Add `static const` to both `keys` and `morekeys`?

lib/btree.bpf.c
```diff
@@ -0,0 +1,513 @@
...
+
+	arrzero(&btn->keys[0], BT_LEAFSZ);
+
+	btn->flags = flags;
+	btn->parent = parent;
+
+	return btn;
+}
+
+static bool btnode_isroot(bt_node *btn)
```
hodgesds: nit: why inline one and not the other?

etsal: Omission, the compiler ends up inlining both so it didn't matter but I will explicitly annotate for uniformity.

lib/btree.bpf.c
```diff
@@ -0,0 +1,513 @@
...
+	return i;
+}
+
+static bt_node *bt_find_leaf(btree_t __arg_arena *btree, u64 key)
+{
+	bt_node *btn = btree->root;
+	u64 ind;
+	int i = 0;
+
+	while (!btnode_isleaf(btn) && can_loop && i++ < 10) {
```
hodgesds: It would be interesting to play around with some veristat values of different levels, but this seems fine.

lib/selftests/selftest.h
```diff
@@ -13,3 +13,4 @@ int scx_selftest_atq(void);
 int scx_selftest_bitmap(void);
 int scx_selftest_minheap(void);
 int scx_selftest_topology(void);
+int scx_selftest_btree(void);
```
hodgesds, line 16: nit: keep sorted

hodgesds (approved):

lib/btree.bpf.c
etsal: Great catch, I'll remove  it - it turns out it was necessary initially because of a bug in the code, but not anymore. It's a good general point imo: We can't run buggy BPF programs to debug them because they fail verification, but verifier errors aren't clear enough to help with the bug.

PR #2473, opened 2025-07-25
scx_utils: prevent rustfmt for generated skeleton

etsal:
Speed up compilation for Rust schedulers by disabling rustfmt for the generated skeleton by default. The libbpf-rs command for generating a Rust skeleton calls out into `rustfmt` for the generated bpf_builder.rs, but most of the time the file is directly linked into the scheduler without being read. The formatting step takes a significant amount of time (25% of total compilation time for scx_layered), and is especially expensive for schedulers that use separate compilation units for the schedulers (50% of total compilation time for scx_p2dq).

Pass an invalid rustfmt executable name to the SkeletonBuilder instance to prevent rustfmt from running. There is no option to disable rustfmt, so we just make the operation fail by passing a nonexistent binary name to libbpf-rs for now.


likewhatevs:
ahh so that was it! kinda surprised rust/sccache/ccache etc. don't handle this better.

likewhatevs (approved):

etsal:
I think the main problem is that even a single chnge invalidates the entire skeleton, we should be able to avoid this issue altogether with subskeletons but we need to expand the BPF builder logic accordingly for this.

arighi (approved):
Interesting! Thanks for this, anything that can speed up build is very much welcome.

JakeHillion:
Makes perfect sense! For future reference, this is explicitly handled in the libbpf-cargo source here: https://github.com/libbpf/libbpf-rs/blob/b5a3bd6fe47cba6206858124153a351167f4a137/libbpf-cargo/src/gen/mod.rs#L221-L223

JakeHillion (approved):

hodgesds:
It's annoying that rustfmt is so slow ([issue](https://github.com/rust-lang/rustfmt/issues/6091)), but this is a good fix!

etsal:
> Makes perfect sense! For future reference, this is explicitly handled in the libbpf-cargo source here: https://github.com/libbpf/libbpf-rs/blob/b5a3bd6fe47cba6206858124153a351167f4a137/libbpf-cargo/src/gen/mod.rs#L221-L223

Exactly, this is why we're passing an invalid path :) We cannot pass `None`, and even if we did, libbpf-rs assumes we want the default rustfmt so it helpfully finds it for us. There is no way to explicitly communicate to libbpf-rs we do not want `rustfmt` to run at all. The "right way" would be to add an extra flag to SkeletonBuilder and the internal gen_* function, but this is easier and does not have side effects even if it is a minor abuse of their API.

PR #2481, opened 2025-07-25
scx_lavd: Throttling wake frequency of batch wakers, and more.

multics69:
There are wakers that perform many wakeup operations of other tasks in a very short duration. Our interval-based frequency calculation does not work well because consecutive, repetitive wakeups make the wake frequency meaninglessly high. Instead, we check if there is at least one wakeup within a certain time interval (LAVD_LC_WAKE_INTERVAL_MIN, 1 msec), then calculate the wake frequency between the intervals where at least one wakeup happened. This prevents excessive prioritization of batch wakers.

In addition to this key change, this PR leverages the number of pinned tasks tracked at the CPU level to eagerly include such CPU to the overflow set, and tweaks time slice calculation for accuracy.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
@@ -343,7 +343,7 @@ static void calc_sys_stat(struct sys_stat_ctx *c)
 
 static void calc_sys_time_slice(void)
 {
-	u64 nr_queued, slice;
+	u64 nr_q, slice;
 
 	/*
 	 * Given the updated state, recalculate the time slice for the next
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
```diff
@@ -343,17 +343,21 @@ static void calc_sys_stat(struct sys_stat_ctx *c)
...
 	/*
 	 * Given the updated state, recalculate the time slice for the next
 	 * round. The time slice should be short enough to schedule all
 	 * runnable tasks at least once within a targeted latency using the
 	 * active CPUs.
 	 */
-	nr_queued = sys_stat.nr_queued_task + 1;
-	slice = (LAVD_TARGETED_LATENCY_NS * sys_stat.nr_active) / nr_queued;
-	slice = clamp(slice, slice_min_ns, slice_max_ns);
+	nr_q= sys_stat.nr_queued_task;
```
arighi: nit: space after `nr_q`

PR #2486, opened 2025-07-26
scx_loader: Add performance profiles for scx_cosmos

arighi:
Define all the supported performance profiles for scx_cosmos using the following settings:

 - auto: disable deferred wakeups to reduce power consumption
 - gaming: always use shared DSQ and deadline scheduling for better system responsiveness
 - lowlatency: same as gaming, but also prioritize fast CPUs on systems with hybrid cores and disable sync wakeups for better performance predictability
 - powersave: prioritize slow CPUs on hybrid systems, disable deferred wakeups to reduce power consumption and increase CPU usage polling time to 5s
 - server: increase task's time slice for better throughput and enable address space affinity for better cache locality

### DIFF
--- a/tools/scx_loader/configuration.md
+++ b/tools/scx_loader/configuration.md
@@ -71,11 +71,11 @@ powersave_mode = ["-f", "50", "-p"]
 server_mode = ["-f", "100"]
 
 [scheds.scx_cosmos]
-auto_mode = []
-gaming_mode = []
-lowlatency_mode = []
-powersave_mode = []
-server_mode = []
+auto_mode = ["-d"]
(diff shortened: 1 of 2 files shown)


sirlucjan (approved):
LGTM, thanks!

sirlucjan:
One small nit: https://github.com/sched-ext/scx/blob/main/tools/scx_loader/configuration.md#example-configuration 
Could you add an example here?

arighi:
> One small nit: https://github.com/sched-ext/scx/blob/main/tools/scx_loader/configuration.md#example-configuration Could you add an example here?

Added, thanks.

sirlucjan (approved):
Great

PR #2487, opened 2025-07-26
scxtop: adding uid to the process list.

devnexen:


yaakov-stein (commented):
Is this really helpful/necessary? We have limited space at the moment and I think we'd prefer to keep it for scheduler specific information. I can hear an argument for adding in the owning user as a string (root, etc.) but I don't really see the benefit here of adding a u32. Let me know what you think.

Also, for future PR's, it would be great if you could include a screenshot of what the changes look like - thanks!

devnexen:
>  I can hear an argument for adding in the owning user as a string (root, etc.) but I don't really see the benefit here of adding a u32. Let me know what you think.

I feared the other way around, if user as string, it could possibly take way too much space.

yaakov-stein:
> I feared the other way around, if user as string, it could possibly take way too much space.

Ya, that's a good point. Maybe let's revisit this in a week or two once a few more of the core features are out for the process view.

htejun:

PR #2490, opened 2025-07-27
sync-to-kernel.sh: Create missing files and skip unwanted headers

EricccTaiwan:
Previously, missing target files were skipped with a warning. This patch ensures those files are created by making parent directories and copying them. Also fixes inconsistent indentation in the sync logic.

Additionally, skips syncing `include/arch/` and `.gitignore` files, as they are not intended to be copied into the kernel tree.

See: https://lore.kernel.org/all/aINzTW6Hi40EPSt8@gpd4/

[Screencast from 2025-07-27 21-09-53.webm](https://github.com/user-attachments/assets/459cdc46-c15a-42e1-9bd7-9164dd637a28)

But even after syncing all files, still fails ... :cry: 
```shell
$ # kernel
$ cd tools/testing/selftests/sched_ext
$ make -j$(nproc)
```

Edit: 
Full log : https://gist.github.com/EricccTaiwan/9d3f47762486e5453db627ed5b0d74ae

### DIFF
--- a/scheds/sync-to-kernel.sh
+++ b/scheds/sync-to-kernel.sh
@@ -11,7 +11,12 @@ fi
 rust_scheds=()
 c_scheds=(scx_simple scx_qmap scx_central scx_flatcg)
 
-headers=($(git ls-files include | grep -v include/vmlinux))
+headers=($(
+    git ls-files include |
+    grep -v include/vmlinux |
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
LGTM

PR #2492, opened 2025-07-27
scx_p2dq: Refactor dispatch path for ATQs

hodgesds:
When popping from an ATQ in the dispatch path requeue the task to the local LLC DSQ if the task popped is not the same task as the peeked task. This fixes issues where ATQs could potentially lose tasks and cause stalls.

Before doing pick two load balancing always try to consume from the migration eligible ATQ/DSQ. This gives slightly better throughput at the LLC level as fewer llc_ctx lookups are needed in the hot path.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -1592,12 +1592,12 @@ static __always_inline void p2dq_dispatch_impl(s32 cpu, struct task_struct *prev
 			 * the ATQ. Otherwise there may be priority inversions.
 			 * This probably needs to be done for the DSQs as well.
 			 */
+			if (!(taskc = lookup_task_ctx(p))) {
+				bpf_task_release(p);
+				scx_bpf_error("failed to get task ctx");
+				return;
+			}
 			if (p->pid == peeked_pid) {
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -1609,11 +1609,13 @@ static __always_inline void p2dq_dispatch_impl(s32 cpu, struct task_struct *prev
 				 * The task that was popped was already
 				 * consumed. The next task that was popped
 				 * might have a higher vtime so reenqueue it
-				 * back to the ATQ.
+				 * back to the LLC DSQ.
 				 */
-				scx_atq_insert_vtime(min_atq,
-						     (u64)p->pid,
-						     p->scx.dsq_vtime);
+				scx_bpf_dsq_insert_vtime(p,
```
arighi, line 1614: I'm failing to see how we lose tasks if we re-enqueue them to ATQ. Can you elaborate more on this?

hodgesds: ATQs have fixed sizes, so the current problem is that if the ATQ is full then the task gets lost and ends up stalling because it ends up being on no ATQ/DSQ. The code could check the return code to be safer, but I figured it's just as easy to throw the task back into the LLC DSQ.

PR #2493, opened 2025-07-28
scxtop: refactor columns part 2

yaakov-stein:
This is part 2 of #2469. We now add the Columns struct, which maintains a hashmap of headers to their index so we can update a column in O(1) time, and also gives an easy way to get a visible columns iterator or all columns. This can now fully be used to turn on and off columns easily with the `UpdateColVisibilityAction` and is almost ready to be used for any ratatui table we will use in the future (ColumnFn's reliance on ProcData needs to be generalized).

### DIFF
--- a/tools/scxtop/src/columns.rs
+++ b/tools/scxtop/src/columns.rs
@@ -1,0 +1,222 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+//
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use crate::ProcData;
+use crate::VecStats;
+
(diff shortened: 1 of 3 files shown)


tools/scxtop/src/columns.rs
```diff
@@ -0,0 +1,227 @@
...
+        if let Some(&idx) = self.header_to_index.get(header) {
+            self.columns[idx].visible = visible;
+            true
+        } else {
+            false
+        }
+    }
+
+    /// Return a slice of only the visible columns
+    pub fn visible_columns(&self) -> impl Iterator<Item = &Column<K, D>> {
```
hodgesds, line 51: nice!

PR #2494, opened 2025-07-28
scxtop: add layered support

yaakov-stein:
[Relies on #2493] This one was pretty fun! This adds the optionality, with the `--layered` arg, to see the current `layer_id` for each process. 

How does this work?
* Since bpf maps are global, we are actually allowed to (with a bit of `libbpf-rs` and `bpftool` magic) reuse the file descriptor of another map as long as we have some identifying feature of the map - in this case the name. Once we "get" the map in our scope, we simply need to access the correct location in the data (currently, that's the third element in a `u32` pointer).

When does this break/not work?
* If `bpftool` isn't installed (we'll prevent `scxtop` from starting and show an error)
* If `scx_layered` isn't running (we'll prevent `scxtop` from starting and show an error)
* If `scx_layered`'s `task_ctx` had an element added above `layer_id` and the `LAYER_ID_INDEX` const in `scxtop` was not updated. This will SILENTLY error - i.e. we'll just get the wrong data as it will be pointing at some other piece of data. I don't believe this should be such a problem as those three fields in `scx_layered` should not be updated often, and it's likely the data will be extremely off to the point that users will be aware it is incorrect.

Example while running `scx_layered --run-example`:
Screenshot 2025-07-27 at 8 52 05 PM

Future Work:
* At the moment, the layer_id is what is on `scx_layered`'s `task_ctx`. It would be nice to have the layer name, but that seems like it'll be a bit more complicated without any `scx_layered` changes.
* Make `--layered` configurable to get other data from `task_ctx`
* Remove reliance on `bpftool`
* Add process view features (filtering, dynamic sorting, thread view, etc.)

Shoutout to @hodgesds for the idea to use the layered map.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -28,7 +28,6 @@ use crate::PerfettoTraceManager;
 use crate::ProcData;
 use crate::ProfilingEvent;
 use crate::Search;
-use crate::UpdateColVisibilityAction;
 use crate::VecStats;
 use crate::ViewState;
 use crate::APP;
--- a/tools/scxtop/src/lib.rs
+++ b/tools/scxtop/src/lib.rs
@@ -371,7 +371,6 @@ pub enum Action {
     IncTickRate,
     InputEntry(String),
     IPI(IPIAction),
-    UpdateColVisibility(UpdateColVisibilityAction),
     MangoApp(MangoAppAction),
     NextEvent,
     NextViewState,


hodgesds:
I think this is fine for a initial implementation, on later revisions we can drop the `bpftool` requirement and make the parsing of the `task_ctx` more configurable.

likewhatevs:
@xerothermic btw this is sick

PR #2503, opened 2025-07-28
lib: add Lev-Chase work stealing queues

etsal:
Add Lev-Chase work stealing queues to the library. These queues are useful for load balancing via work stealing for schedulers. The current implementation does not include shrinking the queues when empty because it uses the static bump allocator that has been proven stable.


hodgesds:
Might need to rerun CI because of GH issues, but otherwise LGTM

hodgesds (approved):

PR #2509, opened 2025-07-29
scx_layered: Make dump easier to verify

hodgesds:
Update dump so that some helpers return ints for global verification. Add extra bounds checks to make verification easier.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -3154,19 +3154,21 @@ __hidden void dump_cpumask_word(s32 word, struct cpumask *cpumask)
 	scx_bpf_dump("%08x", v);
 }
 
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3154,19 +3154,21 @@ __hidden void dump_cpumask_word(s32 word, struct cpumask *cpumask)
 	scx_bpf_dump("%08x", v);
 }
 
-static void dump_layer_cpumask(int id)
+int dump_layer_cpumask(int id)
```
JakeHillion, line 3157: Do you want to put `__weak` here to ensure it becomes global at every callsite?

kkdwvd: +1.

kkdwivedi (approved):

PR #2510, opened 2025-07-29
scx_lavd: Improve the ops.enqueue() path.

multics69:
This PR improves the ops.enqueue() path in two ways:
1) Remove redundant pick_idle_cpu() selection when a CPU was already picked at ops.select_cpu().
2) Use the Power of 2 Random Choices technique when a proper CPU is not chosen to choose a less loaded CPU.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
@@ -257,6 +257,27 @@ s32 cpumask_any_dsitribute(struct pick_ctx *ctx)
 }
 
 static
+s32 pick_random_cpu(struct pick_ctx *ctx)
+{
+	/*
+	 * Pick a less loaded CPU using the random of two choices technique.
+	 */
+	s32 cpu0 = cpumask_any_dsitribute(ctx);
+	s32 cpu1 = cpumask_any_dsitribute(ctx);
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_lavd/src/bpf/idle.bpf.c
```diff
@@ -256,6 +256,27 @@ s32 cpumask_any_dsitribute(struct pick_ctx *ctx)
...
+
+	if (cpu0 == cpu1 && cpu0 != -ENOENT)
+		return cpu0;
+
+	cpuc0 = get_cpu_ctx_id(cpu0);
+	cpuc1 = get_cpu_ctx_id(cpu1);
+	if (!cpuc0 || !cpuc1)
+		return ctx->prev_cpu;
+
+	return (cpuc0->cur_sc_util < cpuc1->cur_sc_util) ? cpu0 : cpu1;
```
hodgesds, line 277: Similar to load balancing across LLCs when the CPUs are close to saturation it might be worth considering the queue depths as well.

hodgesds (approved):
This is an interesting idea!

scheds/rust/scx_lavd/src/bpf/idle.bpf.c
multics69: That's a good idea. I will add it in the next round. Thanks for the suggestion!

PR #2513, opened 2025-07-29
scx.service: drop ConditionPathIsDirectory

sirlucjan:
`ConditionPathIsDirectory` was added at a time when sched-ext was not yet upstream (around kernel version 6.7) as a safeguard against running this on LTS version 6.6. Since the latest LTS kernel having sched-ext is 6.12 we can simplify the syntax of the systemd service.

The scx_loader.service also has no such protection and no errors have been reported because of this.

### DIFF
--- a/services/systemd/scx.service
+++ b/services/systemd/scx.service
@@ -1,6 +1,5 @@
 [Unit]
 Description=Start scx_scheduler
-ConditionPathIsDirectory=/sys/kernel/sched_ext
 StartLimitIntervalSec=30
 StartLimitBurst=2
 


services/systemd/scx.service
```diff
@@ -1,6 +1,5 @@
 [Unit]
 Description=Start scx_scheduler
-ConditionPathIsDirectory=/sys/kernel/sched_ext
```
arighi, line 3: What if someone is running a recent kernel without `CONFIG_SCHED_CLASS_EXT=y`?

sirlucjan: The service is not activated by default, it is a conscious decision by the user. In that case - user doesn't use scx anyway so he won't feel the difference.

sirlucjan: Alternatively, the same solution could be considered for the scx_loader.service, so as to standardise the configuration as much as possible.

arighi: Yeah I'm not sure if we should check the dir or not, but in any case it should be consistent with the scx loader service.

arighi (approved):

services/systemd/scx.service
sirlucjan: I doubt that this is a problem, we have not received a single report regarding scx_loader.service that it would cause an error. Thanks for the review!

PR #2514, opened 2025-07-30
scxtop: add process filtering

yaakov-stein:
This adds process filtering by reusing much of the existing filtering mechanisms. The filtering works for both process name and TGID. 

As this doesn't fit cleanly into the idea of an AppState, a new Filter Action was created that turns a filtering struct variable on. The filter UI is modeled after btop's process filtering.

Ex.:

Screenshot 2025-07-30 at 12 40 16 AM

Note: 
* Since we are changing the default 'f' from toggling frequency -> filtering, if you have a saved config, it will still have the binding of 'f' to toggling frequency. You will either need to update this in the config or delete the config and resave it.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -245,7 +245,7 @@ impl<'a> App<'a> {
         // try to infer from the fd
         let hw_pressure = skel.progs.on_hw_pressure_update.as_fd().as_raw_fd() > 0;
 
-        let app = Self {
+        let mut app = Self {
             config,
             localize: true,
             hw_pressure,
@@ -296,6 +296,9 @@ impl<'a> App<'a> {
             max_fps: 1,
         };
 
+        // Set the initial filter state
+        app.filter_events();
+
         Ok(app)
     }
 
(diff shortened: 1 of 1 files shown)


tools/scxtop/src/lib.rs
```diff
@@ -123,15 +123,36 @@ impl std::fmt::Display for ViewState {
...
+            FilterItem::String(_) => None,
+        }
+    }
+}
+
 #[derive(Default, Debug, Clone)]
-pub struct FilteredEventState {
-    pub list: Vec<String>,
+pub struct FilteredState {
+    pub list: Vec<FilterItem>,
```
yaakov-stein, line 149: I was going back and forth on whether this is the best approach or if there is a better way to do this (maybe just have two seperate filtered states, one for events and one for processes, in App?). Nothing was super clean so I settled on this, but I'm very open to hearing other ideas or opinions if anyone has.

JakeHillion (approved):
Implementation LGTM. Glad to see reuse of the fuzzy search work.

hodgesds (commented):
Nice!

PR #2515, opened 2025-07-30
scx_mitosis: Fix -Wunused-function warning

EricccTaiwan:


JakeHillion:
@dschatzberg is it better to clean these up rather than mark them as unused?

dschatzberg:
@JakeHillion Oh, let's delete these instead. Do you want to change your PR For that? I'd love if there was a way to get CI to flag these - what method are you using to see warnings - if my code compiles I don't get any warnings so I'm missing these while developing

JakeHillion:
> @JakeHillion Oh, let's delete these instead. Do you want to change your PR For that? I'd love if there was a way to get CI to flag these - what method are you using to see warnings - if my code compiles I don't get any warnings so I'm missing these while developing

@EricccTaiwan would be great to see an update!

@dschatzberg warnings should be output on a `cargo build`, and it's been that way for quite a while. We don't have `-Werror` on because Clang is so annoying at changing what constitutes an error, but the CI does always use the same reproducible clang version. Perhaps it would be worth turning on a small set like `-Werror=unused-function` in the CI to catch things like this but not be too annoying?

PR #2518, opened 2025-07-30
scxtop: use procfs 0.15.1

yaakov-stein:
This switches to using the more commonly used 0.15.1 version of procfs. Functionality remains the same.


JakeHillion (approved):
Nice and easy!

PR #2520, opened 2025-07-30
scx_lavd: turn on libbpf error logging for verifier failures

etsal:
Turn on libbpf logging for scx_lavd. This enables both setup-related warnings and emitting the verifier log in case of verification failures by default and independently of any scheduler-related diagnostics.


multics69 (approved):
Thanks, @etsal ! LGTM.

PR #2521, opened 2025-07-30
scx_lavd: initial refactoring for separate unit compilation

etsal:
scx_lavd currently separates code into files, but not into compilation units. This has historically been done because of bugs in the libraries used for BPF object linking. However, these bugs have now been removed and other schedulers (scx_layered, scx_rusty/scx_wd40, scx_p2dq) have been using separate unit compilation for quite some time.

Start separating scx_lavd's code into compilation units by splitting out the sys_stat.bpf.c module out of the rest of the code. This only requires moving or adjusting the linkage of any shared symbols along with adding declarations in the shared header as necessary. Subsequent patches will further refactor the rest of the  code modules into separate compilation units.

No functional change intended.

### DIFF
--- a/kernel-versions.json
+++ b/kernel-versions.json
@@ -18,10 +18,10 @@
   "stable/6_12": {
     "repo": "https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git",
     "branch": "linux-6.12.y",
-    "commitHash": "8f5ff9784f3262e6e85c68d86f8b7931827f2983",
-    "lastModified": 1754282193,
-    "narHash": "sha256-KIUrzpGMDSsYjFlWaluZ5LA1MpunLiF+KHEq/4W9y3s=",
-    "kernelVersion": "6.12.41"
+    "commitHash": "880e4ff5d6c8dc6b660f163a0e9b68b898cc6310",
+    "lastModified": 1755403529,
+    "narHash": "sha256-erVU8SHoCxsDk3OuaS5TzWNvL71k1aN7k6vECTGmAgY=",
+    "kernelVersion": "6.12.42"
   },
   "arighi/6_12": {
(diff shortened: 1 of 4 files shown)


etsal:
The change in some type signatures from `void fn(...)` to `int fn(...)` is because we have to make them global and BPF global functions have to return an integer type.

multics69:
Thanks for the changes, @etsal . I like the direction!

I found that after this change, the verifier raises an error -- `BPF program is too large. Processed 1000001 insn` -- on ARM kernel. It is okay with the x86 kernel. I will revisit this PR after the August release and see if the verifier error can be easily addressed.

etsal:
Interesting! Wondering why the verifier is behaving differently between the two systems. I'll try to get an ARM64 system and test it. Just in case it's the verifier for a specific kernel version, which kernel release did you run it under?

multics69:
> Interesting! Wondering why the verifier is behaving differently between the two systems. I'll try to get an ARM64 system and test it. Just in case it's the verifier for a specific kernel version, which kernel release did you run it under?

I tested it on 6.12 kernel. But in my experience, the ARM kernel hits the `too large program` error more frequently than x86.

multics69:
@etsal -- I sent a PR to your working branch to fix the compile error.

etsal:
Merged, also merged with the mainline.

etsal:
Should we merge, since we cut a new release on Thursday? Or do we still have the ARM verification error?

multics69:
> Should we merge, since we cut a new release on Thursday? Or do we still have the ARM verification error?

I will test it on my ARM platform today.

multics69:
Hmm... it still has a verification error on my ARM setup. Let me fix it after merging it to the main.

PR #2522, opened 2025-07-30
scxtop: fix event rendering bug

yaakov-stein:
Initially, the event view was rendering each NUMA as two blocks, one for the title and one for the body. Since the percentage was set at 2 for the title, it was getting cut out when the screen was even somewhat small. This uses a single block, adding a title and then uses .inner() to get the inner area to be used for the CPU data. 

Before:

Screenshot 2025-07-30 at 3 53 56 PM

After:

Screenshot 2025-07-30 at 4 18 06 PM


hodgesds (approved):
LGTM

PR #2525, opened 2025-07-31
scx_p2dq: Deprecate interactive DSQs/ATQs

hodgesds:
Separate DSQs for interactive tasks adds unneeded overhead as tasks are already vtime ordered. Removing them improves overall throughput and latency.


JakeHillion (approved):
LGTM

PR #2527, opened 2025-07-31
ci: add no_capture flag for run_command

JakeHillion:
Some extra abstractions for `run_command` and `run_command_in_vm` were added a while ago. They're always swallowing output but some commands, such as the `format` one, are explicitly expecting to show their output.

Add a `no_capture` argument to `run_command` and `run_command_in_vm` so we can see the output sooner in the CI and get any useful output at all for the format job.

Test plan:
- CI

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -12,12 +12,15 @@ import sys
 from typing import List, Optional
 
 
-async def run_command(cmd: List[str], cwd: Optional[str] = None) -> str:
+async def run_command(
+    cmd: List[str], cwd: Optional[str] = None, no_capture: bool = False
+) -> Optional[str]:
     print(f"Running: {' '.join(cmd)}", flush=True)
+
     proc = await asyncio.create_subprocess_exec(
         *cmd,
-        stdout=asyncio.subprocess.PIPE,
-        stderr=asyncio.subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


.github/include/ci.py
```diff
@@ -106,28 +113,29 @@ async def run_format():
...
+    await run_command(["git", "diff", "--exit-code"], no_capture=True)
     print("✓ Format completed successfully", flush=True)
 
 
 async def run_build():
     """Build all targets."""
     print("Running build...", flush=True)
 
-    await run_command(["cargo", "build", "--all-targets", "--locked"])
+    await run_command(["cargo", "build", "--all-targets", "--locked"], no_capture=True)
```
hodgesds, line 138: Will this hide build failures? Might be worth keeping a buffer of lines and conditionally print if the exit code is non zero.

JakeHillion: It puts it straight out on the terminal now regardless of success/failure, which for these commands that run linearly is more useful (and adds liveness to the data). See https://github.com/sched-ext/scx/actions/runs/16646001138/job/47106751428?pr=2527 for the example.

Previously it would output the stderr on failure but not the stdout, which is why the format output was useless.

hodgesds: Ah, that makes sense!

sirlucjan (approved):

PR #2529, opened 2025-07-31
scx_chaos: also skip select_cpu if next_trait is CHAOS_TRAIT_KPROBE_RANDOM_DELAYS

aoli-al:
scx_chaos only skips `select_cpu` when `next_trait` is `CHAOS_TRAIT_RANDOM_DELAYS`.


yaakov-stein:
Nice catch - thanks!

PR #2532, opened 2025-07-31
scx_chaos: add README.md

JakeHillion:
Add a README for chaos with a logo. The logo may need a little iteration but I'm generally happy with it. The README is probably bad but the best way to achieve correct things on the Internet is to wait for somebody to correct it :)

Will do my best to keep this up to date.


hodgesds (approved):

scheds/rust/scx_chaos/README.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+### Performance Degradation
+```bash
+sudo scx_chaos --degradation-frequency 0.15 --degradation-frac7 64
+```
+
+### Kprobe Delays
+```bash
+sudo scx_chaos --kprobes-for-random-delays schedule,do_exit --kprobe-random-delay-frequency 0.05 --kprobe-random-delay-min-us 50 --kprobe-random-delay-max-us 500
```
yaakov-stein, line 62:
```suggestion
sudo scx_chaos --kprobes-for-random-delays schedule do_exit --kprobe-random-delay-frequency 0.05 --kprobe-random-delay-min-us 50 --kprobe-random-delay-max-us 500
```

This arg uses clap value parser so it parses values until it hits the next `--`. It actually won't work with commas thought, it'll only work if there is some space between the kprobes.

sirlucjan:
@JakeHillion I love this logo! Piece of art!

scheds/rust/scx_chaos/README.md
JakeHillion: > The README is probably bad but the best way to achieve correct things on the Internet is to wait for somebody to correct it :)

Success! I hate documentation... Would you mind throwing the fix up?

PR #2535, opened 2025-08-01
scxtop: add thread view

yaakov-stein:
This adds a thread view that can be accessed from the process view by selecting a process and clicking enter. As of now, most of the header fields are copied over from the process table. The threads are currently sorted by their CPU utilization.

Process View Selection:
Screenshot 2025-08-01 at 11 08 01 AM

Selecting `scxtop` and entering thread view:
Screenshot 2025-08-01 at 11 08 17 AM

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -2375,6 +2375,14 @@ impl<'a> App<'a> {
                 ))
                 .style(self.theme().title_style())
                 .centered(),
+            )
+            .title_top(
+                Line::from(format!(
+                    "sample rate {}",
+                    self.skel.maps.data_data.as_ref().unwrap().sample_rate
+                ))
+                .style(self.theme().text_important_color())
+                .right_aligned(),
             );
 
         let mut threads = proc_data.threads.iter().collect::<Vec<_>>();


tools/scxtop/src/columns.rs
```diff
@@ -58,91 +59,184 @@ impl<K, D> Columns<K, D> {
     }
 }
 
-pub fn get_process_columns() -> Vec<Column<i32, ProcData>> {
-    vec![
+/// Macros to generate individual common columns shared between process and thread views
```
yaakov-stein, line 62: Using macros here lets us define a given column once and ensure that it is consistent across tables.

tools/scxtop/src/proc_data.rs
```diff
@@ -39,12 +40,6 @@ impl ProcData {
             .expect("proc_stats should have processor");
         let cmdline = process.cmdline().unwrap_or_default();
 
-        let mut threads = BTreeMap::new();
-        for thread in process.tasks()?.flatten() {
```
yaakov-stein, line 43: As opposed to storing all of the threads for all processes always, we'll just load the threads lazily whenever a process is selected (see `init_threads` that is called in `on_enter`).

hodgesds (approved):
LGTM

PR #2547, opened 2025-08-02
lib/btree: Handle alloc failure in bt_split()

EricccTaiwan:
Avoid null dereference when allocating new internal or root nodes fails.

### DIFF
--- a/lib/btree.bpf.c
+++ b/lib/btree.bpf.c
@@ -297,6 +297,10 @@ int bt_split(btree_t __arg_arena *btree, bt_node *btn_old)
 
 		btn_parent = btn_old->parent;
 		btn_new = btnode_alloc(btn_parent, btn_old->flags);
+		if (!btn_new) {
+			// scx_buddy_free(buddy, btn_new);
+			return -ENOMEM;
+		}
 
 		if (btn_old->flags & BT_F_LEAF)
 			key = btnode_split_leaf(btn_new, btn_old);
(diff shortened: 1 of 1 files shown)


lib/btree.bpf.c
```diff
@@ -308,6 +312,10 @@ int bt_split(btree_t __arg_arena *btree, bt_node *btn_old)
 			btn_new->flags &= ~BT_F_ROOT;
 
 			btn_root = btnode_alloc(NULL, BT_F_ROOT);
+			if (!btn_root) {
+				//scx_buddy_free(buddy, btn_root);
```
etsal, line 316: This path leaks btn_new.

EricccTaiwan: Would it make sense to add `// scx_buddy_free(buddy, btn_new);` here now?

etsal: Actually, looking at the commented out lines, neither make sense - they are freeing the variable we failed to allocate. Why were they there in the first place?

EricccTaiwan: I followed `bt_create_internal()` -- if that pattern applies, should we free `btn_new` when `btn_root` or `btnode_add_internal()` fails? Want to confirm before fixing.

```c
__weak
u64 bt_create_internal(void)
{
	btree_t __arg_arena *btree;

	btree = scx_static_alloc(sizeof(*btree), 1);
	if (!btree)
		return (u64)NULL;

	btree->root = btnode_alloc(NULL, BT_F_ROOT | BT_F_LEAF);
	if (!btree->root) {
		/* XXX Fix once we use the buddy allocator. */
		//scx_buddy_free(buddy, btree);
		return (u64)NULL;
	}

	return (u64)btree;
}
```

etsal: btree_internal() frees the already allocated btree, not btree->root, which would have failed to allocate. We should alwayas free anything that was _succesfully_ allocated.

EricccTaiwan: OK, I think I misunderstood something here — thanks for the clarification!

PR #2551, opened 2025-08-03
scxtop: Add network TUI view

hodgesds:
Add TUI view that displays network stats.


tools/scxtop/src/network_stats.rs
```diff
@@ -0,0 +1,322 @@
...
+    pub prev_interfaces: BTreeMap<String, InterfaceStats>,
+    pub last_update_time: std::time::Instant,
+}
+
+impl Default for NetworkStatSnapshot {
+    fn default() -> Self {
+        Self {
+            interfaces: BTreeMap::default(),
+            prev_interfaces: BTreeMap::default(),
+            last_update_time: std::time::Instant::now(),
```
yaakov-stein: nit: Doesn't seem that this is used.

tools/scxtop/src/network_stats.rs
```diff
@@ -0,0 +1,322 @@
...
+                    sent_errs: stats.sent_errs,
+                    sent_drop: stats.sent_drop,
+                },
+            );
+        }
+
+        Ok(())
+    }
+
+    pub fn get_delta_recv_bytes(&self, interface: &str) -> u64 {
```
yaakov-stein: It looks like these all have the exact same logic, maybe something like this would reduce the duplication and make it easier to make changes in the future if necessary:

```
impl NetworkStatSnapshot {
    fn get_delta<F>(&self, interface: &str, field_accessor: F) -> u64
    where
        F: Fn(&InterfaceStats) -> u64,
    {
        let current = self
            .interfaces
            .get(interface)
            .map_or(0, &field_accessor);
        let previous = self
            .prev_interfaces
            .get(interface)
            .map_or(0, &field_accessor);
        if current >= previous {
            current - previous
        } else {
            // Handle counter reset/overflow
            current
        }
    }

    pub fn get_delta_recv_bytes(&self, interface: &str) -> u64 {
        self.get_delta(interface, |s| s.recv_bytes)
    }

    pub fn get_delta_sent_bytes(&self, interface: &str) -> u64 {
        self.get_delta(interface, |s| s.sent_bytes)
    }

    pub fn get_delta_recv_packets(&self, interface: &str) -> u64 {
        self.get_delta(interface, |s| s.recv_packets)
    }

    pub fn get_delta_sent_packets(&self, interface: &str) -> u64 {
        self.get_delta(interface, |s| s.sent_packets)
    }

    pub fn get_delta_recv_errs(&self, interface: &str) -> u64 {
        self.get_delta(interface, |s| s.recv_errs)
    }

    pub fn get_delta_sent_errs(&self, interface: &str) -> u64 {
        self.get_delta(interface, |s| s.sent_errs)
    }
}
```

tools/scxtop/src/network_stats.rs
```diff
@@ -0,0 +1,322 @@
...
+    pub fn get_total_recv_errs(&self) -> u64 {
+        self.interfaces.values().map(|stats| stats.recv_errs).sum()
+    }
+
+    pub fn get_total_sent_errs(&self) -> u64 {
+        self.interfaces.values().map(|stats| stats.sent_errs).sum()
+    }
+}
+
+impl From<&DeviceStatus> for InterfaceStats {
```
yaakov-stein, line 236: Good idea! I should use this approach instead of the way I did it for cpu_stats/mem_stats.

tools/scxtop/src/network_stats.rs
```diff
@@ -0,0 +1,322 @@
...
+            .map_or(0, |stats| stats.sent_errs);
+        if current >= previous {
+            current - previous
+        } else {
+            // Handle counter reset/overflow
+            current
+        }
+    }
+
+    pub fn get_total_delta_recv_bytes(&self) -> u64 {
```
yaakov-stein, line 177: Same thing here possibly:

```
impl NetworkStatSnapshot {
    fn get_total_delta<F>(&self, field_accessor: F) -> u64
    where
        F: Fn(&InterfaceStats) -> u64,
    {
        self.interfaces
            .keys()
            .map(|iface| self.get_delta(iface, &field_accessor))
            .sum()
    }

    pub fn get_total_delta_recv_bytes(&self) -> u64 {
        self.get_total_delta(|s| s.recv_bytes)
    }

    pub fn get_total_delta_sent_bytes(&self) -> u64 {
        self.get_total_delta(|s| s.sent_bytes)
    }

    pub fn get_total_delta_recv_packets(&self) -> u64 {
        self.get_total_delta(|s| s.recv_packets)
    }

    pub fn get_total_delta_sent_packets(&self) -> u64 {
        self.get_total_delta(|s| s.sent_packets)
    }

    pub fn get_total_delta_recv_errs(&self) -> u64 {
        self.get_total_delta(|s| s.recv_errs)
    }

    pub fn get_total_delta_sent_errs(&self) -> u64 {
        self.get_total_delta(|s| s.sent_errs)
    }
}
```

tools/scxtop/src/app.rs
```diff
@@ -582,6 +586,11 @@ impl<'a> App<'a> {
             }
         }
 
+        // Update network stats
+        if let Err(e) = self.network_stats.update() {
+            eprintln!("Failed to update network stats: {}", e);
```
yaakov-stein: Because we've taken over the screen with crossterm, I don't think this will be seen or accessible IIUC. Should we maybe use the `render_error_msg`? Or possibly fail as something has gone wrong?

tools/scxtop/src/app.rs
```diff
@@ -1788,9 +1797,98 @@ impl<'a> App<'a> {
...
+            Layout::vertical([Constraint::Ratio(2, 3), Constraint::Ratio(1, 3)]).areas(left);
 
         self.render_event(frame, right)?;
-        self.render_table(frame, left)
+        self.render_table(frame, left_top)?;
+        self.render_network_summary(frame, left_bottom)
+    }
+
+    /// Renders a simplified network summary for the default view.
+    fn render_network_summary(&mut self, frame: &mut Frame, area: Rect) -> Result<()> {
```
yaakov-stein, line 2086: If you want to use the Columns struct and add a `get_network_summary_columns`, that would keep some of the logic there (and a bit easier to work with going forward). After a brief look through the `render_network_summary` and `render_network`, if you use the Columns struct, it seems like it should be pretty simple to combine these into one method with a bool passed in to render the full network view (because you can just call visible columns etc. and then add the non-summary info if `render_full` == true). But I can also hear an argument to keep them seperate if we think the logic will continue to diverge.

tools/scxtop/src/network_stats.rs
```diff
@@ -0,0 +1,322 @@
...
+    pub sent_bytes: u64,
+    pub sent_packets: u64,
+    pub sent_errs: u64,
+    pub sent_drop: u64,
+}
+
+impl NetworkStatSnapshot {
+    pub fn update(&mut self) -> Result<()> {
+        // Save current interfaces as previous
+        std::mem::swap(&mut self.interfaces, &mut self.prev_interfaces);
```
yaakov-stein, line 51: nit: Just wondering, is there any benefit of a swap here over a take given that we're going to be clearing self.interfaces? I'd imagine a swap is a bit more work than just a take.

tools/scxtop/src/network_stats.rs
```diff
@@ -0,0 +1,322 @@
...
+            .get(interface)
+            .map_or(0, |stats| stats.recv_bytes);
+        let previous = self
+            .prev_interfaces
+            .get(interface)
+            .map_or(0, |stats| stats.recv_bytes);
+        if current >= previous {
+            current - previous
+        } else {
+            // Handle counter reset/overflow
```
yaakov-stein: If we get here, something probably went wrong, no? Should we error out?

tools/scxtop/src/network_stats.rs
hodgesds: Yeah, that's much cleaner

hodgesds:
Updated with improvements to the detailed view:
image

JakeHillion (approved):
LGTM. Might have to switch to using scxtop as my default monitoring tool :)

PR #2552, opened 2025-08-03
scxtop: use more idiomatic style

yaakov-stein:
As a follow-up to #2551, this uses the more idiomatic `From` trait.


JakeHillion (approved):
It's beautiful

PR #2553, opened 2025-08-03
scxtop: fix various bugs + add tests

yaakov-stein:
Previously, avg/max data was showing up as 50/100 whenever there wasn't any data. Ex. (see the threads that don't have NUMA/LLC):
Screenshot 2025-08-01 at 11 08 17 AM

This was due to the `EventData::event_data_immut` returning a vec of `[0, max_data_size]` whenever an entry didn't exist. Instead, we should just return an empty vec. Now, the data is correct:

Screenshot 2025-08-03 at 6 21 15 PM

To hopefully prevent these issues from coming up in the future, test suites were added to `EventData` and `VecStats`. These tests already caught an possible overflow bug in `VecStats` where we were adding u64's into a single u64.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -764,7 +764,7 @@ impl<'a> App<'a> {
             ProfilingEvent::CpuUtil(_) => llc_data.num_cpus,
             _ => 1,
         };
-        let data = llc_data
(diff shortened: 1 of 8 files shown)


hodgesds (approved):
tests good

PR #2554, opened 2025-08-04
scx_lavd: Reduce the competition window to prevent stalls.

multics69:
Reduce the competition window that allows later enqueued tasks to be scheduled earlier by 1/4. This prevents unexpected stalls in some cases and fixes the issue 2491 [1] without causing performance regression.

[1] https://github.com/sched-ext/scx/issues/2491


Franck666666:
Woe, good, BUt one Question remains 

Why no AOCC  Compatibility ?????? or u plan it ?

multics69:
> Why no AOCC Compatibility ?????? or u plan it scx?

I don't think anyone tested scx schedulers with AOCC compilers. I will test it and see if it can be trivially fixed or not.

Franck666666:
Very much thx if u do really, , hes really so fast, Got  with the Nvidia Driver compiled with the AOCC nearly double Framrates, in Games like Quake and Doom, the first ones ... Yesterday seen after edited the dkms.conf with CC= and also LD=


HUGE DIFFERENCE

Franck666666:
But in Facts its nkown since MD releades their Compiler, many do secretly use  for best Reasons, also the Kernel is compiabla sry i am french, with it. No Problrms

But Wayland lost Focus, also Hyperland

PR #2555, opened 2025-08-04
Use scx_bpf_remote_curr() instead of scx_bpf_cpu_rq()

cloehle:
This is the conversion series for using the safer scx_bpf_remote_curr(),
which now also requires the schedulers to call bpf_task_release().

This is dependent on:
https://lore.kernel.org/lkml/20250804112743.711816-1-christian.loehle@arm.com/
so obviously do not merge yet, but feel free to review.


etsal:
Even when the kernel patch gets merged, we need compat macros to keep the schedulers runnable on existing kernels. Can you wrap this around a macro? There's examples in the scheds/include/scx headers.

cloehle:
Closing this in favor of:
https://github.com/sched-ext/scx/pull/2734/

PR #2556, opened 2025-08-04
rustland_core: remove duplicate builder

JakeHillion:
scx_rustland_core's build.rs runs the builder from scx_utils. It uses the output to fill in a `bindings` module, but that module is private and unused. If it were to be used, we should re-export `scx_utils`'s bindings instead.

Clean up the build code and the unused module.

Test plan:
- `cargo build && sudo target/debug/scx_rustland --stats 10`
- CI


JakeHillion:
> I don't know how to test if the scx_rustland_core crate (or rather any potential external scheduler using that crate) is still happy with this change, but we'll see.
> 
> For now, LGTM.

Should be all good, the missing `pub` means it was never exported, and you can see it's not listed on docs.rs: https://docs.rs/scx_rustland_core/latest/scx_rustland_core/index.html

You would have to work around the build system a lot to access this, and I don't think that's an API surface we have to worry about maintaining.

PR #2557, opened 2025-08-04
scxtop: Add memory stats

hodgesds:
Add a pane to the default TUI view for memory related stats. The memory stats are displayed in a tristate where they can be configured to be visible on the default TUI view or a detailed view.

image
image


tools/scxtop/src/lib.rs
```diff
@@ -123,6 +125,37 @@ impl std::fmt::Display for ViewState {
     }
 }
 
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum ComponentViewState {
```
yaakov-stein, line 129: Cool idea!

tools/scxtop/src/mem_stats.rs
```diff
@@ -13,15 +15,163 @@ pub struct MemStatSnapshot {
...
+    pub delta_pgfault: u64,
+    pub delta_pgmajfault: u64,
 }
 
 impl MemStatSnapshot {
     pub fn update(&mut self) -> Result<()> {
         let meminfo = Meminfo::new()?;
-        *self = (&meminfo).into();
+
+        // Save previous values for delta calculation
```
yaakov-stein, line 80: This probably can all be added into impl From and we can keep update the same as before, no?

tools/scxtop/src/mem_stats.rs
```diff
@@ -45,14 +195,64 @@ impl From<&Meminfo> for MemStatSnapshot {
         Self {
             total_kb: meminfo.mem_total,
             free_kb: meminfo.mem_free,
-            available_kb: meminfo
-                .mem_available
-                .expect("available memory should be present"),
+            available_kb: meminfo.mem_available.unwrap_or(0),
```
yaakov-stein: I'm wondering what's better here - the unwrap might lead to data being confusing or incorrect. Should we leave it as expects when the data should exist? Or maybe let's just make the UI not display that line of info when the data is missing?

tools/scxtop/src/app.rs
```diff
@@ -1790,7 +1805,155 @@ impl<'a> App<'a> {
...
+            }
+            ComponentViewState::Default => {
+                // Memory summary is shown, split the left area
+                let [table_area, memory_area] =
+                    Layout::vertical([Constraint::Percentage(80), Constraint::Percentage(20)])
+                        .areas(left);
+                self.render_table(frame, table_area)?;
+                self.render_memory_summary(frame, memory_area)?;
+            }
+            ComponentViewState::Detail => {
```
yaakov-stein: nit: this can be added above to avoid duplication.
```suggestion
            ComponentViewState::Hidden | ComponentViewState::Detail => {
```

tools/scxtop/src/app.rs
```diff
@@ -1790,7 +1805,155 @@ impl<'a> App<'a> {
...
+                // This shouldn't happen in default view, but handle it gracefully
```
yaakov-stein: Is handling this gracefully the right move here? Should we either render_error_msg or panic if its *never* supposed to happen?

tools/scxtop/src/app.rs
```diff
@@ -1790,7 +1805,155 @@ impl<'a> App<'a> {
...
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Renders a memory summary in the default view.
+    fn render_memory_summary(&mut self, frame: &mut Frame, area: Rect) -> Result<()> {
+        // Only render memory summary if the memory view state is Default
+        if self.memory_view_state != ComponentViewState::Default {
```
yaakov-stein, line 1843: I'd imagine here we should atleasy render_error_msg (if not panic) as without anything rendering the screen area will just be blank, which can be quite confusing for a user.

tools/scxtop/src/app.rs
```diff
@@ -1790,7 +1805,155 @@ impl<'a> App<'a> {
...
+        // Calculate memory usage percentages
+        let available_percent = (mem_stats.available_kb as f64 / mem_stats.total_kb as f64) * 100.0;
+        let used_percent = 100.0 - available_percent;
+        let swap_used_percent = 100.0 - mem_stats.swap_ratio() * 100.0;
+        let slab_percent = (mem_stats.slab_kb as f64 / mem_stats.total_kb as f64) * 100.0;
+        let reclaimable_percent =
+            (mem_stats.sreclaimable_kb as f64 / mem_stats.slab_kb as f64) * 100.0;
+
+        // Create a compact multi-line display
+        let text = vec![
```
yaakov-stein: Should this just be a table with headers and one row and all contstraints as Fill? That might look better for different screen sizes. Also probably would be simpler code and easier to change in the future.

tools/scxtop/src/app.rs
```diff
@@ -2446,12 +2631,612 @@ impl<'a> App<'a> {
...
+            .border_type(BorderType::Rounded)
+            .style(self.theme().border_style());
+
+        let table = Table::new(rows, constraints).header(header).block(block);
+
+        frame.render_widget(table, left);
+
+        // Create memory usage gauges and additional stats for the right side
+        let [right_top, right_middle, right_bottom] = Layout::vertical([
+            Constraint::Percentage(8),
```
yaakov-stein: I think we should probably use min here (not sure if 2 is the right number), as otherwise Percentage(8) will probably be cut off on small screens.
```suggestion
            Constraint::Min(2),
```

tools/scxtop/src/app.rs
```diff
@@ -3219,7 +4004,21 @@ impl<'a> App<'a> {
...
             Action::SetState(state) => {
-                if *state == self.state {
+                if *state == AppState::Memory {
+                    // Handle memory view tristate cycling
+                    self.memory_view_state = self.memory_view_state.next();
+                    match self.memory_view_state {
+                        ComponentViewState::Detail => {
+                            self.set_state(AppState::Memory);
+                        }
+                        ComponentViewState::Hidden | ComponentViewState::Default => {
```
yaakov-stein, line 3652: If I am in AppState::Scheduler and I click on 'M' and then 'M' again, is the expected behavior to return to AppState::Scheduler (as I believe it would behave now)? Or do we want it to cycle through the tristate?

tools/scxtop/src/app.rs
hodgesds: I think it's fine to bail in this case since it's an application logic error

tools/scxtop/src/mem_stats.rs
yaakov-stein: Following up here. Can't we just switch this to be `let snapshot = meminfo.into()?;`

tools/scxtop/src/mem_stats.rs
yaakov-stein: Should we maybe just leave them as an option so we can surface an empty datapoint as opposed to unwrap_or(0)? Or we think leaving it as 0 won't be confusing (ie it might show 0 but it's really some other number, we just failed parsing)?

tools/scxtop/src/app.rs
```diff
@@ -307,7 +320,48 @@ impl<'a> App<'a> {
         };
 
         // Set the initial filter state
-        app.filter_events();
+        app.filtering = true;
+        app.event_input_buffer.clear();
+        let filtered_events_list = match app.state {
```
yaakov-stein: Why do we split this out of `filter_events()` now?

yaakov-stein (approved):
Had a few comments (confused why we got rid of `filter_events` and wondering about some error handling mainly), but besides for that, the rest makes sense to me. Thanks for adding!

PR #2558, opened 2025-08-04
scx_rustland_core: Cleanup task storage on init_task failure

EricccTaiwan:
If L2 or L3 cpumask allocation fails in rustland_init_task(), the allocated task storage and cpumask must be properly cleaned up to avoid memory leaks and invalid state.

This patch ensures that on failure, l2_cpumask is released if set, and the task_ctx is deleted using bpf_task_storage_delete().


arighi (commented):
I think this is not needed, the cpumasks should be automatically released via the lifetime tracking in BPF, so we don't need to explicitly free them.

EricccTaiwan:
> I think this is not needed, the cpumasks should be automatically released via the lifetime tracking in BPF, so we don't need to explicitly free them.

Ok! Thanks for the clarification!

PR #2560, opened 2025-08-04
ci: move veristat job from ubuntu runners to python script

JakeHillion:
Currently we run veristat when we run a full set of stress tests for a
scheduler. There are a few flaws with this approach:
- Any failures of veristat are completely ignored and still allow the PR
  to merge.
- We spin up an entire virtme-ng VM to run a single veristat invocation,
  even though veristat is very quick and the VMs are quite expensive.
- We only run veristat against the kernel(s) we have chosen for full
  stress tests, which are few because the tests are expensive.

This moves the veristat jobs from the Ubuntu runners to the ci.py Python
script. In doing that, it adds the plumbing for a job internal matrix
and many efficiency improvements. This allows us to run veristat for all
relevant kernels for all schedulers on every PR with a single quick job.

Adds a new `ci.disable_veristat` option to `package.metadata.scx` in
Cargo.toml for each scheduler, and automatically locates all Rust crates
under `scheds/`. This makes this the first opt out bit of verification
in the CI. Also places the heavy Ubuntu runner integration tests behind
this CI job, as verification failures must be dealt with and this will
save some of that limited capacity.

Test plan:
- CI

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -8,10 +8,11 @@ import os
 import random
 import shlex
 import shutil
+import string
 import subprocess
 import sys
 import tempfile
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 
 async def run_command(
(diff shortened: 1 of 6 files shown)


.github/include/ci.py
```diff
@@ -84,11 +87,98 @@ async def run_command_in_vm(
...
+    # maximum length. write all commands to a shell script and run that instead,
+    # allowing us to bypass this restriction.
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".sh") as file:
+        # VM gets a different PATH to this program so fix the path of the binary.
+        # Other args left to the caller.
+        arg0 = shutil.which(command[0])
+        cmd_quoted = " ".join(shlex.quote(arg) for arg in [arg0] + command[1:])
 
-    return await run_command(cmd, no_capture=no_capture)
+        file.write(f"exec {cmd_quoted}\n")
```
hodgesds: 🦺

hodgesds (approved):

.github/include/ci.py
```diff
@@ -185,7 +275,309 @@ async def run_tests():
...
+            return result
+
+    matrix = await generate_scheduler_kernel_matrix()
+    matrix = [x for x in matrix if not x["disable_veristat"]]
+
+    print(f"Testing {len(matrix)} scheduler-kernel combinations:")
+    for item in matrix:
+        print(f"  - {item['scheduler']} on {item['kernel']}")
+
+    # Group by kernel to reuse VMs
```
hodgesds, line 343: nice!

JakeHillion:
Note for reviewers: this is a stacked PR, only the final commit is relevant. See https://github.com/sched-ext/scx/pull/2559 for the first commit.

PR #2565, opened 2025-08-05
scxtop: add util tests

yaakov-stein:
Adds some tests to the util methods to 1) ensure they are behaving as expected, and 2) ensure changes don't break them.


hodgesds (approved):
tests good

PR #2567, opened 2025-08-06
scxtop: Add additional test coverage

hodgesds:
Add test coverage for common and rather stable code.


JakeHillion (approved):
Nice! I'm sure these aren't perfect, but better to have tests we need to update in future than no tests at all.

PR #2568, opened 2025-08-06
chaos: generate random numbers with less bias

JakeHillion:
chaos currently uses a mod to move the uniform random number from the kfunc into the requested range. This has a potential bias which increases as you move towards larger numbers.

Any solution to this requires a loop which means it's impossible to completely remove bias within BPF's termination guarantees, but we can get a lot closer. This approach uses Lemire's Algorithm 5 which avoids a mod in all cases by using 128 bit arithmetic. Unfortunately, BPF doesn't support 128 bit arithmetic, so we roll it by hand.

Test plan:
- CI
- `cargo build --release && sudo target/release/scx_chaos --random-delay-frequency 0.5 --random-delay-min-us 1000 --random-delay-max-us 10000`

[0] https://arxiv.org/pdf/1805.10941v2

### DIFF
--- a/scheds/rust/scx_chaos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_chaos/src/bpf/main.bpf.c
@@ -88,6 +88,46 @@ struct chaos_task_ctx *lookup_create_chaos_task_ctx(struct task_struct *p)
 	return bpf_task_storage_get(&chaos_task_ctxs, p, NULL, BPF_LOCAL_STORAGE_GET_F_CREATE);
 }
 
+static __always_inline u64 chaos_get_prandom_u64_limit(u64 s)
+{
+	// Implementation of Lemire's algorithm 5 without 128-bit arithmetic.
+	// See https://arxiv.org/pdf/1805.10941v2 for details.
+	// Uses a bounded loop given this is BPF, but given the loop should
+	// rarely be entered this is fine.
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_chaos/src/bpf/main.bpf.c
```diff
@@ -88,6 +88,51 @@ struct chaos_task_ctx *lookup_create_chaos_task_ctx(struct task_struct *p)
 	return bpf_task_storage_get(&chaos_task_ctxs, p, NULL, BPF_LOCAL_STORAGE_GET_F_CREATE);
 }
 
+static __always_inline u64 chaos_get_prandom_u64()
+{
+	return ((u64)bpf_get_prandom_u32() << 32) | bpf_get_prandom_u32();
+}
+
+static __always_inline u64 chaos_get_prandom_u64_limit(u64 s)
```
hodgesds, line 91: might be worth throwing into `common.bpf.h`

JakeHillion: Good reminder. I'll put `get_prandom_u64` into common for now as I think that's unobjectionable, will bed in this implementation for a while for `_u64_limit` - the fact it's still not _technically_ removing the bias makes me a bit loathe to share it just yet.

Issue #2548, opened 2025-08-03
scx_lavd: locks up relatively easily with PREEMPT_RT


Arcterus:
I don't have logs for this (at least at the moment) since I switched to `scx_p2dq` and continued mucking around with my laptop setup, but my system locked up loading a game, accessing the GNOME notification bar (while loading something in the background, although I don't quite remember what), and IIRC when launching Firefox.  I don't believe I had any particular issues prior to enabling `PREEMPT_RT`, but admittedly I hadn't used `scx_lavd` particularly long.  `scx_bpfland` and `scx_p2dq` appear to work fine (and EEVDF, unsurprisingly).

Note that when I say locked up, I mean I had to force the computer off because it refused to accept any input.

multics69:
@Arcterus -- Thank you for reporting the problem. Could you try with the current main? https://github.com/sched-ext/scx/pull/2554 might be related to this problem.

multics69:
And, what is your processor?

multics69:
@Arcterus -- If possible, could you test LAVD with https://github.com/sched-ext/scx/pull/2570?

Arcterus:
Sorry for the delay, I've been trying to fix issues with compiling kernels locally in `nixpkgs`. I'll test the change tomorrow.

Arcterus:
> And, what is your processor?

It's a Ryzen 9 5900 HS (laptop CPU from a few years ago).

multics69:
Thanks @Arcterus ! Now you can test the main since the PR is merged to the main.

Arcterus:
I tested main just now, and it locked up pretty much immediately after starting a game. It's somewhat repeatable, since I tried a second time and it locked up shortly after starting again (although it got further along the second time). The game is Wuthering Waves running through GE-Proton10-10, in case it helps having a specific test case.

Also, when I say shortly after starting, I mean the screen was still black before loading the game company logo.

multics69:
@Arcterus -- For Wuthering Waves, it does not seem like a scheduler problem (including LAVD). I ran the game with EEVDF and any other schedulers, but the system suddenly reboots around the first screen. Since the problem exists even in the default EEVDF scheduler, I guess the root cause would be somewhere in Wine/Proton layers. Did you try the game with EEVDF on your setting?

Arcterus:
[...]

PR #2570, opened 2025-08-07
scx_lavd: Hardening to avoid stalls when extremely over-loaded.

multics69:
Harden the deadline calculation and slice boost when the system is extremely overloaded to avoid stalls. It includes the following tuning and changes that won't affect the normal use cases:

1) Give 2x weight boost for a kernel or pinned task. This helps a pinned kernel
   thread be scheduled before the stall.

2) Reduce LAVD_TARGETED_LATENCY_NS from 20 msec to 10 msec. Under extremely
   heavy load, the scheduler shortens the time slices more aggressively. In the
   normal use up to around 3x overload (one running + 2 waiting tasks with
   5 msec time slice), this should not affect.

3) Reduce LAVD_SLICE_BOOST_MAX from 1 sec to 0.5 sec, which is a more realistic
   maximum value.

4) Previously, the slice boost is triggered when slice_max_ns <=
   sys_stat.slice. This is not ideal since it creates a dependency on the time
   slice calculation. Let's change this sys_stat.nr_queued_task <=
   sys_stat.nr_active. The changed condition is still aggressive enough since
   it will try slice boosting until 2x overload (1 for running + 1 for
   waiting).

The changes tested on a 16-CPU AMD Ryzen 9 PRO 6950H with PREEMPT_DYNAMIC and PREEMPT_RT kernels. The scheduler runs well without stalling while running the following all at the same time:

- Linux kenrel compilation with make -j64
- schbench -L -m 4 -M auto -t 64 -n 0
- perf bench sched messaging -t -g 16 -l 3000000
- stress-ng -c 2048
- nice -n 19 stress-ng -c 2048
- Running WebGL Aquarium on Chromium with 30000 fishes (https://webglsamples.org/aquarium/aquarium.html)

This might fix the issue [1], but it has not been tested due to the lack of hardware.

[1] https://github.com/sched-ext/scx/issues/2548


scheds/rust/scx_lavd/src/bpf/util.bpf.c
```diff
@@ -225,7 +225,7 @@ static bool have_scheduled(struct task_ctx *taskc)
 
 static bool can_boost_slice(void)
 {
-	return slice_max_ns <= sys_stat.slice;
+	return sys_stat.nr_queued_task <= sys_stat.nr_active;
```
hodgesds, line 228: This seems much safer, nice!

PR #2574, opened 2025-08-07
scx_p2dq: Fix segfault

hodgesds:
Fix a segfault in some versions of libbpf/clang by removing some tracing code.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -786,7 +786,6 @@ static s32 pick_idle_cpu(struct task_struct *p, task_ctx *taskc,
 		    scx_bpf_test_and_clear_cpu_idle(pref_cpu)) {
 			*is_idle = true;
 			cpu = pref_cpu;
-			trace("PREF idle %s->%d", p->comm, pref_cpu);
 			goto found_cpu;
 		}
 	}


yaakov-stein:
Was this breaking even when tracing was off?

multics69:
Hmm... The problem still exists. This is the gdb backtrace with a debug build.

image

JakeHillion:
> Hmm... The problem still exists. This is the gdb backtrace with a debug build.

Thank you for the backtrace! I got the same in my build where this fails so this is really helpful. Could you take a look at the value of `relo->map_idx` and check that it's ridiculous? On my box I saw values in the millions.

Unfortunately I couldn't work out where this is getting set. The only fix I had was updating libbpf (the C bit) to the latest master, but unfortunately a bisect doesn't work in the libbpf GitHub repo due to its structure.

multics69:
> > Hmm... The problem still exists. This is the gdb backtrace with a debug build.
> 
> Thank you for the backtrace! I got the same in my build where this fails so this is really helpful. Could you take a look at the value of `relo->map_idx` and check that it's ridiculous? On my box I saw values in the millions.

This is what I got:

```
(gdb) p relo->map_idx
$1 = 1717986878
(gdb) p/x relo->map_idx 
$2 = 0x6666663e
```

hodgesds:
@etsal pointed out this [libbpf patch]( https://lore.kernel.org/all/20250718001009.610955-1-andrii@kernel.org/) which I think is the source of the problem. Unfortunately, it looks like the `libbpf-sys` will need to update the libbpf version. I'm not sure if there is a way to work around the issue maybe changing the layout to where it relocation works, but it seems not very deterministic.

JakeHillion:
> @etsal pointed out this [libbpf patch](https://lore.kernel.org/all/20250718001009.610955-1-andrii@kernel.org/) which I think is the source of the problem. Unfortunately, it looks like the `libbpf-sys` will need to update the libbpf version. I'm not sure if there is a way to work around the issue maybe changing the layout to where it relocation works, but it seems not very deterministic.

This aligns with my testing, updating the vendored libbpf on my machine to the latest `master` was the only way to reliably get rid of this issue. Unfortunately I have no idea how we can publish a crate with that constraint.

hodgesds:
This should be resolved with #2592

PR #2578, opened 2025-08-08
scxtop: add process appstate

yaakov-stein:
This adds a detailed process view and sets the stage to enable additional columns on the wider screen and column toggling. Also removes unnecessary `AppState::Memory` checks.

Screenshot 2025-08-08 at 1 11 55 PM

Testing:
* Played around with it. Went Default -> Process -> Scheduler and Default -> Process -> Thread -> Process -> Default, behavior was as expected.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -365,7 +365,6 @@ impl<'a> App<'a> {
             || self.state == AppState::Default
             || self.state == AppState::Llc
             || self.state == AppState::Node
-            || self.state == AppState::Memory
             || self.state == AppState::Process
         {
             self.filtered_state.lock().unwrap().reset();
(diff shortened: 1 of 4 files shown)


tools/scxtop/src/app.rs
```diff
@@ -3019,7 +3059,7 @@ impl<'a> App<'a> {
                 self.filter_events();
                 self.handle_action(&Action::SetState(self.prev_state.clone()))?;
             }
-            AppState::Default | AppState::Llc | AppState::Node | AppState::Memory => {
```
yaakov-stein, line 3022: Are these Memory checks here for a reason (as far as I understand, these seem specific to process level data)?

hodgesds: Nah, this is just bad codegen.

tools/scxtop/src/app.rs
```diff
@@ -2632,8 +2656,13 @@ impl<'a> App<'a> {
             )
             .title_top(
                 Line::from(format!(
-                    "sample rate {}",
-                    self.skel.maps.data_data.as_ref().unwrap().sample_rate
+                    "sample rate {}{}",
```
hodgesds, line 2659: Maybe we should have a helper function to generate the sample/tick rate.

hodgesds (approved):
LGTM

PR #2580, opened 2025-08-08
scxtop: make help more helpful

yaakov-stein:
This updates help, moves things around into clearer sections, and removes a few duplications/inaccuracies (more on that below). Here is what it looks like now:

Screenshot 2025-08-08 at 1 45 00 PM

Inaccuracies:
* Before memory view was listed twice
* Page down and Page Up still had profiling in the help page (they also work for process view)
* Up and Down were missing


hodgesds (approved):
helpful

PR #2582, opened 2025-08-08
lib/btree: add free bnode cache

etsal:
Add a linked list of freed btree nodes to each tree. This lets us implement free without switching off of the static allocator.


multics69 (approved):
This looks correct to me.

PR #2583, opened 2025-08-08
scxtop: Remove dependency on bpftool

hodgesds:
Use a libbpf map iterator when searching for the layered map. This removes the dependency on bpftool.


etsal (approved):

xerothermic (approved):
Nice!!

PR #2586, opened 2025-08-11
Fix s390 support

iii-i:
* Update the vmlinux.h generator script to fix numerous generic and s390-related issues, most importantly, missing symbols.
* Regenerate vmlinux.h.
* Slightly adjust tests to work with the updated vmlinux.h.
* Port two qspinlock changes from BPF selftests.
* Update perfetto_protos to a version that supports s390.


JakeHillion:
Would you mind splitting this into 4 PRs? The unittests macro seems & qspinlock backport seem easy to stamp independently, and the perfetto_protos dependency might should be fine too. The vmlinux changes I am a little more confused about, I don't see why we're reverting from 6.16 vmlinuxes to 6.14. If we get the other stuff out of the way I think we can have a more useful discussion here - thanks!

iii-i:
Will do.

It was not supposed to be a revert, it's the same kernel commit: `9b30400ff652`.
But apparently the generator script has attributed it to a wrong kernel release.
I will try to find what caused this and fix it.

iii-i:
So, v6.14 comes from:

```
$ git rev-parse HEAD
9b30400ff6527c4601fefa2e9cb0b311c6f902d0

$ git describe --tags --match="v*"
v6.14-13427-g9b30400ff652
```

This is somewhat fair, because when the commit in question was authored, v6.16 was not released yet.
And the commit corresponding to v6.15 is not on the master branch.

Now the question is, what to do with this.
If we want the kernel version number in the file name to represent the release that will contain the respective kernel commit, then I don't think there is an easy way to automate this.
What about adding an extra argument to the script instead?

JakeHillion:
> So, v6.14 comes from:
> 
> ```
> $ git rev-parse HEAD
> 9b30400ff6527c4601fefa2e9cb0b311c6f902d0
> 
> $ git describe --tags --match="v*"
> v6.14-13427-g9b30400ff652
> ```
> 
> This is somewhat fair, because when the commit in question was authored, v6.16 was not released yet. And the commit corresponding to v6.15 is not on the master branch.
> 
> Now the question is, what to do with this. If we want the kernel version number in the file name to represent the release that will contain the respective kernel commit, then I don't think there is an easy way to automate this. What about adding an extra argument to the script instead?

Thanks for the explanation - totally fine as is! Could you rebase this on main now the others have merged & update the description? Otherwise the vmlinux generation changes look fine to me.

iii-i:
Thanks for taking a look, will do.
In order to avoid confusion, I started generation of vmlinux.h files for the v6.16 release itself, which is properly resolved by the script.
Once it's finished, I will rebase, retest and push the result.

iii-i:
After rebasing I had to add an extra commit that replaces `regs->ip` with `PT_REGS_IP(regs)`. Please let me know if you'd rather see this as a separate PR.

JakeHillion:
Looks good to me.

@htejun could you describe the cpumask dance you mentioned in the office hours? I can't find a reference to it.

etsal:
> Looks good to me.
> 
> @htejun could you describe the cpumask dance you mentioned in the office hours? I can't find a reference to it.

It's just a manual modification of the struct cpumask struct in vmlinux.h so that the array of longs has length 128.

htejun:
> @htejun could you describe the cpumask dance you mentioned in the office hours? I can't find a reference to it.

So, in each arch's `vmlinux.h`, there's `struct cpumask` definition:
```
struct cpumask {
        unsigned long bits[COUNT];
};
```
The `COUNT` should be edited to 128.

iii-i:
I've added a script to perform this fixup.

After a rebase, I see a build fail due to `set_rlimit_infinity()` unused result. I've included the fix here for completeness, and will also send it as a separate PR.

iii-i:
Rebased, now that the `set_rlimit_infinity()` PR is merged.

JakeHillion:
Hmm, it looks like the CI failure is legit (caused by an extra job added after you put up your PR). scxtop doesn't build on aarch64 now, though it's fine on x86_64. Logs at: https://github.com/sched-ext/scx/actions/runs/17037301148/job/48292652217.  Not sure exactly what's causing it yet.

scripts/gen_vmlinux_h.sh
```diff
@@ -85,8 +85,26 @@ generate_vmlinux_for_arch() {
...
+        cp .config .config.orig
+    else
+        make ARCH=${ARCH} CROSS_COMPILE=${CROSS_COMPILE} KCFLAGS=-Wno-error defconfig &>> ${LOG}
+        echo CONFIG_DEBUG_INFO_REDUCED=n >>.config
+        echo CONFIG_DEBUG_INFO_DWARF4=y >>.config
+        echo CONFIG_BPF_SYSCALL=y >>.config
+        echo CONFIG_DEBUG_INFO_BTF=y >>.config
+        echo CONFIG_BPF_JIT=y >>.config
+        echo CONFIG_SCHED_CLASS_EXT=y >>.config
+        echo CONFIG_CGROUP_SCHED=y >>.config
```
frelon, line 101: I think we are missing `CONFIG_BPF_EVENTS=y` here. Without it I get the following error building on [arm64](https://build.opensuse.org/package/live_build_log/home:flonnegren:branches:Base:System/scx/openSUSE_Factory/aarch64):

```
...
[  316s]       3: src/bpf/main.bpf.c:1146:64: error: incomplete definition of type 'struct bpf_perf_event_data'
[  316s]           1146 |         event->event.perf_sample.instruction_pointer = PT_REGS_IP(&ctx->regs);
...
```

iii-i: Thanks for taking a look. I think I need to add `CONFIG_FTRACE=y`, since it's a prepreq and `CONFIG_BPF_EVENTS` itself is not user-selectable.

frelon: Yes, `CONFIG_FTRACE=y` is even better :+1:

iii-i:
* Added `CONFIG_FTRACE=y`, rebased.

JakeHillion:
Built this on aarch64 and everything looks okay now. Thanks again!

PR #2587, opened 2025-08-11
scx_cosmos: Add null check for scx_bpf_cpu_rq()

cloehle:
scx_bpf_cpu_rq() may return NULL so check for it.
Additionally also add a NULL check for ->curr.


JakeHillion (approved):
This will unblock https://github.com/sched-ext/scx/pull/2584 to update the sched_ext/for-next kernel, and this does return NULL sometimes so should be checked. As mentioned in https://lore.kernel.org/sched-ext/aJIWHk_4IuOxyh5d@gpd4/T/#m8e663f569b713377557976be11c4ce20ff88be07 though, I think we should implement these changes in a way that doesn't break existing scheduler binaries for a couple of kernel versions.

JakeHillion:
@cloehle does this change still make sense with the differences in the kernel patches? Happy to merge if so.

arighi:
Hm.. this would still break the logic in the scheduler, because it would consider CPUs as busy (in the NULL case) while they're not.

What we should do instead is introduce a compat helper to check if the new API to get the curr task is available and rely on that.

And if we get NULL I'd rather error the scheduler or drop this logic completely.

cloehle:
>What we should do instead is introduce a compat helper to check if the new API to get the curr task is available and rely on that.

I'll do that with the rest of the changes of the new scx_bpf_cpu_rq_locked().
I've updated this PR to be the best what we can do until then.

PR #2591, opened 2025-08-11
scx_bpf_unittests: rename TEST to SCX_BPF_UNITTEST

iii-i:
The scx_bpf_unittests TEST macro conflicts with the dw_dma_regs::TEST struct member. Change its name to SCX_BPF_UNITTEST.


JakeHillion (approved):
I can't believe an enum variant named TEST landed. Thanks for this!

PR #2592, opened 2025-08-11
scx_utils: Update libbpf-rs version

hodgesds:
Update libbpf-rs version for scx_utils and all uses of scx_utils to use the latest libbpf-rs version. This version contains this [patch]( https://lore.kernel.org/all/20250718001009.610955-1-andrii@kernel.org/) which seems to remediate random segfault issues when using arenas.


JakeHillion (approved):
Nice!

PR #2595, opened 2025-08-12
scxtop: Add power view

hodgesds:
Add a power view similar to `turbostat` that provides information on C-State data.

image

### DIFF
--- a/tools/scxtop/src/power_data.rs
+++ b/tools/scxtop/src/power_data.rs
@@ -6,8 +6,533 @@
 use anyhow::{anyhow, Result};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
-use std::fs;
+use std::fs::{self, File};
+use std::io::{Read, Seek, SeekFrom};
 use std::path::Path;
+use std::sync::Mutex;
+
+// MSR-based RAPL (Running Average Power Limit) constants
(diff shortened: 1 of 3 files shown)


tools/scxtop/src/power_data.rs
```diff
@@ -0,0 +1,886 @@
...
+    sysfs_cpuidle_path: String,
+}
+
+impl PowerDataCollector {
+    pub fn new() -> Result<Self> {
+        let cpu_count = Self::detect_cpu_count()?;
+
+        Ok(Self {
+            cpu_count,
+            sysfs_power_path: "/sys/class/power_supply".to_string(),
```
yaakov-stein, line 1039: I think most of these (possibly all) should be available in the already-imported[ `sysinfo`](https://docs.rs/sysinfo/latest/sysinfo/index.html) crate if you want to use that instead of manually parsing.

hodgesds: Hmm there doesn't seem to be many [power](https://docs.rs/sysinfo/latest/sysinfo/index.html?search=power) relate interfaces.

hodgesds:
Updated layout now has CPU temp working properly and with RAPL it's able to read RAM power usage as well.
image

JakeHillion (approved):
Looks useful to me!

PR #2599, opened 2025-08-13
scx_utils: include sched version in /sys

JakeHillion:
When a sched_ext scheduler is running the name field in the `struct_ops` field is presented at `/sys/kernel/sched_ext/root/ops`. This is useful to see what scheduler is running, but the name is quite limited.

Add a new `ops_version` method and add the version information of the scheduler in this `name` field. This means that we get the full git build information in `/sys`, really useful when trying to work out exactly what scheduler binary was running on a machine at a given time.

Test plan:
- CI
- `cargo build --release`

```
$ sudo target/release/scx_lavd &
$ cat /sys/kernel/sched_ext/root/ops
lavd_1.0.14_gce7e69c0_dirty_x86_64_unknown_linux_gnu
```


JakeHillion:
> LGTM, as long as it fits within the limit should be fine.

Yeah, that's all checked. Even without fitting though, it's only appending to the string, so you won't lose any data over what was there before even if your previous scheduler name was 127 characters long.

PR #2603, opened 2025-08-14
scx_utils, scx_lavd: Handle an error in set_rlimit_infinity() gracefully.

multics69:
Setting rlimit for MEMLOCK to infinity would fail with -ENOPERM.
However, it does not necessarily mean catastrophic failure. Especially
when rlimits are already set to a large enough value, it is okay to
ignore the error.
    
So let's ignore the error and print the current rlimits. If the current
rlimits are insufficient, the program will fail anyway later in the
initialization, allocating BPF objects.

### DIFF
--- a/rust/scx_utils/Cargo.toml
+++ b/rust/scx_utils/Cargo.toml
@@ -28,12 +28,13 @@ sscanf = "0.4"
 tar = "0.4"
 walkdir = "2.5"
 version-compare = "0.1"
-libc = "0.2.137"
 zbus = { version = "5.3.1", optional = true }
 const_format = "0.2.34"
 num = "0.4.3"
 tracing = "0.1"
 tracing-subscriber = "0.3"
+libc = "0.2.175"
+nix = { version = "0.30.1", features = ["resource"] }
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


rust/scx_utils/Cargo.toml
```diff
@@ -28,12 +28,13 @@ sscanf = "0.4"
 tar = "0.4"
 walkdir = "2.5"
 version-compare = "0.1"
-libc = "0.2.137"
 zbus = { version = "5.3.1", optional = true }
 const_format = "0.2.34"
 num = "0.4.3"
 tracing = "0.1"
 tracing-subscriber = "0.3"
+rlimit = "0.10.2"
```
hodgesds: Why a new crate? If not using the libc there's also the [nix crate](https://tidelabs.github.io/tidext/nix/sys/resource/fn.setrlimit.html) which is already a dependency elsewhere.

multics69: Ah, that's right. I will change to use the nix crate.

multics69: Fixed.

hodgesds (approved):
LGTM

PR #2604, opened 2025-08-14
scxtop: Fix total power calculation in TUI

hodgesds:
Don't try to use power estimates as they are inaccurate only report the total package power.

image


JakeHillion (approved):
nice.

PR #2605, opened 2025-08-14
lib/btree: add remove and merge operations

etsal:
Add merging and removal operations to the btree data structures.

### DIFF
--- a/lib/btree.bpf.c
+++ b/lib/btree.bpf.c
@@ -9,6 +9,8 @@
 #include <lib/sdt_task.h>
 #include <lib/btree.h>
 
+#define BTREE_MAX_DEPTH (20)
+
 /*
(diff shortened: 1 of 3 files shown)


lib/btree.bpf.c
```diff
@@ -101,8 +104,8 @@ u64 bt_create_internal(void)
 	return (u64)btree;
 }
 
-static
-u64 btn_node_index(bt_node *btn, u64 key)
+__weak
```
hodgesds, line 109: why no longer static if `btn_node_index_by_val` is static?

etsal: IIRC It helped with verification at some point because inlining the loop caused one too many jumpst, but I will revert it and try again.

etsal: Unfortunately it's even worse than that, inlining causes test failures.

PR #2608, opened 2025-08-15
scheds: fix unused set_rlimit_infinity() result

iii-i:
Currently the build fails with a number of warnings like this:

    warning: unused `Result` that must be used
       --> scheds/rust/scx_tickless/src/main.rs:117:9
        |
    117 |         set_rlimit_infinity();
        |         ^^^^^^^^^^^^^^^^^^^^^
        |
        = note: this `Result` may be an `Err` variant, which should be handled

Fix by expanding the scx_lavd approach to other schedulers.


iii-i:
I noticed this on x86 with `rustc 1.82.0 (f6e511eec 2024-10-15)`. One weird thing is that sometimes recompilation without any changes makes the build error go away. But I figured it was worth fixing anyway.

iii-i:
The build is apparently failing due to:

```
[1/3] meson compile -C build stress_tests_scx_layered
/home/runner/work/_temp/d42af2cb-3dab-45c0-8f82-03273eedc822.sh: line 4: meson: command not found
```

But I don't think this can be caused by this change.

EricccTaiwan:
> The change looks good to me. But I didn't have the error on my setting (x86, rustc 1.86.0) even without this change. Is there anything special in you setting?

As mentioned in PR #2609 , I also observed this warning on my device.
```shell
 ~  rustc --version                                                                       
rustc 1.88.0 (6b00bc388 2025-06-23)
 ~  uname -r                                                           
6.14.0-28-generic
 ~  lscpu                                     
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             46 bits physical, 48 bits virtual
  Byte Order:                Little Endian
CPU(s):                      24
  On-line CPU(s) list:       0-23
Vendor ID:                   GenuineIntel
  Model name:                12th Gen Intel(R) Core(TM) i9-12900K
```

multics69:
It is weird. Anyway, the change looks better than the precise one. So let's merge it to the main.

Issue #2474, opened 2025-07-25
scx_wd40: Build failure on ARM due to missing `mempolicy`


EricccTaiwan:
Hi, I recently encountered a build failure on ARM (`aarch64`) after rebasing to the latest commit.

```shell
$  uname -r
6.14.0-15-generic
$ lscpu 
Architecture:             aarch64
CPU(s):                   224
Model name:               ThunderX2-99xx
NUMA node(s):             2
```

Interestingly, the failing code involving `mempolicy` was introduced around 5 months ago, and I’m quite certain I was able to build successfully on ARM before the rebase.

The build now fails with errors like:
```c
error: no member named 'mempolicy' in 'struct task_struct'; did you mean 'policy'?
```

<details>
<summary>Full build log</summary>

```c
$ meson compile -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /usr/bin/ninja -C /home/underdog/eric/scx/build
ninja: Entering directory `/home/underdog/eric/scx/build'
[3/3] Generating rust_all with a custom command (wrapped by meson to set env)
FAILED: Cargo.toml.__PHONY__ 
/usr/bin/meson --internal exe --unpickle /home/underdog/eric/scx/build/meson-private/meson_exe_cargo_3733d421a119686c95c70b55b1a7acfef1777bdd.dat
while executing ['/home/underdog/.cargo/bin/cargo', 'build', '--manifest-path=../Cargo.toml', '--target-dir=.', '--quiet']
--- stdout ---

--- stderr ---
error: failed to run custom build command for `scx_wd40 v1.0.14 (/home/underdog/eric/scx/scheds/rust/scx_wd40)`

Caused by:
  process didn't exit successfully: `/home/underdog/eric/scx/build/./debug/build/scx_wd40-3194eabaf8eb44a9/build-script-build` (exit status: 101)
  --- stdout
[...]

PR #2610, opened 2025-08-15
scx_wd40: No-op task_set_preferred_mempolicy_dom_mask() on ARM64

EricccTaiwan:
Make task_set_preferred_mempolicy_dom_mask() a no-op on ARM64 to fix build warnings on non-NUMA systems, with possible impact on NUMA-aware locality for multi-node ARM64.

Suggested-by: Nudelsalad

close: #2474


hodgesds (approved):
LGTM

PR #2612, opened 2025-08-15
ci: add basic build check for aarch64

JakeHillion:
I recently landed a PR which broke ARM because of some different Rust types in the different architectures. Add a very basic ARM CI job that simply builds the schedulers (no kernel/verification yet) to catch things like this in future.

Test plan:
- Ran from a commit before the ARM fix, it failed: https://github.com/sched-ext/scx/actions/runs/16996832515/job/48189455830
- This PR's CI

### DIFF
--- a/.github/include/flake.nix
+++ b/.github/include/flake.nix
@@ -30,7 +30,7 @@
   };
 
   outputs = { self, nixpkgs, flake-utils, fenix, nix-develop-gha, libbpf-src, veristat-src, ... }:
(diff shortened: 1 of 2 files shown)


hodgesds (approved):
more coverage

JakeHillion:
Hmm, this passed a couple of times when merging, but now appears to be failing on main/in future merge queues. Completely impossible to debug. Will give it another go then backout if it's still causing problems.

PR #2614, opened 2025-08-15
scxtop: Cleanup scheduler TUI view

hodgesds:
When rendering the scheduler TUI view for barcharts or sparkline add the option to control rendering the scheduler name. Only use the option to render the scheduler on the scheduler stats to keep the scheduler TUI view readable.

`main` hard to view with long scheduler names:

PR:


yaakov-stein (approved):
LGTM

PR #2615, opened 2025-08-15
scxtop: Add last waker to thread view

hodgesds:
In the TUI thread view add the last waker to the thread view. This is useful for understanding waker/wakee relationships.

Example output:


likewhatevs (approved):
super nice

PR #2616, opened 2025-08-16
scxtop: Add colors to charts

hodgesds:
Add colors to charts based on values using the gradient helpers. This makes it easier to visually spot changes in values.

### DIFF
--- a/tools/scxtop/src/app.rs
+++ b/tools/scxtop/src/app.rs
@@ -777,8 +777,8 @@ impl<'a> App<'a> {
         Ok(())
     }
 
-    /// Generates a CPU bar chart.
-    fn cpu_bar(&self, cpu: usize, event: &str) -> Bar {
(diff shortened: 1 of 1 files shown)


tools/scxtop/src/app.rs
```diff
@@ -1391,8 +1470,29 @@ impl<'a> App<'a> {
 
     /// Generates a LLC bar chart.
     fn event_bar(&self, id: usize, value: u64, avg: u64, max: u64, min: u64) -> Bar {
+        // Calculate gradient color based on relative position between min and max
+        let gradient_color = if max > min {
```
yaakov-stein: Should we condense this repetitive code into a method that returns the gradient color?

yaakov-stein (approved):
Looks nice and helpful!

tools/scxtop/src/app.rs
hodgesds: Sounds good!

PR #2626, opened 2025-08-17
scxtop: Fix clippy warnings

hodgesds:
Fix clippy warnings for lifetime elision by giving explict lifetime annotations to helper methods.


hodgesds:
> Thanks for fixing (noticed these started recently, was it due to a clippy update? It doesn't seem like there was a code change that should've caused it)

Yeah, I upgraded my clippy and started seeing them as well.

PR #2627, opened 2025-08-18
p2dq: Refactor llc_runs handling

hodgesds:
Refactor the handling of llc_runs on the task_ctx so that tasks become migration eligible when the number of llc_runs reaches zero. If a task reaches zero then llc runs a refreshed. This allows for tasks to be more sticky to a llc.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -261,6 +261,11 @@ static u64 task_dsq_slice_ns(struct task_struct *p, int dsq_index)
 	return task_slice_ns(p, dsq_time_slice(dsq_index));
 }
 
+static void task_refresh_llc_runs(task_ctx *taskc)
+{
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -261,6 +261,11 @@ static u64 task_dsq_slice_ns(struct task_struct *p, int dsq_index)
 	return task_slice_ns(p, dsq_time_slice(dsq_index));
 }
 
+static void task_set_llc_runs(task_ctx *taskc)
```
JakeHillion: nit: I'd expect a setter to take `nr_llc_runs` as an argument, perhaps `task_refresh_llc_runs` is a better name?

hodgesds: yeah, `min_llc_runs_pick2` is a global that gets updated on idle so felt kind of weird to pass it in.

PR #2630, opened 2025-08-18
scx_lavd: continue separate unit compilation refactoring

etsal:
Further adjust lavd's function definitions to treat each source file as a separate compilation unit. For this patch series, separate out the introspection, locking, power, and preemption modules.

### DIFF
--- a/rust/scx_utils/src/cpumask.rs
+++ b/rust/scx_utils/src/cpumask.rs
@@ -269,7 +269,7 @@ impl Cpumask {
     ///     info!("cpu {} was set", cpu);
     /// }
     /// ```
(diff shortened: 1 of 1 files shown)


EricccTaiwan:
> I tested it on x86 and it worked well. There would be problem on ARM but let's fix it after merge on the main.

I tested on ARM. It can successfully build, but when loading `scx_lavd`, I hit some errors.
See: https://gist.github.com/EricccTaiwan/89911ac1d0b07596e10bb1922d62ea31

Edit: I also encountered the error on x86 ...
https://gist.github.com/EricccTaiwan/a24bd7a0af82e73289b6eb99d5506df4

PR #2634, opened 2025-08-18
rust/scx_lib_selftest: add BPF Streams support

etsal:
Add support for the BPF streams error reporting interfaces when testing the arena library.

### DIFF
--- a/rust/scx_lib_selftests/src/main.rs
+++ b/rust/scx_lib_selftests/src/main.rs
@@ -10,8 +10,12 @@ use std::mem::MaybeUninit;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
+use std::os::fd::AsRawFd;
+
+use std::ffi::c_void;
 
 use std::ffi::c_ulong;
+use std::os::fd::AsFd;
 use std::sync::Arc;
 
 use scx_utils::init_libbpf_logging;
@@ -21,6 +25,8 @@ use scx_utils::Topology;
 use scx_utils::NR_CPU_IDS;
 
 use simplelog::{ColorChoice, Config as SimplelogConfig, TermLogger, TerminalMode};
+
+use libbpf_sys;
 
 use libbpf_rs::skel::OpenSkel;
 use libbpf_rs::skel::SkelBuilder;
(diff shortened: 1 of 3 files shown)


rust/scx_lib_selftests/src/main.rs
```diff
@@ -10,8 +10,12 @@ use std::mem::MaybeUninit;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
+use std::os::fd::AsRawFd;
+
+use std::ffi::c_void;
```
hodgesds, line 15: nit: keep all `std::` imports together?

hodgesds (approved):
nice

rust/scx_lib_selftests/src/main.rs
```diff
@@ -152,6 +158,35 @@ fn setup_topology(skel: &mut BpfSkel<'_>) -> Result<()> {
...
+        if ret < 0 {
+            println!("Error {} reading stream {}", ret, stream_id);
+            break;
+        }
+
+        if ret == 0 {
+            break;
+        }
+
+        println!("{}", String::from_utf8_lossy(&buf[..ret as usize]));
```
kkdwvd: We should just print here, not insert a newline, if necessary it can be inserted after dumping all contents, otherwise the output may look odd.

rust/scx_lib_selftests/src/main.rs
```diff
@@ -152,6 +158,35 @@ fn setup_topology(skel: &mut BpfSkel<'_>) -> Result<()> {
     Ok(())
 }
 
+fn print_stream(skel: &mut BpfSkel<'_>, stream_id: u32) -> () {
+    let mut buf = vec![0u8; 4096];
+
+    println!("===BEGIN STREAM {}===", stream_id);
```
kkdwvd: Should we print the string name corresponding to stream_id?

kkdwivedi (changes requested):

rust/scx_lib_selftests/src/main.rs
etsal: Sounds good, skipped it because there's only two right now. I'll hardcode the 1 -> stdout, 2 -> stderr correspondence  - same as in bpftool.

rust/scx_lib_selftests/src/main.rs
etsal: Will try, I passed this by cargo fmt which normally rearranges the imports.

rust/scx_lib_selftests/src/main.rs
etsal: Ack, slipped in.

rust/scx_lib_selftests/src/main.rs
```diff
@@ -181,10 +185,10 @@ fn print_stream(skel: &mut BpfSkel<'_>, stream_id: u32) -> () {
             break;
         }
 
-        println!("{}", String::from_utf8_lossy(&buf[..ret as usize]));
+        print!("{}", String::from_utf8_lossy(&buf[..ret as usize]));
     }
 
-    println!("====END STREAM {}====", stream_id);
+    println!("\n====END STREAM {}====", stream_id);
```
kkdwvd: s/stream_id/name/g

PR #2635, opened 2025-08-19
include/lib: port page fragment allocator from the kernel

etsal:
Port the page fragment allocator from the kernel. This is mainly to be used for debugging as an alternative to existing allocators.


scheds/include/lib/pagefrag_alloc.h
```diff
@@ -0,0 +1,72 @@
+/* SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause) */
+/* Copyright (c) 2024 Meta Platforms, Inc. and affiliates. */
```
hodgesds:
```suggestion
/* Copyright (c) 2025 Meta Platforms, Inc. and affiliates. */
```

scheds/include/lib/pagefrag_alloc.h
```diff
@@ -0,0 +1,72 @@
...
+#endif
+
+static void __arena * __arena page_frag_cur_page[NR_CPUS];
+static int __arena page_frag_cur_offset[NR_CPUS];
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARENA);
+	__uint(map_flags, BPF_F_MMAPABLE);
+#if defined(__TARGET_ARCH_arm64) || defined(__aarch64__)
+	__uint(max_entries, 1 << 16); /* number of pages */
```
hodgesds: It might be nice to have some of these as consts in a header as the arch specific values might get reused elsewhere.

etsal: Agreed, the main problem is that we need the exact same definition for the arena map that we can include everywhere. I will spin it out into a separate include file we can add.

PR #2636, opened 2025-08-19
scx_lavd: Fix a verification error on ARM.

multics69:
On ARM, the verifier hits the "BPF program is too large" error. Let's split a large loop body into two to lower the verification burden and work around the problem before the verifier becomes smarter.


adamdmoss:
AFAICT, same problem on x64 now since the recent lavd refactor.

adamdmoss:
... and I've verified that this PR also fixes x64.  Cheers.

EricccTaiwan:
This fix works on both ARM and x86. Thanks.

PR #2637, opened 2025-08-19
scx_p2dq: Refactor saturation handling

hodgesds:
Rafactor various components of saturation handling in order to improve both throughput and locality.

1) Update the can_migrate function to always respect the number of LLC
   runs even during saturation to give better locality.
2) During idle CPU selection allow tasks that are migration eligible to
   find idle CPUs outside the local LLC.
3) Always refresh tasks runs during intialization.
4) Update the default number of runs during saturation.


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -803,6 +803,14 @@ static s32 pick_idle_cpu(struct task_struct *p, task_ctx *taskc,
 		goto found_cpu;
 	}
 
+	if (saturated && taskc->llc_runs < 1) {
```
JakeHillion:
```suggestion
	if (saturated && taskc->llc_runs == 0) {
```

?

PR #2641, opened 2025-08-19
scx_p2dq: Add task slice tracking

hodgesds:
Add a mode to scale individual tasks by slice usage rather than relying on existing slice values. This gives more stable performance for tasks that would jump between short durations and long durations.


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -1273,7 +1275,15 @@ void BPF_STRUCT_OPS(p2dq_stopping, struct task_struct *p, bool runnable)
 		if (p->scx.weight < 100 && taskc->dsq_index > 1)
 			taskc->dsq_index = 1;
 
-		taskc->slice_ns = task_dsq_slice_ns(p, taskc->dsq_index);
+		if (p2dq_config.task_slice) {
+			if (used >= ((9 * last_dsq_slice_ns) / 10)) {
```
JakeHillion: `7/8` or `15/16` to optimise the division?

JakeHillion (approved):

PR #2643, opened 2025-08-19
High frequency dynamic layer switching using hints

kkdwivedi:
Patches to enable layer switching on wake up and slice boundaries based on the passed in task hints. Please see commit logs for details. Still requires an additional change to correctly handle vruntime adjustment across layer DSQs.


kkdwvd:
Updated to avoid membership change in stopping(), and consolidate slice boundary checks in dispatch() with some explanation on why that's okay. Instead of performing the layer switch in dispatch() directly, it is deferred to enqueue(), so that the accounting we do in stopping() still takes into account the previous layer of the task and charges to it correctly.

Notably slice refresh is skipped in case we notice a discrepancy between hint and the task's layer in dispatch(), so we trigger re-enqueue by not refreshing the slice when `p->scx.flags & SCX_TASK_QUEUED` is true.

kkdwvd:
The vruntime delta scaling will be done in a follow up, so the left the comment about it as-is.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -3157,6 +3181,27 @@ fn verify_layer_specs(specs: &[LayerSpec]) -> Result<()> {
                             bail!("Spec {:?} has too long a process name prefix", spec.name);
                         }
                     }
+                    LayerMatch::HintEquals(hint) => {
+                        if *hint > 1024 {
+                            bail!(
```
hodgesds, line 3187: Should this bail or just warn? I guess maybe it's good to fail fast, but ignoring would seem fine as well.

kkdwvd: The lookup table (hint_to_layer_id_map) is sized to 1025 entries, so not warning would just mean we silently ignore invalid settings, and do not pick up the hint at runtime for the layer.

hodgesds: Ahh that makes sense, thanks!

PR #2648, opened 2025-08-20
scxtop: Add async symbolization

hodgesds:
Add separate thread from symbolization to remove symbolization from the TUI render thread. This fixes stalls in TUI rendering from symbolization.

### DIFF
--- a/tools/scxtop/src/symbol_data.rs
+++ b/tools/scxtop/src/symbol_data.rs
@@ -8,6 +8,191 @@ use blazesym::symbolize::{Input, Sym, Symbolizer};
 use blazesym::Pid;
 use std::collections::{BTreeMap, HashMap};
 use std::path::PathBuf;
+use std::sync::mpsc;
+use std::thread;
+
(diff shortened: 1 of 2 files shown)


tools/scxtop/src/symbol_data.rs
```diff
@@ -8,6 +8,191 @@ use blazesym::symbolize::{Input, Sym, Symbolizer};
...
+    }
+}
+
+impl AsyncSymbolizer {
+    pub fn new() -> Self {
+        let (req_tx, req_rx) = mpsc::channel::<SymbolizationRequest>();
+        let (resp_tx, resp_rx) = mpsc::channel::<SymbolizationResponse>();
+
+        // Spawn the worker thread
+        thread::spawn(move || {
```
JakeHillion, line 56: Might be worth switching to `tokio::spawn` and `tokio::mpsc` for uniformity, you can still do a receive in a sync context with `try_recv` which I think is appropriate for this use case.

PR #2650, opened 2025-08-20
lib/sdt_alloc: Add loop bound check with `can_loop`

EricccTaiwan:


kkdwvd:
Why is this necessary?

etsal:
I agree with @kkdwivedi, this is not necessary in this case. If it were, none of the code in `lib' would be working. The verifier can reason about the index variable just fine.

Is this just for the sake of uniformity? I'd rather not have non-functional changes just for that.

EricccTaiwan:
> Is this just for the sake of uniformity? I'd rather not have non-functional changes just for that.

Yes, I added it because I saw other for-loops also use `can_loop`, so I was trying to keep it consistent. 
But you’re right, this is a non-functional change, so I’ll drop it. Thanks.

PR #2653, opened 2025-08-21
scx_lavd: Check Per-CPU dsq on select cpu path

daidavid:
When Per-CPU DSQ option is enabled, the check on the number of queued tasks on Per-LLC DSQ will always return 0. We should check the CPU specific DSQ instead.


multics69:
I guess this bug would be a source of worse latency when using per-cpu DSQ.

PR #2659, opened 2025-08-22
scx_lavd: --per-cpu-dsq shares DSQ between SMT siblings

daidavid:
When SMT is enabled, map each SMT sibling to share the same CPU DSQ. This should help with preserving better vtime ordering correctness while still maintaining l1/l2 cache locality.


multics69 (approved):
LGTM.

PR #2662, opened 2025-08-22
lib: introduce red-black tree

etsal:
Add a red-black tree data structure with nodes that are embeddable into other structs. This tree type can be used to back ATQs and the fact it is embeddable allows us to add it to task_ctx structs directly.

### DIFF
--- a/lib/rbtree.bpf.c
+++ b/lib/rbtree.bpf.c
@@ -1,0 +1,803 @@
+/*
+ * SPDX-License-Identifier: GPL-2.0
+ * Copyright (c) 2025 Meta Platforms, Inc. and affiliates.
+ * Copyright (c) 2025 Emil Tsalapatis <etsal@meta.com>
+ */
+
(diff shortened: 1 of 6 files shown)


etsal:
TODO: The tree insert/remove operations should directly take a node argument so that we defer the management of node instances to the caller, isnce we need to make it embeddable.

multics69 (approved):
LGTM. Picking an item with the smallest key would be needed to pick the smallest vtime task. 
Also, ideally, it would be great to have a feature to pick a task after checking the cpumask. maybe, it would be a part of ATQ with a user-provided filter function?

etsal:
> LGTM. Picking an item with the smallest key would be needed to pick the smallest vtime task. Also, ideally, it would be great to have a feature to pick a task after checking the cpumask. maybe, it would be a part of ATQ with a user-provided filter function?

Yes, the idea is that we add locking and integration with cpumasks/other scheduler state at the ATQ level.

PR #2663, opened 2025-08-22
scxtop: Refactor on_tick method

hodgesds:
Refactor on_tick method to only update relevant collectors based on the current TUI view.


JakeHillion (approved):
Efficient. We'll have to be careful to keep this in sync.

PR #2664, opened 2025-08-22
lib/btree: fix btnode cached allocation/frees

etsal:
Fix the btnode allocation/free code to properly check the result of the `cmpxchg' operation.


etsal:
@JakeHillion This removes all the may_goto timeouts being spewed by the selftests.

PR #2669, opened 2025-08-23
scxtop: Support TUI as non root user

hodgesds:
Make scxtop TUI mode work as a non root user. Some views are not available as root, but most functionality is preserved.

Non root some functionality such as perf events can still be used depending on system configuration:

Root has extra columns by default (DSQ):

### DIFF
--- a/tools/scxtop/src/main.rs
+++ b/tools/scxtop/src/main.rs
@@ -12,7 +12,10 @@ use scxtop::layered_util;
 use scxtop::mangoapp::poll_mangoapp;
 use scxtop::search;
 use scxtop::tracer::Tracer;
-use scxtop::util::{get_clock_value, read_file_string};
+use scxtop::util::{
+    check_bpf_capability, get_capability_warning_message, get_clock_value, is_root,
+    read_file_string,
+};
 use scxtop::Action;
 use scxtop::App;
 use scxtop::CpuStatTracker;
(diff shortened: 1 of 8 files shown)


tools/scxtop/src/main.rs
```diff
@@ -106,68 +109,99 @@ fn handle_input_entry(app: &App, s: String) -> Option<Action> {
...
+                            $name,
+                            e
+                        ));
+                    } else {
+                        warnings.push(format!("Failed to attach {}: {}", $name, e));
+                    }
+                }
+            }
+        };
+    }
```
JakeHillion, line 140: I don't think you need a macro here. An `&dyn libbpf_rs::ProgramMut` should make this a loop or function rather than a macro, unless I'm misreading the generated skeleton (attach is a method on that trait).

tools/scxtop/src/main.rs
```diff
@@ -106,68 +109,99 @@ fn handle_input_entry(app: &App, s: String) -> Option<Action> {
...
-    Ok(links)
+    Ok((links, warnings))
 }
 
 fn run_trace(trace_args: &TraceArgs) -> Result<()> {
+    // Trace function always requires root privileges
+    if !is_root() {
+        return Err(anyhow!(
+            "Trace functionality requires root privileges. Please run as root:\n\
+            sudo /home/hodgesd/scx/target/release/scxtop trace [OPTIONS]"
```
JakeHillion: Should this be?:
```
std::env::args().next().unwrap()
```
for arg0?

PR #2670, opened 2025-08-23
scx_utils: few helpers simplifications.

devnexen:
- get_pd_by_cpu_id() using map iterator find instead of manual lookup.
- try_set_rlimit_infinity() using a little more rust-y idiom for error checks.

### DIFF
--- a/rust/scx_utils/src/energy_model.rs
+++ b/rust/scx_utils/src/energy_model.rs
@@ -73,12 +73,10 @@ impl EnergyModel {
     }
 
     pub fn get_pd_by_cpu_id(&self, cpu_id: usize) -> Option<&PerfDomain> {
-        for (_, pd) in self.perf_doms.iter() {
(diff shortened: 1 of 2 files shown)


rust/scx_utils/src/energy_model.rs
```diff
@@ -73,12 +73,11 @@ impl EnergyModel {
...
 
     pub fn get_pd_by_cpu_id(&self, cpu_id: usize) -> Option<&PerfDomain> {
-        for (_, pd) in self.perf_doms.iter() {
-            if pd.span.test_cpu(cpu_id) {
-                return Some(pd);
-            }
-        }
-        None
+        self.perf_doms
+            .iter()
```
hodgesds: Why not use `.values()` if this is trying to be more idiomatic?

hodgesds (approved):

PR #2672, opened 2025-08-24
c/scx_userland: Fix wrong return value

EricccTaiwan:
This PR has no tail benefit; feel free to drop.


htejun:
The return value is passed to `strerror()` which expects a positive error number. I think the lack of `-` is intended.

EricccTaiwan:
> The return value is passed to `strerror()` which expects a positive error number. I think the lack of `-` is intended.

Oh, I see, thanks for the clarification!

PR #2673, opened 2025-08-24
scx_p2dq: Add scheduler mode configuration

hodgesds:
Add different modes for gaming, efficiency and default scheduler performance. For the the implementation is rather simple and only effects the select path.


etsal (approved):

multics69 (approved):
LGTM

PR #2675, opened 2025-08-25
scx_lavd: Use Peek operation in lavd if available

rrnewton:
See commit description for the first diff, copied below.

------
[SCX][lavd] modify lavd to use new peek operation
Summary:

The per-cpu queues mode in LAVD checks multiple queues for potential steal targets.
These checks can be made more efficient with the forthcoming O(1) lockless peek operation.
This patch updates LAVD to use peek in three places:

- try_to_steal: picking cpu with lowest vtime
- on consume_task: checking the cpu DSQ
- on consume_task: checking the domain DSQ

Note that these usages are necessarily racy because no lock is held. That is, we
could peek a low vtime, but the task may have moved or completed by the time we
attempt to steal it.

Test Plan:
 - ran schbench under a VM [1 cpdom, 50cpus]

Reviewers: daidavid

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -215,38 +215,6 @@ static int pick_cpu_with_lowest_vtime(struct cpdom_ctx *cpdomc)
 	return pick_cpu;
 }
 
-/*
- * For simplicity, try to just steal from the CPU with
- * the highest number of queued_tasks in this domain.
- */
-static int pick_most_loaded_cpu(struct cpdom_ctx *cpdomc) {
-	u64 highest_queued = 0;
-	int pick_cpu = -ENOENT;
-	int cpu, i, j;
-
-	if (!per_cpu_dsq)
-		return -ENOENT;
-
-	bpf_for(i, 0, LAVD_CPU_ID_MAX/64) {
-		u64 cpumask = cpdomc->__cpumask[i];
-		bpf_for(j, 0, 64) {
-			if (cpumask & 0x1LLU << j) {
-				u64 queued;
-				cpu = (i * 64) + j;
-				if (cpu >= __nr_cpu_ids)
-					break;
-				queued = scx_bpf_dsq_nr_queued(cpu_to_dsq(cpu));
-				if (queued > highest_queued) {
-					highest_queued = queued;
-					pick_cpu = cpu;
-				}
-			}
-		}
-	}
-
-	return pick_cpu;
-}
-
 static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
 {
 	struct cpdom_ctx *cpdomc_pick;


rrnewton:
@daidavid - I'll post some more baremetal perf results here. Under a VM with 50 cpus, 1 cpudom, 120 GB ram, it really doesn't move schbench score.

daidavid:
@rrnewton Yes, let's get the numbers with baremetal. VM numbers are noisy and not very meaningful, a few things to keep in mind:
- I wouldn't expect the new load balancing to help with schbench that much since vtime ordering isn't that important for a microbenchmark. I'd test it on some real workloads if you can.
- I have doubts about how often the task steal path actually gets used given all of the recent changes, it would be interesting to instrument that. We might need to update the enqueue logic as well.

scheds/include/scx/common.bpf.h
```diff
@@ -75,6 +75,7 @@ u32 scx_bpf_reenqueue_local(void) __ksym;
 void scx_bpf_kick_cpu(s32 cpu, u64 flags) __ksym;
 s32 scx_bpf_dsq_nr_queued(u64 dsq_id) __ksym;
 void scx_bpf_destroy_dsq(u64 dsq_id) __ksym;
+struct task_struct *scx_bpf_dsq_peek(u64 dsq_id) __ksym __weak;
```
multics69, line 78: We need a COMPAT macro to bear with the system without scx_bpf_dsq_peek(). It should be something similar to __COMPAT_scx_bpf_get_idle_cpumask_node() in compat.bpf.h, first checking if the symbol of scx_bpf_dsq_peek() exists.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -173,6 +173,46 @@ static bool consume_dsq(struct cpdom_ctx *cpdomc, u64 dsq_id)
 	return ret;
 }
 
+/*
+ * Attempts to peek the vtime of the task at the head of the DSQ, or returns U64_MAX if the DSQ is empty.
+ */
+static int peek_vtime(u64 dsq_id) {
```
multics69: We roughly follow the Linux kernel coding style. So let's place `{` in a new line. 
```
static int peek_vtime(u64 dsq_id)
{
    /* ... */
}

```

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -173,6 +173,46 @@ static bool consume_dsq(struct cpdom_ctx *cpdomc, u64 dsq_id)
...
+	struct task_struct *task;
+    task = scx_bpf_dsq_peek(dsq_id);
+    return task ? task->scx.dsq_vtime : U64_MAX;
+}
+
+/*
+ * Racy operation that returns the cpu that which appears to have the lowest vtime at its head.
+ */
+static int pick_cpu_with_lowest_vtime(struct cpdom_ctx *cpdomc) {
```
multics69: ditto.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -173,6 +173,46 @@ static bool consume_dsq(struct cpdom_ctx *cpdomc, u64 dsq_id)
 	return ret;
 }
 
+/*
+ * Attempts to peek the vtime of the task at the head of the DSQ, or returns U64_MAX if the DSQ is empty.
+ */
+static int peek_vtime(u64 dsq_id) {
+	struct task_struct *task;
+    task = scx_bpf_dsq_peek(dsq_id);
+    return task ? task->scx.dsq_vtime : U64_MAX;
```
multics69: The indentation of these two lines looks different. Space used instead of tab?

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -256,7 +296,7 @@ static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
 			if (!READ_ONCE(cpdomc_pick->is_stealee) || !cpdomc_pick->is_valid)
 				continue;
 
-			pick_cpu = pick_most_loaded_cpu(cpdomc_pick);
```
multics69, line 259: `pick_most_loaded_cpu()` is no longer used, so would it be better to remove it?

multics69 (changes requested):
Thank you, Ryan! I left a few comments for the next version.

One nitpicking is to use "scx_lavd" instead of "[SCX][LAVD]" as a prefix of the commit message for consistency.

scheds/include/scx/common.bpf.h
rrnewton: Oh, cool, I didn't know about that protocol. So that lets us make the change and stay agnostic to when it rolls out in the kernel. Thanks.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
rrnewton: @multics69  - Do we have any equivalent of `checkpatch.pl` for this repo or any auto-formatter support? 

[Huh, funny, pick_most_loaded_cpu was already violating this and I copied it. Oops.]

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
rrnewton: The second commit does that.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
rrnewton: Good eye. I also fixed a few other places in the file that had four spaces instead of a tab.  [Sad to not have auto-formatting though!]

rrnewton:
@multics69 - ok, comments resolved. Let me know if that looks good now.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
multics69: I don't think we have the auto-formatter for the BPF code. We use `rustfmt` for the rust code.

multics69:
> @multics69 - ok, comments resolved. Let me know if that looks good now.

Thanks for addressing the comments. It looks good now. Let's merge it to the main once the kernel patch is submitted.

scheds/include/scx/compat.bpf.h
```diff
@@ -230,6 +230,13 @@ static inline bool __COMPAT_is_enq_cpu_selected(u64 enq_flags)
 	 scx_bpf_pick_any_cpu_node(cpus_allowed, node, flags) :			\
 	 scx_bpf_pick_any_cpu(cpus_allowed, flags))
 
+#define __COMPAT_scx_bpf_dsq_peek(dsq_id)                                      \
+  (bpf_ksym_exists(scx_bpf_dsq_peek) ? scx_bpf_dsq_peek(dsq_id) : ({           \
+    struct task_struct *p = NULL;                                              \
+    bpf_for_each(scx_dsq, p, dsq_id, 0) { break; }                             \
+    p;                                                                         \
+  }))
+
```
JakeHillion, line 239: The statement expression here is non-standard C I believe, do we already use it widely? IMO it would be better to declare either this whole compat function as a `static __always_inline` function or declare the iterator based case as an always inline function and keep the macro for consistency.

htejun: Statement exprs are widely used in kernel source, so I don't think it'd be useful to try to avoid them on the account of it not being standard. I don't we'll ever use a compiler which doesn't support it.

htejun:

PR #2677, opened 2025-08-25
lib/selftest: add C based runner

etsal:
Introduce a C-based selftest runner for the arena library code. This additional runner is both more convenient than the Rust-based one and improves our coverage:

1) It compiles 20x faster than the Rust-based runner (1.5s vs 30s).
2) It loads and runs the BPF selftests directly from C code. This tests whether our headers are `include`able in C userspace code.
3) Provides an example use of the arena libraries from a pure C environment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3087,6 +3087,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "scxcash"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "ctrlc",
+ "libbpf-rs",
+ "libc",
+ "log",
+ "num_cpus",
+ "scx_utils",
+ "serde",
+ "serde_json",
+ "simplelog",
+]
+
+[[package]]
 name = "scxctl"
 version = "1.0.15"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
     "scheds/rust/scx_tickless",
     "scheds/rust/scx_wd40",
     "tools/scx_loader",
+    "tools/scxcash",
     "tools/scxctl",
     "tools/scxtop",
     "tools/vmlinux_docify",
--- a/kernel-versions.json
+++ b/kernel-versions.json
@@ -10,9 +10,9 @@
   "bpf/bpf-next": {
     "repo": "https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git",
     "branch": "master",
-    "commitHash": "4223bf833c8495e40ae2886acbc0ecbe88fa6306",
-    "lastModified": 1756094238,
-    "narHash": "sha256-5Pf0lh9+w7RR5A6xTeATk4XEe+E2RgZkIHOiqNOOE6k=",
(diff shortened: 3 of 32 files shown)


lib/selftests/selftest.c
```diff
@@ -0,0 +1,341 @@
...
+#define TOPO_LEVELS 5
+#define CPUMASK_MAX_SIZE 64
+#define TOPO_MAX_CPU (64  * CPUMASK_MAX_SIZE)
+#define TOPO_MAX_ID (4096)
+
+u64 topo_top[CPUMASK_MAX_SIZE];
+u64 topology[TOPO_LEVELS - 1][TOPO_MAX_ID][CPUMASK_MAX_SIZE];
+u64 topo_max_offset[TOPO_LEVELS];
+
+#define VALIDATE_PERROR(errval, perrstr) do {				\
```
hodgesds: the `\` spacing looks all over, not sure if that is the github UI though

lib/selftests/selftest.c
```diff
@@ -0,0 +1,341 @@
...
+	int progret;
+	int ret;
+
+	args = (struct arena_init_args) {
+		.static_pages = 512,
+		.task_ctx_size = sizeof(struct task_ctx_nonarena),
+	};
+
+	memset(&opts, 0, sizeof(opts));
+	opts =  (struct bpf_test_run_opts) {
```
hodgesds:
```suggestion
	opts = (struct bpf_test_run_opts) {
```

hodgesds (approved):
LGTM

lib/selftests/Makefile
```diff
@@ -0,0 +1,37 @@
+.PHONY: clean test
+BPF_SOURCES = $(wildcard ../*.bpf.c) $(wildcard *.bpf.c)
+BPF_OBJECTS = $(notdir $(BPF_SOURCES:.bpf.c=.bpf.o))
+BPFTOOL=bpftool
+
+INCLUDES=-I../../scheds/include -I../../scheds/include/arch -I ../../scheds/include/arch/bpf-compat
+
+BPF_CFLAGS=-O2 -g -target bpf -D__TARGET_ARCH_x86 -mcpu=v3 -Wall
+BPF_CFLAGS+=-I/usr/include/bpf -I/usr/include/$(shell uname -m)-linux-gnu
```
hodgesds, line 9: I think this should work for most systems

hodgesds:
I think CI might need some tweaks to make this work

lib/selftests/Makefile
etsal: Yeah, the only thing we can change is modify the target to work for all architectures and not just for X86

lib/selftests/selftest.c
etsal: Yeah not sure why, it looks good in my editor - may skip the alignment just in case

etsal:
Not sure what is going on with the CI. The C runner works, the Rust runner works, p2dq works.

etsal:
Turning off the p2dq selftests after consulting with @hodgesds. The tests do not compile even if p2dq compiles and runs. Both C and Rust runners for the data structures return success. We can turn them back on once we figure out why the selftests behave differently compared to the scheduler itself.

PR #2680, opened 2025-08-26
scx_layered: Added --hi-fb-thread-name to give more CPU time for affinitized task

xerothermic:
By default, affinitized tasks are put in lo_fb queue to guarantee a small slice of CPU time on the CPU core to avoid stalls. However, we have an important thread that has CPU affinity but we still want to give it more CPU time. Created an option to throw matching thread to hi_fb to allow it to run.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1691,8 +1691,9 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
 					hi_fb_thread_name, p->pid);
 			taskc->dsq_id = task_cpuc->hi_fb_dsq_id;
 		}
-		else
+		else {
 			taskc->dsq_id = task_cpuc->lo_fb_dsq_id;
+		}
 		/*
 		 * Start a new lo fallback queued region if the DSQ is empty.
 		 * While the following is racy, all that's needed is at least
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -66,6 +66,9 @@ const volatile bool enable_gpu_support = false;
 /* Delay permitted, in seconds, before antistall activates */
 const volatile u64 antistall_sec = 3;
 const u32 zero_u32 = 0;
+/* Move matching thread name to hi_fb */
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1679,7 +1682,17 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
...
-		taskc->dsq_id = task_cpuc->lo_fb_dsq_id;
+		// Special handle for thread that has affinity set, but need more CPU time.
+		// XXX: If we need to support more than one thread (which is probably bad from
+		// deterministic performance perspective), we can change this to a map later.
+		if (enable_hi_fb_thread_name_match &&
+				bpf_strncmp(p->comm, sizeof(p->comm), (const char *)hi_fb_thread_name) == 0) {
+			trace("Put %s[%d] in hi_fb_dsq",
+					hi_fb_thread_name, p->pid);
+			taskc->dsq_id = task_cpuc->hi_fb_dsq_id;
+		}
```
htejun, line 1693: formatting:
```
if () {
} else {
}
```

scheds/rust/scx_layered/src/main.rs
```diff
@@ -1437,7 +1441,7 @@ impl GpuTaskAffinitizer {
                         self.tasks_affinitized += 1;
                     }
                     Err(_) => {
-                        warn!(
+                        debug!(
```
htejun, line 1444: Is this change intentional?

scheds/rust/scx_layered/src/main.rs
```diff
@@ -2316,6 +2320,13 @@ impl<'a> Scheduler<'a> {
             rodata.task_hint_map_enabled = true;
         }
 
+        if opts.hi_fb_thread_name.is_empty() == false {
```
htejun: This is preference but I'm not a big fan of `== false`. Can we do `!` instead?

scheds/rust/scx_layered/src/main.rs
likewhatevs: yeah that was making a lot of spam

scheds/rust/scx_layered/src/bpf/main.bpf.c
htejun: Closing '}' and 'else {' on the same line.

htejun (approved):

PR #2689, opened 2025-08-27
Bump slab to 0.4.11

shunghsiyu:
Current slab 0.4.10 is affected by CVE-2025-55159/RUSTSEC-2025-0047/GHSA-qx2v-8332-m4fv, use 0.4.11 which has the corresponding fix instead.


hodgesds (approved):
lgtm, thanks for updating

PR #2693, opened 2025-08-27
lib/atq: Back with rbtrees

etsal:
Change the backing data structure for ATQs from minheaps to rbtrees. This allows us to have unlimited capacity and lets us implement lookups and in-place updates in the future.

### DIFF
--- a/kernel-versions.json
+++ b/kernel-versions.json
@@ -10,9 +10,9 @@
   "bpf/bpf-next": {
     "repo": "https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git",
     "branch": "master",
-    "commitHash": "4223bf833c8495e40ae2886acbc0ecbe88fa6306",
-    "lastModified": 1756094238,
-    "narHash": "sha256-5Pf0lh9+w7RR5A6xTeATk4XEe+E2RgZkIHOiqNOOE6k=",
+    "commitHash": "737433c6a559c4e8acb065cfe9b6e2ff45ad655c",
+    "lastModified": 1756352757,
+    "narHash": "sha256-29swWp1vdMe89kNjtCxL6uTGwISTkApyKJFIIat6YtI=",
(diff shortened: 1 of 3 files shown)


scheds/include/lib/rbtree.h
```diff
@@ -38,19 +38,25 @@ struct rbtree {
 
 typedef struct rbtree __arena rbtree_t;
 
+enum rbtree_insert_mode {
+	RB_DEFAULT,
+	RB_UPDATE,
+	RB_DUPLICATE,
```
multics69, line 59: It will be nice to add some description about what those options are. For me, it took a bit of time to figure it out.

lib/atq.bpf.c
```diff
@@ -17,57 +17,114 @@ u64 scx_atq_create_internal(bool fifo, size_t capacity)
...
+		ret = -ENOSPC;
+		goto error;
+	}
+
+	/*
+	 * "Leak" the seq on error. We only want
+	 * sequence numbers to be monotonic, not
+	 * consecutive.
+	 */
+	node->key = atq->seq++;
```
multics69, line 64: It seems the `aqt->seq` is not used anywhere. Is it for debugging?

etsal: It is used for FIFO queues to order the tasks by order of arrival. We set the node key to `seq` after locking the ATQ lock to ensure we read and increment it atomically (we could do this during node creation with __fetch_and_add, but that would be an extra atomic).

scheds/include/lib/rbtree.h
etsal: For sure! To be clear, these flags are rbtree-specific, they're not necessary when using the ATQs.

etsal:
Addressed feedback, additionally I changed the behavior on RB_UPDATE to be consistent with the other modes and replace the _node_ in the tree, not the _value_.

multics69 (approved):
Thanks! Looks great!

PR #2694, opened 2025-08-27
scx_mitosis: dynamic cell creation/destruction

dschatzberg:
This commit adds dynamic cell creation and destruction. Cells are defined by having a cpuset applied to a cgroup. The way we determine this is through writes to cpuset.cpus (e.g. a tracepoint) or on cgroup_init. There's a lot of concern about racy behavior here, so the logic is anytime we see a change, we trigger a subsequent tick() to actually evaluate which cells should exist. This enables the logic to be serialized.


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -874,63 +747,319 @@ static inline int cgroup_init_with_cpuset(struct cgrp_ctx *cgc,
...
+		struct cgrp_ctx *cgrp_ctx;
+		if (!(cgrp_ctx = lookup_cgrp_ctx_fallible(cur_cgrp)))
+			continue;
+
+		int rc = get_cgroup_cpumask(cur_cgrp, entry);
+		if (!rc) {
+			/*
+			 * TODO: If this was a cell owner that just had its cpuset removed,
+			 * it should free the cell. Doing so would require draining
+			 * in-flight tasks scheduled to the dsq.
```
kkdwvd, line 861: Is this more difficult than scx_bpf_dsq_move() to the cell DSQ of the new parent cell? I assume we don't have to touch CPU DSQs since they will now be managed by the parent cell?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -874,63 +747,319 @@ static inline int cgroup_init_with_cpuset(struct cgrp_ctx *cgc,
...
+	/* scale the execution time by the inverse of the weight and charge */
+	p->scx.dsq_vtime += used * 100 / p->scx.weight;
+
+	if (cidx != 0 || tctx->all_cell_cpus_allowed) {
+		u64 *cell_cycles = MEMBER_VPTR(cctx->cell_cycles, [cidx]);
+		if (!cell_cycles) {
+			scx_bpf_error("Cell index is too large: %d", cidx);
+			return;
+		}
+		*cell_cycles += used;
```
kkdwvd, line 1059: I know this is not part of the main changes, and just a hunk moving around, but why don't we charge the runtime in tctx->all_cell_cpus_allowed is false? Is it because in that case task to cell association is broken and we may not run on such a CPU? We probably don't make use of this utilization information yet such that it has any material effect but just wondering.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -982,8 +1119,12 @@ s32 BPF_STRUCT_OPS(mitosis_cgroup_exit, struct cgroup *cgrp)
 		return -ENOENT;
 	}
 
-	if (cgc->cell_owner)
-		return free_cell(cgc->cell);
+	if (cgc->cell_owner) {
+		int ret;
+		if ((ret = free_cell(cgc->cell)))
+			return ret;
+		__atomic_add_fetch(&configuration_seq, 1, __ATOMIC_RELEASE);
```
kkdwvd, line 1140: Can you explain why this reconfiguration is needed here? Is it to update the cpumasks for parent cgroup's cell?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -874,63 +747,319 @@ static inline int cgroup_init_with_cpuset(struct cgrp_ctx *cgc,
...
+
+		bpf_cpumask =
+			bpf_kptr_xchg(&cell_cpumaskw->tmp_cpumask, bpf_cpumask);
+		if (bpf_cpumask) {
+			scx_bpf_error("tmp_cpumask should be null");
+			bpf_cpumask_release(bpf_cpumask);
+			goto out_rcu_unlock;
+		}
+
+		barrier();
```
kkdwvd, line 952: I think these are probably not sufficient beyond x86, and we need proper acquire/release barriers, but this is a problem elsewhere in the code as well; I'll take a look in fixing this later, for now this is fine I guess (since on x86 after suppression of compiler optimizations a load/store is acquire/release anyway).

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -874,63 +747,319 @@ static inline int cgroup_init_with_cpuset(struct cgrp_ctx *cgc,
...
+		WRITE_ONCE(cgrp_ctx->cell, cell_idx);
+		u32 level = cur_cgrp->level;
+		if (level <= 0 || level >= MAX_CG_DEPTH) {
+			scx_bpf_error("Cgroup hierarchy is too deep: %d",
+				      level);
+			goto out_rcu_unlock;
+		}
+		level_cells[level] = cell_idx;
```
kkdwvd, line 960: Pardon my ignorance, but can't multiple cgroups appear at the same level? In such a case, wouldn't level_cells overwrite the previous entry if we get a refresh request for a cgroup at the same level, but in another subtree?

I think everything makes sense but I can't wrap my head around how the level_cells logic is ok.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: I think it is a bit more difficult. You have to worry about concurrent select_cpu() and enqueue() with this tick() logic. So if you just scx_bpf_dsq_move() here, nothing guarantees that you won't see further enqueues to the dsq. This was the original rationale for the RCU critical section.

I haven't thought deeply about this particular scenario but perhaps you can do something like (in tick) mark the dying cell's target cell, scx_bpf_dsq_move(), and then on the select_cpu/enqueue path you need to check after enqueueing if there's a target cell and move to that. It's not clear to me when you're actually allowed to free the cell here though.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Actually, this cell_cycles is no longer used. Originally I was using it for debugging to ensure that the right cpus were running threads from the correct cells (and reporting that from userspace). But we no longer do that reporting since some refactoring from Tommy.

I think this can all be deleted. I'll do that in a followup PR

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Not just the cpumasks for the parent cell but also the cell associated with the cpus. Without this we'll just idle cpus of this dead cell until the next creation.

I'll add a comment here to clarify

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Yes, that's correct. I know I've heard about adding a memory model to BPF but I was not aware it was sufficiently mature to use. I've been using these barriers precisely as you describe - because it should be sufficient on x86. But you're completely right - we should use acquire/release semantics instead.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Yeah, please sanity check this for me - but I believe with pre-order traversal we'll always traverse an entire subtree before the next - so while its true that we'll overwrite level_cells[level] for a cgroup at the same level, we'll only do so before traversing that entire subtree. 

So if the cgroup tree is:
```
      1
  2       3
4 5     6 7
```

Then any time we are visiting 4 or 5 (at level 2) the previous cgroup we visited at level 1 must have been 2. After visiting 5, we'll visit 3 and then 6 and 7.

dschatzberg:
Thanks for the thorough review.

> Overall, I guess we don't really have hierarchical cells and flatten the whole hierarchy and create and free based on the cpuset groups created through the cgroup hierarchy, right?

Yeah, everything is flat. The use-cases we're looking at don't even worry about hierarchical cpusets - though in some instances that would "just work" (though it would also break in other cases).

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: Ok, so basically the move needs to be happen once we make sure the wake up path won't enqueue stuff into this cell's DSQ. So this will happen once applied_configuration_seq is noticed to be out of sync (that only happens once tick() terminates and applies updates from configuration_seq). Subsequently, we would rely on this write to be release all other updates such that when we read the new applied_configuration_seq, the cgrp->cell assignment for the parent is visible and the task moves there and picks that cell DSQ.

Hence, the DSQ move from the older cell's DSQ can happen at this point, since nothing would enqueue new tasks from the enqueue path into it anymore.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: I think it's fine to keep it, I am actually going to be probably hooking into this place for also tracking load on mem bw / caches etc. so it's probably fine, I was just wondering. I think it's probably ok to also use this just to log per-cell utilization metrics, esp. because in the future we might use it as a component to scale the size.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: Ohh, right, it makes sense now, so we're guaranteed to have the parent cell in the level - 1 index even if we move to another subtree, we basically use it to track the nearest ancestor in some sense, right? I guess a comment would help, it was not very obvious.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: > Ok, so basically the move needs to be happen once we make sure the wake up path won't enqueue stuff into this cell's DSQ. So this will happen once applied_configuration_seq is noticed to be out of sync (that only happens once tick() terminates and applies updates from configuration_seq).

Right, so I *think* the RCU critical section approach is still pretty sensible - you protect the wakeup path with an rcu "read" critical section and then you can drain the DSQ once you've reached a quiescent point because you'll be sure no further enqueues can happen.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Yeah, at the time I was just very interested in knowing that the particular cpus were running tasks from the expected cells and not polluting the information about pinned tasks. I'll leave it in then and you can hack it apart as you like.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Sure, I do have a comment describing this a bit above, but it doesn't explain how it works sufficiently. Let me do better there.

PR #2695, opened 2025-08-28
scx: Make kconfig symbols weak

hodgesds:
See the docs for __kconfig[1], on some systems if __weak is not used it can cause libbpf to error out on loading.

  [1] https://docs.ebpf.io/ebpf-library/libbpf/ebpf/__kconfig/

### DIFF
--- a/scheds/include/scx/bpf_arena_spin_lock.h
+++ b/scheds/include/scx/bpf_arena_spin_lock.h
@@ -16,7 +16,7 @@
 #define EOPNOTSUPP 95
 #define ETIMEDOUT 110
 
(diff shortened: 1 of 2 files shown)


multics69:
@hodgesds -- I would be a bit out of context. In the first place, why is the kconfig not properly defined? That seems problematic already.

hodgesds:
> @hodgesds -- I would be a bit out of context. In the first place, why is the kconfig not properly defined? That seems problematic already.

On bazzite and some fedora variants they don't ship with kernel configs built in or installed so libbpf fails. Going to resolve this with #2704 as it seems like a better solution.

PR #2696, opened 2025-08-28
scx_cosmos: Consider wakeup frequency when determining task deadlines

arighi:
To prioritize tasks that wake up frequently over those with long sleep periods, adjust the amount of time-slice credit a task can accumulate based on its wakeup frequency.

This allows frequently sleeping tasks to accumulate a bigger time slice credit than tasks that sleep infrequently for long intervals.

This concept is similar to the approach used in scx_flash, which limits the accumulated time slice credit based on the rate of voluntary context switches.

This mechanism helps to further prioritize latency-sensitive tasks that typically exhibit frequent sleep and wake-up patterns.

Example
=======

Recompile the kernel while running a WebGL app and measure fps:
 - before this change: ~25 fps
 -  after this change: ~45 fps

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -129,6 +129,8 @@ static u64 vtime_now;
 struct task_ctx {
 	u64 last_run_at;
 	u64 exec_runtime;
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
lgtm

multics69 (approved):
Looks good to me.

PR #2705, opened 2025-08-29
Auto partitioning support for LAVD

daidavid:
Suggests some partitioning which searches for some sane CPU count within an LLC domain while accounting
for Big/Little sizes of CPUs. This has been shown to improve performance on workloads on machines that have outsized CPUs per LLC.


rust/scx_utils/src/topology.rs
```diff
@@ -103,6 +105,14 @@ lazy_static::lazy_static! {
...
     /// IDs of other CPUs that may be onlined.
     pub static ref NR_CPUS_POSSIBLE: usize = libbpf_rs::num_possible_cpus().unwrap();
+
+    /// The range to search for when finding the number of physical cores
+    /// assigned to a partition to split a large number of cores that share
+    /// an LLC domain. The suggested split for the cores isn't a function of
+    /// the underlying hardware's capability, but rather some sane number
+    /// to help determine the number of CPUs that share the same DSQ.
+    pub static ref NR_PARTITION_MIN_CORES: usize = 4;
+    pub static ref NR_PARTITION_MAX_CORES: usize = 8;
```
htejun, line 115: Having default values is fine but I think schedulers / users should be able to supply these as parameters.

rust/scx_utils/src/topology.rs
```diff
@@ -137,6 +147,9 @@ pub struct Cpu {
     pub node_id: usize,
     pub package_id: usize,
     pub cluster_id: isize,
+
+    /// Recommended partition ID
+    pub partition_rec_id: usize,
```
htejun: I wonder whether it'd be easier for schedulers to consume if these show up as pseudo llc boundaries. There are a couple scenarios to think about but the most pressing one is machines with a single llc domain per socket, in which case, representing these as llc splits would make things pretty straightforward for all topology aware schedulers.

Representing these as fake llc domains would introduce complications on machines which actually have multiple llc domains per socket. However, I don't think this necessarily needs to be a strict hierarchy. ie. We can just represent them as LLC domains and give distance information that schedulers can optionally consume if they want to treat this specifically (or just mark some llcs to be virtual belonging to some physical id).

scheds/rust/scx_lavd/src/cpu_order.rs
```diff
@@ -284,6 +290,11 @@ impl CpuOrderCtx {
                 llc_adx: cpu_id.llc_adx,
                 llc_rdx: cpu_id.llc_rdx,
                 is_big: cpu_id.big_core,
+                partition_adx: if use_llc_partitions {
```
multics69: It is necessary to update the comments above since the criteria to create a domain is now changed.

multics69: Also, I wonder if we need to update the `dist()` function too. I guess it may not be needed. Please double-check. :-)

rust/scx_utils/src/topology.rs
multics69: `partition_rec_id` is a mouthful. Would `subllc_id` be a better name?

multics69: Currently, the kernel exposes the distance between NUMA nodes. Logically, an LLC domain is under a NUMA node, and a partition/sub-LLC domain is under an LLC domain. Since there is no SCX scheduler that leverages NUMA distance as far as I know, it would be fine to leave the details of using LLC ID and partition/sub-LLC ID to a specific scheduler at this point. Once we have a real use case of leveraging NUMA distance, we will have a better understanding how to model NUMA/LLC/partition distances.

daidavid: I was originally envisioning that the recommended ID would give schedulers more flexibility in how they wanted to use this "virtual llc".  ex. They might not want to use this as partition info at all when it comes to task placement decisions. That being said, experiments show that the extra l2 cache locality benefits are translating to tangible performance gains too, so it would make sense to just swap the physical LLC domain with the virtual ones to make it easier for schedulers to use.

So while the "sub-llc" might be the right designation for this, since we don't expect "virtual LLC" to cross physical LLC boundaries, I also don't see a need to add another layer to the hierarchy if it's adding complexity without a clear need.

IIUC, I think we're all on the same page about creating some "virtual" LLC domains with a physical LLC id for schedulers to optionally use as distance info.

scheds/rust/scx_lavd/src/cpu_order.rs
daidavid: I think if we use the virtual LLC as described above, the dist() probably just need to be updated to also use the physical LLCs ids.

rust/scx_utils/src/topology.rs
daidavid: removed due to re-implementation as virtual llcs

hodgesds:
Feels weird to overwrite the topology, why not have a `shard` method on the `Node` or `Llc` structs that returns a map with the shard id as the key and the value being whatever unit (ex: `shard_cores`).

rust/scx_utils/src/topology.rs
```diff
@@ -692,10 +711,100 @@ fn is_smt_active() -> Option<bool> {
...
+        }
+    }
+
+    let mut virt_llcs: BTreeMap<usize, Arc<Llc>> = BTreeMap::new();
+
+    for partition_id in 0..total_partitions {
+        let kernel_id = partition_to_kernel_id
+            .get(&partition_id)
+            .copied()
+            .unwrap_or(partition_id);
```
htejun: When would `.unwrap_or()` trigger? Shouldn't the lookup always succeed?

rust/scx_utils/src/topology.rs
```diff
@@ -692,10 +711,100 @@ fn is_smt_active() -> Option<bool> {
...
+
+    Ok(())
+}
+
 fn create_default_node(
     online_mask: &Cpumask,
     topo_ctx: &mut TopoCtx,
     flatten_llc: bool,
+    min_cores: Option<usize>,
+    max_cores: Option<usize>,
```
htejun: Given that these params are meaningful only when both are specified: sth like `nr_cores_per_vllc: Option<(usize, usize)>`?

rust/scx_utils/src/topology.rs
```diff
@@ -692,10 +711,100 @@ fn is_smt_active() -> Option<bool> {
     Some(smt_on == 1)
 }
 
+fn replace_with_virt_llcs(node: &mut Node, min_cores: usize, max_cores: usize) -> Result<()> {
```
htejun, line 724: I think adding more step-by-step comments on what's happening would make it easier to read this function.

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -201,6 +202,12 @@ struct Opts {
     /// Show descriptions for statistics.
     #[clap(long)]
     help_stats: bool,
+
+    /// Enable virtual LLC partitioning. Use without arguments for default limits
+    /// (2-8 cores per partition), or specify min,max values like --virt-llcs=6-12
+    /// for custom partition limits.
+    #[clap(long = "virt-llcs", value_delimiter = '-', num_args = 0..=2)]
```
htejun: I wonder whether providing similar common option mechanism like `LibbpfOpts` would be better so that individual schedulers don't have to deal with custom options.

htejun (commented):
Generally looks good to me. Left some minor comments.

htejun:
> Feels weird to overwrite the topology, why not have a `shard` method on the `Node` or `Llc` structs that returns a map with the shard id as the key and the value being whatever unit (ex: `shard_cores`).

We can make something more proper in the future (e.g. what you're suggesting or an extra intermediate layer) but I think this is a good compromise to enable quick experimentation across different schedulers. If we encapsulate option handling, we gain the ability to experiment with splitting llcs without doing anything invasive to each scheduler. Also, I think having the mechanism to specify topological options shared across schedulers would be useful - e.g. we can also add options to mock topology for testing and so on.

hodgesds:
> I think this is a good compromise to enable quick experimentation across different schedulers.

Makes sense the more I though about it. I think I was coming at it more from the perspective that the load balancer needs to be smart enough to balance within the LLC first to not introduce extra X-LLC migrations.

htejun:
> Makes sense the more I though about it. I think I was coming at it more from the perspective that the load balancer needs to be smart enough to balance within the LLC first to not introduce extra X-LLC migrations.

I think we'll get there eventually when we further split already not-so-big LLCs but the more pressing use case is working around scalability issues on large single LLC CPUs, where there will only be virtual LLCs anyway. Even for cases where there are both real and virtual LLCs, I wonder whether it'd be simpler to just handle them as LLCs at different distances from each other rather than introducing a new layer.

scheds/rust/scx_lavd/src/main.rs
daidavid: Just so I understand this correctly, we can provide some common Opt to share with the schedulers, not that there's some magical way to concat the cmdline args such that there are no modifications required on the scheduler arg parsing at all(Or is there some rust way to do this?). We could introduce some environment variables, but that doesn't sound great.

htejun: Yeah, just a common opts that each scheduler doesn't have to know the details about. Just add it to its params and pass the resulting opts struct to topo init.

scheds/rust/scx_lavd/src/cpu_order.rs
```diff
@@ -48,6 +48,7 @@ pub struct CpuId {
     pub big_core: bool,
     pub turbo_core: bool,
     pub cpu_sibling: usize,
+    pub llc_kernel_id: usize,
```
multics69: As we have three different LLC IDs now, please add comments about what `llc_kernel_id` is and move `llc_kernel_id` below the 'llc_rdx` field to cluster similar things together.

multics69:
@daidavid -- I added one comment. Overall it looks good to me.

multics69 (approved):

rust/scx_utils/src/topology.rs
```diff
@@ -692,10 +721,105 @@ fn is_smt_active() -> Option<bool> {
...
+    let mut partition_to_kernel_id: BTreeMap<usize, usize> = BTreeMap::new();
+    let mut total_partitions = 0;
+
+    // First pass: determine core to partition mapping, partition to kernel_id mapping, and total partitions needed
+    for (_llc_id, llc) in node.llcs.iter() {
+        // Group cores by type (big/little) to partition separately
+        let mut cores_by_type: BTreeMap<CoreType, Vec<usize>> = BTreeMap::new();
+
+        for (core_id, core) in llc.cores.iter() {
+            let core_type = core.core_type.clone();
```
multics69: I found that this does not work well in a system with a turbo core. In such a system, all cores are Big but one core with the highest clock frequency has a turbo flag on. In the end, the cores are classified into two groups: `Big {turbo: false}` vs. `Big {turbo: ture}`. So we should ignore the `turbo` flag when classifying the core types.

The fix would be something like this:

```
let mut cores_by_type: BTreeMap<bool, Vec<usize>> = BTreeMap::new();
for (core_id, core) in llc.cores.iter() {
    let core_type = core.core_type == CoreType::Little;
```

daidavid: Thanks for testing! Updated.

PR #2707, opened 2025-08-30
scxtop: Reduce redundant if-else

EricccTaiwan:
Simplify the logic in handle_input_entry by merging identical branches for AppState::Memory and AppState::PerfTop into a single match arm with a guard condition.


hodgesds (approved):
lgtm

PR #2712, opened 2025-09-01
scx_bpf_unittests: fix build and run tests again

JakeHillion:
These tests were disabled recently because of missing arena stubs. Add them as stubs (for now) and re-enable the tests.

I want to start using these tests in scx_chaos to test for correct behaviour without the verifier interfering, which will likely require not stubbing the arena code. But this is good enough for now.

Test plan:
- `cargo test -p scx_bpf_unittests`
- CI

### DIFF
--- a/lib/scxtest/overrides.c
+++ b/lib/scxtest/overrides.c
@@ -118,3 +118,68 @@ int scx_minheap_pop(void *heap_ptr __attribute__((unused)),
 {
 	return 0;
 }
+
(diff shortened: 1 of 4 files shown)


JakeHillion:
> LGTM, I think @etsal maybe wanted to rip some of these out

Yeah there are a tonne of stubs, my thinking is it's much easier to move to stubbing the arena fundamentals if things are already building.

PR #2715, opened 2025-09-02
scx_cargo: add specific crate for build time utilities

JakeHillion:
scx_utils currently provides both runtime and build-time utilities for sched_ext schedulers. This doesn't fit very nicely with Cargo, because Cargo builds all dependencies used at build time with `cfg(build)` defined, effectively building everything in the `scx_utils` dependency tree twice (and there's quite a lot).

Following the pattern of `libbpf-{rs,cargo}`, start solving this problem by splitting the `scx_utils` crate into `scx_utils` and `scx_cargo`. This commit does very little towards achieving this goal, re-exporting the `BpfBuilder` from `scx_utils` as `scx_cargo::BpfBuilder`. In our monorepo this is enough to switch all of our crates to depending on `scx_cargo` at build time (making the `scx_utils` dependency transitive), without making any invasive changes.

New build time features can go directly into `scx_cargo`, and eventually the BpfBuilder contents can be migrated to it, switching the `pub use` to `scx_utils` to ensure we don't break existing consumers without warning.

Test plan:
- CI


arighi (approved):
LGTM

PR #2716, opened 2025-09-02
scx_loader: Update args for p2dq

hodgesds:
Add some more reasonable args for scx_loader configuration for p2dq.


sirlucjan:
Just small nit: Could you update configuration.md also here?

Zrzut ekranu_20250902_163122

PR #2717, opened 2025-09-02
scx_p2dq: Undeprecate min_llc_runs_pick2

hodgesds:
This option is sometimes useful for more aggressive task migration for lower latency. When min_llc_runs_pick2 is set it becomes the lower bound for how many times a tasks should try to run in an LLC.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -80,30 +80,30 @@ const volatile struct {
 };
 
 const volatile struct {
-	u64 slack_factor;
 	u64 backoff_ns;
-	u64 min_nr_queued_pick2;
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_p2dq/src/bpf/main.bpf.c
```diff
@@ -1669,10 +1669,10 @@ void BPF_STRUCT_OPS(p2dq_update_idle, s32 cpu, bool idle)
 	saturated = percent_idle < p2dq_config.saturated_percent;
 
 	if (saturated)
-		min_llc_runs_pick2 = 2;
+		min_llc_runs_pick2 = min(2, lb_config.min_llc_runs_pick2);
 	else {
```
JakeHillion: unrelated, but if we have curlies on the else can we get it on the if too?

PR #2718, opened 2025-09-02
scxtop: use clang-format

JakeHillion:
scxtop gets a lot of contributors (yay!) and it's becoming quite tricky to keep the formatting consistent. Copy the kernel's clang-format and apply it to scxtop in the CI.

This formatting is more consistent, probably worse in places, and probably better in places (and this PR specifically is quite churny). Let's try it out for scxtop only and see how it goes.

Test plan:
- CI


hodgesds (approved):
nice!

PR #2719, opened 2025-09-03
scx_p2dq: Update llc initialization

hodgesds:
When initializing llc_ctx structs make sure fields such as load are properly zeroed. This ensures when the scheduler is started that the first few cycles for load balancing are done with reasonable values.


arighi (commented):
Is this really needed? I thought BPF map values are guaranteed to be zero-initialized at creation time.

PR #2723, opened 2025-09-03
scx_rustland_core: Add API to set scheduler name

EricccTaiwan:
Expose a `name: &str` parameter in `BpfScheduler::init` and write it into `struct_ops.rustland.name` before `scx_ops_load!()`.

Still wondering if we should also append the build version to the name, or just keep the exact string?

image

### DIFF
--- a/rust/scx_rustland_core/assets/bpf.rs
+++ b/rust/scx_rustland_core/assets/bpf.rs
@@ -14,6 +14,7 @@ use std::ffi::c_int;
 use std::ffi::c_ulong;
 
 use std::collections::HashMap;
+use std::os::raw::c_char;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
@@ -192,6 +193,7 @@ impl<'cb> BpfScheduler<'cb> {
         partial: bool,
         debug: bool,
         builtin_idle: bool,
+        name: &str,
     ) -> Result<Self> {
         let shutdown = Arc::new(AtomicBool::new(false));
         set_ctrlc_handler(shutdown.clone()).context("Error setting Ctrl-C handler")?;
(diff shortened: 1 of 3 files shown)


EricccTaiwan:
Hmm, this patch works on my x86, but failed on ARM, which I didn't notice. 
I'll follow up with a PR to deal with it.

```c
diff --git a/rust/scx_rustland_core/assets/bpf.rs b/rust/scx_rustland_core/assets/bpf.rs
index 39ed07e2..b1c07fd5 100644
--- a/rust/scx_rustland_core/assets/bpf.rs
+++ b/rust/scx_rustland_core/assets/bpf.rs
@@ -14,6 +14,7 @@ use std::ffi::c_int;
 use std::ffi::c_ulong;
 
  --> scheds/rust/scx_rlfifo/src/bpf.rs:248:32
    |
248 |         Self::set_scx_ops_name(&mut skel.struct_ops.rustland_mut().name, name);
    |         ---------------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&mut [u8]`, found `&mut [i8; 128]`
    |         |
    |         arguments to this function are incorrect
    |
    = note: expected mutable reference `&mut [u8]`
               found mutable reference `&mut [i8; 128]`
note: associated function defined here
   --> scheds/rust/scx_rlfifo/src/bpf.rs:297:8
    |
297 |     fn set_scx_ops_name(dst: &mut [c_char], s: &str) {
    |        ^^^^^^^^^^^^^^^^ ------------------

error[E0308]: mismatched types
   --> scheds/rust/scx_rustland/src/bpf.rs:248:32
    |
248 |         Self::set_scx_ops_name(&mut skel.struct_ops.rustland_mut().name, name);
    |         ---------------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&mut [u8]`, found `&mut [i8; 128]`
    |         |
    |         arguments to this function are incorrect
    |
    = note: expected mutable reference `&mut [u8]`
               found mutable reference `&mut [i8; 128]`
note: associated function defined here
   --> scheds/rust/scx_rustland/src/bpf.rs:297:8
    |
297 |     fn set_scx_ops_name(dst: &mut [c_char], s: &str) {
    |        ^^^^^^^^^^^^^^^^ ------------------
```

PR #2724, opened 2025-09-03
meson: add deprecation warning

JakeHillion:
Meson as a build system is being replaced by a split between Cargo/gnumake for Rust/C schedulers. Add a warning before the next release to ask packagers to switch over to the replacements.

In a future release we'll add an error that can be overridden with a configuration option, and in the release after that we'll remove the Meson build entirely.

Test plan:
- CI - the new warning shows up.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,6 +4,10 @@ project('sched_ext schedulers', 'c',
         license: 'GPL-2.0',
         meson_version : '>= 1.2.0',)
 
+warning('Meson builds are deprecated. Please switch to `cargo build` for Rust'
+        + ' schedulers and `make` for C schedulers. Meson will be removed in an'
+        + ' upcoming release.')
+
 fs = import('fs')
 
 cc = meson.get_compiler('c')


meson.build
```diff
@@ -4,6 +4,10 @@ project('sched_ext schedulers', 'c',
         license: 'GPL-2.0',
         meson_version : '>= 1.2.0',)
 
+warning('Meson builds are deprecated. Please switch to `cargo build` for Rust'
```
EricccTaiwan, line 7: `cargo build --release` ?

https://github.com/sched-ext/scx/blob/a97b5dc56fc33088a7ec8f579867e1a15637d2cd/README.md?plain=1#L153-L155

JakeHillion: I said `cargo build` to mean "use cargo build with whatever flags you deem necessary" but perhaps that's not super clear. What do you think?

EricccTaiwan: Oh got it, I think this info is enough for anyone who knows how to build it. thanks!

PR #2726, opened 2025-09-03
scx_layered: userspace logic for memory-bandwidth based layer growth

etsal:
Add a new config option and related logic for throttling CPU growth based on its memory bandwidth consumption. The new logic uses the size of memory traffic reported by the kernel to prevent the growth or even shrink layers that are consuming memory bus bandwidth above a user-defined limit.

This code is currently a no-op, as the BPF side does not do the necessary tracking. The code for tracking memory memory bandwidth consumption on the BPF side will be added as a separate patch.

### DIFF
--- a/scheds/rust/scx_layered/examples/membw.json
+++ b/scheds/rust/scx_layered/examples/membw.json
@@ -1,0 +1,47 @@
+[
+	{
+		"name": "first",
+		"matches": [
+			[{ "PcommPrefix": "htop" }],
+			[{ "PcommPrefix": "scxtop" }],
+			[{ "PcommPrefix": "bash" }],
+			[{ "PcommPrefix": "yes" }]
+		],
+		"kind": {
+			"Grouped": {
+				"cpus_range":  [5, 5],
+				"util_range": [0.4, 0.85],
+				"growth_algo": "RandomTopo",
+				"disallow_preempt_after_us": 0,
+				"protected": true,
+				"membw_gb": 25
+			}
+		}
+	},
+	{
+		"name": "second",
+		"matches": [
+			[{ "PcommPrefix": "stress-ng" }]
+		],
+		"kind": {
+			"Grouped": {
+				"cpus_range":  [5, 5],
+				"util_range": [0.4, 0.85],
+				"growth_algo": "RandomTopo",
+				"protected": false,
+				"membw_gb": 40
+			}
+		}
+	},
+	{
+		"name": "third",
+		"matches": [
+			[]
+		],
+		"kind": {
+			"Open": {
+				"growth_algo": "RandomTopo"
+			}
+		}
+	}
+]
--- a/scheds/rust/scx_layered/src/config.rs
+++ b/scheds/rust/scx_layered/src/config.rs
@@ -160,6 +160,9 @@ pub enum LayerKind {
 
         #[serde(flatten)]
         common: LayerCommon,
+
+        #[serde(default)]
+        membw_gb: Option<u64>,
     },
     Grouped {
         util_range: (f64, f64),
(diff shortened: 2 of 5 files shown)


scheds/rust/scx_layered/examples/membw.json
```diff
@@ -0,0 +1,47 @@
...
+			[{ "PcommPrefix": "yes" }]
+		],
+		"kind": {
+			"Grouped": {
+				"cpus_range":  [5, 5],
+				"util_range": [0.4, 0.85],
+				"growth_algo": "RandomTopo",
+				"disallow_preempt_after_us": 0,
+				"protected": true,
+				"membw_gb": 25
```
likewhatevs, line 17: Would a frac or pct make more sense for this, if possible/practical?

likewhatevs (commented):
This looks pretty sick. 

It does (and has to) include open execution bandwidth use because of how we're measuring bandwidth though, right?

scheds/rust/scx_layered/src/config.rs
```diff
@@ -155,6 +155,9 @@ pub enum LayerKind {
         #[serde(default)]
         cpus_range_frac: Option<(f64, f64)>,
 
+        #[serde(default)]
+        membw_gb: Option<u64>,
```
htejun: If in u64, wouldn't bytes be a better unit for internal purpose? I get that for interface, GB in float would work well but I don't think GB in int is a good unit for internal representation.

etsal: We only use GiB for the user-provided option, internally we do * 1024^3 before using it for BW control (`clamp_target_by_membw`) so the BW history is in bytes. The reason we have f64 for `prev_layer_membw_agg` is for uniformity with the CPU usage history, but since it is in bytes and expected to be huge we can just round it to u64.

EDIT: To further clarify, the future BPF side stats to be in bytes so that's what the history will be reported in. This is not implicit now by the fact `clamp_target_by_membw` converts the BW limit from GiB to bytes, but not the history.

htejun: I'm a bit slow, so `membw_gb: Option<u64>` is in bytes?

scheds/rust/scx_layered/examples/membw.json
etsal: As in, pct of the total BW of the bus? There's definitely a way to grab this from `/proc`, will find it and add that as an option the same way we have it for # of CPUs in the system.

scheds/rust/scx_layered/src/config.rs
etsal: Sorry I may have misunderstood something - membw_gb is in GiB, and is always a u64 integer. Internally, it gets translated to bytes before getting used only in integer calculations, so we don't really store floating-point gigabyte values. There's a single point where we use floating point bytes for the bandwidth, but we can remove it. Were you referring to a specific variable that's f64?

htejun: Isn't integer gb too course a number? Is there a reason to not use floating point if the unit is in gb? ie. Use floating + gb or use u64 bytes?

etsal: I thought the average memory BW is large enough that 1GiB steps per layer are as fine-grained as people would like. I will change this to option to use floating point and translate to (integer) bytes when we set up the scheduler.

scheds/rust/scx_layered/examples/membw.json
etsal: It looks like actually it's not that readily available, so maybe better to leave it for a future patch.

likewhatevs (approved):

PR #2727, opened 2025-09-03
[WIP] scx_rustland_core: Fix ARM build failure

EricccTaiwan:
Fixes ARM build by casting `&mut [i8; 128]` $\to$ `&mut [c_char]`.
Init is bloated or move `unsafe` into `set_scx_ops_name`.


JakeHillion:
Sorry, I thought I'd tested this on ARM when you put the PR up, but apparently I forgot to check the final outcome :(

If you take a look at https://github.com/sched-ext/scx/blob/c1d2355456d27641566c666398cd08c2ddc87c44/rust/scx_utils/src/compat.rs#L390-L415, it should be possible to avoid the `unsafe` here by using `as_bytes()` on the string.

More generally though, I wouldn't expect you to pass a compatible slice to `set_scx_ops_name`. Keep `set_scx_ops_name` taking the high level Rust String, and do the work inside it to place it into the ops name. We don't want all rustland_core users to have to write unsafe code to use this (on the surface) simple feature.

EricccTaiwan:
> Sorry, I thought I'd tested this on ARM when you put the PR up, but apparently I forgot to check the final outcome :(

Nope, it's my bad lol. I'll double-check next time. 

> If you take a look at
> 
> https://github.com/sched-ext/scx/blob/c1d2355456d27641566c666398cd08c2ddc87c44/rust/scx_utils/src/compat.rs#L390-L415
> 
> , it should be possible to avoid the `unsafe` here by using `as_bytes()` on the string.
> More generally though, I wouldn't expect you to pass a compatible slice to `set_scx_ops_name`. Keep `set_scx_ops_name` taking the high level Rust String, and do the work inside it to place it into the ops name. We don't want all rustland_core users to have to write unsafe code to use this (on the surface) simple feature.

Ok, I’ll need a day (or less) to take a look, fix, and test this. 
Also, when are we planning to release the next version? (I don’t want to mess it up.)

JakeHillion:
> > Sorry, I thought I'd tested this on ARM when you put the PR up, but apparently I forgot to check the final outcome :(
> 
> Nope, it's my bad lol. I'll double-check next time.
> 
> > If you take a look at
> > https://github.com/sched-ext/scx/blob/c1d2355456d27641566c666398cd08c2ddc87c44/rust/scx_utils/src/compat.rs#L390-L415
> > 
> > , it should be possible to avoid the `unsafe` here by using `as_bytes()` on the string.
> > More generally though, I wouldn't expect you to pass a compatible slice to `set_scx_ops_name`. Keep `set_scx_ops_name` taking the high level Rust String, and do the work inside it to place it into the ops name. We don't want all rustland_core users to have to write unsafe code to use this (on the surface) simple feature.
> 
> Ok, I’ll need a day (or less) to take a look, fix, and test this. Also, when are we planning to release the next version? (I don’t want to mess it up.)

Hoping for a release today or tomorrow, I'll consider this PR a blocker for it so don't worry. If you need any help getting it fixed up let me know.

rust/scx_rustland_core/assets/bpf.rs
```diff
@@ -294,11 +293,17 @@ impl<'cb> BpfScheduler<'cb> {
     }
 
     // Set the name of the scx ops.
-    fn set_scx_ops_name(dst: &mut [c_char], s: &str) {
+    fn set_scx_ops_name(dst: &mut [i8], s: &str) {
+        if !s.is_ascii() {
+            panic!("name must be an ASCII string");
```
JakeHillion: Could you make this function return an `anyhow::Result<()>` and use `bail!` as a direct replacement for `panic!`?

EricccTaiwan: Ok!

EricccTaiwan:
Tested on both x86 and ARM, it works. :tada:

Edit:
> Still wondering if we should also append the build version to the name, or just keep the exact string?
> Ref #2723 

I think we should add a version suffix as well, but we can release first. (hope so)
I'll follow up with another PR soon.

EricccTaiwan:
@arighi @JakeHillion plz merge this, thanks!

PR #2728, opened 2025-09-04
scx_p2dq: Refactor enqueue for unselected tasks

hodgesds:
Update enqueue so CPU selection is always run in enqueue if a task has not been selected. This is needed because p2dq always reenqueues on cpu_release.

### DIFF
--- a/scheds/rust/scx_chaos/src/lib.rs
+++ b/scheds/rust/scx_chaos/src/lib.rs
@@ -431,10 +431,6 @@ impl Builder<'_> {
 
         let rodata = open_skel.maps.rodata_data.as_mut().unwrap();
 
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_chaos/src/lib.rs
```diff
@@ -431,10 +431,6 @@ impl Builder<'_> {
 
         let rodata = open_skel.maps.rodata_data.as_mut().unwrap();
 
-        // TODO: figure out how to abstract waking a CPU in enqueue properly, but for now disable
-        // this codepath
-        rodata.p2dq_config.select_idle_in_enqueue = MaybeUninit::new(false);
-
```
JakeHillion, line 437: yay

PR #2735, opened 2025-09-04
scx_cosmos: Introduce --avoid-smt

arighi:
On some architectures, scheduling contending tasks on the same SMT core can severely hurt performance, while on others the impact is much less significant.

To address this, introduce an option that lets the user choose whether to aggressively avoid SMT contention.


scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -71,6 +71,11 @@ const volatile bool cpufreq_enabled = true;
  */
 const volatile bool numa_enabled;
 
+/*
+ * Aggressively try to avoid SMT contention.
+ */
+const volatile bool avoid_smt;
```
JakeHillion: Might be worth defaulting to `true` so veristat takes the more complicated path.

arighi: Good idea!

hodgesds (approved):

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -488,7 +495,8 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags, bo
 	 * there first.
 	 */
 	if (!primary_all && mask) {
-		cpu = scx_bpf_select_cpu_and(p, prev_cpu, wake_flags, mask, 0);
+		cpu = scx_bpf_select_cpu_and(p, prev_cpu, wake_flags, mask,
```
hodgesds, line 498: is it worth doing `SCX_PICK_IDLE_CORE` and then 0 (idle cpu)  if `avoid_smt` is set for better locality? It's another if statement, but might be worth it.

multics69 (approved):
LGTM. It is interesting that delaying an execution to use the SMT core less aggressively helps. Do you have some insight about what types of workloads matter?

arighi:
> LGTM. It is interesting that delaying an execution to use the SMT core less aggressively helps. Do you have some insight about what types of workloads matter?

Actually on AMD the "aggressive SMT avoidance" seems to hurt performance. On Intel with hybrid cores it seems to help a lot, as it seems better to use a fast and a slow core than running 2 tasks in the same SMT fast core.

Moreover, for some reasons (that I'm still trying to understand) I see better performance with sysbench (max and avg latency) both on AMD and Intel with this new option enabled.

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
arighi: It's already doing that internally, `scx_bpf_select_cpu_and()` first tries to pick a full-idle SMT core, following the hierarchy: prev_cpu, llc, numa node, all CPUs. Then it's doing the same but also for partially-idle SMT cores.

arighi: Oh wait, you mean another:
```
    if (avoid_smt)
         cpu = scx_bpf_select_cpu_and(p, prev_cpu, wake_flags, mask, 0);
```

In that case, no, the idea is to go outside the primary domain if there's no full-idle SMT core inside the primary domain.

hodgesds: Ok, that makes sense!

cloehle:
>Moreover, for some reasons (that I'm still trying to understand) I see better performance with sysbench (max and avg latency) both on AMD and Intel with this new option enabled.

Latency should always be better if you avoid SMT right? It's throughput that should decrease by nosmt.

arighi:
> > Moreover, for some reasons (that I'm still trying to understand) I see better performance with sysbench (max and avg latency) both on AMD and Intel with this new option enabled.
> 
> Latency should always be better if you avoid SMT right? It's throughput that should decrease by nosmt.

It depends... on my AMD Ryzen AI 9 HX 370 with hybrid cores I get better latency performance (fps tail latency) prioritizing the fast cores, allocating both SMT siblings, before starting to use the slow cores.

But I guess in the majority of the cases avoiding SMT should give better performance, so maybe we should enable the smt avoidance by the default. But then the primary domain can also be used to reduce power consumption and prioritize the slow cores, in this case we want to pack tasks on the primary domain before moving outside, so probably the default is good as it is.

PR #2737, opened 2025-09-04
scx_layered: Add regex template matching

hodgesds:
Add support for cgroup regex matching in the template config. This allows for matching complex cgroup layouts which currently only supports suffix matches.

Example config:
```
  {
    "name": "my_regex_layer",
    "comment": "match some cgroups",
    "template": {
            "CgroupRegex" : ".*prod_job.*task-main.service"
    },
    "matches": [
      [
        {
          "PcommPrefix": "myservice"
        }
      ]
    ],
    ...
  },
 ```

### CODE REVIEW

**Review by JakeHillion — APPROVED** [review]

File: scheds/rust/scx_layered/src/main.rs [hunk 1]
```diff
@@ -3323,6 +3327,27 @@ fn expand_template(rule: &LayerMatch) -> Result<Vec<(LayerMatch, Cpumask)>> {
                 )
             })
             .collect()),
+        LayerMatch::CgroupRegex(expr) => Ok(traverse_sysfs(Path::new("/sys/fs/cgroup"))?
```
hodgesds, line 3330: At some point we should get the mount point with [`get_fs_mount`](https://github.com/sched-ext/scx/blob/6b4fcd28b667b3184f37106a310d4804c4c1f304/rust/scx_utils/src/compat.rs#L183).

PR #2741, opened 2025-09-05
versions: bump versions for 1.0.16 release

JakeHillion:
Usual version bump. A couple of changes to dependency versions to sync them up.


JakeHillion:
Slight issue, the new vmlinuxs (I think) have scx_utils failing to publish because the blob of sources is too large for crates.io. Figuring out how to work around this.

hodgesds (approved):

arighi:
> Slight issue, the new vmlinuxs (I think) have scx_utils failing to publish because the blob of sources is too large for crates.io. Figuring out how to work around this.

Should we fetch them from github via build.rs? Not sure if this would make packaging more complicated for certain distros...

JakeHillion:
> Should we fetch them from github via build.rs? Not sure if this would make packaging more complicated for certain distros...

As we said offline, this probably isn't the way to go. Would make life really hard for any distros relying on `cargo vendor` to do the online bit. Working on getting these from a checked in `tar.zst` at the minute. Will put up a separate PR with that change when it's done and use it to unblock the release.

PR #2747, opened 2025-09-08
rust: add generic arena library setup code and consolidate arena library code to subdir

etsal:
Consolidate code related to the arena library into its own subdirectory. The new subdirectory has two modules:

1) The existing, renamed selftests module for the arena library.
2) A new module for common arena initialization code.

The latter module replaces scheduler-specific code for arena initialization and makes it easy to include arenalib into new schedulers. Up until now we have been copy-pasting the init code across schedulers because we have been managing BPF state exclusively through the BPF skeleton. These skeletons are autogenerated at compile time and are scheduler-specific, even if the underlying BPF code is not.

We avoid this problem using introspection directly into the object underlying the BPF skeleton. The module receives the opaque BPF object from the scheduler, then dynamically retrieves the arena library BPF programs used for initializing arena state. Schedulers that want to use the arena library now only need to create a new ArenaLib object using .init(), then call its .setup() method.

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -282,7 +282,7 @@ async def run_tests():
     print("Running tests...", flush=True)
 
     # Make sure the selftest is built in case the build was not already run.
-    await run_command(["cargo", "build", "-p", "scx_lib_selftests"], no_capture=True)
+    await run_command(["cargo", "build", "-p", "arenalib_selftests"], no_capture=True)
 
     await run_command(
         [
@@ -702,7 +702,7 @@ async def run_tests_in_vm():
         check=True,
     )
 
-    subprocess.run(["target/debug/scx_lib_selftests"], check=True)
+    subprocess.run(["target/debug/arenalib_selftests"], check=True)
 
 
 async def run_all():


rust/arenalib/arenalib/src/arenalib.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        let ret = self.run_prog_by_name("arena_topology_node_init", input)?;
+        if ret != 0 {
+            bail!("arena_topology_node_init returned {}", ret);
+        }
+
+        Ok(())
+    }
+
+    fn setup_topology(&self) -> Result<()> {
```
hodgesds, line 151: In the future it might be nice to have this pass in the a `Topology` struct rather than creating a new one, especially if different options are used to initialize it.

etsal: We can add that as an option in the future, for now I didn't add it because the existing consumers don't create the topology themselves.

multics69:
This is awesome! Thanks @etsal !

JakeHillion:
Could you refactor this into the existing library `scx_userspace_arena`? There's not much in it at the minute, and we can technically publish an `arenalib` crate at the minute (the name is unused) but it's not very descriptive for the average Rust user.

etsal:
I think it's better to keep those separate, the codebases have different goals (exposing arena memory-mapped regions to a Rust allocator vs userspace handling for the arena library itself). For now providing it as an independent crate is out of scope imo: The only user is the scx scheduler repo. In the future when the arena library is more self-contained we can revisit this.

JakeHillion:
> I think it's better to keep those separate, the codebases have different goals (exposing arena memory-mapped regions to a Rust allocator vs userspace handling for the arena library itself). For now providing it as an independent crate is out of scope imo: The only user is the scx scheduler repo. In the future when the arena library is more self-contained we can revisit this.

Out of scope? It has to be provided as a separate crate published to crates.io else we can't publish any of the schedulers that depend on it.

etsal:
> > I think it's better to keep those separate, the codebases have different goals (exposing arena memory-mapped regions to a Rust allocator vs userspace handling for the arena library itself). For now providing it as an independent crate is out of scope imo: The only user is the scx scheduler repo. In the future when the arena library is more self-contained we can revisit this.
> 
> Out of scope? It has to be provided as a separate crate published to crates.io else we can't publish any of the schedulers that depend on it.

Oh in that case it isn't. With all the crate publishing I understand why we are dumping every single utility we can imagine into scx_utils.

JakeHillion:
> Oh in that case it isn't. With all the crate publishing I understand why we are dumping every single utility we can imagine into scx_utils.

Yeah, it's a bit of a PITA. On the one hand Cargo naturally pushes you to create lots of crates. On the other hand you have to push them all into a flat namespace if you want to use crates.io at all. Sounds like (from office hours) the fix here is going to be a pretty simple rename.

PR #2757, opened 2025-09-10
arenalib: make imports a bit easier

daidavid:
Use the libbpf_sys from libbpf_rs instead to make importing code easier when using alternative build systems.


JakeHillion (approved):
simple.

etsal (approved):
Makes sense

PR #2761, opened 2025-09-11
scx_mitosis: add l3 awareness and work stealing

tommy-u:
Work in progress...

This PR introduces L3 cache domain awareness into scx_mitosis. Now, cell tasks are affinitized to a single L3 within that cell's cpuset. When a CPU executed `dispatch()`, it preferentially executes tasks affinitized to that L3. When a CPU is about to go idle, it first attempts to steal work from other L3 domains associated with the cell. Work stealing can be enabled or disabled at compile time. Because work stealing can paper over a number of scheduler issues, we preserve the option to disable work stealing for debugging purposes.

Performance testing showed <5% of scheduling decisions resulted in steals (and a task staying on a core does not count as a scheduling decision).

Unchanged:
- A cell's cpuset is still dictated by the cpuset of the cgroup it corresponds to
- There are still per CPU DSQs for use by pinned system threads

BPF Side Changes:
- Cell DSQs are no more; they are replaced by per-cell-per-L3 DSQs. 
-- There are MAX_CELLS x MAX_L3S many of these in the scheduler. 
- Cell vtime is no longer tracked, it is tracked by these per-cell-per-L3 DSQs
- Cell tasks are randomly affinitized to an L3 weighted by the number of CPUs the cell has in that L3. 
-- The chance of being affinitized to L3_i is (#cell cpus in L3_i / #cell cpus) 
- The `dispatch()` path preferentially pulls work from cell tasks within its L3
- Work stealing: In our model, a given CPU is associated with exactly one cell. Before a CPU goes idle (empty local, per-CPU, and per-cell-per-L3 DSQs), it first checks all of the other L3s of the cell for queued work. If it finds any queued work, it steals it from that per-cell-per-L3 DSQ and begins execution.
- Steal counts and timstamps are collected for future rate limiting
- Formalized 32-bit DSQ layout with typed unions + helpers replacing ad-hoc bit math
- Organization, code factored into multiple headers: dsq.bpf.h, l3_aware.bpf.h, mitosis.bpf.h

Rust Side Changes:
- Populating maps with CPU -> L3 mappings and the inverse
- More and more configurable debugging counters
- Organization, creating mitosis_topology_utils.rs

TODOs:
Clean up all TODOs
Performance testing
Run formatters
READ_ONCE and WRITE_ONCE on all shared variables
Tighten up sloppy races in work stealing

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/intf.h
+++ b/scheds/rust/scx_mitosis/src/bpf/intf.h
@@ -20,9 +20,7 @@ typedef _Bool bool;
 #endif
 
 /* ---- Work stealing config (compile-time) ------------------------------- */
-#ifndef MITOSIS_ENABLE_STEALING
 #define MITOSIS_ENABLE_STEALING 1
-#endif
 /* ----------------------------------------------------------------------- */
 
 enum consts {
--- a/scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
+++ b/scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
@@ -20,9 +20,7 @@ typedef u32 l3_id_t;
 // Configure how aggressively we steal work.
 // When task is detected as a steal candidate, skip it this many times
 // On a web server workload, 100 reduced steal count by ~90%
-#ifdef MITOSIS_ENABLE_STEALING
 #define PREVENT_N_STEALS 0
-#endif
 
 /* Work stealing statistics map - accessible from both BPF and userspace */
 struct steal_stats_map {
@@ -213,7 +211,7 @@ static inline s32 pick_l3_for_task(u32 cell_id)
 	return ret;
 }
 
-#ifdef MITOSIS_ENABLE_STEALING
+#if MITOSIS_ENABLE_STEALING
 
 static inline bool try_stealing_this_task(struct task_ctx *task_ctx,
 					  s32 local_l3, u64 candidate_dsq)
@@ -280,7 +278,7 @@ static inline bool try_stealing_work(u32 cell, s32 local_l3)
 
 		// Optimization: skip if faster than constructing an iterator
 		// Not redundant with later checking if task found (race)
-		if (scx_bpf_dsq_nr_queued(candidate_dsq))
+		if (!scx_bpf_dsq_nr_queued(candidate_dsq))
 			continue;
 
 		// Just a trick for peeking the head element
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -1625,11 +1625,6 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(mitosis_init)
 
 void BPF_STRUCT_OPS(mitosis_exit, struct scx_exit_info *ei)
 {
-	// int i;
-	// bpf_for(i, 0, MAX_CELLS); {
-	// 	dump_cell_state((u32)i);
-	// }
-
 	UEI_RECORD(uei, ei);
 }
 


scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
```diff
@@ -0,0 +1,174 @@
...
+}
+
+static inline u64 get_cell_l3_dsq_id(u32 cell, u32 l3)
+{
+	if (cell >= MAX_CELLS || l3 >= MAX_L3S)
+		scx_bpf_error("cell %u or l3 %u too large\n", cell, l3);
+	dsq_id_t dsq_id = { .cell_l3_dsq = { .l3 = l3, .cell = cell, .type = DSQ_TYPE_CELL_L3 } };
+
+	return dsq_id.raw;
+}
```
dschatzberg, line 201: I love all of this. Super nice defensive coding and encapsulation via interfaces and types.

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -18,6 +18,12 @@ typedef _Bool bool;
 #include <scx/ravg.bpf.h>
 #endif
 
+/* ---- Work stealing config (compile-time) ------------------------------- */
```
dschatzberg, line 22: I think it might be best to have this as a runtime option - e.g. a flag passed to the user space binary that writes to a global static variable in the bpf code before running it

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -39,6 +45,14 @@ enum cell_stat_idx {
 	NR_CSTATS,
 };
 
+/* Function invocation counters */
+enum counter_idx {
```
dschatzberg: Maybe make this type name a bit more descriptive? `fn_counter_idx` or so

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
```diff
@@ -0,0 +1,150 @@
...
+ * statistics for cross-L3 task migrations.
+ */
+#pragma once
+
+#include "mitosis.bpf.h"
+#include "intf.h"
+
+// It's also an option to just compute this from the cpu_to_l3 map.
+struct l3_cpu_mask {
+	unsigned long cpumask[CPUMASK_LONG_ENTRIES];
```
dschatzberg: Is there a reason not to use struct cpumask? I guess this is populated from userspace or something?

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
```diff
@@ -0,0 +1,150 @@
...
+static inline const struct cpumask *lookup_l3_cpumask(u32 l3)
+{
+	struct l3_cpu_mask *mask;
+
+	if (!(mask = bpf_map_lookup_elem(&l3_to_cpus, &l3))) {
+		scx_bpf_error("no l3 cpumask, l3: %d, %p", l3, &l3_to_cpus);
+		return NULL;
+	}
+
+	return (const struct cpumask *)mask;
```
dschatzberg: Hmm, I don't think you'll be able to pass these to the various bpf cpumask helpers because it won't be a "trusted" type due to this cast. I wonder if you're better off storing it as an actual cpumask in the map?

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/* Recompute cell->l3_cpu_cnt[] after cell cpumask changes (no persistent kptrs). */
```
dschatzberg: What does "no persistent kptrs" mean?

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
```diff
@@ -0,0 +1,150 @@
...
+	{
+		const struct cpumask *l3_mask =
+			lookup_l3_cpumask(l3); // plain map memory
+		if (!l3_mask) {
+			cell->l3_cpu_cnt[l3] = 0;
+			continue;
+		}
+
+		/* ok: dst is bpf_cpumask*, sources are (RCU cpumask*, plain cpumask*) */
+		bpf_cpumask_and(tmp, cell_mask, l3_mask);
```
dschatzberg: Hmm, I'm surprised this verifies - @kkdwivedi my understanding is that this l3_mask type is untrusted and so I would have expected bpf_cpumask_and to require an RCU ptr.

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * Weighted random selection of an L3 cache domain for a task.
+ *
+ * Uses the CPU count in each L3 domain within the cell as weights to
+ * probabilistically select an L3. L3 domains with more CPUs in the cell
+ * have higher probability of being selected.
+ *
+ * @cell_id: The cell ID to select an L3 from
+ * @return: L3 ID on success, INVALID_L3_ID on error, or 0 as fallback
```
dschatzberg: I didn't see where you ever return 0 here?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
```diff
@@ -0,0 +1,134 @@
...
+	/* (tasks cpumask anded with its cell cpumask) */
+	struct bpf_cpumask __kptr *cpumask;
+	/* started_running_at for recording runtime */
+	u64 started_running_at;
+	u64 basis_vtime;
+	/* For the sake of monitoring, each task is owned by a cell */
+	u32 cell;
+	/* For the sake of scheduling, a task is exclusively owned by either a cell
+	 * or a cpu */
+	u64 dsq;
```
dschatzberg: You can probably change this to your new dsq type

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -301,11 +281,24 @@ static inline int update_task_cpumask(struct task_struct *p,
 	if (!tctx->cpumask)
 		return -EINVAL;
 
+	/*
+	 * Calculate the intersection of CPUs that are both:
+	 * 1. In this task's assigned cell (cell_cpumask)
+	 * 2. Allowed by the task's CPU affinity (p->cpus_ptr)
+	 * Store result in tctx->cpumask - this becomes the effective CPU set
+	 * where this task can actually run.
+	 */
```
dschatzberg, line 288: Naively, I would have thought you'd make this restricted to the particular L3 to preserve affinity

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -317,16 +310,56 @@ static inline int update_task_cpumask(struct task_struct *p,
 	 * Revisit if high frequency dynamic cell switching
 	 * needs to be supported.
 	 */
+
+	// We want to set the task vtime to that of the cell it's joining.
+	// This used to be done by looking up the cell's dsq
+	// but now each cell has potentially multiple per l3 dsqs.
```
dschatzberg: I'd avoid comments about historical status of the code - this comment is probably going to live longer than our memories of how it used to be which is going to provide limited usefulness.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -317,16 +310,56 @@ static inline int update_task_cpumask(struct task_struct *p,
...
 	if (tctx->all_cell_cpus_allowed) {
-		tctx->dsq = cell_dsq(tctx->cell);
-		if (!(cell = lookup_cell(tctx->cell)))
+
+		const struct cpumask *l3_mask = NULL;
+		if (tctx->l3 != INVALID_L3_ID) {
+			l3_mask = lookup_l3_cpumask((u32)tctx->l3);
+			/* If the L3 no longer intersects the cell's cpumask, invalidate it */
```
dschatzberg, line 317: Hmm, this made me wonder why not just always invalidate the cpumask. I think most of the cases are probably all just fine to invalidate - e.g. cpumask of the task changed, task init, cgroup move, etc. The only case where it's dumb to invalidate is when the call here is superfluous (the cell cpumask hasn't changed).

Then again, I think in the long run, we'd expect the cell "cpumask" to change dynamically and you wouldn't want to just throw tasks around randomly but rather do some load balancing.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Oh, I think you do that later

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -317,16 +310,56 @@ static inline int update_task_cpumask(struct task_struct *p,
...
+				return -ENODEV;
+			tctx->l3 = new_l3;
+			l3_mask = lookup_l3_cpumask((u32)tctx->l3);
+			if (!l3_mask)
+				return -ENOENT;
+		}
+
+		/* --- Narrow the effective cpumask by the chosen L3 --- */
+		/* tctx->cpumask already contains (task_affinity ∧ cell_mask) */
+		if (tctx->cpumask)
```
dschatzberg, line 338: I think copying this tctx->cpumask to a local variable (just the pointer itself) and then validating that it's non null once should be sufficient and avoids the need to keep checking it.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -317,16 +310,56 @@ static inline int update_task_cpumask(struct task_struct *p,
...
+			bpf_cpumask_and(tctx->cpumask, (const struct cpumask *)tctx->cpumask, l3_mask);
+
+		/* If empty after intersection, nothing can run here */
+		if (tctx->cpumask && bpf_cpumask_empty((const struct cpumask *)tctx->cpumask))
+			return -ENODEV;
```
dschatzberg, line 347: This should probably give an scx_bpf_error(), right? We should only be in this logic if we're not all_cell_cpus_allowed

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -317,16 +310,56 @@ static inline int update_task_cpumask(struct task_struct *p,
...
+		/* --- Point to the correct (cell,L3) DSQ and set vtime baseline --- */
+		tctx->dsq = get_cell_l3_dsq_id(tctx->cell, tctx->l3);
+
+		struct cell *cell = lookup_cell(tctx->cell);
+		if (!cell)
 			return -ENOENT;
-		p->scx.dsq_vtime = READ_ONCE(cell->vtime_now);
+
+		if (tctx->l3 < 0 || tctx->l3 >= MAX_L3S)
+			return -EINVAL;
```
dschatzberg, line 361: Also, scx_bpf_error this

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -563,71 +615,107 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
-	bpf_for_each(scx_dsq, p, cell, 0) {
-		min_vtime = p->scx.dsq_vtime;
-		min_vtime_dsq = cell;
-		found = true;
-		break;
+
+	// Get L3
+	u32 cpu_key = (u32)cpu;
+	u32 *l3_ptr = bpf_map_lookup_elem(&cpu_to_l3, &cpu_key);
+	s32 l3 = l3_ptr ? (s32)*l3_ptr : INVALID_L3_ID;
```
dschatzberg: I think we should store this in the cpu_ctx to avoid having to do another lookup

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -563,71 +615,107 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
-	 */
-
-	if (!scx_bpf_dsq_move_to_local(min_vtime_dsq) && min_vtime_dsq != dsq)
-		scx_bpf_dsq_move_to_local(dsq);
+	* The move_to_local can fail if we raced with some other cpu in the cell
+	* and now the cell is empty. We have to ensure to try the cpu_dsq or else
+	* we might never wakeup.
+	*/
+	// TODO: The upstream side has "&& min_vtime_dsq != dsq" as part of this condition.
+	// Do we care?
```
dschatzberg: If min_vtime_dsq == local_dsq then we're just calling move to local twice in a row on the same dsq. Not going to break anything but totally unnecessary.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -563,71 +615,107 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+	if (!scx_bpf_dsq_move_to_local(min_vtime_dsq)) {
+#if MITOSIS_ENABLE_STEALING
+		/* Dead-simple work stealing:
+		 * If our local choices are empty, scan sibling (cell,L3) DSQs in the
+		 * same cell and steal the head task if it can run on @cpu.
+		 * No thresholds/cooldowns/lag heuristics—just the first eligible head.
+		 */
```
dschatzberg: Would be great to get feedback from @kkdwivedi about what work stealing strategies have worked well here. This one seems particularly aggressive at stealing instead of trying to preserve locality.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -563,71 +615,107 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+		bool moved = false;
+		if (l3 != INVALID_L3_ID) {
+			// TODO: This math is kinda dumb and confusing.
+			u32 start = ((u32)l3 + 1) % nr_l3;
+			u32 off;
+			// TODO: This might try a bunch of L3s outside of the cell
```
dschatzberg: Yeah, I think this should still stay within the cell

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -563,71 +615,107 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+							/* Retag to thief L3 */
+							qt->pending_l3 = l3;
+						}
+						/* Increment steal counter in map */
+						u32 key = 0;
+						u64 *count = bpf_map_lookup_elem(&steal_stats, &key);
+						// NOTE: This could get expensive, but I'm not
+						// anticipating that many steals. Percpu if we care.
+						if (count)
+							__sync_fetch_and_add(count, 1);
```
dschatzberg: I think if you make this all stay within the cell, you can just treat this as another per-cell counter like we do with CSTAT_LOCAL, etc.

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
```diff
@@ -0,0 +1,150 @@
...
+		u32 cnt = bpf_cpumask_weight((const struct cpumask *)tmp);
+		cell->l3_cpu_cnt[l3] = cnt;
+		total_cpus += cnt;
+		if (cnt)
+			present++;
+	}
+	bpf_rcu_read_unlock();
+
+	cell->l3_present_cnt = present;
+	cell->cpu_cnt = total_cpus;
```
dschatzberg, line 150: This logic runs off tick() and it is concurrent with select_cpu, task_init, etc. It's not obvious to me that all the concurrency here is safe. You might even need to protect all of this with a per-cell spinlock or rwsem. At the very least explain how you reason about the safety of these writes vs the below reads

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1017,12 +1125,44 @@ void BPF_STRUCT_OPS(mitosis_running, struct task_struct *p)
...
 	/*
-	 * Update both the CPU's cell and the cpu's vtime so the vtime's are
-	 * comparable at dispatch time.
+	 * If this task was stolen across L3s, retag to thief L3 and recompute
+	 * effective cpumask+DSQ. Preserve vtime to keep fairness.
 	 */
-	if (time_before(READ_ONCE(cell->vtime_now), p->scx.dsq_vtime))
-		WRITE_ONCE(cell->vtime_now, p->scx.dsq_vtime);
+#if MITOSIS_ENABLE_STEALING
+	if (tctx->pending_l3 >= 0 && tctx->pending_l3 < MAX_L3S) {
```
dschatzberg: I'd probably create a "validate_l3()" helper that outputs an scx_bpf_error() if outside this range - you end up doing this conditional a lot and often without an error output but wrapping it in a helper will make it easier.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1057,6 +1197,18 @@ void BPF_STRUCT_OPS(mitosis_stopping, struct task_struct *p, bool runnable)
...
+		 * For cell-schedulable tasks, also accumulate vtime into
+		 * per-cell per-L3 queues
+		 */
+		if (tctx->all_cell_cpus_allowed && tctx->l3 >= 0 &&
+		    tctx->l3 < MAX_L3S) {
+			/* Accumulate weighted execution time into per-(cell, L3) vtime */
+			cell->l3_vtime_now[tctx->l3] +=
+				used * DEFAULT_WEIGHT_MULTIPLIER /
+				p->scx.weight;
+		}
```
dschatzberg, line 1200: Do you really need this? It seems like advancing vtime at running() should be sufficient.

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
tommy-u: Nice, looks like that works as a drop in

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
tommy-u: Nothing interesting, going to delete. Had a version before that didn't use bpf_cpumask_create

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
tommy-u: I think the deal is `l3_mask` is safe because it points to persistent map storage (array type). If it were in a hash map for instance, it could be relocated and you'd need to use bpf_rcu_read_lock. That sound right?

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
tommy-u: dropped that

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: deleted

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: Ya, this is unnecessarily confusing. It does not try any DSQs outside of the cell. But it does try the cell's DSQs outside its cpuset which we know are empty. I'll clean it up

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
tommy-u: Hmm, I tried that, but then I had to cast it or use .raw every time we pass it to a bpf helper function. I don't hate it, but I wasn't sure I wanted others to have to think about it.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: done

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
dschatzberg: Yeah, you're right. It's not trusted but it is RCU protected which is all you need for the bpf cpumask functions (because its a densely packed bitfield, this is safe).

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
dschatzberg: Hmm, without having tried it, I don't mind the .raw use myself, but whatever you decide

tommy-u: Went with your suggestion here

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: I added a configurable knob to stealing here which can reduce the number of times a task is stolen.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: I cleaned this up

scheds/rust/scx_mitosis/src/bpf/l3_aware.bpf.h
tommy-u: I added per-cell spinlocks, which took some refactoring because each cell entry now had to be in its own element of a map. I locked around recalc_cell_l3_counts(), pick_l3_for_task(), and allocate_cell(). I would have liked to user rw locks, but it appears bpf only supports spin locks.

htejun:

PR #2763, opened 2025-09-11
scx_rustland_core: Re-enqueue tasks in cpu_release

EricccTaiwan:
Add scx_bpf_reenqueue_local() to rustland_cpu_release() so that when a higher scheduler class steals the CPU, all tasks waiting on the local DSQ are re-enqueued and given a chance to run on other CPUs.

This patch significantly reduces max wakeup latency, while request latency and RPS remain within statistical variance.

Test plan:
- terminal#1: $ sudo cyclictest -a 0 -p 80 -t1	# RT task
- terminal#2: $ ./schbench -m 4 -t 4 -r 10

Before:
```
Wakeup Latencies percentiles (usec) runtime 10 (s) (59403 total samples)
	  50.0th: 3          (31803 samples)
	  90.0th: 3          (0 samples)
	* 99.0th: 10         (2984 samples)
	  99.9th: 17         (472 samples)
	  min=1, max=3445
Request Latencies percentiles (usec) runtime 10 (s) (59407 total samples)
	  50.0th: 2148       (16596 samples)
	  90.0th: 3588       (29323 samples)
	* 99.0th: 3588       (0 samples)
	  99.9th: 3596       (326 samples)
	  min=2030, max=7028
RPS percentiles (requests) runtime 10 (s) (11 total samples)
	  20.0th: 5928       (4 samples)
	* 50.0th: 5944       (6 samples)
	  90.0th: 5944       (0 samples)
	  min=5927, max=5954
sched delay: message 0 (usec) worker 0 (usec)
current rps: 5945.68
```

After:
```
Wakeup Latencies percentiles (usec) runtime 10 (s) (59174 total samples)
	  50.0th: 3          (32279 samples)
	  90.0th: 3          (0 samples)
	* 99.0th: 4          (1822 samples)
	  99.9th: 8          (242 samples)
	  min=1, max=19
Request Latencies percentiles (usec) runtime 10 (s) (59186 total samples)
	  50.0th: 2156       (15702 samples)
	  90.0th: 3588       (28923 samples)
	* 99.0th: 3588       (0 samples)
	  99.9th: 4664       (148 samples)
	  min=2033, max=8868
RPS percentiles (requests) runtime 10 (s) (11 total samples)
	  20.0th: 5896       (3 samples)
	* 50.0th: 5912       (3 samples)
	  90.0th: 5928       (5 samples)
	  min=5899, max=5932
sched delay: message 0 (usec) worker 0 (usec)
current rps: 5931.68
```

Ref: scheds/rust/scx_cosmos/src/bpf/main.bpf.c

Suggested-by: Andrea Righi <arighi@nvidia.com>

Edit: Fix grammar

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -833,7 +833,7 @@ void BPF_STRUCT_OPS(rustland_enqueue, struct task_struct *p, u64 enq_flags)
 	 */
 	if ((is_kthread(p) && p->nr_cpus_allowed == 1) || is_kswapd(p) || is_khugepaged(p)) {
 		cpu = scx_bpf_task_cpu(p);
-                scx_bpf_dsq_insert_vtime(p, cpu_to_dsq(cpu),
+		scx_bpf_dsq_insert_vtime(p, cpu_to_dsq(cpu),
 					 SCX_SLICE_DFL, p->scx.dsq_vtime, enq_flags);
 		__sync_fetch_and_add(&nr_kernel_dispatches, 1);
 		return;
@@ -841,9 +841,10 @@ void BPF_STRUCT_OPS(rustland_enqueue, struct task_struct *p, u64 enq_flags)
 
 	/*
 	 * Give the task a chance to be directly dispatched if
-	 * ops.select_cpu() was skipped.
+	 * ops.select_cpu() was skipped or if it has been re-enqueued due
+	 * to a higher scheduling class stealing the CPU.
 	 */
-	if (builtin_idle && is_queued_wakeup(p, enq_flags)) {
+	if (builtin_idle && (is_queued_wakeup(p, enq_flags) || (enq_flags & SCX_ENQ_REENQ))) {
 		bool dispatched = false;
 
 		cpu = try_direct_dispatch(p, scx_bpf_task_cpu(p), enq_flags, &dispatched);
@@ -1068,6 +1069,13 @@ void BPF_STRUCT_OPS(rustland_cpu_release, s32 cpu,
 	dbg_msg("cpu preemption: pid=%d (%s)", p->pid, p->comm);
 	if (is_usersched_task(p))
 		set_usersched_needed();
+
+	/*
+	 * A higher scheduler class stole the CPU, re-enqueue all the tasks
+	 * that are waiting on this CPU and give them a chance to pick
+	 * another idle CPU.
+	 */
+	scx_bpf_reenqueue_local();
 }
 
 /*


rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -118,6 +118,11 @@ const volatile bool builtin_idle;
  */
 const volatile bool smt_enabled = true;
 
+/*
+ * Default time slice.
+ */
+const volatile u64 slice_ns = 10000ULL;
```
arighi: We are already using `SCX_SLICE_DFL` as default time slice, why do we need to introduce another one?

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -807,6 +812,14 @@ static bool is_queued_wakeup(const struct task_struct *p, u64 enq_flags)
 	return !__COMPAT_is_enq_cpu_selected(enq_flags) && !scx_bpf_task_running(p);
 }
 
+/*
+ * Return a time slice scaled by the task's weight.
+ */
+static u64 task_slice(const struct task_struct *p)
```
arighi: I see that scaling the task slice by weight can be nice, but it should treated as a separate PR if we want to introduce such built-in logic.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -832,13 +846,25 @@ void BPF_STRUCT_OPS(rustland_enqueue, struct task_struct *p, u64 enq_flags)
...
 					 SCX_SLICE_DFL, p->scx.dsq_vtime, enq_flags);
 		__sync_fetch_and_add(&nr_kernel_dispatches, 1);
 		return;
 	}
 
+	/*
+	 * If the task is being re-enqueued, attempt to dispatch
+	 * directly on an idle CPU.
+	 */
+	if (builtin_idle && (enq_flags & SCX_ENQ_REENQ)) {
```
arighi: I would reuse `try_direct_dispatch()` that also has extra logic to prevent potential stalls, see below.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -832,13 +846,25 @@ void BPF_STRUCT_OPS(rustland_enqueue, struct task_struct *p, u64 enq_flags)
...
+		cpu = pick_idle_cpu(p, prev_cpu);
+		if (cpu >= 0) {
+			scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | cpu, task_slice(p), enq_flags);
+			scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
+			return;
+		}
+	}
+
```
arighi: I would change the condition below like this:
```
        /*
         * Give the task a chance to be directly dispatched if
         * ops.select_cpu() was skipped or if it has been re-enqueued due
         * to a higher scheduling class stealing the CPU.
         */
        if (builtin_idle && (is_queued_wakeup(p, enq_flags) || (enq_flags & SCX_ENQ_REENQ))) {
                bool dispatched = false;

                cpu = try_direct_dispatch(p, scx_bpf_task_cpu(p), enq_flags, &dispatched);
                if (dispatched) {
                        return;
                }
        }
```

In this way the "direct dispatch" path is triggered either in case of a queued wakeup (when `ops.select_cpu()` was skipped) or when a task has been re-enqueued. Either way, if we can't do direct dispatch because there are other tasks queued to the local or shared DSQs, or because no idle CPU is found, we just bounce the task to the user-space scheduler again that will take the appropriate scheduling decisions.

arighi:
Hm... actually, do you see any `SCX_ENQ_REENQ` at all? I've done some tests and the re-enqueue doesn't seem to work correctly. This might be a kernel bug, I'll investigate.

EricccTaiwan:
> Hm... actually, do you see any `SCX_ENQ_REENQ` at all? I've done some tests and the re-enqueue doesn't seem to work correctly. This might be a kernel bug, I'll investigate.

Yes, I used `bpf_printk` to check, and I do see the log.
```diff
	if (builtin_idle && (enq_flags & SCX_ENQ_REENQ)) {
+		bpf_printk("Re-enqueue: pid=%d (%s)", p->pid, p->comm);
		cpu = pick_idle_cpu(p, prev_cpu);
		if (cpu >= 0) {
			scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | cpu, task_slice(p), enq_flags);
			scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
			return;
		}
	}
```

https://drive.google.com/file/d/1etb90oanrmaNAF-VJOeyIAnu_EVwpp43/view?usp=drive_link

rust/scx_rustland_core/assets/bpf/main.bpf.c
EricccTaiwan: Oh, I didn't notice this, thanks!

rust/scx_rustland_core/assets/bpf/main.bpf.c
EricccTaiwan: Got it!

arighi:
Just for the records, as discussed offline, the reason why I'm not seeing re-enqueue events is related to this kernel change introduced in v6.15: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=97e13ecb026684dcd428910701662275175a2315

We might decide to revert the commit, here's a proposal with some details about the problem: https://lore.kernel.org/all/20250912161438.529446-1-arighi@nvidia.com/

The rest of my review, about the code changes, remains valid.

EricccTaiwan:
> Just for the records, as discussed offline, the reason why I'm not seeing re-enqueue events is related to this kernel change introduced in v6.15: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=97e13ecb026684dcd428910701662275175a2315
> 
> We might decide to revert the commit, here's a proposal with some details about the problem: https://lore.kernel.org/all/20250912161438.529446-1-arighi@nvidia.com/
> 
> The rest of my review, about the code changes, remains valid.

Since `v6.14` doesn’t include `97e13ecb02668` commit, would it still make sense to test this PR on `v6.14`? Or would it be better to use `v6.15` and add the above patch (or wait for it to be merged)?

rust/scx_rustland_core/assets/bpf/main.bpf.c
EricccTaiwan: Got it! I added the `(enq_flags & SCX_ENQ_REENQ)` check to the original "ops.select_cpu() skipped" logic.

arighi:
> Since `v6.14` doesn’t include `97e13ecb02668` commit, would it still make sense to test this PR on `v6.14`? Or would it be better to use `v6.15` and add the above patch (or wait for it to be merged)?

It's ok to test it on v6.14. Once patched the v6.15+ kernels will behave like v6.14 with the re-enqueued tasks.

arighi (approved):

arighi:
@EricccTaiwan patch looks good, so we could merge as it is already.

Just a minor comment about the commit message: I think the test isn't really meaningful for the change. What we should do instead, to better highlight the benefits of this change, is to run a CPU-intensive RT tasks together with a non-RT task, for example see the test that I did here: https://lore.kernel.org/all/20250912161438.529446-1-arighi@nvidia.com/

You can reuse the same test with v6.14, or maybe just remove the test & results from the commit message, thanks!

EricccTaiwan:
> Just a minor comment about the commit message: I think the test isn't really meaningful for the change. What we should do instead, to better highlight the benefits of this change, is to run a CPU-intensive RT tasks together with a non-RT task, for example see the test that I did here: https://lore.kernel.org/all/20250912161438.529446-1-arighi@nvidia.com/

Understood! But I tried the same experiment with/without my PR, and when I observed the behavior with `htop`, the results looked pretty much the same. I’m not really sure why that’s the case.

Without PR:
image

With PR:
image

arighi:
Alright, here's a better test, start scx_rustland in a v6.14 kernel using virtme-ng:
```
$ vng --ssh -v -r v6.14 --cpu 16 -m 4G
...
$ sudo ./target/release/scx_rustland
```

On another session run a single CPU hogger that can only run either on CPU0 or CPU1:
```
$ vng --ssh-client
$ schedtool -a 0,1, -e stress-ng -c 1
```

Wait for the CPU hogger to settle on a CPU (i.e., using 100% of CPU0).

On another session run an RT CPU hogger on the same CPU (i.e., CPU0). Use the 5s timeout to prevent hitting the sched_ext watchdog timeout (to prevent this we need the dl_server patch in the kernel, but that's another story):
```
$ vng --ssh-client
$ sudo timeout 5 schedtool -F -p 99 -a 0, -e stress-ng -c 1
```

Look at htop on another `vng --ssh-client` session.

Without this patch applied both tasks are running on CPU0:
```
    0[|||||||||||||||||||||||||||||100.0%]   8[                               0.0%]
    1[                               0.0%]   9[                               0.0%]
    2[                               0.0%]  10[                               0.0%]
    3[                               0.0%]  11[                               0.0%]
    4[                               0.0%]  12[                               0.0%]
    5[                               0.0%]  13[                               0.0%]
    6[                               0.0%]  14[                               0.0%]
    7[                               0.0%]  15[                               0.0%]
  Mem[||||                     226M/3.82G] Tasks: 35, 3 thr, 198 kthr; 7 running
  Swp[                              0K/0K] Load average: 4.65 1.34 0.47
```

With this patch applied the regular CPU hogger is correctly migrated to the other CPU:
```
    0[|||||||||||||||||||||||||||||100.0%]   8[                               0.0%]
    1[|||||||||||||||||||||||||||||100.0%]   9[                               0.0%]
    2[                               0.0%]  10[                               0.0%]
    3[|                              0.7%]  11[                               0.0%]
    4[                               0.0%]  12[                               0.0%]
    5[                               0.0%]  13[                               0.0%]
    6[                               0.0%]  14[                               0.0%]
    7[                               0.0%]  15[                               0.0%]
  Mem[||||                     218M/3.82G] Tasks: 28, 3 thr, 197 kthr; 7 running
  Swp[                              0K/0K] Load average: 4.49 1.19 0.41
                                           Uptime: 00:01:01
```

So, everything seems to work as expected. :+1:

EricccTaiwan:
Thanks, now I can reproduce it and better understand how my patch works!

I was a bit confused at first, but now it’s clearer. In normal use cases with `rustland` (like running Chrome or games), it’s quite rare to actually encounter higher scheduler class tasks, so the effect is hard to observe.

This PR essentially provides a safeguard: even if a higher scheduler class preempts a CPU, `scx_bpf_reenqueue_local()` makes sure tasks can migrate to another idle CPU, rather than sticking to the previously “idle” CPU (which is already occupied by the higher class).

PR #2772, opened 2025-09-15
Instructions for building and managing schedulers using cargo

sirlucjan:
Since the migration from Meson to Cargo is already a done deal, it is a good idea to add a relevant guide for testers and new project members.

The following guide contains:

- an explanation of what profiles are included in Cargo.toml and what their characteristics are

- how to build the entire project or just one part of it (taking into account the use of profiles)

- how to install a given scheduler/tool directly from crates.io

The guide will be gradually expanded if necessary.

### DIFF
--- a/CARGO_BUILD.md
+++ b/CARGO_BUILD.md
@@ -121,10 +121,45 @@ Besides schedulers, the workspace includes several tools:
 
 ---
 
-## 5. Summary
+## 5. Installing from crates.io
+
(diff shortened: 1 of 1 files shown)


arighi (approved):
Maybe add a reference to CARGO_BUILD.md in the main README.md under the section "Building and installing" for the Rust schedulers. 

Other than that LGTM.

PR #2776, opened 2025-09-16
scx_rustland_core: Return prev_cpu if no idle CPU

EricccTaiwan:
If no idle CPU is available, fall back to prev_cpu to ensure a valid execution target and prevent errors.

Ref: scx/scheds/rust/scx_bpfland/src/bpf/main.bpf.c

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -539,14 +539,15 @@ static s32 pick_idle_cpu(const struct task_struct *p, s32 prev_cpu)
 	if (cpu >= 0)
 		goto out_put_cpumask;
 
-	/*
-	 * If all the previous attempts have failed, dispatch the task to the
-	 * first CPU that will become available.
-	 */
-	cpu = -EBUSY;
-
 out_put_cpumask:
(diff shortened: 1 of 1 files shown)


EricccTaiwan:
I discussed this with @ianchen0119 offline. The previous approach (returning `-EBUSY`) gave more flexibility in scheduling, e.g., choosing a CPU that shares cache with `prev_cpu` but is less busy. This PR instead directly returns `prev_cpu`, which I took from the `cpu_release` implementation in bpfland.

That said, I’ve been thinking: if we are exposing this as a public API, would this PR still be useful?

EricccTaiwan:
Oh, I see. So you think it’s better to keep returning `-EBUSY`?
At least that way we can still tell when no CPU is idle.

arighi:
> Oh, I see. So you think it’s better to keep returning `-EBUSY`? At least that way we can still tell when no CPU is idle.

Right, I think returning `-EBUSY` gives more information. Another approach could be to return a CPU and set a flag true/false to determine if the returned CPU was idle, or if it was just a "suggestion". But based on how the framework is currently designed, it'd always return `prev_cpu` in this case with `is_idle == false` and this information can already be determined by the caller when it gets `-EBUSY`...

EricccTaiwan:
Got it, thanks for your explanation!

PR #2779, opened 2025-09-16
lib/scx_layered: add raw PMU tracking module and integrate with scx_layered membw tracking

etsal:
Add a library for tracking and reading raw counting counter PMUs. These PMUs are read by a tracepoint on every context switch and aggregated per-task on a task map. Schedulers can read the task map to track the value of the PMU over time.

Use the library from scx_layered to back its memory barandwidth throttling logic. The scheduler turns on PMU tracking if any tasks have memory bandwidth throttling enabled.

Also adjust the scx_raw_pmu crate to make searching for performance counters easier: Add the codename of the uarch as a separate field, and organize the JSON definitions into a HashMap indexable by the name of the counter.


etsal:
Still debugging this, but the overall structure of the patch is stable. Have left multiple counter support as a followup - most of the code already supports it, but haven't found an elegant way to define mulitple perf event arrays.

kkdwivedi (approved):

PR #2780, opened 2025-09-16
scx_raw_pmu: use include_dir to embed json

JakeHillion:
scx_raw_pmu needs a bunch of JSON files. Currently these are expected to be distributed with the crate/scheduler that uses them, but this is annoying.

Use the `include_dir!` macro to package all of these resources into the crate's library, such that they can be used directly in other crates. Add an abstraction `ResourceDir` which wraps both this `include_dir::Dir` method and a regular Filesystem method for cases where the user wants to override. It's not clear this is needed, but it's less of an API break.

Test plan:
- CI
- Enabled the existing test of the filesystem based method by extracting the included files into a tempdir.
- Added a new similar test that uses the included files from memory.

### DIFF
--- a/rust/scx_raw_pmu/src/json.rs
+++ b/rust/scx_raw_pmu/src/json.rs
@@ -1,18 +1,19 @@
 use anyhow::bail;
 use csv::Reader;
 use procfs::CpuInfo;
-use procfs::FromBufRead;
 use regex::Regex;
 use serde::Deserialize;
 use serde::Deserializer;
 use serde::Serialize;
 use std::env;
 use std::env::consts::ARCH;
+use std::ffi::OsString;
 use std::fs;
 use std::fs::File;
 use std::io::BufReader;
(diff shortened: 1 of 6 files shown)


rust/scx_raw_pmu/src/json.rs
```diff
@@ -211,60 +188,99 @@ impl PMUManager {
...
+
+        let mut pmus = HashMap::new();
+        for file in spec_dir.files()? {
+            // metricgroups.json isn't actually PMU definitions.
+            if file.path().ends_with("metricgroups.json") {
+                continue;
+            }
 
-        let jsondir = dataroot.join(&tuple).join(&arch).join(&path);
+            println!("Reading {}", file.path());
```
etsal: Can you remove this, has been removed from the repo

JakeHillion: Whoops, thanks for catching it!

PR #2785, opened 2025-09-17
scx_rustland_core: Append ops.version suffix

EricccTaiwan:
Ref: scx/rust/scx_utils/src/compat.rs

Before:
image

After:
image


arighi (approved):
Looks good, thanks!

PR #2791, opened 2025-09-18
scx_rustland/rlfifo: Upgrade procfs version

EricccTaiwan:


arighi (approved):

EricccTaiwan:
~Also, update the procfs version in scxtop.~

Never mind, I’ll handle this later, gotta do some cleanup first.

JakeHillion:
> ~Also, update the procfs version in scxtop.~
> 
> 
> 
> Never mind, I’ll handle this later, gotta do some cleanup first.

We were holding back the procfs version in scxtop and some other crates because it makes packaging a bit easier for us at Meta. Is there a good reason to upgrade them?

EricccTaiwan:
> We were holding back the procfs version in scxtop and some other crates because it makes packaging a bit easier for us at Meta. Is there a good reason to upgrade them?

Nah, I only did the upgrade because I noticed `scx_raw_pmu` was already on `0.18`, so I figured it might be a good time to bump the version here as well. Thanks for explaining the background though!

PR #2793, opened 2025-09-18
scx_rustland_core/alloc.rs: Sync with upstream

EricccTaiwan:
Sync with the latest upstream updates and remove unused code.

Ref: https://github.com/jjyr/buddy-alloc

### DIFF
--- a/rust/scx_rustland_core/src/alloc.rs
+++ b/rust/scx_rustland_core/src/alloc.rs
@@ -107,21 +107,25 @@ struct Node {
 impl Node {
     fn init(list: *mut Node) {
         unsafe {
-            (*list).next = list;
-            (*list).prev = list;
+            list.write(Node {
+                next: list,
+                prev: list,
+            });
         }
     }
 
     fn remove(list: *mut Node) {
         unsafe {
-            (*(*list).prev).next = (*list).next;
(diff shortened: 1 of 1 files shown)


arighi (changes requested):
Hm... no, please. We shouldn't blindly sync with upstream, our use case is quite special and it may introduce subtle bugs. We should only apply clear fixes with a proper description of what they're fixing. And cleanups are also fine, but again, with a clear description of what they're cleaning up.

So, if we can split this PR in multiple commits each one describing the change that it's adding or the cleanup, that'd be fine.

Otherwise it might be problematic to bisect and identify potential issues.

EricccTaiwan:
I see.

I noticed that some of the issues you mentioned earlier seem to have been fixed. I’ll take a closer look at those and send separate commits accordingly.

arighi:
> I see.
> 
> I noticed that some of the issues you mentioned earlier seem to have been fixed. I’ll take a closer look at those and send separate commits accordingly.

Maybe check also `stress-ng --race-sched 0` and `stress-ng -c 0` while running https://webglsamples.org/aquarium/aquarium.html and see if the changes are affecting the fps. This is a good test to quickly detect performance regressions. Thanks!

htejun:

PR #2797, opened 2025-09-19
scx_rustland_core: Keep return style consistent

EricccTaiwan:
Return the actual error code in usersched_timer_fn for consistency, though scx_bpf_error() already forces scheduler exit.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -1168,7 +1168,7 @@ static int usersched_timer_fn(void *map, int *key, struct bpf_timer *timer)
 	if (err)
 		scx_bpf_error("Failed to arm stats timer");
 
-	return 0;
+	return err;
 }
 
 /*


EricccTaiwan:
I don't get it, why did this fail? :no_mouth:

JakeHillion:
> I don't get it, why did this fail? 😶

You have to return 0 in BPF timer callbacks.

EricccTaiwan:
> > I don't get it, why did this fail? 😶
> 
> You have to return 0 in BPF timer callbacks.

So we have to return 0 explicitly? Since if there’s no error, `err` would be 0 anyway.

JakeHillion:
> > You have to return 0 in BPF timer callbacks.
> 
> So we have to return 0 explicitly? Since if there’s no error, `err` would be 0 anyway.

I don't believe the verifier can track that `scx_bpf_error` is noreturn (actually, I think it does return, even though it involves the scheduler being kicked soon after). So that path can still occur.

PR #2798, opened 2025-09-19
scx_mitosis: Use RAII style for releasing resources

dschatzberg:
Much of the logic was pretty confusing with many gotos and various release_cpumask/group/rcu lock, etc. This creates a few macros for common types and uses the gcc/clang __attribute__(cleanup, ...) to do the proper destruction of the resource once the variable goes out of scope.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -21,6 +21,8 @@
 #include <scx/common.bpf.h>
 #endif
 
+#include "guard.h"
+
 char _license[] SEC("license") = "GPL";
 
 /*
@@ -152,9 +154,8 @@ static inline struct cgroup *task_cgroup(struct task_struct *p)
 		 *
 		 * p->cgroups is RCU-protected, so we need RCU lock.
 		 */
-		bpf_rcu_read_lock();
+		guard(rcu)();
 		cgrp = bpf_cgroup_acquire(p->cgroups->dfl_cgrp);
-		bpf_rcu_read_unlock();
 	}
 
 	if (!cgrp)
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -483,11 +516,9 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu,
 	 */
 	if (!bpf_cpumask_test_cpu(prev_cpu, cast_mask(tctx->cpumask)) &&
 	    tctx->cpumask)
-		cpu = bpf_cpumask_any_distribute(cast_mask(tctx->cpumask));
-	else
-		cpu = prev_cpu;
+		return bpf_cpumask_any_distribute(cast_mask(tctx->cpumask));
```
tommy-u: [nit] I know the `tctx->cpumask` part of the condition above this line is redundant with the check before it & probably just there for the verifier. But it still looks weird to see the check in that order bc the short circuit implications of:

`!bpf_cpumask_test_cpu(prev_cpu, cast_mask(tctx->cpumask)) && tctx->cpumask`
vs
` tctx->cpumask && !bpf_cpumask_test_cpu(prev_cpu, cast_mask(tctx->cpumask))`

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -402,44 +442,37 @@ static s32 pick_idle_cpu_from(struct task_struct *p,
...
 static __always_inline s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu,
 					 struct cpu_ctx *cctx,
 					 struct task_ctx *tctx)
 {
 	struct cpumask *task_cpumask;
-	const struct cpumask *idle_smtmask;
-	s32 cpu;
+	DECLARE_IDLE_CPUMASK(idle_smtmask) = scx_bpf_get_idle_smtmask();
 
 	if (!(task_cpumask = (struct cpumask *)tctx->cpumask) ||
```
tommy-u: Know this is outside the diff, but the bpf_cpumask is larger than the task_cpumask. I think we'll be fine using it as a pointer as long as we never e.g. copy size of bpf_cpumask into a cpumask 

https://elixir.bootlin.com/linux/v6.16.8/source/kernel/bpf/cpumask.c#L25

htejun:
Can we copy the guard() macros from the kernel and use them? They're pretty much the same things and it'd be nice to have some commonality.

dschatzberg:
I wasn't aware of the guard() macros, let me look those up

htejun:

PR #2805, opened 2025-09-21
scx_rustland_core: Use dbg_msg instead of scx_bpf_error in cpu_to_dsq()

EricccTaiwan:
Since cpu_to_dsq() should return SHARED_DSQ on invalid CPUs, using scx_bpf_error() would abort the scheduler. Switch to dbg_msg() to warn without exiting.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -372,7 +372,7 @@ static bool usersched_has_pending_tasks(void)
 static u64 cpu_to_dsq(s32 cpu)
 {
 	if (cpu < 0 || cpu >= MAX_CPUS) {
-		scx_bpf_error("Invalid cpu: %d", cpu);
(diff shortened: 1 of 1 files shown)


EricccTaiwan:
> Hm... NACK, I think we actually want to error here, otherwise we would hide potential errors in the scheduler.
> 
> We're returning `SHARED_DSQ` just as a temporary "valid" DSQ so that we can reliably exit reporting the "invalid cpu" error.

But doesn’t `scx_bpf_error` request the scheduler to exit? In that case, is returning `SHARED_DSQ` still valid?

arighi:
> > Hm... NACK, I think we actually want to error here, otherwise we would hide potential errors in the scheduler.
> > We're returning `SHARED_DSQ` just as a temporary "valid" DSQ so that we can reliably exit reporting the "invalid cpu" error.
> 
> But doesn’t `scx_bpf_error` request the scheduler to exit? In that case, is returning `SHARED_DSQ` still valid?

Yes, but the scheduler doesn't exit immediately, the exit is not perfectly sync.

PR #2811, opened 2025-09-23
scx_layered: add TTL for layer membership

etsal:
Layer membership for scx_layered is based on rules that are checked during tCertain layer membership conditions may change over the course of the scheduler's runtime. For example, the GPU matcher marks tasks that use NVIDIA API calls as "GPU tasks", but there are workloads that use GPUs once and never again. We want to mark some layers as having a Time-To-Live for its members, where layer membership is re-evaluated after a user-defined number of schedules. Add this per-layer TTL and check it during the .stopping() call to periodically trigger match_layer() for those tasks.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2460,7 +2460,6 @@ static __noinline bool match_one(struct layer_match *match,
 	}
 	case MATCH_IS_KTHREAD:
 		return p->flags & PF_KTHREAD;
-	case MATCH_USED_GPU_TID: {
 			u32 tid;
 			bool pid_present = false;
 
@@ -2473,20 +2472,23 @@ static __noinline bool match_one(struct layer_match *match,
 				pid_present = true;
 
 			return pid_present == match->used_gpu_tid;
-	}
+	case MATCH_USED_GPU_TID:
 	case MATCH_USED_GPU_PID: {
-			u32 tgid;
-			bool pid_present = false;
+			u32 key = (match->kind == MATCH_USED_GPU_TID) ? p->pid : p->tgid;
+			bool used = (match->kind == MATCH_USED_GPU_TID) ? match->used_gpu_tid : match->used_gpu_pid;
+			u32 *present;
 
 			if (!enable_gpu_support)
 				scx_bpf_error("UsedGpuPid requires --enable_gpu_support");
 
-			tgid = p->tgid;
+			if (match->kind == MATCH_USED_GPU_TID)
+				present = bpf_map_lookup_elem(&gpu_tid, &key);
+			else
+				present = bpf_map_lookup_elem(&gpu_tgid, &key);
 
-			if (bpf_map_lookup_elem(&gpu_tgid, &tgid))
-				pid_present = true;
+			/* If not found, we want the used predicate to be required false. */
 
-			return pid_present == match->used_gpu_pid;
+			return (present != NULL) == used;
 	}
 	case MATCH_AVG_RUNTIME: {
 			struct task_ctx *taskc = lookup_task_ctx_may_fail(p);
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -706,23 +706,35 @@ int save_gpu_tgid_pid(void) {
 	pid_tgid = bpf_get_current_pid_tgid();
 	pid = pid_tgid >> 32;
 	tid = pid_tgid;
-	// To avoid race condition where the GPU process is discovered before kprobe is called
-	// If we discover new GPU tid, force task refresh_layer to match_layer again
+
 	if ((p = (struct task_struct*)bpf_get_current_task_btf()) && p) {
 		if ((taskc = lookup_task_ctx_may_fail(p)) && taskc) {
+			/* If it doesn't exist, recheck. If the timestamp is */
 			if(!bpf_map_lookup_elem(&gpu_tid, &tid)) {
 				trace("New GPU tid: %d, force to refresh layer", tid);
 				taskc->refresh_layer = true;
 			}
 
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -588,6 +588,8 @@ struct task_ctx {
 
 	char 			join_layer[SCXCMD_COMLEN];
 	u64			layer_refresh_seq;
+
+	u64			ttl_max;
```
htejun: Maybe just `ttl`? Given that this is just a decrementing counter, `_max` part seems like a misnomer.

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2919,6 +2922,13 @@ void BPF_STRUCT_OPS(layered_stopping, struct task_struct *p, bool runnable)
 		((RUNTIME_DECAY_FACTOR - 1) * taskc->runtime_avg + runtime) /
 		RUNTIME_DECAY_FACTOR;
 
+	/* If TTL matching is on, decrement TTL and force matching refresh if applicable. */
+	if (taskc->ttl_max) {
+		taskc->ttl_max -= 1;
+		if (!taskc->ttl_max)
+			taskc->refresh_layer = true;
```
htejun: Would it be better to make this based on time instead of the number of stopping events? Stopping event frequency can vary widely, so it can be a bit difficult to predict the behavior and configure. Using time wouldn't make it noticeably more expensive either. We can set valid_until timestamp on layer match to be `now + ttl` and trigger refresh from a scheduling event which already reads the current time.

etsal: Sounds good, in that case I'll add the checks in .stopping() and .tick() since they already call scx_bpf_now() and change the argument to milliseconds.

etsal:
@kkdwivedi the new version adds an scx_bpf_now() call in the hinting function is that ok?

etsal:
Added timestamp recording for GPU tasks

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1757,6 +1761,19 @@ void BPF_STRUCT_OPS(layered_enqueue, struct task_struct *p, u64 enq_flags)
 	}
 }
 
+static inline void ttl_tick(struct task_ctx *taskc, u64 now)
+{
+	/* If TTL matching is on, decrement TTL and force matching refresh if applicable. */
+	if (taskc->member_expire == MEMBER_NOEXPIRE)
+		return;
+
+	/* Use GiB to make the division cheap. */
```
xerothermic: This comment is for something else?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -588,6 +588,8 @@ struct task_ctx {
 
 	char 			join_layer[SCXCMD_COMLEN];
 	u64			layer_refresh_seq;
+
+	u64			member_expire;
```
xerothermic: just a thought, will `member_expire_at` be more explicit about the intent of this variable?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2456,20 +2474,38 @@ static __noinline bool match_one(struct layer_match *match,
 				pid_present = true;
 
 			return pid_present == match->used_gpu_tid;
```
xerothermic: stale code? please fix

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2456,20 +2474,38 @@ static __noinline bool match_one(struct layer_match *match,
...
+			if (match->kind == MATCH_USED_GPU_TID)
+				timestamp = bpf_map_lookup_elem(&gpu_tid, &key);
+			else
+				timestamp = bpf_map_lookup_elem(&gpu_tgid, &key);
 
-			if (bpf_map_lookup_elem(&gpu_tgid, &tgid))
-				pid_present = true;
+			/* If not found, we want the used predicate to be required false. */
+			if (!timestamp)
+				return !used;
```
xerothermic: I'm not sure how does this condition work? Can you please elaborate?

etsal: If we don't have an entry for the task in the GPU map, the matching rule should say "is not GPU task". We compute used above.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: Yeah that was for a calculation from ns to ms that was removed.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: sure, can do

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -246,15 +246,15 @@ struct {
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__type(key, u32);
-	__type(value, u32);
+	__type(value, u64);
```
htejun, line 257: Add a comment explaining how the map and its values are being used?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1775,6 +1792,8 @@ static void account_used(struct cpu_ctx *cpuc, struct task_ctx *taskc, u64 now,
 	cpuc->used_at = now;
 	cpuc->usage += used;
 
+	ttl_tick(taskc, now);
```
htejun: Maybe a bit of misnomer now? Would something like `check_member_expired()` work better?

htejun (approved):

xerothermic (approved):
Thanks for the great work!

PR #2815, opened 2025-09-24
scx_lavd: Improve the calculation of performance criticality threshold

multics69:
Previously, only active CPUs were considered when calculating the performance criticality threshold, which is used to determine if a task should be assigned to a big core or a little core. However, in reality, some cores are on anyway due to pinned tasks. Let's embrace this reality and consider the computing capacity of overflow CPUs, too.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/power.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/power.bpf.c
@@ -251,7 +251,7 @@ int do_core_compaction(void)
 	struct cpdom_ctx *cpdomc;
 	int nr_active, cpu, i;
 	u32 sum_capacity = 0, big_capacity = 0, nr_active_cpdoms = 0;
-	bool need_kick;
+	bool need_kick, cpu_on;
 	u64 cpdom_id;
 
 	bpf_rcu_read_lock();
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_lavd/src/bpf/power.bpf.c
```diff
@@ -328,6 +324,7 @@ int do_core_compaction(void)
 				 * add this CPU{ to the overflow set.
 				 */
 				bpf_cpumask_set_cpu(cpu, ovrflw);
+				cpu_on = true;
```
arighi, line 327: Do we need to test `scx_bpf_dsq_nr_queued(cpu_to_dsq(cpu)) + scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | cpu)` here?

multics69: No. The intention of the current design is to keep using the overflow CPU for one full interval to avoid CPU frequency fluctuation.

multics69: Thanks @arighi for the review!

PR #2816, opened 2025-09-24
cargo: add xtask crate for repo scripts

JakeHillion:
`cargo xtask` [0] is a pattern for running arbitrary commands in your Cargo workspace. You create a crate for `xtask` and add an alias for `cargo run -p xtask --` as `cargo xtask`. This means we can run "scripts" like `cargo xtask bump-versions --all` with only Cargo installed.

Add this pattern and a first couple of useful scripts to demonstrate the pattern. First we have `versions`, which emulates pretty closely the existing `version-tool.py` (with no arguments). Then we have `bump-versions`, which allows you to pass one or more crates and bumps their patch versions like we do for a full sched_ext release. Intermediate bumps are handled by the crate authors directly.

This will let us add more useful features but are a good base. After this lands I want to add more formats for `versions`, and in the future enable `bump-versions` to not bump crates that are identical to their published version (there are specific cases where this is useful).

Test plan:

- CI
- `cargo xtask bump-versions -p scx_utils` bumps scx_utils and all of its dependencies that are in the workspace. It then updates every reference to scx_utils everywhere.
- `cargo xtask bump-versions --all` bumps everything in the repo and all of its internal usages. This is what we do as part of the monthly release process.

```sh
jake@rooster:/data/users/jake/repos/scx/ > cargo xtask versions
warning: /data/users/jake/repos/scx/rust/scx_rustland_core/Cargo.toml: unused manifest key: lib.include
warning: /data/users/jake/repos/scx/Cargo.toml: unused manifest key: profile.release-fast.target-cpu
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
     Running `/home/jake/cargo_builds/scx/debug/xtask versions`
12:14:04 [WARN] [/data/users/jake/repos/scx/rust/arenalib/arenalib/Cargo.toml:16] crate "scx_utils" 1.0.19 mismatches existing 1.0.22
...
12:14:04 [WARN] [/data/users/jake/repos/scx/tools/xtask/Cargo.toml:15] crate "log" 0.4 mismatches existing 0.4.17
{
  "00-versions": {
    "meson": "1.0.16"
  },
  "01-rust-versions": {
    "arenalib": "0.1.1",
    "arenalib_selftests": "1.0.3",
    "scx_bpf_compat": "1.0.16",
...
    "scxtop": "1.0.18",
    "vmlinux_docify": "0.1.5",
    "xtask": "0.1.0"
  },
  "02-rust-deps": {
    "affinity": "0.1",
...
    "zvariant": "5.1"
  }
}
```

[0] https://github.com/matklad/cargo-xtask

### DIFF
--- a/tools/xtask/src/versions.rs
+++ b/tools/xtask/src/versions.rs
@@ -1,0 +1,276 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+//
+// This software may be used and distributed according to the terms of the
+// GNU General Public License version 2.
+
+use anyhow::Context;
+use anyhow::Result;
+use regex::Regex;
+use serde::Serialize;
+
+use std::collections::HashMap;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+use crate::get_cargo_metadata;
+use crate::get_rust_paths;
+
+pub fn version_command() -> anyhow::Result<()> {
+    #[derive(Serialize)]
+    struct Output {
+        rust_versions: HashMap<String, String>,
+        rust_dep_versions: HashMap<String, String>,
+    }
+
+    let mut output = Output {
+        rust_versions: HashMap::new(),
+        rust_dep_versions: HashMap::new(),
+    };
+
+    // Get rust paths and process versions
+    let rust_paths = get_rust_paths().context("failed to get rust paths")?;
+
+    for path in &rust_paths {
+        if let Some((name, version)) = process_rust_version(path)
+            .with_context(|| format!("failed to process rust version for {}", path.display()))?
+        {
+            output.rust_versions.insert(name, version);
+        }
+    }
+
+    // Include tree crates as deps
+    output.rust_dep_versions.extend(
+        output
+            .rust_versions
+            .iter()
+            .map(|(k, v)| (k.clone(), v.clone())),
+    );
+
+    // Process rust dependencies
+    for path in &rust_paths {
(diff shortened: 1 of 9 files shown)


tools/xtask/src/versions.rs
```diff
@@ -0,0 +1,312 @@
...
+
+pub fn version_command() -> anyhow::Result<()> {
+    let workspace_root = crate::get_workspace_root()?;
+
+    let mut versions = HashMap::new();
+    let mut rust_versions = HashMap::new();
+    let mut rust_deps = HashMap::new();
+
+    // Process meson version
+    if let Err(e) = process_meson_version(workspace_root.as_path(), &mut versions) {
```
arighi: Should we just ignore meson at this point?

arighi (approved):
Left a small comment. This looks useful, thanks!

tools/xtask/src/versions.rs
JakeHillion: Yeah, let's ditch it. I was going to keep this as the v1 format matching `version-tool.py` output, but given we (maybe I?) are the only consumers of this script, let's break the format now. I'll rename the entries to not include the numbers too.

JakeHillion:
The Rust dependency collection is currently done through parsing Cargo.toml files instead of from `cargo metadata` output which is somewhat flaky, can be fixed later though.

PR #2818, opened 2025-09-24
scx_rustland: fix min_vruntime in update_enqueue

charliechiou:
The scheduler was previously updating min_vruntime using the maximum vruntime. Fix the logic to consistently track the minimum observed vruntime as the baseline. Improve the consistency of fairness semantics.

### DIFF
--- a/scheds/rust/scx_rustland/src/main.rs
+++ b/scheds/rust/scx_rustland/src/main.rs
@@ -247,13 +247,13 @@ impl<'a> Scheduler<'a> {
     // This method implements the main task ordering logic of the scheduler.
     fn update_enqueued(&mut self, task: &mut QueuedTask) -> u64 {
         // Update global minimum vruntime based on the previous task's vruntime.
-        if self.min_vruntime < task.vtime {
+        if self.min_vruntime > task.vtime {
             self.min_vruntime = task.vtime;
         }
 
         // Update task's vruntime re-aligning it to min_vruntime (never allow a task to accumulate
         // a budget of more than a time slice to prevent starvation).
-        let min_vruntime = self.min_vruntime.saturating_sub(self.slice_ns);
+        let min_vruntime = self.min_vruntime;
         if task.vtime == 0 {
             // Slightly penalize new tasks by charging an extra time slice to prevent bursts of such
             // tasks from disrupting the responsiveness of already running ones.


scheds/rust/scx_rustland/src/main.rs
```diff
@@ -247,13 +247,13 @@ impl<'a> Scheduler<'a> {
     // This method implements the main task ordering logic of the scheduler.
     fn update_enqueued(&mut self, task: &mut QueuedTask) -> u64 {
         // Update global minimum vruntime based on the previous task's vruntime.
-        if self.min_vruntime < task.vtime {
+        if self.min_vruntime > task.vtime {
```
arighi, line 250: Actually we want to move min_vruntime forward.

You can see `self.min_vruntime`  as a clock that is always moving forward, and the "current time" is determined by the task that is enqueued. Ideally we would like to update the `min_vruntime` using the vtime of the task that is running on a CPU, but we don't a have direct callback from `ops.running()` to the Rust part of the scheduler, so this is an approximation.

Maybe we should clarify better this concept renaming `self.min_vruntime` to `self.vruntime_now` or something similar...

scheds/rust/scx_rustland/src/main.rs
```diff
@@ -247,13 +247,13 @@ impl<'a> Scheduler<'a> {
...
             self.min_vruntime = task.vtime;
         }
 
         // Update task's vruntime re-aligning it to min_vruntime (never allow a task to accumulate
         // a budget of more than a time slice to prevent starvation).
-        let min_vruntime = self.min_vruntime.saturating_sub(self.slice_ns);
+        let min_vruntime = self.min_vruntime;
```
arighi, line 256: In this case we want to give some "buffer" to the vruntime, basically a sleeping task can accumulate a vruntime credit up to `self.slice_ns`. If we remove this "buffer" the scheduler becomes similar to a FIFO...

scheds/rust/scx_rustland/src/main.rs
charliechiou: Got it, thanks for clarifying!
I'll update the code by renaming `self.min_vruntime` to `self.vruntime_now` to better reflect the conceptual.

scheds/rust/scx_rustland/src/main.rs
charliechiou: Understood — I’ll keep the buffer here. Thanks!

charliechiou:
I have made further changes in #2822 , therefore I will close this PR.

PR #2823, opened 2025-09-25
versions: add starlark output

JakeHillion:
Quality of life improvement allowing us to emit version metadata in the Starlark format. This can be consumed by the Bazel & BUCK build systems wherever that metadata is needed.

Test plan:
- Ran it and used in a BUCK build. It works!


likewhatevs (approved):
wild, but cool.

Issue #2825, opened 2025-09-25
MAP_TYPE_QUEUE->LOCAL_DSQ insert and scx_central


cloehle:
scx_central is broken (reproducing while running hackbench, every 2-3 minutes of running):

```
Xorg[341392] triggered exit kind 1024:
  runtime error (SCX_DSQ_LOCAL[_ON] cannot move migration disabled Xorg[341392] from CPU 5 to 0)

Backtrace:
  scx_exit+0x58/0x84
  task_can_run_on_remote_rq+0x16c/0x1a0
  dispatch_to_local_dsq+0x7c/0x204
  flush_dispatch_buf+0x208/0x218
  balance_scx+0x254/0x490
  __schedule+0x540/0xf6c
  schedule_idle+0x28/0x48
  do_idle+0x194/0x28c
  cpu_startup_entry+0x38/0x3c
  rest_init+0xfc/0x18c
  start_kernel+0x720/0x7e8
  __primary_switched+0x88/0x90

CPU states
----------

CPU 0   : nr_run=2 flags=0x3 cpu_rel=0 ops_qseq=17776131 pnt_seq=29451
          curr=Xorg[341392] class=ext_sched_class
  idle_to_kick   : 040

 *R Xorg[341392] +0ms
      scx_state/flags=3/0x5 dsq_flags=0x0 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=(n/a)
      dsq_vtime=0 slice=0 weight=100
      cpus=fff

    scx_dump_state+0x7d0/0x924
    scx_error_irq_workfn+0x4c/0x68
    irq_work_single+0x6c/0xa8
    irq_work_run_list+0x4c/0x68
    irq_work_run+0x38/0x5c
    ipi_handler+0x20c/0x334
    handle_percpu_devid_irq+0xc0/0x208
    handle_irq_desc+0x40/0x58
    generic_handle_domain_irq+0x1c/0x28
    gic_handle_irq+0x4c/0x11c
    call_on_irq_stack+0x30/0x48
    do_interrupt_handler+0xd4/0xd8
    el1_interrupt+0x34/0x64
    el1h_64_irq_handler+0x18/0x24
    el1h_64_irq+0x6c/0x70
    finish_task_switch.isra.0+0xbc/0x2c8
    el1_interrupt+0x4c/0x64
    el1h_64_irq_handler+0x18/0x24
    el1h_64_irq+0x6c/0x70
[...]

PR #2842, opened 2025-09-30
scx_qmap: Handle migration-disabled tasks in ops.enqueue()

arighi:
When dispatching a task to a remote DSQ in ops.enqueue(), we must ensure the task is allowed to migrate.

Failing to do so can trigger errors such as:
```
  EXIT: runtime error (SCX_DSQ_LOCAL[_ON] cannot move migration disabled Xorg[341392] from CPU 5 to 0)
```
This fixes issue #2825.

Reported-by: Christian Loehle <christian.loehle@arm.com>

### DIFF
--- a/scheds/c/scx_qmap.bpf.c
+++ b/scheds/c/scx_qmap.bpf.c
@@ -231,7 +231,7 @@ void BPF_STRUCT_OPS(qmap_enqueue, struct task_struct *p, u64 enq_flags)
 	}
 
 	/* if select_cpu() wasn't called, try direct dispatch */
-	if (!__COMPAT_is_enq_cpu_selected(enq_flags) &&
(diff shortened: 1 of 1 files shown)


cloehle:
Isn't this a noop?
scx_qmap doesn't set SCX_OPS_ENQ_MIGRATION_DISABLED so enqueue() shouldn't be called if p has migration disabled?

arighi:
> Isn't this a noop? scx_qmap doesn't set SCX_OPS_ENQ_MIGRATION_DISABLED so enqueue() shouldn't be called if p has migration disabled?

That seems right... but now I'm wondering how you're triggering this error with qmap...

Edit: ohhh I see, that's because qmap is bouncing some tasks in ops.dispatch() with the `HIGHPRI_DSQ`. Oh well, then this is not fixing anything. Let's close it.

PR #2844, opened 2025-09-30
nix: update rust toolchain

JakeHillion:
Update the locked Nix toolchain from Rust 1.88.0->1.90.0. This brings the ability to run `cargo publish --workspace --dry-run`, which seems to replace our `cargo-publish.py` script - it runs a topological sort of the crates, and even better it allows you to dry-run the whole repo to ensure you can publish atomically.

Test plan:
- CI
- `cargo publish --workspace --dry-run` - it fails horribly :)

### DIFF
--- a/.github/include/flake.lock
+++ b/.github/include/flake.lock
@@ -8,11 +8,11 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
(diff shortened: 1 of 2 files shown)


JakeHillion:
Hmm, this is failing way more than expected. Will come back when it actually works.

PR #2852, opened 2025-10-05
scx_utils: fix clippy error.

devnexen:
```
error: you are deriving `Hash` but have implemented `PartialEq` explicitly
   --> rust/scx_utils/src/energy_model.rs:27:28
    |
 27 | #[derive(Debug, Clone, Eq, Hash, Ord, PartialOrd)]
```


JakeHillion (changes requested):
Why is this the correct fix over deriving PartialEq? It seems to make the code more complex and more likely to get out of sync.

devnexen:
It s not a required fix but I thought you would prefer to keep the PartialEq implementation (which seems to be just the same as the default behavior anyway).

htejun:
This changes how equality is tested for `PerfState`. Before, `inefficient` was not included in equality testing. After, all fields are tested. It looks like the omission wasn't intentional and this actually is a bug fix. @multics69, can you please confirm? @devnexen can you please update the commit title and description? The fact that clippy pointed this out is not that interesting. Please describe the actual changes and rationale.

htejun:
Closing from lack of activity.

PR #2854, opened 2025-10-05
scx_rustland_core: Introduce QueuedTask:comm_str()

arighi:
Add a helper to QueuedTask to retrieve the task's command name.

Having direct access to the executable name from user space can be useful for classifying workloads based on the application or task name.t


EricccTaiwan (approved):
LGTM! 😃

PR #2855, opened 2025-10-06
scx_loader: set scx_cosmos as suggested scheduler

sirlucjan:
CachyOS users have repeatedly suggested that scx_cosmos is currently the most efficient scheduler and should be the one suggested in the scx_loader configuration.


arighi (approved):
Makes sense to me, scx_cosmos seems to perform pretty well both with server and desktop workloads right now and has positive user feedback. We can always switch the default to another scheduler if a better option emerges and in the meantime we can experiment new ideas in the other schedulers.

PR #2857, opened 2025-10-06
nix: move nixpkgs input from JakeHillion to upstream nixpkgs

JakeHillion:
Currently Nix pulls its nixpkgs input from JakeHillion/nixpkgs/virtme-ng where I maintain a virtme-ng package. This does work and was done with the intention of upstreaming this commit, but is a pain to maintain - I have to push a merge commit to that branch to update nixpkgs in sched_ext and it's tricky to do atomically.

Grab the virtme-ng package from that branch and vendor it here. This allows the normal `nix flake update` to work without requiring changes in a different GitHub repository.

Some refactoring here too, the nix flake is getting a bit too big. Add a `.nix/pkgs` directory to store the few package definitions. Will move the flake.nix and flake.lock there as a follow up, it's a bigger change.

Test plan:
- CI


hodgesds (approved):

JohnRTitor:
@JakeHillion can you make a PR on upstream nixpkgs and tag me there?

I wouldn't be using the package though, so you'll need to maintain it. However I can help merging the updates.

JakeHillion:
> @JakeHillion can you make a PR on upstream nixpkgs and tag me there?
> 
> I wouldn't be using the package though, so you'll need to maintain it. However I can help merging the updates.

Thanks! I haven't put it up yet because I want to add a feature to vng to make it useful on NixOS. Currently it's basically useless, you can't keep your path when invoking vng (the feature I need) and you can't run `nix develop` in the vng VM. The intention is still to upstream this, but I'm a bit bandwidth constrained at the moment.

PR #2866, opened 2025-10-07
versions: bump versions for 1.0.17 release

JakeHillion:


sirlucjan (approved):

JakeHillion:
For future version cutters: `cargo publish --workspace` is firmly not up to the task. `cargo publish --workspace --dry-run` is very useful, but it doesn't verify constraints that crates.io has that are only verified when you publish. When `cargo publish --workspace` breaks in the middle it is nowhere near as easy to pick back up as our existing `cargo-publish.py`.

`cargo xtask bump-versions --all` however worked excellently. Will migrate `cargo-publish.py` into Rust too before the next cycle.

PR #2875, opened 2025-10-09
scx_rustland_core: Rely on built-in idle CPU selection policy

arighi:
Use the built-in idle CPU selection policy (when builtin_idle is enabled) and prioritize CPUs based on their capacity: on task wakeup, if the waker's CPU has higher capacity than the wakee's and they're on different LLCs, attempt to migrate the wakee closer to the waker's CPU.

This naturally favors higher-capacity CPUs on hybrid architectures, providing better performance overall.

User-space schedulers based on scx_rustland_core can still override the default policy setting builtin_idle to false.

### DIFF
--- a/rust/scx_rustland_core/assets/bpf/main.bpf.c
+++ b/rust/scx_rustland_core/assets/bpf/main.bpf.c
@@ -34,6 +34,7 @@
 #include <scx/common.bpf.h>
 #endif
 
+#include <scx/percpu.bpf.h>
 #include "intf.h"
 
 char _license[] SEC("license") = "GPL";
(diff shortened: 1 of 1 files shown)


rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -919,6 +850,25 @@ static long handle_dispatched_task(struct bpf_dynptr *dynptr, void *context)
...
+	if (!is_task_queued(p))
+		return false;
+
+	/*
+	 * If the task can only run on this CPU, keep it running.
+	 */
+	if (is_pcpu_task(p))
+		return true;
+
+	return true;
```
EricccTaiwan: Should this be `return false`?

arighi: Actually it's ok to return true here, it's the `is_pcpu_task()` check that is redundant. Thanks for noticing!

arighi:
@EricccTaiwan commit updated. But I'm afraid that we're stuck in a situation where scx_rustland is failing in the CI for who knows what reason... I'm wondering if we should just disable the stress tests in the CI for scx_rustland

JakeHillion:
> But I'm afraid that we're stuck in a situation where scx_rustland is failing in the CI for who knows what reason... I'm wondering if we should just disable the stress tests in the CI for scx_rustland

I think this might be a real stall, but the weirdest thing is it seems to be rustland specific. rlfifo is being tested in the same way and passing. Very strange considering only rustland_core changed. Not sure how we debug this further though, does the equivalent test pass locally?

Disabling rustland tests is easy though, just remove this line (it'll still get veristat):

https://github.com/sched-ext/scx/blob/2808f23752ea7deec2919542a7d8377abc4e8bed/.github/include/list-integration-tests.py#L132

arighi:
> > But I'm afraid that we're stuck in a situation where scx_rustland is failing in the CI for who knows what reason... I'm wondering if we should just disable the stress tests in the CI for scx_rustland
> 
> I think this might be a real stall, but the weirdest thing is it seems to be rustland specific. rlfifo is being tested in the same way and passing. Very strange considering only rustland_core changed. Not sure how we debug this further though, does the equivalent test pass locally?

Oh yes, it is a real stall indeed, I'm just saying with user-space scheduling debugging these stalls is 100x more difficult, so it's probably going to take a while (and multiple iterations) to figure out a fix...

arighi:
Looks like CI is finally happy. Sorry for the spam of failures. @EricccTaiwan if you have time and you wanna take a look I won't push any other changes to this PR. Everything seems to work fine on my side.

rust/scx_rustland_core/assets/bpf/main.bpf.c
```diff
@@ -382,14 +384,9 @@ static u64 cpu_to_dsq(s32 cpu)
  * to handle these mistakes in favor of a more efficient response and a reduced
  * scheduling overhead.
  */
-static s32 pick_idle_cpu(const struct task_struct *p, s32 prev_cpu)
+static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags)
```
EricccTaiwan, line 387: Overall LGTM.

Just one question - I noticed that `pick_idle_cpu()` changed from `const struct task_struct *p` to `struct task_struct *p`, but it doesn’t seem like `*p` is modified.
I’m a bit curious - is there a specific reason const was dropped here?

arighi: Oh that's because `scx_bpf_select_cpu_dfl()` and `scx_bpf_select_cpu_and()` want a `struct task_struct *p`.

EricccTaiwan: Oh, I see. Thanks for explaination. Let's merge this :fire:

PR #2876, opened 2025-10-09
scx_bpfland: Rework idle selection and scheduling policy

arighi:
scx_bpfland hasn't been touched in a while, so let's modernize the scheduler a bit with the following changes:

- drop the custom idle CPU selection policy and rely on the built-in idle CPU selection (and remove related topology cmdline options),
-  CPU capacity awareness: on task wakeup attempt to migrate the wakee to the waker's CPU if the CPU is faster
-  always attempt to migrate to full-idle SMT cores,
- drop NUMA rebalancing (rebalance across nodes can now can happen only on task wakeup),
- use a constant time slice (scaled by task weight) and drop `--slice-min`,
- use all CPUs when running balanced_power energy profile in auto mode (and prioritize slow cores only in powersave mode),
-  keep track of task wakeup frequency and consider this metric to evaluate task's deadline,
- tasks that are spamming re-enqueues (due to instense read ops or explicitly calling sleep) are forced to stay on the local DSQ to reduce locking contention on the global DSQs (also introduce the new option `--no-sticky-tasks` to disable this behavior),
- reuse the same logic for epp domains from scx_flash,
- adjust default time slice values to more optimal settings after these changes.

Some results comparing bpfland before and after this change:
```
WebGL acquarium 15K (idle) # more is better:
 - bpfland-old: avg fps: 135
 - bpfland-new: avg fps: 143

WebGL acquarium 15K + stress-ng -c 0 # more is better:
 - bpfland-old: avg fps: 41
 - bpfland-new: avg fps: 65

WebGL acquarium 15K + stress-ng --sleep 0 --timeout 10s # more is better:
 - bpfland-old: avg fps:  0 (kaboom)
 - bpfland-new: avg fps: 55

> schbench -L -m 4 -M auto -t 64 -n 0 # more is better
 - bpfland-old: current rps:  412624.75
 - bpfland-new: current rps: 1510736.89

> sudo perf bench -f simple sched messaging -t -g 24 -l 6000 # less is better
 - bpfland-old: 8.653
 - bpfland-new: 4.565

> time vng -r v6.16 --cpu 4 -m 256M -- uname -r # less is better
 - bpfland-old: 3.85 secs
 - bpfland-new: 2.02 secs
```

These changes have been tested also by the CachyOS community with positive results.

### DIFF
--- a/tools/scx_loader/configuration.md
+++ b/tools/scx_loader/configuration.md
@@ -18,10 +18,17 @@ If no configuration file is found at any of these paths, `scx_loader` will use t
 The configuration file has the following structure:
 
 ```toml
-default_sched = "scx_flash"
+default_sched = "scx_cosmos"
 default_mode = "Auto"
 
-[scheds.scx_rustland]
+[scheds.scx_bpfland]
+auto_mode = []
+gaming_mode = []
+lowlatency_mode = ["-m", "performance", "-w"]
+powersave_mode = ["-s", "20000", "-m", "powersave", "-I", "100", "-t", "100"]
+server_mode = ["-s", "20000"]
+
+[scheds.scx_rusty]
 auto_mode = []
 gaming_mode = []
 lowlatency_mode = []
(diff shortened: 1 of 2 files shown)


arighi:
We didn't trigger any stall or crash with this one and literally every single benchmark is showing better results, so let's go ahead and merge this. :)

arighi:
@ColinIanKing if you're going to stress test scx_bpfland make sure this one is applied. Thanks!

PR #2880, opened 2025-10-10
scx_layered: add memory bandwidth-based layer sizing

etsal:
Implement the PMU reading and resizing logic stubs in scx_layered to enable memory bandwidth-based layer resizing. The code uses resctl and hardware performance counters to estimate the memory bandwidth consumption of each task in the system. The scheduler then uses layer resizing to control this bandwidth.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('sched_ext schedulers', 'c',
         default_options : ['c_std=gnu11'], # Use gnu11 as the C standard to compile vmlinux.h without errors.
-        version: '1.0.16',
+        version: '1.0.17',
         license: 'GPL-2.0',
         meson_version : '>= 1.2.0',)
 


lib/pmu.bpf.c
```diff
@@ -27,6 +27,7 @@ struct {
 struct scx_pmu_counters {
 	u64 start[SCX_MAX_PMU_COUNTERS];
 	u64 agg[SCX_MAX_PMU_COUNTERS];
+	bool switched;
```
xerothermic, line 30: Is this a debug switch?

lib/pmu.bpf.c
```diff
@@ -106,6 +112,9 @@ int scx_pmu_event_start(struct task_struct __arg_trusted *p)
 		if (ret)
 			return ret;
 
+		if (update)
+			cntrs->agg[idx] += value.counter - cntrs->start[idx];
+
 		/* Add the delta for this scheduling interval. */
 		cntrs->start[idx] = value.counter;
```
xerothermic:
```suggestion
		if (update) {
			/* Add the delta for this scheduling interval. */
			cntrs->agg[idx] += value.counter - cntrs->start[idx];
		}

		cntrs->start[idx] = value.counter;
```

lib/pmu.bpf.c
```diff
@@ -243,32 +252,44 @@ int scx_pmu_read(struct task_struct __arg_trusted *p, u64 event, u64 *value, boo
...
 		return ret;
 
 next:
-	next = bpf_task_from_pid(args->next_pid);
-	if (!next)
-		return -ENOENT;
+	if (!next->pid)
+		return 0;
+
+	return scx_pmu_event_start(next, false);
```
xerothermic, line 277:
```suggestion
	/* Skip update when there was no previous task to obtain delta */
	return scx_pmu_event_start(next, false);
```

xerothermic: Is the added comment accurate?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -499,8 +499,15 @@ __weak s32 refresh_cpumasks(u32 layer_id)
 			} else {
 				if (layer->kind == LAYER_KIND_OPEN)
 					cpuc->in_open_layers = false;
-				else if (cpuc->layer_id == layer_id)
+				else if (cpuc->layer_id == layer_id) {
 					cpuc->layer_id = MAX_LAYERS;
+					cpuc->in_open_layers = true;
```
xerothermic, line 508: This was the fix for the extra CPU utilization in the confined layer, right?

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -1819,10 +1842,16 @@ static inline void check_member_expired(struct task_ctx *taskc, u64 now)
...
+static void account_used(struct task_struct *p, struct cpu_ctx *cpuc, struct task_ctx *taskc, u64 now)
 {
 	s32 task_lid;
 	u64 used;
+	u64 bytes;
+
+	/* Try to get the memory bandwidth, actively ignore the error if we fail. */
+	if (scx_pmu_read(p, membw_event, &bytes, true))
+		bytes = 0;
+	bytes *= 64;
```
xerothermic, line 1881: I think this is assuming there are 64 bytes in a cache line? 
Maybe put a comment and re-visit if when we support architecture that has non-64 bytes cacheline.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -2635,28 +2693,22 @@ impl<'a> Scheduler<'a> {
     fn clamp_target_by_membw(
         &self,
         layer: &Layer,
-        membw_limit: usize,
-        last_membw: usize,
-        low: usize,
-        high: usize,
+        membw_limit: f64,
+        membw: f64,
+        curtarget: u64,
```
xerothermic, line 2767:
```suggestion
        cur_target: u64,
```

scheds/rust/scx_layered/src/main.rs
```diff
@@ -3813,6 +3886,7 @@ fn create_perf_fds(skel: &mut BpfSkel, event: u64) -> Result<()> {
     attr.type_ = perf::bindings::PERF_TYPE_RAW;
     attr.config = event;
     attr.sample_type = 0u64;
+    attr.__bindgen_anon_1.sample_period = 0u64;
     attr.set_disabled(0);
```
xerothermic, line 4045: This is the first time I see this usage. What is `__bindgen_anon_1`?

etsal: It's the unnamed union struct that includes sample_period, basically an unfortunate side-effect of the way we interface Rust and C.

lib/pmu.bpf.c
etsal: Not exactly debug but we're not seeing it in practice, it's a flag we set when we notice our perf counter is being switched in and out of the PMU.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: Part of it, this explicitly sets the flag because it wasn't being updated after the CPU was removed from the layer.

scheds/rust/scx_layered/src/bpf/main.bpf.c
etsal: Will do, do more recent CPUs have larger cache lines? I think I've never seen a value other than 64.

xerothermic: Some of the ARM machines have 128 byte cachelines.

etsal: Makes sense. Unfortunately I don't see an easy way to grab the current machine's cache line size from BPF, because it's set using #define so it's not in vmlinux.h. I think we have to create a new kfunc to futureproof this code, but can't do much about this right now.

PR #2887, opened 2025-10-13
scx_bpfland: Disable sticky tasks by default

arighi:
Some users have reported laggy behavior when sticky tasks are enabled, so disable this feature for now, as this scheduler is primarily intended for gaming and desktop workloads.


sirlucjan:
I've tested, everything was OK.

PR #2902, opened 2025-10-16
nix: support aarch64-linux in flake

JakeHillion:
This makes my dev environment work on ARM. Don't apply in the CI yet given last time it didn't work, but it should be enabled at some point to populate the cache.

Test plan:
- Loaded the dev environment on an aarch64 NixOS machine. Can then run `cargo build`.


hodgesds (approved):
LGTM

PR #2903, opened 2025-10-16
scx_userland: refreshing output option.

devnexen:


htejun:
Ditto, please post on the kernel side. Thanks.

PR #2904, opened 2025-10-17
mitosis: bugfixes

likewhatevs:
Fix bugs in mitosis.

The test script added passes on the last commit in this PR.

It fails due to various issues on commits in this PR. 

W/o this PR at all, mitosis will get kicked w/ `missing cell in cpu map`.

W/o drain commit, mitosis will get kicked w/ `No available cells to allocate`.

### DIFF
--- a/scheds/rust/scx_mitosis/test/cleanup_test_cgroups.sh
+++ b/scheds/rust/scx_mitosis/test/cleanup_test_cgroups.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+# Cleanup all test cgroups created by test scripts
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m'
+
+if [ "$EUID" -ne 0 ]; then
+    echo -e "${RED}Must run as root${NC}"
+    exit 1
+fi
+
+echo -e "${YELLOW}========================================${NC}"
+echo -e "${YELLOW}Cleaning up all test cgroups${NC}"
+echo -e "${YELLOW}========================================${NC}\n"
+
+# Find all test-related cgroups with pattern matching
+TEST_PATTERNS=(
+    "test_mitosis*"
+    "test_cell*"
+    "test_simple*"
+    "test_brutal*"
+    "test_absolute*"
+    "test_working*"
+    "test_reuse*"
+    "test_verify*"
+    "test_cycle*"
+    "test_16*"
+    "test_15*"
+    "test_*"
+    "scx_mitosis_test*"
+)
+
+CLEANED=0
+FAILED=0
+
+# First pass: Find all matching cgroups at root level
+ALL_TEST_CGROUPS=()
+for pattern in "${TEST_PATTERNS[@]}"; do
+    shopt -s nullglob  # Make glob return nothing if no matches
+    for cg in /sys/fs/cgroup/${pattern}; do
+        if [ -d "$cg" ]; then
+            ALL_TEST_CGROUPS+=("$cg")
+        fi
+    done
+    shopt -u nullglob
+done
+
+# Remove duplicates and sort
+UNIQUE_CGROUPS=($(printf '%s\n' "${ALL_TEST_CGROUPS[@]}" | sort -u))
+
+if [ ${#UNIQUE_CGROUPS[@]} -eq 0 ]; then
+    echo -e "${GREEN}No test cgroups found to clean up${NC}"
+    exit 0
+fi
+
+echo -e "${YELLOW}Found ${#UNIQUE_CGROUPS[@]} test cgroups to clean up${NC}\n"
+
+# Clean each cgroup
+for test_root in "${UNIQUE_CGROUPS[@]}"; do
+    if [ ! -d "$test_root" ]; then
+        continue
+    fi
+
+    echo -e "${YELLOW}Cleaning $(basename $test_root)...${NC}"
+
+    # Find all child cgroups, deepest first
+    if [ -d "$test_root" ]; then
+        find "$test_root" -mindepth 1 -type d 2>/dev/null | sort -r | while read -r cg; do
+            # Kill all processes in the cgroup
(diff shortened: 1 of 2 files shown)


likewhatevs:
Reasonably confident this also fixes:

```
WARNING: CPU: 42 PID: 3899049 at kernel/sched/ext.c:2177 scx_dsq_insert_commit+0xf1/0x120
```

which i encountered while testing.

dschatzberg:
> Implement cell draining when cpuset removed

Let's drop this commit from the PR - I don't think we've hit this issue except in synthetic cases. It also won't work because we reassign the cpus to other cells so there's no guarantee that those cells will get drained.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -357,14 +357,17 @@ static inline int update_task_cell(struct task_struct *p, struct task_ctx *tctx,
...
+	 * By reading in reverse order with a barrier between, we ensure that
+	 * if we see a new configuration_seq, we've seen all the data updates,
+	 * and if we see old configuration_seq, any stale data we read will be
+	 * corrected on the next refresh.
 	 */
-	tctx->configuration_seq = READ_ONCE(applied_configuration_seq);
-	barrier();
 	tctx->cell = cgc->cell;
+	barrier();
+	tctx->configuration_seq = READ_ONCE(applied_configuration_seq);
```
dschatzberg, line 444: This whole commit looks wrong to me. The ordering of reading the seq and then reading the data ensures that the data we saw is at least as new as the seq. If we do it the other way around we might read old data, and then read the new seq but never actually apply the data from the new seq.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -484,8 +484,9 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu,
 	 * All else failed, send it to the prev cpu (if that's valid), otherwise any
 	 * valid cpu.
 	 */
-	if (!bpf_cpumask_test_cpu(prev_cpu, cast_mask(tctx->cpumask)) &&
-	    tctx->cpumask)
+	if (bpf_cpumask_test_cpu(prev_cpu, cast_mask(tctx->cpumask)))
+		cpu = prev_cpu;
+	else if (tctx->cpumask)
```
dschatzberg: The commit message is misleading - I don't think this is a functional change at all, just a stylistic change. I think I prefer the existing code since there's one less branch.

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -469,7 +471,13 @@ impl<'a> Scheduler<'a> {
...
             let cell_idx = i as u32;
             let bpf_cell = cells[i];
-            let in_use = unsafe { std::ptr::read_volatile(&bpf_cell.in_use as *const u32) };
+            let in_use = unsafe {
+                let ptr = &bpf_cell.in_use as *const u32;
+                (ptr as *const std::sync::atomic::AtomicU32)
+                    .as_ref()
+                    .unwrap()
+                    .load(std::sync::atomic::Ordering::Acquire)
+            };
```
dschatzberg: Hmm, I don't actually know the rust rules here - I kind of suspect you technically need both volatile and atomic since the values are getting changed outside of the rust binary. I doubt there's any practical difference though. Might as well leave it as atomic

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1136,11 +1136,13 @@ s32 BPF_STRUCT_OPS(mitosis_cgroup_init, struct cgroup *cgrp,
...
-		scx_bpf_error("cgrp_ctx creation failed for cgid %llu",
-			      cgrp->kn->id);
-		return -ENOENT;
+	/*
+	 * Use lookup without CREATE since this is the exit path. If the cgroup
+	 * doesn't have storage, it's not a cell owner anyway.
+	 */
+	if (!(cgc = lookup_cgrp_ctx_fallible(cgrp))) {
+		/* No storage means no cell to free, just return */
+		return 0;
```
dschatzberg, line 1225: I think we can probably use the non-fallible version and scx_bpf_error here - the storage should exist. The change makes sense though

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -471,38 +471,43 @@ impl<'a> Scheduler<'a> {
...
+
+        // Cell 0 (root cell) always exists even if it has no CPUs temporarily
+        let cells_with_cpus: HashSet<u32> = cell_to_cpus.keys().copied().collect();
+        let mut active_cells = cells_with_cpus.clone();
+        active_cells.insert(0);
+
+        for cell_idx in &active_cells {
+            let cpus = cell_to_cpus.get(cell_idx).cloned().unwrap_or_else(|| {
+                // Only cell 0 can have no CPUs (when all CPUs assigned to other cells)
+                if *cell_idx == 0 {
```
dschatzberg: I don't think you need this check - if cell 0 has no CPUs everything is broken

likewhatevs: removed

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: done

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: dropped

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: dropped

PR #2905, opened 2025-10-17
scx_mitosis: Add debug event recording

dschatzberg:
We've seen incidents which indicate cgroup_init and init_task not following the expected ordering. This commit adds a flag that enables recording of such events to be output at dump time. This should make it easier to track down the issue.


likewhatevs:
will update the githook template to include clangd on mitosis once i update my local copy  #2908

PR #2908, opened 2025-10-17
update fmt githook template

likewhatevs:
update template pre-commit hook which runs format to account for new ci formatting rules (i.e. mitosis clangd linting).

you install this by symlinking or copying it to where it should be.


etsal (approved):
Makes sense, we should discuss adding linting by default to all C schedulers.

PR #2909, opened 2025-10-17
tools/xtask: Downgrade cargo_metadata

arighi:
The cargo_metadata 0.23 dependency breaks `cargo vendor` on Ubuntu 24.04, due to the usual error:
```
  feature `edition2024` is required
```
This prevents including all required crates in the source package.

Downgrade to cargo_metadata 0.19 in xtask to avoid maintaining a custom patch for Ubuntu Noble.

All the `cargo xtask` commands seem to be still functional after downgrading the dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -460,22 +460,13 @@ dependencies = [
 ]
 
 [[package]]
-name = "cargo-platform"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "122ec45a44b270afd1402f351b782c676b173e3c3fb28d86ff7ebfb4d86a4ee4"
-dependencies = [
- "serde",
-]
-
-[[package]]
 name = "cargo_metadata"
 version = "0.18.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


JakeHillion:
Could you exclude this crate instead? It's not a dependency of any of the schedulers & would be convenient if we could stay flexible on dependencies as it's for repository management tools, CI, etc.

arighi:
Unfortunately `cargo vendor` doesn't support `--exclude`, it just grabs everything. The only way is to maintain a custom patch in `debian/patches`, which is not a big deal, but I wish there was a better way....

JakeHillion:
Ahh that sucks. We're hitting the problem we discussed offline recently: I just pulled the latest version in to Meta to build this tool, and now we need a different version for Ubuntu. As this downgrade is source compatible though I think it'll be fine.

arighi:
Oh I see, no worries, if it's a problem I can manage this via debian/patches for now. Ignore this PR,

arighi:
Sometimes I wish we had a separate git repo just with the scx schedulers and have all the tooling etc. in another one. The fact that `cargo vendor` grabs everything is quite annoying for the packaging. Maybe there's a way to split them via `Cargo.toml`?

sirlucjan:
What about separate ```[workspace]``` for it? But we've to drop meson first.

PR #2912, opened 2025-10-17
scx_lavd: Add configurable migration threshold with avg utilization

daidavid:
The current utilization based migration scheme look back period is too short and results in overly aggressive load balancing. For example, in certain workloads on a machine that has an average utilization at ~90%, sampling load deltas between compute domains still showed > 50% differences in load.

Allow users to define some threshold for task stealing using sum of util_avg of CPUs on a compute domain instead to smooth out load tracking and reduce overall task migrations.

These are sampled on a ~90% workload to measure the cost of cross CCX/LLC domain migrations.

bpftop sample:
+-----------------+--------------+---------------+
| lavd_dispatch   | forced_steal | no_task_steal |
+-----------------+--------------+---------------+
| avg_runtime(ns) | 3593         | 762           |
| cputime(%)      | 250.99       | 40.87         |
| events/s        | 704720       | 746275        |
+-----------------+--------------+---------------+

latencies(ns):
+-----+---------------+--------------------+----------------------+
|     | dispatch(avg) | consume(local_llc) | consume(remote_llc)  |
+-----+---------------+--------------------+----------------------+
| avg | 3884          | 595.9              | 2216.6               |
| p90 | 7721          | 1952.3             | 5709.0               |
| p99 | 14715         | 6900.0             | 13023.5              |
+-----+---------------+--------------------+----------------------+

In general, cross CCX/LLC domain migrations are expensive at high utilization and results in lower IPC due to cache locality. Disabling forced task stealing allows us to trade off some work conservation for better cache characteristics and scheduler overhead.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -8,6 +8,8 @@
  * To be included to the main.bpf.c
  */
 
+extern const volatile u8	mig_delta_pct;
+
 u64 __attribute__ ((noinline)) calc_mig_delta(u64 avg_sc_load, int nz_qlen)
 {
 	/*
@@ -66,7 +68,13 @@ int plan_x_cpdom_migration(void)
 			continue;
 		}
 
-		util = (cpdomc->cur_util_sum << LAVD_SHIFT) / cpdomc->nr_active_cpus;
+		/*
+		 * Use avg_util_sum when mig_delta_pct is set, otherwise use cur_util_sum.
+		 */
+		if (mig_delta_pct > 0)
+			util = (cpdomc->avg_util_sum << LAVD_SHIFT) / cpdomc->nr_active_cpus;
+		else
+			util = (cpdomc->cur_util_sum << LAVD_SHIFT) / cpdomc->nr_active_cpus;
 		qlen = cpdomc->nr_queued_task;
 		sc_qlen = (qlen << (LAVD_SHIFT * 3)) / cpdomc->cap_sum_active_cpus;
 		cpdomc->sc_load = util + sc_qlen;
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -85,8 +93,14 @@ int plan_x_cpdom_migration(void)
 	/*
 	 * Determine the criteria for stealer and stealee domains.
 	 * The more the system is loaded, the tighter criteria will be chosen.
+	 * When mig_delta_pct is set (non-zero), use it as a fixed percentage
+	 * instead of the dynamic calculation.
 	 */
-	x_mig_delta = calc_mig_delta(avg_sc_load, nz_qlen);
+	if (mig_delta_pct > 0) {
+		x_mig_delta = (avg_sc_load * mig_delta_pct) / 100;
```
multics69: The LAVD code internally uses 1024 scale (LAVD_SHIFT, LAVD_SCALE).  I suggest following the same convention for consistency. For an option from the user, [0-100] makes more sense.

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -119,6 +119,16 @@ struct Opts {
     #[clap(long = "slice-min-us", default_value = "500")]
     slice_min_us: u64,
 
+    /// Migration delta threshold percentage (0-100). When set to a non-zero value,
+    /// uses average utilization for threshold calculation instead of current
+    /// utilization, and the threshold is calculated as: avg_load * (mig-delta-pct / 100).
+    /// Additionally, disables force task stealing in the consume path, relying only
+    /// on the is_stealer/is_stealee thresholds for more predictable load balancing.
+    /// Default is 0 (disabled, uses dynamic threshold based on load with both
+    /// probabilistic and force task stealing enabled).
```
multics69: It would be useful to add this is an experimental feature.

PR #2914, opened 2025-10-18
README: Add missing kernel option

### LINKED ISSUE #2911: missing kernel option
Extent the kernel option file with:

CONFIG_BPF_LSM=y

otherwise there are permission issues.

EricccTaiwan:
Close #2911

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ The kernel has to be built with the following configuration:
 - `CONFIG_BPF_JIT_ALWAYS_ON=y`
 - `CONFIG_BPF_JIT_DEFAULT_ON=y`
 - `CONFIG_SCHED_CLASS_EXT=y`
+- `CONFIG_BPF_LSM=y`
 
 The [`scx/kernel.config`](./kernel.config) file includes all required and other recommended options for using `sched_ext`.
 You can append its contents to your kernel `.config` file to enable the necessary features.


arighi (commented):
Is it really true? I just recompiled a kernel with `CONFIG_SECURITY=y` and `CONFIG_BPF_LSM not set` and I don't get any permission error. However, I tested it inside virtme-ng, so maybe other components are involved to trigger the issue (likely systemd).

So, I'm ok to recommend this option, but it'd be nice to understand exactly what is happening.

etsal:
> Is it really true? I just recompiled a kernel with `CONFIG_SECURITY=y` and `CONFIG_BPF_LSM not set` and I don't get any permission error. However, I tested it inside virtme-ng, so maybe other components are involved to trigger the issue (likely systemd).
> 
> So, I'm ok to recommend this option, but it'd be nice to understand exactly what is happening.

Second that, looking at the logs in the original issue (https://github.com/Frogging-Family/linux-tkg/issues/1156) The problem seems lavd-specific and related to ftrace support. 

Should we wait till the problem gets triaged a bit better before merging? The original problem (lavd not finding some pretty basic kfuncs) is not that obviously related to CONFIG_BPF_LSM.

multics69:
Let me reproduce the original issue (https://github.com/Frogging-Family/linux-tkg/issues/1156) with your changes.

PR #2925, opened 2025-10-20
scx_rustland: small latency-oriented improvements

arighi:
Small changes to scx_rustland focused at improving system responsiveness:
- do not penalize newly created task (re-align newly created tasks to the current vruntime, without charging extra time)
- always consume/dispatch one task at a time (reduce work-conservation at saturation, but it improves responsiveness)

Both changes help to achieve better performance at system saturation (tested with parallel kernel builds, stress-ng, etc.).


arighi:
> LGTM, would be interesting to profile some of the userspace code paths to see how it performs at different utilization/saturation.

Yes, that's something I've been thinking about, and being in userspace it should be easy, we can use any "regular" profiling tool.

PR #2929, opened 2025-10-21
scx_mitosis: Make timer_cb more robust

dschatzberg:
We've seen cases where this rootcg context lookup can fail in a timer. It feels unlikely for initialization to take so long but in theory it is possible since the timer is armed in init and there's no synchronization with the rest.


kkdwvd:
Also a better approach might be to not make it fallible to catch look up failures in general, but use an atomic flag to signal completion of init side racing with the timer.

tommy-u (approved):

dschatzberg:
Please take another look - implemented this a bit differently by initializing the rootcg storage before the timer ever gets armed. Should avoid any concerns about this papering over real issues

kkdwivedi (approved):
LGTM

PR #2930, opened 2025-10-22
scx_lavd: Add --pinned-slice-us option for improved pinned task respo…

daidavid:
scx_lavd: Add --pinned-slice-us option for improved pinned task responsiveness

Add an option --pinned-slice-us that improves responsiveness for pinned tasks by reducing the time slice for tasks running on the contended CPU. This also allows for a longer default slice configuration while maintaining better latencies for workloads using pinning. When enabled, Per-CPU DSQs are created and pinned tasks are enqueued into their own queues.

For workloads running erlang or employ other per-cpu pinning, we observed long scheduling delays from non-pinned tasks competing with pinned tasks waiting on the same CPU. Employing a short slice time effectively allows the pinned task to pre-empt the task that's currently running on the CPU. This will allow unpinned tasks to re-enqueue on shared DSQs and find a less contended CPU while still maintaining some fairness between tasks when the system is close to saturation.

We also observed poor task placement due to congestion when using shared DSQs for pinned tasks due to:

- Having more/heavier pinned tasks increases contention on the DSQ lock, as CPUs require more time to iterate through ineligible tasks that are at the front of the queue while holding the lock

- Completely random task stealing from all CPUs will self balance to steal from more congested queues(as task stealing will more likely to succeed from longer queues)

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -213,6 +213,13 @@ const volatile u64	slice_max_ns = LAVD_SLICE_MAX_NS_DFL;
  */
 const volatile u8	mig_delta_pct = 0;
 
+/*
+ * Slice time for all tasks when pinned tasks are running on the CPU.
+ * When this is set (non-zero), pinned tasks always use per-CPU DSQs and
+ * the dispatch logic compares vtimes across DSQs.
+ */
+const volatile u64	pinned_slice_ns = 0;
+
 static volatile u64	nr_cpus_big;
 
 /*
@@ -265,6 +272,17 @@ static u64 calc_time_slice(struct task_ctx *taskc, struct cpu_ctx *cpuc)
 	 */
 	if (!taskc || !cpuc)
 		return LAVD_SLICE_MAX_NS_DFL;
+
+	/*
+	 * If pinned_slice_ns is enabled and there are pinned tasks waiting
+	 * to run on this CPU, unconditionally reduce the time slice for
+	 * all tasks to ensure pinned tasks can run promptly.
+	 */
+	if (pinned_slice_ns && cpuc->nr_pinned_tasks) {
+		taskc->slice = pinned_slice_ns;
+		reset_task_flag(taskc, LAVD_FLAG_SLICE_BOOST);
+		return taskc->slice;
+	}
 
 	/*
 	 * If the task's avg_runtime is greater than the regular time slice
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -562,7 +580,10 @@ s32 BPF_STRUCT_OPS(lavd_select_cpu, struct task_struct *p, s32 prev_cpu,
 
 		if (!scx_bpf_dsq_nr_queued(dsq_id)) {
 			p->scx.dsq_vtime = calc_when_to_run(p, taskc);
-			p->scx.slice = LAVD_SLICE_MAX_NS_DFL;
+			if (pinned_slice_ns && cpuc->nr_pinned_tasks)
+				p->scx.slice = pinned_slice_ns;
+			else
+				p->scx.slice = LAVD_SLICE_MAX_NS_DFL;
```
multics69: Actually, this change is not necessary. LAVD calculates the time slice at ops.running() since we can calculate the system load most precisely right before the running. In other words, the time slice given at ops.select_cpu() and ops.enqueue() will be overridden at ops.running() on purpose. So the change should be moved to the ops.running(). 

```
	/*
	 * If the sched_ext core directly dispatched a task, calculating the
	 * task's deadline and time slice was also skipped. In this case, we
	 * set the deadline to the current logical lock.
	 *
	 * Note that this is necessary when the kernel does not support
	 * SCX_OPS_ENQ_MIGRATION_DISABLED or SCX_OPS_ENQ_MIGRATION_DISABLED
	 * is not turned on.
	 */
	if (p->scx.slice == SCX_SLICE_DFL)
		p->scx.dsq_vtime = READ_ONCE(cur_logical_clk);

	/*
	 * Calculate the task's time slice here,
	 * as it depends on the system load.
	 */
	p->scx.slice = calc_time_slice(taskc, cpuc);

```

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -651,15 +672,21 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 		__sync_fetch_and_add(&cpuc->nr_pinned_tasks, 1);
 	}
 
+	if (pinned_slice_ns && cpuc->nr_pinned_tasks)
+		p->scx.slice = pinned_slice_ns;
+
```
multics69: Same as the previous comment, this also needs to be moved to the ops.running().

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1077,6 +1104,16 @@ void BPF_STRUCT_OPS(lavd_tick, struct task_struct *p)
 	now = scx_bpf_now();
 	account_task_runtime(p, taskc, cpuc, now);
 
+	/*
+	 * If pinned_slice_ns is enabled and there are pinned tasks waiting
+	 * to run on this CPU, unconditionally reduce the time slice for
+	 * all tasks to ensure pinned tasks can run promptly.
+	 */
+	if (pinned_slice_ns && cpuc->nr_pinned_tasks) {
+		p->scx.slice = pinned_slice_ns;
```
multics69, line 1108: `p->scx.slice` could be smaller than `pinned_slice_ns`. So, we need to check if `p->scx.slice > pnned_slice_ns` before overriding the slice.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
daidavid: The ops.running() case should be handled in calc_time_slice. I was still seeing some long slices in the previous iteration, so I was a bit heavy handed in replacing all the p->scx.slice overrides. I redid some testing and it looks okay now.

PR #2935, opened 2025-10-22
scx_mitosis: Fix enqueue vtime stale read

dschatzberg:
The existing logic reads vtime from the process before maybe_refresh_cell which could reassign the process to a cell and reset its vtime in doing so. This stale vtime is then used in later comparisons. I also extended the error message with some more context in case we see such issues more often.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -575,8 +575,8 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
 	struct task_ctx *tctx;
 	struct cell	*cell;
 	s32		 task_cpu = scx_bpf_task_cpu(p);
-	u64		 vtime	  = p->scx.dsq_vtime;
-	s32		 cpu	  = -1;
+	u64		 vtime;
+	s32		 cpu = -1;
 	u64		 basis_vtime;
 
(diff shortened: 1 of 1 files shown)


dschatzberg:
@JakeHillion - the formatting for local variables is a little gross to me. First, the spacing is just weird but I can get over that. Take a look at this commit - changing one variable causes formatting changes to another unrelated variable which I think makes code reviewing a lot more difficult. Any thoughts on that?

JakeHillion:
> @JakeHillion - the formatting for local variables is a little gross to me. First, the spacing is just weird but I can get over that. Take a look at this commit - changing one variable causes formatting changes to another unrelated variable which I think makes code reviewing a lot more difficult. Any thoughts on that?

Yeah, agreed, that's a bit weird. I think the kernel's clang-format is focusing on improving the readability of code at the cost of diff readability. One interesting thing here is if you review this diff with GitHub's "Hide whitespace" option enabled it looks fine, but obviously still messes up the git history. Another is that it seems to only align equals signs when there are several in a row, so if the `=` are all grouped this specific diff wouldn't have altered indentation, maybe that's a kernel formatting thing that isn't enforced by clang-format?

Not sure what the cost is of modifying that option & diverging from the kernel's clang-format vs suffering it. @hodgesds any thoughts?

dschatzberg:
> Not sure what the cost is of modifying that option & diverging from the kernel's clang-format vs suffering it. @hodgesds any thoughts?

I wonder if this is because typical behavior in kernel space is only to format the diff and not everything. I kind of like that we format everything (keeps things consistent) - but that divergence has the implications seen in this PR.

PR #2938, opened 2025-10-23
mitosis: add grace period to startup errs

likewhatevs:
add a configurable grace period, in the form of iterations over which to reduce errs to warns. this should resolve 'No queueing decisions made globally' around startup. also address lints and log formatting around this.


likewhatevs:
this is probably for a misconfiguration

PR #2939, opened 2025-10-23
scx_bpfland: Add --preferred-idle-scan option to prioritize faster CPUs

arighi:
Similarly to scx_cosmos, add a `--preferred-idle-scan/-P` option to scx_bpfland as well.

When enabled, the scheduler attempts to select idle CPUs in order from fastest (highest capacity) to slowest (lower capacity), ignoring the built-in idle CPU selection policy.

This option has shown consistent improvements in topologies where the cost of a cross-LLC migration is negligible compared to picking the fastest CPUs.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -7,6 +7,12 @@
 #include "intf.h"
 
 /*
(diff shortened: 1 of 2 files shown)


arighi:
Updated the PR dropping the the first commit (that requires some rework to avoid triggering the SCX_DSQ_LOCAL_ON / migration disabled tasks issue). I'll send another PR for that.

Issue #2934, opened 2025-10-22
scx_lavd E2BIG error with --per-cpu-dsq


meflove:
System & Kernel Information:
- OS: NixOS (nixpkgs master branch)
- Kernel: 6.17.3-cachyos (x86-64-v3)

scx version:
```
scx_lavd --version
scx_lavd 1.0.17 x86_64-unknown-linux-gnu
```

Problem: When attempting to start `scx_lavd` with the `--per-cpu-dsq` argument (specifically, using the command below), I encounter an error related to loading BPF programs:

Command Causing Error:
```bash
sudo scx_lavd --performance --no-core-compaction --per-cpu-dsq
```

```
-- END PROG LOAD LOG --

23:37:37 [WARN] libbpf: prog 'lavd_dispatch': failed to load: -E2BIG

23:37:37 [WARN] libbpf: failed to load object 'bpf_bpf'

23:37:37 [WARN] libbpf: failed to load BPF skeleton 'bpf_bpf': -E2BIG

Error: Failed to load BPF program

Caused by:
    Argument list too long (os error 7)
```

[scx_lavd.log](https://github.com/user-attachments/files/23057827/scx_lavd.log)

multics69:
It would be related to an older version of clang. Would you share the output of `clang --version`?

meflove:
I use NixOS, if that matters
clang version 21.1.2

multics69:
@meflove -- I think I fixed the problem. Could you try with https://github.com/sched-ext/scx/pull/2942 ?

meflove:
@multics69  still the same problem

multics69:
Hmm... that is interesting. I think I should set up a NixOS environment to reproduce your problem.

meflove:
my config for kernel
```Nix
    environment.systemPackages = with pkgs; [
      config.boot.kernelPackages.v4l2loopback
      v4l-utils
    ];

    hardware.xone.enable = true;

    services.scx = {
      enable = true;
      package = pkgs.callPackage ../../../packages/scx_rustcheds {};

      scheduler = "scx_lavd";
      extraArgs = [
        "--performance"
        "--no-core-compaction"
        # "--per-cpu-dsq"
      ];
    };

    boot = {
      kernelPackages = cfg.kernelPackage; # pkgs.linuxPackages_cachyos.cachyOverride {mArch = "GENERIC_V3";} from chaotic-cx/nyx

      initrd = {
        compressor = "cat";
      };

[...]

PR #2942, opened 2025-10-24
scx_lavd: Fix BPF verification errors.

multics69:
BPF verifier emits errors when the code is compiled with an old version of clang (18.**) or when the --per-cpu-dsq option is given. Let's avoid the errors by simplying the loop or preventing function inlining.

This should solve https://github.com/sched-ext/scx/issues/2934


multics69:
This requires some more work since it does not solve https://github.com/sched-ext/scx/issues/2934 .
I will reproduce the problem on NixOS.

multics69:
Let's merge the first patch that does not cause a veristat failure first (https://github.com/sched-ext/scx/pull/2965). So close this one for now.

PR #2943, opened 2025-10-24
scx_bpfland: Introduce per-CPU DSQs

arighi:
When targeting specific CPUs with SCX_DSQ_LOCAL_ON in ops.enqueue(), errors like the following may occur:
```
  runtime error (SCX_DSQ_LOCAL[_ON] cannot move migration disabled execsnoop[1423] from CPU 9 to 3)
```
This can happen even when explicitly checking for migration-disabled tasks or CPU affinity, since a task may be dequeued and re-enqueued during the enqueue path, changing its affinity or migration-disabled state in the process.

Because of this, ops.enqueue() is not a reliable place to use SCX_DSQ_LOCAL_ON and this limitation likely needs to be addressed in the kernel.

As a temporary workaround, introduce per-CPU DSQs and implement the proper machinery in ops.dispatch(), where task affinity and migration state are reliable, to validate dispatches to specific CPUs.

If a task is dispatched to an invalid CPU from ops.enqueue(), it will automatically be rerouted to the local DSQ of its assigned CPU from ops.dispatch().

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -237,6 +237,32 @@ struct task_ctx *try_lookup_task_ctx(const struct task_struct *p)
 }
 
 /*
+ * Return the DSQ id of the corresponding @cpu.
+ */
+static inline u64 cpu_dsq(s32 cpu)
+{
+	return cpu;
+}
+
+/*
+ * Return the DSQ id of the corresponding @cpu.
+ */
+static inline u64 node_dsq(s32 cpu)
+{
+	int node = __COMPAT_scx_bpf_cpu_node(cpu);
+
+	return nr_cpu_ids + node;
+}
+
+/*
+ * Return true if @dsq_id is a per-CPU DSQ, false otherwise.
+ */
+static inline bool is_cpu_dsq(u64 dsq_id)
+{
+	return dsq_id < nr_cpu_ids;
+}
+
+/*
  * Return true if the target task @p is a kernel thread.
  */
 static inline bool is_kthread(const struct task_struct *p)
@@ -651,49 +677,13 @@ s32 BPF_STRUCT_OPS(bpfland_select_cpu, struct task_struct *p, s32 prev_cpu, u64 
 	cpu = pick_idle_cpu(p, prev_cpu, is_this_cpu_allowed ? this_cpu : -1,
 			    wake_flags, false);
 	if (cpu >= 0) {
-		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, task_slice(p), 0);
+		scx_bpf_dsq_insert(p, cpu_dsq(cpu), task_slice(p), 0);
 		__sync_fetch_and_add(&nr_direct_dispatches, 1);
 
 		return cpu;
 	}
 
 	return prev_cpu;
-}
-
-/*
- * Return true if we should attempt a task migration to an idle CPU, false
- * otherwise.
- */
-static bool try_direct_dispatch(const struct task_struct *p, s32 prev_cpu, u64 enq_flags)
-{
-	/*
-	 * Per-CPU tasks are not allowed to migrate, so don't attempt a
-	 * direct dispatch on a different CPU.
-	 */
-	if (is_pcpu_task(p))
-		return false;
-
-	/*
-	 * Always attempt to migrate to a different CPU if the task was
-	 * re-enqueued due to a higher scheduling class stealing the CPU it
-	 * was using or queued on.
-	 */
-	if (enq_flags & SCX_ENQ_REENQ)
-		return true;
-
-	/*
-	 * Always attempt to dispatch directly to another CPU if we're
-	 * contending an SMT core.
-	 */
-	if (is_smt_contended(prev_cpu))
-		return true;
-
-	/*
-	 * Attempt a direct dispatch on wakeup (if ops.select_cpu() was
-	 * skipped) or if the task was re-enqueued due to a higher
-	 * scheduling class stealing the CPU it was queued on.
-	 */
-	return !__COMPAT_is_enq_cpu_selected(enq_flags) && !scx_bpf_task_running(p);
 }
 
 /*
@@ -775,7 +765,6 @@ static u64 update_freq(u64 freq, u64 interval)
 void BPF_STRUCT_OPS(bpfland_enqueue, struct task_struct *p, u64 enq_flags)
 {
 	s32 prev_cpu = scx_bpf_task_cpu(p);
-	int node = __COMPAT_scx_bpf_cpu_node(prev_cpu);
 	u64 slice = task_slice(p);
 	struct task_ctx *tctx;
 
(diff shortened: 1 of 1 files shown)


arighi:
@hodgesds do you think something like this can work for scx_p2dq as well, instead of always using `SCX_DSQ_LOCAL` (to fix #2879)?

In any case, I think we need to address this in the kernel, otherwise it's just not possible to use `SCX_DSQ_LOCAL_ON` from `ops.enqueue()`.

hodgesds:
> @hodgesds do you think something like this can work for scx_p2dq as well, instead of always using SCX_DSQ_LOCAL

Yeah, I think this is closer to what I've been testing (unpublished). I've been looking into the kernel side as well as I discovered you can easily livelock a system when unloading a scheduler under load and when heavily using SCX_DSQ_LOCAL.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -862,21 +873,90 @@ static bool keep_running(const struct task_struct *p, s32 cpu)
...
+	if (is_cpu_dsq(dsq_id)) {
+		s32 cpu = dsq_cpu(dsq_id), task_cpu = scx_bpf_task_cpu(p);
+
+		if (!p->cpus_ptr)
+			return false;
+		/*
+		 * If the task can't run on this CPU, move it to the local
+		 * DSQ of its previously assigned CPU.
+		 */
+		if (!bpf_cpumask_test_cpu(cpu, p->cpus_ptr) ||
```
hodgesds: Really easy to miss this bit 👍

hodgesds (approved):

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Actually there's something that I still don't understand, do you see before this check I'm also checking if `p->cpus_ptr == NULL`? This shouldn't be required, because `bpf_cpumask_test_cpu()` should be able to handle NULL, but I can actually crash the kernel if I don't add that check, which I think it's a kernel bug:
```
[   14.257828] BUG: kernel NULL pointer dereference, address: 0000000000000000
[   14.258032] #PF: supervisor read access in kernel mode
[   14.258141] #PF: error_code(0x0000) - not-present page
[   14.258206] PGD 0 P4D 0
[   14.258252] Oops: Oops: 0000 [#1] SMP NOPTI
[   14.258314] CPU: 14 UID: 0 PID: 61 Comm: migration/14 Not tainted 6.18.0-rc2-1-cachyos-rc #1 PREEMPT(full)  2a7889c09d4652cf7a79d70b3ad9174e55006d5b
[   14.258488] Hardware name: Bochs Bochs, BIOS Bochs 01/01/2011
[   14.258585] Stopper: 0x0 <- 0x0
[   14.258684] Sched_ext: bpfland_1.0.17_g24ee315f_dirty_x86_64_unknown_linux_gnu (enabled+all)
[   14.258686] RIP: 0010:bpf_cpumask_test_cpu+0x15/0x30
[   14.258874] Code: cc cc 0f 1f 00 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 0f 1f 40 d6 0f 1f 44 00 00 31 c0 3b 3d 9f 1c 32 02 73 09 89 ff <48> 0f a3 3e 0f 92 c0 c3 cc cc cc cc 0f 1f 40 00 66 66 2e 0f 1f 84
[   14.259087] RSP: 0018:ffffcea2c0283c28 EFLAGS: 00010083
[   14.259171] RAX: 0000000000000000 RBX: 000000000000000e RCX: ffff8a8a0107f638
[   14.259267] RDX: 0000000000000000 RSI: 0000000000000000 RDI: 000000000000000e
[   14.259360] RBP: ffffcea2c0283c78 R08: 0000000000000000 R09: 0000000000000001
[   14.259462] R10: 0000000000000000 R11: 0000000000000000 R12: ffffeea2bfd89088
[   14.259565] R13: ffff8a8a12afa180 R14: 0000000000000000 R15: ffff8a8a05302180
[   14.259675] FS:  0000000000000000(0000) GS:ffff8a8ab0a6f000(0000) knlGS:0000000000000000
[   14.259767] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[   14.259853] CR2: 0000000000000000 CR3: 00000001092f7000 CR4: 0000000000750ef0
[   14.259949] PKRU: 55555554
[   14.259986] Call Trace:
[   14.260025]  <TASK>
[   14.260064]  bpf_prog_7e561be414cf44e0_consume_first_task+0x5d/0xe5
[   14.260153]  ? scx_bpf_dsq_peek+0x2a/0x80
[   14.260214]  bpf_prog_cdadde6056166037_bpfland_dispatch+0xcf/0x248
[   14.260291]  ? scx_bpf_kick_cpu+0x40/0x50
[   14.260347]  ? __bpf_prog_enter+0x2b/0xa0
[   14.260394]  bpf__sched_ext_ops_dispatch+0x4b/0xa8
[   14.260466]  pick_task_scx+0x208/0x5b0
[   14.260514]  __pick_next_task.part.0+0xed/0x1d0
[   14.260583]  ? __pfx_smpboot_thread_fn+0x10/0x10
[   14.260662]  ? __pfx_smpboot_thread_fn+0x10/0x10
[   14.260733]  __schedule+0x5c1/0x1750
[   14.260794]  ? move_queued_task+0x41/0x120
[   14.260851]  ? complete_all+0x51/0xa0
[   14.260908]  ? migration_cpu_stop+0x210/0x3a0
[   14.260980]  ? __pfx_smpboot_thread_fn+0x10/0x10
[   14.261052]  schedule+0x27/0xd0
[   14.261111]  smpboot_thread_fn+0x190/0x220
[   14.261168]  kthread+0xfc/0x240
[   14.261218]  ? __pfx_kthread+0x10/0x10
[   14.261274]  ? __pfx_kthread+0x10/0x10
[   14.261329]  ret_from_fork+0x1c3/0x1f0
[   14.261385]  ? __pfx_kthread+0x10/0x10
[   14.261441]  ret_from_fork_asm+0x1a/0x30
[   14.261499]  </TASK>
```

I'm wondering if the right thing here is to add a `bpf_rcu_read_lock/unlock()`, which also shouldn't be required because we should be in an RCU-safe place... so, I'm confused.

arighi: Oh I think I understand, `p->cpus_ptr` is protected by `p->pi_lock` which is not held here, so it isn't safe to check... meh.

hodgesds: > do you see before this check I'm also checking if p->cpus_ptr == NULL? 

I thought that was just for the verifier.

> but I can actually crash the kernel if I don't add that check, which I think it's a kernel bug:

I'll check this as well.

arighi:
@hodgesds I think I found a reliable and bug-free fix, just updated the PR. See if it makes sense to you.

We'll discuss about a potential in-kernel approach separately.

hodgesds:
LGTM, kind of annoying that the DSQ iterator has to be used everywhere to be safe.

arighi:
Alright, even if I'm still not 100% convinced and I need to understand exactly what is happening, but this one passed all my stress tests, not stall, bugs, etc. and performance wise seems also pretty good, so let's merge this. We'll improve things in tree if needed.

PR #2944, opened 2025-10-24
scx_layered: Add per-layer LLC stickiness control

hodgesds:
Add a new configuration option `llc_sticky_runs` that controls how many times a task must run on its current LLC before being allowed to migrate to a different LLC. This provides finer-grained control over LLC locality and complements the existing `xllc_mig_min_us` option.

The implementation tracks the number of consecutive runs on the current LLC in the task context (`taskc->llc_runs`). When the scheduler attempts to find an idle CPU and no idle CPUs are available on the current LLC, it checks if the task has run fewer times than the configured threshold. If so, cross-LLC migration is prevented, keeping the task sticky to its current LLC to preserve cache locality.

Key changes:

- Add `llc_runs` counter to `task_ctx` to track consecutive runs on current LLC
- Add `llc_sticky_runs` configuration field to `layer` struct and `LayerConfig` in config.rs
- Increment `llc_runs` in `layered_running()` when stickiness is enabled
- Reset `llc_runs` to 0 in `maybe_update_task_llc()` on LLC migration
- Add stickiness check in `pick_idle_cpu()` to prevent cross-LLC migration when task hasn't run enough times on current LLC
- Add `LSTAT_LLC_STICKY_SKIP` statistic to track prevented migrations
- Initialize `llc_runs` to 0 in `layered_init_task()`

The feature is disabled by default (llc_sticky_runs = 0) and can be configured per-layer. Example configuration:

  {
    "name": "batch",
    "kind": {
      "Confined": {
        "llc_sticky_runs": 20,
      }
    }
  }

This would keep tasks on their current LLC for 20 runs before allowing cross-LLC migration, providing stronger LLC affinity for workloads that benefit from cache locality.

The new statistic appears in the output as:
  xllc_mig/skip/sticky_skip={xllc_migration%}/{xllc_skip%}/{llc_sticky_skip%}

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -148,6 +148,7 @@ enum layer_stat_id {
 	LSTAT_XNUMA_MIGRATION,
 	LSTAT_XLLC_MIGRATION,
 	LSTAT_XLLC_MIGRATION_SKIP,
+	LSTAT_LLC_STICKY_SKIP,
 	LSTAT_XLAYER_WAKE,
 	LSTAT_XLAYER_REWAKE,
 	LSTAT_LLC_DRAIN_TRY,
@@ -348,6 +349,7 @@ struct layer {
 	u64			disallow_open_after_ns;
 	u64			disallow_preempt_after_ns;
 	u64			xllc_mig_min_ns;
+	u32			llc_sticky_runs;
 
 	int			kind;
 	bool			preempt;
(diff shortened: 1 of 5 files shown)


etsal (approved):

hodgesds:
~This doesn't work 100% of the time because during dispatch task can be forcefully migrated off their LLC DSQs, but that's maybe ok.~ Added support at dispatch time using iterators.

hodgesds:
Added the dispatch side and running `stress-ng` with `--run-example` shows sticky skips:
```
open_idle= 0.00 mig= 1.38 xnuma_mig= 0.10 xllc_mig/skip/sticky_skip= 0.36/ 0.00/ 0.01 affn_viol= 0.00
```

hodgesds:
Not sure this approach works very well, the overhead of the DSQ iterator is pretty expensive:
image
compared to main:
image
The overhead is almost 1K ns using iterators.

hodgesds:
Closing as I don't think the overhead is worth it.

PR #2948, opened 2025-10-25
scx_bpfland: Prevent starvation and improve performance consistency

arighi:
A series of commits aimed at preventing per-CPU DSQs from starving per-node DSQs.

To achieve this, when selecting a task in `ops.dispatch()`, we now always pick the one with the earliest deadline, comparing the first task from both the per-CPU and per-node DSQs.

Additionally, task deadlines are now updated also during direct dispatch, ensuring this logic remains effective in all cases.

With these changes in place, the scheduler shows more consistent performance results and it fixes a starvation issue (stall) observed on a 72-cores system running `stress-ng --tlb-shootdown 75% -t 60s`.


arighi:
cc: @colinianking

PR #2950, opened 2025-10-27
scx_utils, scx_lavd: Harden the futex tracing.

multics69:
When neither tracepoint nor ftrace is enabled by the kernel, the existing futex tracing code raises runtime errors. This PR is to handle such error cases gracefully. It first add a missing error condition check (compat::tracepoint_exists), and check if ftraced is available before attaching ftrace points (compat::tracer_available). 

This should resolve the problem reported at https://github.com/Frogging-Family/linux-tkg/issues/1156
Once this is merged, https://github.com/sched-ext/scx/pull/2914 should be closed.

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -214,6 +214,25 @@ pub fn debugfs_mount() -> Result<std::path::PathBuf> {
     mounts.into_iter().next().context("No debugfs mount found")
 }
 
+pub fn tracer_available(tracer: &str) -> Result<bool> {
+    let base_path = tracefs_mount().unwrap_or_else(|_| debugfs_mount().unwrap().join("tracing"));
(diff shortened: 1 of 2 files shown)


arighi (approved):
LGTM

AdelKS:
Hey,

Thanks for looking into it ! Can disabling FTRACE negatively affect performance, I.e. is it used for non-debugging purposes, e.g. tracing processes to get insights on better scheduling decisions ?

multics69:
> Thanks for looking into it ! Can disabling FTRACE negatively affect performance, I.e. is it used for non-debugging purposes, e.g. tracing processes to get insights on better scheduling decisions ?

It depends on workloads, of course. If the workload heavily relies on futex, it will help a bit. Otherwise, it shouldn't matter much.

PR #2955, opened 2025-10-27
scxtop: Fix P99 Runtime chart width in BPF detailed view

hodgesds:
The P99 Runtime sparkline was not filling the entire horizontal pane width like the other charts (Runtime and Calls/sec). This was caused by calculating p99_data from the already-trimmed runtime_data instead of directly from prog_data.runtime_history.

Changed the p99_data calculation to use prog_data.runtime_history directly with VecDeque iterator methods, ensuring it generates the same number of data points as the other charts and fills the available width.


hodgesds:
image

PR #2958, opened 2025-10-27
ci: implement selective test matrix based on changed files

hodgesds:
- Add change detection logic to list-integration-tests.py
- Build dependency graph from Cargo metadata
- Filter test matrix to only affected schedulers
- Test all schedulers when core files or libraries change
- Add local testing script and documentation

This reduces CI time for PRs that only touch specific schedulers.

Technical changes:
- Added get_changed_files() to detect file changes in PR context
- Added should_test_all_schedulers() to categorize changes
- Added build_scheduler_dependency_map() to track Cargo dependencies
- Added get_affected_schedulers() to determine which schedulers to test
- Modified main() to filter the scheduler list based on changes
- Preserved backward compatibility: non-PR contexts test everything

Examples:
- PR changing only scx_chaos: tests 1 scheduler (85-90% CI time saved)
- PR changing scx_utils: tests all schedulers (conservative)
- PR changing scx_rustland_core: tests scx_rustland and scx_rlfifo only

### DIFF
--- a/.github/include/list-integration-tests.py
+++ b/.github/include/list-integration-tests.py
@@ -1,10 +1,31 @@
 #!/usr/bin/env python3
+"""
+Generate a test matrix for integration tests, filtering based on changed files.
+
+This script determines which schedulers need to be tested based on:
+1. Files changed in the PR (for pull_request events)
+2. Dependency relationships from Cargo.toml
+3. Kernel requirements from package metadata
+
+Behavior:
+- In non-PR contexts (push to main, scheduled runs): Tests all schedulers
+- When core files change (rust/scx_utils, .github/, etc.): Tests all schedulers
+- When only specific scheduler directories change: Tests only those schedulers
+- When a library changes: Tests all schedulers that depend on it
+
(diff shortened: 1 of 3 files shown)


hodgesds:
Not sure how to test this besides landing really, but this should significantly speed up CI.

.github/include/list-integration-tests.py
```diff
@@ -104,6 +125,205 @@ def get_kernel_trailers_from_commits():
...
+        "rust/scx_bpf_compat/",
+        "rust/scx_raw_pmu/",
+        "scheds/include/",
+        ".github/",
+        ".nix/",
+        "Cargo.toml",  # Root workspace Cargo.toml
+        "Cargo.lock",
+        "meson.build",
+        "kernel-versions.json",
+    ]
```
JakeHillion, line 192: Do you think we could invert this? I wonder whether it would be better to say that changes in `scheds/rust` don't necessarily run everything, but every other change does. This list scares me a bit!

htejun:

PR #2965, opened 2025-10-28
scx_lavd: Fix a BPF verifier error (E2BIG) at do_core_compaction().

multics69:
With an old version clang (18.*), the BPF verifier spits a E2BIG error around do_core_compaction(). So, let's simplify the loop inside do_core_compaction() without changing the logic to make the verifier happy.


hodgesds (approved):
Makes sense, looping on nr_cpus and keeping state is hard make the verifier happy.

PR #2971, opened 2025-10-28
scx_mitosis: Workaround exiting tasks in dead cgs

dschatzberg:
We've observed cgroup local storage lookup failures with scx_mitosis that we believe we've root caused to the fact that some tasks can be initialized or even get scheduling events after their cgroup has been offlined. This should be fixed in the kernel, but a workaround in scx_mitosis is to just special case tasks with PF_EXITING and fallback to the root cgroup.

This commit adds a flag (default enabled) that implements such a fallback.


dschatzberg:
@htejun - FYI this is the mitosis side fix for the issue we discussed

tommy-u (approved):

PR #2977, opened 2025-10-29
scx_loader/scxctl: remove from repository

JakeHillion:
scx_loader and scxctl expose a very different security surface than the rest of the sched_ext schedulers in this repo. They also don't benefit from the shared CI that the rest of the schedulers use, which is slow and often unhelpful.

scx_loader and scxctl have been migrated to https://github.com/sched-ext/scx-loader Remove them from the scx repository.

Test plan:
- CI


JakeHillion:
@vnepogodin @frap129 would appreciate your ACK on this. New repo should be accessible to you both.

sirlucjan (approved):

sirlucjan:
@JakeHillion We should remove also: https://github.com/sched-ext/scx/blob/main/meson-scripts/install_rust_user_scheds#L69-L72 and https://github.com/sched-ext/scx/tree/main/services/systemd and https://github.com/sched-ext/scx/blob/main/meson.build#L503-L507 and https://github.com/sched-ext/scx/blob/main/meson.options#L29-L34

frap129:
Checked out the new repo, everything looks good to me

JakeHillion:
> @JakeHillion We should remove also: https://github.com/sched-ext/scx/blob/main/meson-scripts/install_rust_user_scheds#L69-L72 and https://github.com/sched-ext/scx/tree/main/services/systemd and https://github.com/sched-ext/scx/blob/main/meson.build#L503-L507 and https://github.com/sched-ext/scx/blob/main/meson.options#L29-L34

Thanks, there were a few more references in the docs. I think I've got everything now, but we can always follow up.

PR #2978, opened 2025-10-29
scx_mitosis: Fix tasks getting vtime too far ahead

tommy-u:
We've observed "vtime too far ahead" errors with scx_mitosis that occur when task vtimes run far ahead of cell/CPU dsq clocks. In a single call to `mitosis_stopping()`, we've observed tasks accumulating hundreds of seconds of vtime. This happened when a task with a large nice value (minimum weight) ran on a cpu for over 1.5 seconds without a scheduling event.

The error "vtime too far ahead" happens in the `enqueue()` callback. The vtime gap is detected because the task vtime has already been updated in `stopping()`, but the cell/CPU dsq vtimes have not yet been updated in `running()`.

This commit eliminates that vtime gap by unifying the three vtime updates (cell, CPU, and task) to all occur in `stopping()`.

This is done by changing the default behavior to unify vtime updates and adding a --split-vtime-updates flag to restore the legacy behavior.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -36,6 +36,7 @@ const volatile u64	     slice_ns;
 const volatile u64	     root_cgid			     = 1;
 const volatile bool	     debug_events_enabled	     = false;
 const volatile bool	     exiting_task_workaround_enabled = true;
+const volatile bool	     split_vtime_updates	     = false;
 
 /*
  * CPU assignment changes aren't fully in effect until a subsequent tick()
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -36,6 +36,7 @@ const volatile u64	     slice_ns;
 const volatile u64	     root_cgid			     = 1;
 const volatile bool	     debug_events_enabled	     = false;
 const volatile bool	     exiting_task_workaround_enabled = true;
+const volatile bool	     unify_vtime_updates	     = false;
```
dschatzberg: Let's actually default this to true. My logic is that we expect this to "just work" and be a strict improvement so letting the new behavior just roll out with new releases is preferred. The flag really just gives us the option to turn off the behavior if there's some issue.

If this were a more heavy-weight feature or something that had a higher chance of causing issues, we would want to use default disabled and a more explicit testing pattern

tommy-u: Ok, this makes sense.

PR #2983, opened 2025-10-30
scx_lavd: gate scx_bpf_dsq_nr_queued call in can_direct_dispatch

JakeHillion:
When checking whether a CPU should receive a direct dispatch lavd checks the contents of the per_cpu DSQ to ensure direct dispatching won't starve that queue. However, this check occurs even when that DSQ does not exist.

This is confirmed by adding printing of the `scx_bpf_dsq_nr_queued` result. It's always 4294967294, which is -ENOENT saying the queue does not exist. `!(-ENOENT)` is false, so `can_direct_dispatch` always returns false if `per_cpu_dsq` is false.

Modify the check to confirm whether `per_cpu_dsq` is enabled before checking the size of the queue.

Test plan:
- CI
- `bpf_printk` - before only the `else` branch was hit, now I get a mixture.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -596,8 +596,12 @@ static bool can_direct_dispatch(u64 dsq_id, s32 cpu, bool is_idle)
 	/*
 	 * If the chosen CPU is idle and there is nothing to do
 	 * in the domain, we can safely choose the fast track.
+	 *
+	 * When per_cpu_dsq is false, we only check if the CPU is idle.
+	 * When per_cpu_dsq is true, we also need to check if the per-CPU
+	 * DSQ is empty.
 	 */
-	return is_idle && cpu >= 0 && !scx_bpf_dsq_nr_queued(dsq_id);
+	return is_idle && cpu >= 0 && (!per_cpu_dsq || !scx_bpf_dsq_nr_queued(dsq_id));
 }
 
 void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)


JakeHillion:
Hmm, this might not be the correct change. It seems this can jump the shared DSQ if there's an idle CPU breaking fairness. The behaviour before per-CPU DSQs was instead to pass in the shared DSQ and check it, which I don't think this is doing now.

daidavid:
Sorry, I was a bit trigger happy with the approval. I think we should handle both the per-cpu-dsq enabled/disabled in this optimization path. Just a heads up that I'm currently reworking this section to check for both if we're to re-using the per-cpu DSQs for affinitized tasks.

In theory, when the per-cpu-dsq is enabled, we can potentially dispatch from both cpdom_dsq or the cpu_dsq. However, in reality, pick_proper_dsq will always return a cpu, so only the cpu dsqs are used today when per-cpu-dsq is enabled. That relationship is a bit fragile and needs to be revisited.

So I think the proper logic is to check for both cpu-dsq && cpdom_dsq when per-cpu dsq is available and only cpdom_dsq when it's not. WDYT?

When affinitized tasks on per-cpu DSQ is supported, this will simplify the check for always both cpu/cpdom dsqs during this and the other direct dispatch path on select_cpu.

multics69:
Thanks, @JakeHillion, for catching this up!

I think the correct fix would be as follows, since we do not use per-CPU DSQ and domain CPU at the same time.
Later, when we decide to use both DSQ at the same time (e.g., repurposing the per-CPU DSQ for fastpath for pinned tasks), we should check the both DSQs.

What do you think, @daidavid ? 

```.c
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -605,7 +605,7 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
        struct task_ctx *taskc;
        struct cpu_ctx *cpuc, *cpuc_cur;
        s32 task_cpu, cpu = -ENOENT;
-       u64 cpdom_id;
+       u64 cpdom_id, dsq_id;
        bool is_idle = false;
 
        taskc = get_task_ctx(p);
@@ -677,14 +677,12 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
         * When pinned_slice_ns is enabled, pinned tasks always use per-CPU DSQ
         * to enable vtime comparison across DSQs during dispatch.
         */
-       if (can_direct_dispatch(cpu_to_dsq(cpu), cpu, is_idle)) {
+       dsq_id = per_cpu_dsq ? cpu_to_dsq(cpu) : cpdom_to_dsq(cpdom_id);
+       if (can_direct_dispatch(dsq_id, cpu, is_idle)) {
                scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | cpu, p->scx.slice,
                                   enq_flags);
-       } else if (per_cpu_dsq) {
-               scx_bpf_dsq_insert_vtime(p, cpu_to_dsq(cpu), p->scx.slice,
-                                        p->scx.dsq_vtime, enq_flags);
        } else {
-               scx_bpf_dsq_insert_vtime(p, cpdom_to_dsq(cpdom_id), p->scx.slice,
+               scx_bpf_dsq_insert_vtime(p, dsq_id, p->scx.slice,
                                         p->scx.dsq_vtime, enq_flags);
        }
```

daidavid:
This section was reworked in https://github.com/sched-ext/scx/pull/3014, since we need to consider both DSQs with pinned_slice_ns. Let's continue the discussion there.

PR #2984, opened 2025-10-30
scx_bpfland: Classify sticky tasks based on their average runtime

arighi:
Tasks that generate a massive volume of scheduling events (such as `stress-ng --sleep 0`) can create excessive locking pressure on the per-node and per-CPU DSQs, potentially rendering the system unresponsive.

To mitigate this issue, the "sticky tasks" classification feature (option --sticky-tasks / -S) was introduced: when a task is classified as "scheduling intensive" it is dispatched for @slice_lag ns to the same CPU (SCX_DSQ_LOCAL). This allows reducing the locking contention on the per-CPU and per-node DSQs and it also prioritizes frequent sleepers that typically release the CPU quickly.

However, the classification logic relies on intercepting specific syscalls, such as sys_read() and sys_nanosleep(), which proved to be a fragile approach and insufficiently generic.

The new implementation takes a more robust approach by tracking the average runtime of each task: tasks with an average runtime below a defined threshold (10 us) are classified as "scheduling intensive" and dispatched directly to the local DSQ (SCX_DSQ_LOCAL).

This runtime-based classification is considerably more generic and enables more robust handling of diverse workloads.

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -226,7 +226,7 @@ struct task_ctx {
 	u64 last_run_at;
 	u64 wakeup_freq;
 	u64 last_woke_at;
-	u64 pinned_at;
+	u64 avg_runtime;
 };
 
 /* Map that contains task-local storage. */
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -752,48 +752,12 @@ s32 BPF_STRUCT_OPS(bpfland_select_cpu, struct task_struct *p, s32 prev_cpu, u64
...
-
-/*
- * Return true if the task is forced to stay on the same CPU, false
+ * Return true if the task should be forced to stay on the same CPU, false
  * otherwise.
  */
 static bool is_task_sticky(const struct task_ctx *tctx)
 {
-	return sticky_tasks && time_after(tctx->pinned_at + slice_lag, bpf_ktime_get_ns());
+	return sticky_tasks && tctx->avg_runtime < 10 * NSEC_PER_USEC;
```
hodgesds, line 760: maybe make this configurable from the CLI?

hodgesds (approved):
I've seen this pattern before 😄

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
arighi: Yep, I'm planning to do that, but at the same time I don't want people to start using random numbers, especially large values are a bit risky... so I need to figure a range of "reasonable values", and then it'll become a cmdline option. :)

PR #2991, opened 2025-10-31
scheds: Introduce scx_beerland 🍻

arighi:
Introduce scx_beerland (BPF-Enhanced Execution Runtime Locality-Aware Non-blocking Dispatcher... alright, just a funny name), a deadline-based scheduler designed to maximize locality and scalability.

With the introduction of the lockless peek API (`scx_bpf_dsq_peek()`), it has become more practical to design schedulers that rely exclusively on per-CPU DSQs, reducing locking contention on global / shared DSQs, while preserving the effectiveness of deadline-based scheduling.

The scheduler maintains one deadline-ordered DSQ per CPU. When selecting the next task to run, it normally chooses the task with the earliest deadline from the per-CPU DSQ. However, when the system becomes saturated, it can also peek into other CPUs' DSQs to find the task with the earliest deadline system-wide.

While scx_beerland takes full advantage of the `scx_bpf_dsq_peek()` API available in newer kernels, it remains functional on older kernels as well. Most of its scheduling and deadline logic is shared with scx_bpfland, and future work could involve factoring common code into shared components.

Early versions of this scheduler have been tested by the CachyOS community with promising results in both throughput and responsiveness.

Merging this scheduler can help establish a larger test base, enabling further improvements, and provide a real-world stress test for per-CPU DSQ patterns (potentially surfacing kernel issues in the process).

### DIFF
--- a/scheds/rust/scx_beerland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_beerland/src/bpf/main.bpf.c
@@ -773,7 +773,7 @@ void BPF_STRUCT_OPS(beerland_dispatch, s32 cpu, struct task_struct *prev)
 	/*
 	 * Try to consume a task from a remote CPU.
 	 *
-	 * Do no trigger any rebalance unless the system is completely
+	 * Do not trigger any rebalance unless the system is completely
 	 * saturated.
 	 */
 	if (is_system_busy() && dispatch_from_any_cpu(cpu)) {


scheds/rust/scx_beerland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,596 @@
...
+	 */
+	if (scx_bpf_dsq_move_to_local(cpu)) {
+		__sync_fetch_and_add(&nr_local_dispatch, 1);
+		return;
+	}
+
+	/*
+	 * Try to consume a task from a remote CPU.
+	 *
+	 * Do no trigger any rebalance unless the system is completely
```
multics69: "Do no" => "Do not"?

scheds/rust/scx_beerland/src/bpf/main.bpf.c
```diff
@@ -756,13 +773,15 @@ void BPF_STRUCT_OPS(beerland_dispatch, s32 cpu, struct task_struct *prev)
 
 void BPF_STRUCT_OPS(beerland_runnable, struct task_struct *p, u64 enq_flags)
 {
-	u64 now = scx_bpf_now(), delta_t;
+	u64 now = bpf_ktime_get_ns(), delta_t;
```
multics69, line 804: Just for curiosity. Any reason for switching to bpf_ktime_get_ns()?

multics69 (approved):
Nice! Really interesting set of policies!

PR #3002, opened 2025-11-03
vmlinux: Update to v6.18-rc4

etsal:
Update the bundled vmlinux to v6.18-rc4. Also fix the generation script for Fedora and update it to handle relative paths.

cc @cloehle


scripts/gen_vmlinux_h.sh
```diff
@@ -23,15 +23,15 @@ LINUX_VER=$(git describe --tags --abbrev=0 --match="v*")
 declare -A ARCHS
 ARCHS=(
     [arm]="arm-linux-gnueabi-"
-    [arm64]="aarch64-linux-gnu-"
-    [mips]="mips64-linux-gnu-"
-    [powerpc]="powerpc64le-linux-gnu-"
-    [riscv]="riscv64-linux-gnu-"
-    [s390]="s390x-linux-gnu-"
-    [x86]="x86_64-linux-gnu-"
+#    [arm64]="aarch64-linux-gnu-"
```
hodgesds: is this intended?

etsal: Nice catch, no this is leftover testing (the arm target wasn't compiling properly on Fedora due to some issue with the cross-compile toolchain).

etsal:
It also looks like mempolicy is not included, I will add it in.

hodgesds (approved):
Nice

PR #3011, opened 2025-11-05
arenas: Make sure arena is visible to the verifier for sdt_alloc

hodgesds:
When trying to alloc on the arena make it is initialized from the verifiers perspective.

@arighi I think this might fix the verifier errors you saw on the ubuntu kernels.

### DIFF
--- a/lib/sdt_alloc.bpf.c
+++ b/lib/sdt_alloc.bpf.c
@@ -148,6 +148,8 @@ int scx_alloc_attempt(struct scx_alloc_stack __arena *stack)
 {
 	int i;
 
+	scx_arena_subprog_init();
+
 	/*
 	 * Use can_loop to help with verification. The can_loop macro was
 	 * introduced in kernel commit ba39486 and wraps around the may_goto


arighi:
@hodges unfortunately the verifier is still unhappy.

Just for the records, this is happening with a 6.14 kernel, more exactly the linux-nvidia-6.14 kernel in Ubuntu (6.14.0-1008-nvidia-64k). I tried to repro with v6.14 mainline inside vng, but I couldn't...

The error is:
```
165: (18) r1 = 0xffff18d94ae11508     ; frame3: R1_w=map_value(map=.data.LOCK,ks=4,vs=4)
167: (85) call bpf_spin_lock#93       ; frame3:
168: (b4) w0 = 0                      ; frame3: R0_w=0
; if (stack->idx >= SDT_TASK_ALLOC_STACK_MIN) @ sdt_alloc.bpf.c:114
169: (79) r1 = *(u64 *)(r6 +0)
R6 invalid mem access 'scalar'
processed 110 insns (limit 1000000) max_states_per_insn 1 total_states 10 peak_states 10 mark_read 3
```

PR #3013, opened 2025-11-06
bpf_unittests: Remove p2dq tests

hodgesds:
Remove the p2dq bpf unit tests as they aren't being maintained and p2dq will be migrating more logic into arenas which would require more extensive refactoring of the tests to support.


arighi (approved):
Makes sense

PR #3014, opened 2025-11-07
scx_lavd: Reintroduce using Per-Cpu DSQs for pinned tasks

daidavid:
Use Per-Cpu DSQs when pinned_slice_ns is enabled, this also fixes a bug where U64_MAX was not defined correctly causing out of order dispatches when comparing DSQs to select tasks to run.

In workloads using erlang that relies on running "real-time" tasks on Per-CPU erlang scheduling threads, scheduling latencies that exceed 5ms results in poor service behavior at higher loads. Further increase the overall responsiveness of the system by keeping the pinned task counter until the pinned task becomes quiescent and aggressively kick out unpinned tasks during tick to avoid high p99 scheduling latencies for pinned tasks.


multics69:
The CI broke for some other reason. I built it, tested it locally, and it worked fine, so merge it into the main.

PR #3015, opened 2025-11-07
scx_mitosis: Fix runnable task stall

tommy-u:
This fixes an error I created in PR #2978, which lead to mitosis getting kicked out due to runnable task stalls.

When we update cell dsq vtime in stopping(), we need to update the cell dsq that corresponds to the CPU, not the task.

These are often the same thing, but there are special cases where a task belonging to one cell is pinned to a CPU that belongs to another cell, e.g. a pinned kworker thread.


dschatzberg (approved):
Does this change the accounting on line 1199 now? I think we only use that cell_cycles for monitoring so it's not terribly important but it would be good to confirm we aren't making another unintended change here.

In any event, either you'll need to do two cell lookups or leave the code as is so I'm approving - feel free to merge with whatever solution you see fit.

tommy-u:
Merging per our convo. We will address counters separately.

PR #3017, opened 2025-11-07
p2dq/mitosis/lavd/layered -- cli and logging productionization

likewhatevs:
This pr productionizes lavd/mitosis/p2dq/layered a bit.

It has CLI args be handled by clap_main and adds a run_id argument to enable passing in a run id to enable log correlation etc. wrt when scheduler binaries are exec'ed.

It has logging in those schedulers be handled by tracing, following conventions such as `--log-level` and `RUST_LOG`, while ensuring that logging log logs are not silently dropped.

### DIFF
--- a/scheds/rust/scx_lavd/src/main.rs
+++ b/scheds/rust/scx_lavd/src/main.rs
@@ -63,7 +63,7 @@ use stats::SchedSamples;
 use stats::StatsReq;
 use stats::StatsRes;
 use stats::SysStats;
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 use tracing_subscriber::filter::EnvFilter;
 
 const SCHEDULER_NAME: &str = "scx_lavd";
@@ -74,6 +74,10 @@ const SCHEDULER_NAME: &str = "scx_lavd";
 /// See the more detailed overview of the LAVD design at main.bpf.c.
 #[derive(Debug, Parser)]
 struct Opts {
+    /// Depricated, noop, use RUST_LOG or --log-level instead.
+    #[clap(short = 'v', long, action = clap::ArgAction::Count)]
+    verbose: u8,
+
     /// Automatically decide the scheduler's power mode (performance vs.
     /// powersave vs. balanced), CPU preference order, etc, based on system
     /// load. The options affecting the power mode and the use of core compaction
@@ -948,6 +952,10 @@ fn main(mut opts: Opts) -> Result<()> {
     }
 
     init_log(&opts);
+
+    if opts.verbose > 0 {
+        warn!("Setting verbose via -v is depricated and will be an error in future releases.");
+    }
 
     if let Some(run_id) = opts.run_id {
         info!("scx_lavd run_id: {}", run_id);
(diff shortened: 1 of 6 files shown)


multics69 (approved):
LGTM

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -209,15 +209,19 @@ struct Opts {
...
+    #[clap(long, default_value = "info")]
+    log_level: String,
 
     /// Print scheduler version and exit.
     #[clap(short = 'V', long, action = clap::ArgAction::SetTrue)]
     version: bool,
 
+    /// Optional run ID for tracking scheduler instances.
+    #[clap(long)]
+    run_id: Option<u64>,
```
hodgesds, line 234: I don't really get the run id, is it just for logging? Why not just print the pid, unless you want something deterministic?

likewhatevs: That needs to be very unique -- more/less that'd be the join key for systems which aggregate logs etc. pid could work until it didn't and things came out wrong.

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -90,6 +88,10 @@ struct Opts {
     #[clap(short = 'V', long, action = clap::ArgAction::SetTrue)]
     version: bool,
 
+    /// Optional run ID for tracking scheduler instances.
+    #[clap(long)]
+    run_id: Option<u64>,
```
dforsyth, line 97: If we're going to take run_id, should we update the log formatter to always insert it?

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -564,27 +567,37 @@ fn main() -> Result<()> {
...
-        .set_location_level(simplelog::LevelFilter::Off)
-        .set_target_level(simplelog::LevelFilter::Off)
-        .set_thread_level(simplelog::LevelFilter::Off);
-    simplelog::TermLogger::init(
-        llv,
-        lcfg.build(),
-        simplelog::TerminalMode::Stderr,
-        simplelog::ColorChoice::Auto,
-    )?;
+    let env_filter = EnvFilter::try_from_default_env()
```
dforsyth, line 574: maybe for a later change -- these are the same in all the schedulers you touched here (I think). Can we just share?

scheds/rust/scx_mitosis/src/main.rs
likewhatevs: My thinking is no -- for many use cases (i.e. my bazzite handheld), this would just be log noise. For those use cases where this would be used though, this would lessen (possibly eliminate, likely only lessen because of how rust logging works, IIUC) the code delta to enable the use of this flag.

scheds/rust/scx_mitosis/src/main.rs
likewhatevs: tbh not sure. i *think* for these 4 that makes sense, but at the same time there's a constant tension between having schedulers be independent and sharing code. idk.

PR #3026, opened 2025-11-10
Add `cpu.max` Support for the SCX Scheduler

multics69:
This patch set adds **`cpu.max` support** [[1]](#1-reference) to the **SCX scheduler**.  
It introduces a new library for CPU bandwidth control and integrates it with LAVD.  
The series consists of four major parts:

1. `cpu.max` library implementation  
2. ATQ/rbtree modifications required by the library  
3. LAVD integration for `cpu.max`  
4. Utility scripts for debugging and testing  

---

## 1. Overview of the `cpu.max` Feature

The `cpu.max` interface controls CPU bandwidth for non-root cgroups using three parameters:

- `$QUOTA`
- `$PERIOD`
- `$BURST`

A cgroup may consume up to `$QUOTA` runtime in each `$PERIOD`.  
Using `max` for `$QUOTA` removes the limit.  
At the end of each period, unused time is discarded unless `$BURST` is specified—allowing carryover up to `$BURST`.

---

## 2. Design Overview

Our implementation follows the cgroup v2 `cpu.max` semantics but differs from the in-kernel CFS version for **efficiency** and **simplicity**.

### (1) Interpreting Quota and Period as CPU Utilization

We treat `quota / period` as CPU utilization rather than time-based replenishment.  

For example, if a cgroup has:
```
quota = 20000 µs
period = 200000 µs
```
it may use up to **10% of a single CPU**.  
All cgroups are replenished by a **single BPF timer**, avoiding one timer per cgroup.

### (2) Eventual Quota Enforcement via Task Admission Control

To reduce overhead in the critical path, we enforce quota **eventually** rather than immediately.  

- In CFS, quota is checked when selecting a task.  
- In SCX, we check it **on enqueue** (*task admission control*).

Tasks from throttled cgroups are deferred to a **backlog queue** instead of
DSQ. When quota is replenished by the BPF timer, deferred tasks are **moved
back** to their DSQ, ensuring zero additional overhead in `ops.dispatch()`.

---

## 3. API for BPF Schedulers

BPF schedulers interact with the library using the following API.

### Initialization

- `scx_cgroup_bw_lib_init()` → called in `ops.init()`  
- `scx_cgroup_bw_init()` / `scx_cgroup_bw_exit()` → called in `ops.cgroup_init()` / `ops.cgroup_exit()`  
- `scx_cgroup_bw_set()` → called when `cpu.max` parameters change  

### Runtime Checks

Before enqueueing (`ops.enqueue()`) or selecting (`ops.select_cpu()`), call  
`scx_cgroup_bw_throttled()` to check if the cgroup is throttled.

If throttled:
- Defer task using `scx_cgroup_bw_put_aside()`  
- When replenished, call `scx_cgroup_bw_reenqueue()` (from `ops.dispatch()`)

Re-enqueued tasks use scheduler-defined policies (e.g., time slice, vtime, DSQ).  
A callback must be registered with:

```c
REGISTER_SCX_CGROUP_BW_ENQUEUE_CB(eqcb)
```

where:

```c
int eqcb(u64 pid);
```

If a throttled task is dequeued before being unthrottled (e.g., due to `sched_setaffinity(2)`),  
the scheduler should call `scx_cgroup_bw_cancel()` to maintain proper state.

### Reporting the Used Time

A task must report its consumed execution time periodically (`ops
.tick()`) and when it stops (`ops.stopping()`) using `scx_cgroup_bw_consume()`.

### API Summary

See `scheds/include/lib/cgroup.h` for detailed documentation.

```c
int scx_cgroup_bw_lib_init(struct scx_cgroup_bw_config *config);
int scx_cgroup_bw_init(struct cgroup *cgrp, struct scx_cgroup_init_args *args);
int scx_cgroup_bw_exit(struct cgroup *cgrp);
int scx_cgroup_bw_set(struct cgroup *cgrp, u64 period_us, u64 quota_us, u64 burst_us);
int scx_cgroup_bw_throttled(struct cgroup *cgrp);
int scx_cgroup_bw_consume(struct cgroup *cgrp, u64 consumed_ns);
int scx_cgroup_bw_put_aside(struct task_struct *p, u64 taskc, u64 vtime, struct cgroup *cgrp);
int scx_cgroup_bw_reenqueue(void);
int scx_cgroup_bw_cancel(u64 taskc);
#define REGISTER_SCX_CGROUP_BW_ENQUEUE_CB(eqcb)
```

---

## 4. Prerequisites for BPF Schedulers

Schedulers must allocate their per-task context (e.g., `struct task_ctx` in LAVD) **in the BPF arena**,  
as the library internally uses **ATQ (Arena-based Task Queue)** for managing throttled tasks.

Each task context must begin with `struct scx_task_common` (defined in `scheds/include/lib/atq.h`)  
or reserve equivalent space (`struct atq_ctx` in LAVD).  
This structure embeds a red-black tree node and related state required by ATQ.

---

### (1) ATQ / Rbtree Changes
- lib/atq: factor out task insertion into scx_atq_insert_node
- lib/rbtree: add noalloc/nofree variants of the API
- lib/rbtree: turn rbtree_insert_mode from a per-insert into a per-tree attribute
- lib/rbtree: adjust inlining to pass verification
- lib/atq: only use embedded rbnodes on scx_atq_insert_*()
- lib/rbtree: remove RB_ALLOC check out of rb_insert codepath and update selftests
- lib/selftests: add selftests for embedded rbtree node
- lib/rbtree: initialize node color to red for embedded nodes
- include/lib: expose rb_integrity_check as public API
- lib/rbtree: wipe the values of ->left and ->right in embedded rbnode_t instances before use
- selftests/atq: expand and fix tests
- lib/selftests: exclude cgroup_bw.bpf.c from selftests
- lib/selftests: hardcode Clang version to be the system one
- include: fix userspace-side spinlock definitions
- lib/atq: change signature to require scx_task_common
- lib/atq: add atq_remove_node call
- include/atq: add ATQ lock/unlock calls
- lib/atq: add unlocked insert ATQ calls

### (2) `cpu.max` Library Implementation
- lib: cgroup_bw: Add skeleton for CPU bandwidth control.
- lib: cgroup_bw: Implement scx_cgroup_bw_lib_init().
- lib: cgroup_bw: Implement scx_cgroup_bw_init().
- lib: cgroup_bw: Implement scx_cgroup_bw_set().
- lib: cgroup_bw: Implement scx_cgroup_bw_exit().
- lib: cgroup_bw: Implement scx_cgroup_bw_throttled().
- lib: cgroup_bw: Implement scx_cgroup_bw_consume().
- lib: cgroup_bw: Implement scx_cgroup_bw_put_aside().
- lib: cgroup_bw: Implement replenish timer and cbw_reenqueue_cgroup().
- lib: cgroup_bw: Implement scx_cgroup_bw_cancel().

### (3) LAVD Integration
- scx_lavd: Initial integration with CPU bandwidth control.
- scx_lavd: Move task_ctx to the BPF arena.
- scx_lavd: Support cpu.max at enqueue-like paths.
- scx_lavd: Implement ops.deque() for throttling cancelling.

### (4) Scripts
- scripts: Add a script to get a cgroup path from it id
- scripts: Add a script to set up cgroups for basic testing of cpu.max.

---

## 6. Reference

[1] [Cgroup v2 Documentation — CPU controller](https://docs.kernel.org/admin-guide/cgroup-v2.html)

---

Signed-off-by: Emil Tsalapatis <etsal@meta.com>

### DIFF
--- a/scripts/cg_setup_test.sh
+++ b/scripts/cg_setup_test.sh
@@ -1,0 +1,319 @@
+#!/usr/bin/bash
+
+echo "+cpu" > /sys/fs/cgroup/cgroup.subtree_control
+
+# ==========================================================
+# Config 01: a single level, half CPU
+mkdir -p /sys/fs/cgroup/test01_l1
+
+echo "+cpu" > /sys/fs/cgroup/test01_l1/cgroup.subtree_control
+
+echo "50000 100000" > /sys/fs/cgroup/test01_l1/cpu.max
+
+# -------------
+# Test 01-01
+# echo $$ > /sys/fs/cgroup/test01_l1/cgroup.procs
+
+# ==========================================================
+# Config 02: a single level, two CPUs
+mkdir -p /sys/fs/cgroup/test02_l1
+
+echo "+cpu" > /sys/fs/cgroup/test02_l1/cgroup.subtree_control
+
+echo "200000 100000 " > /sys/fs/cgroup/test02_l1/cpu.max
+
+# -------------
+# Test 02-01
+# echo $$ > /sys/fs/cgroup/test02_l1/cgroup.procs
+
+# ==========================================================
+# Config 03: two-level, half CPU
+mkdir -p /sys/fs/cgroup/test03_l1
+mkdir -p /sys/fs/cgroup/test03_l1/l2-a
+mkdir -p /sys/fs/cgroup/test03_l1/l2-b
+mkdir -p /sys/fs/cgroup/test03_l1/l2-c
+
+echo "+cpu" > /sys/fs/cgroup/test03_l1/cgroup.subtree_control
+echo "+cpu" > /sys/fs/cgroup/test03_l1/l2-a/cgroup.subtree_control
+echo "+cpu" > /sys/fs/cgroup/test03_l1/l2-b/cgroup.subtree_control
+echo "+cpu" > /sys/fs/cgroup/test03_l1/l2-c/cgroup.subtree_control
+
+echo "50000 100000" > /sys/fs/cgroup/test03_l1/cpu.max
+echo "50000 100000" > /sys/fs/cgroup/test03_l1/l2-a/cpu.max
+echo "50000 100000" > /sys/fs/cgroup/test03_l1/l2-b/cpu.max
+echo "50000 100000" > /sys/fs/cgroup/test03_l1/l2-c/cpu.max
+
+# -------------
+# Test 03-01: running on a single cgroup
+# echo $$ > /sys/fs/cgroup/test03_l1/l2-a/cgroup.procs
+
+# -------------
+# Test 03-02 (l1, l2): running on two cgroups
+# echo $$ > /sys/fs/cgroup/test03_l1/l2-a/cgroup.procs
+# echo $$ > /sys/fs/cgroup/test03_l1/l2-b/cgroup.procs
+
+# -------------
+# Test 03-03: running on three cgroups
+# echo $$ > /sys/fs/cgroup/test03_l1/l2-a/cgroup.procs
+# echo $$ > /sys/fs/cgroup/test03_l1/l2-b/cgroup.procs
+# echo $$ > /sys/fs/cgroup/test03_l1/l2-c/cgroup.procs
+
+
+# ==========================================================
+# Config 04: two-level, two CPUs
+mkdir -p /sys/fs/cgroup/test04_l1
+mkdir -p /sys/fs/cgroup/test04_l1/l2-a
+mkdir -p /sys/fs/cgroup/test04_l1/l2-b
+
+echo "+cpu" > /sys/fs/cgroup/test04_l1/cgroup.subtree_control
+echo "+cpu" > /sys/fs/cgroup/test04_l1/l2-a/cgroup.subtree_control
+echo "+cpu" > /sys/fs/cgroup/test04_l1/l2-b/cgroup.subtree_control
+
+echo "200000 100000" > /sys/fs/cgroup/test04_l1/cpu.max
+echo "200000 100000" > /sys/fs/cgroup/test04_l1/l2-a/cpu.max
+echo "200000 100000" > /sys/fs/cgroup/test04_l1/l2-b/cpu.max
+
+# -------------
+# Test 04-01: running on a single cgroup
+# echo $$ > /sys/fs/cgroup/test04_l1/l2-a/cgroup.procs
+
+# -------------
+# Test 04-02: running on two cgroups
+# echo $$ > /sys/fs/cgroup/test04_l1/l2-a/cgroup.procs
+# echo $$ > /sys/fs/cgroup/test04_l1/l2-b/cgroup.procs
+
+
+# ==========================================================
+# Config 05: three-level, half CPU
+mkdir -p /sys/fs/cgroup/test05_l1
+mkdir -p /sys/fs/cgroup/test05_l1/l2-a
+mkdir -p /sys/fs/cgroup/test05_l1/l2-a/l3-x
+mkdir -p /sys/fs/cgroup/test05_l1/l2-a/l3-y
+mkdir -p /sys/fs/cgroup/test05_l1/l2-b
+
+echo "+cpu" > /sys/fs/cgroup/test05_l1/cgroup.subtree_control
+echo "+cpu" > /sys/fs/cgroup/test05_l1/l2-a/cgroup.subtree_control
+echo "+cpu" > /sys/fs/cgroup/test05_l1/l2-a/l3-x/cgroup.subtree_control
+echo "+cpu" > /sys/fs/cgroup/test05_l1/l2-a/l3-y/cgroup.subtree_control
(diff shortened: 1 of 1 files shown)


htejun:
Let's land this one after the november release is cut.

lib/atq.bpf.c
```diff
@@ -28,102 +28,105 @@ u64 scx_atq_create_internal(bool fifo, size_t capacity)
...
-	if (ret) {
-		rb_node_free(atq->tree, node);
-		return ret;
-	}
-
-	if (unlikely(atq->size == atq->capacity)) {
-		ret = -ENOSPC;
-		goto error;
-	}
+	if ((vtime == SCX_ATQ_FIFO) != atq->fifo)
```
hodgesds, line 40: This seems like a weird API, it's doing FIFO by passing `SCX_ATQ_FIFO` as vtime?

lib/selftests/Makefile
```diff
@@ -9,14 +10,16 @@ BPF_CFLAGS=-O2 -g -target bpf -D__TARGET_ARCH_x86 -mcpu=v3 -Wall
 BPF_CFLAGS+=-I/usr/include/bpf -I/usr/include/$(shell uname -m)-linux-gnu 
 BPF_CFLAGS+=$(INCLUDES)
 
+CC=/usr/bin/clang
```
hodgesds: I think you can specify `CC=clang` and rely on the users $PATH as users may have clang installed elsewhere.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1016,15 +1176,15 @@ void BPF_STRUCT_OPS(lavd_runnable, struct task_struct *p, u64 enq_flags)
 	 */
 	if (is_monitored) {
 		p_taskc->waker_pid = waker->pid;
-		__builtin_memcpy_inline(p_taskc->waker_comm, waker->comm,
-					TASK_COMM_LEN);
+		for (i = 0; i < TASK_COMM_LEN && can_loop; i++)
```
hodgesds, line 1178: Why not `bpf_probe_read_kernel_str`?

multics69: Thanks for the suggestion, @hodgesds !

When using `bpf_probe_read_kernel_str`, the verifier spits the following error:

```
; bpf_probe_read_kernel_str((void *)p_taskc->waker_comm, TASK_COMM_LEN, waker->comm); @ main.bpf.c:1179                                                                                                                                                                                                                        
87: (07) r9 += 236                    ; R9_w=scalar(id=2+236)                                                                                                                                                                                                                                                                  
88: (bf) r1 = addr_space_cast(r9, 0, 1)       ; R1_w=arena R9_w=scalar(id=2+236)                                                                                                                                                                                                                                               
89: (b4) w2 = 16                      ; R2_w=16                                                                                                                                                                                                                                                                                
90: (85) call bpf_probe_read_kernel_str#115                                                                                                                                                                                                                                                                                    
R1 type=arena expected=fp, pkt, pkt_meta, map_key, map_value, mem, ringbuf_mem, buf, trusted_ptr_  
```

In short, the arena pointer is not an expected destination of `bpf_probe_read_kernel_str`. 

In contrast, when using `__builtin_memcpy_inline()` (the original code), the verifier complains as follows:

```
; __builtin_memcpy_inline(p_taskc->waker_comm, waker->comm, TASK_COMM_LEN); @ main.bpf.c:1179
88: (61) r2 = *(u32 *)(r3 +12)        ; R2_w=scalar(smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) R3_w=trusted_ptr_task_struct(off=3248)
89: (63) *(u32 *)(r9 +248) = r2
R9 invalid mem access 'scalar'
```

So it is not allowed to copy between the kernel memory and arena memory.

---

There are two possible workarounds -- (1) copying byte by byte (the current change) or (2) using a stack memory as a staging buffer (kernel memory --> stack memory --> arena memory), and the first one was chosen.

Does it make sense to you, @hodgesds ?

hodgesds: Yeah that makes sense, kind of unfortunate that it doesn't work with arenas.

lib/atq.bpf.c
etsal: The SCX_ATQ_FIFO value is only visible internally, it's used only by scx_atq_insert() when it calls into scx_atq_insert_vtime().

lib/selftests/Makefile
multics69: Dropped "/usr/bin/" as @hodgesds suggested.

JakeHillion:
> Short note: This feature will temporarily break schedulers that use the old ATQ API, notably p2dq. There will be a subsequent patch that fixes this issue.

This is kind of a pain, CI is failing on main because of this. Can we please make sure to land the fixes today?

sirlucjan:
This is fix: https://github.com/sched-ext/scx/pull/3038 @JakeHillion

PR #3028, opened 2025-11-11
scxtop: Do not use is_multiple_of() to avoid an E0658 error.

multics69:
`is_multiple_of()` is recently introduced, so it could cause an unstable feature error (E0658) in not-so-up-to-date toolchains. So let's replace `X.is_multiple_of(Y)` to `X % Y == 0` to avoid the compilation error.


multics69:
Ha, it generates a lint error: 

https://rust-lang.github.io/rust-clippy/master/index.html#manual_is_multiple_of

So, fixed the lint error by adding `#[allow(clippy::manual_is_multiple_of)]`.

hodgesds (approved):
Thanks for fixing!

PR #3031, opened 2025-11-11
versions: bump versions for 1.0.18 release

JakeHillion:


sirlucjan:
@JakeHillion Maybe you should update version in meson.build.

PR #3016, opened 2025-11-07
scx_lavd: Optimize CPU mask iteration with __builtin_ctzll

bboymimi:
scx_lavd: Optimize CPU mask iteration with __builtin_ctzll

Replace bit-by-bit iteration through CPU masks with __builtin_ctzll()
intrinsic across three critical paths: pick_most_loaded_cpu(),
init_per_cpu_ctx(), and collect_sys_stat().

The optimization uses count-trailing-zeros to jump directly to set bits
instead of checking all 64 bits sequentially, particularly beneficial
for sparse CPU masks common in scheduling scenarios.

Measurement:

i). bpf_for(j, 0, 64) measurement

Ther are three loops inside collect_sys_stat:

1. number of domains (3 domains on my system: big, medium, little)
2. 8 cpumask chunks (# cpu mask = LAVD_CPU_ID_MAX/64 = 512/64)
3. bpf_for(j, 0, 64)

The first measurement is around the inner bpf_for by taking timestamps
with bpf_ktime_get_ns() before the bpf_for loop and print the delta
after the loop via, for example:
```c
+       if (cnt < 10000ULL)
+               before = bpf_ktime_get_ns();
+       bpf_for(k, 0, 64) {
+               if (!cpumask)
+                       break;
+               j = __builtin_ctzll(cpumask);
+               cpu = (i * 64) + j;
+               cpumask &= cpumask - 1;
+               if (cpu >= __nr_cpu_ids)
+                       break;
+               cpdomc->nr_queued_task += scx_bpf_dsq_nr_queued(cpu_to_dsq(cpu));
+       }
+       if (cnt++ < 10000ULL)
+               bpf_printk("collect_sys_stat_lat: %llu", time_delta(bpf_ktime_get_ns(), before));
```
Then, the following table is generated by the two 10k data samples
captured before and after the optimization:

Performance improvement (10k samples) [1]:
|               | Before   | After    | Change (%)  |
|---------------|----------|----------|-------------|
| Mean (ns)     | 375.59   | 80.72    | -78.51%     |
| Median (ns)   | 333.00   | 38.00    | -88.59%     |
| Min (ns)      | 74       | 11       | -85.14%     |
| Max (ns)      | 10837    | 8130     | -24.98%     |
| Std Dev       | 316.12   | 192.71   | -39.04%     |
| P25 (ns)      | 232.00   | 16.00    | -93.10%     |
| P75 (ns)      | 386.00   | 53.00    | -86.27%     |
[...]

PR #3032, opened 2025-11-12
scx_utils, scx_lavd: Use a portable ctzll().

multics69:
[PR 3016](https://github.com/sched-ext/scx/pull/3016) uses __builtin_ctzll() for performance optimzation. However, unfortunately, the BPF backend of clang on ARM64 does not support lowering __builtin_ctzll() intrinsics, causing the following compile-time error. 

> fatal error: error in backend: unimplemented opcode: 191

This PR fixes the problem and consists of three parts:
- Passing the target architecture definition (e.g., __SCX_TARGET_ARCH_x86) for BPF code.
- Add a portable implementation of __builtin_ctzll() based on the De Bruijn sequence.
- Replace __builtin_ctzll() to ctzll().

The changes were tested on ARM64 with clang 18.1.8.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
+++ b/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
@@ -6,6 +6,8 @@
 #ifndef __LAVD_H
 #define __LAVD_H
 
+#include <scx/common.bpf.h>
+
 /*
  * common macros
  */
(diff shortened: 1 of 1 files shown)


scheds/include/scx/common.bpf.h
```diff
@@ -748,6 +748,69 @@ static inline u64 __sqrt_u64(u64 x)
...
+	 * the constant 0x03f79d71b4cb0a89ULL is carefully chosen to be a
+	 * De Bruijn sequence, ensuring no collisions in the table index.
+	 */
+	index = (lowest_bit * DEBRUIJN_CONSTANT) >> 58;
+
+	/*
+	 * Lookup in a precomputed table. No collision is guaranteed by the
+	 * De Bruijn property.
+	 */
+	lt = MEMBER_VPTR(lookup_table, [index]);
```
etsal, line 810: I assume we're using the macro bc it includes the automatic bounds check?

etsal (approved):
TIL, really interesting addition. Is the root problem that the ARM JIT does not implement the instruction? AFAIK LLVM only generates code for the BPF bytecode VM, so it doesn't matter on which architecture we end up running the object.

scheds/include/scx/common.bpf.h
multics69: > I assume we're using the macro bc it includes the automatic bounds check?

Yes. Without this, the verifier complains. :-(

multics69:
> TIL, really interesting addition. Is the root problem that the ARM JIT does not implement the instruction? AFAIK LLVM only generates code for the BPF bytecode VM, so it doesn't matter on which architecture we end up running the object.

After some more digging, I (actaully Gavin Guo) found that clang on x86 translates `__builtin_ctzll()` into the De Bruijn sequence based emulation using BPF bytecode. The clang's emulation is essentially the same as what I did, while it uses a different magic constant and lookup table. In contrast, clang on ARM64 does not generate such an emulation code and just gives up.

Besides the performance, in my opinion, the real value is the elimination of a loop that is often the source of E2BIG error!

PR #3033, opened 2025-11-12
scx_lavd: Lower the verifier pressure.

multics69:
scx_lavd has suffered from verifier errors in some kernel/compiler/architecture combinations, failing with `-E2BIG` (the BPF program is too complex) or `-EACCESS` (the stack consumption is beyond 512 bytes).

This PR aims to lower the verifier pressure:
- For the `-E2BIG` error
    - Replace a loop-based pick_any_bit() to a loop-less version using ctzll().
    - Disable inlining a looo-heavy pick_most_loaded_cpu() function.

- For `-EACCESS` error
    - Reduce the size of `struct pick_ctx`, which is allocated on the stack.
    -  Clean up stack variables in `lavd_select_cpu()`.

No functional changes are intended.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -529,27 +529,23 @@ static void update_stat_for_refill(struct task_struct *p,
 s32 BPF_STRUCT_OPS(lavd_select_cpu, struct task_struct *p, s32 prev_cpu,
 		   u64 wake_flags)
 {
-	bool found_idle = false;
-	struct task_ctx *taskc = get_task_ctx(p);
-	struct cpu_ctx *cpuc_cur = get_cpu_ctx();
-	struct cpu_ctx *cpuc;
-	u64 dsq_id, nr_queued = 0;
-	s32 cpu_id;
 	struct pick_ctx ictx = {
 		.p = p,
-		.taskc = taskc,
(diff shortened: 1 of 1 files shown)


multics69:
> This looks good, but have you tried setting -Os or -Oz in BPF_BASE_CFLAGS for lavd? Maybe in this way clang can already produce smaller code without adding any extra overhead or requiring code changes (in some cases it might even bring better performance).

Thanks for the review, @arighi ! I didn't but I found that -E2BIG is more related to the number of branch instructions (especially loop) rather than the actual binary size. That makes sense considering the search space that the verifier should explore is exponentially increases as the number of branch instructions grows.

multics69:
Currently, CI fails due to the `veristat`. More specifically, `veristat` fails at `lavd_dispatch`. However, I found that there is no actual verification error when loading the scx_lavd on all the tested kernels. After some more digging, the culprit is `veristat` itself. `lavd_dispatch` fails only with `-o csv` option. When giving an extra `-v` or `-d` or giving a function name, `lavd_dispatch` as an argument of `veristat`, it works fine.

@JakeHillion -- What do you think? Do you think if it is okay to merge the PR and exempt lavd from veristat for a while until the bug is fixed?

multics69:
Hmm... `veristat` fails for `scx_lavd` of PR https://github.com/sched-ext/scx/pull/3026 . Let me merge this PR first, then fix the `veristat` problem later.

PR #3044, opened 2025-11-14
scx_lavd: Revise the use of per-CPU DSQ.

multics69:
Now, the use of per-CPU DSQ gets complicated because the --per-cpu-dsq and --pinned-slice-us options were introduced. It is easy to make a mistake when checking whether per-CPU DSQs are used. We introduce utility functions around per-CPU DSQ and apply them to the code. We also found a few bugs where per-CPU DSQs were not correctly checked and fixed them.


daidavid:
Thanks for untangling some of the logic with Per CPU and Cpdom DSQs. LGTM, I also did some sanity testing with it and doesn't seem like it broken anything.

multics69:

Issue #3047, opened 2025-11-14
p2dq is broken on 6.12


JakeHillion:
Starting p2dq on 6.12, which is LTS, gives the following error:

```
cyclone% sudo target/release/scx_p2dq
18:36:50 [INFO] Running scx_p2dq (build ID: 1.0.23-ga885d393 x86_64-unknown-linux-gnu)
18:36:50 [INFO] Single LLC detected - disabling multi-LLC optimizations
18:36:50 [INFO] DSQ[0] slice_ns 100000
18:36:50 [INFO] DSQ[1] slice_ns 3200000
18:36:50 [INFO] DSQ[2] slice_ns 6400000
18:36:50 [INFO] libbpf: struct_ops p2dq: member cgroup_set_bandwidth not found in kernel, skipping it as it's set to zero

18:36:50 [INFO] libbpf: struct_ops p2dq: member priv not found in kernel, skipping it as it's set to zero

18:36:50 [WARN] libbpf: prog 'p2dq_dequeue': BPF program load failed: -EINVAL

18:36:50 [WARN] libbpf: prog 'p2dq_dequeue': -- BEGIN PROG LOAD LOG --
Validating scx_atq_cancel() func#3...
55: R1=scalar() R10=fp0
; int scx_atq_cancel(scx_task_common __arg_arena *taskc) @ atq.bpf.c:193
55: (b4) w7 = 0                       ; R7_w=0
56: (bf) r9 = addr_space_cast(r1, 0, 1)       ; R1=scalar() R9_w=arena
; atq = taskc->atq; @ atq.bpf.c:203
57: (79) r6 = *(u64 *)(r9 +48)        ; R6_w=scalar() R9_w=arena
; if (!atq) @ atq.bpf.c:204
58: (15) if r6 == 0x0 goto pc+65      ; R6_w=scalar(umin=1)
59: (b4) w7 = -95                     ; R7_w=0xffffffa1
; if (CONFIG_NR_CPUS > _Q_MAX_CPUS) @ bpf_arena_spin_lock.h:506
60: (18) r2 = 0xffffcf09c0e2e000      ; R2_w=map_value(map=bpf_bpf.kconfig,ks=4,vs=9)
62: (79) r2 = *(u64 *)(r2 +0)         ; R2_w=384
63: (25) if r2 > 0x4000 goto pc+21    ; R2_w=384
[...]

Issue #3046, opened 2025-11-14
lavd is broken on 6.12


JakeHillion:
Starting lavd on 6.12, which is LTS, gives the following error:

```
cyclone% sudo target/release/scx_lavd
18:32:06 [INFO] Autopilot mode is enabled.
18:32:06 [INFO] Energy model won't be used for CPU preference order.
18:32:06 [INFO] Opts {
    autopilot: true,
    autopower: false,
    performance: false,
    powersave: false,
    balanced: false,
    slice_max_us: 5000,
    slice_min_us: 500,
    mig_delta_pct: 0,
    pinned_slice_us: None,
    preempt_shift: 6,
    cpu_pref_order: "",
    no_use_em: true,
    no_futex_boost: false,
    no_preemption: false,
    no_wake_sync: false,
    no_slice_boost: false,
    per_cpu_dsq: false,
    enable_cpu_bw: false,
    no_core_compaction: false,
    no_freq_scaling: false,
    stats: None,
    monitor: None,
    monitor_sched_samples: None,
    verbose: 0,
    version: false,
    help_stats: false,
    libbpf: LibbpfOpts {
        relaxed_maps: None,
        pin_root_path: None,
        kconfig: None,
        btf_custom_path: None,
        bpf_token_path: None,
    },
    topology: None,
}
18:32:06 [INFO] capacity bound:  1024 (25%)
  primary CPUs:  [0]
  overflow CPUs: [1, 2, 3]
18:32:06 [INFO] capacity bound:  4096 (100%)
  primary CPUs:  [0]
  overflow CPUs: [1, 2, 3]
18:32:06 [WARN] libbpf: struct_ops init_kern lavd_ops: Cannot find member cgroup_set_bandwidth in kernel BTF

18:32:06 [WARN] libbpf: failed to load BPF skeleton 'bpf_bpf': -EOPNOTSUPP

Error: Failed to load BPF program

Caused by:
    Operation not supported (os error 95)
```

Veristat also reports this. For now 6.12 CI testing is disabled for lavd, but we need to decide whether lavd no longer supports 6.12 or if we want to fix this before the next release.

multics69:
https://github.com/sched-ext/scx/pull/3068 should fix the problem described above when --enable-cpu-bw is not given.

However, when --enable-cpu-bw is given, it still causes the following verification error (as same as P2DQ):

```
returning from callee:
[...]

PR #3048, opened 2025-11-14
lavd/p2dq: disable 6.12 CI testing

JakeHillion:
lavd and p2dq are both currently broken on 6.12. Created #3046 and #3047 to track this, but in the meantime we need to disable their CI on 6.12 to get things working again.

Test plan:
- CI - this gets the Nix build-and-test passing, but naturally the Ubuntu runners are failing. Will follow up.

### DIFF
--- a/scheds/rust/scx_lavd/Cargo.toml
+++ b/scheds/rust/scx_lavd/Cargo.toml
@@ -6,6 +6,11 @@ edition = "2021"
 description = "A Latency-criticality Aware Virtual Deadline (LAVD) scheduler based on sched_ext, which is a Linux kernel feature which enables implementing kernel thread schedulers in BPF and dynamically loading them. https://github.com/sched-ext/scx/tree/main"
 license = "GPL-2.0-only"
 
+[package.metadata.scx.ci.kernel]
+blocklist = [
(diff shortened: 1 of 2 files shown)


etsal:
For lavd, I think this is the only way to get veristat passing on 6.12, unless we add build flags that remove cgroup_bw.bpf.c and all invocations into it from the scheduler just for 6.12. For p2dq as I mentioned in the issue I don't see how we can avoid turning off 6.12.

JakeHillion:
> For lavd, I think this is the only way to get veristat passing on 6.12, unless we add build flags that remove cgroup_bw.bpf.c and all invocations into it from the scheduler just for 6.12. For p2dq as I mentioned in the issue I don't see how we can avoid turning off 6.12.

Yeah, this PR isn't related to whether those schedulers _should_ run on 6.12, it's to unbreak main. If we want to follow up with a more targeted disable of lavd's veristat we can do that, but we need to unbreak main first.

JakeHillion:
After this lands we can land #3050 then the CI will finally be green.

multics69:
Thanks @JakeHillion ! LGTM

multics69 (approved):

PR #3053, opened 2025-11-16
scx_lavd: Revise the time slice calculation when pinned_slice_ns is on.

multics69:
This PR made three changes:
- When the system is heavily loaded, the system-wide time slice could be smaller than the pinned_slice_ns. In this case, choose the system-wide time slice.
- At `ops.tick()`, the task's time slice was set to 0 when its accumulated time slice is greater than pinned_slice_ns (`taskc->acc_runtime`). However, `taskc->acc_runtime` is the total execution time accumulated once the task becomes runnable. Instead, calculate the real execution from `cpuc->running_clk`.
- Merge the time slice shrink logic into `shrink_boosted_slice_remote()` for readability.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/preempt.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/preempt.bpf.c
@@ -278,27 +278,44 @@ int shrink_boosted_slice_remote(struct cpu_ctx *cpuc, u64 now)
 }
 
 __hidden
-void shrink_boosted_slice_at_tick(struct task_struct *p,
-					 struct cpu_ctx *cpuc, u64 now)
+void shrink_slice_at_tick(struct task_struct *p, struct cpu_ctx *cpuc, u64 now)
 {
-	u64 dur, new_slice = 0;
+	u64 ub, duration, new_slice;
 
 	/*
-	 * Shrink the time slice of the slice-boosted task into a regular
-	 * slice by reducing the remaining time slice to the regular one.
+	 * Calculate the upper bound of running this task
+	 * since it is scheduled this time.
 	 */
-	reset_cpu_flag(cpuc, LAVD_FLAG_SLICE_BOOST);
-
-	dur = time_delta(now, cpuc->running_clk);
-	if (sys_stat.slice > dur)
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_lavd/src/bpf/preempt.bpf.c
```diff
@@ -278,28 +278,55 @@ int shrink_boosted_slice_remote(struct cpu_ctx *cpuc, u64 now)
...
+		 */
+		p->scx.slice = 0;
+		reset_cpu_flag(cpuc, LAVD_FLAG_SLICE_BOOST);
+		cpuc->nr_preempt++;
+		return true;
+	}
+	/* Note: duration < ub */
 
-	cpuc->nr_preempt++;
+	if (test_cpu_flag(cpuc, LAVD_FLAG_SLICE_BOOST)) {
```
daidavid: I'm curious if it makes sense to remove this section since eventually the previous check will kick in when the task has ran long enough on subsequent ticks. It does result in potentially lagging in kicking out the longer task if the tick isn't frequent enough. It's not much of an issue on a 1000HZ system, but for 250HZ, this can lag by up to 4ms. WDYT?

multics69:
> > When the system is heavily loaded, sys_stat.slice would be even smaller
> > than the given pinned_slice_ns.
> 
> Overall changes look fine, but can you describe this usecase a bit more? How does a busy system result in smaller sys_stat.slice?

`calc_sys_time_slice()` decreases the slice (`sys_stat.slice`) and the number of queued tasks increases. So, when the system is heavily loaded, there will be more queued tasks, so it could be shorter than the `pinned_slice_ns`.

scheds/rust/scx_lavd/src/bpf/preempt.bpf.c
multics69: That makes sense. Regardless of whether a task is slice-boosted or not, it should not go beyond the current upper bound. I will change it.

multics69: @daidavid -- I updated it. Please take a look.

PR #3055, opened 2025-11-16
scx_lavd: Tracking the number of pinned tasks on a CPU precisely.

multics69:
Previously, we didn't correctly track the number of pinned tasks when a task changed its pin location. Now, we track the pinned CPU ID in task_ctx (pinned_cpu_id), which is set to -ENOENT when the task is not pinned or not runnable. Then increment and decrement the count when the task becomes enqueued and dequeued, respectively.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -722,8 +722,13 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 	/*
 	 * Increase the number of pinned tasks waiting for execution.
 	 */
-	if (is_pinned(p)) {
+	if (is_pinned(p) && (taskc->pinned_cpu_id == -ENOENT)) {
+		taskc->pinned_cpu_id = cpu;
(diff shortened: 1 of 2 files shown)


etsal:
Am I missing something, or are we never setting pinned_cpu_id? If we need to add it, we can do it during enqueue() to also account for affinity changes.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -722,8 +722,13 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 	/*
 	 * Increase the number of pinned tasks waiting for execution.
 	 */
-	if (is_pinned(p)) {
+	if (is_pinned(p) && (taskc->pinned_cpu_id == -ENOENT)) {
+		taskc->pinned_cpu_id = cpu;
```
multics69, line 726: > Am I missing something, or are we never setting pinned_cpu_id? If we need to add it, we can do it during enqueue() to also account for affinity changes.

@etsal -- Exactly, it is set at ops.enqueue().

etsal: Ah, turns out I'm blind. Thanks!

PR #3062, opened 2025-11-18
scx_lavd: Turn on --pinned-slice-us option by default.

multics69:
Managing pinned tasks in a per-CPU DSQ is advantageous. When pinned tasks are on the shared per-domain DSQ, it takes longer to pop a task from the DSQ because the CPU must skip pinned tasks that cannot be executed on the CPU. That stretches the critical section, increasing the contention on the shared DSQ. So, let's avoid this problem by managing per-CPU DSQ for pinned tasks.


hodgesds (approved):
LGTM

daidavid (approved):

PR #3074, opened 2025-11-20
scx_lavd: Move task_ctx to lavd.bpf.h from intf.h

multics69:
Previously, struct task_ctx was defined in intf.h because struct task_ctx information was passed over the ring buffer for introspection (--monitor-sched-samples). However, intf.h is shared between BPF code and userspace code, so adding a new field of a struct type can complicate things due to header file dependencies.

To remove such a header file dependency, we made the following changes:
- Move struct task_ctx to lavd.bpf.h, which is included only in the BPF code.
- Extend struct task_ctx_x to contain all the necessary information for introspection.
- Change the type of atq in task_ctx from a dummy type (struct atq_ctx) to a real type (struct scx_task_common).

There is no functional change intended.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/intf.h
+++ b/scheds/rust/scx_lavd/src/bpf/intf.h
@@ -21,28 +21,11 @@ typedef signed long s64;
 
 typedef int pid_t;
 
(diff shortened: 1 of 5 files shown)


multics69:

PR #3076, opened 2025-11-20
scx_p2dq: Fix high wakeup latency on idle systems with DHQ enabled

hodgesds:
On mostly idle systems, tasks were experiencing 100+ ms wakeup latencies. Analysis of trace data revealed that affinitized tasks were being enqueued to per-CPU affinity queues (affn_dsq) that were only checked by their assigned CPU, not by other idle CPUs in the same affinity mask.

For example, a task enqueued to CPU 2's affn_dsq would wait for CPU 2 to become available, even when CPU 3 (also in the task's affinity mask and in the same LLC) was idle. This caused severe latency on mostly idle systems where some CPUs had queued work while others sat idle.

This commit adds affn_dsq work stealing to the dispatch path. During vtime comparison, each CPU now checks other CPUs' affn_dsq within the same LLC and dispatches the task with the lowest vtime. This enables idle CPUs to steal affinitized work from busy CPUs while maintaining:

- Vtime fairness (only steal if task has lowest vtime)
- Affinity constraints (only steal tasks that can run on current CPU)
- LLC locality (only check within same LLC, same vtime domain)

The fix allows proper work conservation on mostly idle systems without sacrificing fairness or violating affinity constraints.

Reported-by: trace analysis showing 103ms wakeup latency (PID 1072628)
Tested-on: System with DHQ enabled showing 23% of wakeup events as outliers

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -1922,6 +1922,42 @@ static void p2dq_dispatch_impl(s32 cpu, struct task_struct *prev)
 		}
 	}
 
+	// Check other CPUs' affn_dsq in same LLC for affinitized work stealing
+	// This prevents high wakeup latency when tasks are queued on busy CPUs
+	// but other CPUs in the affinity mask are idle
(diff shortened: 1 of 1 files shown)


arighi (approved):
Makes sense.

PR #3079, opened 2025-11-21
scx_lavd: Don't try to migrate a task from non-migratable DSQs.

multics69:
When pinned_slice_ns is on but per_cpu_dsq is not, per-CPU DSQ is used only for pinned tasks, which cannot be migrated to another CPU. In this case, we don't need to check the load of per-CPU DSQs of a domain.

To express the intention of the code more explicitly, add a utility function, is_per_cpu_dsq_migratable(), and use it in pick_most_loaded_dsq().


multics69:
Thanks @daidavid and @hodgesds for reviews!

PR #3086, opened 2025-11-23
scx_cosmos: Avoid unnecessary CPU wakeups

arighi:
In ops.enqueue() send CPU wakeup events only when a task is migrated to an idle CPU or when the task was previously sleeping and the wakeup wasn't already handled in ops.select_cpu().


hodgesds (approved):
LGTM

sirlucjan (approved):

PR #3093, opened 2025-11-25
scx_lavd: Revise the performance criticality metric.

multics69:
Instead of directly using the average runtime value (avg_runtime), let's use a bigger one between the average runtime (avg_runtime) vs. the current runtime accumulated (acc_runtime) to react faster.

In addition, revise the math used to calculate the threshold of performance criticality in update_thr_perf_cri().


multics69:

PR #3094, opened 2025-11-25
scx_lavd: Revise the load-balancing metrics.

multics69:
We revise the load-balancing metrics. Specifically, there are places where we did not count queued tasks on per-CPU DSQ or local DSQ. Revised those places in the following commits:
  - scx_lavd: Precisely count the queued tasks for a CPU.
  - scx_lavd: Use the utility function to count the # of queued tasks for a waker's CPU.

Also, when picking an idle CPU, we compare the load of the previous domain and the waker's domain. To correctly account for the load of a domain encompassing per-CPU/per-domain/local DSQs with different CPU computing capacity, we use the scaled load metric (sc_load) of a domain.
 scx_lavd: Revise the load comparison when deciding a sticky CPU/domain.
   - scx_lavd: Stay on the sticky CPU (if there is) when there is no idle CPU.
  - scx_lavd: Reduce the size of struct sticky_ctx.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
@@ -532,11 +532,8 @@ bool is_sync_waker_idle(struct pick_ctx * ctx, s64 *cpdom_id)
 	if (!can_run_on_cpu(ctx, ctx->sync_waker_cpu))
 		return false;
 
-	if (scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | ctx->sync_waker_cpu))
-		return false;
-
 	cpuc_waker = get_cpu_ctx_id(ctx->sync_waker_cpu);
-	if (!cpuc_waker || scx_bpf_dsq_nr_queued(cpuc_waker->cpdom_id))
+	if (!cpuc_waker || nr_queued_on_cpu(cpuc_waker))
 		return false;
 
 	if (nr_cpdoms > 1) {


scheds/rust/scx_lavd/src/bpf/idle.bpf.c
```diff
@@ -424,11 +425,12 @@ s32 find_sticky_cpu_and_cpdom(struct pick_ctx *ctx, s64 *sticky_cpdom)
...
-		q0 = sctx.cpuc_match[0]->cpdom_id; /* prev_cpu */
-		q1 = sctx.cpuc_match[1]->cpdom_id; /* sync_waker_cpu */
-		if (q0 != q1 &&
-		    (scx_bpf_dsq_nr_queued(q0) > scx_bpf_dsq_nr_queued(q1))) {
+		p0 = sctx.cpuc_match[0]; /* prev_cpu */
+		p1 = sctx.cpuc_match[1]; /* sync_waker_cpu */
+		d0 = MEMBER_VPTR(cpdom_ctxs, [p0->cpdom_id]);
+		d1 = MEMBER_VPTR(cpdom_ctxs, [p1->cpdom_id]);
+
+		if ((p0 != p1) && (d0 && d1) && (d0->sc_load > d1->sc_load)) {
```
daidavid, line 433: I know this is a bit tangential to the patch itself. But as I'm looking over the sc_load tracking and sc_time metrics while I'm reworking the time accounting for tasks, I don't see why we shouldn't be using the rq->clock_pelt to do the capacity/frequency scaling for us. I can understand if we want our own custom load tracking functions for accumulate/delay since there's room for us to tune. The clock_pelt should provide better accuracy and better granularity since it can leverage HW support if FIE exists on that platform.

daidavid (approved):

scheds/rust/scx_lavd/src/bpf/idle.bpf.c
multics69: There is no serious reason for that. Initially, it was unclear what needed to be measured. Later, the capacity/frequency scaling ideas were merged. If we are confident that clock_pelt serves our purpose with less complexity and enough flexibility, why not?

daidavid: So to fix IRQ time accounting, I'm using p->se.sum_exec_runtime as a proxy for since update_curr_scx in sched_ext occurs frequently enough that it doesn't severely affect the precision. Though that's a bit awkward since last_measure_clk will look out of sync with everything else, I'll put up a preliminary PR so we can continue discussion there. I think accessing the rq->clock* members requires holding on the rq lock, so we might have to cache those values in scx and update it in update_curr_scx as required.

PR #3095, opened 2025-11-25
document release process

JakeHillion:
I've been completing the release process for a while so understand the
joys (pain) of shipping sched_ext releases. Sharing this so the next
victim can have more information for the first time.

Test plan:
- Reading.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,141 @@
+This doc explains both how our releases work and the steps the developer performing the release needs to take.
+
+## What's a release?
+
+sched_ext cuts a monthly release of the monorepo. This currently has a semantic
+version versioning scheme where we always bump the last number. It may change in
+the future, the key here is it's monotonically increasing. These releases are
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,141 @@
...
+$ git tag -s -m  'vX.Y.Z' vX.Y.Z
+```
+
+This git tag is also important. It should come from the head of the PR you're
+merging, _not_ the merge commit into main. This is needed to best replicate the
+crates.io versions in the tagged version - you don't want to race with someone
+merging and tag from a Git commit that includes other changes that weren't in the
+published crate sources.
+
+Same story with the signature as above. If CachyOS aren't away of your GPG key
```
blezsan:
```suggestion
Same story with the signature as above. If CachyOS aren't aware of your GPG key
```

blezsan (approved):
LGTM

PR #3098, opened 2025-11-26
scx_lavd: Account for time stolen from tasks from steal/irq_time

daidavid:
In high traffic networking services, irq_time can account for up to 30% of the entire system's cpu time. The current time accounting does not take into account time stolen by IRQ or steal which can make tasks appear as though it ran longer on CPU than reality. This can skew vtime deadline calculations and other metrics.

For kernels with IRQ_TIME_ACCOUNTING and/or PARAVIRT_TIME_ACCOUNTING enabled, rq->clock_task already deducts irq/time_steal when accounting for task runtimes. p->se.sum_exec_runtime serves as a good proxy for rq->clock_task when used to determine task's runtime as it's frequently updated via update_curr_scx().

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -393,6 +393,7 @@ static void update_stat_for_running(struct task_struct *p,
 	reset_task_flag(taskc, LAVD_FLAG_IS_SYNC_WAKEUP);
 	taskc->last_running_clk = now;
 	taskc->last_measured_clk = now;
+	taskc->last_sum_exec_clk = task_exec_time(p);
 
 	/*
 	 * Reset task's lock and futex boost count
@@ -454,7 +455,7 @@ static void account_task_runtime(struct task_struct *p,
 				 struct cpu_ctx *cpuc,
 				 u64 now)
 {
-	u64 sus_dur, runtime, svc_time, sc_time;
+	u64 sus_dur, runtime, svc_time, sc_time, task_time, exec_delta;
 
 	/*
 	 * Since task execution can span one or more sys_stat intervals,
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -530,12 +531,19 @@ static void update_stat_for_refill(struct task_struct *p,
...
 				   struct cpu_ctx *cpuc)
 {
-	u64 now = scx_bpf_now();
+	/*
+	 * Unlike the other functions(tick/stopping) where we invoke
+	 * account_task_runtime, update_curr_scx() doesn't occur right before
+	 * dispatch, so the task_exec_time here can be slightly out of date.
+	 * A possible improvement could be to add update_curr_scx() before
+	 * calling into dispatch.
+	 */
```
multics69: In my understanding, `update_curr_scx()` is called before calling `ops.dispatch()` from `balance_one()` (see [this](https://github.com/torvalds/linux/blob/30f09200cc4aefbd8385b01e41bde2e4565a6f0e/kernel/sched/ext.c#L2067)). So it should be okay. \\^^/

scheds/rust/scx_lavd/src/bpf/util.bpf.c
```diff
@@ -329,3 +329,8 @@ u64 get_target_dsq_id(struct task_struct *p, struct cpu_ctx *cpuc)
 		return cpu_to_dsq(cpuc->cpu_id);
 	return cpdom_to_dsq(cpuc->cpdom_id);
 }
+
+u64 task_exec_time(struct task_struct __arg_trusted *p)
+{
+	return p->se.sum_exec_runtime;
```
multics69: I like the idea of repurposing `p->se.sum_exec_runtime`.

In my understanding, `p->se.sum_exec_runtime` does not increase when the system is suspended. If so, we can remove the code related to suspended duration tracking: `get_suspended_duration_and_reset()` , `get_suspended_duration_and_reset()`, `cpuc->online_clk`, and `cpuc->offline_clk`.

multics69 (changes requested):
Overall I like the idea. I left two minor comments.

scheds/rust/scx_lavd/src/bpf/util.bpf.c
multics69: Also, it would be good to add some explanation about `p->se.sum_exec_runtime` and `clock_task` and relationship with substracting IRQ time.

PR #3099, opened 2025-11-26
scx_lavd: Fix cpu_dsq_id typo in vtime comparison

bboymimi:
The second __COMPAT_scx_bpf_dsq_peek() call was mistakenly using cpu_dsq_id instead of cpdom_dsq_id, making the code to peek at the same dsq twice instead of comparing vtimes between the cpu dsq and the cpdom dsq as intended.


multics69 (approved):
Thanks @bboymimi ! This is a nice catch!

PR #3100, opened 2025-11-26
layered: Add a new NUMANode Matcher

xerothermic:
We have multi-socket systems with CPU pinned processes (one per numa node). We found it's better to separate these processes to different layers for better performance.

~~As part of the development, I noticed the GPU Matcher is not working (e.g. `[{"UsedGpuPid": true},{"IsGroupLeader":true}]`)
Had a patch to restore the original behavior (Finding the main thread via child thread pulling the GPU). Needs to re-visit this to see if it covers all intent use cases.~~ Dropping the GPU Matcher changes, as @etsal and I looked at this problem, and we can't re-pro the behavior at the moment.

### DIFF
--- a/scheds/rust/scx_layered/examples/numa_node.json
+++ b/scheds/rust/scx_layered/examples/numa_node.json
@@ -1,0 +1,74 @@
+[
+	{
+		"name": "stress-ng numa 0",
+		"comment": "Match stress-ng processes pinned to NUMA node 0",
+		"matches": [
+			[
+				{ "PcommPrefix": "stress-ng" },
+				{ "NumaNode": 0 }
+			]
+		],
+		"kind": {
+			"Confined": {
+				"cpus_range": [4, 8],
+				"util_range": [0.5, 0.9]
+			}
+		}
+	},
+	{
+		"name": "stress-ng numa 1",
+		"comment": "Match stress-ng processes pinned to NUMA node 1",
+		"matches": [
+			[
+				{ "PcommPrefix": "stress-ng" },
+				{ "NumaNode": 1 }
+			]
+		],
+		"kind": {
+			"Confined": {
+				"cpus_range": [4, 8],
+				"util_range": [0.5, 0.9]
+			}
+		}
+	},
+	{
+		"name": "other_numa0_tasks",
+		"comment": "Match other tasks pinned to NUMA node 0",
+		"matches": [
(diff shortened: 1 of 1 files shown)


likewhatevs (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -2689,6 +2687,21 @@ static __noinline bool match_one(struct layer *layer, struct layer_match *match,
...
+		/* Get the NUMA node context */
+		if (!(nodec = lookup_node_ctx(match->numa_node_id)))
+			return false;
+
+		/* Get the node's CPU mask */
+		if (!(node_cpumask = cast_mask(nodec->cpumask)))
+			return false;
+
+		/* Check if task's affinity is a subset of the NUMA node's CPUs */
+		return bpf_cpumask_subset(p->cpus_ptr, node_cpumask);
```
likewhatevs, line 2705: depending on how affinities of tasks are set, you might want to make the criteria for allow_node_aligned use this also (possibly adding some configurable threshold on num_cpus for when lo fb's are used to avoid stalls).

etsal:
The change to save_gpu_tgid_pid() assumes that reading the clock is essentially free. The BPF kfunc call is also not guaranteed to be monotonic because it calls ktime_get_mono_fast_ns(), so we should use scx_bpf_now() instead.

xerothermic:
Dropping the GPU matcher update, as it seems like it's not a universal repro. Landing the safe NumaNode change first.

scheds/rust/scx_layered/src/bpf/main.bpf.c
xerothermic: Are you suggesting that we can also relax how we match allow_node_aligned from the entire numa node to have at least N cpus set in the affinity to reduce the lo fb usage?

xerothermic:
> The change to save_gpu_tgid_pid() assumes that reading the clock is essentially free. The BPF kfunc call is also not guaranteed to be monotonic because it calls ktime_get_mono_fast_ns(), so we should use scx_bpf_now() instead.

Thanks for the context. I hit the bpf violation when trying to call scx_bpf_now() in the kprobe context. Now, we decide to drop this change from the patchset, we can re-visit later.

PR #3101, opened 2025-11-26
scx_utils: transfer cpu list from argument off scx_beerland/scx_cosmos.

devnexen:


htejun:
`cpumask.rs` already has cpulist parsing. Maybe switch over to that instead? The special keyword handling is not generic, so I'm not sure that fits well in scx_utils.

PR #3120, opened 2025-12-02
scx_layered: prevent timestamp 0 for GPU kprobe map

etsal:


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -766,8 +766,13 @@ int save_gpu_tgid_pid(void) {
 				taskc->refresh_layer = true;
 
 			timestamp = taskc->running_at;
+		} else {
+			bpf_prink("[BUG] %s:%d", __func__, __LINE__);
```
hodgesds: if it's a bug why not use `trace` macro? I guess you always want it logged?

etsal: Sorry for the late reply, this is a draft test diff to debug an issue. Will mark it as such to prevent confusion.

PR #3123, opened 2025-12-02
scx_p2dq: Add conservative wakeup preemption for latency-critical tasks

hodgesds:
Implement optional wakeup preemption to reduce latency for high-priority tasks when no idle CPU is available.

Core changes:
- Add running_weight field to cpu_ctx to track current task's priority
- Update p2dq_running_impl() to store task weight when task starts running
- Implement wakeup preemption in p2dq_select_cpu_impl() with strict criteria

Preemption conditions (all must be satisfied):
1. Waking task has very high priority (scx.weight >= 2847, nice <= -15)
2. No idle CPU was found during idle CPU search
3. Task can run on prev_cpu (affinity check)
4. prev_cpu is NOT running an interactive task
5. Waking task has HIGHER priority than task running on prev_cpu

The preemption mechanism doesn't bypass normal queueing - it only provides a CPU target hint (prev_cpu) for cache affinity while letting vtime ordering handle priority scheduling.

Configuration flags:
- --latency-priority: Enable latency priority system (unused currently)
- --wakeup-preemption: Enable wakeup preemption feature

This conservative approach ensures we only interrupt lower-priority work and never disrupt interactive or equally important tasks, minimizing impact on tail latency.

### DIFF
--- a/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_p2dq/src/bpf/main.bpf.c
@@ -151,6 +151,15 @@ const volatile struct {
 	.pelt_enabled = true,
 };
 
+/* Latency priority and preemption configuration */
+const volatile struct {
+	bool latency_priority_enabled;
+	bool wakeup_preemption_enabled;
+} latency_config = {
+	.latency_priority_enabled = false,
+	.wakeup_preemption_enabled = false,
+};
+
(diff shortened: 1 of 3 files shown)


hodgesds:
> Is the queued wakeup case handled as well (task wakeup without calling ops.select_cpu())?

Still need to fully handle that case, will do a separate PR.

PR #3126, opened 2025-12-03
layered: fix GPU matching when member_expire_ms=0

xerothermic:
Recently, we introduced the TTL changes, which allow GPU matcher to "expire" a task to force a rematch. By default, member_expire_ms=0, which means a task should never expire, or fail the expieration check.

We noticed after the TTL change, we lost the matching ability when member_expire_ms=0. After debugging, we see that member_expire_ms=0 was treated as a valid timestamp for expieration check.

This diff makes sure we only compare valid member_expire_ms.


etsal:
Nice catch!

PR #3132, opened 2025-12-04
scx_layered: fix kprobe race

etsal:
We currently attach kprobes along with the rest of the BPF programs in the scx_layered skeleton, but that can lead to the kprobes being attached before the struct_ops of the scheduler itself. This in turns introduces a window of time during which tasks that use the GPU may trigger the krpobes and attempt to retrieve their task context, before said context is created in .init(). This in turn causes spurious lookup failures that falsely imply the kprobe logic is flawed and prevents us from putting in checks to catch any actual latent bugs.

Defer kprobe attach until after the rest of the progs in the skeleton are ready. First, introduce a variant of the kprobe_enable function in scx_utils that allows the kprobe to be loaded but not attached. Then, modify scx_layered to only trigger kprobe attach after the scheduler is up-and-running.


hodgesds (approved):
LGTM, linter might need to be appeased

PR #3145, opened 2025-12-11
mitosis: make independent of cpu cg controller presence

likewhatevs:
this pr makes mitosis work more/less the same with cpu controller disabled.

it addresses the following issues:

1) cgroup_init/move/exit are never called when cpu controller is disabled.
2) bpf/scx cgroup helpers return root cg when cpu controller is disabled.
3) use cpus effective instead of cpus allowed when sourcing cpusets via co-re.
4) if a cell exists for a cpuset already, use it instead of creating a new one.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -35,7 +35,7 @@ const volatile u64	     root_cgid			     = 1;
 const volatile bool	     debug_events_enabled	     = false;
 const volatile bool	     exiting_task_workaround_enabled = true;
 const volatile bool	     split_vtime_updates	     = false;
-const volatile bool	     uses_cpu_controller	     = true;
+const volatile bool	     cpu_controller_disabled	     = false;
 
 /*
  * CPU assignment changes aren't fully in effect until a subsequent tick()
@@ -142,7 +142,7 @@ static inline struct cgroup *task_cgroup(struct task_struct *p)
 {
 	struct cgroup *cgrp;
 
-	if (uses_cpu_controller) {
+	if (!cpu_controller_disabled) {
 		cgrp = __COMPAT_scx_bpf_task_cgroup(p);
 	} else {
 		/*
@@ -183,7 +183,7 @@ struct task_ctx {
 	u32 configuration_seq;
 	/* Is this task allowed on all cores of its cell? */
 	bool all_cell_cpus_allowed;
-	/* Last known cgroup ID for detecting cgroup moves (used when !uses_cpu_controller) */
+	/* Last known cgroup ID for detecting cgroup moves (used when cpu_controller_disabled) */
 	u64 cgid;
 };
 
@@ -555,7 +555,7 @@ static __always_inline int maybe_refresh_cell(struct task_struct *p,
 	 * The cgroup is already initialized by tp_cgroup_mkdir which
 	 * fires before the task can be scheduled in the new cgroup.
 	 */
-	if (!uses_cpu_controller) {
+	if (cpu_controller_disabled) {
 		u64 current_cgid;
 
 		bpf_rcu_read_lock();
@@ -1378,7 +1378,7 @@ static int init_cgrp_ctx_with_ancestors(struct cgroup *cgrp)
 s32 BPF_STRUCT_OPS(mitosis_cgroup_init, struct cgroup *cgrp,
 		   struct scx_cgroup_init_args *args)
 {
-	if (!uses_cpu_controller)
+	if (cpu_controller_disabled)
 		return 0;
 	return init_cgrp_ctx(cgrp);
 }
@@ -1388,7 +1388,7 @@ s32 BPF_STRUCT_OPS(mitosis_cgroup_exit, struct cgroup *cgrp)
 	struct cgrp_ctx *cgc;
 	int		 ret;
 
-	if (!uses_cpu_controller)
+	if (cpu_controller_disabled)
 		return 0;
 
 	record_cgroup_exit(cgrp->kn->id);
@@ -1421,7 +1421,7 @@ void BPF_STRUCT_OPS(mitosis_cgroup_move, struct task_struct *p,
 {
 	struct task_ctx *tctx;
 
-	if (!uses_cpu_controller)
+	if (cpu_controller_disabled)
 		return;
 
 	if (!(tctx = lookup_task_ctx(p)))
@@ -1438,7 +1438,7 @@ SEC("tp_btf/cgroup_mkdir")
 int BPF_PROG(tp_cgroup_mkdir, struct cgroup *cgrp, const char *cgrp_path)
 {
 	int ret;
-	if (uses_cpu_controller)
+	if (!cpu_controller_disabled)
 		return 0;
 
 	ret = init_cgrp_ctx_with_ancestors(cgrp);
@@ -1455,7 +1455,7 @@ int BPF_PROG(tp_cgroup_rmdir, struct cgroup *cgrp, const char *cgrp_path)
 {
 	struct cgrp_ctx *cgc;
 
-	if (uses_cpu_controller)
+	if (!cpu_controller_disabled)
 		return 0;
 
 	/*
@@ -1543,7 +1543,7 @@ s32 BPF_STRUCT_OPS(mitosis_init_task, struct task_struct *p,
 	 * We also need to ensure the cgroup hierarchy is initialized since
 	 * SCX cgroup callbacks won't fire.
 	 */
-	if (!uses_cpu_controller) {
+	if (cpu_controller_disabled) {
 		struct cgroup *cgrp = task_cgroup(p);
 		if (!cgrp)
 			return -ENOENT;
@@ -1759,7 +1759,7 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(mitosis_init)
 	 * cgroups get initialized in hierarchical order during scheduler attach.
 	 * The tracepoint handles new cgroups created after attach.
 	 */
-	if (!uses_cpu_controller) {
+	if (cpu_controller_disabled) {
 		struct cgroup *iter_root = NULL;
 
 		bpf_rcu_read_lock();
--- a/scheds/rust/scx_mitosis/src/main.rs
+++ b/scheds/rust/scx_mitosis/src/main.rs
@@ -113,10 +113,10 @@ struct Opts {
     #[clap(long, action = clap::ArgAction::SetTrue)]
     split_vtime_updates: bool,
 
-    /// Use SCX cgroup callbacks (requires CPU cgroup controller enabled).
-    /// When not specified, uses tracepoints and cgroup iteration instead.
+    /// Disable SCX cgroup callbacks (for when CPU cgroup controller is disabled).
+    /// Uses tracepoints and cgroup iteration instead.
     #[clap(long, action = clap::ArgAction::SetTrue)]
-    uses_cpu_controller: bool,
+    cpu_controller_disabled: bool,
 
     #[clap(flatten, next_help_heading = "Libbpf Options")]
     pub libbpf: LibbpfOpts,
@@ -217,7 +217,11 @@ impl<'a> Scheduler<'a> {
             .unwrap()
             .exiting_task_workaround_enabled = opts.exiting_task_workaround;
         skel.maps.rodata_data.as_mut().unwrap().split_vtime_updates = opts.split_vtime_updates;
-        skel.maps.rodata_data.as_mut().unwrap().uses_cpu_controller = opts.uses_cpu_controller;
+        skel.maps
+            .rodata_data
+            .as_mut()
+            .unwrap()
+            .cpu_controller_disabled = opts.cpu_controller_disabled;
 
         skel.maps.rodata_data.as_mut().unwrap().nr_possible_cpus = *NR_CPUS_POSSIBLE as u32;
         for cpu in topology.all_cpus.keys() {


likewhatevs:
added a bunch of folks as reviewers because this needs to ahh well be perfect and test results are suspiciously good.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -44,6 +44,14 @@ const volatile bool	     split_vtime_updates	     = false;
 u32 configuration_seq;
 u32 applied_configuration_seq;
 
+/*
+ * Track whether CPU controller is enabled. Set to true when cgroup_init
+ * callback is invoked for a non-root cgroup (which only happens with CPU
+ * controller enabled). We check non-root because mitosis_init always creates
+ * cgrp_ctx for root cgroup.
+ */
+bool cpu_ctrl_enabled;
```
dschatzberg: I think I would make the entirety of mitosis just operate without the need for the cpu controller, but just make it conditional on a flag to make roll out easier.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -438,27 +491,48 @@ static inline int update_task_cell(struct task_struct *p, struct task_ctx *tctx,
...
-				scx_bpf_error(
-					"Unexpected uninitialized rootcg");
-				return -ENOENT;
-			}
-
-			cgc = lookup_cgrp_ctx(rootcg);
+		if (!READ_ONCE(cpu_ctrl_enabled) &&
+		    (cgc = bpf_cgrp_storage_get(
+			     &cgrp_ctxs, cg, 0,
+			     BPF_LOCAL_STORAGE_GET_F_CREATE))) {
```
dschatzberg: I really don't want cgrp_ctx creation happening on this path. I would have expected something similar to what's in layered that basically juts replicates cgroup_init off a tracepoint (https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/main.bpf.c#L812). Is there a reason that can't be made to work?

dschatzberg (changes requested):
It seems like most of your changes are working around the fact that cgroup_init won't be called for cgroups w/ cpu controller enabled but I think we can just do the same thing as layered does by using a tracepoint instead (https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/src/bpf/main.bpf.c#L812) - aside from that, the only other change you need to make is in task_cgroup, right?

That should make this a much smaller change - I'd still like the logic gated behind a flag just to make rollout more manageable but it should be possible to make it unconditional once rollout has completed.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: I can give that a try, looks like it should work.

likewhatevs:
> aside from that, the only other change you need to make is in task_cgroup, right?

I doubt this -- I think at the least you'd need the replacement for cg init + the change from cpus_allowed to cpus_effective + the co-re change to enable that + not creating duplicate cells for the same cpuset. 

I don't super follow how this works w/o those atm, but init running at just the right time always is probably carrying a lot of water I think.

dschatzberg:
What's the rationale for using cpus_effective? that's all part of the cpuset controller (different from the cpu controller) - so as long as you have the right cgroup (i.e. not the rootcg), the existing logic should just work. The idea I think is that with a tracepoint you are hooking in at effectively the same point as cgroup_init but just using the default cgroup hierarchy to get the right cgroup

likewhatevs:
> What's the rationale for using cpus_effective?

I was getting stalls and seeing the wrong number of cpus in ways that super did not make sense (i.e. neither container cpus nor total cpus) w/o that. will try this w/o that nor cpuset/cell dedup anyhow though.

likewhatevs:
i think this is the minimum which could work, will give it a spin and see if no stalls/no regressions.

likewhatevs:
> i think this is the minimum which could work, will give it a spin and see if no stalls/no regressions.

seems fine w/ this change

dschatzberg:
@likewhatevs - could we literally have the tracepoint and cgroup_init logic be identical except have only one of the function at a time? I'd like to have a flag just set from userspace that dictates if we use the tracepoint logic or the cgroup callback

likewhatevs:
> could we literally have the tracepoint and cgroup_init logic be identical except have only one of the function at a time? I'd like to have a flag just set from userspace that dictates if we use the tracepoint logic or the cgroup callback

I think they can be close -- iiuc `tp_btf/cgroup_attach_task` isn't a replacement for `scx_cgroup_init` wrt/ tasks already running in cg's at scheduler attach time, will play w/ it.

likewhatevs:
> could we literally have the tracepoint and cgroup_init logic be identical except have only one of the function at a time?

I think this both gets close and is maybe a bigger change than we want for this. LMK if we should go w/ more divergent paths to enable a smaller delta, will start testing this.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -438,19 +442,28 @@ static inline int update_task_cell(struct task_struct *p, struct task_ctx *tctx,
...
+			struct cgroup *rootcg = READ_ONCE(root_cgrp);
+			if (rootcg)
+				cgc = lookup_cgrp_ctx_fallible(rootcg);
+			if (!cgc) {
+				tctx->cell		= 0;
+				tctx->configuration_seq = (u32)-1;
+				return update_task_cpumask(p, tctx);
+			}
+		} else if (exiting_task_workaround_enabled &&
+			   (p->flags & PF_EXITING)) {
```
dschatzberg: Hmm, is the only reason you need all this stuff because we won't get a cgroup_init for already created/attached cgroups on initialization? Maybe it would be cleaner to just iterate through all cgroups in mitosis_init and just initialize them explicitly?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -965,10 +978,23 @@ static int update_timer_cb(void *map, int *key, struct bpf_timer *timer)
...
+					     &cgrp_ctxs, cur_cgrp, 0,
+					     BPF_LOCAL_STORAGE_GET_F_CREATE))) {
+					/* Will be properly initialized below */
+					cgrp_ctx->cell	     = 0;
+					cgrp_ctx->cell_owner = false;
+				}
+			}
+			if (!cgrp_ctx)
+				continue;
+		}
```
dschatzberg: It's the same here, right? If we proactively initialize cgroups then we should be good

dschatzberg (commented):
This is looking much closer to what I had in mind - check out the comments inline though, I think we can move some logic into init to mimic what's happening on initialization anyways

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: > Hmm, is the only reason you need all this stuff because we won't get a cgroup_init for already created/attached cgroups on initialization?

Yes once i figure out the minimal that works.

> Maybe it would be cleaner to just iterate through all cgroups in mitosis_init and just initialize them explicitly?

That would work, will figure it out, thx.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: I think rollout/testing is simpler (well, after initial confirmation) w/o cpu controller so I just outright removed the need for this in latest.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1519,6 +1570,33 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(mitosis_init)
...
+	 */
+	{
+		struct cgroup *iter_root = NULL;
+
+		bpf_rcu_read_lock();
+		if (root_cgrp)
+			iter_root = bpf_cgroup_acquire(root_cgrp);
+		bpf_rcu_read_unlock();
+
+		if (iter_root) {
```
dschatzberg: Let's error out on !iter_root rather than silently not initialize

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -804,17 +798,25 @@ typedef struct cpumask___local *cpumask_var_t___ptr;
...
- * Given a cgroup, get its cpumask (populated in entry), returns 0 if no
- * cpumask, < 0 on error and > 0 on a populated cpumask.
+ * Given a cgroup, get its effective cpumask (populated in entry).
+ * Returns 0 if no cpumask or cpumask contains all CPUs, < 0 on error,
+ * and > 0 on a populated restricted cpumask.
+ *
+ * We use effective_cpus rather than cpus_allowed because effective_cpus
+ * reflects the actual usable CPUs after considering ancestor restrictions.
+ * This allows cells to be naturally deduplicated - child cgroups that inherit
+ * their parent's cpuset will have the same effective_cpus and can share a cell.
```
dschatzberg: Why do we need this to be changed? The intent of all this logic was only to create a new cell if cpus_allowed is explicitly configured from userspace

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1009,7 +1011,50 @@ static int update_timer_cb(void *map, int *key, struct bpf_timer *timer)
 			goto out_root_cgrp;
 
 		/*
-		 * cgroup has a cpumask, allocate a new cell if needed, and assign cpus
+		 * cgroup has a cpumask. Check if parent's cell already has the same
+		 * cpumask - if so, reuse it to avoid duplicate cells for inherited
+		 * cpusets.
```
dschatzberg: This bit seems orthogonal to whether or not you need the cpu cg controller enabled/disabled no?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1215,74 +1259,83 @@ int BPF_PROG(fentry_cpuset_write_resmask, struct kernfs_open_file *of,
 	return 0;
 }
 
-s32 BPF_STRUCT_OPS(mitosis_cgroup_init, struct cgroup *cgrp,
-		   struct scx_cgroup_init_args *args)
+/*
+ * Cgroup initialization - creates cgrp_ctx with cell=0 and bumps
+ * configuration_seq. The timer (update_timer_cb) handles the actual cell
+ * assignment based on cpuset configuration.
+ */
```
dschatzberg, line 1285: Why change this initialization? The pre-existing code would:

1) bump the configuration_seq if cpus_allowed was configured
2) Initialize to parent's cell

Why can't we do all that still?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1215,74 +1259,83 @@ int BPF_PROG(fentry_cpuset_write_resmask, struct kernfs_open_file *of,
...
+SEC("tp_btf/cgroup_attach_task")
+int BPF_PROG(tp_cgroup_attach_task, struct cgroup *cgrp, const char *cgrp_path,
+	     struct task_struct *task, bool threadgroup)
+{
+	struct task_ctx *tctx;
+
+	init_cgrp_ctx_minimal(cgrp);
+
+	/* Force task cell refresh on next scheduling event */
```
dschatzberg: We shouldn't need to do this - if the init_cgrp_ctx needs to bump the configuration_seq, then that will just trigger task cell refresh

dschatzberg (changes requested):
This is getting there, just looking for ways we can minimize the changes even more. I really liked that originally this was all configurable via a flag (set from userspace) that controlled the logic - can we add that back in?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: dug a bit more, this (using cpus_allowed like it's used in mitosis) rly should work, will figure it out.

* was just bad test setup, works perfectly w/ cpus_allowed and cell inheritance w/o an explicit dedup pass.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: This is closer but still different:

https://github.com/sched-ext/scx/pull/3145/changes#diff-c9fe76136c7770967146e44f3d711b06044d7ca4f63beb1e5085b8ae1470fd98R1285-R1299

on cpu controller disabled path, in cg init, fallback to cell 0 if parent cg is not initialized. this needs to be different because cg init in tp path is on task attach (not actual cg init, like in scx path) and cg's can be empty/that can be not-called-yet, i think.

dschatzberg: Good call. How about this though - on the tp path - we iterate through all cgroup ancestors and init_cgrp_ctx in order (allowing for already initialized cgroups)? That more closely replicates what happens with current cgroup_init semantics too.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1305,6 +1372,68 @@ void BPF_STRUCT_OPS(mitosis_cgroup_move, struct task_struct *p,
...
+			cgrp->kn->id, ret);
+		return 0;
+	}
+
+	/*
+	 * Bump configuration_seq to trigger task cell refresh via maybe_refresh_cell()
+	 * on the next scheduling event. We can't call update_task_cell() directly
+	 * because tracepoints have read-only access to task arguments.
+	 */
+	__atomic_add_fetch(&configuration_seq, 1, __ATOMIC_RELEASE);
```
dschatzberg: Why do we need to trigger task cell refresh here? init_cgrp_ctx now bumps if it needs to create a new cell, otherwise there shouldn't be any need to trigger the timer.

dschatzberg (commented):
This is looking really good - just a couple comments inline

likewhatevs:
> Good call. How about this though - on the tp path - we iterate through all cgroup ancestors and init_cgrp_ctx in order (allowing for already initialized cgroups)? That more closely replicates what happens with current cgroup_init semantics too.

Do we want to still initialize already existing cgs after adding this? We wouldn't need to w/ this. I think this is asking do we care more about the semantic delta vs code delta.

dschatzberg:
> Do we want to still initialize already existing cgs after adding this?

No, if they're already initialized we should skip it.

>  I think this is asking do we care more about the semantic delta vs code delta.

If I'm interpreting the distinction you're making here correctly - I really want to minimize the semantic delta. There was quite a lot of trickiness in the cgroup initialization, lookup, and destruction paths with a few kernel bugs found along the way. The less we mutate the semantic differences of the mitosis logic, the more likely we are to avoid introducing or stumbling upon new bugs here.

So in this case, I'd love if we could always ensure initialization happened once and only once in ancestor order but I don't see a good way to do that but to add some code to check if initialization already happened and skip

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: This is in leiu of update_task_cell in cgroup_move. I think this fixed affinity violations going high when churning through containers but will confirm.

~~wrong reference...~~ nope this was the right reference and what I tested, confused myself (didn't see the extended comment there after testing/updating and thought i mixed up callsites).

likewhatevs:
Cleaned things up a bit and added lazy creation of parent cg's in tp path, while retaining (now technically unneeded, i think) cg creation during init to reduce semantic delta of the tp/non-tp paths.

Also pulled is_cg_dying into a fn and used it.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -436,14 +450,19 @@ static inline int update_task_cell(struct task_struct *p, struct task_ctx *tctx,
...
-		 * for tasks that were forked before the scheduler was loaded,
-		 * whose cgroups went offline before scx_cgroup_init() ran.
-		 * Only fall back to root cgroup if the workaround is enabled
-		 * and the task is exiting.
+		 * Cgroup lookup failed - this can happen:
+		 * - During scheduler load for tasks forked before the scheduler
+		 *   was loaded, whose cgroups went offline before init ran.
+		 * - When CPU controller is disabled and a cgroup is destroyed
+		 *   while tasks are still referencing it (race between
+		 *   tp_cgroup_rmdir and task scheduling).
```
dschatzberg: Can you unpack this case for me a bit? It shouldn't be possible to destroy a cgroup that still has (non EXITING) processes

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1264,6 +1301,52 @@ s32 BPF_STRUCT_OPS(mitosis_cgroup_init, struct cgroup *cgrp,
...
+	/* Initialize ancestors first (replicates SCX cgroup_init order) */
+	bpf_for(level, 1, target_level)
+	{
+		struct cgroup *ancestor = lookup_cgrp_ancestor(cgrp, level);
+		if (!ancestor)
+			return -ENOENT;
+
+		/* Skip if already initialized or dying */
+		if (!lookup_cgrp_ctx_fallible(ancestor) &&
+		    !cgrp_is_dying(ancestor)) {
```
dschatzberg: I'd flip these two conditionals since if its dying you'd be able to avoid the lookup

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -35,6 +35,7 @@ const volatile u64	     root_cgid			     = 1;
 const volatile bool	     debug_events_enabled	     = false;
 const volatile bool	     exiting_task_workaround_enabled = true;
 const volatile bool	     split_vtime_updates	     = false;
+const volatile bool	     cpu_controller_disabled	     = false;
```
dschatzberg: Can we make this flag "uses_cpu_controller" or something and actually also disable the cgroup_init paths when its flipped off? Basically I'd like to either have cgroup_init and friends used, or the tracepoint. Once we feel confident that the tracepoint approach works we can remove the flag and also the cgroup_init callbacks

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1305,6 +1388,70 @@ void BPF_STRUCT_OPS(mitosis_cgroup_move, struct task_struct *p,
...
+	}
+
+	/*
+	 * Bump configuration_seq to trigger task cell refresh via maybe_refresh_cell()
+	 * on the next scheduling event. We can't call update_task_cell() directly
+	 * because tracepoints have read-only access to task arguments.
+	 * Without this bump, affn violations on cell 0 will increase as tasks
+	 * are not updated as needed after move.
+	 */
+	__atomic_add_fetch(&configuration_seq, 1, __ATOMIC_RELEASE);
```
dschatzberg: I'm a little nervous about bumping this on every cgroup_attach_task - it basically means we're going to be triggering the timer very frequently. I wonder if we're better off just recording (per task) the cgroup we've updated to, and check if it changes (in select_cpu, enqueue, etc.) and run update_task_cell.

I think it might even not be correct to do this asynchronously  (the vtime won't be updated until after we schedule possibly)

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: Let me test this without this constraint, I think I might have added that while debugging test setup/wrong binary running (w/o knowing that was the case).

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: uses_cpu_controller == current behavior with this flag enabled, just a thing we need to track is all, added + gated.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: added this check to maybe refresh cell, calling update if needed:

https://github.com/sched-ext/scx/pull/3145/changes#diff-c9fe76136c7770967146e44f3d711b06044d7ca4f63beb1e5085b8ae1470fd98R543-R561

added some logging to allocate_cell using this, suspect we want to use this in most error logging to help w/ figuring out what is going on:

https://github.com/sched-ext/scx/pull/3145/changes#diff-c9fe76136c7770967146e44f3d711b06044d7ca4f63beb1e5085b8ae1470fd98R478-R479

likewhatevs:
Still works w/ the latest changes (leveraging cfg seq less). 

In theory init w/ ancestors could be dropped w/ the latest changes (most specifically, moving init tracking from cg task attach to cg mkdir), but i think init w/ ancestors does a lot of work wrt/ keeping this approach from being a race condition finder (i.e. the first which comes to mind is tp's firing before mitosis_init completes if init w/ ancestors were not a thing, and then polling locks (to prevent that) from trace points etc.)).

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -516,7 +539,30 @@ static __always_inline int maybe_refresh_cell(struct task_struct *p,
...
+		bpf_rcu_read_unlock();
+
+		if (current_cgid != tctx->cgid) {
+			if (!(cgrp = task_cgroup(p)))
+				return -1;
+			if (update_task_cell(p, tctx, cgrp))
+				ret = -1;
+			bpf_cgroup_release(cgrp);
+		}
+	}
```
dschatzberg, line 567: I think you can probably restructure this a bit just to avoid duplicating the few lines here that call update_task_cell() - e.g. something like:

```
needs_update = false;
if (tctx->configuration_seq != READ_ONCE(applied_configuration_seq)) {
  needs_update = true;
} else if (!uses_cpu_controller) {
  ...
}
```

Maybe there's something even cleaner - not a big deal either way just take a look at it

likewhatevs: good call, added a commit cleaning that up, will test cg ctrl on and cg ctrl off then merge.

likewhatevs:
replaced uses_cpu_controller (flag presence disables new capability) w/ cpu_controller_disabled (flag presence enables new capability) to keep behavior w/ the same flags the same across the addition of new arg.

PR #3146, opened 2025-12-12
scx_mitosis: Add Optional LLC Awareness and Work Stealing

tommy-u:
## Summary

Mitosis previously wasn't topology aware, dispatching tasks to a single DSQ per cell. This work adds
LLC (Last Level Cache) domain awareness to scx_mitosis. When enabled via `--enable-llc-awareness`,
tasks are assigned to LLC domains within their cell and preferentially scheduled on CPUs sharing that LLC.
Cross-LLC work stealing allows for some degree of load balancing and work conservation.

## Examples

```bash
# Original behavior, no LLC awareness
scx_mitosis

# LLC-aware scheduling, no work stealing
scx_mitosis --enable-llc-awareness

# LLC-aware with work stealing (recommended)
scx_mitosis --enable-llc-awareness --enable-work-stealing
```

### BPF Side

- **DSQ typed encoding**: Introduced `struct dsq_type` and encoding helpers to create unique DSQ IDs that embed both cell index and LLC index, enabling per-(cell, LLC) dispatch queues.
- **LLC-aware task assignment**: Tasks are assigned to an LLC within their cell using weighted random selection based on CPU count per LLC. The task's cpumask is narrowed to CPUs in its assigned LLC domain.
- **Per-LLC vtime tracking**: Each cell now maintains `llc_vtime_now[MAX_LLCS]` for fair scheduling within LLC domains, replacing the single cell-wide vtime.
- **Work stealing**: Idle CPUs can steal tasks from sibling LLC DSQs within the same cell. Stolen tasks are detected via LLC mismatch in `running()` and retagged to the thief's LLC. 
- **Steal statistics**: Steal events are tracked as `CSTAT_STEAL` in per-cell cstats.
- **Header restructuring**: LLC-aware logic is isolated in `llc_aware.bpf.h` with `mitosis.bpf.h` providing shared types and utilities.

### Rust Side

- **Topology initialization**: Populates `cpu_to_llc` and `llc_to_cpus` maps from system topology before BPF init.
- **CLI options**: Added `--enable-llc-awareness`, `--enable-work-stealing` to control the feature.
- **Cell LLC recalculation**: Calls `recalc_cell_llc_counts()` after cell cpumask changes to update per-LLC CPU counts.

## Unchanged Behavior

When `--enable-llc-awareness` is not set:
- Tasks dispatch to flat per-cell DSQs as before
- No LLC domain assignment or cross-LLC stealing occurs
- `FAKE_FLAT_CELL_LLC` (0) is used for DSQ encoding

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/intf.h
+++ b/scheds/rust/scx_mitosis/src/bpf/intf.h
@@ -71,6 +71,7 @@ enum cell_stat_idx {
 	CSTAT_CPU_DSQ,
 	CSTAT_CELL_DSQ,
 	CSTAT_AFFN_VIOL,
+	CSTAT_STEAL,
 	NR_CSTATS,
 };
 
@@ -86,16 +87,6 @@ struct cgrp_ctx {
 	u32  cell;
 	bool cell_owner;
 };
-
-/*
-* Cell struct shared between kernel and userspace.
-* Kernel uses spinlock for atomic updates.
-* Userspace must read with BPF_F_LOCK to avoid torn reads.
-* Lock field is padding (kernel zeros it to avoid leaking pointers).
-*
-* map.lookup(&key, MapFlags::ANY)  -> userspace may see torn state
-* map.lookup(&key, MapFlags::LOCK) -> safe read
-*/
 
 /*
  * Per-LLC data is cacheline-aligned to prevent false sharing when
@@ -120,8 +111,15 @@ _Static_assert(sizeof(struct cell_llc) >= CACHELINE_SIZE,
 		u32 __pad; \
 	}
 #endif
+
 /*
- * cell is the per-cell book-keeping
+* Cell struct shared between kernel and userspace.
+* Kernel uses spinlock for atomic updates.
+* Userspace must read with BPF_F_LOCK to avoid torn reads.
+* Lock field is padding (kernel zeros it to avoid leaking pointers).
+*
+* map.lookup(&key, MapFlags::ANY)  -> userspace may see torn state
+* map.lookup(&key, MapFlags::LOCK) -> safe read
 */
 struct cell {
 	// This is a lock in the kernel and padding in userspace
@@ -142,7 +140,6 @@ struct cell {
 
 // Putting the lock first in the struct is our convention.
 // We pad this space when in Rust code that will never see the lock value.
-// We intentionally avoid it in copy_cell_skip_lock to keep the verifier happy.
 // It is a BPF constraint that it is 4 byte aligned.
 
 // All assertions work for both BPF and userspace builds
--- a/scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
+++ b/scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
@@ -6,7 +6,8 @@
  * This header assists adding LLC cache awareness to scx_mitosis by defining
  * maps and fns for managing CPU-to-LLC domain mappings. It provides code to
  * recalculate per-LLC CPU counts within cells and implements weighted
- * random LLC selection for tasks.
+ * random LLC selection for tasks. It also tracks work-stealing
+ * statistics for cross-LLC task migrations.
  */
 #pragma once
 
@@ -35,6 +36,12 @@ static inline bool	  llc_is_valid(u32 llc_id)
 static inline void init_task_llc(struct task_ctx *tctx)
 {
 	tctx->llc = LLC_INVALID;
+
+	if (!enable_work_stealing)
+		return;
+
+	tctx->steal_count    = 0;
+	tctx->last_stolen_at = 0;
 }
 
 static inline const struct cpumask *lookup_llc_cpumask(u32 llc)
@@ -142,9 +149,10 @@ static inline s32 pick_llc_for_task(u32 cell_id)
 	bpf_spin_lock(&cell->lock);
 	for (u32 i = 0; i < MAX_LLCS; i++)
 		llc_cpu_cnt[i] = cell->llcs[i].cpu_cnt;
-	bpf_spin_unlock(&cell->lock);
 
 	u32 total_cpu_cnt = cell->cpu_cnt;
+	bpf_spin_unlock(&cell->lock);
+
 	if (!total_cpu_cnt) {
 		scx_bpf_error(
 			"pick_llc_for_task: cell %d has no CPUs accounted yet",
@@ -191,6 +199,105 @@ static void zero_cell_vtimes(struct cell *cell)
 	}
 }
 
+/*
+ * Detect and handle cross-LLC task migration.
+ * Called from running() to check if task's assigned LLC differs
+ * from the CPU's LLC (indicating work stealing occurred).
+ *
+ * Caller must ensure enable_llc_awareness is true.
+ */
+static inline int maybe_retag_stolen_task(struct task_struct *p,
+					  struct task_ctx    *tctx,
+					  struct cpu_ctx     *cctx)
+{
+	/* No mismatch = no steal, fast path */
+	if (tctx->llc == cctx->llc)
+		return 0;
+
+	/* Task was stolen to a different LLC - update accounting */
+	tctx->steal_count++;
+	tctx->last_stolen_at = scx_bpf_now();
+
+	/* Assign task to new LLC */
+	tctx->llc = cctx->llc;
+
+	/*
+	 * New LLC, need new cpumask. This updates the task vtime
+	 * to that of the new cell-LLC DSQ.
+	 */
+	return update_task_cpumask(p, tctx);
+}
+
+/* Work stealing:
+ * Scan sibling (cell,LLC) DSQs in the same cell and steal the first queued task if it can run on this cpu
+ * Returns:
+ *  true == 1;  task was stolen
+ *  false == 0; no tasks were stolen
+ *  error <0;   error encountered
+*/
+static inline s32 try_stealing_work(u32 cell, s32 local_llc)
+{
+	if (!llc_is_valid(local_llc)) {
+		scx_bpf_error("try_stealing_work: invalid local_llc: %d",
+			      local_llc);
+		return -EINVAL;
+	}
+
+	struct cell *cell_ptr = lookup_cell(cell);
+	if (!cell_ptr)
+		return -EINVAL;
+
+	// Loop over all other LLCs, looking for a queued task to steal
+	u32 i;
+	bpf_for(i, 1, nr_llc)
+	{
+		// Start with the next one to spread out the load
+		u32 candidate_llc = (local_llc + i) % nr_llc;
+
+		// Prevents the optimizer from removing the following conditional return
+		// so that the verifier knows the read will be safe
+		barrier_var(candidate_llc);
+
+		if (candidate_llc >= MAX_LLCS)
+			continue;
+
+		/*
+    * Skip if the cell doesn't have CPUs in this LLC.
+    * Note: rechecking cell_ptr for verifier.
+    * This is racy with try_stealing_this_task, but we don't care -
+    * if the LLC actually doesn't have CPUs come steal time,
+    * we will fail the steal and continue to the next LLC.
+    */
+		if (cell_ptr &&
+		    READ_ONCE(cell_ptr->llcs[candidate_llc].cpu_cnt) == 0)
+			continue;
+
+		dsq_id_t candidate_dsq =
+			get_cell_llc_dsq_id(cell, candidate_llc);
+		if (dsq_is_invalid(candidate_dsq))
+			return -EINVAL; // already errored in get_cell_llc_dsq_id
+
+		// Optimization: skip if faster than constructing an iterator
+		// Not redundant with later checking if task found (race)
+		if (!scx_bpf_dsq_nr_queued(candidate_dsq.raw))
+			continue;
+
+		/*
+		 * Attempt the steal - can fail because it's a race.
+		 * We don't update task_ctx here because the peeked task_ctx
+		 * may be stale (a different task may now be at head of DSQ).
+		 * Actual retag and accounting happens in running() via
+		 * mismatch detection.
+		 */
+		if (!scx_bpf_dsq_move_to_local(candidate_dsq.raw))
+			continue;
+
+		// Success, we got a task
+		return true;
+	}
+	return false;
+}
+
 static inline int update_task_llc_assignment(struct task_struct *p,
 					     struct task_ctx	*tctx)
 {
@@ -227,7 +334,9 @@ static inline int update_task_llc_assignment(struct task_struct *p,
 	}
 
 	/* --- Point to the correct (cell,LLC) DSQ and set vtime baseline --- */
-	tctx->dsq	  = get_cell_llc_dsq_id(tctx->cell, tctx->llc);
+	tctx->dsq = get_cell_llc_dsq_id(tctx->cell, tctx->llc);
+	if (dsq_is_invalid(tctx->dsq))
+		return -EINVAL;
 
 	struct cell *cell = lookup_cell(tctx->cell);
 	if (!cell)
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -423,7 +423,7 @@ static inline int update_task_cpumask(struct task_struct *p,
 
 		tctx->dsq = get_cpu_dsq_id(cpu);
 		if (dsq_is_invalid(tctx->dsq))
-			return -EINVAL; // scx_bpf_error() already called in get_cpu_dsq_id
+			return -EINVAL;
 
 		p->scx.dsq_vtime = READ_ONCE(cpu_ctx->vtime_now);
 		return 0;
@@ -438,7 +438,7 @@ static inline int update_task_cpumask(struct task_struct *p,
 	/* Non-LLC aware version */
 	tctx->dsq = get_cell_llc_dsq_id(tctx->cell, FAKE_FLAT_CELL_LLC);
 	if (dsq_is_invalid(tctx->dsq))
-		return -EINVAL; // scx_bpf_error() already called in get_cell_llc_dsq_id
+		return -EINVAL;
 
 	struct cell *cell;
 	if (!(cell = lookup_cell(tctx->cell)))
@@ -773,7 +773,7 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
 	dsq_id_t cpu_dsq  = get_cpu_dsq_id(cpu);
 
 	if (dsq_is_invalid(cell_dsq) || dsq_is_invalid(cpu_dsq)) {
-		return; // scx_bpf_error already called in get_*_dsq_id
+		return;
 	}
 
 	/* Peek at cell-LLC DSQ head */
@@ -793,12 +793,23 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
 	}
 
 	/*
-	 * If we failed to find an eligible task, scx will keep running prev if
-	 * prev->scx.flags & SCX_TASK_QUEUED (we don't set SCX_OPS_ENQ_LAST), and
-	 * otherwise go idle.
+	 * If we failed to find an eligible task, try work stealing if enabled.
+	 * Otherwise, scx will keep running prev if prev->scx.flags &
+	 * SCX_TASK_QUEUED (we don't set SCX_OPS_ENQ_LAST), and otherwise go idle.
 	 */
-	if (!found)
+	if (!found) {
+		/* Try work stealing if enabled */
+		if (enable_llc_awareness && enable_work_stealing) {
+			/* Returns: <0 error, 0 no steal, >0 stole work */
+			s32 ret = try_stealing_work(cell, llc);
+			if (ret < 0)
+				return;
+			if (ret > 0) {
+				cstat_inc(CSTAT_STEAL, cell, cctx);
+			}
+		}
 		return;
+	}
 
 	/*
 	 * The move_to_local can fail if we raced with some other cpu in the cell
@@ -1269,16 +1280,21 @@ void BPF_STRUCT_OPS(mitosis_running, struct task_struct *p)
 	struct task_ctx *tctx;
 	struct cell	*cell;
 
-	if (!(tctx = lookup_task_ctx(p)))
+	if (!(cctx = lookup_cpu_ctx(-1)) || !(tctx = lookup_task_ctx(p)))
 		return;
+
+	/* Handle stolen task retag (LLC-aware mode only) */
+	if (enable_llc_awareness && enable_work_stealing) {
+		if (maybe_retag_stolen_task(p, tctx, cctx) < 0)
+			return;
+	}
 
 	/*
 	 * Legacy approach: Update vtime_now before task runs.
 	 * Only used when split vtime updates is enabled.
 	 */
 	if (split_vtime_updates) {
-		if (!(cctx = lookup_cpu_ctx(-1)) ||
-		    !(cell = lookup_cell(cctx->cell)))
+		if (!(cell = lookup_cell(cctx->cell)))
 			return;
 
 		advance_dsq_vtimes(cell, cctx, tctx, p->scx.dsq_vtime);
@@ -1593,6 +1609,13 @@ s32 validate_flags()
 		return -EINVAL;
 	}
 
+	/* Work stealing only makes sense when enable_llc_awareness. */
+	if (enable_work_stealing && (!enable_llc_awareness)) {
+		scx_bpf_error(
+			"Work stealing requires LLC-aware mode to be enabled");
+		return -EINVAL;
+	}
+
 	return 0;
 }
 
@@ -1732,7 +1755,7 @@ void BPF_STRUCT_OPS(mitosis_dump, struct scx_dump_ctx *dctx)
 		/* Per-LLC stats deferred: FAKE_FLAT_CELL_LLC used for now */
 		dsq_id_t dsq_id = get_cell_llc_dsq_id(i, FAKE_FLAT_CELL_LLC);
 		if (dsq_is_invalid(dsq_id))
-			return; // scx_bpf_error already called by get_cell_llc_dsq_id
+			return;
 
 		scx_bpf_dump(
 			"CELL[%d] vtime=%llu nr_queued=%d\n", i,
@@ -1747,7 +1770,7 @@ void BPF_STRUCT_OPS(mitosis_dump, struct scx_dump_ctx *dctx)
 
 		dsq_id = get_cpu_dsq_id(i);
 		if (dsq_is_invalid(dsq_id))
-			return; // scx_bpf_error already called by get_cpu_dsq_id
+			return;
 		scx_bpf_dump("CPU[%d] cell=%d vtime=%llu nr_queued=%d\n", i,
 			     cpu_ctx->cell, READ_ONCE(cpu_ctx->vtime_now),
 			     scx_bpf_dsq_nr_queued(dsq_id.raw));
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
@@ -38,6 +38,7 @@ enum mitosis_constants {
  * Variables populated by userspace
  */
 const volatile bool	   enable_llc_awareness = false;
+const volatile bool	   enable_work_stealing = false;
 const volatile u32	   nr_llc		= 1;
 
 static inline struct cell *lookup_cell(int idx)
@@ -77,6 +78,9 @@ struct task_ctx {
 	u64 cgid;
 	/* Which LLC this task is assigned to */
 	s32 llc;
+
+	u32 steal_count; /* how many times this task has been stolen */
+	u64 last_stolen_at; /* ns timestamp of the last steal (scx_bpf_now) */
 };
 
 static inline const struct cpumask *lookup_cell_cpumask(int idx);
--- a/scheds/rust/scx_mitosis/src/main.rs
+++ b/scheds/rust/scx_mitosis/src/main.rs
@@ -129,6 +129,10 @@ struct Opts {
     #[clap(long, action = clap::ArgAction::SetTrue)]
     enable_llc_awareness: bool,
 
+    /// Enable work stealing. This is only relevant when LLC-awareness is enabled.
+    #[clap(long, action = clap::ArgAction::SetTrue)]
+    enable_work_stealing: bool,
+
     #[clap(flatten, next_help_heading = "Libbpf Options")]
     pub libbpf: LibbpfOpts,
 }
@@ -168,6 +172,7 @@ struct DistributionStats {
     cpu_q_pct: f64,
     cell_q_pct: f64,
     affn_viol_pct: f64,
+    steal_pct: f64,
 
     // for formatting
     global_queue_decisions: u64,
@@ -189,20 +194,31 @@ impl Display for DistributionStats {
         };
         write!(
             f,
-            "{:width$} {:5.1}% | Local:{:4.1}% From: CPU:{:4.1}% Cell:{:4.1}% | V:{:4.1}%",
+            "{:width$} {:5.1}% | Local:{:4.1}% From: CPU:{:4.1}% Cell:{:4.1}% | V:{:4.1}% S:{:4.1}%",
             self.total_decisions,
             self.share_of_decisions_pct,
             self.local_q_pct,
             self.cpu_q_pct,
             self.cell_q_pct,
             self.affn_viol_pct,
+            self.steal_pct,
             width = descisions_width,
         )
     }
 }
 
 impl<'a> Scheduler<'a> {
+    fn validate_args(opts: &Opts) -> Result<()> {
+        if opts.enable_work_stealing && !opts.enable_llc_awareness {
+            bail!("Work stealing requires LLC-aware mode (--enable-llc-awareness)");
+        }
+
+        Ok(())
+    }
+
     fn init(opts: &Opts, open_object: &'a mut MaybeUninit<OpenObject>) -> Result<Self> {
+        Self::validate_args(opts)?;
+
         let topology = Topology::new()?;
 
         let nr_llc = topology.all_llcs.len().max(1);
@@ -240,6 +256,7 @@ impl<'a> Scheduler<'a> {
         // Set nr_llc in rodata
         rodata.nr_llc = nr_llc as u32;
         rodata.enable_llc_awareness = opts.enable_llc_awareness;
+        rodata.enable_work_stealing = opts.enable_work_stealing;
 
         match *compat::SCX_OPS_ALLOW_QUEUED_WAKEUP {
             0 => info!("Kernel does not support queued wakeup optimization."),
@@ -305,6 +322,7 @@ impl<'a> Scheduler<'a> {
         global_queue_decisions: u64,
         scope_queue_decisions: u64,
         scope_affn_viols: u64,
+        scope_steals: u64,
     ) -> Result<DistributionStats> {
         // First % on the line: share of global work
         // We know global_queue_decisions is non-zero.
@@ -329,6 +347,12 @@ impl<'a> Scheduler<'a> {
             100.0 * (scope_affn_viols as f64) / (scope_queue_decisions as f64)
         };
 
+        let steal_pct = if scope_queue_decisions == 0 {
+            0.0
+        } else {
+            100.0 * (scope_steals as f64) / (scope_queue_decisions as f64)
+        };
+
         const EXPECTED_QUEUES: usize = 3;
         if queue_pct.len() != EXPECTED_QUEUES {
             bail!(
@@ -345,6 +369,7 @@ impl<'a> Scheduler<'a> {
             cpu_q_pct: queue_pct[1],
             cell_q_pct: queue_pct[2],
             affn_viol_pct: affinity_violations_percent,
+            steal_pct,
             global_queue_decisions,
         });
     }
@@ -371,12 +396,19 @@ impl<'a> Scheduler<'a> {
             .map(|&cell| cell[bpf_intf::cell_stat_idx_CSTAT_AFFN_VIOL as usize])
             .sum::<u64>();
 
+        // Sum steals over all cells
+        let scope_steals: u64 = cell_stats_delta
+            .iter()
+            .map(|&cell| cell[bpf_intf::cell_stat_idx_CSTAT_STEAL as usize])
+            .sum::<u64>();
+
         // Special case where the number of scope decisions == number global decisions
         let stats = self.calculate_distribution_stats(
             &queue_counts,
             global_queue_decisions,
             global_queue_decisions,
             scope_affn_viols,
+            scope_steals,
         )?;
 
         self.metrics.update(&stats);
@@ -417,11 +449,16 @@ impl<'a> Scheduler<'a> {
             let scope_affn_viols: u64 =
                 cell_stats_delta[cell][bpf_intf::cell_stat_idx_CSTAT_AFFN_VIOL as usize];
 
+            // Steals for this cell
+            let scope_steals: u64 =
+                cell_stats_delta[cell][bpf_intf::cell_stat_idx_CSTAT_STEAL as usize];
+
             let stats = self.calculate_distribution_stats(
                 &queue_counts,
                 global_queue_decisions,
                 cell_queue_decisions,
                 scope_affn_viols,
+                scope_steals,
             )?;
 
             self.metrics
--- a/scheds/rust/scx_mitosis/src/stats.rs
+++ b/scheds/rust/scx_mitosis/src/stats.rs
@@ -30,6 +30,8 @@ pub struct CellMetrics {
     pub cell_q_pct: f64,
     #[stat(desc = "Affinity violations % of global")]
     pub affn_violations_pct: f64,
+    #[stat(desc = "Steal %")]
+    pub steal_pct: f64,
     #[stat(desc = "Decision share % of global")]
     pub share_of_decisions_pct: f64,
     #[stat(desc = "Cell scheduling decisions")]
@@ -42,6 +44,7 @@ impl CellMetrics {
         self.cpu_q_pct = ds.cpu_q_pct;
         self.cell_q_pct = ds.cell_q_pct;
         self.affn_violations_pct = ds.affn_viol_pct;
+        self.steal_pct = ds.steal_pct;
         self.share_of_decisions_pct = ds.share_of_decisions_pct;
         self.total_decisions = ds.total_decisions;
     }
@@ -61,6 +64,8 @@ pub struct Metrics {
     pub cell_q_pct: f64,
     #[stat(desc = "Affinity violations % of global")]
     pub affn_violations_pct: f64,
+    #[stat(desc = "Steal %")]
+    pub steal_pct: f64,
     #[stat(desc = "Decision share % of global")]
     pub share_of_decisions_pct: f64,
     #[stat(desc = "Cell scheduling decisions")]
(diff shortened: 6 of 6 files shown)


scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -0,0 +1,374 @@
...
+
+#include "mitosis.bpf.h"
+#include "intf.h"
+
+typedef u32 llc_id_t;
+#define LLC_INVALID ((llc_id_t)~0u)
+
+// A CPU -> LLC cache ID map
+struct cpu_to_llc_map {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
```
likewhatevs: should this be percpu? i vaguely recall that maybe mattering w/ layered. not so much this one in particular (although it looks like it could be), but iiuc more/less that which can be percpu array is best if it is.

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
```diff
@@ -0,0 +1,200 @@
...
+ * The use of these bitfields depends on compiler defined byte AND bit ordering.
+ * Make sure we're only building with Clang/LLVM and that we're little-endian.
+ */
+#ifndef __clang__
+#error "This code must be compiled with Clang/LLVM (eBPF: clang -target bpf)."
+#endif
+
+#if __BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__
+#error "dsq64 bitfield layout assumes little-endian (bpfel)."
+#endif
```
dschatzberg, line 69: It's probably worth making this work for big endian machines relatively soon

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
```diff
@@ -0,0 +1,200 @@
...
+		u64 rsvd : RSVD_B;
+	} user_dsq;
+
+	/* Built-in DSQ view */
+	struct {
+		u64 value : 32;
+		u64 rsvd : 30;
+		u64 local_on : 1;
+		u64 builtin : 1;
+	} builtin_dsq;
```
dschatzberg, line 116: In the past I've seen gcc/llvm produce really unoptimized code with these bitfields. Stylistically this differs quite a bit from how the Linux kernel typically uses bitfields (instead relying on macros and shifts/masks). I think I'd prefer aligning with the approach used in the kernel but I don't feel too strongly - just take a closer look here.

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
```diff
@@ -0,0 +1,200 @@
...
+static inline bool is_cpu_dsq(dsq_id_t dsq_id)
+{
+	return is_user_dsq(dsq_id) && dsq_id.user_dsq.type == DSQ_TYPE_CPU;
+}
+
+// If this is a per cpu dsq, return the cpu
+static inline u32 get_cpu_from_dsq(dsq_id_t dsq_id)
+{
+	if (!is_cpu_dsq(dsq_id))
+		scx_bpf_error("trying to get cpu from non-cpu dsq\n");
```
dschatzberg: We should print the dsq getting passed in here and we also need some return value to indicate that callers here should error out

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
```diff
@@ -0,0 +1,200 @@
...
+
+	return dsq_id.cpu_dsq.cpu;
+}
+
+/* Helper functions to construct DSQ IDs */
+static inline dsq_id_t get_cpu_dsq_id(u32 cpu)
+{
+	// Check for valid CPU range, 0 indexed so >=.
+	if (cpu >= MAX_CPUS)
+		scx_bpf_error("invalid cpu %u\n", cpu);
```
dschatzberg: Similarly here, we need a way to error out of a caller - scx_bpf_error doesn't just halt the program it gets checked after the entire bpf call returns

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
```diff
@@ -0,0 +1,200 @@
...
+
+	return (dsq_id_t){ .cpu_dsq = { .cpu = cpu, .type = DSQ_TYPE_CPU } };
+}
+
+static inline dsq_id_t get_cell_l3_dsq_id(u32 cell, u32 l3)
+{
+	if (cell >= MAX_CELLS || l3 >= MAX_L3S)
+		scx_bpf_error("cell %u or l3 %u too large\n", cell, l3);
```
dschatzberg: And here again

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1487,9 +1478,9 @@ void BPF_STRUCT_OPS(mitosis_dump_task, struct scx_dump_ctx *dctx,
 		return;
 
 	scx_bpf_dump(
-		"Task[%d] vtime=%llu basis_vtime=%llu cell=%u dsq=%x all_cell_cpus_allowed=%d\n",
+		"Task[%d] vtime=%llu basis_vtime=%llu cell=%u dsq=%llu all_cell_cpus_allowed=%d\n",
```
dschatzberg: If you're printing the raw value I think I prefer hex - it's easier to read, no?

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -25,6 +26,8 @@ enum consts {
 	MAX_CG_DEPTH	      = 256,
 	MAX_L3S		      = 16,
 
+	MITOSIS_ENABLE_STEALING = 1,
```
dschatzberg: Let's avoid having compile-time configuration. I really think we want to enable features at runtime so we can ship a single binary and just dynamically choose the feature set we enable

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -74,14 +77,66 @@ struct cgrp_ctx {
...
 
+#if defined(__BPF__)  // ← ADD THIS LINE
+#define CELL_LOCK_T struct bpf_spin_lock
+#else
+/* userspace placeholder: kernel won’t copy spin_lock */
+#define CELL_LOCK_T        \
+        struct {           \
+                u32 __pad; \
+        } /* 4-byte aligned as required */
+#endif
```
dschatzberg, line 113: This bit of code probably requires some inline comment explaining why its needed - very confusing to me in isolation

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -74,14 +77,66 @@ struct cgrp_ctx {
...
+        u32 l3_present_cnt;
+
+        // Number of CPUs from each L3 assigned to this cell
+        u32 l3_cpu_cnt[MAX_L3S];
+
+        // per-L3 vtimes within this cell
+        u64 l3_vtime_now[MAX_L3S];
+
+		// XXX Remove me when shift to llc is complete.
+		u64 vtime_now;
```
dschatzberg: The formatting here looks a bit odd

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -74,14 +77,66 @@ struct cgrp_ctx {
...
+        u32 cpu_cnt;
+
+        // Number of L3s with at least one CPU in this cell
+        u32 l3_present_cnt;
+
+        // Number of CPUs from each L3 assigned to this cell
+        u32 l3_cpu_cnt[MAX_L3S];
+
+        // per-L3 vtimes within this cell
+        u64 l3_vtime_now[MAX_L3S];
```
dschatzberg: I wonder if you would rather have a new struct cell_llc that has the number of CPUs and the vtimes in it - you may even want to cacheline align such structs to reduce cache ping-ponging

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1125,17 +1158,14 @@ void BPF_STRUCT_OPS(mitosis_running, struct task_struct *p)
 	/*
 	 * Legacy approach: Update vtime_now before task runs.
 	 * Only used when split vtime updates is enabled.
+	 * Expect "vtime too far ahead" errors if this is enabled.
```
dschatzberg: Can we just remove the flag? Seems like we're confident the code works so no reason to make it conditional

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
dschatzberg: e.g. take a look at https://github.com/torvalds/linux/blob/master/include/linux/bitfield.h for how the linux kernel helpers look here - might be worth replicating that

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: Hmm, for cacheing purposes? I think this should be fine because after init it's read only & thus trivially cacheable. Maybe you have other concerns?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: Totally agree it should be eliminated. I assumed this would need to be done in a separate PR.

tommy-u:
> I didn't quite get through half of this. It's a lot to review all at once. Can you split out the work stealing related bits and have this PR only cover LLC locality? I'll keep working through this but in general the quality of my reviews is inversely proportional to the amount of code in the PR.

I simplified this PR into 3 commits. I cleanly separated stealing out into its own commit, but kept it in the same PR because they do go together. Stealing is intended to better load balance tasks which would otherwise be statically assigned to LLCs.

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
tommy-u: I thought bitfields would be a clear way to express these structures. I didn't consider that using bitfields like this this could have negative performance implications. 

Shifting to the style in bitfield.h would bring us into alignment with the kernel, potentially improve performance, and presumably fix our endianness restriction. Let's decide if we're going this route, and if it should happen in this PR or a new one.

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: I have removed this.

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: Removed

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: Ha, yes, agreed.

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
tommy-u: fixed

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
tommy-u: fixed

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
tommy-u: fixed

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: done

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -259,7 +225,7 @@ static inline int allocate_cell()
 			return -1;
 
 		if (__sync_bool_compare_and_swap(&c->in_use, 0, 1)) {
-			WRITE_ONCE(c->vtime_now, 0);
+			WRITE_ONCE(c->llc_vtime_now[FAKE_FLAT_CELL_LLC], 0);
```
tommy-u: This is inconsistent in the LLC case. I think we don't need to zero vtime here, it's just going to get pulled forward to the cpu dsq vtime in stopping.

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: Done. I like the encapsulation into cell_llc. I'm not certain this path gets enough write traffic to justify the padding, but I don't see any harm.

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
dschatzberg: This doesn't actually look fixed

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
dschatzberg: I'll leave it up to you to make the decision here

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -65,23 +65,95 @@ enum cell_stat_idx {
...
+	       "in_use must follow 4-byte lock/padding");
+
+// Verify these are the same size in both BPF and Rust.
+// 16 bytes header + 48 bytes padding (to align llcs) + 64*16 bytes llcs = 1088 bytes
+_Static_assert(sizeof(struct cell) ==
+		       (CACHELINE_SIZE + (CACHELINE_SIZE * MAX_LLCS)),
+	       "struct cell size must be stable for Rust bindings");
+
+_Static_assert(sizeof(struct cell) == 1088,
+	       "struct cell must be exactly 1088 bytes");
```
dschatzberg: Is this assert all that useful? If we extend struct cell we are forced to keep changing this?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -0,0 +1,262 @@
...
+#pragma once
+
+#include "mitosis.bpf.h"
+#include "intf.h"
+
+typedef u32 llc_id_t;
+#define LLC_INVALID ((llc_id_t)~0u)
+
+// A CPU -> LLC cache ID map
+struct cpu_to_llc_map {
```
dschatzberg: Is there a reason to map this a BPF map instead of just a global array that userspace populates?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -0,0 +1,262 @@
...
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, u32);
+	__type(value, struct cpumask);
+	__uint(max_entries, MAX_LLCS);
+};
+
+extern struct cpu_to_llc_map  cpu_to_llc;
+extern struct llc_to_cpus_map llc_to_cpus;
+
+static inline bool	      llc_is_valid(u32 llc_id)
```
dschatzberg: whitespace here looks a little odd

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -0,0 +1,262 @@
...
+
+	return mask;
+}
+
+/* Recompute cell->llc_cpu_cnt[] after cell cpumask changes */
+static __always_inline void recalc_cell_llc_counts(u32 cell_idx)
+{
+	struct cell *cell = lookup_cell(cell_idx);
+	if (!cell) {
+		scx_bpf_error("recalc_cell_llc_counts: invalid cell %d",
```
dschatzberg: lookup_cell already has an scx_bpf_error (that's the point having making these wrapper lookup functions) - is there any value added by outputting a different error here?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -0,0 +1,262 @@
...
+			      cell_idx);
+		return;
+	}
+
+	CPUMASK_GUARD(tmp_guard);
```
dschatzberg, line 72: Just FYI I have a commit pending to create guard macros in the Linux kernel style (https://github.com/dschatzberg/scx/commit/4aee5b1286b9aea28c2b9e4f5720f8e359f85575) - I'll wait until after you land this PR and then do a sweep to change these to the kernel style guard macros

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
dschatzberg: I also see a few more extra scx_bpf_error calls after lookups - juts scan through and check if they're adding any value

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -0,0 +1,262 @@
...
+	/*
+	 * Read only what we need under the lock to avoid putting the
+	 * large cell struct on the stack (would exceed BPF stack limit).
+	 */
+	u32 llc_cpu_cnt[MAX_LLCS];
+
+	bpf_spin_lock(&cell->lock);
+	for (u32 i = 0; i < MAX_LLCS; i++)
+		llc_cpu_cnt[i] = cell->llcs[i].cpu_cnt;
+	bpf_spin_unlock(&cell->lock);
```
dschatzberg: I wonder if locking is really the best concurrency construct for cell manipulation. If I followed the logic correctly you are only ever mutating the cell struct under update_timer_cb but everywhere else is read-only? I guess this is only really executing when reconfiguration happens so locking might be justified? Some description of the concurrency model is probably warranted somewhere in here

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -0,0 +1,262 @@
...
+	if (tctx->llc != LLC_INVALID) {
+		llc_mask = lookup_llc_cpumask((u32)tctx->llc);
+		/* If the LLC no longer intersects the cell's cpumask, invalidate it */
+		if (!llc_mask ||
+		    !bpf_cpumask_intersects(cell_cpumask, llc_mask))
+			tctx->llc = LLC_INVALID;
+	}
+
+	/* --- Pick a new LLC if needed --- */
+	if (tctx->llc == LLC_INVALID) {
```
dschatzberg: I'm a bit surprised by this check - under what conditions do we need to update a task's llc assignment? It should only be if the task's cell was updated, no? In that case wouldn't we want to unconditionally pick a new LLC for the task given that the previous assignment had no bearing on how cpus were assigned to its current cell?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -755,22 +753,30 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
 	struct task_struct *p;
-	bpf_for_each(scx_dsq, p,
-		     get_cell_llc_dsq_id(cell, FAKE_FLAT_CELL_LLC).raw, 0) {
+
+	/* Check the cell-LLC DSQ (use FAKE_FLAT_CELL_LLC when not LLC-aware) */
+	u32	 llc	= enable_llc_awareness ? cctx->llc : FAKE_FLAT_CELL_LLC;
+
+	dsq_id_t dsq_id = get_cell_llc_dsq_id(cell, llc);
+	if (dsq_is_invalid(dsq_id))
+		return;
```
dschatzberg, line 776: Does this need an scx_bpf_error?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -755,22 +753,30 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+		min_vtime_dsq = dsq_id;
 		found	      = true;
 		break;
 	}
 
-	dsq_id_t dsq = get_cpu_dsq_id(cpu);
-	bpf_for_each(scx_dsq, p, dsq.raw, 0) {
+	dsq_id = get_cpu_dsq_id(cpu);
+	if (dsq_is_invalid(dsq_id))
+		return;
```
dschatzberg: and here as well

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1767,6 +1831,19 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(mitosis_init)
...
 			return -EINVAL;
 		}
+
+		/* Store the LLC that each cpu belongs to. Used in Dispatch. */
+		struct cpu_ctx *cpu_ctx = lookup_cpu_ctx(i);
+		if (cpu_ctx) {
+			if (enable_llc_awareness) {
+				u32 *llc_ptr;
+				llc_ptr = bpf_map_lookup_elem(&cpu_to_llc, &i);
+				// TODO: Should this error out here if NULL?
```
dschatzberg: Yeah for sure

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: There's no such thing as a global array in BPF, right? That's just syntactic sugar for another map.

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
tommy-u: I'm punting this to future work.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: Don't get me started... the clang-format spec forces this spacing. If you want to be truly horrified, look at what it does to the variable definitions in dispatch()

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: Ah man, that's right. I also need to return errors up the chain

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -0,0 +1,262 @@
...
+	u32 llc_cpu_cnt_tmp[MAX_LLCS] = { 0 };
+
+	{ // RCU context
+		RCU_READ_GUARD();
+		const struct cpumask *cell_mask =
+			lookup_cell_cpumask(cell_idx); // RCU ptr
+
+		if (!cell_mask) {
+			scx_bpf_error(
+				"recalc_cell_llc_counts: invalid cell mask");
```
tommy-u: This is not checked elsewhere in the codebase, perhaps assumed good by construction

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: I'm not the best at this, but here's my reasoning:

Agreed, one mutator, one reader. Maybe userspace could write this directly in the future.

[Like here](https://github.com/sched-ext/scx/pull/3146/changes/9569e4e4d6aabb51bcef2661d9110b7beb16095e#diff-4912ad70de152bdff68aa3fd3cc09f86ccd91c304820d84f07938d41eda5abb4R81), I'm worried about e.g. torn reads. Getting half way through this loop, then the other cpu overwrites the array, then we finish the read loop.

Do you have a certain construct in mind? A seqlock seems like overkill, but might fit well. We use rcu elsewhere, I could figure that out.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: I agree, let's reassign the LLC unconditionally. While update_task_llc_assignment is downstream of mitosis_set_cpumask, I don't think that path would make it here because update_task_cpumask already guaranteed that all_cell_cpus_allowed == true.

Whatever tiny optimization happens when the task happens to be in the same LLC isn't even worth the complexity of this condition. Does no favors for load balancing either.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: Perhaps these could hang off the cpu contexts?

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
tommy-u: Noted. When we decide this, I'll start the work.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: yes, added

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: You will also see I cleaned up the dispatch code on this commit, it used to come later.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: We discussed this. It seems good enough for now

scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
tommy-u: Ok, it (and all others I could find) are now actually fixed

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: Removed

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: I changed this to do the "global array" syntax and changed userspace to update it correctly

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: Woah, surprisingly no. It's actually already called in get_cell_llc_dsq_id

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: No actually, but commenting for clarity

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: commenting for clarity

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: done

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: I fixed these

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -35,6 +36,13 @@ static inline bool	  llc_is_valid(u32 llc_id)
 static inline void init_task_llc(struct task_ctx *tctx)
 {
 	tctx->llc = LLC_INVALID;
+
+	if (!enable_work_stealing)
+		return;
+	tctx->pending_llc      = LLC_INVALID;
+	tctx->steal_count      = 0;
+	tctx->last_stolen_at   = 0;
+	tctx->steals_prevented = 0;
```
tommy-u: Perhaps these could be collected into a struct

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +191,136 @@ static void zero_cell_vtimes(struct cell *cell)
...
+   * Caller must ensure enable_llc_awareness is true.
+   */
+static inline void apply_pending_llc_retag(struct task_struct *p,
+					   struct task_ctx    *tctx)
+{
+	/* This is the normal do nothing path */
+	if (tctx->pending_llc == LLC_INVALID)
+		return;
+
+	if (!llc_is_valid(tctx->pending_llc)) {
```
tommy-u: Slight inconsistency between:
- llc_is_valid
- dsq_is_invalid

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +191,136 @@ static void zero_cell_vtimes(struct cell *cell)
...
+
+	/* Assign task to new LLC */
+	tctx->llc	  = tctx->pending_llc;
+	tctx->pending_llc = LLC_INVALID;
+
+	/*
+	 * New LLC, need new cpumask. This updates the task vtime
+	 * to that of the new cell-LLC dsq.
+	*/
+	update_task_cpumask(p, tctx);
```
tommy-u: Dan: In your LLC review you mentioned "under what conditions do we need to update a task's llc assignment? It should only be if the task's cell was updated, no?"

I had forgotten we invoke that path through update_task_cpumask when stealing.

I don't think this changes anything about our decision to make it unconditionally pick a new LLC.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +191,136 @@ static void zero_cell_vtimes(struct cell *cell)
...
+static inline s32 try_stealing_work(u32 cell, s32 local_llc)
+{
+	if (!llc_is_valid(local_llc)) {
+		scx_bpf_error("try_stealing_work: invalid local_llc");
+		return -EINVAL;
+	}
+
+	struct cell *cell_ptr = lookup_cell(cell);
+	if (!cell_ptr) {
+		scx_bpf_error("try_stealing_work: invalid cell");
```
tommy-u: remove extra error

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +191,136 @@ static void zero_cell_vtimes(struct cell *cell)
...
+    * This is racy with try_stealing_this_task, but we don't care -
+    * if the LLC actually doesn't have CPUs come steal time,
+    * we will fail the steal and continue to the next LLC.
+    */
+		if (cell_ptr &&
+		    READ_ONCE(cell_ptr->llcs[candidate_llc].cpu_cnt) == 0)
+			continue;
+
+		u64 candidate_dsq =
+			get_cell_llc_dsq_id(cell, candidate_llc).raw;
```
tommy-u: add return on error case

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1251,6 +1262,11 @@ void BPF_STRUCT_OPS(mitosis_running, struct task_struct *p)
 	if (!(tctx = lookup_task_ctx(p)))
 		return;
 
+	/* Handle stolen task retag (LLC-aware mode only) */
+	if (enable_llc_awareness && enable_work_stealing) {
+		apply_pending_llc_retag(p, tctx);
```
tommy-u: return if this errors

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +191,136 @@ static void zero_cell_vtimes(struct cell *cell)
 	}
 }
 
+/*
+   * Apply pending LLC retag if task was stolen across LLC domains.
+   * Called from running() when a task starts executing.
+   *
+   * Caller must ensure enable_llc_awareness is true.
+   */
+static inline void apply_pending_llc_retag(struct task_struct *p,
```
tommy-u: make this return int so we can communicate error case

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +192,139 @@ static void zero_cell_vtimes(struct cell *cell)
 	}
```
tommy-u, line 199: maybe the error check should be in the critical section because it's racy here.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +192,139 @@ static void zero_cell_vtimes(struct cell *cell)
...
+ * Scan sibling (cell,LLC) DSQs in the same cell and steal the first queued task if it can run on this cpu
+ * Returns:
+ *  true == 1;  task was stolen
+ *  false == 0; no tasks were stolen
+ *  error <0;   error encountered
+*/
+static inline s32 try_stealing_work(u32 cell, s32 local_llc)
+{
+	if (!llc_is_valid(local_llc)) {
+		scx_bpf_error("try_stealing_work: invalid local_llc");
```
dschatzberg: Output the value so its easier to debug

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +192,139 @@ static void zero_cell_vtimes(struct cell *cell)
...
+		struct task_ctx	   *task_ctx;
+		// I'm only using this for the verifier
+		bool found_task = false;
+
+		// Optimization: skip if faster than constructing an iterator
+		// Not redundant with later checking if task found (race)
+		if (!scx_bpf_dsq_nr_queued(candidate_dsq.raw))
+			continue;
+
+		// Just a trick for peeking the head element
```
dschatzberg: I think we now have some peek call we can use, I believe lavd is using it already?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +192,139 @@ static void zero_cell_vtimes(struct cell *cell)
...
+			found_task = true;
+			break;
+		}
+
+		// No task? Try next LLC
+		if (!found_task)
+			continue;
+
+		// This knob throttles stealing.
+		if (task_ctx->steals_prevented++ < steal_throttle) {
```
dschatzberg: Walk me through this a bit - why throttle on a per-task basis?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -183,6 +192,139 @@ static void zero_cell_vtimes(struct cell *cell)
...
+	// Attempt the steal, can fail because it's a race.
+	if (!scx_bpf_dsq_move_to_local(candidate_dsq))
+		return false;
+
+	// We got the task!
+	task_ctx->steal_count++;
+	task_ctx->last_stolen_at = scx_bpf_now();
+	/* Retag to thief LLC (the one for this cpu) */
+	task_ctx->pending_llc	   = local_llc;
+	task_ctx->steals_prevented = 0;
```
dschatzberg: This all looks buggy to me - how do you know the thing you've stolen is this task_ctx?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: @JakeHillion suggested this as a naive way to control stealing rates, it was easy to implement.

The main idea is just to have some knob to turn that allows interpolation stealing rate from whatever rate the stealing policy achieves on its own, down to zero.

The per-task basis approach is biased to reduce the chance of cores stealing the same task back and forth which may hurt cache perf.

This could lead to worse work conservation when a cpu sleeps instead of stealing work.

It could lead to longer LLC searches, but I don't think this timescale matters.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: Absolutely right. My plan is to drop any stats updates here and check in running if the task llc differs from the cpu llc

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -753,25 +753,33 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+	bool		    found	  = false;
+	dsq_id_t	    min_vtime_dsq = DSQ_INVALID;
+	u64		    min_vtime	  = 0;
 
 	struct task_struct *p;
-	bpf_for_each(scx_dsq, p, cell, 0) {
+
+	dsq_id_t dsq_id = get_cell_llc_dsq_id(cell, FAKE_FLAT_CELL_LLC);
+	if (dsq_is_invalid(dsq_id))
+		return; // scx_bpf_error already called in get_cell_llc_dsq_id
```
dschatzberg: These comments feel a bit superfluous to me. Certainly on first read I'd want to make sure these helpers are calling scx_bpf_error but at every call site it feels unnecessary.

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -65,23 +77,90 @@ enum cell_stat_idx {
...
+*/
+
+/*
+ * Per-LLC data is cacheline-aligned to prevent false sharing when
+ * CPUs on different LLCs update their vtime concurrently.
+ */
+struct cell_llc {
+	u64 vtime_now;
+	u32 cpu_cnt;
+} __attribute__((aligned(CACHELINE_SIZE)));
```
dschatzberg, line 98: In the past I've seen bindgen not respect this attribute. Can you just confirm that it does the right thing?

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -65,23 +77,90 @@ enum cell_stat_idx {
...
 };
 
 struct cgrp_ctx {
 	u32  cell;
 	bool cell_owner;
 };
 
+/*
+* Cell struct shared between kernel and userspace.
+* Kernel uses spinlock for atomic updates.
```
dschatzberg: There's no spinlock now so this comment is erroneous

scheds/rust/scx_mitosis/src/bpf/intf_rust.h
```diff
@@ -0,0 +1,16 @@
+/*
+ * This is a preprocessor workaround so we can differentiate
+ * bindgen using intf.h for rust and the BPF program using it for C.
+ *
+ * __KERNEL__ doesn't work because it's not defined when our BPF program is
+ * compiled.
+ *
+ * __BPF__ doesn't work because it IS defined when our rust program is
```
dschatzberg, line 8: It's probably worth digging into why this is - it probably shouldn't behave that way

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: There's still a spinlock in struct cell

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: My belief is that the property we care about (not having multiple on a cache line) is actually checked by the static assert that follows it (in both the C and Rust environments).

As an example, the following code barfs when building the Rust code.

If you flip `#ifdef` to `#ifndef`, you get a barf when building the C code. 

```
struct test_struct {
	u32 a;
#ifdef __BINDGEN_RUNNING__
	u32 b;
#endif
};
_Static_assert(sizeof(struct test_struct) == 4, "test_struct must be 4 bytes");
```

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: Absolutely right, sorry

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: Happy to spot check in a debugger or the binary if there's a remaining concern

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: There's still a spinlock in struct cell.

dschatzberg: Ah, I was thrown off by this comment being above struct cell_llc. It's a bit oddly placed?

scheds/rust/scx_mitosis/src/bpf/intf.h
dschatzberg: My concern is that the output of bindgen (the rust struct) won't be sized the same way. You can just spot check the generated code to confirm.

scheds/rust/scx_mitosis/src/bpf/intf_rust.h
tommy-u: So it appears that it's intentionally designed this way. For yucks, I tried removing the flag, and I hit challenges with other flags that depend on it. BPF-specific flag -mcpu=v3 is only valid under that target.

I imagine it's possible to remove this, but I'm guessing it would be good to chat with Tejun first. 

https://github.com/sched-ext/scx/blob/3f893fdec09f0b4167e7b69319538186f3bdf3a3/rust/scx_cargo/src/bpf_builder.rs#L312-L320

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: I see, thanks. I ran bindgen by hand to confirm this, and the Rust code it produces does respect the alignment (and thus sizing) constraint. Learned a lot. **Looks like this is doing what we want.**

```
#[repr(C)]
#[repr(align(64))]
#[derive(Debug, Copy, Clone)]
pub struct cell_llc {
    pub vtime_now: u64_,
    pub cpu_cnt: u32_,
}
```

This static assert from the generated Rust code is why we know the structs are the same in both Rust and C. It is totally separate from the static assert that I put in the intf.h file. My understanding is that the right-hand side of this subtraction is the independent check from bindgen calling into libclang. 
```
#[allow(clippy::unnecessary_operation, clippy::identity_op)]
const _: () = {
    ["Size of cell_llc"][::std::mem::size_of::<cell_llc>() - 64usize];
```

scheds/rust/scx_mitosis/src/bpf/intf_rust.h
tommy-u: Followup: replacing 
`.chain(["-target".into(), "bpf".into()].iter()),` 
with 
`.filter(|f| *f != "-mcpu=v3"),`

Still builds the whole project without any issues. And `__BPF__` is not defined when bindgen is running. Perhaps I could float it to the team to get some clarity on whether we actually want these flags (and others) passed into bindgen.

scheds/rust/scx_mitosis/src/bpf/intf.h
tommy-u: Yes, this comment is in an insane location 🙈

scheds/rust/scx_mitosis/src/bpf/intf_rust.h
likewhatevs: this doesn't add up to me but at the same time i'd say clean it up later to get this merged tbh, the reason it's needed sounds like something is wrong with something.

scheds/rust/scx_mitosis/src/bpf/intf.h
dschatzberg: Nice, it seems bindgen has fixed this issue. I wasn't aware it also translated the static asserts - that's cool too.

tommy-u: It actually doesn't translate the static asserts. Those are separately generated to make sure that the Rust and Clang structs are the same size.

PR #3151, opened 2025-12-12
Add perfmon based scheduling to scx_cosmos

balbirs-nv:
Add support for optimizing a single perf counter passed in via -e and a threshold via -E. This gives the scheduler to distribute or keep tasks sticky based on the counter and it's threshold. To help with further optimization add NUMA awareness, so that perfmon based optimization can be contained in the same NUMA node if tasks are being distributed, find a suitable candidate within the same node as the current CPU (current CPU w.r.t task scheduling, which is being optimized for the perf counter)

The patches enable numa by default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


balbirs-nv:
> Makes sense!
> 
> The perf event tracking code is very close to what we have in lib/pmu.bpf.c. Would you be open to a PR that switches cosmos to it? If the module is missing anything, can we expand it to make it work?

Thanks for the feedback, Happy to do it as a follow up once this is merged!

PR #3160, opened 2025-12-18
lib: assorted bugfixes

etsal:
Fix an assorted collection of errors for the scheduler libraries.


hodgesds (approved):
LGTM

lib/topology.bpf.c
```diff
@@ -209,12 +209,12 @@ topo_ptr topo_find_sibling(topo_ptr topo, u32 cpu)
...
 
 	if (!parent) {
-		bpf_printk("parent has no sibling");
+		bpf_printk("node is at the top level");
 		return NULL;
 	}
 
-	for (i = 0; i < topo->nr_children && can_loop; i++) {
-		child = topo->children[i];
+	for (i = 0; i < parent->nr_children && can_loop; i++) {
```
hodgesds, line 216: easy coverage

lib/lvqueue.bpf.c
```diff
@@ -86,7 +86,7 @@ int lvq_push(lv_queue_t __arg_arena *lvq, u64 val)
 		lvq->cur = newarr;
 	}
 
-	lv_arr_put(arr, b, val);
+	lv_arr_put(newarr, b, val);
```
arighi: This should be `lv_arr_put(lv->cur, b, val)` because newarr is only initialized inside the if block.

arighi (changes requested):

lib/lvqueue.bpf.c
etsal: Ack, I'm surprised I didn't get a compiler warning for this.

PR #3162, opened 2025-12-18
scx_bpfland: Re-introduce --slice-min-us

arighi:
Some workloads may benefit from setting a minimum allowed time slice, so re-introdce the option --slice-min-us (default = no minimum time slice).


hodgesds (approved):
LGTM

PR #3169, opened 2025-12-21
Downgrade crate dependencies to more stable / portable versions

arighi:
Aggressively updating crates to the latest versions inevitably breaks builds on more stable distributions that do not aggressively update their toolchains (such as Ubuntu).

Revert the latest dependency upgrades, as they break the build for all schedulers on Ubuntu, due to some crates relying on unstable Rust features, like:
```
error[E0658]: use of unstable library feature `unsigned_is_multiple_of`
   --> /home/arighi/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ruzstd-0.8.2/src/huff0/huff0_decoder.rs:248:51
    |
248 |                 let bytes_needed = if num_weights.is_multiple_of(2) {
    |                                                   ^^^^^^^^^^^^^^
    |
    = note: see issue #128101 <https://github.com/rust-lang/rust/issues/128101> for more information
```

As a general guideline for the future I think we should avoid being overly aggressive at updating to the latest crate versions and prioritize portability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.25.1"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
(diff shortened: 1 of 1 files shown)


sirlucjan:
I checked Fedora, openSUSE, and Arch. We should probably stick with the current crates, as it is somewhat difficult to ensure stable functionality for all distributions.

Issue #3182, opened 2025-12-27
scx_lavd: No debugfs mount found


name-tar-xz:
log:
```
thread 'main' (139436) panicked at rust/scx_utils/src/energy_model.rs:237:44:
called `Result::unwrap()` on an `Err` value: No debugfs mount found

Stack backtrace:
   0: scx_utils::compat::debugfs_mount
   1: scx_utils::energy_model::get_pd_paths
   2: scx_lavd::clap_main_fn
   3: scx_lavd::main
   4: std::sys::backtrace::__rust_begin_short_backtrace
   5: std::rt::lang_start::{{closure}}
   6: std::rt::lang_start_internal
   7: main
   8: __libc_start_call_main
   9: __libc_start_main_alias_1
  10: _start
stack backtrace:
   0:     0x55b1c44794ab - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h5a7965721e20362f
   1:     0x55b1c4201c2f - core::fmt::write::hb2be3bc45c2132e7
   2:     0x55b1c448ff0d - std::io::Write::write_fmt::hefe7258ac2f8f77a
   3:     0x55b1c4477973 - std::sys::backtrace::BacktraceLock::print::hf8c144f475d7afb9
   4:     0x55b1c448bab2 - std::panicking::default_hook::{{closure}}::h20495c8c5cc77186
   5:     0x55b1c448b992 - std::panicking::default_hook::h26df0833967641ff
   6:     0x55b1c448bc61 - std::panicking::panic_with_hook::h88667df88367c84a
   7:     0x55b1c447806a - std::panicking::panic_handler::{{closure}}::hc49779adfcdeb73f
   8:     0x55b1c4477a89 - std::sys::backtrace::__rust_end_short_backtrace::ha021440587b8be26
   9:     0x55b1c4487add - __rustc[bd3b3f48d2381c32]::rust_begin_unwind
  10:     0x55b1c412cfd0 - core::panicking::panic_fmt::h4594a04b5e6b3550
  11:     0x55b1c412c876 - core::result::unwrap_failed::h8522f703098ca08b
  12:     0x55b1c4441d85 - scx_utils::energy_model::get_pd_paths::h6b2a9d34f39a78d4
  13:     0x55b1c433c0f3 - scx_lavd::clap_main_fn::h7a14d3f025684f54
  14:     0x55b1c434017d - scx_lavd::main::h1f7a2e40ff45a53b
[...]

PR #3184, opened 2025-12-30
utils: Graceful error handling for EnergyModel.

multics69:
When debugfs is not mounted (or disabled), creating EnergyModel caused a panic with the following error:

  thread 'main' (465) panicked at rust/scx_utils/src/energy_model.rs:237:44:
  called `Result::unwrap()` on an `Err` value: No debugfs mount found

Instead of unwrap(), let's use ? to propagate an error, so handling it gracefully without panicking.

This should fix https://github.com/sched-ext/scx/issues/3182

### DIFF
--- a/rust/scx_utils/src/energy_model.rs
+++ b/rust/scx_utils/src/energy_model.rs
@@ -65,7 +65,7 @@ impl EnergyModel {
         };
 
         for (pd_id, pd_path) in pd_paths {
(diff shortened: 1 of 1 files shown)


etsal (approved):

multics69:
Thanks @etsal and @oxyzenQ  for the review!

hodgesds (approved):

PR #3186, opened 2025-12-31
scx_beerland: Reduce scheduler overhead

arighi:
Prefer using SCX_DSQ_LOCAL[_ON] whenever possible to reduce scheduling overhead, and use __COMPAT_scx_bpf_dsq_peek() to check if any task is waiting on a per-CPU DSQ, preventing starvation.

Additionally, in opt.dispatch(), always try to consume from a remote DSQ if the current DSQ is empty. While this may slightly affect locality, it improves responsiveness and work conservation when the system is underutilized, making it beneficial overall.

With this change in place, we can slightly improve schbench performance (in the 1% range, but consistently), allowing to outperform EEVDF/BORE, without regressing any other benchmarks (at least those that I tested).


hodgesds (approved):
LGTM

PR #3188, opened 2026-01-01
arena: initial ASAN support and new allocators

etsal:
Add an arena ASAN BPF library module for tracking spatial (pointer arithmetic and temporal (use-after-free) memory errors. Writing arena code is error-prone and complicated because there is no way to determine whether a memory access is valid. Even though recent kernels provide arena page-fault reporting, invalid memory accesses within page-backed memory still succeed and can cause arbitrary memory corruption. Address sanitization ensures that unallocated memory accesses are properly tracked and reported, making debugging easier.

Also refactor, fix, and test the existing byte-granularity memory allocators. Memory allocation in the repository is either slab-based, or completely static. The buddy allocator, while functional, had multiple bugs that were unreasonably labor-intensive to fix without memory error reporting. Using BPF Streams arena stack traces and ASAN, amend and expand the buddy allocator to make it more useful for scheduler writing.

ASAN and the allocators are fully functional but highly experimental and subject to breaking changes. While the feature is unstable, it is worth adding it to the repository so we can have it interact with the codebase. Also, sched-ext is the largest codebase that uses BPF arenas and testing ASAN against it will help us adjust it to be effective and easy to use for general arena code.

The feature requires using a bleeding-edge Clang compiler and bpf-next. This requirement will remain for the rest of the feature's development, as our design will require further kernel patches  for support.

### DIFF
--- a/lib/selftests/alloc/README.md
+++ b/lib/selftests/alloc/README.md
@@ -1,0 +1,227 @@
+ARENA ASAN FRAMEWORK
+====================
+
+This is a short document logging the development of ASAN for BPF arena memory.
+The document will be removed once development is complete and the feature is
+upstreamed to the kernel. 
+
+Initial development for the feature was done in https://github.com/etsal/aasan/tree/fixup.
+The initial code contribution here in the scx/ repo squashes the development history for
+clarity. Later upstreaming will split the code into separate features.
+
+STATUS
+======
+
+Fully functional, still in development. The three allocators are passing their selftests,
+both for the allocators themselves and for ASAN support. Reporting is currently manual
+because it makes testing easier, but can be trivially changed to produce backtraces. This
+is due to BUG()-style reporting being implicitly included as a feature in the patchset.
+
+The allocators currently require ASAN, but adding a no-op ASAN header is to avoid this is
+trivial.
+
+ASAN support requires an LLVM compiler built from HEAD and bpf-next.
+
+DESIGN
+======
+
+We have two axes on which we need to work:
+
+- MECHANISM: The main KASAN mechanism.
+- ALLOCATOR: The consumer of the mechanism. It is co-designed with KASAN because
+  it is the same consumer.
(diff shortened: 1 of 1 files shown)


oxyzenQ (approved):

multics69:
Thanks, @etsal ,  for your efforts in supporting ASAN for the BPF area. I have two questions:

1) I was not able to find places to call __asan_storeN() and __asan_loadN() in the code base. I presume those will be emitted by the compiler when generating load and store instructions accessing the BPF arena. Am I correct? If so, it would be nice to add comments to make it easier to find the missing link.

2) Do you assume that the buddy allocator is always used when ASAN is on? If the buddy allocator can run with ASAN off, would it be nice to check whether the allocator metadata is corrupted? Assuming the ASAN would be off in production due to overhead, this would help spot the problem early, even with limited precision.

etsal:
Hi Changwoo,

> Thanks, @etsal , for your efforts in supporting ASAN for the BPF area. I have two questions:
> 
> 1. I was not able to find places to call __asan_storeN() and __asan_loadN() in the code base. I presume those will be emitted by the compiler when generating load and store instructions accessing the BPF arena. Am I correct? If so, it would be nice to add comments to make it easier to find the missing link.

Yes they are there just in case the compiler generates the calls. I will add a comment.

> 2. Do you assume that the buddy allocator is always used when ASAN is on? If the buddy allocator can run with ASAN off, would it be nice to check whether the allocator metadata is corrupted? Assuming the ASAN would be off in production due to overhead, this would help spot the problem early, even with limited precision.

Do you mean that the buddy allocator always runs with ASAN on? It is on my TODO list to add a `#ifdef` guards that turn 
`asan_poison/asan_unpoison` into no-ops, to effectively turn ASAN off.

multics69:
Thanks for the clarification, Emil. While I didn't take a look at the code super deep, it looks good to me.

PR #3206, opened 2026-01-09
mitosis: dynamic cpus for affinitized tasks

dschatzberg:
Tasks that are affinitized (e.g. to a subset of their cell) are presently simply bound to a single cpu out of their affinity mask. This mostly works since most affinitized tasks are just affinitized to a single cpu. We've observed instances where tasks are affinitized to multiple cpus and the initialization-time binding can lead to significant cpu imbalance.

This commit adds a new flag --dynamic-affinity-cpu-selection which will dynamically re-assign such tasks (only affinitized tasks) to random cpus within their affinity mask in order to reduce load imbalance. This doesn't completely eliminate cpu load imbalance for these kinds of tasks but avoids load-balancing or other complexities and ultimately prevents long-term imbalance due to making random decisions at each scheduling point.

Additionally I've added a test script to reproduce this behavior (may take a few runs due to the random assignment).

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -36,6 +36,7 @@ const volatile bool	     debug_events_enabled	     = false;
 const volatile bool	     exiting_task_workaround_enabled = true;
 const volatile bool	     split_vtime_updates	     = false;
 const volatile bool	     cpu_controller_disabled	     = false;
+const volatile bool	     dynamic_affinity_cpu_selection  = false;
 
 /*
  * CPU assignment changes aren't fully in effect until a subsequent tick()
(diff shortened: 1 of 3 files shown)


oxyzenQ (approved):

tommy-u (approved):
Goal is to improve load balancing for affinitized tasks after observing workloads that pin to multiple cpus.

Changes are in `mitosis_select_cpu` and `mitosis_enqueue`.

Select now spreads work potentially across all cpus the thread is allowed on.

`mitosis_select_cpu` used to throw affinitized task on whatever cpu dsq was chosen in `update_task_cpumask`. Now when `dynamic_affinity_cpu_selection` is enabled, it 
1. tries for an idle cpu in the task's cpuset
2. failing that it grabs a random cpu from the set
3. when moved to a new cpu, updates task vtime to the new cpu dsq's vtime.

Enqueue randomly redistributes the task when the selected cpu is busy and also has a queued task.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -616,7 +617,50 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+		if (dynamic_affinity_cpu_selection) {
+			const struct cpumask *idle_smtmask;
+
+			/*
+			 * Dynamic affinity balancing: use pick_idle_cpu_from
+			 * for SMT-aware idle CPU selection. If no idle CPU,
+			 * randomly pick to balance load over time.
+			 */
+			idle_smtmask = scx_bpf_get_idle_smtmask();
+			if (!idle_smtmask) {
```
likewhatevs, line 630: I think this is meant to be `bpf_cpumask_empty(idle_smtmask)`

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -616,7 +617,50 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu,
...
+				}
+			}
+		} else {
+			cpu = dsq_to_cpu(tctx->dsq);
+		}
+
+affn_done:
+
 		if (scx_bpf_test_and_clear_cpu_idle(cpu))
 			scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, slice_ns, 0);
```
likewhatevs, line 665: should affn_done be here instead? all paths in `pick_idle_cpu_from` clear idle/claim cpu.

likewhatevs (changes requested):
LGTM, couple of minor maybe-perf-relevant things.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: No, pick_idle_cpu_from handles an empty idle mask - this is just error handling in case scx_bpf_get_idle_smtmask were to return NULL. I don't believe it can do that presently but the verifier doesn't know that. I should add an scx_bpf_error here though - I'll do that

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Hmm, good catch - I wonder if we even need this conditional with the new logic. Let me study it a bit more closely

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: oof, these api's lol, right.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -616,7 +617,50 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu,
 
 	if (!tctx->all_cell_cpus_allowed) {
 		cstat_inc(CSTAT_AFFN_VIOL, tctx->cell, cctx);
-		cpu = dsq_to_cpu(tctx->dsq);
+
```
tommy-u, line 620: From a performance standpoint, might it make sense to have a shortcut case for when the task only runs on a single CPU?

likewhatevs:
linking this because the idea here is probably worth considering wrt/ handling of affine tasks: https://github.com/sched-ext/scx/commit/86b36ec982950fbd25c0cfdac0f4c745ead02bf3

in a sentence, it is "treat tasks with affinities greater than 1 as if they have the affinity of the cell". 

the reasoning is "i think there's a rapid roll-off wrt/ stalls wrt/ num eligible cpus wrt/ workloads and stall handling can get involved w/ more complex approaches". 

think definitely carries a lot of water in that last sentence, but yeah idk just a thing worth considering.

htejun:

PR #3217, opened 2026-01-12
scx_lavd: Misc code clean-up and hardening.

multics69:
This patchset contains one hardening and one clean-up change:

* Check the CPU’s local DSQ before removing it from the overflow CPU set to be more conservative.
* Remove an unused function, pick_any_bit().


etsal (approved):
IIRC the change in power.bpf.c also tries to elide the pseudorandom check whenever possible?

multics69:
> IIRC the change in power.bpf.c also tries to elide the pseudorandom check whenever possible?

You are right. I will update the commit message and merge it.

PR #3218, opened 2026-01-12
scx_lavd: Improve latency criticality propagation among tasks.

multics69:
This patch set improves how we propagate latency-criticality across waker-wakee relationships and how we use it to achieve better overall latency. It includes the following three changes:

* Previously, we propagated (inherited) the waker’s latency-criticality to the wakee (i.e., forward propagation). This is good to carry forward the momentum to the wakee. In addition to this, we add the backward propagation — propagating the wakee’s to the waker side. The idea is to move the waker so it doesn't block the road.

* When a waker is an RT or DL task, increase the latency-criticality of its wakee. This is necessary because we do not manage the latency criticality of non-SCX (RT/DL) tasks.

* Previously, we penalized greedy tasks and provided no reward for underserved tasks. This approach actually treats an underserved wakee as a fairly served, regular background task. This is not ideal, especially for a task that mostly sleeps but wakes up periodically to finish a job. To address this problem, we reward underserved tasks by assigning earlier deadlines.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/util.bpf.h
+++ b/scheds/rust/scx_lavd/src/bpf/util.bpf.h
@@ -2,11 +2,6 @@
 #ifndef __UTIL_H
 #define __UTIL_H
 
-extern struct bpf_cpumask __kptr *turbo_cpumask; /* CPU mask for turbo CPUs */
-extern struct bpf_cpumask __kptr *big_cpumask; /* CPU mask for big CPUs */
-extern struct bpf_cpumask __kptr *active_cpumask; /* CPU mask for active CPUs */
-extern struct bpf_cpumask __kptr *ovrflw_cpumask; /* CPU mask for overflow CPUs */
-
 extern const volatile u64	nr_llcs;	/* number of LLC domains */
 extern const volatile u64	__nr_cpu_ids;	/* maximum CPU IDs */
 extern volatile u64		nr_cpus_onln;	/* current number of online CPUs */


scheds/rust/scx_lavd/src/bpf/lat_cri.bpf.c
```diff
@@ -148,21 +156,30 @@ static void calc_lat_cri(struct task_struct *p, task_ctx *taskc)
 
 	/*
 	 * Determine latency criticality of a task in a context-aware manner by
-	 * considering which task wakes up this task. If its waker is more
-	 * latency-critcial, inherit waker's latency criticality partially.
+	 * considering its waker and wakee's latency criticality.
```
hodgesds, line 159: Does this take into account the number of wakees. If a task wakes a number of latency critical tasks should it get extra latency criticality?

scheds/rust/scx_lavd/src/bpf/lat_cri.bpf.c
```diff
@@ -50,6 +50,14 @@ static u64 calc_weight_factor(struct task_struct *p, task_ctx *taskc)
 	if (is_kernel_worker(p))
 		weight_boost += LAVD_LC_WEIGHT_BOOST;
 
+	/*
+	 * Prioritize a task woken by an RT/DL task.
+	 */
+	if (test_task_flag(taskc, LAVD_FLAG_WOKEN_BY_RT_DL)) {
+		reset_task_flag(taskc, LAVD_FLAG_WOKEN_BY_RT_DL);
+		weight_boost += 4 * LAVD_LC_WEIGHT_BOOST;
```
etsal: Why 4 here? If it's a parameter can we add it to the enums?

multics69: > Why 4 here? If it's a parameter can we add it to the enums?

I chose 4 because it should be as important as ksoftirqd, which also uses 4.

Yes, it should be better presented rather than hardcoding the parameter. I added one more commit with LAVD_LC_WEIGHT_BOOST_REGULAR, LAVD_LC_WEIGHT_BOOST_MEDUIM, and LAVD_LC_WEIGHT_BOOST_HIGH.

scheds/rust/scx_lavd/src/bpf/lat_cri.bpf.c
multics69: The current implementation chooses the wakee with the maximum latency-criticality because it is simple to implement. :-) 

In ops.runnable(), we have:
```.c
if (waker_taskc->lat_cri_wakee < p_taskc->lat_cri)
		waker_taskc->lat_cri_wakee = p_taskc->lat_cri;
```

PR #3225, opened 2026-01-13
scx_lavd: Improve scaled load calculation and core compaction.

multics69:
Previously, we calculated the scaled load, a capacity- and frequency-invariant CPU utilization, using the (scaled) runtime of scx tasks. However, this approach underestimates utilization because it does not include other CPU time, such as RT/DL tasks and IRQs. Such an underestimated load results in suboptimal decisions in load balancing and core compaction.

Fix this problem by incorporating non-SCX compute time. And tuning a core compaction knob to better serve bursty workloads.


daidavid (approved):
Overall looks good, sc_non_scx_time will lack a bit of precision due to lack of frequency info and isn't solved by exposing rq->clock_pelt. The kernel has a similar issue when calculating irq_load, so I think it's fine.

multics69:

PR #3226, opened 2026-01-13
services: re-introduce basic systemd support (make systemd service great again)

sirlucjan:
During office hours, we agreed with @arighi and @multics69  that in certain situations, such as servers, a simple systemd service may work better than the complex scx_loader mechanism. So let's restore a simple management mechanism on servers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,10 @@ $ scx_rustland --monitor 5
 [RustLand] tasks -> r:  1/4  w: 1 /1  | pf: 0     | dispatch -> u: 33178 k: 0     c: 0     b: 0     f: 0     | cg: 0
 ```
 
+## systemd services
+
+See: [services](services/README.md)
+
 ## Kernel Feature Status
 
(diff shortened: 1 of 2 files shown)


services/scx
```diff
@@ -0,0 +1,5 @@
+# List of scx_schedulers: scx_beerland scx_bpfland scx_chaos scx_cosmos scx_flash scx_lavd scx_p2dq scx_rustland scx_rusty scx_tickless
+SCX_SCHEDULER=scx_cosmos
+
+# Set custom flags for each scheduler, below is an example of how to use
+#SCX_FLAGS='-d'
```
arighi: Can we add (uncommented):
```
SCX_FLAGS='-s 20000 -d -c 0 -p 0'
```

This should be the new default for scx_cosmos, I'm planning to change it for the next version, so maybe we can do it here for now and we provide also a real example of how to use SCX_FLAGS.

PR #3242, opened 2026-01-19
scx_lavd: Improve core compaction to use effective capacity.

multics69:
The core compaction determines the number of active CPUs based on the required compute capacity. It is assumed that each CPU can achieve its maximum architectural capacity (i.e., maximum capacity). However, the effective capacity that a CPU can achieve at a certain moment can be restricted by various factors, such as thermal throttling, limited power budget, and user-enforced policy (e.g., setting “Power Mode” to “Power Save” mode or setting the “TDP Limit” to “3W”). If the effective capacity falls below the maximum capacity, the core compaction logic underestimates the required number of active CPUs, leading to performance issues.

This PR addresses such a problem with the following main changes:
1) Estimate the effective capacity of a CPU.
2) Use the effective capacity in determining the number of active CPUs in core compaction.
3) Always give 25% headroom when calculating the required capacity to handle load spikes better.

Besides these main changes, there are shotgun changes here and there to make the verifier happy. :-(

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/power.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/power.bpf.c
@@ -270,9 +270,11 @@ static int calc_nr_active_cpus(void)
 	u16 cpu;
 
 	/*
-	 * First, calculate the required compute capacity:
+	 * First, calculate the required compute capacity. Give some (say 25%)
+	 * headroom to handle sudden load spikes smoothly.
 	 */
 	req_cap = calc_required_capacity();
+	req_cap += (req_cap * LAVD_CC_REQ_CAPACITY_HEADROOM) >> LAVD_SHIFT;
 
 	/*
 	 * Then, determine the number of active CPUs that meet the required
--- a/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
+++ b/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
@@ -86,14 +86,14 @@ enum consts_internal {
 	LAVD_LC_INH_RECEIVER_SHIFT	= 2, /* 25.0% of receiver's latency criticality */
 	LAVD_LC_INH_GIVER_SHIFT		= 3, /* 12.5 of giver's latency criticality */
 
-	LAVD_CPU_UTIL_MAX_FOR_CPUPERF	= p2s(85), /* 85.0% */
-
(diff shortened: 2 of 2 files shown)


scheds/rust/scx_lavd/src/bpf/power.bpf.c
```diff
@@ -231,25 +323,45 @@ static int calc_nr_active_cpus(void)
...
-				WRITE_ONCE(pco_idx, i);
-				return pco_nr_primary[i];
+				const volatile u16 *cpu_order = pco_table[i];
+				sum_eff_cap = 0;
+
+				bpf_for(j, 0, pco_nr_primary[i]) {
+					if (j >= LAVD_CPU_ID_MAX)
+						break;
+
+					cpu = cpu_order[i];
```
bboymimi: With energy model awareness, 'i' indicates the pco_index. 'j' should be the index to fetch the cpu_order[j].

multics69: You are right. Fixed it.

scheds/rust/scx_lavd/src/bpf/power.bpf.c
```diff
@@ -92,6 +93,91 @@ volatile int		power_mode;
...
+	 *
+	 * Note that cpuc->max_freq is [0:1024].
+	 */
+	mfo = __sync_val_compare_and_swap(&cpuc->max_freq_observed,
+					  cpuc->max_freq_observed, 0);
+	if ((mfo > 0) && ((cpuc->max_freq < mfo) ||
+	    (cpuc->cur_util >= LAVD_CPU_UTIL_THR_FOR_MAX_FREQ))) {
+		cpuc->max_freq = calc_avg32(cpuc->max_freq, mfo);
+	}
+	capacity_observed = (cpuc->max_capacity * cpuc->max_freq) >> LAVD_SHIFT;
```
bboymimi, line 165: I like the idea of retrieving the observed capacity based on the running frequency average.

multics69: This was an inevitable choice because much of the CPU frequency management is dedicated to the system-management microcontroller in many systems. Some partially expose their frequency decision information to sw, but others don't do so at all. The approach is more like a best-effort approach to find the effective capacity based on exposed constraints and observed capacity.

bboymimi (approved):

PR #3243, opened 2026-01-20
Update service README

Sisyphus1813:
Updated the service folder README to the most recent instructions and configuration settings for the system service. Some of these instructions are behavior I have observed whilst using the service, other of these were posted in (the now gone) scx/services/systemd/MIGRATION_GUIDE.md provided by commit 22fe5e758587e70b10dadc1a94541d3fa8663f59. This is not fully fleshed out but it is a start; and the current services README contains deprecated information.

### DIFF
--- a/services/README.md
+++ b/services/README.md
@@ -4,11 +4,29 @@ This guide provides instructions for running the SCX schedulers as a systemd ser
 
 ## Getting Started
 
-At the very beginning, configure the /etc/default/scx file:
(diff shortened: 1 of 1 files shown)


sirlucjan:
scx_loader is no longer part of the scx repository and is provided separately: https://github.com/sched-ext/scx-loader  The scx repository provides scx.service again, so README.md is correct.

PR #3250, opened 2026-01-22
scx_mitosis: Reject partial multi-CPU pinning by default

tommy-u:
Add check to error out when a task from cell >0 has multi-CPU affinity that doesn't cover the entire cell. Single-CPU pinning remains allowed.

Use --enable-multicpu-pinning to opt-in to the old behavior at the risk of reduced performance

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -36,6 +36,7 @@ const volatile bool	     debug_events_enabled	     = false;
 const volatile bool	     exiting_task_workaround_enabled = true;
 const volatile bool	     split_vtime_updates	     = false;
 const volatile bool	     cpu_controller_disabled	     = false;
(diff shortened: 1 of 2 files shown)


likewhatevs:
This probably should not be default because kicking the scheduler comes with a performance cost also and this introduces the risk that some random library affinitizing it's threadpool causes mass scheduler kicks.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -36,6 +36,7 @@ const volatile bool	     debug_events_enabled	     = false;
 const volatile bool	     exiting_task_workaround_enabled = true;
 const volatile bool	     split_vtime_updates	     = false;
 const volatile bool	     cpu_controller_disabled	     = false;
+const volatile bool	     enable_multicpu_pinning	     = false;
```
dschatzberg: Let's make the default true which is consistent with the current behavior (no changes) - just to enable us to deploy the scheduler binary independently from the cli argument changes

PR #3264, opened 2026-01-28
scx_beerland: Reduce scheduling overhead

arighi:
A set of small cleanups and optimization to reduce scheduling overhead in scx_beerland, reducing unnecessary migrations when there's not too contention on the CPUs and exploiting more the benefits of direct dispatch in ops.select_cpu().

With these changes in place the scheduler seems to provide better fps (average and 1p low) in games without regressing the usual stress-ng / schbench / perf bench tests.

### DIFF
--- a/scheds/rust/scx_beerland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_beerland/src/bpf/main.bpf.c
@@ -609,20 +609,39 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, s32 this_cpu, u64 
 }
 
 /*
+ * Return true if @p can run on @cpu, false otherwise.
+ */
+static bool is_cpu_allowed(const struct task_struct *p, s32 cpu)
+{
+	return p->nr_cpus_allowed == nr_cpu_ids ||
+	       bpf_cpumask_test_cpu(cpu, p->cpus_ptr);
+}
+
+/*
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_beerland/src/bpf/main.bpf.c
```diff
@@ -630,21 +608,40 @@ static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, s32 this_cpu, u64
...
+
+	/*
+	 * If there's no task queued either in the previous or the current
+	 * CPU's DSQ, assume the CPUs are not too contended, so we can
+	 * directly dispatch to the local DSQ and save some unnecessary
+	 * scheduling overhead.
+	 */
+	if (!__COMPAT_scx_bpf_dsq_peek(prev_cpu) ||
+	    !__COMPAT_scx_bpf_dsq_peek(this_cpu))
+		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, task_slice(p), 0);
```
hodgesds, line 637: do you want to return after this?

arighi: No return, I still want to find an idle CPU. Essentially I'm saying: let's direct dispatch in advance, even before looking for an idle CPU, because if there's nothing queued in prev_cpu or this_cpu, then the system is not too overloaded and I can still benefit from bypassing the scheduler's logic.

PR #3270, opened 2026-01-30
scx_mitosis: Initial support for non-cpuset cell assignment

dschatzberg:
Currently Mitosis creates cells from bpf purely based on cpusets. We want to make mitosis capable of creating cells without strictly cpu pinning cgroups. This PR implements an initial naive version that:

1) Adds a new CLI flag: --cell-parent-cgroup <path>: When specified, direct children of this cgroup path automatically become separate cells rather than creating cells only from cpuset cgroups.
2) Creates a userspace cell manager (cell_manager.rs) which observes creation/destruction of cgroups and triggers reconfiguration from BPF (as opposed to triggered from BPF as is done today).

Future work here will extend this to make cell <=> cpu assignment topology aware, utilization aware, etc. And enable allocation of CPUs through configuration.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -62,6 +62,7 @@ struct llc_cpumask llc_to_cpus[MAX_LLCS];
  */
 u32 configuration_seq;
 u32 applied_configuration_seq;
+u32 cpuset_seq;
 
 /*
  * Debug events circular buffer
@@ -1422,10 +1423,17 @@ int BPF_PROG(fentry_cpuset_write_resmask, struct kernfs_open_file *of,
 	     char *buf, size_t nbytes, loff_t off, ssize_t retval)
 {
 	/*
-	 * On a write to cpuset.cpus, we'll need to configure new cells, bump
-	 * configuration_seq so tick() does that.
+	 * On a write to cpuset.cpus, we'll need to reconfigure cells.
+	 *
+	 * In userspace-managed mode, bump cpuset_seq so userspace re-reads
+	 * cpusets and recomputes CPU assignments.
+	 *
+	 * In auto mode, bump configuration_seq so tick() reconfigures cells.
 	 */
-	__atomic_add_fetch(&configuration_seq, 1, __ATOMIC_RELEASE);
+	if (userspace_managed_cell_mode)
+		__atomic_add_fetch(&cpuset_seq, 1, __ATOMIC_RELEASE);
+	else
+		__atomic_add_fetch(&configuration_seq, 1, __ATOMIC_RELEASE);
 	return 0;
 }
 
--- a/scheds/rust/scx_mitosis/src/cell_manager.rs
+++ b/scheds/rust/scx_mitosis/src/cell_manager.rs
@@ -434,33 +434,15 @@ impl CellManager {
     fn create_cell_for_cgroup(&mut self, path: &Path, cgid: u64) -> Result<(u64, u32)> {
         let cell_id = self.allocate_cell_id()?;
 
-        // Try to read cpuset.cpus for this cgroup
-        let cpuset_path = path.join("cpuset.cpus");
-        let cpuset = match std::fs::read_to_string(&cpuset_path) {
-            Ok(content) => {
-                let content = content.trim();
-                if content.is_empty() {
-                    None
-                } else {
-                    let mask = Cpumask::from_cpulist(content).with_context(|| {
-                        format!(
-                            "Failed to parse cpuset '{}' from {}",
-                            content,
-                            cpuset_path.display()
-                        )
-                    })?;
-                    debug!(
-                        "Cell {} has cpuset: {} (from {})",
-                        cell_id,
-                        content,
-                        cpuset_path.display()
-                    );
-                    Some(mask)
-                }
-            }
-            // File doesn't exist - cpuset controller is not enabled for this cgroup
-            Err(_) => None,
-        };
+        let cpuset = Self::read_cpuset(path)?;
+        if let Some(ref mask) = cpuset {
+            debug!(
+                "Cell {} has cpuset: {} (from {})",
+                cell_id,
+                mask.to_cpulist(),
+                path.join("cpuset.cpus").display()
+            );
+        }
 
         self.cells.insert(
             cgid,
@@ -481,6 +463,30 @@ impl CellManager {
         );
 
         Ok((cgid, cell_id))
+    }
+
+    /// Read cpuset.cpus from a cgroup path. Returns None if empty or unavailable.
+    fn read_cpuset(cgroup_path: &Path) -> Result<Option<Cpumask>> {
+        let cpuset_path = cgroup_path.join("cpuset.cpus");
+        match std::fs::read_to_string(&cpuset_path) {
+            Ok(content) => {
+                let content = content.trim();
+                if content.is_empty() {
+                    Ok(None)
+                } else {
+                    let mask = Cpumask::from_cpulist(content).with_context(|| {
+                        format!(
+                            "Failed to parse cpuset '{}' from {}",
+                            content,
+                            cpuset_path.display()
+                        )
+                    })?;
+                    Ok(Some(mask))
+                }
+            }
+            // File doesn't exist - cpuset controller is not enabled for this cgroup
+            Err(_) => Ok(None),
+        }
     }
 
     fn allocate_cell_id(&mut self) -> Result<u32> {
@@ -794,6 +800,30 @@ impl CellManager {
             }
         }
         parts.join(" ")
+    }
+
+    /// Re-read cpuset.cpus for all cells and update stored cpusets.
+    /// Returns true if any cell's cpuset changed.
+    pub fn refresh_cpusets(&mut self) -> Result<bool> {
+        let mut changed = false;
+        for info in self.cells.values_mut() {
+            let Some(ref cgroup_path) = info.cgroup_path else {
+                continue; // cell 0 has no cgroup
+            };
+            let new_cpuset = Self::read_cpuset(cgroup_path)?;
+            if new_cpuset != info.cpuset {
+                info!(
+                    "Cell {} cpuset changed: {:?} -> {:?} ({})",
+                    info.cell_id,
+                    info.cpuset.as_ref().map(|m| m.to_cpulist()),
+                    new_cpuset.as_ref().map(|m| m.to_cpulist()),
+                    cgroup_path.display(),
+                );
+                info.cpuset = new_cpuset;
+                changed = true;
+            }
+        }
+        Ok(changed)
     }
 }
 
--- a/scheds/rust/scx_mitosis/src/main.rs
+++ b/scheds/rust/scx_mitosis/src/main.rs
@@ -18,6 +18,7 @@ use std::fmt::Display;
 use std::mem::MaybeUninit;
 use std::os::fd::AsFd;
 use std::sync::atomic::AtomicBool;
+use std::sync::atomic::AtomicU32;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::Duration;
@@ -216,6 +217,8 @@ struct Scheduler<'a> {
     metrics: Metrics,
     stats_server: Option<StatsServer<(), Metrics>>,
     last_configuration_seq: Option<u32>,
+    /// Last observed cpuset_seq for cpuset change detection
+    last_cpuset_seq: u32,
     /// Optional cell manager for --cell-parent-cgroup mode
     cell_manager: Option<CellManager>,
     /// Whether CPU borrowing is enabled
@@ -409,6 +412,7 @@ impl<'a> Scheduler<'a> {
             metrics: Metrics::default(),
             stats_server: Some(stats_server),
             last_configuration_seq: None,
+            last_cpuset_seq: 0,
             cell_manager,
             enable_borrowing: opts.enable_borrowing,
             enable_rebalancing: opts.enable_rebalancing,
@@ -477,6 +481,7 @@ impl<'a> Scheduler<'a> {
 
             // Periodic work on every iteration
             self.refresh_bpf_cells()?;
+            self.check_cpuset_changes()?;
             self.collect_metrics()?;
 
             if self.enable_rebalancing && self.cell_manager.is_some() {
@@ -1142,6 +1147,38 @@ impl<'a> Scheduler<'a> {
         self.metrics
             .update_demand(global_util_pct, global_borrow_pct, global_lent_pct);
 
+        Ok(())
+    }
+
+    /// Check if any cell's cpuset was modified and recompute if so.
+    fn check_cpuset_changes(&mut self) -> Result<()> {
+        let Some(ref mut cm) = self.cell_manager else {
+            return Ok(());
+        };
+
+        let current_seq = unsafe {
+            let ptr = &self.skel.maps.bss_data.as_ref().unwrap().cpuset_seq as *const u32;
+            (ptr as *const AtomicU32)
+                .as_ref()
+                .unwrap()
+                .load(Ordering::Acquire)
+        };
+
+        if current_seq == self.last_cpuset_seq {
+            return Ok(());
+        }
+        self.last_cpuset_seq = current_seq;
+
+        if !cm.refresh_cpusets()? {
+            return Ok(()); // seq changed but no cpusets on our cells changed
+        }
+
+        let cpu_assignments = self.compute_and_apply_cell_config(&[])?;
+        let cell_manager = self.cell_manager.as_ref().unwrap();
+        info!(
+            "Cpuset change detected, recomputed config: {}",
+            cell_manager.format_cell_config(&cpu_assignments)
+        );
         Ok(())
     }
 
--- a/scheds/rust/scx_mitosis/test/test_cell_isolation.sh
+++ b/scheds/rust/scx_mitosis/test/test_cell_isolation.sh
@@ -10,7 +10,7 @@
 #   ./test_cell_isolation.sh [--cpuset] [--proportional] [--num-cells N] [--workers N]
 #                            [--test-dynamic] [--test-borrowing]
 #                            [--test-enqueue-borrowing]
-#                            [--test-rebalancing] [--test-all]
+#                            [--test-rebalancing] [--test-cpuset-change] [--test-all]
 #
 # Options:
 #   --cpuset        Test cpuset-based cell isolation (requires cpuset configuration)
@@ -19,8 +19,9 @@
 #   --workers N     Number of stress-ng workers per cell (default: 5)
 #   --test-dynamic  Run dynamic cell creation/destruction tests
 #   --test-rebalancing Run demand-based CPU rebalancing test
+#   --test-cpuset-change Run cpuset change detection test
 #   --test-new-cell-demand Run new cell demand seeding test
-#   --test-all      Run all tests (basic + dynamic + borrowing + rebalancing + new-cell-demand)
+#   --test-all      Run all tests (basic + dynamic + borrowing + rebalancing + cpuset-change + new-cell-demand)
 #   --help          Show this help message
 #
 
@@ -41,6 +42,7 @@ TEST_DYNAMIC=0
 TEST_BORROWING=0
 TEST_ENQUEUE_BORROWING=0
 TEST_REBALANCING=0
+TEST_CPUSET_CHANGE=0
 TEST_NEW_CELL_DEMAND=0
 TEST_ALL=0
 
@@ -105,6 +107,10 @@ while [[ $# -gt 0 ]]; do
             TEST_REBALANCING=1
             shift
             ;;
+        --test-cpuset-change)
+            TEST_CPUSET_CHANGE=1
+            shift
+            ;;
         --test-new-cell-demand)
             TEST_NEW_CELL_DEMAND=1
             shift
@@ -152,7 +158,7 @@ cleanup() {
     done
 
     # Remove dynamic test cgroups
-    for cell_name in test_cell_dynamic test_cell_temp test_cell_3 test_cell_reuse_a test_cell_reuse_b test_cell_borrow_busy test_cell_borrow_idle test_cell_rebal_busy test_cell_rebal_idle test_cell_demand_a test_cell_demand_b test_cell_demand_c; do
+    for cell_name in test_cell_dynamic test_cell_temp test_cell_3 test_cell_reuse_a test_cell_reuse_b test_cell_borrow_busy test_cell_borrow_idle test_cell_rebal_busy test_cell_rebal_idle test_cell_cpuset_a test_cell_cpuset_b test_cell_demand_a test_cell_demand_b test_cell_demand_c; do
         cell_path="$CGROUP_BASE/$cell_name"
         if [[ -d "$cell_path" ]]; then
             for pid in $(cat "$cell_path/cgroup.procs" 2>/dev/null); do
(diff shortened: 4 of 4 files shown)


scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -0,0 +1,646 @@
...
+    /// CPUs are divided equally among cell 0 and all user cells.
+    /// Any remainder CPUs go to cell 0.
+    ///
+    /// Returns a Vec of (cell_id, Cpumask), or an error if any cell would
+    /// receive zero CPUs (which indicates too many cells for available CPUs).
+    pub fn compute_cpu_assignments(&self) -> Result<Vec<(u32, Cpumask)>> {
+        let mut cell_cpus: HashMap<u32, Cpumask> = HashMap::new();
+
+        // All CPUs are distributed among cell 0 + all cells
+        let all_cell_ids: Vec<u32> = self.cells.values().map(|c| c.cell_id).collect();
```
tommy-u: If I'm getting this right `all_cell_ids` means "all cell ids except the root cell." That's definitely too confusing for me. How about `all_workload_cells`?

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -0,0 +1,646 @@
...
+                        .or_insert_with(Cpumask::new)
+                        .set_cpu(cpu)
+                        .ok();
+                }
+            }
+        }
+
+        // Verify all cells have at least one CPU assigned
+        // Cell 0 always exists
+        if !cell_cpus.contains_key(&0) || cell_cpus.get(&0).map_or(true, |m| m.weight() == 0) {
```
tommy-u: How about a constant to keep things self documenting
const SYSTEM_CELL_ID: u32 = 0;

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -0,0 +1,646 @@
...
+    }
+
+    /// Compute CPU assignments for all cells.
+    ///
+    /// CPUs are divided equally among cell 0 and all user cells.
+    /// Any remainder CPUs go to cell 0.
+    ///
+    /// Returns a Vec of (cell_id, Cpumask), or an error if any cell would
+    /// receive zero CPUs (which indicates too many cells for available CPUs).
+    pub fn compute_cpu_assignments(&self) -> Result<Vec<(u32, Cpumask)>> {
```
tommy-u: Love this approach!

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -264,19 +290,53 @@ impl<'a> Scheduler<'a> {
...
+        let stats_bridge = std::thread::spawn(move || {
+            while req_ch.recv().is_ok() {
+                // Wake up main loop via eventfd
+                let _ = stats_waker.write(1);
+            }
+        });
 
         while !shutdown.load(Ordering::Relaxed) && !uei_exited!(&self.skel, uei) {
+            let mut events = [EpollEvent::empty(); 1];
+            let timeout_ms = self.monitor_interval.as_millis() as u16;
```
tommy-u: I might be reading this wrong, but would putting a monitor_interval of greater than 65 seconds cause this to roll over?

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -246,67 +277,155 @@ impl CellManager {
...
+        let mut contention: HashMap<usize, Vec<u32>> = HashMap::new();
+        for cell_info in self.cells.values() {
+            if let Some(ref cpuset) = cell_info.cpuset {
+                for cpu in cpuset.iter() {
+                    contention.entry(cpu).or_default().push(cell_info.cell_id);
+                }
+            }
+        }
+
+        // Phase 2: Categorize CPUs and build initial assignments
```
tommy-u: I forget who I am halfway down 200 line functions. Could these phases be natural breaks for fn calls?

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -246,67 +277,155 @@ impl CellManager {
...
+                        .entry(cell_id)
+                        .or_insert_with(Cpumask::new)
+                        .set_cpu(cpu)
+                        .ok();
+                }
+                Some(_) => contested_cpus.push(cpu),
+            }
+        }
+
+        // Phase 3: Distribute contested CPUs proportionally among claimants
```
tommy-u: Split the baby. very king solomon

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -79,8 +79,36 @@ struct cgrp_ctx {
...
 
+/* Cell assignment entry: maps a cgroup to a cell */
+struct cell_assignment {
+	u64 cgid; /* cgroup ID (from cgroup file inode) */
+	u32 cell_id; /* cell ID to assign */
+};
+
+/* Cell cpumask data for a single cell */
+struct cell_cpumask_data {
+	unsigned char mask[MAX_CPUS_U8];
```
tommy-u, line 172: Opportunity to use arenas?

scheds/rust/scx_mitosis/src/cell_manager.rs
tommy-u: Personal preference, but as this fn grows longer, it would be easier for me to understand and modify if it were decomposed into separate functions.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1871,6 +1884,267 @@ void BPF_STRUCT_OPS(mitosis_exit, struct scx_exit_info *ei)
...
+ *
+ * The function operates in five phases:
+ * 1. Mark all cells (except cell 0) as not in use
+ * 2. Apply cell assignments for owner cgroups
+ * 3. Walk cgroup hierarchy to propagate cells to children
+ * 4. Apply cell cpumasks and CPU-to-cell mappings
+ * 5. Bump applied_configuration_seq to signal completion
+ *
+ * Note: This is not atomic - tasks may observe intermediate states during
+ * execution. On error, the scheduler may be left in a partially-configured
```
tommy-u, line 2184: This sounds bad, but probably fine in practice. Can you say another word about why we should feel chill about it?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1871,6 +1884,267 @@ void BPF_STRUCT_OPS(mitosis_exit, struct scx_exit_info *ei)
...
+	struct cgrp_ctx		    *cgc;
+	struct cell		    *cell;
+	struct cpu_ctx		    *cctx;
+	struct cell_cpumask_wrapper *cpumaskw;
+	struct bpf_cpumask	    *new_cpumask;
+	struct cgroup_subsys_state  *root_css, *pos;
+	struct cgroup		    *cur_cgrp, *root_cgrp_ref;
+	u32			     i, cell_id;
+	int			     ret = 0;
+
```
tommy-u, line 2198: Personal preference, but I struggle to hold this many ideas in my head at a time. Maybe can procedurally decompose like on the userspace side.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1871,6 +1884,267 @@ void BPF_STRUCT_OPS(mitosis_exit, struct scx_exit_info *ei)
...
+ * 4. Apply cell cpumasks and CPU-to-cell mappings
+ * 5. Bump applied_configuration_seq to signal completion
+ *
+ * Note: This is not atomic - tasks may observe intermediate states during
+ * execution. On error, the scheduler may be left in a partially-configured
+ * state. This is acceptable because userspace treats errors as fatal and
+ * exits, causing the scheduler to be unloaded.
+ */
+SEC("syscall")
+int apply_cell_config(void *ctx)
```
tommy-u, line 2189: I really like this overall. It feels like we're moving way closer to a simpler world driven by userspace.

But can't we still move most of this function to userspace? We're almost entirely looking up and processing data that is available in userspace. Why don't we precompute all that and let this function be like 4 trivial loops that write it into an essential bpf datastructures?

Also, possible opportunity to pass data via ctx?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1871,6 +1884,267 @@ void BPF_STRUCT_OPS(mitosis_exit, struct scx_exit_info *ei)
...
+
+		u64 cgid   = assignment->cgid;
+		cell_id	   = assignment->cell_id;
+
+		if (cell_id >= MAX_CELLS)
+			return -EINVAL;
+
+		cgrp = bpf_cgroup_from_id(cgid);
+		if (!cgrp)
+			return -ENOENT;
```
tommy-u: scx_bpf_error here?

tommy-u (changes requested):
I think this is a huge step toward a scheduler that's more understandable and easier to modify, fix, and extend. It's got me excited about how to design across the kernel boundary.

High level things for me are terminology and opportunities to shift code out of apply_cell_config.

1) Terminology. Right now, terminology is a bit messy. We have "Cell 0" or the "root cell." I saw "all_cell_ids" which was only the workload cell ids (I think this went away in a later commit).

This is a little tricky because the terms we need are doing double duty between 2 views:
- cgroups - hierarchical
- cells - flat

So how about we have a "system" cgroup(s)/cell and multiple "workload" cgroups/cells. Not married to these words, but I'd really like to drop "Cell 0" (implementation specific) and "root" which fits for the cgroup view, but is misleading in the cell view.

2) I really like what you're cooking in apply_cell_config(). Why not take it further though? It seems like almost all the data you're accessing and processing in there is available for userspace to precompute and hand over for apply_cell_config to apply trivially.

I think user code is >10x easier to change and maintain than bpf, so I think it would be way easier to test new ideas. Making stuff up here, but e.g. if we didn't hardcode assigning child cgroup's to same cell as the workload owner, someone could play with extracting child cgroups into their own cells, maybe there's an application with virtualization... idk. Policy <-> Mechanism

scheds/rust/scx_mitosis/Cargo.toml
```diff
@@ -19,6 +19,8 @@ itertools = "0.13.0"
 lazy_static = "1.5.0"
 libbpf-rs = "=0.26.0-beta.1"
 libc = "0.2.175"
+inotify = "0.10"
```
likewhatevs: layered uses 0.11, easier using that imo.

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -264,23 +327,216 @@ impl<'a> Scheduler<'a> {
...
+
+            match self
+                .epoll
+                .wait(&mut events, EpollTimeout::try_from(timeout_ms)?)
+            {
+                Ok(n) => {
+                    for event in &events[..n] {
+                        match event.data() {
+                            INOTIFY_TOKEN => {
```
likewhatevs, line 465: IIUC, to make things simpler/less fragile, we're doing a kinda-sorta figure out the state of the world/reconcile thing here vs something more granular/delta based.

If so, I think there should be some debounce here for when things thrash/to avoid reconcile amplifying that.

scheds/rust/scx_mitosis/src/cell_manager.rs
dschatzberg: Yeah, it might be worth some additional logic just to create cell 0. I get you on the terminology point but I don't love the workload/system distinction (nothing seems to indicate to me that we wouldn't want multiple "system" cells). I could maybe call it the default cell? But for whatever reason root seems most natural to me. I acknowledge that its coupled with the cgroup hierarchy, but that's also how I think about it?

scheds/rust/scx_mitosis/src/cell_manager.rs
dschatzberg: > Personal preference, but as this fn grows longer, it would be easier for me to understand and modify if it were decomposed into separate functions.

Yeah, I think this is just my personal preference that I like the linearity of a single function (with comments to make it easier to follow) rather than breaking it up into multiple functions that requires jumping around to understand the entire logic.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Are you asking about the error case or the non-atomic aspect?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: I'm curious - how do multiple functions differ from the phases w/ block comments differ for you? I prefer the latter because I read the code linearly rather than needing to jump around.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: What parts are you suggesting could be moved to userspace?

All the cgroup lookups and walking the cgroup hierarchy would be heinously expensive in userspace. While this isn't a particularly perf sensitive path, it's pretty egregious to have to do so many filesystem syscalls compared to the in-kernel iteration.

The rest is really just atomically publishing the cpumasks - I'm not sure how that can be done from userspace.

I guess I could bump the configuration_seq from userspace but it seems cleaner to keep the parts that actually synchronize in the same code base.

scheds/rust/scx_mitosis/src/main.rs
dschatzberg: Can you expand on what your concern is regarding thrashing?

scheds/rust/scx_mitosis/src/cell_manager.rs
tommy-u: I don't mean to bikeshed, anything is better than cell 0. I can live with with "root" and something to mean non-root. Let's chat offline :)

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: Sorry for the ambiguity in highlighting two ideas here.

My question was about the non-atomic aspect here. 

I think there is an implicit statement that might help a reader, something like: "The worst case is tasks transiently running on stale CPUs while applying cell config runs."

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: Skill issue on my part, probably. But here's my case for it not being arbitrary.

The one-big-function approach is like a cookbook that lists every ingredient for every recipe upfront—then you read each recipe with no idea what you actually need until you're done.

One big function:
- 20+ local variables, any of which could be used at any point
- You can't mentally set any of them down until you've read everything

Small functions:
- Natural summary of steps with none of the implementation detail
- Explicit data flow (prepared by one, consumed by another)
- Compiler complains if you pass unused variables to minimize scope

Keep the summary open on one window to keep the big picture, jump around to each fn on the other.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: I think Phase 3 is the best candidate. I tried to sketch an example in my overall note where we might be able to remove a hard-coded policy assumption that child cgroups should be mapped to the same cell as the owning ancestor.

I was thinking of this as something that only happens on the ~hours time scale. And that you'd only have to walk cgroup in the workload path because all the rest map to the root cell, which would be pretty cheap?

scheds/rust/scx_mitosis/src/main.rs
likewhatevs: So, we use inotify to trigger a full "figure out where things should run/what CPUs they should get".

If we have some container that keeps getting created/destroyed (say, something crashlooping and getting destroyed/created over and over), we could keep re-figuring out where things should run with different outcomes (i.e. split 4 ways, split 3 ways).

I think this issue applies even w/o a crashlooping container, but it'd be tolerable given spacing between create/destroy (i.e. if u care about perf, you aren't doing that already), but I think the crashloop case would be the particularly bad one because we'd have relentless migrations and task spraying, iiuc.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -2017,6 +2167,331 @@ void BPF_STRUCT_OPS(mitosis_exit, struct scx_exit_info *ei)
...
+	struct cgrp_ctx		    *cgc;
+	struct cell		    *cell;
+	struct cpu_ctx		    *cctx;
+	struct cell_cpumask_wrapper *cpumaskw;
+	struct cgroup_subsys_state  *root_css, *pos;
+	struct cgroup		    *cur_cgrp;
+	u32			     i, cell_id;
+
+	/* Read configuration from global struct (populated by userspace) */
+	struct cell_config *config = &cell_config;
```
tommy-u, line 2200: Can't we pass `cell_config` as an arg to the bpf syscall `ctx` parameter

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -251,6 +329,14 @@ impl<'a> Scheduler<'a> {
         rodata.enable_llc_awareness = opts.enable_llc_awareness;
         rodata.enable_work_stealing = opts.enable_work_stealing;
 
+        skel.maps
+            .rodata_data
+            .as_mut()
+            .expect("rodata_data must be available during init")
+            .userspace_managed_cell_mode = opts.cell_parent_cgroup.is_some();
+
+        skel.maps.rodata_data.as_mut().unwrap().enable_borrowing = opts.enable_borrowing;
```
tommy-u: can use `rodata`

scheds/rust/scx_mitosis/src/bpf/intf.h
```diff
@@ -71,13 +71,15 @@ enum cell_stat_idx {
...
 	CSTAT_CELL_DSQ,
 	CSTAT_AFFN_VIOL,
+	CSTAT_BORROWED,
 	CSTAT_STEAL,
 	NR_CSTATS,
 };
 
 struct cpu_ctx {
 	u64 cstats[MAX_CELLS][NR_CSTATS];
 	u64 cell_cycles[MAX_CELLS];
```
tommy-u, line 81: The history of the commit stack makes this confusing. I think `cell_cycles` dead code now.

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -288,23 +428,306 @@ impl<'a> Scheduler<'a> {
...
+            self.cells
+                .get(&a.cell_id)
+                .map_or(true, |cell| cell.cpus != a.primary)
+        });
+
+        if !changed {
+            return Ok(());
+        }
+
+        let cpu_assignments = self.compute_and_apply_cell_config_demand(&cell_demands)?;
```
tommy-u: nit: Seems redundant to recompute the cell assignment you already got in new_assignments?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -390,6 +411,15 @@ static inline int update_task_cpumask(struct task_struct *p,
...
 			bpf_cpumask_subset(cell_cpumask, p->cpus_ptr);
 
+	if (tctx->all_cell_cpus_allowed && enable_borrowing) {
+		const struct cpumask *borrowable =
+			lookup_cell_borrowable_cpumask(tctx->cell);
+		if (!borrowable)
+			return -ENOENT;
+		if (!bpf_cpumask_subset(borrowable, p->cpus_ptr))
+			tctx->all_cell_cpus_allowed = false;
+	}
```
tommy-u, line 421: This looks like a bug if I'm understanding.

If borrowable is not a subset of p->cpus_ptr, we map it down to a single cpu?

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -288,23 +428,306 @@ impl<'a> Scheduler<'a> {
...
+        let stats_waker = unsafe { EventFd::from_owned_fd(stats_waker_fd) };
+        let stats_bridge = std::thread::spawn(move || {
+            while req_ch.recv().is_ok() {
+                // Wake up main loop via eventfd
+                let _ = stats_waker.write(1);
+            }
+        });
 
         while !shutdown.load(Ordering::Relaxed) && !uei_exited!(&self.skel, uei) {
+            let mut events = [EpollEvent::empty(); 1];
```
tommy-u, line 453: Claude thinks this should have 2 elements:
inotify fd — for cpuset changes
stats eventfd — for stats requests

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -127,6 +140,40 @@ struct Opts {
...
+    /// Utilization spread (max - min) that triggers rebalancing (default: 20%)
+    #[clap(long, default_value = "20.0")]
+    rebalance_threshold: f64,
+
+    /// Minimum seconds between rebalancing events (default: 5)
+    #[clap(long, default_value = "5")]
+    rebalance_cooldown_s: u64,
+
+    /// EWMA smoothing factor for demand tracking. Higher = more responsive (default: 0.3)
+    #[clap(long, default_value = "0.3")]
```
tommy-u: check range or this will do weird stuff :)

#[clap(long, default_value = "0.1", value_parser = clap::value_parser!(f64).range(0.0..=1.0))]

tommy-u (changes requested):
I really like this & can't wait to get to testing it.

My only concern is marking a task !all_cell_cpus_allowed when borrowable cpuset is not a subset of the task cpuset. This seems like another version of the multi-cpu-pinning issue. Do we just need a second variable like "can_borrow" which we disable here?

Or maybe I misunderstand the borrowable cpuset or an assumption that we'll only be here when there is no pinning. https://github.com/sched-ext/scx/pull/3270/changes#r2850203114

Anyways, looks awesome. Let's fire it up

scheds/rust/scx_mitosis/src/bpf/intf.h
dschatzberg: cell_cycles is currently unused by userspace. I went ahead and repurposed it for the utilization tracking that was added in this stack

scheds/rust/scx_mitosis/src/main.rs
dschatzberg: Hmm, this is just a buffer for a single epoll event. It gets populated in the self.epoll.wait call and then processed after which it will loop and wait again. It just means we only process one even per loop iteration. What you're proposing is that we potentially process two events per loop but I don't see why that matters unless we were worried about epoll syscall overheads.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: Discussed offline - this is just because a cell's cpuset is now the union of its owned cpus (its cell_cpumask) and its borrowable cpus so to make sure a process can use any cpu in the cell's cpuset, we need this check as well

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: @kkdwvd maybe has more context here but I tried this and it failed with some verifier errors that I've since lost. To the best my understanding the verifier requires that offsets into any BPF_PROG_TYPE_SYSCALL are compile-time constants but we do a bunch of indexing into arrays (which create variable-offsets).

scheds/rust/scx_mitosis/src/main.rs
dschatzberg: @likewhatevs - sorry I missed this response. I think it's a valid concern but I don't see this "thrashing" as being so impactful to actual workloads since we'll continue to have preference for prev CPU and other idle CPU detection.

But I'm also not sure what the alternative would be - if a cell leaves the machine or joins the machine, we need to react to that.

PR #3278, opened 2026-02-02
scx_lavd: Remove an obsolete global variable, __nr_cpu_ids.

multics69:
Also, move set_on_cpu_type() inside RCU read lock to avoid the verifier error.


bboymimi:
The PR removes redundant code and fixes an RCU protection issue. It looks good to me.

PR #3280, opened 2026-02-02
scx_lavd: Inherit scheduling context from parent on fork.

multics69:
When a task is initialized under SCX control (e.g. at fork), initialize its LAVD task context by inheriting the parent’s properties where possible.

Freshly forked tasks typically share memory footprint and execution characteristics with their parent. Inheriting latency and performance-criticality parameters helps preserve scheduling behavior until the child establishes its own runtime profile.

If the parent context cannot be safely accessed (e.g. RT/DL parent or lookup failure), fall back to initializing the task as an average fair task.

This uses a best-effort parent lookup via bpf_task_from_pid() to work around current BPF restrictions on trusting p->real_parent.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -1675,9 +1675,10 @@ void BPF_STRUCT_OPS(lavd_enable, struct task_struct *p)
 s32 BPF_STRUCT_OPS_SLEEPABLE(lavd_init_task, struct task_struct *p,
 			     struct scx_init_task_args *args)
 {
-	task_ctx *taskc;
-	u64 now;
+	task_ctx *taskc, *taskc_parent;
+	struct task_struct *parent;
 	int i;
+	u64 now;
 
 	/*
 	 * When @p becomes under the SCX control (e.g., being forked), @p's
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1699,38 +1700,54 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(lavd_init_task, struct task_struct *p,
...
+	 * its parent is an RT/DL task), set its properties to the average
+	 * fair task as a best guess.
+	 *
+	 * Note that this is a hack to bypass the restriction of the
+	 * current bpf not trusting the pointer p->real_parent.
 	 */
-	for (i = 0; i < sizeof(*taskc) && can_loop; i++)
-		((char __arena *)taskc)[i] = 0;
+	parent = bpf_task_from_pid(p->real_parent->pid);
+	if (parent && (taskc_parent = get_task_ctx(parent))) {
```
daidavid, line 1721: Are there use cases where we may not want to inherit properties? i.e. Should we respect SCHED_RESET_ON_FORK?

multics69: SCHED_RESET_ON_FORK seems more about preventing the parent’s nice priority, scheduler class, and uclmap settings from being inherited. It seems less relevant to our change, since it concerns whether to inherit parents’ statistics, such as wake/wake frequencies and average runtime. However, there are applications that use SCHED_RESET_ON_FORK in creative ways. :-) For now, I think it is better to move on, and let’s revisit later when we have concrete cases that matter. I left comments in the code to recall SCHED_RESET_ON_FORK easily later.

PR #3281, opened 2026-02-02
scx_lavd: Optimize task migration for execve.

multics69:
This series optimizes task placement in the LAVD scheduler with aggressive load balancing during execve().

- Execve-Driven Migration: Triggers immediate migration for tasks calling execve(), as address space replacement makes current cache locality irrelevant.
- Code Refactoring: Modularizes neighbor-domain migration logic to support these new policies.


multics69:
@daidavid -- As we discussed offline, I dropped the fork part (preventing cross-domain migration during a fork) because some workloads create more worker threads than an LLC can handle. In this case, cross-domain migration will balance the load much more quickly.

daidavid (approved):

PR #3288, opened 2026-02-04
include: reduce arena size for compatibility with older kernels

etsal:
Bump the size of the arena map below 512MiB for compatibility with older kernels. Newer versions of libbpf place arena globals at the very end of the arena, but older kernel versions forbid map sizes above 512MiB. Since we don't use significant amount of arena memory, limit the maximum size to below this threshold.


etsal:
Yes, even HEAD is failing so it's no worse than what we already have.

PR #3294, opened 2026-02-07
scx_lavd: Standardize naming conventions for time, utilization, and load metrics

multics69:
This patch series standardizes the naming convention for time-, utilization-, and load-related variables in the scx_lavd scheduler. The primary goal is to eliminate ambiguity and prevent the accidental mixing of different time domains, such as raw wall-clock time and frequency-invariant time.

Rationale
--------------
The scheduler currently operates across multiple time domains: raw wall-clock time, time scaled by task weight, and time scaled by CPU capacity or frequency. Mixing these units (e.g., adding a frequency-invariant duration to a wall-clock timestamp) can lead to subtle accounting and scheduling errors.

Furthermore, terms previously used in the codebase, such as "scaled," were often ambiguous regarding what they were scaled by. This series introduces a strict 4-letter suffix convention to make the domain of any variable or function immediately apparent.

Naming Convention
-----------------------------
The following suffixes are applied throughout the series:

* _wall: Wall-clock time in nanoseconds.
* _invr: CPU capacity and frequency-invariant time.
* _wwgt: Weighted wall-clock time, scaled by the inverse of the task's weight.
* _iwgt: Weighted invariant time, scaled by both task weight and hardware capacity.

Series Overview
-----------------------
* [1/4] Introduces the suffix convention and renames core time-related variables in the task and CPU contexts.
* [2/4] Renames utility functions to reflect their source and destination domains (e.g., scale_cap_freq() becomes conv_wall_to_invr()).
* [3/4] Standardizes CPU utilization metrics, replacing ambiguous names like cpu_sutil with cpu_util_invr.
* [4/4] Renames load-related variables, specifically changing sc_load to load_invr to match the utilization naming scheme.

There are no functional changes intended in this series.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -19,16 +19,16 @@
 
 extern const volatile u8	mig_delta_pct;
 
-u64 __attribute__ ((noinline)) calc_mig_delta(u64 avg_sc_load, int nz_qlen)
+u64 __attribute__ ((noinline)) calc_mig_delta(u64 avg_load_invr, int nz_qlen)
 {
 	/*
 	 * Note that added "noinline" to make the verifier happy.
 	 */
 	if (nz_qlen >= sys_stat.nr_active_cpdoms)
-		return avg_sc_load >> LAVD_CPDOM_MIG_SHIFT_OL;
+		return avg_load_invr >> LAVD_CPDOM_MIG_SHIFT_OL;
 	if (nz_qlen == 0)
-		return avg_sc_load >> LAVD_CPDOM_MIG_SHIFT_UL;
-	return avg_sc_load >> LAVD_CPDOM_MIG_SHIFT;
+		return avg_load_invr >> LAVD_CPDOM_MIG_SHIFT_UL;
+	return avg_load_invr >> LAVD_CPDOM_MIG_SHIFT;
 }
 
 __weak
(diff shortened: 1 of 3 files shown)


multics69:
**This PR will be merged after the February release.**

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1736,8 +1739,8 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(lavd_init_task, struct task_struct *p,
 		taskc->last_running_clk = now; /* for avg_runtime */
```
bboymimi: I grep the avg_runtime, and it seems that avg_runtime still exists in many places in the comments. It may be a minor change. Could be better to correct the naming as well.

scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
bboymimi: How about sys_stat.{avg_util,avg_sc_util}? Are these two also considered to align with the naming?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Updated all. Thanks!

scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
multics69: I renamed them into `avg_util_wall` and `avg_util_invr` following the suffix convention.

scheds/rust/scx_lavd/src/stats.rs
bboymimi, line 223: Do you feel that the names—such as CPU_UTIL and CPU_SUTIL—of the column headers should also be changed? Maybe CPU_UTIL_WALL and CPU_UTIL_INVR. I feel the naming could be too long and optional. I'll leave it to your discretion. Overall, it looks good to me.

multics69: Thanks for the suggestion. I also thought about that. The userspace tools, such as htop, report CPU utilization based on wall-clock time, so I thought CPU_UTIL would be more natural for user-exposed information.  As not changing CPU_UTIL to CPU_UTIL_WALL, I thought let's leave CPU_SUTIL as it is. Hehe, this is a bit hand-wavy. Let's revisit this when more information is exposed.

multics69:
Thanks @bboymimi and @daidavid for the review.

PR #3298, opened 2026-02-07
scx_flatcg: use online CPUs instead of possible CPUs for nr_cpus

yunwei37:
On systems where cpu_possible != cpu_online (e.g., VMs with CPU hotplug support), libbpf_num_possible_cpus() returns a larger value than the actual online CPUs. This causes all tasks to be routed to FALLBACK_DSQ because the BPF check `p->nr_cpus_allowed != nr_cpus` always evaluates to true (e.g., 4 != 128).

Fix this by using sysconf(_SC_NPROCESSORS_ONLN) to get the online CPU count, which matches what normal processes have in nr_cpus_allowed.

Also fix fcg_read_stats() to use libbpf_num_possible_cpus() for reading per-CPU BPF maps, as these are sized by possible CPUs.

Tested on a system with:
  cpu_possible: 0-127 (128 CPUs)
  cpu_online: 0-3 (4 CPUs)

Before fix: global:649 local:18 (weight scheduling broken)
After fix:  global:151 local:38 (weight scheduling working, 0.20% deviation)

### DIFF
--- a/scheds/c/scx_flatcg.c
+++ b/scheds/c/scx_flatcg.c
@@ -102,7 +102,12 @@ static float read_cpu_util(__u64 *last_sum, __u64 *last_idle)
 
 static void fcg_read_stats(struct scx_flatcg *skel, __u64 *stats)
 {
-	__u64 cnts[FCG_NR_STATS][skel->rodata->nr_cpus];
+	/*
+	 * Per-CPU BPF maps are sized by libbpf_num_possible_cpus(),
+	 * not the number of online CPUs.
+	 */
+	int nr_possible_cpus = libbpf_num_possible_cpus();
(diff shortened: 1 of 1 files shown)


yunwei37:
The test framework: https://github.com/yunwei37/agentcgroup/tree/27580d5501e3f197ae436d8b5a889628f9c8b86c/scx_flatcg

etsal:
Please send the change to scx_flatcg to the sched_ext kernel mailing list instead, the C schedulers in this directory are being phased out: https://github.com/sched-ext/scx/pull/3195. The canonical version of the C schedulers is in linux/tools/sched_ext.

htejun:
We're dropping C scheds from this repo. Please post the patch against the kernel's `tools/sched_ext`.

PR #3300, opened 2026-02-08
scx_lavd: Revise load-balancing conditions

bboymimi:
LAVD's periodic load balancer runs unconditionally even when the system is lightly loaded, and balancing provides no benefit. This wastes CPU cycles and can cause unnecessary task migrations that hurt cache locality.

This series makes load balancing more selective by skipping it when the system is idle enough, and by simplifying the dispatch path at very low utilization where deadline ordering adds overhead without benefit. Task stealing at ops.dispatch() and task migration at ops.select_cpu() remain active for work conservation and handling bursty workloads.                                                                                                    

Two new tunable command line options are introduced:                                                                                                                                                                                     
  - `--lb-low-util-pct` (default 25): skip periodic LB below this utilization percentage
  - `--lb-local-dsq-util-pct` (default 10): bypass deadline scheduling and use FIFO local DSQ below this utilization percentage

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -226,6 +226,14 @@ const volatile u8	mig_delta_pct = 0;
 const volatile u64	lb_low_util_wall = 0;
 
 /*
+ * Bypass deadline-based scheduling and dispatch directly to local DSQ
+ * when average system utilization is below this threshold.
+ * The value is pre-scaled by userspace. 0 = disabled.
+ * Default: p2s(10) = 102.
+ */
+const volatile u64	lb_local_dsq_util_wall = 0;
+
+/*
  * Slice time for all tasks when pinned tasks are running on the CPU.
  * When this is set (non-zero), pinned tasks always use per-CPU DSQs and
  * the dispatch logic compares vtimes across DSQs.
@@ -563,6 +571,13 @@ static void update_stat_for_refill(struct task_struct *p,
 					   taskc->acc_runtime_wall);
 }
 
+static bool can_direct_dispatch(struct cpu_ctx *cpuc, bool is_cpu_idle)
+{
+	return (is_cpu_idle && !queued_on_cpu(cpuc)) ||
+	       (lb_local_dsq_util_wall > 0 &&
+		sys_stat.avg_util_wall < lb_local_dsq_util_wall);
+}
+
 s32 BPF_STRUCT_OPS(lavd_select_cpu, struct task_struct *p, s32 prev_cpu,
 		   u64 wake_flags)
 {
@@ -634,7 +649,7 @@ s32 BPF_STRUCT_OPS(lavd_select_cpu, struct task_struct *p, s32 prev_cpu,
 			goto out;
 		}
 
-		if (!queued_on_cpu(cpuc)) {
+		if (can_direct_dispatch(cpuc, true)) {
 			p->scx.dsq_vtime = calc_when_to_run(p, ictx.taskc);
 			p->scx.slice = LAVD_SLICE_MAX_NS_DFL;
 			scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, p->scx.slice, 0);
@@ -789,7 +804,7 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 	 * When pinned_slice_ns is enabled, pinned tasks always use per-CPU DSQ
 	 * to enable vtime comparison across DSQs during dispatch.
 	 */
-	if (is_idle && !queued_on_cpu(cpuc)) {
+	if (can_direct_dispatch(cpuc, is_idle)) {
 		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | cpu, p->scx.slice,
 				   enq_flags);
 	} else {
--- a/scheds/rust/scx_lavd/src/main.rs
+++ b/scheds/rust/scx_lavd/src/main.rs
@@ -143,6 +143,15 @@ struct Opts {
     #[clap(long = "lb-low-util-pct", default_value = "25", value_parser=Opts::lb_low_util_pct_range)]
     lb_low_util_pct: u8,
 
+    /// Low utilization threshold percentage (0-100) for bypassing deadline
+    /// scheduling. When set to a non-zero value, tasks are dispatched directly
+    /// to the local DSQ (FIFO) instead of using deadline-based ordering when
+    /// average system utilization is below this percentage.
+    /// Default is 10 (bypass deadline scheduling below 10% utilization).
+    /// Set to 0 to disable. Set to 100 to always bypass deadline scheduling.
+    #[clap(long = "lb-local-dsq-util-pct", default_value = "10", value_parser=Opts::lb_local_dsq_util_pct_range)]
+    lb_local_dsq_util_pct: u8,
+
     /// Slice duration in microseconds to use for all tasks when pinned tasks
     /// are running on a CPU. Must be between slice-min-us and slice-max-us.
     /// When this option is enabled, pinned tasks are always enqueued to per-CPU DSQs
@@ -383,6 +392,10 @@ impl Opts {
     }
 
     fn lb_low_util_pct_range(s: &str) -> Result<u8, String> {
+        number_range(s, 0, 100)
+    }
+
+    fn lb_local_dsq_util_pct_range(s: &str) -> Result<u8, String> {
         number_range(s, 0, 100)
     }
 }
@@ -658,6 +671,7 @@ impl<'a> Scheduler<'a> {
         rodata.preempt_shift = opts.preempt_shift;
         rodata.mig_delta_pct = opts.mig_delta_pct;
         rodata.lb_low_util_wall = ((opts.lb_low_util_pct as u64) << 10) / 100;
+        rodata.lb_local_dsq_util_wall = ((opts.lb_local_dsq_util_pct as u64) << 10) / 100;
         rodata.no_use_em = opts.no_use_em as u8;
         rodata.no_wake_sync = opts.no_wake_sync;
         rodata.no_slice_boost = opts.no_slice_boost;


scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -55,6 +56,14 @@ int plan_x_cpdom_migration(void)
 	 * throughput when the system is overloaded.
 	 */
 
+	/*
+	 * When system utilization is low, periodic load balancing across
+	 * LLC domains is unnecessary since there is plenty of idle capacity.
+	 */
+	if (lb_low_util_pct > 0 &&
+	    sys_stat.avg_util < ((u64)lb_low_util_pct << LAVD_SHIFT) / 100)
```
multics69: Please use the `p2s()` macro for `((u64)lb_low_util_pct << LAVD_SHIFT) / 100)`.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -783,7 +790,9 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 	 * When pinned_slice_ns is enabled, pinned tasks always use per-CPU DSQ
 	 * to enable vtime comparison across DSQs during dispatch.
 	 */
-	if (is_idle && !queued_on_cpu(cpuc)) {
+	if ((is_idle && !queued_on_cpu(cpuc)) ||
+	    (lb_local_dsq_util_pct > 0 &&
+	     sys_stat.avg_util < ((u64)lb_local_dsq_util_pct << LAVD_SHIFT) / 100)) {
```
multics69: Ditto. Please use `p2s()`.

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -145,6 +145,16 @@ struct Opts {
     #[clap(long = "lb-low-util-pct", default_value = "25", value_parser=Opts::lb_low_util_pct_range)]
     lb_low_util_pct: u8,
 
+    /// Low utilization threshold percentage (0-100) for bypassing deadline
+    /// scheduling. When set to a non-zero value, tasks are dispatched directly
+    /// to the local DSQ (FIFO) instead of using deadline-based ordering when
+    /// average system utilization is below this percentage. The threshold is
+    /// calculated as: avg_util < (lb-local-dsq-util-pct << 10) / 100.
```
multics69: Adding `avg_util < (lb-local-dsq-util-pct << 10) / 100` to the help message would be too implementation detail. Please keep the description high-level from a user's point of view.

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -136,6 +136,15 @@ struct Opts {
     #[clap(long = "mig-delta-pct", default_value = "0", value_parser=Opts::mig_delta_pct_range)]
     mig_delta_pct: u8,
 
+    /// Low utilization threshold percentage (0-100) for periodic load balancing.
+    /// When set to a non-zero value, periodic load balancing is skipped when
+    /// average system utilization is below this percentage. The threshold is
+    /// calculated as: avg_util < (lb-low-util-pct << 10) / 100.
```
multics69: `avg_util < (lb-low-util-pct << 10) / 100` is too low-level for the help output. Please keep it high-level for users.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -78,12 +78,9 @@ int plan_x_cpdom_migration(void)
 		}
 
 		/*
-		 * Use avg_util_sum when mig_delta_pct is set, otherwise use cur_util_sum.
+		 * Use avg_util_sum for stable load balancing decisions.
```
multics69: Please also update the description of the `mig-delta-pct` option:

```
    /// Migration delta threshold percentage (0-100). When set to a non-zero value,
    /// uses average utilization for threshold calculation instead of current
    /// utilization, and the threshold is calculated as: avg_load * (mig-delta-pct / 100).
```

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
multics69: > Please use the `p2s()` macro for `((u64)lb_low_util_pct << LAVD_SHIFT) / 100)`.

Instead of calculating the same computation again and again, can we pre-calculate the value when initializing the scheduler (on the rust side)?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: > Ditto. Please use `p2s()`.

Instead of calculating the same computation again and again, can we pre-calculate the value when initializing the scheduler (on the rust side)?

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -78,12 +78,9 @@ int plan_x_cpdom_migration(void)
...
 		 */
-		if (mig_delta_pct > 0)
-			util = (cpdomc->avg_util_sum << LAVD_SHIFT) / cpdomc->nr_active_cpus;
-		else
-			util = (cpdomc->cur_util_sum << LAVD_SHIFT) / cpdomc->nr_active_cpus;
+		util = (cpdomc->avg_util_sum << LAVD_SHIFT) / cpdomc->nr_active_cpus;
```
daidavid: When I was experimenting with this earlier, I recall that avg_util_sum might've converged a bit too slowly. I think we might need some better functions to accumulate/decay avg_util_sum in a more nuanced way. Do you have any benchmark results that could highlight the impact of this change?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -783,7 +790,9 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 	 * When pinned_slice_ns is enabled, pinned tasks always use per-CPU DSQ
 	 * to enable vtime comparison across DSQs during dispatch.
 	 */
-	if (is_idle && !queued_on_cpu(cpuc)) {
+	if ((is_idle && !queued_on_cpu(cpuc)) ||
```
daidavid: I recall that at low utilization, most of the tasks will get dispatched via direct_dispatch on select_cpu path anyways(It's been a while since I instrumented this, so my view of this might be outdated). I'm curious how often this actually gets exercised at low util.

bboymimi: Good observation. You're right that tasks get directly dispatched to local DSQ in select_cpu when an idle cpu with empty DSQ is found, which bypasses enqueue entirely. The main benefit of the lb_local_dsq_util_pct bypass is to avoid the overhead of vtime-based deadline ordering when it provides no scheduling advantage. At low utilization, maintaining deadline order via scx_bpf_dsq_insert_vtime() adds cost, such as vtime computation, ordered insertion, without improving scheduling efficiency. Dispatching FIFO via scx_bpf_dsq_insert() to the local dsq is cheaper and sufficient in this condition.

I think you're right that to add some instrumentation to measure how often the direct dispatch path triggers and how much benefit the strategy introduces. If the numbers show it's negligible in practice, we could drop the patch or adjust the default threshold. While I'll conduct the tracing for the impact, appreciate if you have suggestions which numbers of measurement you would expect to evaluate.

While I'm currently on the Lunar New Year vacation, time is limited for work. Sorry for any inconvenience.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Great point, David. The avg_util_sum calculated from avg_util using
 calc_asym_avg(), which applies different decay rates depending on the
 direction of change—increasing fast and decreasing slow (updated every 10ms):

 Load increasing 0% to 100% (decay=2):

`__calc_avg(new_val=target, old_val=current, 2)`

Step-by-step (target=1024, representing 100% in LAVD scale):
10ms:  512  (50.0%)
20ms:  896  (87.5%)
30ms:  992  (96.9%)
40ms:  1016 (99.2%)
50ms:  1022 (99.8%)

Load decreasing 100% to 0% (decay=3):

`__calc_avg(old_val=current, new_val=target, 3)`

Step-by-step:
10ms: 896  (87.5%)
20ms: 784  (76.6%)
30ms: 686  (67.0%)
50ms: 525  (51.3%)
100ms:  269  (26.3%)
150ms:  156  (15.2%)
200ms:   71  ( 6.9%)

 So upward convergence reaches ~97% of target within 30ms, while downward
 convergence takes ~50ms just to around halve. 

 I was thinking about the two informations that would be helpful:

 1. **Latency tolerance**: 
	Is it possible to share what's the acceptable reaction time or latency for load
	balancer decisions in the scenarios?  If utilization shifts, how many ms of lag
	is tolerable before the load balance state should reflect the new reality? It will impact
	the reaction of role, stealers or stealees, classification and result in the
	slow load balance decision.

 2. **Benchmarking methodology**:
	 when you observed the slow convergence earlier, what workload or test scenario
	 were you using to reproduce it? Or we just use DCPerf to simulate the
	 environment?

 Currently, I don't have the benchmark results yet. With the latency tolerance
 indicator,  we could, for example, use a symmetric decay=2 EWMA or more
 aggressive decay=1 , or design something more nuanced approach.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Yes, I will prepare another version to pre-calculate the value on the RUST side.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Thanks for the review. Will do.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
bboymimi: Will submit another revision.

scheds/rust/scx_lavd/src/main.rs
bboymimi: Acked. Will provide another user-friendly version.

scheds/rust/scx_lavd/src/main.rs
bboymimi: Will provide a user-friendly version.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
daidavid: Thanks for doing the detailed analysis Gavin! 

1. The tolerance for this is very dependent on the workload. I don't have an exact answer for you here.
2. It was for an internal work similar to Django bench, it would be interesting to tune this and see the effects of this on the benchmarks. I'd recommend turning forced stealing off for experimentation since that makes load balancing completely unpredictable.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
daidavid: No rush, I suspect this will be fairly negligible, but any optimization is nice. Overall, I appreciate you looking into this.

bboymimi: ## Patch 4 Analysis: Dispatch Counter Instrumentation (bypass deadline scheduling at low util)
### Methodology

- BPF counter patch: https://gavinguo.cc/scx/local_dsq_insert/local_dsq_counter.diff
Added per-condition atomic counters (`nr_local_idle`, `nr_local_low_util`, `nr_enq_vtime`) to `lavd_enqueue()` in `main.bpf.c` to track which branch each dispatch takes. Split the original combined `if` into separate branches:

- **idle fast path**: `is_idle && !queued_on_cpu(cpuc)` → `scx_bpf_dsq_insert` (existing)
- **low-util bypass**: `lb_local_dsq_util > 0 && avg_util_wall < lb_local_dsq_util` → `scx_bpf_dsq_insert` (Patch 4)
- **vtime (deadline)**: else → `scx_bpf_dsq_insert_vtime`

Counters read from `.bss` map via `bpftool` polling script.

### Instrumentation Scripts

- BPF counter patch: https://gavinguo.cc/scx/local_dsq_insert/local_dsq_counter.diff
- bpftrace script: https://gavinguo.cc/scx/local_dsq_insert/lavd_dsq_trace.bt
- BPF counter polling: https://gavinguo.cc/scx/local_dsq_insert/lavd_enq_stats.sh
- Flamegraph: https://gavinguo.cc/scx/local_dsq_insert/lavd_dsq.svg

### A/B Test Results

- https://gavinguo.cc/scx/local_dsq_insert/lavd_stat.sh
Workload: desktop idle (low utilization), 20 seconds per test.

**Test A — Patch 4 disabled** (`--lb-local-dsq-util-pct 0`, 21 samples):

| Metric        | Value |
| ------------- | ----- |
| Avg local/sec | 5,863 |
| Avg vtime/sec | 319   |
| Avg total/sec | 6,182 |
| Avg local %   | 94.8% |

**Test B — Patch 4 enabled** (`--lb-local-dsq-util-pct 25`, 20 samples):

| Metric        | Value |
| ------------- | ----- |
| Avg local/sec | 5,885 |
| Avg vtime/sec | 218   |
| Avg total/sec | 6,102 |
| Avg local %   | 96.4% |

### Per-Condition Dispatch Breakdown (with counters)

Running `--lb-local-dsq-util-pct 10` (default threshold, `p2s(10) = 102`) with instrumented build under mixed desktop workload:

```bash
sudo target/release/scx_lavd --performance --lb-local-dsq-util-pct 10 \
    --slice-min-us 3000 --slice-max-us 10000 --pinned-slice-us 3000 --stats 1
```

**Cumulative over full run**:

- BPF counter patch: https://gavinguo.cc/scx/local_dsq_insert/local_dsq_counter.diff
```
  idle fast path:   84,161   ( 19.0%)
  low-util bypass:   4,423   (  1.0%)  <- Patch 4
  vtime (deadline): 353,503  ( 80.0%)
  avg_util_wall: 29.4%
```

### Flamegraph Call Stack Analysis

I was also curious about the origin of the local insert and used bpftrace `kstack` tracing on `kfunc:scx_bpf_dsq_insert` to identify the origin of local DSQ dispatches and found that most of the local insert is from the select_cpu:

- Flamegraph: https://gavinguo.cc/scx/local_dsq_insert/lavd_dsq.svg
- **~96.6%** from `bpf__sched_ext_ops_select_cpu` — the idle-CPU fast path that finds an idle CPU and dispatches directly to local DSQ, bypassing `enqueue` entirely
- **~2.1%** from `bpf__sched_ext_ops_enqueue` — the non-idle path where Patch 4's bypass condition applies
### Conclusion

At ~29% average utilization, Patch 4's low-util bypass accounts for only **1.0%** of enqueue dispatches cumulatively. The vtime (deadline) path dominates at **80%** because `avg_util_wall` (29.4%) exceeds the `lb_local_dsq_util` threshold (`p2s(10)` ≈ 10%), so the bypass condition is rarely met. The idle fast path handles **19%**.

I think you are correct: at low utilization, `select_cpu` handles ~96.6% of all local DSQ dispatches before `enqueue` is even reached. Within `enqueue`, the idle fast path and vtime dominate, leaving Patch 4 with minimal impact (~1% of enqueue volume). On the other hand, I would also want to see if there is any performance gain on the production side.

daidavid: Thanks Gavin, I really liked the detailed analysis/breakdown. That being said, optimization at the low util is really not that meaningful for production workloads. I'm not going to block on this if you have a different usecase where this is beneficial. Overall LGTM.

bboymimi: Maybe wait a while to let @multics69 make the decision or provide more insights. Either we can drop or keep this one. I'm fine.

multics69: Sorry for the delayed review. I am slowly getting back into the swing of things after the packing and unpacking.

Definitely, patch 4 is not for major target workloads, either in fleet and AAA games in SteamDeck, where CPU utilization is fairly high. But it will be useful when running casual, not CPU-demanding games on beefy machines. Also, the cost is cheap (one more comparison). I'd like to keep it.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -789,7 +797,9 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
...
+	    (lb_local_dsq_util > 0 &&
+	     sys_stat.avg_util_wall < lb_local_dsq_util)) {
```
multics69: It makes sense to add this condition to the `ops.select_cpu()` path as a shortcut. If we use the same condition in two functions, I think it is worth adding a utility function expressing the intention more clearly. For example, it would be something like this:

```
bool can_direct_dispatch(struct cpu_ctx *cpuc, bool is_cpu_idle)
{
    return is_cpu_idle && queued_on_cpu(cpuc) &&
           lb_local_dsq_util > 0 && (sys_stat.avg_util_wall < lb_local_dsq_util);
}
```

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -225,6 +225,14 @@ const volatile u8	mig_delta_pct = 0;
  */
 const volatile u64	lb_low_util = 0;
```
multics69: Please rename `lb_low_util` to `lb_low_util_wall` to follow the suffix convention.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -225,6 +225,14 @@ const volatile u8	mig_delta_pct = 0;
...
 
+/*
+ * Bypass deadline-based scheduling and dispatch directly to local DSQ
+ * when average system utilization is below this threshold.
+ * The value is pre-scaled by userspace. 0 = disabled.
+ * Default: p2s(10) = 102.
+ */
+const volatile u64	lb_local_dsq_util = 0;
```
multics69: Please rename `lb_local_dsq_util` to `lb_local_dsq_util_wall` to follow the suffix convention.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
multics69: Thanks @daidavid for the review. Since there is no regression, let's merge this PR. Anyway, we need to revisit the load balancing metric when addressing the aggresiveness of load balancing.

bboymimi: Great point, David. The per-CPU avg_util that feeds into avg_util_sum uses
calc_asym_avg(), which applies different decay rates depending on the
direction of change. Here are the convergence profiles for both directions
(updated every 10ms):

**Load increasing 0% → 100% (decay=2):**

```
__calc_avg(new_val=target, old_val=current, 2):
  normal case:  result = target - (target >> 2) + (current >> 2)
                       = 0.75 * target + 0.25 * current
  small-value:  result = 0.5 * target + 0.5 * current  (when current < 4)

Step-by-step (target=1024, representing 100% in LAVD scale):
  10ms:  current=0 < thr(4) → 0.5*1024 + 0.5*0     = 512  (50.0%)
  20ms:  0.75*1024 + 0.25*512                        = 896  (87.5%)
  30ms:  0.75*1024 + 0.25*896                        = 992  (96.9%)
  40ms:  0.75*1024 + 0.25*992                        = 1016 (99.2%)
  50ms:  0.75*1024 + 0.25*1016                       = 1022 (99.8%)
```

**Load decreasing 100% → 0% (decay=3):**

```
__calc_avg(old_val=current, new_val=target, 3):
  normal case:  result = current - (current >> 3) + (target >> 3)
                       = 0.875 * current + 0.125 * target
  with target=0: result = 0.875 * current

General formula: value(n) = 0.875^n * 1024

Step-by-step:
  10ms:   0.875^1  * 1024 = 0.875  * 1024 = 896  (87.5%)
  20ms:   0.875^2  * 1024 = 0.766  * 1024 = 784  (76.6%)
  30ms:   0.875^3  * 1024 = 0.670  * 1024 = 686  (67.0%)
  50ms:   0.875^5  * 1024 = 0.513  * 1024 = 525  (51.3%)
  100ms:  0.875^10 * 1024 = 0.263  * 1024 = 269  (26.3%)
  150ms:  0.875^15 * 1024 = 0.152  * 1024 = 156  (15.2%)
  200ms:  0.875^20 * 1024 = 0.069  * 1024 =  71  ( 6.9%)
```

So upward convergence reaches ~97% of target within 30ms, while downward
convergence takes ~100ms just to halve. The asymmetry is intentional in the
original design (fast reaction to load spikes, slow decay to prevent
oscillation), but it may not be the right trade-off for LB decisions.

To design a better accumulation/decay function for this use case, it would
help to know two things from your side:

1. **Latency tolerance** — what's the acceptable reaction time for LB
   decisions in the scenarios you care about? For example, if utilization
   shifts, how many ms of lag is tolerable before the LB state should
   reflect the new reality? That would help us pick the right decay
   parameter (or design a multi-rate approach).

2. **Benchmarking methodology** — when you observed the slow convergence
   earlier, what workload or test scenario were you using to reproduce it?
   If you could share the methodology (or a reproducer), I can use the
   same setup to validate improvements and produce comparable numbers.

With those two data points we could, for example, use a symmetric decay=2
EWMA (~30ms to 97% convergence in both directions), or design something
more nuanced like a dual-rate EWMA that adapts based on the magnitude of
change. Happy to prototype either approach once we align on the target
behavior.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
bboymimi: Good observation — you're right that at low utilization, most tasks get
directly dispatched in select_cpu when an idle CPU with empty DSQs is found,
which bypasses enqueue entirely per the sched_ext framework.

The main benefit of the lb_local_dsq_util_pct bypass is avoiding the overhead
of vtime-based deadline ordering when it provides no scheduling advantage.
At low utilization, tasks rarely compete for CPUs, so maintaining deadline
order via scx_bpf_dsq_insert_vtime() adds cost (vtime computation, ordered
insertion) without improving scheduling quality. Dispatching FIFO via
scx_bpf_dsq_insert() to the local DSQ is cheaper and sufficient when
contention is negligible.

For the tasks that do reach enqueue at low utilization, this bypass catches
several specific cases:

1. **select_cpu found an idle CPU but its DSQs had queued tasks** — the
   `is_idle && queued_on_cpu(cpuc)` case. The task still goes through enqueue
   and would normally get vtime-ordered insertion, which is unnecessary at
   low utilization.

2. **Tasks that enter enqueue without going through select_cpu** — this
   includes re-enqueued tasks (`SCX_ENQ_REENQ`, e.g., after yielding or
   tick-based re-queuing), tasks preempted by higher-priority RT/DL classes
   (`SCX_ENQ_PREEMPT`), and `SCX_ENQ_LAST`. In all these cases,
   `__COMPAT_is_enq_cpu_selected()` returns false, so enqueue runs its own
   `pick_idle_cpu()` but the task still benefits from skipping vtime ordering.

3. **Micro-burst wakeups where idle CPUs' DSQs are transiently non-empty** —
   e.g., a timer expiry waking a batch of workers or a network packet burst
   triggering multiple task wakeups in quick succession. Average utilization
   stays low, but the burst fills DSQs temporarily faster than CPUs drain
   them. These tasks hit enqueue with queued DSQs and would otherwise pay
   the vtime insertion cost for no benefit.

That said, I think you're right that at truly steady low utilization, the
select_cpu direct dispatch path dominates and this bypass fires infrequently.
I'll add some instrumentation to measure how often each dispatch path fires
at different utilization levels to validate. If the numbers show it's
negligible in practice, we could simplify or raise the default threshold.
Appreciate the sanity check.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
bboymimi: Thanks for the suggestion on can_direct_dispatch! Before implementing the function, I have a question about the function. The statement restricts the condition more narrowly to `is_cpu_idle && queued_on_cpu(cpu) &&`, which is different from the original logic, and this makes the condition more difficult to trigger. Could you elaborate on the idea behind the logic?

multics69: Haha, that's a typo. It should be something like this:

```
bool can_direct_dispatch(struct cpu_ctx *cpuc, bool is_cpu_idle)
{
    return (is_cpu_idle && !queued_on_cpu(cpuc)) ||
           (lb_local_dsq_util > 0 && (sys_stat.avg_util_wall < lb_local_dsq_util));
}
```

PR #3302, opened 2026-02-09
scx_utils: Add is_scx_running for sysfs check

EricccTaiwan:
Print an info log to confirm the scheduler execution has begun, making it more intuitive for users to verify the status.

### DIFF
--- a/rust/scx_utils/src/build_id.rs
+++ b/rust/scx_utils/src/build_id.rs
@@ -4,6 +4,8 @@
 // GNU General Public License version 2.
 
 use std::fmt::Write;
+use std::fs;
+use std::path::Path;
 
 lazy_static::lazy_static! {
     static ref GIT_VERSION: String = {
@@ -67,8 +69,31 @@ lazy_static::lazy_static! {
     pub static ref SCX_FULL_VERSION: String = full_version(*SCX_CARGO_VERSION);
 }
 
+fn match_name(cur_ops: &str, expected: &str) -> bool {
+    let base = expected.strip_prefix("scx_").unwrap_or(expected);
+    cur_ops.contains(base)
+}
+
+pub fn is_scx_running(expected: &str) -> bool {
+    let ops_path = "/sys/kernel/sched_ext/root/ops";
+
+    if !Path::new(ops_path).exists() {
+        return false;
+    }
+
+    match fs::read_to_string(ops_path) {
+        Ok(ret) => match_name(ret.trim(), expected),
+        Err(e) => {
+            log::debug!("Failed to read {}: {}", ops_path, e);
+            false
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
+    use super::*;
+
     #[test]
     fn test_cargo_ver() {
         //assert_eq!(super::*SCX_CARGO_VERSION, 1);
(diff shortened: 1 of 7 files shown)


EricccTaiwan:
> It might be nicer to have something in scx_utils that reads /sys/kernel/sched_ext/root/ops to actively check whether the scheduler is running. The end result would probably be similar, but this would at least give us a truly generic check.

Thanks for pointing this out, this makes a lot more sense! I agree and have updated the PR to add the `is_scx_running` check in `scx_utils`.
Let me know how it looks! :+1:

htejun:
I'm not sure what this actually adds. Is this solving any actual problem?

EricccTaiwan:
> I'm not sure what this actually adds. Is this solving any actual problem?

I mainly wanted to align the behavior with other schedulers that provide visual confirmation upon startup. A generic "is running" message reassures the user that the scheduler is active and hasn't stalled.

I implemented the `ops` check to ensure the message is backed by the actual system state. But I understand if you prefer to keep the output minimal. If you think this doesn't add enough value, please let me know and I'll close it.

arighi:
@EricccTaiwan thanks for adding the check in sysfs. However, as also pointed by Tejun, I'm still unsure if this would provide any benefit in practice, if `Scheduler::init()` doesn't trigger any error then the scheduler is always started, we don't need to print it. I would at least print the message only in verbose mode, in regular mode there's really no point printing that the scheduler is running.

rust/scx_utils/src/build_id.rs
```diff
@@ -67,8 +69,31 @@ lazy_static::lazy_static! {
     pub static ref SCX_FULL_VERSION: String = full_version(*SCX_CARGO_VERSION);
 }
 
+fn match_name(cur_ops: &str, expected: &str) -> bool {
```
arighi, line 72: Maybe `rust/scx_utils/src/misc.rs` is a better place for these functions?

scheds/rust/scx_beerland/src/main.rs
```diff
@@ -588,6 +588,14 @@ fn main() -> Result<()> {
     let mut open_object = MaybeUninit::uninit();
     loop {
         let mut sched = Scheduler::init(&opts, &mut open_object)?;
+        if build_id::is_scx_running(SCHEDULER_NAME) {
```
arighi, line 591: I'm wondering if we could hit false positives here: if we're checking for the sysfs file too soon and the file is not populated in sysfs yet, we may trigger an error, while the scheduler is started instead.

But, again, do we really need to deal with this extra complexity?

Don't get me wrong, it might be useful to have the `is_scx_running()` as a generic lib function (maybe in misc.rs as I mentioned), but I wouldn't use it in all the schedulers when they're started.

EricccTaiwan: I completely missed the "false positive" timing issue. Given that we'd have to add retry logic to handle it, dealing with this definitely adds unnecessary complexity. I'm wondering if it's even worth the effort just for a log message.

I'll go ahead and close this PR to keep things clean. But if you think the `is_scx_running` function is still worth keeping as a standalone utility, I'm open to giving it another try.
I just don't know of a good way to solve the sysfs delay without making the code too complex. If you have any ideas, let me know!

PR #3108, opened 2025-11-30
scx_utils: optimise gpu NUMA node id info retrieval.

devnexen:
using the wrapper but still falling back to the slow(er) path in case.

PR #3312, opened 2026-02-12
scx_utils: use proper NVML API for GPU NUMA node lookup

devnexen:
using the wrapper but still falling back to the slow(er) path in case. new attempt after #3108 only this time a new nvml-wrapper version had been released.

### DIFF
--- a/rust/scx_utils/src/gpu.rs
+++ b/rust/scx_utils/src/gpu.rs
@@ -118,12 +118,18 @@ pub fn create_gpus() -> BTreeMap<usize, Vec<Gpu>> {
                 .performance_state()
                 .unwrap_or(PerformanceState::Unknown);
 
-            // The NVML library doesn't return a PCIe bus ID compatible with sysfs. It includes
-            // uppercase bus ID values and an extra four leading 0s.
-            let bus_id = pci_info.bus_id.to_lowercase();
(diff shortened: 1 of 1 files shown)


devnexen:
cc @arighi

htejun (changes requested):
Please prove that this optimization is meaningful.

devnexen:
To be honest, the win when I tried it was not that much performance wise mostly saving the FS operation time, I m digging the benchmark (hopefully still have it somewhere); but to summarize. The primary motivation is correctness rather than raw speed. The old code reverse-engineers a sysfs path from NVML's bus ID format by stripping a "0000" prefix — this is fragile and the FIXME comment in the
  code acknowledges it may not always be 4 leading zeros.

  numa_node_id() uses the proper NVML API (nvmlDeviceGetNumaNodeId) directly. The fallback to the sysfs path is kept for older NVML versions that don't support it.

devnexen:
here it is, [as gist](https://gist.github.com/devnexen/91bc6818b999fcf963b420055c055e9e). Locally it is only ~1.1 faster on average (I have only one nvidia gpu tough). Anyhow, using the now available direct wrapper NVML call alone, is a valid reason IMHO.

htejun:

PR #3320, opened 2026-02-13
scx_cosmos: Add CPU capacity awareness

arighi:
On task wakeup, when using the default idle CPU selection policy, if the waker's CPU has higher capacity than the wakee's CPU, attempt to place the wakee on a CPU closer in capacity to the waker.

NOTE: This logic is already used in scx_bpfland and scx_beerland with positive results, so extend the same approach to scx_cosmos as well.

cc: @balbirs-nv


multics69 (approved):
LGTM

PR #3324, opened 2026-02-13
Cargo: limit the number of compiled workspace elements

sirlucjan:
It turns out that our workspace is large. It won't get any smaller, and it will only get bigger. By default, Rust/Cargo builds a lot of objects at once. I was thinking of putting a limit in config.toml.


multics69 (approved):
LGTM

PR #3331, opened 2026-02-16
scx_cake: add to integration-tests

sirlucjan:
Add scx_cake to integration tests


sirlucjan:
Merge after scx_cake fixes.

PR #3336, opened 2026-02-17
[layered]: fix UsedGpuPid + IsGroupLeader matchers

xerothermic:
PR #3132 fixed a kprobe race condition by deferring kprobe attachment to after layered_init(). This exposed a hidden bug in save_gpu_tgid_pid(): it was refreshing p->parent instead of p->group_leader.

Before the refactor, kprobes attached early, so the GPU maps were already populated by the time `layered_init()` enumerated tasks. The init-time scan would correctly match GPU tasks to their layers, masking the incorrect parent refresh.

After the refactor, kprobes attach after init, so the GPU maps are empty during the initial task scan. Layer matching now depends entirely on the kprobe refresh_layer path — which was refreshing the wrong task. For any multithreaded process, p->parent (the forking process, e.g. a shell) differs from p->group_leader (the main thread of the thread group), so UsedGpuPid + IsGroupLeader rules would never match.

This fix changes the refresh target from p->parent to p->group_leader.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -739,7 +739,7 @@ int save_gpu_tgid_pid(void) {
 	if (!enable_gpu_support)
 		return 0;
 	struct task_struct *p = NULL;
-	struct task_ctx *taskc, *parent;
+	struct task_ctx *taskc, *leader;
 	u64 pid_tgid;
 	u32 pid, tid;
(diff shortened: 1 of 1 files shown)


etsal:
Interesting, so the behavior was sane but wrong - we were essentially handling multiprocess applications instead of multithreaded ones (which is what we actually wanted), am I getting this right?

xerothermic:
Yeah, most of the training applications have multithreaded child that's pulling the Nvidia driver.

PR #3338, opened 2026-02-18
Made netdev-irq-balance works

xerothermic:
Added a script generation to log original IRQ affinities, and restore it upon exit. If restore failed, leave the initial IRQ affinity file in `/run/scx_layered_netdev_irq_restore.sh` so user has a chance to manually restore it.

While testing the irq rebalance, I noticed some hardware interrupts are still overlapped with cores that run in grouped / confined layers. Turned out, `while_take` will give up after first non-match (https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while) Fixed the issue by using .find(...)

### DIFF
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -2826,8 +2826,7 @@ impl<'a> Scheduler<'a> {
                 .topo
                 .nodes
                 .values()
-                .take_while(|n| n.id == netdev.node())
-                .next()
+                .find(|n| n.id == netdev.node())
                 .ok_or_else(|| anyhow!("Failed to get netdev node"))?;
             let node_cpus = node.span.clone();
             for (irq, irqmask) in netdev.irqs.iter_mut() {
@@ -3534,7 +3533,9 @@ impl<'a> Scheduler<'a> {
                 empty_layer_ids.len() as u32;
         }
 
-        let _ = self.update_netdev_cpumasks();
+        if let Err(e) = self.update_netdev_cpumasks() {
+            warn!("Failed to update netdev IRQ cpumasks: {:#}", e);
+        }
         Ok(())
     }
 


rust/scx_utils/src/netdev.rs
```diff
@@ -40,8 +42,8 @@ impl NetDev {
 
     pub fn apply_cpumasks(&self) -> Result<()> {
         for (irq, cpumask) in self.irqs.iter() {
-            let irq_path = format!("/proc/irq/{irq}/smp_affinity");
-            fs::write(irq_path, format!("{cpumask:#x}"))?
+            let irq_path = format!("/proc/irq/{irq}/smp_affinity_list");
+            fs::write(irq_path, cpumask.to_cpulist())?
```
hodgesds: shouldn't the cpumask work? I guess this works as well, but doesn't seem necessary to write out a list

scheds/rust/scx_layered/src/main.rs
```diff
@@ -2408,9 +2357,26 @@ impl<'a> Scheduler<'a> {
...
         let netdevs = if opts.netdev_irq_balance {
             warn!(
-                "Experimental netdev IRQ balancing enabled. Reset IRQ masks of network devices after use!!!"
+                "Experimental netdev IRQ balancing enabled. \
+                 Original IRQ affinity will be restored on exit. \
+                 Emergency restore script: {NETDEV_IRQ_RESTORE_SCRIPT}"
             );
-            read_netdevs()?
+            let devs = read_netdevs()?;
+            write_netdev_irq_restore_script(&devs, Path::new(NETDEV_IRQ_RESTORE_SCRIPT))?;
```
hodgesds: I would think this is the job of whatever is running the scheduler with that option enabled, it's a little weird to create a script when running the scheduler.

scheds/rust/scx_layered/src/main.rs
```diff
@@ -3950,6 +3910,29 @@ impl Drop for Scheduler<'_> {
     fn drop(&mut self) {
         info!("Unregister {SCHEDULER_NAME} scheduler");
 
+        if !self.netdevs.is_empty() {
+            match Command::new("sh").arg(NETDEV_IRQ_RESTORE_SCRIPT).status() {
```
hodgesds: what happens if someone adds `rm -rf --no-preserve-root /` to the end of the script?

xerothermic:
Hi @hodgesds , thanks for the review feedbacks. @etsal shared similar feedbacks and I realized it's not a good idea to write out a script to restore the IRQ. I've refactored to restore the initial IRQ affinity using the stored states within scheduler. Please take another look. Thanks!

rust/scx_utils/src/netdev.rs
xerothermic: initially, this was done to make the script more readable, but now we're dropping the script, so I restored and work with smp_affinity.

scheds/rust/scx_layered/src/main.rs
xerothermic: yeah, initially I worried we can't recall the system state if the scheduler crashes, but the likelihood of rust scheduler crash isn't high, so the latest change drop the script approach.

scheds/rust/scx_layered/src/main.rs
xerothermic: Good point. Removed the script now.

PR #3341, opened 2026-02-18
scx_mitosis: add RAII guard framework and replace manual cleanup

dschatzberg:
Add cleanup.bpf.h with kernel-style RAII macros (DEFINE_FREE, __free, no_free_ptr, DEFINE_CLASS, DEFINE_GUARD, DEFINE_LOCK_GUARD_0, guard, scoped_guard) modeled on the kernel's include/linux/cleanup.h.

Define resource types for cgroup, bpf_cpumask, idle_cpumask, rcu, and spin_lock guards. Replace all manual goto-based cleanup and the three ad-hoc guard macros (RCU_READ_GUARD, CPUMASK_GUARD, DECLARE_CPUMASK_ENTRY) with the unified framework.

This eliminates ~21 manual resource release sites across 8 functions, removes 3 goto labels from update_timer_cb, and removes ~45 lines of ad-hoc guard infrastructure from mitosis.bpf.h.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/cleanup.bpf.h
+++ b/scheds/rust/scx_mitosis/src/bpf/cleanup.bpf.h
@@ -1,0 +1,132 @@
+/* Copyright (c) Meta Platforms, Inc. and affiliates. */
+/*
+ * This software may be used and distributed according to the terms of the
+ * GNU General Public License version 2.
+ *
+ * RAII guard framework — subset of the kernel's cleanup.h for BPF.
+ *
+ * Naming and structure follow the kernel's include/linux/cleanup.h.
+ */
+
+#pragma once
+
+/* DEFINE_FREE — scoped cleanup for owned resources (refcounts, allocations) */
+#define DEFINE_FREE(_name, _type, _free)           \
+	static inline void __free_##_name(void *p) \
+	{                                          \
+		_type _T = *(_type *)p;            \
+		_free;                             \
+	}
+
+#define __free(_name) __attribute__((__cleanup__(__free_##_name)))
+
+/*
+ * Wrapper that enforces __must_check semantics so callers cannot
+ * accidentally discard the returned pointer and leak the resource.
+ */
+static inline __attribute__((__warn_unused_result__)) const volatile void *
+__must_check_fn(const volatile void *val)
+{
+	return val;
+}
+
+/* Like a non-atomic xchg(var, NULL), returns the old value. */
+#define __get_and_null(p)                  \
+	({                                 \
+		typeof(p) *__ptr = &(p);   \
+		typeof(p)  __val = *__ptr; \
+		*__ptr		 = NULL;   \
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_mitosis/src/bpf/cleanup.bpf.h
```diff
@@ -0,0 +1,126 @@
+/* Copyright (c) Meta Platforms, Inc. and affiliates. */
+/*
+ * This software may be used and distributed according to the terms of the
+ * GNU General Public License version 2.
+ *
+ * RAII guard framework — subset of the kernel's cleanup.h for BPF.
+ *
+ * Naming and structure follow the kernel's include/linux/cleanup.h.
```
tommy-u, line 8: Could add the version for future code differs

scheds/rust/scx_mitosis/src/bpf/cleanup.bpf.h
```diff
@@ -0,0 +1,126 @@
...
+#define __free(_name) __attribute__((__cleanup__(__free_##_name)))
+
+/* Transfer ownership out, preventing cleanup */
+#define no_free_ptr(p)                          \
+	({                                      \
+		typeof(p) __nfp = (p);          \
+		(p)		= (typeof(p))0; \
+		__nfp;                          \
+	})
+
```
tommy-u: I was talking to claude about the decision to simplify this relative to the kernel version https://elixir.bootlin.com/linux/v6.18.6/source/include/linux/cleanup.h#L227

It seems like all you get for the more complicated kernel version is: 
1) compile time erroring on incorrect usage of return_ptr on an integer type , because `__get_and_null(p, NULL)`
2) a warning when you use `no_free_ptr(p); ` as a complete statement w/o using the result of it.

We decided that 1) is moot because BPF doesn't do integer-type based ownership (like userspace would use FDs where 0 is valid), that sound right?
2) You always pass the ptr to the map when you exchange it. we can just be careful

tommy-u (approved):
This is awesome, 20 gotos highlighted in red.

I went through all of this, no issues I could find. 

You're killing our job security by making this easier to read.

scheds/rust/scx_mitosis/src/bpf/cleanup.bpf.h
dschatzberg: I don't think __get_and_null is materially different from what I've done here. I guess we potentially evaluate the statement `p` multiple times so it seems worth it to avoid doing that.

The warning of not using the result was just laziness but seems worthwhile to do. I'll fix both of those

scheds/rust/scx_mitosis/src/bpf/cleanup.bpf.h
dschatzberg: I decided not to - people can pretty easily check when the file was created and figure it out from there

PR #3345, opened 2026-02-19
flash: Drop ops.cpu_release()

EricccTaiwan:
Follow up #3343.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -1772,16 +1772,6 @@ void BPF_STRUCT_OPS(flash_quiescent, struct task_struct *p, u64 deq_flags)
 	}
 }
 
-void BPF_STRUCT_OPS(flash_cpu_release, s32 cpu, struct scx_cpu_release_args *args)
-{
-	/*
-	 * When a CPU is taken by a higher priority scheduler class,
-	 * re-enqueue all the tasks that are waiting in the local DSQ, so
-	 * that we can give them a chance to run on another CPU.
-	 */
-	scx_bpf_reenqueue_local();
-}
-
 void BPF_STRUCT_OPS(flash_set_cpumask, struct task_struct *p,
 		    const struct cpumask *cpumask)
 {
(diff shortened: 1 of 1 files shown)


sirlucjan (approved):

arighi:
Ok about scx_flash.

For scx_layered, can you create a separate PR so that the scx_layered maintainers can better review and decide about this change? Thanks!

etsal:
If we remove cpu_release (which we should) don't we have to also attach a kprobe to sched_switch() so that we can call scx_bpf_reenqueue_local() from there instead? Without it we open ourselves to starvation from higher-priority tasks.

arighi:
> If we remove cpu_release (which we should) don't we have to also attach a kprobe to sched_switch() so that we can call scx_bpf_reenqueue_local() from there instead? Without it we open ourselves to starvation from higher-priority tasks.

Not necessarily, starting with the v7.0 kernel, with the ext dl_server, the impact of RT tasks is less critical, schedulers may decide to keep tasks on the same local DSQ without re-enqueuing, since they'll still have a 5% CPU bandwidth assigned.

EricccTaiwan:
> Ok about scx_flash.
> 
> For scx_layered, can you create a separate PR so that the scx_layered maintainers can better review and decide about this change? Thanks!

Got it. I'll update this PR to focus only on `scx_flash`.

htejun:
> > If we remove cpu_release (which we should) don't we have to also attach a kprobe to sched_switch() so that we can call scx_bpf_reenqueue_local() from there instead? Without it we open ourselves to starvation from higher-priority tasks.
> 
> Not necessarily, starting with the v7.0 kernel, with the ext dl_server, the impact of RT tasks is less critical, schedulers may decide to keep tasks on the same local DSQ without re-enqueuing, since they'll still have a 5% CPU bandwidth assigned.

Hmmm... so while it won't stall, it can still have latency impacts, possibly significant depending on what kind of RT threads you have on the system.

arighi:
> > > If we remove cpu_release (which we should) don't we have to also attach a kprobe to sched_switch() so that we can call scx_bpf_reenqueue_local() from there instead? Without it we open ourselves to starvation from higher-priority tasks.
> > 
> > 
> > Not necessarily, starting with the v7.0 kernel, with the ext dl_server, the impact of RT tasks is less critical, schedulers may decide to keep tasks on the same local DSQ without re-enqueuing, since they'll still have a 5% CPU bandwidth assigned.
> 
> Hmmm... so while it won't stall, it can still have latency impacts, possibly significant depending on what kind of RT threads you have on the system.

That's true and it can be addressed in a separate PR <- @EricccTaiwan if you have time to look at this go ahead. thanks!

PR #3346, opened 2026-02-19
scx: Include task local data library

ameryhung:
Introduce task local data (TLD) to be used by LAVD and layered.

Task local data is developed to make hinting faster and simpler to use. It defines an abstract storage type on top of task local storage, and provide simple APIs to access it. By using special field UPTR, the scheduler can access the user space thread-specific hint directly. In addition, the API hides the layout from both user space and BPF programs with a key value API so that we can remove the central definition of the map value.

The task local data BPF library is copied from [selftests/bpf/progs](https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/progs/task_local_data.bpf.h) in the kernel tree. There is also user space API implemented in [selftests/bpf/prog_tests/task_local_data.h](https://github.com/torvalds/linux/blob/master/tools/testing/selftests/bpf/prog_tests/task_local_data.h). To pass the hint from the user space, first define a global key using the following macro:

  TLD_DEFINE_KEY(task_hint_key, "task_hint_priority", sizeof(u64));

Then, to get a pointer to the hint specific to the thread:

  u64 *task_hint_p = tld_get_data(tld_data_map_fd, task_hint_key);

The pointer will remain valid until the thread exit. For an basic example of TLD can be used, please refer to https://github.com/torvalds/linux/commit/120f1a950e495d9751fdb5c8b7852d94546dcd03

### DIFF
--- a/scheds/include/scx/task_local_data.bpf.h
+++ b/scheds/include/scx/task_local_data.bpf.h
@@ -1,0 +1,237 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#ifndef __TASK_LOCAL_DATA_BPF_H
+#define __TASK_LOCAL_DATA_BPF_H
+
+/*
+ * Task local data is a library that facilitates sharing per-task data
+ * between user space and bpf programs.
+ *
+ *
+ * USAGE
+ *
+ * A TLD, an entry of data in task local data, first needs to be created by the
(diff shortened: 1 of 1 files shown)


htejun:
Can you update the comment to point to the userspace part so that readers can get the whole picture?

ameryhung:
Links to the kernel tree and an example added

PR #3349, opened 2026-02-20
scx_cosmos: Introduce "sticky" and dynamic PMU thresholds

arighi:
This series extends the scx_cosmos scheduler's PMU-based scheduling with a second perf counter, a dedicated "sticky" event with its own threshold, and optional dynamic adjustment of both thresholds at runtime.

scx_cosmos already had a migration event (-e / -E): when a task's event rate exceeds the threshold, it is moved to the least event-busy CPU. With the "sticky" event it is possible to specify a second independent event used only to make tasks sticky on their current CPU when its threshold is exceeded.

This allows defining rules like "migrate tasks that exceeds 100 iTLB misses/sec, but keep tasks on their CPU if they're cache hit ratio exceeds 10.000 hits/sec" or similar.

Moreover, when the user doesn't specify any threshold for the PMU events, the scheduler will constantly try to find an optimal threshold dynamically at runtime to keep the migration rate in a reasonable range (between 10-100 migrations/sec).

cc: @balbirs-nv

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -110,7 +110,7 @@ const volatile u64 perf_config;
 /*
  * Performance counter threshold to classify a task as event heavy.
  */
-const volatile u64 perf_threshold;
+volatile u64 perf_threshold;
 
 /*
  * Enable deferred wakeup.
@@ -126,7 +126,7 @@ const volatile u64 perf_sticky;
 /*
  * Threshold for sticky event; task is kept on same CPU when exceeded.
  */
-const volatile u64 perf_sticky_threshold;
+volatile u64 perf_sticky_threshold;
 
 /*
  * Enable tick-based preemption enforcement.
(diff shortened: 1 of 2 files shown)


arighi:
@htejun @etsal @likewhatevs can you double check https://github.com/sched-ext/scx/pull/3349/changes/74d0337f854f988c53e3901b2151895c23aaab1b to make sure scx_layered is still happy with it? Thanks!

arighi:
CI is failing, I think we're hitting a kernel bug in the BPF verifier:
```
     verifier bug: REG INVARIANTS VIOLATION (false_reg1): range bounds violation
        u64=[0x0, 0x1] s64=[0x0, 0x1] u32=[0x3, 0x1] s32=[0x0, 0x1] var_off=(0x0, 0x1)
      WARNING: kernel/bpf/verifier.c:2742 at reg_bounds_sanity_check
      Call Trace:
        reg_bounds_sanity_check+0xbc/0x1e0
        reg_set_min_max+0x1a2/0x1f0
        check_cond_jmp_op+0x5d2/0x1980
        do_check_common+0x2b0f/0x3410
        do_check_subprogs+0xcd/0x180
        bpf_check+0x33fe/0x3850
        bpf_prog_load+0x7d7/0xee0
        __sys_bpf+0xea2/0x2e30
```

I'll investigate.

arighi:
Potential fix: https://lore.kernel.org/all/20260220190736.230329-1-arighi@nvidia.com/

arighi:
@etsal what do you think about d59114bf0?

etsal:
> @etsal what do you think about [d59114b](https://github.com/sched-ext/scx/commit/d59114bf08f6f363b0e87d64d3efc2f12f5a74be)?

I think it's the most practical solution for now. Let's go with that because I am not sure what the status is in terms of bug hunting for the other patch.

etsal:
Another note: Can we skip the switch-based macros and try to find another way? I think they are relatively suspect in terms of causing bugs. Can we use modulo arithmetic or bitmasks? Especially bitmasks are really well exercised as a way to enforce bounds.

arighi:
@etsal the verifier is really annoying, anything I try triggers either `R7 32-bit pointer arithmetic prohibited` or `R7 pointer arithmetic with <<= operator prohibited`. The switch/case seems to be the only way to go unfortunately...

arighi:
> @etsal the verifier is really annoying, anything I try triggers either `R7 32-bit pointer arithmetic prohibited` or `R7 pointer arithmetic with <<= operator prohibited`. The switch/case seems to be the only way to go unfortunately...

I think I found a better way, if I replace bpf_for() with a simple for loop + pragma unroll everything works, see commit 6bc75949c.

lib/pmu.bpf.c
```diff
@@ -51,12 +50,14 @@ int scx_pmu_event_stop(struct task_struct __arg_trusted *p)
...
 	int idx;
 	int ret;
+	__u32 key;
 
 	cntrs = bpf_task_storage_get(&scx_pmu_tasks, p, 0, BPF_LOCAL_STORAGE_GET_F_CREATE);
 	if (!cntrs)
 		return -ENOENT;
 
-	bpf_for(idx, 0, SCX_MAX_PMU_COUNTERS) {
+#pragma unroll(2)
```
etsal: Nit: Can we align this with the loop

lib/pmu.bpf.c
```diff
@@ -95,12 +97,14 @@ int scx_pmu_event_start(struct task_struct __arg_trusted *p, bool update)
...
```
etsal: Same

etsal (approved):

PR #3351, opened 2026-02-20
[DRAFT] lib: Fix memory leaks in rbtree, ATQ and cpu.max, add cgroup stress test.

multics69:
This PR adds cgroupv2 stress-testing infrastructure and fixes arena
memory leaks in the rbtree and ATQ data structures.

- Add `cg_stress.sh`: a script that stress-tests cgroupv2 by repeatedly
  creating and deleting cgroups in batches, reporting live throughput
  and memory stats.
- Add `--list` / `--test` CLI options to `scx_arena_selftests` for
  targeted test execution without reading source.
- Add `scx_alloc_get_pages_used()` to track total arena memory usage
  across static and dynamic allocators — enables memory leak assertions
  in selftests.
- Fix memory leak in `lib/rbtree`: add `scx_alloc_get_pages_used()`,
  add leak selftest (`scx_selftest_rbtree_leak`), and transition
  `rb_create_internal()` / `rb_destroy()` from `scx_static_alloc` to
  `scx_alloc` so that `rb_destroy()` actually frees the rbtree.
- Fix memory leak in `lib/atq`: add `scx_atq_destroy()`, add leak
  selftest (`scx_selftest_atq_leak`), and transition
  `scx_atq_create_internal()` / `scx_atq_destroy()` from
  `scx_static_alloc` to `scx_alloc`.
- Fix memory leak in cpu.max by wiring the destroy call into
  `cbw_free_llc_ctx()` to fix the cgroup bandwidth cleanup path.

### DIFF
--- a/lib/atq.bpf.c
+++ b/lib/atq.bpf.c
@@ -3,6 +3,7 @@
 #include <lib/sdt_task.h>
 
 #include <lib/atq.h>
+#include <lib/alloc/buddy.h>
 
 /*
  * Arena task queue implementation.
@@ -13,17 +14,18 @@ u64 scx_atq_create_internal(bool fifo, size_t capacity)
 {
 	scx_atq_t *atq;
 
-	atq = scx_static_alloc(sizeof(*atq), 1);
+	atq = sys_buddy_zalloc(sizeof(*atq));
 	if (!atq)
 		return (u64)NULL;
 
 	atq->tree = rb_create(RB_NOALLOC, RB_DUPLICATE);
-	if (!atq->tree)
+	if (!atq->tree) {
+		sys_buddy_free(atq);
 		return (u64)NULL;
+	}
 
 	atq->fifo = fifo;
 	atq->capacity = capacity;
-	atq->size = 0;
 
 	return (u64)atq;
 }
@@ -37,6 +39,7 @@ int scx_atq_destroy(scx_atq_t __arg_arena *atq)
 		/* Do nothing. Just drain all the queued tasks. */
 	}
 	rb_destroy(atq->tree);
+	sys_buddy_free(atq);
 
 	return 0;
 }


multics69:
It fixes memory leak in rbtree, ATQ, and cpu.max. However, it seems there are more memory leaks. `cg_stress.sh` fails after creating 65520 and 2670 cgroups in 6.19 and 6.17, respectively.

lib/selftests/selftest.h
```diff
@@ -3,6 +3,26 @@
...
 #include <lib/rbtree.h>
 
+/*
+ * Test IDs for arena_selftest's selftest_run_id selector.
+ * SCX_SELFTEST_ID_ALL (0) runs all tests; any other value runs only the
+ * matching test.
+ */
+enum scx_selftest_id {
+	SCX_SELFTEST_ID_ALL			= 0,
+	SCX_SELFTEST_ID_ARENA_TOPOLOGY_TIMER	= 1,
```
etsal, line 13: After the first is set to 0 we don't need to set the reat

lib/sdt_alloc.bpf.c
```diff
@@ -616,9 +617,8 @@ u64 scx_static_alloc_internal(size_t bytes, size_t alignment)
 		 * No free operation so just forget about the previous
 		 * allocation memory.
 		 */
-
-		memory = bpf_arena_alloc_pages(&arena, NULL,
-					       scx_static.max_alloc_bytes / PAGE_SIZE,
+		page_cnt = scx_static.max_alloc_bytes / PAGE_SIZE,
```
etsal: typo?

scheds/include/lib/rbtree.h
```diff
@@ -62,18 +64,22 @@ enum rbtree_insert_mode {
...
 typedef struct rbtree __arena rbtree_t;
+
+int scx_rb_init(void);
+
 #ifdef __BPF__
 u64 rb_create_internal(enum rbtree_alloc alloc, enum rbtree_insert_mode insert);
 #define rb_create(alloc, insert) ((rbtree_t *)rb_create_internal((alloc), (insert)))
 
-int rb_destroy(rbtree_t *rbtree);
+int rb_destroy(rbtree_t __arg_arena *rbtree);
```
etsal, line 77: This is unnecessary, declaration attributes do nothing

multics69: Hmm... without this, the verifier screams. :-(

lib/sdt_alloc.bpf.c
multics69: Fixed!

lib/selftests/selftest.h
multics69: This is intentional to avoid mistakes in the future because those enums are mirrored in the Rust part (enum SelfTestId).

multics69:
> Thani you for working on this! AFAICT this isn't really leaks in the data structures, but rather we're making an assumption ATQs aren't created and destroyed correct?

You are right. That was the original assumption. However, that assumption is broken in the case of cgroup. Because in cpu.max, each cgroup has its associated ATQ. When a cgropu is destroyed, its associated ATQ will remain in the arena. That's why now we need to properly destroy everything. 

> In terms of allocation: I don't think adding per-data structure allocators is the best way to go about this. We can start using the buddy allocator we have in the tree that is generic and supports ASAN.
> 
> This will not change the testing, but it will the changes to the data structures. This means we can keep like 95% of this and rebase on top.

Thanks for the suggestion. That sounds better, although I was a bit worried about fragmentation when using the buddy allocator. I will change it as you suggested.

lib/atq.bpf.c
```diff
@@ -8,15 +8,27 @@
...
-	if (!atq)
+	data = scx_alloc(&scx_atq_allocator);
+	if (unlikely(!data))
 		return (u64)NULL;
 
+	atq = (scx_atq_t *)data->payload;
+	atq->tid = data->tid;
+
 	atq->tree = rb_create(RB_NOALLOC, RB_DUPLICATE);
 	if (!atq->tree)
```
bboymimi: It seems the data allocated before is not freed after the scx_atq_create_internal() returns. And the data will be leaked as well. This should be fixed by:
```c
if (!atq->tree) {
        scx_alloc_free_idx(&scx_atq_allocator, atq->tid.idx);
        return (u64)NULL;
}       
```

multics69:
Close this PR as the next version is submitted: https://github.com/sched-ext/scx/pull/3361

PR #3352, opened 2026-02-20
scx_simulator: A userspace scheduler simulation framework for sched_ext

rrnewton:
# Description

 scx_simulator enables running sched_ext BPF schedulers like LAVD entirely in userspace without requiring a real kernel. The framework compiles scheduler BPF code as regular C code, in a shared library, providing stub implementations of kernel kfuncs (like scx_bpf_dsq_insert, scx_bpf_select_cpu_dfl, etc.) that model kernel scheduling behavior. Workloads are defined using rt-app JSON format, specifying task phases (run/sleep durations), CPU affinities, cgroup membership, and parent-child relationships. The simulator executes these workloads against the loaded scheduler, producing execution traces that capture every scheduling decision, context switch, and DSQ operation.

The simulator models relevant kernel scheduling semantics including per-CPU and per-domain dispatch queues, timer-based preemption, CPU idle tracking, cgroup hierarchies with bandwidth control, and task migration.  For LAVD specifically, it supports the full latency criticality calculation, slice boost mechanism, wake-affine scheduling, and cross-domain load balancing. A probe infrastructure allows inspection of internal scheduler state (e.g., `avg_runtime_wall, lat_cri, sys_stat`) at each scheduling event, enabling detailed analysis of scheduler behavior. The framework also supports real kernel tracing via `--real-run --bpf-trace` mode, allowing statistical comparison between simulated and actual kernel execution. Traces can be export as perfetto traces just like real, live-kernel runs.

# Uses and features

This tool serves multiple purposes: deterministic regression testing of scheduler logic, concurrency stress testing of schedulers (in simulation), and debugging scheduler behavior through standard tools (e.g. debuggers) as well as very easy to add visibility (tracing, visualization, etc) compared to live kernels. The combination of fast feedback loops and deep observability makes it practical to develop and test branches of scheduler code that are otherwise hard to reach. Code coverage is measured using standard tools, making it feasible for agents to automatically crank up code coverage for scheduler testing.

For deterministic concurrency (chaos) testing, the simulator executes the scheduler C code in a single threaded way, but with random preemption/interleaving at any branch point (RBC counter), as well as any kfunc call. This enables stress runs to systematically explore interleavings searching for a concurrency bug.  Though many more features are needed in this department.

# Why's it ready

There's still a long way to go in terms of modeling more and adding fidelity and further tooling. The reason I think this has reached MVP is:
 * I was able to successfully reproduce two non-stall LAVD scheduler crashes found in the fleet with the alpha binary.
 * Stress testing is already able to turn up scheduler failures that are then completely reproducible/debuggable.

# Logic for living in this repo

Like scxtop, this is an isolated tool that lives in its own directory. Alas, there are some specific C-wrappers for each scheduler that needs to change along with the schedulers. This could live in the sched-test repo, but it would make the linkage to the schedulers less tight. However, if a scheduler change breaks the simulator, it is actually extremely easy to ask an agent  to update it so that it passes. I suggest we try it in this repo for a while and if it causes a hassle I'll move it to the other (which would then have to submodule scx to "pin" the scheduler versions).

### DIFF
--- a/rust/scx_simulator/crates/scx_simulator/tests/compare.rs
+++ b/rust/scx_simulator/crates/scx_simulator/tests/compare.rs
@@ -17,9 +17,9 @@ mod common;
 fn test_compare_two_runners_mitosis() {
     let _lock = common::setup_test();
     let json = include_str!("../workloads/two_runners.json");
-    let scenario = load_rtapp(json, 2).unwrap();
+    let scenario = load_rtapp(json, 4).unwrap();
 
-    let trace = Simulator::new(DynamicScheduler::mitosis(2)).run(scenario);
+    let trace = Simulator::new(DynamicScheduler::mitosis(4)).run(scenario);
     trace.dump();
 
     // Sanity: both tasks should run
@@ -56,9 +56,9 @@ fn test_compare_two_runners_mitosis() {
 fn test_mitosis_trace_stats() {
     let _lock = common::setup_test();
     let json = include_str!("../workloads/two_runners.json");
-    let scenario = load_rtapp(json, 2).unwrap();
+    let scenario = load_rtapp(json, 4).unwrap();
 
-    let trace = Simulator::new(DynamicScheduler::mitosis(2)).run(scenario);
+    let trace = Simulator::new(DynamicScheduler::mitosis(4)).run(scenario);
 
     // Compute and print statistics
     let stats = TraceStats::from_trace(&trace);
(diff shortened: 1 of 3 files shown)


rrnewton:
This was started Feb 9 with the "simulator" branch (f1394d1e793e51b052d4efd6c114daa135e31028). `simulator.v2` is just a rebase with adjustment to the latest scheduler changes.

etsal:
This is great! I wonder if it is big enough to merit its own repo.

htejun:
Yeah, given the size and likely high volume of expected changes, I think it'd be better to give it its own repo if possible.

arighi:
I also think this is really interesting, thanks @rrnewton! Definitely worth having its own repo under the sched-ext org.

multics69:
Thanks @rrnewton ! This is fantastic! Definitely it deserves a new repo under the sche-ext.

rrnewton:
Ok, I think the sentiment is clear and I'll put it in the sched-test repo. I'd like to collect tools there.

I'll just warn you that you won't be able to get any CI signal on scheduler bugs introduced! The coverage is >75% for LAVD and I think it could serve for regression testing.  I'm working on validating that with mutation testing.  (I.e. if we introduce a bug on line XYZ, will it be caught by real scheduler smoke tests, and/or in simulation.)

But I'll wait to see if we BELATEDLY find any bugs when the "scx" submodule is updated and the simulator gets the latest scheduler changes.  If so we can reassess in the future.

PR #3357, opened 2026-02-24
scx_layered: Unified per-node CPU allocation

htejun:
## Summary

scx_layered's CPU allocation was node-blind. This series replaces the ad-hoc
approach with a unified per-node allocation system that makes NUMA a
first-class concern, culminating in the removal of the `allow_node_aligned`
hack that was propping up the old design.

## Background

scx_layered has two allocation paths:

**Non-StickyDynamic layers** use `core_order: Vec<usize>`, a flat list of
CPU indices produced by the growth algorithm. Growing iterates forward,
shrinking iterates backward. The growth algorithms (Linear, Reverse, Random,
Sticky, BigLittle, CpuSetSpread) each produce different orderings, but none
were node-aware — the node distribution of allocated CPUs was accidental,
determined entirely by the iteration order of the flat Vec.

**StickyDynamic layers** operate at LLC granularity with `assigned_llcs` and
`free_llcs`. `take_llc` / `return_llc` move LLCs between pools. While
`free_llcs` was technically per-node, `assigned_llcs` was flat, so shrink
popped from the back of a flat Vec without regard for node locality.

Both paths shared the same problem: there was no mechanism to express "this
layer needs X CPUs on node 0 and Y CPUs on node 1" based on where the
layer's tasks are actually pinned.

### Workarounds that accumulated

To mitigate node-blindness, two mechanisms were layered on:

1. **Spread algorithms** (RoundRobin, NodeSpread, NodeSpreadReverse,
   NodeSpreadRandom) — force equal CPU counts across all NUMA nodes, so
   layers aren't accidentally concentrated on one node.

2. **`allow_node_aligned`** — a BPF flag that lets node-aligned tasks (tasks
   whose cpumask covers exactly one NUMA node) bypass the lo-fallback DSQ
   and dispatch on their node's layer DSQ even if the layer doesn't own CPUs
   there. Without this, node-aligned tasks on a layer that happened to get
   CPUs on the wrong node would be throttled in lo-fallback.

Together, spread + allow_node_aligned formed a fragile workaround: spread
ensures the layer has CPUs on every node, and allow_node_aligned prevents
tasks from being throttled when the allocation doesn't match their affinity.
Neither addresses the root cause — the allocator doesn't know about nodes.

### Unified per-node allocation (`alloc.rs`)

A new allocation module decomposes each layer's CPU demand into per-node
pinned and unpinned components, then distributes CPUs using a water-fill
algorithm:

1. **Pinned demand is allocated first.** Pinned tasks can only run on their
   node, so their demand is inflexible. Each layer's pinned demands are
   scaled to fit within its weight-proportional budget, then distributed
   per-node via water-fill against each node's available capacity.

2. **Unpinned demand absorbs the remainder.** After pinned allocation, the
   leftover budget goes to unpinned work. If pinned demand on a full node
   couldn't be fully satisfied, the unspent pinned budget flows automatically
   to the unpinned side, giving the layer more flexibility.

3. **Demand-capped layers free excess.** When a layer's total demand
   (pinned + unpinned) is less than its budget, it is locked at its actual
   usage and the excess is redistributed to layers that can use it. This is
   the water-fill lock step — it repeats until all budget is placed or all
   layers are satisfied.

4. **Spread layers get equal per-node budgets.** After the global water-fill
   determines a spread layer's total, `resolve_spread()` distributes it
   evenly across nodes, capped at the least-available node capacity. This
   replaces the old approach where spread was purely an ordering trick.

The result: pinned tasks get CPUs on the right node, unpinned tasks absorb
the rest, spread layers get balanced budgets, and no capacity is wasted.

### Per-node core_order

`core_order` is split from a flat `Vec<usize>` into per-node
`Vec<Vec<usize>>`. Each growth algorithm now produces a per-node ordering via
`node_order()`, which determines the preferred NUMA node and the order of
remaining nodes. Within each node, the algorithm determines core selection
order. The unified allocator's per-node budgets then determine how many CPUs
to take from each node's ordering.

### Per-node StickyDynamic LLC trading

StickyDynamic's `free_llcs` pool is restructured to be per-node with explicit
node ordering. `take_llc` drains the preferred node first before spilling to
remote nodes. `return_llc` releases remote-node LLCs before home-node LLCs
during shrink, preserving locality.

### Deprecate `allow_node_aligned`

With proper per-node allocation, layers naturally have CPUs on the nodes
where their tasks run. The `allow_node_aligned` flag is no longer necessary
and is deprecated — node-aligned task dispatch is now unconditional.

### Unit tests (185 tests)

The series adds comprehensive Rust unit tests covering:
- CpuPool basics (creation, grow/shrink, topology queries)
- All growth algorithms on 1/2/4-node topologies (core_order correctness,
  node preference, locality properties)
- StickyDynamic LLC trading (take/return, per-node ordering, cross-node
  spillover, multi-layer competition)
- Cross-node overflow (demand exceeding home node spills correctly)
- `unified_alloc` (water-fill distribution, pinned-first allocation, spread
  layer budgets, edge cases)

### Hardware tests (40+ test cases, all passing)

Tested on a dual-socket AMD EPYC 7642 (192 CPUs, 2 NUMA nodes, 32 LLCs) and
a single-node 16-CPU QEMU VM. Five test phases:

**Phase 1 — Single-node equivalence** (VM): Linear, RoundRobin,
StickyDynamic all match baseline behavior within ±2 CPUs.

**Phase 2 — Multi-node sub-overflow** (loads ≤ 96 CPUs): Single-layer
staircases (Linear, RR, SD), pinned layers, two-layer same-algo, two-layer
cross-algo, three-layer mixed — all pass.

**Phase 3 — Multi-node overflow** (loads up to 180 CPUs): Same matrix as
Phase 2 but with loads exceeding single-node capacity. Verifies correct
cross-node spillover. Notable improvements: SD layers that were stuck at
~14 CPUs in baseline now scale to 38-40. Cross-algo resource sharing
significantly improved.

**Phase 4 — Edge cases**: Rapid oscillation (5s phases), zero-to-max-to-zero,
steady-state convergence, growth-from-zero under saturation. All stable, no
oscillation.

**Before/after for `allow_node_aligned` deprecation**: Under overload
(80 workers on 18-CPU layer), baseline showed 71.4% lo-fallback dispatches.
With the patch, lo-fallback drops to 0.15%.

### DIFF
--- a/scheds/rust/scx_layered/examples/cpuset.json
+++ b/scheds/rust/scx_layered/examples/cpuset.json
@@ -32,8 +32,7 @@
             0.8,
             0.9
           ],
-          "min_exec_us": 100,
-          "allow_node_aligned": true
+          "min_exec_us": 100
         }
       }
     }
--- a/scheds/rust/scx_layered/examples/same_over_idle.json
+++ b/scheds/rust/scx_layered/examples/same_over_idle.json
@@ -51,8 +51,7 @@
             0.8,
             0.9
           ],
-          "min_exec_us": 100,
-          "allow_node_aligned": true
+          "min_exec_us": 100
         }
       }
     }
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -355,7 +355,6 @@ struct layer {
 	bool			preempt;
 	bool			preempt_first;
 	bool			excl;
-	bool			allow_node_aligned;
 	bool			has_cpuset;
 	bool			skip_remote_node;
 	bool			prev_over_idle_core;
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1778,13 +1778,13 @@ skip_ddsp:
 	 *   always run on all CPUs the layer will ever use. Guaranteed to
 	 *   be false for non-cpuset layers by refresh_cpus_flags().
 	 *
-	 * - allow_node_aligned && cpus_node_aligned: The task's affinity
-	 *   covers whole NUMA nodes. With per-node allocation, the layer
-	 *   will always have CPUs on those nodes.
+	 * - cpus_node_aligned: The task's affinity covers whole NUMA nodes.
+	 *   With per-node allocation, the layer will always have CPUs on
+	 *   those nodes.
 	 */
 	if ((!taskc->all_cpus_allowed &&
 	     !taskc->all_cpuset_cpus_allowed &&
-	     !(layer->allow_node_aligned && taskc->cpus_node_aligned)) ||
+	     !taskc->cpus_node_aligned) ||
 	    !layer->nr_cpus) {
 		// Special handle for thread that has affinity set, but need more CPU time.
 		// XXX: If we need to support more than one thread (which is probably bad from
--- a/scheds/rust/scx_layered/src/config.rs
+++ b/scheds/rust/scx_layered/src/config.rs
@@ -119,8 +119,8 @@ pub struct LayerCommon {
     pub preempt_first: bool,
     #[serde(default)]
     pub exclusive: bool,
-    #[serde(default)]
-    pub allow_node_aligned: bool,
+    #[serde(default, skip_serializing)]
+    pub allow_node_aligned: Option<bool>,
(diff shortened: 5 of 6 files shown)


htejun:
@kkdwvd Can you please take a look at StickyDynamic changes? It's now contained within each node. This is going to be the last StickyDynamic change for NUMA support.

PR #3358, opened 2026-02-24
layered: Drop ops.cpu_release()

EricccTaiwan:
Ref: #3343, #3345.


htejun:
I think we want to keep the ability to requeue tasks when preempted by higher scheduling classes.

EricccTaiwan:
> I think we want to keep the ability to requeue tasks when preempted by higher scheduling classes.

Got it!

PR #3361, opened 2026-02-25
lib: cgroup ATQ lifecycle cleanup + scx_alloc() migration for rbtree/ATQ

multics69:
This series addresses proper cleanup of ATQ during cgroup exit, motivated by `cg_stress.sh` -- a new cgroupv2 stress script that repeatedly creates and destroys thousands of cgroups to exercise the cgroup lifecycle path in `lib/cgroup_bw`.

Running the stress test revealed that `cbw_free_llc_ctx()` had a TODO where ATQ cleanup should occur: when a cgroup exits, its per-LLC ATQs were never reclaimed. The root cause was that ATQ and rbtree were originally backed by `scx_static_alloc()`, which provides a simple bump allocator suitable for objects allocated once and never freed. That model works well for long-lived, scheduler-global structures, but breaks down for per-cgroup state that must be released on cgroup exit.

To support proper ATQ destruction, this series:

1. **Migrates rbtree to `scx_alloc()`** -- introduces `scx_rb_init()` to set up slab allocators for `rbtree_t` and `rbnode_t`, stores allocation IDs in `tid` fields, and implements `rb_destroy()` to drain and free all nodes and the tree itself. The per-tree freelist is retained as a local cache to reduce contention under concurrent operations.

2. **Migrates ATQ to `scx_alloc()`** -- introduces `scx_atq_init()` and stores the allocation ID in `atq->tid`, allowing `scx_atq_destroy()` to free the ATQ after draining the queue and destroying its backing rbtree.

3. **Fixes `rb_destroy()` parameter** -- adds `__arg_arena` to the `rbtree_t *` parameter so the BPF verifier accepts the call from arena context (e.g., from `scx_atq_destroy()` which operates on `scx_atq_t __arg_arena *`).

4. **Implements ATQ cleanup in `cbw_free_llc_ctx()`** -- replaces the TODO with a call to `scx_atq_destroy()`, properly reclaiming arena memory when a cgroup exits. Also tightens `cbw_free_llc_ctx()` to handle NULL context pointers, uses `cgroup_get_id()` for reliable error logging, and fixes `scx_cgroup_bw_exit()` to correctly propagate return codes from descendant updates.

5. **Improves `scx_arena_selftests`** -- adds `--list` and `--test NAME` CLI options for selective test execution, and colorizes `[ PASS ]` / `[ FAIL ]` output (gated on `IsTerminal` so piped output stays plain text).

5. **Adds `scripts/cg_stress.sh`** -- a bash script that stress tests cgroupv2 by repeatedly creating and deleting configurable batches of cgroups. Prints a live status line with iteration count, cgroup throughput (cgroups/s), current memory usage, delta from start (Δstart), and instantaneous memory rate (kB/s).

**Testing:** `scx_lavd --enable-cpu-bw` survived 10M+ cgroup construction/destruction iterations under `cg_stress.sh` without crashing or leaking arena memory.

### DIFF
--- a/lib/atq.bpf.c
+++ b/lib/atq.bpf.c
@@ -8,22 +8,36 @@
  * Arena task queue implementation.
  */
 
+static struct scx_allocator scx_atq_allocator;
+
+__weak
+int scx_atq_init(void)
+{
+	return scx_alloc_init(&scx_atq_allocator, sizeof(scx_atq_t));
+}
+
 __weak
 u64 scx_atq_create_internal(bool fifo, size_t capacity)
 {
+	struct sdt_data __arena *data = NULL;
 	scx_atq_t *atq;
 
-	atq = scx_static_alloc(sizeof(*atq), 1);
-	if (!atq)
+	/* Note that scx_alloc() returns a zero-initialized memory. */
+	data = scx_alloc(&scx_atq_allocator);
+	if (unlikely(!data))
 		return (u64)NULL;
 
+	atq = (scx_atq_t *)data->payload;
+	atq->tid = data->tid;
+
 	atq->tree = rb_create(RB_NOALLOC, RB_DUPLICATE);
-	if (!atq->tree)
+	if (!atq->tree) {
+		scx_alloc_free_idx(&scx_atq_allocator, atq->tid.idx);
 		return (u64)NULL;
+	}
 
 	atq->fifo = fifo;
 	atq->capacity = capacity;
-	atq->size = 0;
 
 	return (u64)atq;
 }
@@ -38,6 +52,7 @@ int scx_atq_destroy(scx_atq_t __arg_arena *atq)
 	}
 	rb_destroy(atq->tree);
 
+	scx_alloc_free_idx(&scx_atq_allocator, atq->tid.idx);
 	return 0;
 }
 
--- a/lib/arena.bpf.c
+++ b/lib/arena.bpf.c
@@ -11,6 +11,7 @@
 #include <lib/cpumask.h>
 #include <lib/topology.h>
 #include <lib/rbtree.h>
+#include <lib/atq.h>
 
 /*
  * "System-call" based API for arenas.
@@ -54,6 +55,12 @@ int arena_init(struct arena_init_args *args)
 	ret = scx_rb_init();
 	if (ret) {
 		bpf_printk("scx_rb_init failed with %d", ret);
+		return ret;
+	}
+
+	ret = scx_atq_init();
+	if (ret) {
+		bpf_printk("scx_atq_init failed with %d", ret);
 		return ret;
 	}
 
--- a/scheds/include/lib/atq.h
+++ b/scheds/include/lib/atq.h
@@ -22,6 +22,7 @@ enum scx_task_throttle {
 };
 
 struct scx_atq {
+	union sdt_id tid;
 	rbtree_t *tree;
 	arena_spinlock_t lock;
 	u64 capacity;
@@ -32,6 +33,8 @@ struct scx_atq {
 
 
 typedef struct scx_atq __arena scx_atq_t;
+
+int scx_atq_init(void);
 
 struct scx_task_common {
 	struct rbnode node;	/* rbnode for being inserted into ATQs */
--- a/scheds/rust/scx_wd40/build.rs
+++ b/scheds/rust/scx_wd40/build.rs
@@ -22,6 +22,7 @@ fn main() {
         .add_source("../../../lib/sdt_alloc.bpf.c")
         .add_source("../../../lib/topology.bpf.c")
         .add_source("../../../lib/rbtree.bpf.c")
+        .add_source("../../../lib/atq.bpf.c")
         .compile_link_gen()
         .unwrap();
 }


lib/selftests/selftest.h
```diff
@@ -3,6 +3,24 @@
 #include <lib/atq.h>
 #include <lib/rbtree.h>
 
+/*
+ * Test IDs for arena_selftest's selftest_run_id selector.
+ * SCX_SELFTEST_ID_ALL (0) runs all tests; any other value runs only the
+ * matching test.
+ */
+enum scx_selftest_id {
```
etsal, line 20: Just repeating to confirm I got this right from our last conversation: We need to explicitly number the enums not because of any runtime behavior we need, but because this file will be parsed to produce bpf_intf.rs and the way we do it requires we explicitly number them. Is that accurate? I

I am seeing in the selftests that we are copying the enums there too. Do we actually use these exact definitions from Rust?  If not, we can remove all = apart from  = 0, the C standard still guarantees the enums will be consecutive.

lib/atq.bpf.c
```diff
@@ -8,26 +8,53 @@
...
+
 __weak
 u64 scx_atq_create_internal(bool fifo, size_t capacity)
 {
+	struct sdt_data __arena *data = NULL;
 	scx_atq_t *atq;
 
-	atq = scx_static_alloc(sizeof(*atq), 1);
-	if (!atq)
+	data = scx_zalloc(&scx_atq_allocator);
```
etsal: The current allocator should already guarantee all bytes are zeroed: Initially all pages from the kernel are zeroed, and we zero out the memory during a free. This is a bug in the allocator that I will backport the fix of (we fixed it for tools/sched_ext/scx_sdt.bpf.c when we moved it to the kernel).

rust/scx_arena/selftests/src/main.rs
```diff
@@ -36,6 +39,66 @@ use libbpf_rs::ProgramInput;
...
+    SCX_SELFTEST_ID_BITMAP = 9,
+}
+
+const TEST_CASES: &[(&str, u32)] = &[
+    (
+        "arena_topology_timer",
+        SelfTestId::SCX_SELFTEST_ID_ARENA_TOPOLOGY_TIMER as u32,
+    ),
+    ("atq", SelfTestId::SCX_SELFTEST_ID_ATQ as u32),
+    ("dhq", SelfTestId::SCX_SELFTEST_ID_DHQ as u32),
```
etsal: We can remove dhq and arena_topology_timer, they are a bit untested and as you mentioned don't pass.

rust/scx_arena/selftests/src/main.rs
```diff
@@ -207,6 +295,27 @@ fn main() {
...
+            println!("  {}", name);
+        }
+        return;
+    }
+
+    // Validate test names before loading BPF.
+    for name in &opts.tests {
+        if !TEST_CASES.iter().any(|(n, _)| *n == name.as_str()) {
+            eprintln!(
+                "Unknown test: '{}'. Use --list to see available tests.",
```
etsal: Can we print the available tests if the user uses the wrong one?

scheds/include/lib/sdt_task.h
```diff
@@ -58,8 +58,26 @@ int scx_alloc_free_idx(struct scx_allocator *alloc, __u64 idx);
 
 #define scx_alloc(alloc) ((struct sdt_data __arena *)scx_alloc_internal((alloc)))
 
+static __always_inline
```
etsal: I will get the bug fixed after which we can remove this. 

Note (probably won't be an issue here, but for reference): LLVM and BPF have an outstanding issue where the BPF backend crashes when it comes across the memset() intrinsic. Loops of zeros sometimes get promoted to memset by prior passes. Just mentioning it in case our code triggers this behavior at some point.

lib/atq.bpf.c
multics69: Once `tools/sched_ext/scx_sdt.bpf.c` is merged, I will clean the `zalloc` part accordingly.

rust/scx_arena/selftests/src/main.rs
multics69: Okay.

rust/scx_arena/selftests/src/main.rs
multics69: Sure.

scheds/include/lib/sdt_task.h
multics69: > Note (probably won't be an issue here, but for reference): LLVM and BPF have an outstanding issue where the BPF backend crashes when it comes across the memset() intrinsic. Loops of zeros sometimes get promoted to memset by prior passes. Just mentioning it in case our code triggers this behavior at some point.

Wow! If it happens, is there any workaround to avoid the LLVM crash?

etsal: Unfortunately no :( This is one of the reasons array operations are finicky in BPF. This is not an issue most of the time because BPF structs are small,  but with arenas that no longer holds. The right way is to implement the missing intrinsics in LLVM.

lib/atq.bpf.c
multics69: Maybe renaming `scx_alloc` to `scx_zalloc` would be one way to make the API's behavior more explicit. Maybe, in a separate PR?

rust/scx_arena/selftests/src/main.rs
multics69: Do you also want to remove `bitmap` too? The `bitmap` selftest also failed. 

```
[ FAIL ] arena_topology_timer (returned -16)
===BEGIN STREAM OUTPUT===

====END STREAM  OUTPUT====
===BEGIN STREAM ERROR===

====END STREAM  ERROR====
[ FAIL ] dhq (returned -12)
===BEGIN STREAM OUTPUT===

====END STREAM  OUTPUT====
===BEGIN STREAM ERROR===

====END STREAM  ERROR====
[ FAIL ] bitmap (returned -95)
===BEGIN STREAM OUTPUT===

====END STREAM  OUTPUT====
===BEGIN STREAM ERROR===

====END STREAM  ERROR====
One or more selftests failed.
```

scheds/include/lib/sdt_task.h
multics69: Ah, that's why you used `zero` instead of `0`?

lib/selftests/selftest.h
multics69: This is purely for pool proofing. Values are explicitly numbered because this enum is mirrored in Rust as `SelfTestId` in `rust/scx_arena/selftests/src/main.rs`. Keeping explicit numbers makes mismatches between the two copies immediately visible: a wrong  number is a grep-time catch, whereas a wrong ordering would be a silent runtime mismatch that is much harder to detect.

Does it make sense to you? Also, I am okay with the style without explicit numbering.

scheds/include/lib/sdt_task.h
etsal: No, that's a different matter: Using volatile zero ensures the entire loop converges by making i imprecise.

rust/scx_arena/selftests/src/main.rs
etsal: Let's remove it, yes. Error -95 is EOPNOTSUPP, so we're not losing anything by removing it - it was more of a placeholder.

lib/atq.bpf.c
etsal: How about we keep zalloc() as a name only for call sites where we explicitly care about zero initialization, and for now resolve it to alloc()? This way when we move to the proper allocator we don't zero data when not necessary.

lib/selftests/selftest.h
etsal: Fair, I like the idea of being able to easily visually inspect the enum. Let's keep it as-is.

lib/atq.bpf.c
multics69: Make sense. I drop the `scx_zalloc()` for now. Let's reintroduce it later when we are ready.

lib/cgroup_bw.bpf.c
```diff
@@ -540,29 +540,35 @@ static
 void cbw_free_llc_ctx(struct cgroup *cgrp, struct scx_cgroup_ctx *cgx)
 {
 	struct scx_cgroup_llc_ctx *llcx;
+	scx_atq_t *btq;
 	int i;
 
-	if (!cgrp || !cgx || !cgx->has_llcx)
```
etsal, line 545: Just to confirm, this is deliberate that we support running this function even if !cgx, right? Is there a non-bug reason why cgx could be NULL?

lib/cgroup_bw.bpf.c
```diff
@@ -862,17 +868,15 @@ int scx_cgroup_bw_init(struct cgroup *cgrp __arg_trusted, struct scx_cgroup_init
...
-	int ret;
+	int ret = 0;
 
 	cbw_dbg_cgrp();
-	if (cgrp->level > 1) {
-		if ((ret = cbw_update_nr_taskable_descendents(cgrp, -1)))
-			return ret;
-	}
+	if (cgrp->level > 1)
+		ret = cbw_update_nr_taskable_descendents(cgrp, -1);
```
etsal, line 875: Again to confirm, this addresses an existing leak in case cgrp->level == 1 in that we don't actually call the free functions below, correct??

rust/scx_arena/selftests/src/main.rs
```diff
@@ -36,6 +39,65 @@ use libbpf_rs::ProgramInput;
...
+    }
+}
+
+// Mirrors enum scx_selftest_id in lib/selftests/selftest.h.
+// SCX_SELFTEST_ID_ALL (0) is reserved for "run all" and is not listed in
+// TEST_CASES; only the named per-test IDs appear there.
+#[repr(u32)]
+#[allow(non_camel_case_types)]
+enum SelfTestId {
+    #[allow(dead_code)]
```
etsal, line 60: Closing up our discussion of this, I think this is ok for now and ideally in the future we have a single location across Rust and C where we define those.

multics69: I agree. Ideally, the rust enum should be generated from a C header file.

lib/cgroup_bw.bpf.c
multics69: yes, in the error case.

lib/cgroup_bw.bpf.c
multics69: In the `scx_cgroup_bw_exit()`, `cgx` is destructed first (`cbw_del_cgroup_ctx`) then we free `llcx` (`cbw_free_llc_ctx(cgrp, NULL);`).

PR #3364, opened 2026-02-26
lib/sdt_alloc: Fix erroneous condition that prevents zero'ing

etsal:
Address the bug in sdt_alloc.bpf.c fixed upstream by commit 2e06d54ea9a25. The bug prevents freed slab allocator memory from being zeroed and requires manual zero'ing by the caller.


lib/sdt_alloc.bpf.c
multics69: I am a bit confused if it is correct to zero out util `alloc->pool.elem_size`. In the `scx_alloc_init()`, we did `data_size += sizeof(struct sdt_data);`, so the payload size should be `alloc->pool.elem_size - sizeof(struct sdt_data)`. Did I miss something?

etsal:
Good catch, the size we have stored includes the struct sdt_data, so it needs to be adjusted. I've updated the diff.

hodgesds (approved):
LGTM

PR #3365, opened 2026-02-26
scx_cosmos: Add tick hook for preemption enforcement

devnexen:
Without a tick callback, cosmos relies entirely on voluntary yields for CPU handoff. This means a CPU-bound task can hold the CPU well beyond its 10µs time slice, starving latency-sensitive workloads like nginx workers waiting on epoll wakeups.

The tick hook enforces preemption in three cases:

1. Slice enforcement: preempt tasks that exceed their weighted time slice, ensuring no task monopolizes a CPU.

2. Priority-aware preemption: when the system is busy and tasks are queued in the shared DSQ, preempt low-frequency wakers (batch work) so high-frequency sleepers like nginx workers get the CPU sooner.

3. SMT rebalancing: when avoid_smt is enabled and the SMT sibling has gone busy, trigger rescheduling so select_cpu can find a less contended core.

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -124,6 +124,11 @@ const volatile bool deferred_wakeups = true;
 const volatile bool perf_sticky;
 
 /*
+ * Enable tick-based preemption enforcement.
+ */
+const volatile bool tick_preempt = true;
+
+/*
  * Ignore synchronous wakeup events.
  */
 const volatile bool no_wake_sync;
@@ -994,6 +999,37 @@ static inline void wakeup_cpu(s32 cpu)
 	scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
 }
 
+void BPF_STRUCT_OPS(cosmos_tick, struct task_struct *p)
+{
+	struct task_ctx *tctx;
+
+	if (!tick_preempt)
+		return;
+
+	tctx = try_lookup_task_ctx(p);
+	if (!tctx)
+		return;
+
+	/*
+	 * Force preemption if the task has exceeded its time slice and
+	 * either:
+	 * - SMT contention has changed since we started running
+	 *   (sibling went busy or idle), triggering rescheduling so
+	 *   select_cpu can make a better placement decision, or
+	 * - the system is busy and there are tasks waiting in the
+	 *   local or shared DSQ.
+	 */
+	if (time_delta(scx_bpf_now(), tctx->last_run_at) > task_slice(p)) {
+		s32 cpu = scx_bpf_task_cpu(p);
+		bool smt_contention = avoid_smt && is_smt_contended(cpu);
+		bool cpu_busy = scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | cpu) ||
+				scx_bpf_dsq_nr_queued(shared_dsq(cpu));
+
+		if (smt_contention || (is_system_busy() && cpu_busy))
+			p->scx.slice = 0;
+	}
+}
+
 void BPF_STRUCT_OPS(cosmos_enqueue, struct task_struct *p, u64 enq_flags)
 {
 	s32 prev_cpu = scx_bpf_task_cpu(p), cpu;
@@ -1279,6 +1315,7 @@ SCX_OPS_DEFINE(cosmos_ops,
 	       .select_cpu		= (void *)cosmos_select_cpu,
 	       .enqueue			= (void *)cosmos_enqueue,
 	       .dispatch		= (void *)cosmos_dispatch,
+	       .tick                    = (void *)cosmos_tick,
 	       .runnable		= (void *)cosmos_runnable,
 	       .running			= (void *)cosmos_running,
 	       .stopping		= (void *)cosmos_stopping,
--- a/scheds/rust/scx_cosmos/src/main.rs
+++ b/scheds/rust/scx_cosmos/src/main.rs
@@ -244,6 +244,13 @@ struct Opts {
     #[clap(short = 'd', long, action = clap::ArgAction::SetTrue)]
     no_deferred_wakeup: bool,
 
+    /// Disable tick-based preemption enforcement.
+    ///
+    /// By default, the scheduler preempts tasks that exceed their time slice when the system is
+    /// busy or SMT contention is detected. Use this flag to disable this behavior.
+    #[clap(long, action = clap::ArgAction::SetTrue)]
+    no_tick_preempt: bool,
+
     /// Enable address space affinity.
     ///
     /// This option allows to keep tasks that share the same address space (e.g., threads of the
@@ -453,6 +460,7 @@ impl<'a> Scheduler<'a> {
         rodata.nr_node_ids = topo.nodes.len() as u32;
         rodata.no_wake_sync = opts.no_wake_sync;
         rodata.avoid_smt = opts.avoid_smt;
+        rodata.tick_preempt = !opts.no_tick_preempt;
         rodata.mm_affinity = opts.mm_affinity;
 
         // Enable perf event scheduling settings.


scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -994,6 +994,38 @@ static inline void wakeup_cpu(s32 cpu)
...
+	 * Force preemption if:
+	 * - the task has exceeded its time slice, or
+	 * - the system is busy and there are tasks waiting in the
+	 *   shared DSQ with the current task being a low-frequency
+	 *   waker (likely batch work), or
+	 * - SMT contention has changed since we started running
+	 *   (sibling went busy or idle), triggering rescheduling so
+	 *   select_cpu can make a better placement decision.
+	 */
+	if (runtime > task_slice(p) ||
```
arighi: This should be done already by the kernel, is it really needed?

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -994,6 +994,38 @@ static inline void wakeup_cpu(s32 cpu)
...
+	    (is_system_busy() && scx_bpf_dsq_nr_queued(shared_dsq(cpu)) > 0 && tctx->wakeup_freq == 0) ||
```
arighi: Invoking this kfunc from ops.tick() has probably a cost, especially if the system is busy. Do you have some numbers in terms of throughput? I'm wondering how much this is going to slow down certain throughput-intensive workloads...

devnexen:
Thanks for the review.

Addressed both points:

- Dropped the `runtime > task_slice(p)` check, you're right the kernel handles this already.
- Gated the entire tick path behind a `tick_preempt` flag (off by default) so `scx_bpf_dsq_nr_queued()` is never called unless explicitly opted in via `--tick-preempt`.

Ran a quick nginx benchmark (wrk, 4 threads, 256 connections, 30s, static file on localhost):

| | main (no tick) | tick (no-op) | --tick-preempt |
|---|---|---|---|
| Req/sec | 547,714 | 648,890 | 632,573 |
| Avg latency | 239.79µs | 208.40µs | 216.06µs |
| p50 | 232µs | 190µs | 167µs |
| p99 | 468µs | 479µs | 568µs |

Default path (tick registered but no-op) actually improves throughput by ~18% over main. The `--tick-preempt` mode trades some p99 (~19% higher) for better p50 (~12% lower), which makes sense for latency-sensitive workloads but stays opt-in to avoid penalizing throughput.

Also added a `bench_nginx.sh` script to make it easy to reproduce.

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
arighi: > This should be done already by the kernel, is it really needed?

@devnexen actually I take that back, it is done by the kernel, but I did some tests and it looks like this condition can happen, likely because the slice decrementing logic is a bit flaky.

So... can we re-add this check back? Sorry for going back and forth...

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -994,6 +999,31 @@ static inline void wakeup_cpu(s32 cpu)
...
+	int cpu;
+
+	if (!tick_preempt)
+		return;
+
+	tctx = try_lookup_task_ctx(p);
+	if (!tctx)
+		return;
+
+	cpu = bpf_get_smp_processor_id();
```
arighi: This is not right, it should be scx_bpf_task_cpu(p).

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -994,6 +999,31 @@ static inline void wakeup_cpu(s32 cpu)
 	scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
 }
 
+void BPF_STRUCT_OPS(cosmos_tick, struct task_struct *p)
+{
+	struct task_ctx *tctx;
+	int cpu;
+
+	if (!tick_preempt)
```
arighi, line 1006: Thanks for adding this option!

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -994,6 +999,31 @@ static inline void wakeup_cpu(s32 cpu)
...
+		return;
+
+	cpu = bpf_get_smp_processor_id();
+
+	/*
+	 * Preempt batch tasks when higher-priority work is queued,
+	 * or when SMT contention changes warrant rescheduling.
+	 */
+	if ((is_system_busy() && scx_bpf_dsq_nr_queued(shared_dsq(cpu)) > 0 && tctx->wakeup_freq == 0) ||
+	    (avoid_smt && is_smt_contended(cpu))) {
```
arighi: What do you think about doing this:
```
        if (time_delta(scx_bpf_now(), tctx->last_run_at) > task_slice(p)) {
                s32 cpu = scx_bpf_task_cpu(p);
                bool smt_contention = avoid_smt && is_smt_contended(cpu);
                bool cpu_busy = scx_bpf_dsq_nr_queued(SCX_DSQ_LOCAL_ON | cpu) ||
                                scx_bpf_dsq_nr_queued(shared_dsq(cpu));

                if (smt_contention || (is_system_busy() && cpu_busy))
                        p->scx.slice = 0;
        }
```

I'm not very convinced about the `tctx->wakeup_freq == 0` condition, this should applied to all tasks for better fairness.

scheds/rust/scx_cosmos/src/main.rs
```diff
@@ -244,6 +244,13 @@ struct Opts {
     #[clap(short = 'd', long, action = clap::ArgAction::SetTrue)]
     no_deferred_wakeup: bool,
 
+    /// Enable tick-based preemption enforcement.
+    ///
+    /// Preempt batch tasks when higher-priority work is queued and rebalance on SMT contention
+    /// changes. Useful for latency-sensitive workloads.
+    #[clap(long, action = clap::ArgAction::SetTrue)]
+    tick_preempt: bool,
```
arighi: If we agree with the logic above, I think we can we flip this and enable it by default, like having a `--no-tick-preempt` option.

scheds/rust/scx_cosmos/bench_nginx.sh
```diff
@@ -0,0 +1,104 @@
...
+#
+# Benchmark nginx latency/throughput with scx_cosmos, comparing
+# default mode vs --tick-preempt mode.
+#
+# Prerequisites:
+#   - nginx running and serving on TARGET_URL
+#   - wrk installed (https://github.com/wg/wrk)
+#   - scx_cosmos built (cargo build --release -p scx_cosmos)
+#
+# Usage: sudo ./bench_nginx.sh [TARGET_URL] [DURATION] [THREADS] [CONNECTIONS]
```
arighi: Thanks for sharing the benchmark, but we shouldn't include this in the repo.

devnexen: I dropped in the last commit :) but I ll look at your other comments soon-ish

devnexen:
Addressed all review feedback:

- Fixed `bpf_get_smp_processor_id()` → `scx_bpf_task_cpu(p)`.
- Simplified tick logic per your suggestion: dropped `wakeup_freq == 0` filter, gated everything behind `runtime > task_slice(p)`, checking both local and shared DSQ.
- Flipped `--tick-preempt` → `--no-tick-preempt` (enabled by default).

Benchmark results (wrk, 4 threads, 256 connections, 30s, static file on localhost nginx 1.28.0):

**main vs branch:**

| | main (no tick hook) | branch (tick preempt on) |
|---|---|---|
| Req/sec | 612,553 | 623,697 (+1.8%) |
| Avg latency | 225.04µs | 225.56µs |
| p50 | 181µs | 193µs |
| p75 | 266µs | 252µs (-5.3%) |
| p90 | 398µs | 362µs (-9.0%) |
| p99 | 550µs | 524µs (-4.7%) |
| Stdev | 112.13µs | 86.35µs (-23.0%) |

**branch: default vs --no-tick-preempt:**

| | tick preempt (default) | --no-tick-preempt |
|---|---|---|
| Req/sec | 728,052 | 660,126 (+10.3%) |
| Avg latency | 196.48µs | 215.38µs |
| p50 | 172µs | 189µs |
| p99 | 437µs | 481µs (-9.1%) |

No throughput regression vs main, significantly tighter tail latency distribution (stdev -23%, p90 -9%). Enabling by default looks safe — the `--no-tick-preempt` escape hatch is there if needed.

arighi:
@devnexen just one last request before merging, can we squash all the commits in a single one? I don't see any reason to keep all the different commits. Thanks!

arighi:
Awesome thanks @devnexen

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
multics69: > > This should be done already by the kernel, is it really needed?
> 
> @devnexen actually I take that back, it is done by the kernel, but I did some tests and it looks like this condition can happen, likely because the slice decrementing logic is a bit flaky.

@arighi -- I wonder in what condition this can happen. I took a look `update_curr_scx()` in the kernel but it looks okay.

PR #3366, opened 2026-02-26
scx_layered: Per-node LLC drain and fallback dispatch

htejun:
## Summary
- Reduce MAX_NUMA_NODES from 64 to 8 to shrink per-node arrays
- Factor per-node layer fields (nr_cpus, nr_pinned_tasks) into layer_node_ctx
- Make fallback dispatch per-node — each node independently tracks which layers have zero local CPUs and services their stranded DSQ tasks
- Add per-node LLC list to node_ctx for iterating only local LLCs
- Make LLC drain per-node — llcs_to_drain/llc_drain_cnt move into layer_node_ctx, drain only iterates local node's LLCs, idle CPU kick is node-local

## Testing
Tested on 2-node/32-LLC machine with BPF trace instrumentation confirming node locality: 290 drain enable/disable events all correctly map LLCs to their NUMA node, drain kicks only target node-local CPUs, per-node LLC lists are correctly propagated (node 0 gets LLCs 0-15, node 1 gets LLCs 16-31), and per-node layer cpumasks receive equal updates for both nodes.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -362,6 +362,36 @@ struct {
 	__uint(map_flags, 0);
 } layer_cpumasks SEC(".maps");
 
+/*
+ * Per-layer per-node cpumasks (layer cpumask & node cpumask). Stored in a
+ * separate map because struct layer is in BSS which is mmapped and can't
+ * contain kptrs. Keyed by layer_id * MAX_NUMA_NODES + node_id.
+ */
+struct layer_node_cpumask_wrapper {
+	struct bpf_cpumask __kptr *cpumask;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__type(key, u32);
+	__type(value, struct layer_node_cpumask_wrapper);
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -575,6 +637,24 @@ int BPF_PROG(refresh_layer_cpumasks)
...
+	struct node_ctx *nodec;
+
+	if (!(nodec = lookup_node_ctx(arg->node_id)))
+		return -ENOENT;
+
+	nodec->nr_empty_layer_ids = arg->nr_empty_layer_ids;
+	__builtin_memcpy(nodec->empty_layer_ids, arg->empty_layer_ids,
+			 sizeof(nodec->empty_layer_ids));
+
+	nodec->nr_llcs = arg->nr_llcs;
```
hodgesds: this probably only needs updated once right? I guess it's infrequent anyways

htejun: Right, added an init flag to refresh_node_ctx_arg — the LLC list is now only copied when init is set (during startup). Periodic updates skip it.

PR #3377, opened 2026-03-02
scx_cosmos: GPU auto-affinitization

arighi:
Introduce the new options `--gpu` / `--gpu-util-threshold` to enable automatic GPU affinity on multi-GPU systems. When enabled, the scheduler will attempt to keep tasks using a GPU on the NUMA node associated with that device.

In the future, this mechanism can be generalized to support affinity for other device types as well, evolving from GPU-specific affinity into a broader device-aware scheduling policy.

Cc: @balbirs-nv

### DIFF
--- a/scheds/rust/scx_cosmos/src/main.rs
+++ b/scheds/rust/scx_cosmos/src/main.rs
@@ -11,11 +11,13 @@ pub mod bpf_intf;
 pub use bpf_intf::*;
 
 mod stats;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::ffi::{c_int, c_ulong};
+use std::fs;
 use std::fs::File;
 use std::io::{BufRead, BufReader};
 use std::mem::MaybeUninit;
+use std::path::Path;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
@@ -31,6 +33,8 @@ use libbpf_rs::MapFlags;
 use libbpf_rs::OpenObject;
 use libbpf_rs::ProgramInput;
 use log::{debug, info, warn};
+use nvml_wrapper::bitmasks::InitFlags;
+use nvml_wrapper::Nvml;
 use scx_stats::prelude::*;
 use scx_utils::build_id;
 use scx_utils::compat;
@@ -42,6 +46,7 @@ use scx_utils::try_set_rlimit_infinity;
 use scx_utils::uei_exited;
 use scx_utils::uei_report;
 use scx_utils::CoreType;
+use scx_utils::GpuIndex;
 use scx_utils::Topology;
 use scx_utils::UserExitInfo;
 use scx_utils::NR_CPU_IDS;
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_cosmos/src/main.rs
```diff
@@ -678,9 +744,112 @@ impl<'a> Scheduler<'a> {
...
+                        // NotSupported or other: fall back to all running processes.
+                        Self::add_running_gpu_processes_to_set(&device, node, &mut pid_to_nodes);
+                    }
+                }
+            } else {
+                Self::add_running_gpu_processes_to_set(&device, node, &mut pid_to_nodes);
+            }
+        }
+
+        // Only add PIDs that use exactly one GPU to the map.
```
xerothermic, line 799: This is a big assumption. We thought this is the dominate use case initially, but I've seen more usage where single process open multiple GPU contexts, and it broke our assumption and the `--enable-gpu-affinitize` flag in scx_layered.

Something to watch out as you deploy your scheduler...

scheds/rust/scx_cosmos/src/main.rs
```diff
@@ -187,6 +192,17 @@ struct Opts {
...
+    /// Enable GPU-aware scheduling.
+    #[clap(short = 'g', long, action = clap::ArgAction::SetTrue)]
+    gpu: bool,
+
+    /// Only treat a process as GPU-bound if its GPU utilization is at least this percentage (0–100).
+    ///
+    /// Uses NVML process utilization (SM + memory). 0 = no filter (all processes on the GPU are
+    /// considered GPU-bound). Requires driver support (Maxwell or newer).
+    #[clap(long, default_value = "0", value_parser = clap::value_parser!(u32).range(0..=100))]
+    gpu_util_threshold: u32,
```
xerothermic, line 204: Is the idea for gpu_util_threshold to distinguish idle vs. active GPU processes?

scheds/rust/scx_cosmos/src/main.rs
arighi: True. However, if a task is actively utilizing multiple GPUs across different NUMA nodes (based on the GPU utilization threshold) at the same time, it's probably not that beneficial attempting to migrate it elsewhere.

A more refined approach could be to detect whether the task is using multiple GPUs within the same NUMA node and, in that case, record the association in the BPF map. For more complex topologies, we could also support N:M relationships. for example "soft-affinitizing" a PID to a set of N GPUs spanning one or more NUMA nodes. But yeah, that's probably an area where we can improve the policy in the future.

scheds/rust/scx_cosmos/src/main.rs
arighi: Correct, the idea is to avoid affinitizing tasks that "barely" touch a GPU, maybe they're just checking/configuring something and they're not really using it. Hence the utilization threshold.

arighi:
Merging this since it's pretty well self contained, regression potential is minimal and all my basic tests are showing positive results. Of course there's more work to do, but it seems good enough as a first prototype.

PR #3378, opened 2026-03-03
scx_layered: Add cross-NUMA migration gating

htejun:
## Summary

Add userspace-driven, rate-limited gating for cross-NUMA scheduling paths in
scx_layered. Without this, the BPF scheduling paths (idle CPU selection, DSQ
consumption, preemption) freely cross NUMA boundaries, causing catastrophic
memory latency and bandwidth degradation on multi-node systems.

The approach uses per-task duty cycle tracking to measure per-(layer, node) load,
then a userspace water-fill algorithm computes migration rate budgets that the BPF
side enforces via token buckets on three NUMA-crossing paths:

- **pick_idle_cpu()**: After exhausting same-node idle CPUs, cross-NUMA idle
  selection is gated by the token bucket rate.
- **try_consume_layer()**: LLC proximity walk stops at the NUMA boundary unless
  the token bucket allows cross-node consumption.
- **Preemption walk**: CPU proximity walk stops at node_end unless the token
  bucket allows cross-node preemption.

Gate activation uses two-threshold hysteresis: a node becomes a migration source
only when its load/alloc ratio exceeds thresholds AND the allocation system is
denying growth requests. When gates are inactive, scheduling behaves exactly as
before (no overhead — token bucket rate is U64_MAX).

Rate dampening (0.5x per cycle) ensures gradual convergence rather than
single-step overcorrection.

15 patches, 33 unit tests for the water-fill algorithm and rate dampening.

## Patches

1. **Use plain crate** for BPF map serialization (replaces raw pointer casts)
2. **Factor out layer_node_ctx** from struct layer
3. **Add per-node utilization stats** (util, queue time)
4. **Fix LSTAT_PREEMPT_XLLC** not being counted
5. **Add cross-NUMA preemption stat**
6. **Per-task duty cycle tracking** (EWMA load_avg)
7. **Per-CPU per-layer demand tracking** (duty_sum aggregation)
8. **Userspace water-fill computation** (rate budget calculation)
9. **Token bucket gate helpers** (BPF-side rate enforcement)
10. **Per-node proximity map** (node_prox_map for ordered node iteration)
11. **Gate pick_idle_cpu()** at NUMA boundary
12. **Gate try_consume_layer()** at NUMA boundary
13. **Gate preemption walk** at NUMA boundary
14. **xnuma stats and observability**
15. **Dampen migration rates** for gradual convergence

## Testing

Tested on a 2-node (2x AMD EPYC 9654, 192 CPUs) system with stress-ng workloads
saturating fair-class. All scheduling paths (idle, consume, preempt) gate correctly
under load. 33 unit tests pass for the water-fill algorithm. Gate activation
follows the two-threshold hysteresis as designed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
- "gimli 0.31.1",
+ "gimli",
 ]
 
 [[package]]
@@ -36,7 +36,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom 0.3.3",
+ "getrandom 0.3.4",
  "once_cell",
  "version_check",
  "zerocopy",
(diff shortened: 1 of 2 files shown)


htejun:
@hodgesds Once this lands, cross-node behavior defaults to gating migrations. You can get the previous behavior of migrating freely across node boundaries by setting `xnuma_threshold` to `[0, 0]`.

PR #3379, opened 2026-03-03
scx_lavd: guard division-by-zero in sys_stat collect/calc paths

rrnewton:
## Summary

- Fix per-CPU `compute_wall` leak across loop phases in `collect_sys_stat()`: phase 3 was using the last CPU's value from phase 1 instead of computing per-CPU. Recompute with `time_delta()` and guard with `?: 1`.
- Guard `duration_total_wall` divisor in `calc_sys_stat()` with `?: 1` for defensive robustness.

Both use the existing `?: 1` idiom (see `init_sys_stat_ctx()` line 87) for consistency.

## Details

In `collect_sys_stat()`, `compute_wall` is set per-CPU in the phase 1 loop (line 189) but used in the phase 3 loop (line 325) with only the stale value from the **last** CPU processed in phase 1. This has two problems:

1. Phase 3 uses the wrong CPU's `compute_wall` for every CPU it processes.
2. When the last CPU in phase 1 was fully idle (`idle_total_wall >= duration_wall`), `compute_wall` is 0, causing division by zero.

On a real kernel the BPF verifier silently returns 0 for div-by-zero, but the code should guard explicitly for correctness and clarity.

## Test plan

- [ ] Verify LAVD scheduler builds cleanly
- [ ] Run `scx_lavd` on a system with idle CPUs to confirm no behavioral regression

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
@@ -294,6 +294,7 @@ static void collect_sys_stat(void)
 	 * Collect statistics for each CPU (phase 3).
 	 */
 	bpf_for(cpu, 0, nr_cpu_ids) {
+		u64 cpu_compute_wall;
 		struct cpu_ctx *cpuc = get_cpu_ctx_id(cpu);
 		if (!cpuc) {
 			c->compute_total_wall = 0;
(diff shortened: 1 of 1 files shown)


rrnewton:
@multics69 @daidavid -- this may be very unlikely in live kernel but I was hitting it in simulation-testing of the LAVD code.  It made me fix my divide-by-zero (SIGFPE) handling in the simulator ;-).

multics69 (approved):
Thanks @rrnewton for the fix. It looks good to me.

PR #3380, opened 2026-03-03
Change BUFSIZE to use queued_task_ctx type

michaelr0:
This resolves an issue, when extending the `QueuedTask` struct within a `rustland_core based scheduler`
The following error is then thrown

```sh
thread 'main' (74915) panicked at scheds/rust/scx_rustland/src/bpf.rs:255:23:
copy_from_slice: source slice length (88) does not match destination slice length (96)
```

This happens because if you have modified the `QueuedTask` struct, then the original check calculates the length of the struct, which no longer matches the length of the buf from bpf, this changes moves the length restriction to the length of the bpf buf, so that userspace changes/logic can be added without impacting the buf.


EricccTaiwan (approved):
LGTM, thanks.

arighi (approved):

PR #3388, opened 2026-03-05
common.bpf.h, scx_lavd: use task clock for runtime accounting

multics69:
This series improves runtime accounting in scx_lavd by introducing
proper per-CPU clock helpers in `common.bpf.h` and switching lavd's
runtime measurement from a task-lifetime accumulator to the direct
per-CPU clock source.

**Patch 1** adds four BPF clock helpers to `scheds/include/scx/common.bpf.h`:

- `scx_clock_task(cpu)` -- equivalent to `rq_clock_task()`: wall-clock
  time excluding IRQ and hypervisor steal time.
- `scx_clock_pelt(cpu)` -- equivalent to `rq_clock_pelt()`: capacity-
  and frequency-invariant clock, continuous across idle transitions.
- `scx_clock_virt(cpu)` -- lagging shadow of `paravirt_steal_clock()`:
  cumulative stolen time. Returns 0 when `CONFIG_PARAVIRT_TIME_ACCOUNTING`
  is disabled, via `bpf_core_field_exists()`.
- `scx_clock_irq(cpu)` -- equivalent to `irq_time_read()`: cumulative
  hardirq and non-ksoftirqd softirq time. Returns 0 when
  `CONFIG_IRQ_TIME_ACCOUNTING` is disabled, via `bpf_core_type_exists()`.

**Patch 2** replaces lavd's `task_exec_time()` -- which read
`p->se.sum_exec_runtime` as a proxy for `rq->clock_task` -- with
`scx_clock_task()` directly. This removes the indirection through the
per-task accumulator and makes the intent clearer: we are measuring the
CPU clock, not the task's lifetime runtime. The helper `task_exec_time()`
is removed as it no longer has callers, and the `task_ctx` field
`last_sum_exec_clk` is renamed to `last_task_clk` accordingly.

No behavioral changes are intended.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -403,8 +403,8 @@ static void update_stat_for_running(struct task_struct *p,
 	reset_task_flag(taskc, LAVD_FLAG_IS_WAKEUP);
 	reset_task_flag(taskc, LAVD_FLAG_IS_SYNC_WAKEUP);
 	taskc->last_running_clk = now;
-	taskc->last_measured_clk = now;
-	taskc->last_task_clk = scx_clock_task(cpuc->cpu_id);
+	taskc->last_measured_wall_clk = now;
+	taskc->last_measured_task_clk = scx_clock_task(cpuc->cpu_id);
 
 	/*
 	 * Reset task's lock and futex boost count
@@ -466,8 +466,8 @@ static void account_task_runtime(struct task_struct *p,
 				 struct cpu_ctx *cpuc,
 				 u64 now)
 {
-	u64 suspended_wall, runtime_wall, task_time_wwgt, task_time_invr;
-	u64 task_time_wall, exec_delta_wall;
+	u64 task_time_wall, task_time_wwgt, task_time_invr;
+	u64 suspended_wall, duration_wall, now_task;
 
 	/*
 	 * Since task execution can span one or more sys_stat intervals,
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -480,17 +480,16 @@ static void account_task_runtime(struct task_struct *p,
...
-	 * as cpuc->stolen_time_est to try and approximate the total time CPU spent in
-	 * stolen time (irq+steal). Until more accurate irq_time/steal_time snapshots
-	 * are available from the kernel, we can use the samples from sum_exec_runtime
-	 * to extrapolate total stolen time per CPU.
+	 * runtime - exec_delta must equal to irq_time + steal_time barring
+	 * some imprecision when the time was snapshotted. We accumulate the
+	 * delta as cpuc->stolen_time_est to try and approximate the total
+	 * time CPU spent in stolen time (irq+steal). Until more accurate
+	 * irq_time/steal_time snapshots are available from the kernel, we can
+	 * use the samples from sum_exec_runtime to extrapolate total stolen
```
bboymimi: Is it a stale reference to sum_exec_runtime? Should this comment be updated to reference scx_clock_task()? The task_exec_time() has been replaced by scx_clock_task().

daidavid (approved):

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Thanks! Fixed!

PR #3389, opened 2026-03-05
Cargo.lock: Update dependencies to latest versions

EricccTaiwan:


sirlucjan (approved):
LGTM but maybe ask also @htejun and @arighi because last time `cargo update` broke scx sources for Ubuntu.

PR #3393, opened 2026-03-06
Mitosis: dynamic affinity CPU selection for pinned tasks (based on PR #3206)

tommy-u:
scx_mitosis: add dynamic CPU selection for affinitized tasks
    
      Tasks affinitized to a subset of their cell were bound to a single CPU
      from their affinity mask at initialization. This mostly works since most
      affinitized tasks target a single CPU, but we've observed cases where
      tasks affinitized to multiple CPUs cause significant load imbalance.
    
      Add --dynamic-affinity-cpu-selection which re-assigns such tasks to a
      random CPU within their affinity mask when no idle CPU is available.
      This doesn't fully eliminate imbalance but prevents long-term pileup
      without adding load-balancing complexity.
    
      Includes a single-CPU fast path (bpf_cpumask_weight == 1), bounds
      checking on bpf_cpumask_any_distribute, restored cpu < 0 guards on
      legacy paths, and an idle_cpu_cleared flag to avoid double-clearing
      the idle bit after pick_idle_cpu_from. Kick idle in enqueue in case
      its a different CPU from what select picked.
    
      Test script added to reproduce and measure the behavior.

Copying description from @dschatzberg, original author of #3206 this is based on. 

Tasks that are affinitized (e.g. to a subset of their cell) are presently simply bound to a single cpu out of their affinity mask. This mostly works since most affinitized tasks are just affinitized to a single cpu. We've observed instances where tasks are affinitized to multiple cpus and the initialization-time binding can lead to significant cpu imbalance.

This commit adds a new flag --dynamic-affinity-cpu-selection which will dynamically re-assign such tasks (only affinitized tasks) to random cpus within their affinity mask in order to reduce load imbalance. This doesn't completely eliminate cpu load imbalance for these kinds of tasks but avoids load-balancing or other complexities and ultimately prevents long-term imbalance due to making random decisions at each scheduling point.

Additionally I've added a test script to reproduce this behavior (may take a few runs due to the random assignment).

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -48,6 +48,7 @@ const volatile bool reject_multicpu_pinning = false;
 const volatile bool userspace_managed_cell_mode = false;
 const volatile bool enable_borrowing = false;
 const volatile bool use_lockless_peek = false;
+const volatile bool dynamic_affinity_cpu_selection = false;
 
 /*
  * Global arrays for LLC topology, populated by userspace before load.
@@ -634,6 +635,58 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p, s32 prev_cpu
 }
 
 /*
+ * Switch task to a new CPU's per-CPU DSQ with vtime reset.
+ * Returns new_cpu on success, -1 on failure (tctx unchanged).
+ */
+static __always_inline s32 update_pinned_dsq(struct task_struct *p, struct task_ctx *tctx,
+					     s32 new_cpu)
+{
+	s32 current_cpu = get_cpu_from_dsq(tctx->dsq);
+	if (current_cpu < 0)
+		return -1;
+
+	if (current_cpu == new_cpu)
+		return new_cpu; /* already on this DSQ */
+
+	struct cpu_ctx *new_cctx = lookup_cpu_ctx(new_cpu);
+	if (!new_cctx)
+		return -1;
+
+	tctx->dsq = get_cpu_dsq_id(new_cpu);
+	p->scx.dsq_vtime = READ_ONCE(new_cctx->vtime_now);
+	return new_cpu;
+}
+
+static __always_inline s32 select_pinned_cpu(struct task_struct *p, s32 prev_cpu,
+					     struct task_ctx *tctx, bool *idle_cpu_cleared)
+{
+	s32 cpu;
+
+	/*
+	 * Dynamic affinity balancing: use pick_idle_cpu_from
+	 * for SMT-aware idle CPU selection. If no idle CPU,
+	 * return prev_cpu and let enqueue() handle placement.
+	 */
+	const struct cpumask *idle_smtmask __free(idle_cpumask) = scx_bpf_get_idle_smtmask();
+	if (!idle_smtmask) {
+		scx_bpf_error("Failed to get idle smtmask");
+		return -1;
+	}
+
+	cpu = pick_idle_cpu_from(p, p->cpus_ptr, prev_cpu, idle_smtmask);
+
+	if (cpu < 0)
+		return prev_cpu;
+
+	*idle_cpu_cleared = true;
+
+	if (update_pinned_dsq(p, tctx, cpu) < 0) {
+		return -1;
+	}
+	return cpu;
+}
+
+/*
  * select_cpu is where we update each task's cell assignment and then try to
  * dispatch to an idle core in the cell if possible
  */
@@ -651,11 +704,23 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu, u64 
 
 	if (!tctx->all_cell_cpus_allowed) {
 		cstat_inc(CSTAT_AFFN_VIOL, tctx->cell, cctx);
-		cpu = get_cpu_from_dsq(tctx->dsq);
+		bool idle_cpu_cleared = false;
+
+		if (bpf_cpumask_weight(p->cpus_ptr) == 1) {
+			/* If we're pinned to a single CPU, just use that */
+			cpu = get_cpu_from_dsq(tctx->dsq);
+		} else if (dynamic_affinity_cpu_selection) {
+			/* Multicpu pinning, try to find an idle CPU */
+			cpu = select_pinned_cpu(p, prev_cpu, tctx, &idle_cpu_cleared);
+		} else {
+			/* Legacy pinned CPU, stay on DSQ you were initialized to */
+			cpu = get_cpu_from_dsq(tctx->dsq);
+		}
+
 		if (cpu < 0)
 			return prev_cpu;
 
-		if (scx_bpf_test_and_clear_cpu_idle(cpu))
+		if (idle_cpu_cleared || scx_bpf_test_and_clear_cpu_idle(cpu))
 			scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, slice_ns, 0);
 		return cpu;
 	}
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -679,6 +742,30 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu, u64
...
+	 * has tasks queued, randomly pick from allowed CPUs to
+	 * balance load across compatible CPUs over time.
+	 *
+	 * This runs even if select_cpu() was called, because
+	 * select_cpu() may have picked a random CPU when no
+	 * idle CPUs were available, without checking queue depth.
+	 */
+	s32 cpu;
+	cpu = get_cpu_from_dsq(tctx->dsq);
+	if (scx_bpf_dsq_nr_queued(tctx->dsq.raw) > 0) {
```
likewhatevs, line 760: this seems a bit aggressive, is 0 a reasonable number here? Maybe gate on idk how long stuck runnable or queue depth or something?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -679,6 +742,30 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu, u64
...
+static __always_inline s32 enqueue_pinned_cpu(struct task_struct *p, struct task_ctx *tctx)
+{
+	/*
+	 * Dynamic affinity balancing: if current assigned CPU
+	 * has tasks queued, randomly pick from allowed CPUs to
+	 * balance load across compatible CPUs over time.
+	 *
+	 * This runs even if select_cpu() was called, because
+	 * select_cpu() may have picked a random CPU when no
+	 * idle CPUs were available, without checking queue depth.
```
likewhatevs: should we instead propagate that idle was picked w/ a flag on tctx and respect the selection if so?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -633,6 +634,56 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p, s32 prev_cpu
...
+	* Dynamic affinity balancing: use pick_idle_cpu_from
+	* for SMT-aware idle CPU selection. If no idle CPU,
+	* randomly pick to balance load over time.
+	*/
+	idle_smtmask = scx_bpf_get_idle_smtmask();
+	if (!idle_smtmask)
+		return prev_cpu;
+
+	cpu = pick_idle_cpu_from(p, p->cpus_ptr, prev_cpu, idle_smtmask);
+	scx_bpf_put_idle_cpumask(idle_smtmask);
```
dschatzberg: I think we have RAII helpers for this now - if not we can add them

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -633,6 +634,56 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p, s32 prev_cpu
...
+
+	if (cpu >= 0)
+		*idle_cpu_cleared = true;
+	else
+		cpu = bpf_cpumask_any_distribute(p->cpus_ptr);
```
dschatzberg: Is this bit actually important? The logic otherwise was just to return `prev` which would just run enqueue on the last cpu to run p which feels right to me - there's no reason to spray this now and interrupt another CPU - you're going to do that in enqueue().

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -633,6 +634,56 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p, s32 prev_cpu
...
 }
 
+/*
+ * Switch task to a new CPU's per-CPU DSQ with vtime reset.
+ * Returns new_cpu on success, -1 on failure (tctx unchanged).
+ */
+static __always_inline s32 update_pinned_dsq(struct task_struct *p, struct task_ctx *tctx,
+					     s32 new_cpu)
+{
+	if (get_cpu_dsq_id(new_cpu).raw == tctx->dsq.raw)
```
dschatzberg: I thought you might have a helper for this already? get cpu from dsq or something

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: If select_cpu picked an idle task, we would have put it on the local dsq - no enqueue()

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -633,6 +634,56 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p, s32 prev_cpu
...
+	cpu = pick_idle_cpu_from(p, p->cpus_ptr, prev_cpu, idle_smtmask);
+	scx_bpf_put_idle_cpumask(idle_smtmask);
+
+	if (cpu >= 0)
+		*idle_cpu_cleared = true;
+	else
+		cpu = bpf_cpumask_any_distribute(p->cpus_ptr);
+
+	/* When no idle CPU */
+	if (cpu >= nr_possible_cpus || update_pinned_dsq(p, tctx, cpu) < 0) {
```
dschatzberg: I tend to prefer the logical checks like `cpu >= nr_possible_cpus` living on their own conditional + scx_bpf_error + return.

The error case from update_pinned_dsq is actually a bit strange to me - you might have already hit scx_bpf_error here and yet you will give back prev_cpu and do a bunch more work. The return value here should just be -1 and then you'll immediately return prev_cpu in select_cpu which seems right for an error.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -679,6 +742,30 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu, u64
...
+	 * select_cpu() may have picked a random CPU when no
+	 * idle CPUs were available, without checking queue depth.
+	 */
+	s32 cpu;
+	cpu = get_cpu_from_dsq(tctx->dsq);
+	if (scx_bpf_dsq_nr_queued(tctx->dsq.raw) > 0) {
+		s32 new_cpu;
+
+		new_cpu = bpf_cpumask_any_distribute(p->cpus_ptr);
+		if (new_cpu < nr_possible_cpus)
```
dschatzberg: Again, invert conditional, scx_bpf_error, and return. This is silently moving forward on a logically impossible case.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -679,6 +742,30 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu, u64
...
+			if (update_pinned_dsq(p, tctx, new_cpu) >= 0)
```
dschatzberg: Again, invert conditional, hard return on error

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u: Might be a dumb heuristic. Noted

tommy-u:
> I would change the flag situation though. I'd say this should replace multipin reject flag. So like, maybe multipin cpu reject is swapped w/ this? Although idk, if in doubt flag i suppose.

Unfortunately this is insufficient to drop the reject multicpu pinning flag. It's about half of what it takes to do that. I think the other half is dynamic slice shrinking.

tommy-u:
Addressed all comment. I like the idea to do something smarter than redistributing when there are > 0 tasks. I think tracking runnable times is a good suggestion for the future.

PR #3397, opened 2026-03-07
scx_layered: Bump MAX_NUMA_NODES to 32 and consolidate node stats

### LINKED ISSUE #3384: [scx_layered] BPF program fails to load on systems with more than 8 NUMA nodes (AMD EPYC 9654 chiplet, 24 NUMA nodes)
## Environment

| Item | Details |
|---|---|
| Kernel | Linux 6.12 |
| CPU | 2 × AMD EPYC 9654 (96 cores each, 192 cores total) |
| NUMA topology | 24 NUMA nodes (1 node per CCD chiplet, 12 CCDs × 2 sockets) |
| scx version | upstream main branch |
| Scheduler | scx_layered |

## Problem Description

`scx_layered` fails to run on systems with more than 8 NUMA nodes. The failure manifests in two stages depending on the value of `MAX_NUMA_NODES` in `intf.h`.

---

### Stage 1: Runtime panic with default `MAX_NUMA_NODES=8`

Running the unmodified binary on a 24-NUMA-node system causes a panic at startup:

```
thread 'main' panicked at scheds/rust/scx_layered/src/main.rs:2108:22:
index out of bounds: the len is 8 but the index is 8
```

The Rust userspace code iterates over all detected NUMA nodes and accesses `skel.maps.rodata_data.numa_cpumasks[node_id]`. Since the array length is fixed at compile time to `MAX_NUMA_NODES=8`, accessing `node_id=8` (the 9th node) triggers a bounds check panic.

---

### Stage 2: BPF program load failure after increasing `MAX_NUMA_NODES`

Naively bumping `MAX_NUMA_NODES` to either 24 (exact hardware count) or 32 causes the BPF program to fail loading:

```
Failed to load BPF program
caused by: Argument list too long (os error 7)
```

htejun:
## Summary

- Bump `MAX_NUMA_NODES` from 8 to 32 to support AMD EPYC processors with
  per-CCD NUMA (NPS4) that can expose 24+ nodes. Verified with veristat on
  both 6.12 and 6.18 kernels — all 28 BPF programs pass verification.
- Consolidate the separate `node util/load` and `pinned util/tasks` stats
  lines into a single `node pin/ut/ld` line with proper label alignment.

Fixes #3384

## Test plan

- [x] veristat verification on 6.12 and 6.18 kernels
- [x] Build test (`cargo build --release -p scx_layered`)
- [x] Runtime test on 2-node bare-metal (chk) with stress-ng workload

### DIFF
--- a/scheds/rust/scx_layered/src/stats.rs
+++ b/scheds/rust/scx_layered/src/stats.rs
@@ -474,16 +474,11 @@ impl LayerStats {
             fmt_pct(self.skip_remote_node),
         )?;
 
-        // per-node utilization and load (multi-node only)
+        // per-node utilization, load, and pinned utilization (multi-node only)
         if self.node_utils.len() > 1 {
-            let xnuma_tag = if self.xnuma_active != 0 {
-                " [xnuma]"
-            } else {
-                ""
-            };
-            let prefix = format!("  node util/load{xnuma_tag} ");
-            // N99=99999.9/99999.9 = 19 chars + 1 space = 20
-            let cell_width = 21;
+            let prefix = "  node    pin/ut/ld ";
+            // N99=999.9/999.9/99999.9 = 24 chars + 1 space = 25
+            let cell_width = 25;
             let usable = if max_width > prefix.len() {
                 max_width - prefix.len()
             } else {
@@ -494,6 +489,7 @@ impl LayerStats {
             for nid in 0..self.node_utils.len() {
                 let util = self.node_utils[nid];
                 let load = self.node_loads.get(nid).copied().unwrap_or(0.0);
+                let pin = self.node_pinned_utils.get(nid).copied().unwrap_or(0.0);
                 if nid % cells_per_row == 0 {
                     if nid > 0 {
                         writeln!(w)?;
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_layered/src/stats.rs
```diff
@@ -502,35 +498,7 @@ impl LayerStats {
...
-                let tasks = self.node_pinned_tasks.get(nid).copied().unwrap_or(0);
-                if nid % cells_per_row == 0 {
-                    if nid > 0 {
-                        writeln!(w)?;
-                    }
-                    write!(w, "{prefix}")?;
-                } else {
-                    write!(w, " ")?;
-                }
-                write!(w, "N{}={:7.1}/{:5}", nid, util, tasks)?;
```
xerothermic, line 533: While I haven't rely on this stats much, I feel having task counts per node will be useful to understand if the system is balanced.

htejun: Hmmm... does it matter how many tasks are there as long as utils are balanced?

PR #3403, opened 2026-03-08
scx_cosmos: Deprecate deferred CPU wakeups and increase default time slice

arighi:
Deferred CPU wakeups are only increasing power consumption without providing any significant benefit, so disable them by default and deprecate the option (`-d`).

Moreover, increase the default time slice to 1ms, which is a better compromise between throughput and latency.

@sirlucjan the only required change in scx_loader is drop the option `-d`. But we can also keep it for a while for backward compatibility (it's just ignored for now).


hodgesds (approved):
LGTM

PR #3408, opened 2026-03-09
scx_cosmos: Aggressively migrate tasks to the primary domain

arighi:
When a primary domain is defined (option -m), always attempt to migrate tasks to it. Avoid keeping tasks running from ops.dispatch() when they execute outside the primary domain, and instead always trigger a direct dispatch attempt from ops.enqueue().

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -1110,7 +1110,6 @@ s32 BPF_STRUCT_OPS(cosmos_select_cpu, struct task_struct *p, s32 prev_cpu, u64 w
 	return cpu >= 0 ? cpu : prev_cpu;
 }
 
-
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -1142,6 +1141,19 @@ void BPF_STRUCT_OPS(cosmos_tick, struct task_struct *p)
...
+ * Return true if @cpu is in the primary domain, false otherwise.
+ */
+static inline bool is_primary_cpu(s32 cpu)
+{
+	const struct cpumask *mask = cast_mask(primary_cpumask);
+
+	if (primary_all)
+		return true;
+
+	return mask && bpf_cpumask_test_cpu(cpu, mask);
```
etsal, line 1154: The && is to placate the verifier, right? This looks at first glance like a bug (&& instead of &) but I think is actually right.

PR #3415, opened 2026-03-10
scx_cosmos: PMU-based scheduling improvements

arighi:
Set of changes to enhance performance event scheduling in scx_cosmos:
- better logic for choosing idle CPUs when tasks are migrated due to perf events (limit CPU search within the primary domain and same LLC to avoid expensive migrations)
- fix potential starvation with sticky direct dispatches
- replace multiplicative scaling with EMA + hysteresis for smoother threshold adjustments
- support for human-readable perf event names like "cache-misses" in addition of raw hex codes

Examples:
- always migrate tasks, but try keeping the cache-intensive ones on the same CPU:
```
 $ sudo scx_cosmos -c 0 -y cache-references
```
- trigger migrations for tasks that are hitting too many branch mispredictions
```
 $ sudo scx_cosmos -e branch-misses
```

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -993,6 +993,19 @@ is_wake_affine(const struct task_struct *waker, const struct task_struct *wakee)
 }
 
 /*
+ * Return true if @cpu is in the primary domain, false otherwise.
+ */
+static inline bool is_primary_cpu(s32 cpu)
+{
+	const struct cpumask *mask = cast_mask(primary_cpumask);
+
+	if (primary_all)
+		return true;
+
+	return mask && bpf_cpumask_test_cpu(cpu, mask);
+}
+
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_cosmos/src/main.rs
```diff
@@ -54,13 +54,145 @@ use stats::Metrics;
...
+    event_id: u64,
+    /// perf_event_attr.type (PERF_TYPE_RAW, PERF_TYPE_HARDWARE, etc.).
+    type_: u32,
+    /// perf_event_attr.config.
+    config: u64,
+    /// Original string for error messages.
+    display_name: String,
+}
+
+fn parse_hardware_event(s: &str) -> Option<u64> {
```
hodgesds, line 71: might be worth putting some of these helper functions in `utils/src/perf.rs`

arighi: Ack! Will do.

PR #3425, opened 2026-03-13
scx_cosmos: Misc improvements

arighi:
A set of generic improvements for scx_cosmos:
 - Tune default settings for latency (since this is the most common use case for this scheduler)
 - Improve SMT contention avoidance (option -S)
 - Scale used time slice by CPU capacity (improve fairness in hybrid core systems)
 - Avoid migration attempts with per-CPU tasks (reduce unnecessary sched overhead)

### DIFF
--- a/scheds/rust/scx_cosmos/src/main.rs
+++ b/scheds/rust/scx_cosmos/src/main.rs
@@ -282,7 +282,7 @@ struct Opts {
     ///
     /// A higher value is recommended for server-type workloads, while a lower value is recommended
     /// for interactive-type workloads.
-    #[clap(short = 'c', long, default_value = "75")]
+    #[clap(short = 'c', long, default_value = "0")]
     cpu_busy_thresh: u64,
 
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -1060,6 +1075,7 @@ s32 BPF_STRUCT_OPS(cosmos_select_cpu, struct task_struct *p, s32 prev_cpu, u64 w
 {
 	const struct task_struct *current = (void *)bpf_get_current_task_btf();
 	struct task_ctx *tctx;
+	u64 now = scx_bpf_now();
```
hodgesds: might be able to move this after the `is_wake_affine` check for a little extra perf

arighi: Good point! Updated the code to make `scx_bpf_now()` more "local" and call it only when needed. Thanks.

PR #3432, opened 2026-03-16
ci: use vng-action vmlinux.h to fix BTF mismatch

likewhatevs:
## Summary
  - Checked-in vmlinux headers (v7.0-rc2) drift from the for-next kernel, causing `scx_bpf_cpu_rq` func_proto BTF incompatibility (`-EINVAL` on all scheduler loads)
  - Bump vng-action to v0.4.0 which generates and caches vmlinux.h from kernel BTF
  - Each job restores the kernel cache first (`run: true`), copies vmlinux.h to `/tmp` (outside `linux/` to survive subsequent cache restores), applies cpumask fixup, regenerates `vmlinux.tar.zst`,
  builds on host, then re-enters VM for workloads
  - Bump cache version to v18 to force rebuild with BTF enabled
  - Add `stress-ng` to `SCX_PACKAGES` for scheduler test workloads

  ## Test plan
  - [x] veristat jobs compile BPF against per-kernel vmlinux.h from cache
  - [x] test job compiles against for-next vmlinux.h before loading schedulers
  - [x] stress-ng available in VM for scheduler tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,9 +146,9 @@ jobs:
       fail-fast: false
       matrix:
         kernel:
-          - name: scx__sched_ext__for-next
-            url: https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git
-            tag: for-next
+          - name: scx__stable__linux-rolling-stable
+            url: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
+            tag: linux-rolling-stable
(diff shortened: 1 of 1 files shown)


likewhatevs:
https://github.com/sched-ext/scx/actions/runs/23127110298/job/67178810006  made me think PR was needed (i.e. that that failure was due to vmlinux.h/kernel signature mismatches) but this seems worse across the board.

PR #3434, opened 2026-03-16
scx_lavd: Restart scheduler when initially offline CPU comes online.

multics69:
Offline CPUs lack the `/sys/devices/system/cpu/cpuN/topology` directory (see commit 84a202e2a "topology: Skip offline CPUs"), so `scx_utils` skips them during topology initialization. As a result, their capacity is never populated and `cpuc->max_capacity` remains 0.

When such a CPU later comes online via hotplug, `lavd_cpu_online()` would proceed with stale/zeroed topology data, leading to incorrect scheduling decisions. Detect this case by checking `max_capacity == 0` and trigger a scheduler restart via `SCX_ECODE_ACT_RESTART` so the topology is properly re-initialized for the newly online CPU.


etsal:
IIUC this is indisttinguishable from any other scheduler exit, since even if we return SCX_ECODE_ACT_RESTART nobody actually reads the value. Is there a way to transparently restart the scheduler without requiring the userspace process to exit?

htejun:
> IIUC this is indisttinguishable from any other scheduler exit, since even if we return SCX_ECODE_ACT_RESTART nobody actually reads the value. Is there a way to transparently restart the scheduler without requiring the userspace process to exit?

This makes user_exit_info.should_restart() return true, which makes lavd main loop to restart instead of exit.

PR #3436, opened 2026-03-16
ci: reduce pipeline wall time from ~52m to ~5m on cache hit

likewhatevs:
## Summary

  - Run scheduler tests on the host kernel instead of booting a VM for
    each, keeping only arena selftests + cake + pandemonium in a VM
    (need newer kernel for BPF arena features). This works now because
    stable.
  - Split scheduler tests into parallel matrix groups so they run on
    separate runners simultaneously
  - Pre-generate cargo metadata on the host and pass it to veristat via
    --metadata-json, avoiding cargo metadata resolution over virtiofs 
    inside the VM
  - Replace vergen-gitcl with SCX_GIT_SHA env var so scx_utils doesn't
    invalidate every crate on each commit
  - Add sccache with GHA backend for cross-job rustc caching
  - Switch ci linker from lld to mold
  - Batch all scheduler builds into single cargo invocations
  - Mount workspace read-only in VMs so virtiofs uses cache=always
  - Gate BTF dumps on runner.debug
  - Bump actions/checkout to v6, vng-action to v0.5.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: ci
 
 on:
+  workflow_dispatch:
   pull_request:
   merge_group:
   push:


likewhatevs:
the last commit adds a button so that after pushing commits to sched_ext kernel.org repo (or any other we point things at), a run can be triggered which will run ci against that code.

builds will be ehh (cache miss times are ~15 mins atm), but once proper compute is hooked up, cold build times should be 1-2 minutes (i.e. push to upstream dev branch => click button (or have gh cli trigger ci run) => ci results in ~7 mins from push).

arighi (approved):

PR #3437, opened 2026-03-17
cargo: introduce workspace.dependencies

sirlucjan:
Many crates have same dependencies. If scheduler/tool doesn't require any features or it they all require the same one like `serde` we can use `workspace.dependencies` and leave everything else as is. This would make version upgrades much easier. I've checked `cargo build`, `cargo test` and `cargo check` and everything works. 

Checked on Arch Linux, Fedora and openSUSE.

### DIFF
--- a/rust/scx_arena/scx_arena/Cargo.toml
+++ b/rust/scx_arena/scx_arena/Cargo.toml
@@ -6,10 +6,10 @@ description = "Crate for setting up the BPF arena library for sched-ext schedule
 license = "GPL-2.0-only"
 
 [dependencies]
-anyhow = "1"
-libbpf-rs = "=0.26.1"
+anyhow.workspace = true
+libbpf-rs.workspace = true
 libbpf-sys = "=1.6.3"
-simplelog = "0.12"
+simplelog.workspace = true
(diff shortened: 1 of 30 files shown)


EricccTaiwan (approved):
Aside from some `vmlinux.h` warnings and a persistent `scx_cake` failure on ARM, everything looks good.

Tested on Ubuntu x86 and ARM without any issues.

sirlucjan:
> Aside from some `vmlinux.h` warnings and a persistent `scx_cake` failure on ARM, everything looks good.
> 
> Tested on Ubuntu x86 and ARM without any issues.

`vmlinux` warnings are related to latest CI fixes. `scx_cake` failure will be fixed in separated PR (which is accepted, only CI is still unhappy for some reasons...) Thanks!

likewhatevs:
~~I'm a fan of this -- for intentional version differences, folks can set those on their crates. This would simplify dependency management and help us need to compile less crates overall.~~

brought this up during office hours, i think it's better that crate versioning be more independent, like it is now. we have a script to help manage this, https://github.com/sched-ext/scx/blob/main/version-tool.py . I think I am going to open a PR later today ensuring that versions are unified across crates.

PR #3441, opened 2026-03-17
scx-service: add missing quote

sirlucjan:
Add missinig `'` to SCX_FLAGS.


arighi (approved):
Looks good now, thanks! :)

PR #3443, opened 2026-03-18
scx_cosmos: Add --no-early-clear

shelterx:
Try to implement --no-early-clear, it seems to help performance in some games and hinder performance in others.

Shadow of The tomb Raider
With early clear (cosmos default):
176 Avg FPS
Without  early clear:;
191 Avg FPS

Cyberpunk 2077
With early clear (cosmos default):
168 Avg FPS
Without  early clear:
161  Avg FPS

Note:
The use of AI in this MR is heavy, I've done what I can to clean it up.
Take it for what it is.

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -144,6 +144,11 @@ const volatile bool tick_preempt = true;
 const volatile bool no_wake_sync;
 
 /*
+ * Disable early clearing of idle CPU state.
+ */
+const volatile bool no_early_clear;
+
+/*
  * Default time slice.
  */
 const volatile u64 slice_ns = 1000000ULL;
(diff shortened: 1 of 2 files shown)


shelterx:
Some more numbers.
```
SottR - linux-cachyos
--no-early-clear
192 fps Avg

--flat-idle-scan
175 fps Avg

--flat-idle-scan -c 10 -p 100
178 fps Avg
```

arighi:
@shelterx thanks for the PR, looks interesting. I made some adjustments toyour patch and pushed it to https://github.com/sched-ext/scx/tree/cosmos-next (see https://github.com/sched-ext/scx/commit/3e7e7a7912444edc53c2963161caf53c859ca7dd).

I change the way to check if a CPU is idle and allowed to use "no-early-clear" also without forcing the flat or preferred idle scan (even if it's not fully replacing all the test_and_clear).

Can you check / test if it's still working for you? Thanks!

shelterx:
@arighi 
I did the SottR--no-early-clear test and got 192 FPS again, which matches previous results so I'd say it's fine.

PR #3445, opened 2026-03-19
sched: add scx_cognis scheduler

galpt:
This adds `scx_cognis`, a BPF-first `sched_ext` scheduler aimed at desktops, workstations, and general-purpose servers.

Cognis keeps the normal scheduling path in BPF and uses Rust as a control plane for loading, topology export, stats, restart handling, the optional TUI, and a narrow compatibility fallback when work intentionally crosses into userspace.

The current PR branch is synced to Cognis `v2.0.8`.

## Design

- BPF queue hierarchy: `CPU local DSQ -> LLC DSQ -> node DSQ -> shared DSQ`
- Deadline-oriented dispatch with bounded wake credit
- Remote balancing through cross-LLC and cross-node steals before the final shared spill path
- Two built-in profiles:
  - `desktop` for stronger wake responsiveness and locality retention
  - `server` for broader spill behavior under saturation

The common case is intended to avoid a Rust round-trip. Rust remains available for observability and compatibility fallback rather than owning the hot path.

## Integration

- Added as an in-tree Rust scheduler under `scheds/rust/scx_cognis`
- Builds against the current `sched-ext/scx` workspace and uses the repo's `scx_cargo`, `scx_utils`, `scx_stats`, and `scx_rustland_core` crates
- Follows the current compat path for `scx_bpf_select_cpu_and()`
- Wired into workspace metadata and CI package build coverage

## Validation

From an upstream-shaped checkout:

- `cargo fmt --all --check`
- `cargo check -p scx_cognis --locked`
- `cargo test -p scx_cognis --locked`
- `cargo build --profile ci --locked -p scx_cognis`
- `target/ci/scx_cognis --help`

`cargo test -p scx_cognis --locked` currently passes with `35/35` tests.

## Notes

- Cognis is BPF-first, but not BPF-only. It still uses a Rust companion process for control-plane work.
- The README intentionally does not claim universal production readiness across all kernels and hardware without workload-specific validation.

### DIFF
--- a/scheds/rust/scx_cognis/README.md
+++ b/scheds/rust/scx_cognis/README.md
@@ -30,6 +30,7 @@ Cognis uses a BPF queue hierarchy and tries to keep normal scheduling decisions 
 - default CLI mode: `desktop`
 - alternate CLI mode: `server`
 - watchdog-triggered `sched_ext` exits are treated as non-restartable failures so Cognis fails open to the kernel scheduler instead of immediately re-executing itself
+- the headless no-fallback path now backs off more aggressively and does not boost the Cognis userspace thread priority by default
 
 The BPF policy lives in [src/bpf/main.bpf.c](src/bpf/main.bpf.c). The Rust control plane lives in [src/main.rs](src/main.rs) and [src/bpf.rs](src/bpf.rs).
 
@@ -123,6 +124,7 @@ sudo ./target/release/scx_cognis --tui
 - The BPF side uses bounded queue domains and per-task local state.
 - `desktop` and `server` share the same hierarchy so the server profile remains a first-class mode rather than a neglected side path.
 - Watchdog / runnable-task-stall exits fail open to the kernel scheduler instead of being treated as restartable runtime exits.
+- In headless no-fallback periods, the Rust loop backs off instead of trying to run as an aggressively favored userspace thread.
 
 ## Limitations
 
--- a/scheds/rust/scx_cognis/src/main.rs
+++ b/scheds/rust/scx_cognis/src/main.rs
@@ -75,7 +75,7 @@ const DEFAULT_DESKTOP_SLICE_MIN_NS: u64 = 250_000;
 const DEFAULT_SERVER_SLICE_MIN_NS: u64 = 1_000_000;
 const DEFAULT_DESKTOP_SLICE_LAG_NS: u64 = 40_000_000;
 const DEFAULT_SERVER_SLICE_LAG_NS: u64 = 1_500_000;
-const IDLE_BACKOFF: Duration = Duration::from_micros(250);
+const IDLE_BACKOFF: Duration = Duration::from_millis(5);
 const RESTART_BACKOFF: Duration = Duration::from_millis(250);
 const RAPID_FAILURE_WINDOW: Duration = Duration::from_secs(30);
 const RAPID_FAILURE_LIMIT: u32 = 20;
@@ -1553,33 +1553,6 @@ impl Drop for Scheduler<'_> {
     }
 }
 
-fn elevate_scheduler_thread() {
-    // Keep the userspace scheduler responsive without turning it into a
-    // permanent real-time thread.
-    //
-    // The previous implementation promoted the main loop to SCHED_FIFO(1).
-    // That diverged from Andrea Righi's stable reference scheduler and was
-    // unsafe here because Cognis runs a mostly non-blocking control loop
-    // (schedule → housekeeping → optional TUI draw). Under SCHED_FIFO, Linux
-    // runs the thread until it blocks, is preempted by a higher-priority RT
-    // thread, or yields to an equal-priority peer; normal-priority kworkers on
-    // the same CPU can therefore be starved long enough to trip the sched_ext
-    // watchdog. See sched(7): https://man7.org/linux/man-pages/man7/sched.7.html
-    //
-    // A best-effort nice(-20) boost keeps the scheduler favored under CFS
-    // while still preserving fair progress for kernel workers and the rest of
-    // the system.
-    unsafe {
-        if libc::setpriority(libc::PRIO_PROCESS, 0, -20) != 0 {
-            warn!(
-                "Could not raise nice priority to -20 (errno {}); continuing \
-                 with default CFS priority",
-                *libc::__errno_location()
-            );
-        }
-    }
-}
-
 fn panic_payload_to_string(payload: &(dyn std::any::Any + Send)) -> String {
     if let Some(msg) = payload.downcast_ref::<&'static str>() {
         (*msg).to_string()
@@ -1704,8 +1677,6 @@ fn main() -> Result<()> {
         if shutdown.load(Ordering::Relaxed) {
             break;
         }
-
-        elevate_scheduler_thread();
 
         let loop_result = panic::catch_unwind(AssertUnwindSafe(|| -> Result<bool> {
             let mut sched = Scheduler::init(&opts, &mut open_object, shutdown.clone())?;


scheds/rust/scx_cognis/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1636 @@
...
+	return true;
+}
+
+/*
+ * Return the DSQ ID associated to a CPU, or SHARED_DSQ if the CPU is not
+ * valid.
+ */
+static u64 cpu_to_dsq(s32 cpu)
+{
+	if (!is_cpu_valid(cpu))
```
hodgesds, line 460: Does this make sense? In `is_cpu_valid` it is using `scx_bpf_error`, wouldn't that kick the scheduler?

scheds/rust/scx_cognis/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1636 @@
...
+/*
+ * Return true if @this_cpu and @that_cpu are in the same LLC, false
+ * otherwise.
+ */
+static inline bool cpus_share_cache(s32 this_cpu, s32 that_cpu)
+{
+        if (this_cpu == that_cpu)
+                return true;
+
+	if (!is_cpu_valid(this_cpu) || !is_cpu_valid(that_cpu))
```
hodgesds: nit: why does formatting change here?

scheds/rust/scx_cognis/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1636 @@
...
+static s32 pick_idle_cpu(struct task_struct *p, s32 prev_cpu, u64 wake_flags)
+{
+	s32 cpu, this_cpu = bpf_get_smp_processor_id();
+	bool is_this_cpu_allowed = bpf_cpumask_test_cpu(this_cpu, p->cpus_ptr);
+
+	if (no_wake_sync)
+		wake_flags &= ~SCX_WAKE_SYNC;
+
+	/*
+	 * For tasks that can run only on a single CPU, we can simply verify if
```
hodgesds, line 753: Tasks that run on a single CPU can change affinity at times, so this should maybe also check that the `prev_cpu` is still in the task's CPUs.

scheds/rust/scx_cognis/src/bpf/main.bpf.c
galpt: Cleaned up the formatting in this block and the matching helper next to it.

scheds/rust/scx_cognis/src/bpf/main.bpf.c
galpt: Good point. I changed `is_cpu_valid()` to treat a bad CPU id as a recoverable case and fall back with `dbg_msg()` here instead of tripping `scx_bpf_error()`.

scheds/rust/scx_cognis/src/bpf/main.bpf.c
galpt: Good catch. The single-CPU path now also checks that `prev_cpu` is still allowed in `p->cpus_ptr` before trying to reuse it.

galpt:
Addressed the startup UX point in b53059f4.

scx_cognis now logs a short startup line with the Cognis version, active mode, resolved slice settings, and enabled behavioral flags, and it also logs a matching "Registered Cognis scheduler" line once registration succeeds.

sirlucjan:
```
lucjan at cachyos ~ 18:46:57    
❯ sudo scx_cognis --mode desktop
18:47:00 [INFO] Starting Cognis v2.0.3 x86_64-unknown-linux-gnu (mode=desktop, slice=1000us, min_slice=250us, flags=none)
18:47:03 [INFO] Registered Cognis scheduler
^CEXIT: unregistered from user space
18:47:08 [INFO] Unregistered Cognis scheduler
lucjan at cachyos ~ 18:47:08    
❯ sudo scx_cognis --mode server 
18:47:12 [INFO] Starting Cognis v2.0.3 x86_64-unknown-linux-gnu (mode=server, slice=8000us, min_slice=1000us, flags=none)
18:47:12 [INFO] Registered Cognis scheduler
^CEXIT: unregistered from user space
18:47:20 [INFO] Unregistered Cognis scheduler
```

Thanks!

sirlucjan:
@galpt  Once it's been merged, take a look at this: https://github.com/sched-ext/scx-loader/commit/cdf354b10d5b42fe10a2802a30e331d064e5378a. Do you think this is a good solution, or would you suggest using different flags?

galpt:
Thanks. I think that mapping is mostly right for Cognis as it exists today.

`auto`, `gaming`, and `lowlatency` pointing to `--mode desktop` makes sense to me, and `server_mode -> --mode server` does too.

The only one I would treat a bit more carefully is `powersave_mode -> --mode desktop`. Cognis currently exposes `desktop` and `server`, but it does not have a separate power-tuned profile yet. So if loader needs every scheduler to fill every slot, falling back to `desktop` there is still a reasonable practical default. If it can be left unmapped until Cognis grows a real powersave profile, I think that would be the cleaner option.

sirlucjan:
> Thanks. I think that mapping is mostly right for Cognis as it exists today.
> 
> `auto`, `gaming`, and `lowlatency` pointing to `--mode desktop` makes sense to me, and `server_mode -> --mode server` does too.
> 
> The only one I would treat a bit more carefully is `powersave_mode -> --mode desktop`. Cognis currently exposes `desktop` and `server`, but it does not have a separate power-tuned profile yet. So if loader needs every scheduler to fill every slot, falling back to `desktop` there is still a reasonable practical default. If it can be left unmapped until Cognis grows a real powersave profile, I think that would be the cleaner option.

If we don't specify any parameters when we run `scx_cognis`, it will start in `--mode desktop` anyway, so I'll unmap powersave now. 

```
lucjan at cachyos ~ 11:53:19    
❯ sudo scx_cognis
11:53:21 [INFO] Starting Cognis v2.0.3 x86_64-unknown-linux-gnu (mode=desktop, slice=1000us, min_slice=250us, flags=none)
11:53:25 [INFO] Registered Cognis scheduler

```

Edit: Unfortunately, it’s not easy to unmap power-save mode, so either it will remain blank (which will bring up the desktop mode anyway) or I’ll wait for the next release until the power-saving profile will be added.

galpt:
@sirlucjan Thanks, that sounds reasonable to me.

From the Cognis side, `desktop` is still the default when `scx_cognis` is started without an explicit mode, and Cognis currently only exposes `desktop` and `server`. So if powersave ends up falling through to the default startup path for now, I think that is fine.

I just wouldn’t want to describe that as a distinct power-tuned Cognis profile yet, because we don’t actually have one today. If Cognis grows a real powersave-oriented mode later, we can wire that up explicitly at that point.

For now, keeping `server` as the explicit opt-in and letting the default path stay `desktop` makes sense to me.

sirlucjan:
@galpt 

I noticed something unusual. 

```
lucjan at cachyos ~ 17:07:48    
❯ cat /sys/kernel/sched_ext/root/ops     
cognis_2.0.3_x86_64_unknown_linux_gnu
```

cognis does not unregister after stopping, as the only scheduler.

galpt:
@sirlucjan Thanks, good catch. The previous shutdown path could log "Unregistered Cognis scheduler" after dropping struct_ops without verifying that `/sys/kernel/sched_ext/root/ops` had actually stopped reporting Cognis. I tightened that in `74823f99`: Cognis now polls `root/ops` briefly during shutdown and only logs the unregister message if detach is actually visible there. If `root/ops` still reports a Cognis instance after the timeout, it logs a warning instead of claiming success.

sirlucjan:
> @sirlucjan Thanks, good catch. The previous shutdown path could log "Unregistered Cognis scheduler" after dropping struct_ops without verifying that `/sys/kernel/sched_ext/root/ops` had actually stopped reporting Cognis. I tightened that in `74823f99`: Cognis now polls `root/ops` briefly during shutdown and only logs the unregister message if detach is actually visible there. If `root/ops` still reports a Cognis instance after the timeout, it logs a warning instead of claiming success.

Fixed, thanks.

sirlucjan:
@galpt I've benchmarked cognis (with stock kernel and scx_cake). 

[kaboom.log](https://github.com/user-attachments/files/26147517/kaboom.log)
kernel_version_comparison_All

Please take a look on `kaboom.log` because scx_cognis crashed once.

galpt:
@sirlucjan Thanks for attaching `kaboom.log`.

I checked it carefully. What the log clearly shows is a sched_ext watchdog exit (`runnable task stall`) rather than a Rust panic.

At the time of the stall, the dump shows all visible CPUs running `rt_sched_class` tasks named `Kizuna[...]`, while a number of normal sched_ext tasks were still runnable and had been waiting for a couple of seconds.

So from the log alone I can say that Cognis is not handling that condition robustly enough yet, but I don't think the log by itself is enough to pin the root cause down more narrowly than that.

Could you share what `Kizuna` is on that machine, and whether those threads are intentionally running as `SCHED_FIFO` / `SCHED_RR`?

Something like this would help:

```bash
ps -Leo pid,tid,cls,rtprio,pri,comm | grep Kizuna
```

If you have that context, I can check the scheduler side against it and work out whether this needs a Cognis policy change or a more explicit safety path under RT-heavy pressure.

sirlucjan:
```
lucjan at cachyos ~ 18:43:17    
❯ ps -Leo pid,tid,cls,rtprio,pri,comm | grep Kizuna
lucjan at cachyos ~ 18:43:20  
```
I benchmarked with:  https://github.com/CachyOS/cachyos-benchmarker

galpt:
@sirlucjan Thanks, that helps.

I checked `cachyos-benchmarker` on my side, and it doesn't look like the benchmark wrapper itself is explicitly setting `SCHED_FIFO` / `SCHED_RR` or spawning anything named `Kizuna`.

So the `ps ... | grep Kizuna` suggestion I gave was too narrow, and also easy to miss after the crash if those threads were transient. Sorry about that.

A broader snapshot of RT-class tasks during the benchmark would be much more useful. Something like this while the benchmark is running would help:

```bash
ps -eLo pid,tid,cls,rtprio,pri,comm | awk '$3 == "FF" || $3 == "RR" || $4 > 0'
```

Or, if you want to catch short-lived threads:

```bash
while true; do
  date
  ps -eLo pid,tid,cls,rtprio,pri,comm | awk '$3 == "FF" || $3 == "RR" || $4 > 0'
  echo
  sleep 1
done
```

That should tell us whether the watchdog case is happening under a broader RT-heavy condition on the machine, or whether there's something more specific we need to harden in Cognis itself.

sirlucjan:
I'll take a look next week, thanks!

galpt:
@sirlucjan Thanks for pointing me at `cachyos-benchmarker`.

I reproduced the same failure locally with that benchmark on my laptop. It was not `Kizuna` on my side, but the failure mode matched: a sched_ext watchdog stall under heavy load while RT-class threads were active, and Cognis was still treating that `EXIT:` path too much like something it should restart from.

I pushed [9e39307f](https://github.com/galpt/scx/commit/9e39307f4f2ea31d4f9e8a6d83c3f462aaedb2cc) to fix the part that is clearly ours:

- watchdog / runnable-task-stall `sched_ext` exits are now treated as non-restartable
- Cognis fails open to the kernel scheduler instead of re-execing itself back into the same bad state

So this is not a claim that the whole stall class is solved yet. It is a narrower fix: once the watchdog fires, Cognis now gets out of the way cleanly instead of trying to bounce straight back in.

If you get a chance to retest, that is the behavior change I would want checked first.

galpt:
@sirlucjan I did a direct pass against `scx_cake` after reproducing the stall locally.

The main difference was not the DSQ hierarchy. It was that `cake` keeps its steady-state path much quieter on the Rust side, while Cognis was still doing two things that made it a more aggressive userspace companion under load:

- polling the no-fallback headless path with a very short backoff
- raising the Cognis userspace thread to `nice -20` by default

I pushed [1c6e04ae](https://github.com/galpt/scx/commit/1c6e04ae3f32aa9114d1bc9496010862347469cf) to narrow that gap:

- removed the default userspace-thread priority boost
- increased the no-fallback idle backoff so the headless loop stays out of the way more when BPF is carrying the workload

This is still a targeted mitigation, not a claim that the whole watchdog-stall class is solved. But it is a cleaner next step than guessing at a bigger policy rewrite, and it is directly based on how `cake` behaves in the same area.

galpt:
Thanks again for the review and the detailed testing.

I spent more time reproducing this locally with `cachyos-benchmarker` and tried several grounded fixes on my side before deciding what to do next.

At this point, I do not think it is fair to keep this PR open while the current design still has an unresolved RT-heavy `sched_ext` watchdog stall under hostile load. I improved the fail-open behavior locally, but I have not eliminated the underlying stall in a way I would be comfortable asking reviewers to spend more time on.

Rather than keep this thread on a moving target, I am going to close this PR, rework the design locally, and come back with a cleaner submission once I can validate it more thoroughly first.

Thanks again for the time and feedback here.

multics69:
Look forward to seeing the improved version of scx_cognis!

Issue #3446, opened 2026-03-20
Null pointer dereference in `lavd_cgroup_move` due to missing return


Zhuocheng-Lang:
## Description
In `src/bpf/main.bpf.c`, the function `lavd_cgroup_move` checks if `taskc`, the task context, is NULL and logs an error using `scx_bpf_error`. However, it fails to return immediately after doing so and continues execution, leading to illegal access of a NULL pointer on the following line.

## Location
- **File**: `scheds/rust/scx_lavd/src/bpf/main.bpf.c`
- **Function**: `lavd_cgroup_move`
- **Lines**: 2174-2177

## Expected Behavior
In case `get_task_ctx(p)` returns NULL, it should log an error and immediately return, since it would lead to illegal access of a NULL pointer.

## Actual Behavior
It logs an error and then continues execution.
```c
	taskc = get_task_ctx(p);
	if (!taskc)
		scx_bpf_error("Failed to get a task context: %d", p->pid);
	taskc->cgrp_id = to->kn->id; // Dereferences NULL pointer here
```

## Suggested Fix
It should return after calling `scx_bpf_error`.

PR #3447, opened 2026-03-20
scx_lavd: Add missing return after scx_bpf_error() in lavd_cgroup_move()

EricccTaiwan:
Prevent null pointer dereference when get_task_ctx() returns NULL. Without the early return, execution continued and dereferenced taskc on the next line.

Fixes: #3446


multics69 (approved):
Thanks @EricccTaiwan !

PR #3448, opened 2026-03-21
scx_layered: Fix qrt_layer_id init to use MAX_LAYERS instead of MAX_LLCS

cuiweixie:
qrt_layer_id is a layer index, so its sentinel value should be MAX_LAYERS, not MAX_LLCS. The adjacent qrt_llc_id correctly uses MAX_LLCS.


cuiweixie:
please help to trigger the ci.

PR #3449, opened 2026-03-21
scx_layered: Remove duplicate taskc null check in match_one

cuiweixie:
In the MATCH_AVG_RUNTIME case, taskc is already null-checked right after lookup_task_ctx_may_fail(). The second identical check after the dereference is dead code. Remove it.


cuiweixie:
please help to trigger the ci.

PR #3450, opened 2026-03-21
scx_cosmos: Use high-precision time

arighi:
Replace the lightweight scx_bpf_now() with bpf_ktime_get_ns(), that provides a better precision when evaluating task vruntime.

This change seems to provide better results both in terms of throughput and system responsiveness. This shouldn't introduce too much overhead in modern systems, and, considering that scx_cosmos is mostly targetting newer systems, this change should be safe.


arighi:
> Nit: s/`bpf_ktime_get()`/`bpf_ktime_get_ns()` in the commit msg

Udpated, thanks!

PR #3454, opened 2026-03-23
lib/cgroup_bw: Fix task loss under cgroup CPU bandwidth throttling

multics69:
This series fixes task loss in `lib/cgroup_bw.bpf.c`, the shared BPF
library for cgroup CPU bandwidth control in sched_ext schedulers. A
throttled task could be permanently lost from all runqueues, causing a
sched_ext watchdog timeout. Fixing the root cause exposed two missing
implementations for related race scenarios (cgroup exit and task
migration), which are addressed as well. The series also includes
replenishment state machine simplification and a set of smaller bug
fixes found along the way.

#### Patch overview:

  - Patch 1: Reproducer
    Add `scripts/cg_race.sh`, a stress test that concurrently exercises
    task migration between cgroups, cgroup destruction with throttled
    tasks, and task exit while throttled. It monitors the SCX scheduler
    process and exits with an error if the scheduler stops due to a
    watchdog-detected task stall.

  - Patches 2-3: Fix task loss — exiting tasks permanently stuck in the BTQ
    In `do_exit()`, a task is removed from the PID map by
    `__unhash_process()` while still preemptable and before `TASK_DEAD`
    is set. If preempted here with its cgroup throttled, the task enters
    the BTQ with `SCX_TASK_IN_CUSTODY` set. The BTQ drain calls
    `bpf_task_from_pid()`, which returns `NULL` for the unhashed task;
    with no way to reenqueue it, the task is permanently lost, causing a
    sched_ext watchdog timeout. So, we skip throttling in
    `scx_cgroup_bw_throttled()` when `PF_EXITING` is set. As a cleanup,
    `bpf_task_from_pid()` is moved inside `scx_cgroup_bw_enqueue_cb()`
    so the library resolves task pointers centrally.

  - Patches 10-12: Missing implementation for cgroup exit and task migration
    While fixing task loss, two related scenarios were found to lack
    proper handling. For task migration, moving a throttled task between
    cgroups left it in the source BTQ with bandwidth charged to the
    wrong cgroup; `scx_cgroup_bw_move()` is added to cancel the task
    from the source and re-insert it into the destination BTQ. For
    cgroup exit, a cgroup can exit while exiting tasks remain throttled
    under it (the kernel does not count `TASK_DEAD` tasks as living); on
    exit, `cbw_unthrottle_cgroup_for_exit()` sets the budget to infinity and
    `cbw_free_llc_ctx()` drains the BTQ by migrating remaining tasks to
    the root cgroup with `vtime=0` for fast reaping. The cgroup exit
    path also introduces a use-after-free: `cbw_drain_btq_batch()`
    snapshots `llcx->btq` with `READ_ONCE()` and passes the snapshot to
    `scx_atq_pop()`, while `cbw_free_llc_ctx()` can null and destroy the
    BTQ concurrently. This is fixed by deferring `scx_atq_destroy()` via
    `schedule_atq_destroy()`, which uses a 256-slot circular array with
    atomic tail allocation and CAS-based slot eviction to ensure
    destruction is deferred until all in-flight drain operations on that
    BTQ has been completed.

  - Patches 4-9: Replenishment simplification and bug fixes
    The replenishment state machine is redesigned to use a single atomic
    `u64` (`union backlog_stat`) packing a sequence counter, throttled
    cgroup count, and `has_throttled_tasks` flag, reducing the 4-state
    machine to 2 states and closing races between the top and bottom
    halves. Four bugs are also fixed: a read-after-write on
    `is_throttled` that could leave re-throttled tasks stuck in the BTQ;
    a missing `smp_mb()` that could cause spurious re-throttling on
    non-TSO architectures; two `-EAGAIN` paths that failed to set
    `is_throttled`; and a BTQ drain loop that stopped at the first
    exhausted LLC instead of trying others with the remaining budget.

 - Patch 13: Workaround for the BPF verifier bug.

---
  - Patch  1: scripts: Add cg_race.sh — cgroup CPU bandwidth race stress test
  - Patch  2: lib/cgroup_bw: Never throttle an exiting task to prevent permanent task loss
  - Patch  3: lib/cgroup_bw: Move bpf_task_from_pid() into scx_cgroup_bw_enqueue_cb()
  - Patch  4: lib/cgroup_bw: Redesign replenish state tracking with backlog_stat
  - Patch  5: lib/cgroup_bw: Fix read-after-write bug in cbw_replenish_taskable_cgroup
  - Patch  6: lib/cgroup_bw: Fix memory ordering in cbw_replenish_taskable_cgroup
  - Patch  7: lib/cgroup_bw: Set is_throttled in two missing paths of cbw_cgroup_bw_throttled
  - Patch  8: lib/cgroup_bw: Try all LLC contexts when reenqueueing throttled tasks
  - Patch  9: lib/cgroup_bw: Do not rely on is_throttled to stop BTQ drain
  - Patch 10: lib/cgroup_bw: Handle throttled task relocation on cgroup_move
  - Patch 11: lib/cgroup_bw: Handle exiting throttled tasks on cgroup exit
  - Patch 12: lib/cgroup_bw: Defer BTQ destruction to avoid use-after-free
  - Patch 13: lib/cgroup_bw: Mark cbw_calc_budget_tx() as static __noinline

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -487,6 +487,16 @@ macro_rules! scx_ops_load {
             use ::anyhow::Context;
             use ::libbpf_rs::skel::OpenSkel;
 
+            {
+                let ops = $skel.struct_ops.[<$ops _mut>]();
+                if ops.sub_cgroup_id > 0 {
+                    if let Ok(false) | Err(_) = scx_utils::compat::struct_has_field("sched_ext_ops", "sub_cgroup_id") {
+                        ::scx_utils::warn!("kernel doesn't support ops.sub_cgroup_id");
+                        ops.sub_cgroup_id = 0;
+                    }
+                }
+            }
+
             scx_utils::uei_set_size!($skel, $ops, $uei);
             $skel.load().context("Failed to load BPF program")
         }
--- a/scheds/include/scx/compat.h
+++ b/scheds/include/scx/compat.h
@@ -160,7 +160,7 @@ static inline long scx_hotplug_seq(void)
  * COMPAT:
  * - v6.17: ops.cgroup_set_bandwidth()
  * - v6.19: ops.cgroup_set_idle()
- * - v7.1:  ops.sub_attach(), ops.sub_detach()
+ * - v7.1:  ops.sub_attach(), ops.sub_detach(), ops.sub_cgroup_id
  */
 #define SCX_OPS_OPEN(__ops_name, __scx_name) ({					\
 	struct __scx_name *__skel;						\
@@ -191,6 +191,11 @@ static inline long scx_hotplug_seq(void)
 	    !__COMPAT_struct_has_field("sched_ext_ops", "sub_detach")) {	\
 		fprintf(stderr, "WARNING: kernel doesn't support ops.sub_detach()\n"); \
 		__skel->struct_ops.__ops_name->sub_detach = NULL;		\
+	}									\
+	if (__skel->struct_ops.__ops_name->sub_cgroup_id > 1 &&		\
+	    !__COMPAT_struct_has_field("sched_ext_ops", "sub_cgroup_id")) { \
+		fprintf(stderr, "WARNING: kernel doesn't support ops.sub_cgroup_id\n"); \
+		__skel->struct_ops.__ops_name->sub_cgroup_id = 1;		\
 	}									\
 	__skel; 								\
 })


multics69:
Added "lib/cgroup_bw: Mark cbw_calc_budget_tx() as static __noinline" to work around the BPF verifier bug that occurred at TJ's for-next tree.

daidavid (approved):

Issue #3435, opened 2026-03-16
Warnings after latest CI fixes (main)


sirlucjan:
The latest updates have introduced quite a few warnings. 

[scx-scheds-git-1.1.0.r110.g9a6501a1-1-x86_64-build.log](https://github.com/user-attachments/files/26034952/scx-scheds-git-1.1.0.r110.g9a6501a1-1-x86_64-build.log)

PR #3456, opened 2026-03-25
Fix build warnings

frelon:
These commits fix current build warnings when running `cargo build`.

Most warnings seem to be variations of the following:

```
warning: scx_arena_selftests@1.1.0: /home/frelon/src/scx/target/debug/build/scx_arena_selftests-246eeaf3f478df20/out/scx_utils-bpf_h/arch/x86/vmlinux.h:109744:4: warning: declaration does not declare anything [-Wmissing-declarations]
warning: scx_arena_selftests@1.1.0:  109744 |                         struct freelist_counters;
warning: scx_arena_selftests@1.1.0:         |                         ^~~~~~~~~~~~~~~~~~~~~~~~
warning: scx_arena_selftests@1.1.0: 6 warnings generated.
```

Which can be ignored using the `#pragma clang diagnostic ...` in `common.bpf.h`.

Also update includes in scxcash to fix the warnings.

Fixes: #3435

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -14,12 +14,17 @@
  */
 #define BPF_NO_KFUNC_PROTOTYPES
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wmissing-declarations"
+
 #ifdef LSP
 #define __bpf__
 #include "../vmlinux.h"
 #else
 #include "vmlinux.h"
 #endif
+
+#pragma clang diagnostic pop
 
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_tracing.h>


rust/scx_rustland_core/Cargo.toml
```diff
@@ -18,11 +18,6 @@ scx_cargo = { path = "../scx_cargo", version = "1.1.0" }
 [lib]
 name = "scx_rustland_core"
 path = "src/lib.rs"
-include = [
-    "assets/bpf/intf.h",
-    "assets/bpf/main.bpf.c",
-    "assets/bpf.rs",
-]
```
sirlucjan, line 25: That’s actually needed

frelon: Okay, I will drop the commit! Out of curiosity, what breaks when removing this?

sirlucjan: ~~This is required to publish the crate~~  From what @arighi said, it’s simply necessary for crates to work

PR #3458, opened 2026-03-25
mitosis: add dispatch fallback for lockless peek miss

likewhatevs:
## Summary

Tasks stall for 2+ seconds on cell DSQs while their assigned CPUs sit idle, triggering the watchdog. The lockless `scx_bpf_dsq_peek` reads `dsq->first_task` via `rcu_dereference`, which can return NULL while the DSQ list is non-empty — `dispatch_enqueue` calls `list_add` before `rcu_assign_pointer` sets `first_task`, so a concurrent peek can see the stale NULL. Without a fallback, `mitosis_dispatch` returns empty on a NULL peek, dispatch returns empty, and since kicks only fire from enqueue paths (not dispatch), no one wakes it to retry. Also, how we use peek is racy, so use move_to_local to get a lock in there/avoid racing cpu's missing work (i.e. losing move_to_local race after peek).

## Fix

Always fall through to `scx_bpf_dsq_move_to_local()` regardless of peek results. `move_to_local` uses `consume_dispatch_q()` which takes `dsq->lock` and walks the actual list, finding tasks the lockless peek missed. When peek does find a winner, try `move_to_local` on the winner first, then fall back to the other DSQ if the winner was consumed by a racing CPU. Work stealing is moved after the fallback so it only triggers when both DSQs are genuinely empty.

## Test plan

- [x] stt `dispatch_contention` scenario: without fix, 18 tasks stuck on cell DSQ while 4 CPUs idle (reproduces with `nohz=off`), watchdog fires at 2.7s — with fix, passes cleanly


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -855,54 +855,89 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+	 * IMPORTANT: We must always fall back to trying move_to_local
+	 * even when peek returns NULL. There is a race between enqueue
+	 * and dispatch where peek can miss tasks:
+	 *
+	 * 1. CPU A runs ops.enqueue for task T destined for DSQ X.
+	 *    scx_bpf_dsq_insert_vtime() calls mark_direct_dispatch()
+	 *    which sets ddsp_dsq_id but does NOT insert T on DSQ X.
+	 *
+	 * 2. CPU A calls scx_bpf_kick_cpu(B, SCX_KICK_IDLE) to wake
+	 *    CPU B which serves DSQ X. The kick is async (irq_work).
```
dschatzberg: This ordering isn't clear to me - how can we scx_bpf_dsq_insert_vtime() but not insert onto the DSQ and then subsequently kick a cpu?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -855,54 +855,89 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+	 *
+	 * The fallback move_to_local calls use consume_dispatch_q()
+	 * which takes dsq->lock and walks the actual list. By the time
+	 * this locked path runs (nanoseconds later, or on the next
+	 * scheduling event), dispatch_enqueue has completed and the
+	 * task is visible under the lock. The lockless list_empty
+	 * fast-path in consume_dispatch_q (ext.c:1897) may also see
+	 * the list update since list_add happens before first_task
+	 * is set via rcu_assign_pointer, and list_empty reads list.next
+	 * which is updated earlier in the insertion sequence.
```
dschatzberg: What does the dsq lock synchronize with to prevent operation 3 from happening afterwards?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: I think the fix here is right, but the cause is wrong, let me keep digging.

likewhatevs: reproduced this issue w/ the other fixes all applied confirming the need for this fix/that it fixes something and updated the comments/pr w/ my current understanding of the issue.

figuring out the other comment (if there is a need for lock and adding it if so) next.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: applying this question to idk my updated understanding of what this PR is fixing -- in this pr, we fallback to a sychronized path when the lockless one says no work, because sometimes that is going to say no work when there is work.

likewhatevs:
@tommy-u @dforsyth when you get a chance, would this be a flag one or a not flag one? not sure if this falls into obvious fix territory (i think seeing a trace w/ it is necessary to know independent of test stuff).

likewhatevs:
closing in favor of https://github.com/sched-ext/scx/pull/3489

PR #3459, opened 2026-03-25
mitosis: guard against vtime underflow in stopping

likewhatevs:
## Summary

`scx_bpf_now()` is per-CPU (`this_rq()`) and not monotonic across CPUs. In `mitosis_stopping`, if `now < started_running_at` due to cross-CPU clock skew, the unsigned subtraction wraps to ~`U64_MAX`, corrupting vtime for the entire cell.

## Fix

Use `time_delta()` from `common.bpf.h` to clamp negative deltas to zero. Add `CSTAT_CLAMP_USED` counter to track how often this fires in the wild.


htejun:
Does this actually happen? Even across CPUs, if strict ordering can be defined, the clock values shouldn't go backward. cc @multics69

likewhatevs:
> Does this actually happen? 

Wildly, I think so. I can't reproduce this issue on demand to test this PR, but pouring over data, the math worked out and the dumps aligned with this being the cause of this issue (and per the code, I don't think anything else could have been the cause).

dschatzberg:
Could you actually also bump some CSTAT so we can measure if we're seeing this in the wild?

likewhatevs:
> I'd be surprised if we're seeing this in practice, but may as well guard against it

I will concede this is amongst the wildest things I have seen but it is eerie how the math works out like, just right, for this to be the case and the dump logs for this situation align with idk this.

Will add the stat.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1358,8 +1358,21 @@ void BPF_STRUCT_OPS(mitosis_stopping, struct task_struct *p, bool runnable)
...
 		return;
 
-	now			 = scx_bpf_now();
-	used			 = now - tctx->started_running_at;
+	now  = scx_bpf_now();
+	used = now - tctx->started_running_at;
+	/*
+	 * Guard against clock going backward. During scx_ops_enable(),
+	 * the kernel's task iterator calls running()/stopping() on the
+	 * iterator's CPU for tasks current on remote CPUs. scx_bpf_now()
```
multics69: Could you elaborate a bit more about the scx_ops_enable() path? I am not able to find that symbol either in the SCX and kernel code.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1358,8 +1358,21 @@ void BPF_STRUCT_OPS(mitosis_stopping, struct task_struct *p, bool runnable)
...
+	 * uses this_rq(), so started_running_at is set with the iterator
+	 * CPU's clock. When the task later stops on its own CPU, the
+	 * subtraction underflows if that CPU's TSC is behind. The
+	 * underflow wraps to ~U64_MAX, and (U64_MAX * 100 / weight)
+	 * produces a corrupted vtime of ~U64_MAX/100 that poisons the
+	 * cell's vtime domain.
+	 */
+	if (unlikely((s64)used < 0))
+		used = 0;
```
multics69: The recommended practice for time calculation is to use `time_*()` functions defined in common.bpf.h. In this case, you can use `time_delta(u64 after, u64 before)`.

multics69:
> Does this actually happen? Even across CPUs, if strict ordering can be defined, the clock values shouldn't go backward. cc @multics69

In my understanding, the scenario that this PR handles is:

- CPU X calls task's `ops.running()` ; `scx_bpf_now()` inside `ops.running()` calculated using CPU X's `rq clock`.
- Then CPU Y calls the task's `ops.stopping()`; `scx_bpf_now()` inside `ops.stopping()` calculated using CPU Y's `rq clock`.
- If CPU Y's clock is behind CPU X's clock, the delta between two clocks becomes negative. 

First of all, I am not sure in what cases `ops.running()` and `ops.stopping()` can run on two different CPUs. Also, the clock skew between CPU X and Y should be greater than the task's runtime. That seems unlikly to happen even though it is not impossible.

@likewhatevs -- Is my understanding correct?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: Thank you, done!

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: ahh my bad, thanks.

PR #3460, opened 2026-03-25
mitosis: use SCX_DSQ_LOCAL_ON in try_pick_idle_cpu

likewhatevs:
## Summary

`try_pick_idle_cpu` dispatches to `SCX_DSQ_LOCAL` after finding an idle CPU, but `SCX_DSQ_LOCAL` resolves to `task_rq(p)` — the CPU the task last ran on, not the idle CPU that was found. When called from the `put_prev_task_scx` -> enqueue path, the task goes back onto its current CPU's local DSQ while the idle CPU gets kicked and finds nothing.

This causes per-CPU DSQ starvation: the local DSQ on the current CPU always has a task, so `balance_one` (ext.c:2193) skips `ops.dispatch` entirely. Tasks on per-CPU DSQs (pinned kworkers) are never considered and stall indefinitely.

## Fix

Use `SCX_DSQ_LOCAL_ON | cpu` instead of `SCX_DSQ_LOCAL` in both the cell-local and borrowing paths of `try_pick_idle_cpu`. `SCX_DSQ_LOCAL_ON` resolves to the specified CPU's local DSQ regardless of `task_rq(p)`.

## Test plan

- [x] Verified pinned kworkers no longer stall on per-CPU DSQs during scheduling under load

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -648,15 +648,12 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p,
 	if (cpu >= 0) {
 		cstat_inc(CSTAT_LOCAL, tctx->cell, cctx);
 		/*
-		 * Use SCX_DSQ_LOCAL_ON to target the idle CPU we found,
-		 * not SCX_DSQ_LOCAL which resolves to task_rq(p) -- the
-		 * CPU the task just ran on. Without this, when called
-		 * from put_prev_task_scx -> enqueue, the task goes back
-		 * to the current CPU's local DSQ while the idle CPU gets
-		 * kicked and finds nothing. This pins tasks to their
-		 * current CPU, starving per-CPU DSQ tasks (kworkers)
-		 * because ops.dispatch is never called when the local
-		 * DSQ is non-empty.
+		 * Use SCX_DSQ_LOCAL_ON to explicitly target the idle CPU
+		 * we found. In the select_cpu path this is redundant
+		 * (SCX_DSQ_LOCAL already resolves to the selected CPU),
+		 * but from the enqueue path (put_prev_task_scx ->
+		 * enqueue), SCX_DSQ_LOCAL resolves to task_rq(p) -- not
+		 * the idle CPU we picked.
 		 */
 		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | cpu, slice_ns, 0);
 		if (kick)


htejun:
`SCX_DSQ_LOCAL` from `ops.select_cpu()` means the returned selected CPU, not the CPU that `ops.select_cpu()` is running on. Have you confirmed this by reproducing or is this just AI agent spotting a "clear" issue? It's okay to use AI agents to write up PRs but you gotta make sure that the clear issues it found aren't hallucinations. Don't stop at letting it write the code changes, force it to reproduce the problem and verify the fix by comparing before and after.

likewhatevs:
> SCX_DSQ_LOCAL from ops.select_cpu() means the returned selected CPU, not the CPU that ops.select_cpu() is running on. Have you confirmed this by reproducing or is this just AI agent spotting a "clear" issue? It's okay to use AI agents to write up PRs but you gotta make sure that the clear issues it found aren't hallucinations. Don't stop at letting it write the code changes, force it to reproduce the problem and verify the fix by comparing before and after.

Thanks, will do!

I'll PR this once sorting out issues, but more/less, we've been on a bug hunt with mitosis and I wrote (generated, idk) a test framework to help identify all the issues (some have been exceedingly hard to debug thus far):
 https://github.com/likewhatevs/scx/tree/stt-debug-framework

I've been running ~1200 tests of different scenarios against mitosis with a scheduler kick timeout of 2s of stall and using that to help identify issues. This particular change moved a lot of the tests in the following manner consistently (so, idk, but i think there's something to it):

```
● ┌──────────────────────┬───────────────┬───────────────┬───────────────┐
  │       Scenario       │ Peek-fix only │ +DSQ_LOCAL_ON │     Delta     │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ io_borrowing         │ 0             │ 12            │ +12           │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ cpuset_disjoint      │ 0             │ 10            │ +10           │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ cpuset_change        │ 3             │ 12            │ +9            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ rebalancing_dynamic  │ 6             │ 12            │ +6            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ rebalancing          │ 6             │ 12            │ +6            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ dynamic_borrowing    │ 6             │ 12            │ +6            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ borrowing_contention │ 4             │ 10            │ +6            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ no_ctrl_rebalancing  │ 7             │ 12            │ +5            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ nested_borrowing     │ 7             │ 12            │ +5            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ exclude_rebalancing  │ 8             │ 12            │ +4            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ exclude_borrowing    │ 8             │ 12            │ +4            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ cpuset_add           │ 8             │ 12            │ +4            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ borrowing_idle       │ 4             │ 8             │ +4            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ borrowing_bursty     │ 8             │ 12            │ +4            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ affinity_random      │ 1             │ 5             │ +4            │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ + 25 more scenarios  │               │               │ +1 to +3 each │
  ├──────────────────────┼───────────────┼───────────────┼───────────────┤
  │ Total                │ 619           │ 741           │ +122          │
  └──────────────────────┴───────────────┴───────────────┴───────────────┘
```

htejun:
Ah okay, you wrote that this is from enqueue path. Yes, from enqueue(), you would need `LOCAL_ON` and the returned CPU doesn't do anything. Sorry about the noise.

dschatzberg:
Thanks, this looks like a legit bug and I had been hunting down issues in this part of the code. This unconditionally changes the code to use LOCAL_ON - is that appropriate also in select_cpu or not? I wonder if you need to have a flag to use one vs the other

likewhatevs:
> Thanks, this looks like a legit bug and I had been hunting down issues in this part of the code. This unconditionally changes the code to use LOCAL_ON - is that appropriate also in select_cpu or not? I wonder if you need to have a flag to use one vs the other

This is redundant/no-op in select_cpu so I would say this is appropriate? I don't see why we should use one vs the other because in select_cpu SCX_DSQ_LOCAL_ON | cpu == SCX_DSQ_LOCAL.

PR #3453, opened 2026-03-23
sched: add scx_timely scheduler

galpt:
This adds `scx_timely`, a `sched_ext` scheduler built on top of `scx_bpfland` and aimed at adapting TIMELY's delay-driven feedback ideas to CPU scheduling.

## Design

- Keeps the inherited `bpfland` fast path largely intact
- Uses queue delay as the main feedback signal for a narrow Timely-style control layer
- Exposes explicit `Tlow` / `Thigh` delay regions
- Uses additive increase, multiplicative decrease, and HAI-style faster recovery
- Ships `desktop`, `powersave`, and `server` presets over the same controller structure

## Integration

- Added as an in-tree Rust scheduler under `scheds/rust/scx_timely`
- Wired into the workspace and Rust schedulers index
- Uses the in-tree `scx_cargo`, `scx_utils`, `scx_stats`, and `scx_stats_derive` crates
- Includes a short in-tree README and a veristat baseline

## Validation

From an upstream-shaped checkout in a no-space build path:

- `cargo fmt --all --check`
- `cargo check -p scx_timely`
- `cargo build --profile ci -p scx_timely`
- `target/ci/scx_timely --help`

## Notes

- `scx_timely` is still experimental and the README says so directly
- This PR keeps the top-level in-tree docs short; the standalone repo keeps the extra local benchmark helpers and operational notes

PR #3463, opened 2026-03-26
scx_timely v2: Pressure-aware load-balancing with expand/contract mode

galpt:
## Summary

scx_timely v2 introduces pressure-aware load-balancing, a Swift-inspired evolution of the TIMELY scheduler where sustained delay pressure drives scheduler-level policy shifts.

### Core Changes

- **Expand/Contract Mode**: Scheduler switches between locality-first and balance-first behavior based on system-wide delay pressure
- **Global Pressure Tracking**: EMA of delay-pressured tasks drives policy decisions  
- **Hysteresis**: Prevents oscillation between modes with mode-specific thresholds (desktop: 75%/50%, powersave: 65%/40%, server: 80%/55%)

### Design Rationale

Per the TIMELY paper, delay is the main control signal. Swift strengthens this into cleaner policy shifts:
- **Contract (low pressure)**: Locality-first, favor favored CPU set
- **Expand (sustained pressure)**: Balance-first, spread work to reduce queue delay
- Wake-heavy tasks (audio/interactive) preserve locality even in expand mode

### References

1. `sched-ext` maintainers. *scx_bpfland* [Software]. https://github.com/sched-ext/scx/tree/main/scheds/rust/scx_bpfland
2. Mittal, R., Lam, V. T., Dukkipati, N., et al. (2015). *TIMELY: RTT-based congestion control for the datacenter.* https://research.google.com/pubs/timely-rtt-based-congestion-control-for-the-datacenter/
3. Kabbani, A., et al. (2020). *Swift: Delay is Simple and Effective for Congestion Control in the Datacenter.* https://research.google.com/pubs/swift-delay-is-simple-and-effective-for-congestion-control-in-the-datacenter/
4. Ousterhout, A., et al. (2019). *Shenango: Achieving High CPU Efficiency for Latency-sensitive Datacenter Workloads.* https://www.usenix.org/conference/nsdi19/presentation/ousterhout

### Validation

- `cargo build` passes
- `cargo fmt --check` passes  
- `cargo check` passes
- Benchmarks show competitive performance vs bpfland/cake

### Notes

- Version bumped to 2.0.0 (Swift is successor to TIMELY paper)
- This is a separate branch from main v1.1.0


galpt:
For easier code review, all changes are documented line-by-line at:
https://github.com/galpt/scx_timely/blob/timely-v2/docs/design-vs-bpfland.md

scheds/rust/scx_timely/Cargo.lock
sirlucjan, line 1: You shouldn't add extra Cargo.lock file for your scheduler.

sirlucjan (approved):

arighi:
Sorry, I'm a bit late at reviewing all these PRs... Since most of this code is shared with scx_bpfland (just a different time slice evaluation and now a load balancer), would it make sense to just merge the code into bpfland and provide options to enable/disable the scx_timely behavior?

galpt:
@arighi thanks for the review and the suggestion, really appreciate it!

I'm totally fine with merging timely behavior into bpfland if that's what you prefer. That said, since the original bpfland is already considered production ready, I'd also like to keep the original code around as a stable base that people can fork and build other schedulers from — similar to how this one was built on top of it.

I'm also keeping in mind what @htejun [mentioned before](https://github.com/sched-ext/scx/pull/3453#pullrequestreview-3989837138) about potentially separating stable and experimental schedulers in the future. I think keeping them as separate schedulers for now aligns with that direction too.

But you're the maintainer here, so if you think merging into bpfland is cleaner for the repo long-term, I can definitely work on making that merge as clean as possible. What do you think?

arighi:
@galpt thank you for contributing. If the changes are not too big, I think it's worth to include them directly in bpfland, there's really no reason to maintain a separate scheduler for such small changes. And if we guard them behind command line options there's no regression potential for production use cases. So my vote is to merge these changes directly into bpfland with command line options to enable/disable them (if possible). If it's not possible or if the changes become too big, then we can definitely fork off.

@sirlucjan with wd40 it was a bit different scenario, we basically re-implemented big parts of rusty using BPF arenas, that was a quite disruptive change, so in that case I think it made sense to create a separate scheduler.

sirlucjan:
I’ll ask a different question @galpt  will there be further updates, modifications, and new features? If you plan to make more and more changes over time, it might be a good idea to keep this. But @arighi's suggestion makes a lot of sense.

arighi:
One thing doesn't exclude the other. We can merge these changes in bpfland now, keep timely alive and in sync as an experimental playground (up to @galpt) and, again, if things diverge too much, continue their development as separate schedulers.

I also like @htejun idea of having an experimental/staging folder to add newer schedulers. It's really good to have all the schedulers in the same repo, but at the same time I think we need to contain a bit the profileration of new schedulers (and I'm the one always introducing new ones BTW). But it's expensive in terms of testing/CI/build time/maintenance/packaging, so maybe having a separate folder for the new experimental scheds is a good idea.

galpt:
@arighi @sirlucjan thank you both for the feedback, that makes total sense!

I was actually planning to go much deeper into the literature ([Caladan](https://www.usenix.org/conference/osdi20/presentation/fried), [PerfIso](https://www.usenix.org/conference/atc18/presentation/iorgulescu), [Shinjuku](https://www.usenix.org/conference/nsdi19/presentation/kaffes), etc.), but from the current v2 to those papers it would require rewriting parts of bpfland and the changes would become unrecognizable — so I held back and went with the current v2 instead to keep the diff small.

I agree with @arighi's suggestion. I'll merge the timely features into bpfland behind a `--timely` flag or similar, and save the more experimental stuff for a different scheduler later.

@sirlucjan to answer your question — yes, I do plan to keep experimenting with more ideas, but I'll do that in a separate scheduler so bpfland stays clean and focused. Does that work for everyone?

sirlucjan:
@galpt Yes, I'm OK with your proposal.

arighi:
> @arighi @sirlucjan thank you both for the feedback, that makes total sense!
> 
> I was actually planning to go much deeper into the literature ([Caladan](https://www.usenix.org/conference/osdi20/presentation/fried), [PerfIso](https://www.usenix.org/conference/atc18/presentation/iorgulescu), [Shinjuku](https://www.usenix.org/conference/nsdi19/presentation/kaffes), etc.), but from the current v2 to those papers it would require rewriting parts of bpfland and the changes would become unrecognizable — so I held back and went with the current v2 instead to keep the diff small.

Cool, then yes, makes sense to have a separate scheduler. But for now I think it'd be great to have these changes in bpfland as well.

> 
> I agree with @arighi's suggestion. I'll merge the timely features into bpfland behind a `--timely` flag or similar, and save the more experimental stuff for a different scheduler later.

Awesome, thanks!

> 
> @sirlucjan to answer your question — yes, I do plan to keep experimenting with more ideas, but I'll do that in a separate scheduler so bpfland stays clean and focused. Does that work for everyone?

Sounds like a good me to me.

galpt:
As discussed, I've implemented the TIMELY v2 changes in `scx_bpfland` behind a `--timely` flag. The new PR is here: https://github.com/sched-ext/scx/pull/3472

The implementation adds:
- Delay-based time slice adaptation using TIMELY's gain algorithm
- Locality fallback for wake-heavy tasks
- Pressure-aware expand/contract load balancing mode

All TIMELY v2 behavior is guarded behind `timely_enabled`, so the default bpfland behavior is unchanged when `--timely` is not set.

PR #3464, opened 2026-03-26
mitosis: avoid cross-cell CPU-local vtime contamination

dforsyth:
We initialize a task's dsq_vtime from the vtime domain it was enqueued into, but in userspace-managed cell mode, apply_cell_config() can retag a CPU to a different cell while a task is still running on it. mitosis_stopping() would then copy that old-domain dsq_vtime into the retagged CPU's local vtime, tainting the destination cell's local timeline.

This change will record the CPU cell at the start of each running slice and only advance cctx->vtime_now in mitosis_stopping() when the CPU stayed in the same cell for the full slice.

I think that this means we'll over compensate on the per-cpu dsq for a while, but that is probably fine vs the current vtime issue we have after a rebalance.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -584,6 +584,7 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p, s32 prev_cpu
 		 * enqueue), SCX_DSQ_LOCAL resolves to task_rq(p) -- not
 		 * the idle CPU we picked.
 		 */
+		tctx->vtime_charge_cell = tctx->cell;
 		scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | cpu, slice_ns, 0);
 		if (kick)
 			scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
@@ -607,6 +608,7 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p, s32 prev_cpu
 		if (cpu >= 0) {
 			tctx->borrowed = true;
 			cstat_inc(CSTAT_BORROWED, tctx->cell, cctx);
+			tctx->vtime_charge_cell = tctx->cell;
 			scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | cpu, slice_ns, 0);
 			if (kick)
 				scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
@@ -703,8 +705,10 @@ s32 BPF_STRUCT_OPS(mitosis_select_cpu, struct task_struct *p, s32 prev_cpu, u64 
 		if (cpu < 0)
 			return prev_cpu;
 
-		if (idle_cpu_cleared || scx_bpf_test_and_clear_cpu_idle(cpu))
+		if (idle_cpu_cleared || scx_bpf_test_and_clear_cpu_idle(cpu)) {
+			tctx->vtime_charge_cell = tctx->cell;
 			scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL, slice_ns, 0);
+		}
 		return cpu;
 	}
 
@@ -770,6 +774,19 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
 
 	if (maybe_refresh_cell(p, tctx) < 0)
 		return;
+
+	/*
+	 * CPU -> cell mappings can change between enqueue() and stopping().
+	 * If that happens, the task's dsq_vtime may no longer belong to the
+	 * CPU-local or shared cell vtime domains visible at stopping(), and
+	 * advancing either one would charge the wrong domain.
+	 * Direct local insert paths snapshot the same state before inserting.
+	 *
+	 * Snapshot the cell whose vtime domain this placement expects to
+	 * charge. stopping() only advances local and cell vtime if the task
+	 * is not borrowed and the CPU it stops on is still in this same cell.
+	 */
+	tctx->vtime_charge_cell = tctx->cell;
 
 	/* Ensure this is done *AFTER* refreshing cell which might manipulate vtime */
 	vtime = p->scx.dsq_vtime;
@@ -1331,6 +1348,7 @@ void BPF_STRUCT_OPS(mitosis_running, struct task_struct *p)
 			return;
 	}
 
+	/* Record the running slice start time. */
 	tctx->started_running_at = scx_bpf_now();
 
 	/* Shrink our slice if a pinned task is queued on this CPU's DSQ. */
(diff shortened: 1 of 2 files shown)


likewhatevs:
if this fixes it i'm all for it.

that being said, i added an (admittedly, rather extreme) test here to try and trigger this misbehavior and this does not pass it:
https://github.com/sched-ext/scx/pull/3467/changes/1a38dc96e9c873932879c2f63eb6986161f4651e (vtime_contamination), so it could be that the test is a superset of real occurrences of this issue *or* this fix does fix this issue all places we need it fixed.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1383,27 +1385,36 @@ void BPF_STRUCT_OPS(mitosis_stopping, struct task_struct *p, bool runnable)
...
-	 * advancing cctx->vtime_now is correct and prevents staleness.
+	 * A task's dsq_vtime belongs to the domain it was enqueued into.
+	 * tctx->cell != cidx only tells us that the task's nominal cell and
+	 * the CPU's current cell differ at stopping time. That can happen if
+	 * userspace retagged the CPU while the task was running, or if pinned
+	 * cross-cell work (for example a root-cell kworker) is running on a
+	 * CPU that currently belongs to another cell. Neither case means the
+	 * running slice belongs to cidx's vtime domain, so blindly copying its
+	 * dsq_vtime into cctx->vtime_now can taint the destination cell's
+	 * local timeline.
```
dschatzberg: Very nice! This looks like a legit bug here in the case where userspace retags the CPU. As you say in the comment, we had it in the vtime domain of a different cell, updating the cpu's vtime after it moves cells would be incorrect and taint the destination cell's timeline. In a sense, this really just looks like we borrowed the task, so protecting here makes sense.

The comment regarding pinned work is a bit confusing - we do normally want pinned tasks to contribute to the per-cpu vtimeline (and therefore its cell as well) - this is how we compare pinned tasks vs non-pinned tasks on the same vtimeline. This is precisely the way you've implemented it as well so I think it's just that the comment is confusing.

I think a clearer way to reason about this issue is that in apply_cell_config we reset the per-cpu vtimeline (https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c#L2284-L2310) and any application of vtimes from a prior domain to the current domain would be tainted. That's consistent with what you're fixing here.

I think this fix is probably sufficient, but the one thing I couldn't clearly reason about is the case where cpu transitions from A -> B -> A in quick succession - if we started a task before the first transition and it stops after the last transition, we'd update the per-cpu vtime. I think that's probably legit since we always captured vtime within the same domain but might be worth double checking

dschatzberg (approved):

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1383,27 +1385,36 @@ void BPF_STRUCT_OPS(mitosis_stopping, struct task_struct *p, bool runnable)
...
-	} else {
-		vtime_cell = cell;
-	}
-	advance_cell_llc_vtime(vtime_cell, tctx, p->scx.dsq_vtime);
+	/*
+	 * Keep the shared cell cursor in the same domain dispatch compares
+	 * against on this CPU. Cross-cell pinned work can otherwise advance
+	 * the CPU-local cursor without advancing this CPU's current cell DSQ.
+	 */
+	advance_cell_llc_vtime(cell, tctx, p->scx.dsq_vtime);
```
dschatzberg: Huh, I can't wrap my head around this change. The only functional change here is for tasks that are pinned outside their cell - previously we would update the task's cell's vtime but now we update the cpu's cell's vtime. It's not obvious to me now which is correct but given the existing code, I clearly believed the task's cell was the correct one to update. The comment you have here makes no sense to me - can you elaborate?

dschatzberg (dismissed):

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg: @dforsyth - wouldn't this obviously be wrong in the borrowing case? We're clearly at that time polluting the vtime of the wrong domain

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1396,33 +1404,29 @@ void BPF_STRUCT_OPS(mitosis_stopping, struct task_struct *p, bool runnable)
...
-	 * the vtime was initialized in this CPU's cell domain, so
-	 * advancing cctx->vtime_now is correct and prevents staleness.
+	 * Only advance this CPU's local vtime when the slice ends on a CPU
+	 * whose cell matches the cell this task expected at enqueue time and
+	 * the task was not borrowed. If execution ends in some other cell,
+	 * drop the local charge rather than risk charging an unexpected
+	 * domain.
 	 */
-	if (!tctx->borrowed) {
+	if (!tctx->borrowed && tctx->enqueue_cpu_cell == cidx) {
```
likewhatevs: Cstat for vtime drop here?

likewhatevs (approved):
lgtm

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -788,6 +788,13 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
 	if (maybe_refresh_cell(p, tctx) < 0)
 		return;
 
+	/*
+	 * Snapshot the CPU cell this enqueue expects to run in. stopping()
+	 * only advances local and shared vtime when the slice ends in this
+	 * same cell and the task was not borrowed.
+	 */
```
tommy-u, line 788: This comment is too confusing.

Could this work:
CPU -> cell mapping can change between a task's enqueue() and stopping() times. When this happens, we do not want to advance the CPU or cell DSQ vtimes because (reason). We snapshot the cell the task was associated with during enqueue() and reference it in stopping(). We only advance DSQ vtimes in stopping() if the CPU it stops on still matches (there was no CPU retag).

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1396,33 +1404,29 @@ void BPF_STRUCT_OPS(mitosis_stopping, struct task_struct *p, bool runnable)
...
+	if (!tctx->borrowed && tctx->enqueue_cpu_cell == cidx) {
 		if (time_before(READ_ONCE(cctx->vtime_now), p->scx.dsq_vtime))
 			WRITE_ONCE(cctx->vtime_now, p->scx.dsq_vtime);
 	}
 
+	/*
+	 * Only advance the shared cell vtime when the slice ends in the same
+	 * cell this task expected at enqueue time and the task was not
+	 * borrowed.
+	 */
```
tommy-u, line 1419: This is correct, but don't we want to emphasize that the CPU was not retagged to a different cell between enqueue() and stopping()?

Nit: "slice ends" -> "task stops"
Nit: Delete "shared"

tommy-u (approved):
This looks good. Can you take another pass on the comments so they're a bit clearer for a noob.

tommy-u (changes requested):
Sorry for the re-review. My understanding is this will miss direct dispatched tasks which only go through select and not enqueue. I think the best option is to stamp the tasks in running().

tommy-u (approved):
Looks good.  We'll get a counter in here later.

PR #3468, opened 2026-03-27
mitosis: prevent cross-cell vtime contamination in stopping

likewhatevs:
## Summary

`mitosis_stopping` advances `cctx->vtime_now` from `p->scx.dsq_vtime`, but cross-cell tasks (root cell kworkers on user cell CPUs, borrowed tasks, tasks mid-retag) have `dsq_vtime` from a foreign vtime domain. Writing it to `cctx->vtime_now` contaminates the cell's per-CPU vtime, which then propagates to the cell DSQ via `apply_cell_config`'s vtime sync.

The existing `!tctx->borrowed` guard only covers one case. Any task whose `tctx->cell` differs from the CPU's cell has foreign-domain vtime.

## Fix

Add `tctx->cell == cidx` to the vtime advancement guard, covering all cross-cell cases — borrowed tasks, root cell kworkers, and mid-retag tasks.

PR #3469, opened 2026-03-27
mitosis: preempt on tick and check borrow fairness

likewhatevs:
## Summary
- Add ops.tick handler that preempts the running task when per-CPU DSQ
  work is waiting, or when a cross-cell borrower occupies a CPU whose
  native cell has queued work.
- Before borrowing a CPU, check that its native cell and per-CPU DSQs
  are empty — skip the borrow if the target cell needs the CPU.
- Add raw u64 DSQ ID helpers for use in contexts where BPF cannot
  return aggregates.

Together with #3468, this gets the vtime contamination test in #3467
to pass.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -677,6 +677,21 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p,
 		}
 		cpu = pick_idle_cpu_from(p, borrowable, prev_cpu, idle_smtmask);
 		if (cpu >= 0) {
+			/*
+			 * Check if the target CPU's native cell has
+			 * waiting work before borrowing. If so, skip
+			 * — the native cell needs this CPU.
+			 */
+			struct cpu_ctx *target_cctx = lookup_cpu_ctx(cpu);
+			if (target_cctx) {
+				u32 tllc = enable_llc_awareness ?
+						   target_cctx->llc :
+						   FAKE_FLAT_CELL_LLC;
+				if (scx_bpf_dsq_nr_queued(cell_llc_dsq_raw(
+					    target_cctx->cell, tllc)) > 0 ||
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -677,6 +677,21 @@ static __always_inline s32 try_pick_idle_cpu(struct task_struct *p,
...
+			struct cpu_ctx *target_cctx = lookup_cpu_ctx(cpu);
+			if (target_cctx) {
+				u32 tllc = enable_llc_awareness ?
+						   target_cctx->llc :
+						   FAKE_FLAT_CELL_LLC;
+				if (scx_bpf_dsq_nr_queued(cell_llc_dsq_raw(
+					    target_cctx->cell, tllc)) > 0 ||
+				    scx_bpf_dsq_nr_queued(cpu_dsq_raw(cpu)) > 0)
+					goto no_borrow;
+			}
```
dschatzberg, line 694: This doesn't make a ton of sense to me - it seems like papering over some other bug if we're finding idle CPUs in a cell with work in its DSQ

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1803,6 +1819,39 @@ static void dump_cell_cpumask(int id)
...
+		}
+	}
+	if (tctx->cell != cctx->cell) {
+		u32 llc = enable_llc_awareness ? cctx->llc : FAKE_FLAT_CELL_LLC;
+		u64 cldsq = cell_llc_dsq_raw(cctx->cell, llc);
+		if (scx_bpf_dsq_nr_queued(cldsq) > 0) {
+			p->scx.slice = 0;
+		}
+	}
+}
```
dschatzberg, line 1853: I think I rather make this choice at select_cpu or enqueue time rather than on a periodic tick. That being said - this just seems like a fairness change - maybe let's wait on this until we've resolved more obvious correctness issues?

dschatzberg (changes requested):

PR #3471, opened 2026-03-27
mitosis: fix yielder starvation via vtime clamp split

likewhatevs:
Edit: something about pr is off -- stt needs to be able to trigger dumps when things go sideways from userspace independent of watchdog to sort this out.

## Summary

The backward vtime clamp in `mitosis_enqueue` (`basis_vtime - slice_ns`) gives tasks up to one slice of credit below the current cell clock. For tasks waking from sleep, this is correct — they shouldn't be penalized for idling. But for tasks re-enqueued via yield or slice expiry, the clamp resets their vtime to `basis_vtime - slice_ns` on every re-enqueue, giving them perpetually lowest vtime in the PRIQ. Yielding tasks monopolize dispatch and starve all other tasks in the cell.

On oversubscribed cells (e.g. 16 CpuSpin workers on 2 cpuset-constrained CPUs with a YieldHeavy worker), the yielder's vtime is always one slice below the cell clock while CpuSpin workers are at or above it. The yielder wins every dispatch, runs for microseconds, yields, re-enqueues with the clamp resetting its vtime, and wins again. CpuSpin workers stall for 5+ seconds.

## Fix

Split the clamp by enqueue reason:
 - **Wakeups** (`SCX_ENQ_WAKEUP`): clamp to `basis_vtime - slice_ns` (one slice of credit, as before)
 - **Re-enqueues** (yield, slice expiry): clamp to `basis_vtime` (no credit — vtime reflects actual CPU usage)

## Test plan

 - [x] `rebalancing_cpuset/rebal` on odd-3llc (6 CPUs): 0/10 → 10/10 pass (16 CpuSpin + 1 YieldHeavy on 2-CPU cpuset cell)
 - [x] `dispatch_contention`, `oversubscribed`, `stall_detect`, `proportional`, `sched_fifo`, `sched_rr`, `cpuset_aligned`, `cpuset_misaligned`, `rebalancing` — all pass
 - [x] All rebalancing variants pass: `rebalancing_dynamic`, `rebalancing_many`, `rebalancing_oscillate`, `rebalancing_shift`

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -826,11 +826,20 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
 		return;
 	}
 	/*
-	 * Limit the amount of budget that an idling task can accumulate
-	 * to one slice.
+	 * Clamp vtime so tasks don't accumulate too much credit.
+	 * Wakeups get up to one slice of credit (they were sleeping).
+	 * Re-enqueues (yield, slice expiry) get no credit — their
+	 * vtime reflects actual CPU usage. Without this distinction,
+	 * yielders perpetually win the PRIQ by resetting to
+	 * basis_vtime - slice_ns on every re-enqueue.
 	 */
-	if (time_before(vtime, basis_vtime - slice_ns))
-		vtime = basis_vtime - slice_ns;
+	if (enq_flags & SCX_ENQ_WAKEUP) {
+		if (time_before(vtime, basis_vtime - slice_ns))
+			vtime = basis_vtime - slice_ns;
+	} else {
+		if (time_before(vtime, basis_vtime))
+			vtime = basis_vtime;
+	}
 
 	scx_bpf_dsq_insert_vtime(p, tctx->dsq.raw, slice_ns, vtime, enq_flags);
 


dschatzberg (changes requested):
So even though this is pretty egregious userspace behavior, I do see how a perpetually yielding tasks could muck things up. I wonder if it would be more accurate to clamp a perpetually yielding task to the latest vtime we've enqueued - basically insert it in the back of the current set of waiters. That seems to be more appropriate for yielding

likewhatevs:
> So even though this is pretty egregious userspace behavior, I do see how a perpetually yielding tasks could muck things up. I wonder if it would be more accurate to clamp a perpetually yielding task to the latest vtime we've enqueued - basically insert it in the back of the current set of waiters. That seems to be more appropriate for yielding

I think this is actually less accurate/doesn't work in that it penalizes yielder's too heavily, and that is what the test results below show for the following impl of that (i.e. it fails the opposite way):
```
  - /*
  -  * Limit the amount of budget that an idling task can accumulate
  -  * to one slice.
  -  */
  - if (time_before(vtime, basis_vtime - slice_ns))
  -     vtime = basis_vtime - slice_ns;
  + /*
  +  * Clamp vtime so tasks don't accumulate too much credit.
  +  * Wakeups get up to one slice of credit (they were sleeping).
  +  * Re-enqueues (yield, slice expiry): clamp to the DSQ tail
  +  * vtime so the task goes behind current waiters.
  +  */
  + if (enq_flags & SCX_ENQ_WAKEUP) {
  +     if (time_before(vtime, basis_vtime - slice_ns))
  +         vtime = basis_vtime - slice_ns;
  + } else {
  +     struct task_struct *tail;
  +     u64 tail_vtime = basis_vtime;
  +
  +     bpf_for_each(scx_dsq, tail, tctx->dsq.raw, SCX_DSQ_ITER_REV) {
  +         tail_vtime = tail->scx.dsq_vtime;
  +         break;
  +     }
  +     if (time_before(vtime, tail_vtime))
  +         vtime = tail_vtime;
  + }
```

#### pass fail -- multiple runs

```
  ┌────────┬───────┬───────┬───────┬───────────┐
  │        │ Run 1 │ Run 2 │ Run 3 │ Pass rate │
  ├────────┼───────┼───────┼───────┼───────────┤
  │ MAIN   │ FAIL  │ FAIL  │ PASS  │ 1/3       │
  ├────────┼───────┼───────┼───────┼───────────┤
  │ PR     │ PASS  │ PASS  │ PASS  │ 3/3       │
  ├────────┼───────┼───────┼───────┼───────────┤
  │ ALT v2 │ PASS  │ FAIL  │ PASS  │ 2/3       │
  └────────┴───────┴───────┴───────┴───────────┘
```

PR #3472, opened 2026-03-27
scx_bpfland: add TIMELY v2 mode behind --timely flag

galpt:
## Summary

Add TIMELY's delay-driven feedback for adaptive time slices and pressure-aware load balancing as an optional mode in `scx_bpfland`.

When `--timely` is enabled:
- **Delay-based time slice adaptation** using TIMELY's gain algorithm
- **Locality fallback** for wake-heavy tasks to reduce lock contention
- **Pressure-aware expand/contract load balancing** mode (locality-first when healthy, balance-first under pressure)

All TIMELY v2 behavior is guarded behind the `timely_enabled` flag, so the default behavior remains unchanged when `--timely` is not set.

### Design Rationale

This implements the changes requested in PR #3463, merging TIMELY v2 behavior into `scx_bpfland` behind a `--timely` flag, as suggested by @arighi.

Per the TIMELY paper, delay is the main control signal. Swift (TIMELY v2) strengthens this into cleaner policy shifts:
- **Contract (low pressure)**: Locality-first, favor favored CPU set
- **Expand (sustained pressure)**: Balance-first, spread work to reduce queue delay

### Changes

- `scheds/rust/scx_bpfland/src/main.rs`: Add `--timely` flag and TIMELY v2 command-line options
- `scheds/rust/scx_bpfland/src/bpf/main.bpf.c`: Add TIMELY v2 BPF implementation with dispatch wrappers
- `scheds/rust/scx_bpfland/src/stats.rs`: Add TIMELY v2 statistics

### References

- TIMELY paper: https://research.google.com/pubs/timely-rtt-based-congestion-control-for-the-datacenter/
- Swift paper: https://research.google.com/pubs/swift-delay-is-simple-and-effective-for-congestion-control-in-the-datacenter/
- Original scx_timely v2 PR: https://github.com/sched-ext/scx/pull/3463

PR #3473, opened 2026-03-27
scx_timely: drop from tree

sirlucjan:
scx_timely will be merged into scx_bfpland, https://github.com/sched-ext/scx/pull/3472, and https://github.com/sched-ext/scx/pull/3463, so we can remove the standalone version from the main branch.


sirlucjan:
Let's wait until @galpt  submits PRs to merge scx_timely into scx_bpfland, and once they've been merged, I'll mark my commit as “ready for review”

PR #3477, opened 2026-03-28
lib/pelt: Add BPF PELT library with configurable half-life

EricccTaiwan:
## Background
                                                   
Initially this feature was implemented in kernel/sched/ext.c via a new ops field (ops.pelt_lshift = N), letting the kernel track a parallel sched_avg per CPU with a configurable half-life (see [PATCH](https://hackmd.io/@ericisgood/pelt-half-life-patch)).

After discussing with @arighi, we agreed to move the implementation to BPF instead for two reasons:   

1. Adding a dependency between ext.c and pelt.c makes the kernel side harder to maintain. 
2. The PELT math can be fully reimplemented in BPF without any kernel changes. 
                                      
This follows the same pattern as lib/ravg.bpf.c. 

## Summary

This design is heavily inspired by [`[RFC PATCH 0/1] sched/pelt: Change PELT halflife at runtime`](https://lore.kernel.org/all/20220829055450.1703092-1-dietmar.eggemann@arm.com/).

Re-implement the kernel's Per-Entity Load Tracking (PELT) math entirely in BPF. No kernel kfuncs are required. 

The standard kernel PELT uses a fixed 32ms half-life. This library  supports shorter half-lives via time-stretching: the elapsed delta is left-shifted before being fed into the accumulator, making time appear  to pass faster and the signal converge more quickly:
                                                                                           
    lshift = 0  ->  1x stretch ->  32ms half-life (standard) 
    lshift = 1  ->  2x stretch ->  16ms half-life
    lshift = 2  ->  4x stretch ->   8ms half-life 
         
BPF schedulers add lib/pelt.bpf.c as a source, zero-initialize a per-CPU struct pelt_rq, set lshift once at init, and call pelt_update() from ops.running(), ops.stopping(), and ops.tick().

CC: @jserv

### DIFF
--- a/lib/pelt.bpf.c
+++ b/lib/pelt.bpf.c
@@ -1,0 +1,157 @@
+/*
+ * SPDX-License-Identifier: GPL-2.0
+ * Author: Cheng-Yang Chou <yphbchou0911@gmail.com>
+ */
+
+/*
+ * This design is heavily inspired by `[RFC PATCH 0/1] sched/pelt: Change PELT
+ * halflife at runtime` [1].
+ * [1]: https://lore.kernel.org/all/20220829055450.1703092-1-dietmar.eggemann@arm.com/
+ */
+
+#include <scx/common.bpf.h>
+#include <lib/pelt.h>
+
+/*
+ * pelt_do_accumulate_sum - accumulate elapsed time into *_sum fields
+ * @delta: elapsed time in 1024us units, already stretched by lshift
+ * @prq: PELT state to update
+ * @load: load weight (0 if task is not on rq)
+ * @runnable: runnable weight (0 if not runnable)
+ * @running: 1 if the CPU is executing a task, 0 otherwise
+ *
+ * Returns the number of full 1024us periods crossed, or 0 if no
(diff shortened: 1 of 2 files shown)


EricccTaiwan:
I think adding a test for pelt is redundant, so I removed it.

arighi:
It would be nice to show if there are practical benefits, like use this from a scheduler and show some numbers.

EricccTaiwan:
> It would be nice to show if there are practical benefits, like use this from a scheduler and show some numbers.

Ok, I'll do some tests.

multics69:
`ravg` is the BPF implementation of PELT, isn't it?

EricccTaiwan:
> `ravg` is the BPF implementation of PELT, isn't it?

I think they're slightly different.

The main distinction I see is that ravg depends on BPF arena, while PELT avoids that dependency so schedulers without arena support (like scx_bpfland) can use it. PELT also specifically tracks running/runnable/load states to match the kernel's struct sched_avg semantics, which ravg doesn't cover.

That said, I'm totally fine if you think this is redundant and prefer not to add another library.

PR #3478, opened 2026-03-28
scx_flash: Re-modernize scheduler

arighi:
Remove compatibility code to support old kernels, completely rely on the built-in idle CPU selection policy, drop NUMA rebalancing (allow natural rebalance on task wakeup), introduce cgroup weight support, evaluate the task deadline as a function of vruntime and wakeup frequency (as it was originally designed), handle sleep/wakeup intensive tasks by directly dispatching them to the local DSQ (reduce node DSQ locking pressure) and remove deprecated options.

@sirlucjan this change removes some old options, so it breaks the scx_loader profiles. We can just use the default options for all profiles with scx_flash for now, we'll figure out the right options later.

### DIFF
--- a/scheds/rust/scx_flash/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_flash/src/bpf/main.bpf.c
@@ -386,7 +386,7 @@ static inline u64 task_slice(const struct task_struct *p)
  * that sleep frequently accumulate less runtime and they also have a
  * larger budget).
  */
-static u64 task_dl(struct task_struct *p, int node, struct task_ctx *tctx, u64 enq_flags)
+static u64 task_dl(struct task_struct *p, struct task_ctx *tctx, u64 enq_flags)
 {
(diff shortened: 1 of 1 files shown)


hodgesds:
Might need  to update the CI to skip 6.13 tests as it looks like it's no longer verifiable.

arighi:
I think I'll probably remove flash from CI completely, so that we can break things. At the end it's supposed to be an experimental version of scx_bpfland, so personally I'm ok to skip CI for it entirely.

Edit: actually I think I can fix this one, it should be pretty easy. And while at it I may update this PR with more commits.

arighi:
@sirlucjan FYI, this is going to break all the options and profiles for scx_flash, sorry. :) If we merge this we can reset everything and use the default for all profiles for now. We'll figure out some profiles later.

sirlucjan:
I'll update scx_loader tomorrow. Thanks for info

PR #3479, opened 2026-03-29
scx_lavd: Faster load balancing with budget-controlled invariant runtime

bboymimi:
This series is a first draft for discussion. It replaces LAVD's
one-task-per-steal load balancer with budget-controlled batch migration
using invariant runtime as the fundamental element to improve load
distribution accuracy and establish the foundation for future selective
migration. We can iterate on top of this. The patches are organized in
four parts.

### Part 1: Instrumentation

  f4295f7edd22 scx_lavd: Track system-wide average invariant runtime
  5acd23286228 scx_lavd: Track per-domain queued invariant load

Track system-wide average invariant runtime and per-domain queued
invariant load. queued_load_invr is the sum of each queued task's
avg_runtime_invr, maintained incrementally via atomic add/sub across all
enqueue and dequeue paths. This replaces the task-count approximation
with actual task-size-aware load information.

### Part 2: Budget framework

  1ea86ccb4f5b scx_lavd: Add migration budget for stealee domains
  280172ac4706 scx_lavd: Use actual queued task sizes instead of task count
  82ed9c672c5f scx_lavd: Add migration budget for stealer domains

Introduce symmetric migration budgets for both stealee and stealer
domains. Each stealee budget is half the domain's excess queued load
above its capacity-proportional fair share. Each stealer budget is half
its deficit below fair share. Replace the task-count-based load equation
with actual queued task sizes in load_invr to improve stealer/stealee
classification accuracy.

### Part 3: Budget enforcement

  c3cabcd7a2e9 scx_lavd: Budget-controlled task stealing in try_to_steal_task()
  2667d85c1f88 scx_lavd: Decrement budget on force steal for consistent accounting
  8677592e67bd scx_lavd: Remove flag clearing from migrate_to_neighbor()

Enforce budgets in try_to_steal_task() and force_to_steal_task() by
peeking at the DSQ head for actual task size before consuming. Flags are
cleared only when budget is exhausted. Remove flag clearing from the
donation path since donated tasks were never queued in the source
domain.

### Part 4: Migration delta and cleanup

  870cf5a463ee scx_lavd: Capacity-proportional migration delta
  43da9aa72337 scx_lavd: Fix budget decrement unit mismatch with load_invr
  97e1b9b15c64 scx_lavd: Remove probabilistic entry gate from try_to_steal_task()

Replace the flat system-wide threshold with capacity-proportional fair
share targets so heterogeneous domains are correctly classified. Align
budget decrement units with the load_invr space used for budget
computation. Remove the probabilistic entry gate from
try_to_steal_task(), which is now gated by budget control.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -174,9 +174,21 @@ int plan_x_cpdom_migration(void)
 
 		/*
 		 * Under-loaded active domains become a stealer.
+		 * Ingress budget = half the deficit below fair share.
 		 */
 		if (cpdomc->nr_active_cpus &&
 		    cpdomc->load_invr <= stealer_threshold) {
+			u64 stealer_budget = 0;
+
+			if (total_cap_sum > 0) {
+				u64 fair_share = total_load_invr *
+					cpdomc->cap_sum_active_cpus /
+					total_cap_sum;
+				if (fair_share > cpdomc->load_invr)
+					stealer_budget = (fair_share -
+						cpdomc->load_invr) / 2;
+			}
+			WRITE_ONCE(cpdomc->stealer_budget_invr, stealer_budget);
 			WRITE_ONCE(cpdomc->stealee_budget_invr, 0);
 			WRITE_ONCE(cpdomc->is_stealer, true);
 			WRITE_ONCE(cpdomc->is_stealee, false);
@@ -208,6 +220,7 @@ int plan_x_cpdom_migration(void)
 			}
 
 			WRITE_ONCE(cpdomc->stealee_budget_invr, budget);
+			WRITE_ONCE(cpdomc->stealer_budget_invr, 0);
 			WRITE_ONCE(cpdomc->is_stealer, false);
 			WRITE_ONCE(cpdomc->is_stealee, true);
 			nr_stealee++;
@@ -218,6 +231,7 @@ int plan_x_cpdom_migration(void)
 		 * Otherwise, keep tasks as it is.
 		 */
 		WRITE_ONCE(cpdomc->stealee_budget_invr, 0);
+		WRITE_ONCE(cpdomc->stealer_budget_invr, 0);
 		WRITE_ONCE(cpdomc->is_stealer, false);
 		WRITE_ONCE(cpdomc->is_stealee, false);
 	}
@@ -238,6 +252,7 @@ reset_and_skip_lb:
 
 			cpdomc = MEMBER_VPTR(cpdom_ctxs, [cpdom_id]);
 			WRITE_ONCE(cpdomc->stealee_budget_invr, 0);
+			WRITE_ONCE(cpdomc->stealer_budget_invr, 0);
 			WRITE_ONCE(cpdomc->is_stealer, false);
 			WRITE_ONCE(cpdomc->is_stealee, false);
 		}
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -18,6 +18,7 @@
 
 
 extern const volatile u8	mig_delta_pct;
+extern const volatile u8	no_fast_lb;
 extern const volatile u64	lb_low_util_wall;
 
 u64 __attribute__ ((noinline)) calc_mig_delta(u64 avg_load_invr, int nz_qlen,
@@ -56,7 +57,12 @@ classify_cpdom(struct cpdom_ctx *cpdomc, u64 total_load_invr,
 	if (!cpdomc)
 		return 0;
 
-	if (cpdomc->nr_active_cpus && total_cap_sum > 0) {
+	if (no_fast_lb && sys_stat.nr_active_cpdoms) {
+		u64 avg = total_load_invr / sys_stat.nr_active_cpdoms;
+		x_mig_delta = calc_mig_delta(avg, nz_qlen, mig_delta_factor);
+		stealer_threshold = avg - x_mig_delta;
+		stealee_threshold = avg + x_mig_delta;
+	} else if (cpdomc->nr_active_cpus && total_cap_sum > 0) {
 		fair_share_invr = total_load_invr *
 			     cpdomc->cap_sum_active_cpus /
 			     total_cap_sum;
@@ -169,13 +175,21 @@ int plan_x_cpdom_migration(void)
 		 * and queued load (qload_invr, tracked atomically via
 		 * account/unaccount at enqueue/running).
 		 */
-		cpdomc->load_invr = cpdomc->avg_util_invr_sum +
-				    cpdomc->qload_invr;
+		if (no_fast_lb) {
+			u64 qlen = cpdomc->nr_queued_task;
+			u64 qlen_invr = (qlen << (LAVD_SHIFT * 3)) /
+					cpdomc->cap_sum_active_cpus;
+			cpdomc->load_invr = util + qlen_invr;
+			if (qlen)
+				nz_qlen++;
+		} else {
+			cpdomc->load_invr = cpdomc->avg_util_invr_sum +
+					    cpdomc->qload_invr;
+			if (cpdomc->qload_invr)
+				nz_qlen++;
+		}
 		total_load_invr += cpdomc->load_invr;
 		total_cap_sum += cpdomc->cap_sum_active_cpus;
-
-		if (cpdomc->qload_invr)
-			nz_qlen++;
 	}
 
 	/*
@@ -317,6 +331,10 @@ static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
 	if (!cpdomc->nr_active_cpus)
 		return false;
 
+	if (no_fast_lb &&
+	    !prob_x_out_of_y(1, cpdomc->nr_active_cpus * LAVD_CPDOM_MIG_PROB_FT))
+		return false;
+
 	/*
 	 * Traverse neighbor compute domains in distance order.
 	 */
@@ -381,8 +399,13 @@ static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
 			 * helpers.
 			 */
 			if (consume_dsq(cpdomc_pick, dsq_id)) {
-				decrement_stealee_budget(cpdomc_pick, task_load);
-				decrement_stealer_budget(cpdomc, task_load);
+				if (no_fast_lb) {
+					WRITE_ONCE(cpdomc_pick->is_stealee, false);
+					WRITE_ONCE(cpdomc->is_stealer, false);
+				} else {
+					decrement_stealee_budget(cpdomc_pick, task_load);
+					decrement_stealer_budget(cpdomc, task_load);
+				}
 				return true;
 			}
 		}
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -20,11 +20,17 @@
 extern const volatile u8	mig_delta_pct;
 extern const volatile u64	lb_low_util_wall;
 
-u64 __attribute__ ((noinline)) calc_mig_delta(u64 avg_load_invr, int nz_qlen)
+u64 __attribute__ ((noinline)) calc_mig_delta(u64 avg_load_invr, int nz_qlen,
+					      u64 mig_delta_factor)
 {
 	/*
 	 * Note that added "noinline" to make the verifier happy.
+	 * When mig_delta_factor > 0, the user specified a fixed
+	 * migration delta percentage; otherwise use the dynamic
+	 * shift-based heuristic.
 	 */
+	if (mig_delta_factor > 0)
+		return avg_load_invr * mig_delta_factor / LAVD_SCALE;
 	if (nz_qlen >= sys_stat.nr_active_cpdoms)
 		return avg_load_invr >> LAVD_CPDOM_MIG_SHIFT_OL;
 	if (nz_qlen == 0)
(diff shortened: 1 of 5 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1786,6 +1838,10 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(lavd_init_task, struct task_struct *p,
 s32 BPF_STRUCT_OPS(lavd_exit_task, struct task_struct *p,
 		   struct scx_exit_task_args *args)
 {
+	task_ctx *taskc = get_task_ctx(p);
+	if (taskc)
+		unaccount_queued_load(taskc);
```
multics69, line 1956: I am not 100% sure if unaccounting at the ops.exit_task() path is necessary. In my understanding, a task can exit only while running, so it is already unaccounted for at the ops.running() path. Maybe it would incur double unaccounting. Please double-check any potential double-unaccounting issue at ops.exit_task() path.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -885,19 +928,25 @@ void BPF_STRUCT_OPS(lavd_dequeue, struct task_struct *p, u64 deq_flags)
...
+	if (!taskc) {
+		debugln("Failed to lookup task_ctx for task %d", p->pid);
+		return;
+	}
+
+	/*
+	 * Unaccount queued load before the enable_cpu_bw guard to avoid
+	 * leaking queued_load_invr when tasks are dequeued without running.
+	 */
+	unaccount_queued_load(taskc);
```
multics69: Similar to my comments on ops.exit_task path, need to double-check if there is any possibility of double-unaccounting.

In my understanding, when SCX_DEQ_SLEEP is set, we should not unaccount. That's because SCX_DEQ_SLEEP is set when a task was running and now it becomes sleeping. In this case, the task was already unaccounted upon ops.running(), so we should not unaccount again here.

For SCX_DEQ_CORE_SCHED_EXEC, I think we should unaccount here because the task is runnable but not yet running.

For SCX_DEQ_SCHED_CHANGE, things are even trickier since a task can be in a running state (e.g., calling sched_setaffinity) or in a runnable-but-not-running state (e.g., CPU migration, etc).

Considering all these cases, you can simply check if the current task of an rq is the same as task p. If so, you can skip the unaccounting there since task p is already unaccounted when it becomes runnable to running state. You can use __COMPAT_scx_bpf_cpu_curr().

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Ah, there are cases that a task can be ops.exit_task()-ed even when it is not running. So, similar to the ops.dequeue() case, you can use __COMPAT_scx_bpf_cpu_curr() to check if the task is running. If so, you can safely skip unaccounting.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
bboymimi: Thanks for the detailed breakdown of the dequeue flag semantics. You're right that `__COMPAT_scx_bpf_cpu_curr()` is a precise way to handle the various dequeue conditions. I'll adopt your suggestion.

One concern I have of the current design is that the account/unaccount pattern could be fragile. It faces similar leakage and double-free problems as reference counting. As new code paths are added, any missed accounting sites make `queued_load_invr` unreliable. Once the symptom is noticed, I'm thinking of adding a periodic check in `calc_sys_stat()` that detects drift and resets the value, such as:

```c
estimated = cpdomc->nr_queued_task * sys_stat.avg_runtime_invr;
if (actual > estimated * 2)
    WRITE_ONCE(cpdomc->queued_load_invr, estimated);
```

Even if a leakage path surfaces, the damage is bounded by the threshold. However, it's just a preventive discussion. Let me know what you think.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
bboymimi: Thanks for the review!

scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
```diff
@@ -629,8 +631,11 @@ static void calc_sys_stat(void)
 			calc_avg32(sys_stat.max_perf_cri, c->max_perf_cri);
 	}
 
-	if (c->nr_sched > 0)
+	if (c->nr_sched > 0) {
 		avg_svc_time_iwgt = c->tot_task_time_iwgt / c->nr_sched;
+		sys_stat.avg_runtime_invr = calc_avg(sys_stat.avg_runtime_invr,
+						     c->tot_task_time_invr / c->nr_sched);
```
daidavid: I'm not sure if avg_runtime_invr is the best way to represent "load", a few pitfalls that I see:

- c->nr_sched isn't incremented in the refill case. Say there are two identical tasks running in a 100ms run, 100ms sleep cycle. One runs on a cpu completely uninterrupted while the other is scheduled off every 20ms(slice config) due to a competing task. The avg_runtime_invr can severely skew one way or the other depending on busy a domain is and doesn't really help to discover the load.

- Indirect but related to this patch is the use of calc_avg for avg_runtime_invr. I think for the 10ms fixed interval accounting makes sense for a fixed weighting for calc_avg, but the use of calc_avg(taskc->avg_runtime_invr in stopping will decay far too quickly for sporadic task and accumulate too slowly for a longer running task(stats for refill helps this a bit). This would mean a task that alternates between heavy/sporadic would be inaccurately represented.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: For now, running, task_exit, and dequeue are all we need to handle. Instead of adding preventive code to calc_sys_stat(), let's add an error message to see if the problem happens.

scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
multics69: I think this is a valid concern. PELT-based load measurement would be a better alternative. We can land this PR first, then change it to use PELT-based load or vice versa.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -87,16 +87,16 @@ int plan_x_cpdom_migration(void)
 		util = (cpdomc->avg_util_wall_sum << LAVD_SHIFT) / cpdomc->nr_active_cpus;
 		if ((util >> LAVD_SHIFT) > max_avg_util_wall)
 			max_avg_util_wall = util >> LAVD_SHIFT;
-		qlen = cpdomc->nr_queued_task;
-		qlen_invr = (qlen << (LAVD_SHIFT * 3)) / cpdomc->cap_sum_active_cpus;
-		cpdomc->load_invr = util + qlen_invr;
+		qload_invr = (cpdomc->queued_load_invr << LAVD_SHIFT) /
+			     cpdomc->cap_sum_active_cpus;
+		cpdomc->load_invr = util + qload_invr;
```
daidavid: Assuming that we fix up the qload_invr with more accuracy, the fact that we combine util(which takes time to converge) diminishes the value that qload_invr brings, which allows us to account load migration much faster without having to wait. It might make sense to track the qload_invr as an independent metric for load balancing.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -460,8 +460,31 @@ static bool force_to_steal_task(struct cpdom_ctx *cpdomc)
...
+			if (peek_p) {
+				task_ctx *peek_taskc = get_task_ctx(peek_p);
+				if (peek_taskc &&
+				    peek_taskc->avg_runtime_invr)
+					task_load =
+						peek_taskc->avg_runtime_invr;
+			}
+
+			/*
+			 * Force steal is unconditional for work
```
daidavid, line 476: I've found that unconditional work stealing just generally does not behave well for load balancing. IMO I think we should remove forced_task stealing in general and replace it with this more nuanced approach.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -20,28 +20,133 @@
...
+	u64 fair_share_invr = 0;
+	u64 stealer_threshold = 0;
+	u64 stealee_threshold = 0;
+
+	if (!cpdomc)
+		return 0;
+
+	if (cpdomc->nr_active_cpus && total_cap_sum > 0) {
+		fair_share_invr = total_load_invr *
+			     cpdomc->cap_sum_active_cpus /
```
daidavid, line 67: So for heterogeneous systems, I'm not sure that just looking at the load is the best decision for the system. There's generally a tradeoff when placing tasks between Large/Small CPUs, typically:
- Tasks were placed on larger CPUs because smaller CPUs simply aren't big enough, since we're looking at a group of CPU's capacity. We aren't considering whether or not a task even "fits" within a single CPU within this domain, so looking at the cap_sum_active_cpus of the domain isn't sufficient to make this decision. We should probably add a check to see if the task even fits inside a single CPU.
- Stealing smaller tasks from smaller domains to larger one is also non-ideal energy consumption wise.

I'm not sure, load balancing is tricky for asym systems since the optimal load state is not well defined.

daidavid:
Thanks for the patches Gavin. I left some comments for discussion, but overall I'm very excited about the upcoming improvements!

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Great idea. I was thinking about that as well. To include the util is to reflect the current CPU loading of the running task. However, for the load balancer to react to the real load in time without waiting for the util to reflect matters more. Maybe just using the queue load is enough.

bboymimi:
> Thanks for the patches Gavin. I left some comments for discussion, but overall I'm very excited about the upcoming improvements!

Thanks for the detailed review as well. I'll address the comments ASAP.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: I agree. The current force stealing is used to compensate for the underreaction of the load balancer. I've already tested with force steal disabled, and the scheduler runs better (the workload is distributed as expected). Once I addressed the comment and test under production workloads, the force steal can be removed.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Thank you for the review. I think this is a good insight to consider the cpu capacity when migrating the tasks. Maybe we can consider the feature in the future as part of the selective migration. We can discuss afterward.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -225,6 +225,8 @@ struct task_ctx {
 	u64	last_slice_used_wall;	/* time(ns) used in last scheduled interval: [last running, last stopping] */
 	u32	cpu_id;			/* where a task is running now */
 	u32	prev_cpu_id;		/* where a task ran last time */
+	u8	queued_in_cpdom_id;	/* cpdom this task's load is counted in; LAVD_CPDOM_MAX_NR = not queued */
```
multics69, line 229: I think the access to `queued_in_cpdom_id` should use `READ_ONCE()` and `WRITE_ONCE()` because the variable can be accessed across CPUs within a domain.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -615,6 +615,48 @@ static bool can_direct_dispatch(struct cpu_ctx *cpuc, bool is_cpu_idle)
...
+static __always_inline void account_queued_load(task_ctx *taskc,
+						u8 cpdom_id)
+{
+	struct cpdom_ctx *cpdomc;
+
+	if (cpdom_id >= LAVD_CPDOM_MAX_NR)
+		return;
+
+	/*
+	 * Idempotency: if already accounted in some domain, skip.
```
multics69: I suggest dropping "Idempotency" because we prevent the double accounting here.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
multics69: This also needs to be updated (or removed).

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -615,6 +615,48 @@ static bool can_direct_dispatch(struct cpu_ctx *cpuc, bool is_cpu_idle)
...
+
+	/*
+	 * Snapshot the load value at enqueue time. The unaccount path
+	 * subtracts this exact snapshot, preventing drift when util_est
+	 * changes between enqueue and dequeue.
+	 */
+	u32 load = task_load_metric(taskc);
+	cpdomc = MEMBER_VPTR(cpdom_ctxs, [cpdom_id]);
+	if (cpdomc)
+		__sync_fetch_and_add(&cpdomc->qload_invr, load);
```
multics69, line 649: I think it is worth adding comments on why atomic inc/dec was used in accounting/unaccounting, while for others are accounted to per-CPU, then aggregated to per-domain at sys_stat interval. In my understanding, we need to know the fresh (not a snapshot at sys_stat) value of `qload_invr` while making migration decisions.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -174,16 +177,32 @@ int plan_x_cpdom_migration(void)
...
 			WRITE_ONCE(cpdomc->is_stealer, true);
 			WRITE_ONCE(cpdomc->is_stealee, false);
 			continue;
 		}
 
 		/*
 		 * Over-loaded or non-active domains become a stealee.
+		 * Budget = half the excess load above capacity-proportional
+		 * fair share. This prevents draining an overloaded domain
+		 * in one round while still allowing meaningful migration.
```
multics69: Please mention "ping-ping" or "thundering-hred" here to clarify what the "meaningful" migration means.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -346,22 +346,34 @@ static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
 			if (!READ_ONCE(cpdomc_pick->is_stealee) || !cpdomc_pick->is_valid)
 				continue;
 
+			if (READ_ONCE(cpdomc_pick->stealee_budget_invr) == 0)
```
multics69: I would change it as follows:

```
			if (READ_ONCE(cpdomc_pick->stealee_budget_invr) <= 0)
```

See more comments at the decrement_stealee_budget/decrement_stealer_budget.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -291,6 +291,42 @@ struct cpdom_ctx {
...
+						     u64 amount)
+{
+	u64 remaining = READ_ONCE(cpdomc->stealee_budget_invr);
+
+	if (remaining > amount)
+		WRITE_ONCE(cpdomc->stealee_budget_invr, remaining - amount);
+	else {
+		WRITE_ONCE(cpdomc->stealee_budget_invr, 0);
+		WRITE_ONCE(cpdomc->is_stealee, false);
+	}
```
multics69: Updating stealee_budget_invr is racy here. So use atomic dec instead.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -291,6 +291,42 @@ struct cpdom_ctx {
...
+						     u64 amount)
+{
+	u64 remaining = READ_ONCE(cpdomc->stealer_budget_invr);
+
+	if (remaining > amount)
+		WRITE_ONCE(cpdomc->stealer_budget_invr, remaining - amount);
+	else {
+		WRITE_ONCE(cpdomc->stealer_budget_invr, 0);
+		WRITE_ONCE(cpdomc->is_stealer, false);
+	}
```
multics69: Updating stealee_budget_invr is racy here. So use atomic dec instead.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -346,22 +346,34 @@ static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
...
+				if (peek_taskc)
+					task_load = task_load_metric(peek_taskc);
+			}
+
+			/*
+			 * On success, decrement both egress and ingress
+			 * budgets. The stealer stays active for the
+			 * entire round. Budget exhaustion clears the
+			 * is_stealee/is_stealer flags via the decrement
+			 * helpers.
```
multics69: I think it is worth adding a comment about the TOCTOU problem here. 
There is no gurantee that the peeked task will be consumed, and currently, sched_ext does not have a mechanism to guarantee this.

multics69 (changes requested):

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
bboymimi: Thanks! Updated.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
bboymimi: Thanks! Updated. Idempotency has been removed.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Thanks! This description was removed.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
bboymimi: Thanks! Comment updated.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Thanks! Comment updated.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Thanks! Updated and used s64 instead.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
bboymimi: Thanks! Updated with __sync_fetch_and_sub.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
bboymimi: Thanks! Updated with __sync_fetch_and_sub.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Thanks! Comment updated.

multics69 (approved):
Thanks for addressing comments! LGTM.

bboymimi:
@daidavid @multics69 @rrnewton 

As we discussed in the meeting, I added one more knob for the new load balancer. The only change is in the latest commit (c2678ad), and the remaining patches are the same as the previous version. These are the details for the commit:

The knob gates five call sites: 
    
1. try_to_steal_task() entry — restore the per-domain probabilistic gate that
commit 36f3035ae1fe removed. Without it, every CPU in a domain attempts a steal
each LB round; with it, only one in nr_active_cpus * LAVD_CPDOM_MIG_PROB_FT
does, mirroring the pre-fast-lb thundering-herd mitigation.

2. try_to_steal_task() post-steal — instead of decrementing the stealer/stealee
budgets that commit 15674a4789f9 introduced, clear is_stealee and is_stealer
outright. This restores the old one-steal-per-round behavior.

3. migrate_to_neighbor() post-donation — restore the flag clearing that commit
a00fe067d208 removed, so a successful donation excludes the pair from the rest
of the round.

4. plan_x_cpdom_migration() Phase 1 load equation — restore the
pre- 0e898ced0bde equation load_invr = util + (nr_queued_task << (LAVD_SHIFT*3))
/ cap_sum_active_cpus so domains are classified by queue length (and wall-clock
util) rather than the per-task invariant-runtime sum (qload_invr).

5. classify_cpdom() threshold center — restore the pre- d95e68a76966
flat-average threshold (total_load_invr / nr_active_cpdoms) instead of the
per-domain capacity-proportional fair share. The stealer/stealee budgets are
still computed but unused under the legacy mechanics path, so the wasted u64 
multiplies are accepted.

daidavid (approved):
LGTM, thanks @bboymimi !

multics69 (approved):
LGTM. Thanks @bboymimi !

PR #3481, opened 2026-03-29
scx_lavd: Latency-Aware Task Placement

daidavid:
Latency-Aware Task Placement for scx_lavd

Not all CPUs behave the same. Some get hit by frequent IRQs, hypervisor steal time, or RT/DL preemptions — making them less predictable for latency-sensitive work. This series allows scx_lavd to detect these "turbulent" CPUs and steer
important tasks toward steadier ones.

While the most relevant use case we have today is for networking-heavy scenarios that can create an uneven IRQ distribution, certain CPUs can lose significant time to non-SCX work in general. Tasks running on these CPUs experience unexpected delays that hurt tail latency. Today, scx_lavd treats all CPUs in a compute domain (assuming symmetric CPUs capacity-wise) equally when dispatching tasks, so a latency-critical task can land on a CPU that's about to spend quarter of its time in an IRQ storm.

Traditional load balancing has been about fitting tasks into capacities in terms of load/utilization or the task's "size". We instead track how much time each CPU loses to IRQs and steal, then classify CPUs as turbulent or steady based on their latency headroom. Tasks get a preemption vulnerability score derived from their latency criticality and estimated utilization. The more vulnerable a task is to unexpected interruptions, the more we try to keep it on steady CPUs.

Each compute domain gets a second DSQ (the turbulent DSQ) alongside its main DSQ. Tasks above a dynamic vulnerability threshold go to the main DSQ; the rest go to the turbulent DSQ. The threshold adjusts each stats interval to balance per-capacity load between the two tiers, and the same vulnerability score biases idle CPU selection toward steady CPUs for important tasks.


scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -440,41 +450,29 @@ bool consume_task(u64 cpu_dsq_id, u64 cpdom_dsq_id)
...
-			return true;
-	} else if (use_cpdom_dsq()) {
-		if (consume_dsq(cpdomc, cpdom_dsq_id))
-			return true;
-	} else if (use_per_cpu_dsq()) {
-		if (consume_dsq(cpdomc, cpu_dsq_id))
-			return true;
-	}
+	dsqs[0] = (struct dsq_entry){ cpu_dsq_id,       U64_MAX, use_per_cpu_dsq() };
+	dsqs[1] = (struct dsq_entry){ cpdom_dsq_id,     U64_MAX, use_cpdom_dsq() && !turbulent };
```
bboymimi: The default threshold 12.5%. Is it possible that the cpdom is under IRQ storm and there is no non-turbulent cpu? The preemption-vulnerable tasks in cpdom_dsq could suffer from starvation. Shall we have a fallback here to check with the condition and let the turbulent CPU fetch from the cpdom_dsq still?   Or more conservatively, only fall back when the turbo DSQ is empty?
```c
bool turb_dsq_empty = !scx_bpf_dsq_nr_queued(cpdom_turb_dsq_id);
dsqs[1] = (struct dsq_entry){ cpdom_dsq_id, U64_MAX,
                                use_cpdom_dsq() && (!turbulent || turb_dsq_empty) };
```

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi, line 255: The migration here needs to consider the turbo DSQ as well and selectively pick between the two—depending on the dispatching CPU's turbulence.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
daidavid: So in the event that we approach towards only having turbulent CPUs available, the dynamic threshold should divert the vulnerable tasks into the turbulent CPUs. But yes, we should handle the special case of only having turbulent CPUs within in a domain and where there's a race where we queue a vulnerable task into a steady DSQ but then CPUs become turbulent and leads to starving the task.

I think the suggested fix/fallback can still cause starvation in the race scenario described above, I think it's easier to just keep a count of the turbulent cpus or vice versa and just allow consumption of steady DSQs when the corner case happens.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
daidavid: I did consider and experiment task stealing between the turbulent DSQs but it made no difference performance wise and decided it should be in done in a follow on patch.

There are a few things to consider:
- Should only turbulent CPUs be stealing from other turbulent DSQs? or do we allow intermixing?
- What additional metrics should we consider when stealing aside from just load? ex. stolen task time? runnable time/load?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -416,6 +416,13 @@ static void update_stat_for_running(struct task_struct *p,
 	taskc->last_measured_task_clk = scx_clock_task(cpuc->cpu_id);
 	taskc->last_measured_pelt_clk = scx_clock_pelt(cpuc->cpu_id);
 
+	/*
+	 * Mark this CPU as busy in the duty-cycle ravg.
+	 */
+	ravg_accumulate(&cpuc->avg_util_ravg, LAVD_SCALE, now, 128 * NSEC_PER_MSEC);
```
multics69: Better to define an enum constant for `128 * NSEC_PER_MSEC`

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -535,6 +542,13 @@ static void update_stat_for_stopping(struct task_struct *p,
 	taskc->avg_runtime_invr = calc_avg(taskc->avg_runtime_invr,
 					   taskc->acc_runtime_invr);
 
+	/*
+	 * Mark this CPU as idle in the duty-cycle ravg.
+	 */
+	ravg_accumulate(&cpuc->avg_util_ravg, 0, now, 128 * NSEC_PER_MSEC);
```
multics69: Better to define an enum constant for `128 * NSEC_PER_MSEC`

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1194,6 +1208,25 @@ void BPF_STRUCT_OPS(lavd_runnable, struct task_struct *p, u64 enq_flags)
...
+	local_ravg.old = p_taskc->avg_util_ravg.old;
+	local_ravg.cur = p_taskc->avg_util_ravg.cur;
+
+	ravg_accumulate(&local_ravg, LAVD_SCALE, now,
+			128 * NSEC_PER_MSEC);
+
+	p_taskc->avg_util_ravg.val = local_ravg.val;
+	p_taskc->avg_util_ravg.val_at = local_ravg.val_at;
+	p_taskc->avg_util_ravg.old = local_ravg.old;
+	p_taskc->avg_util_ravg.cur = local_ravg.cur;
```
multics69: I think it's better to add a utility function for this. Also, inside the utility function, let's use `ravg_to_arena` and `ravg_from_arena`.  Something like, `update_util_for_running()`?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1194,6 +1208,25 @@ void BPF_STRUCT_OPS(lavd_runnable, struct task_struct *p, u64 enq_flags)
...
+	 */
+	now = scx_bpf_now();
+	struct ravg_data local_ravg;
+	local_ravg.val = p_taskc->avg_util_ravg.val;
+	local_ravg.val_at = p_taskc->avg_util_ravg.val_at;
+	local_ravg.old = p_taskc->avg_util_ravg.old;
+	local_ravg.cur = p_taskc->avg_util_ravg.cur;
+
+	ravg_accumulate(&local_ravg, LAVD_SCALE, now,
+			128 * NSEC_PER_MSEC);
```
multics69: Better to define an enum constant for `128 * NSEC_PER_MSEC`.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1435,6 +1469,26 @@ void BPF_STRUCT_OPS(lavd_quiescent, struct task_struct *p, u64 deq_flags)
...
+	ravg_accumulate(&local_ravg, 0, now, 128 * NSEC_PER_MSEC);
+	u64 avg_util_fp = ravg_read(&local_ravg, now,
+				    128 * NSEC_PER_MSEC);
+
+	taskc->avg_util_ravg.val = local_ravg.val;
+	taskc->avg_util_ravg.val_at = local_ravg.val_at;
+	taskc->avg_util_ravg.old = local_ravg.old;
+	taskc->avg_util_ravg.cur = local_ravg.cur;
+	taskc->util_est = (u32)(avg_util_fp >> RAVG_FRAC_BITS);
+
```
multics69, line 1488: I think it's better to add a utility function for this. Also, inside the utility function, let's use `ravg_to_arena` and `ravg_from_arena`.  Something like, `update_util_for_sleeping()`?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1435,6 +1469,26 @@ void BPF_STRUCT_OPS(lavd_quiescent, struct task_struct *p, u64 deq_flags)
...
+	 */
+	now = scx_bpf_now();
+	local_ravg.val = taskc->avg_util_ravg.val;
+	local_ravg.val_at = taskc->avg_util_ravg.val_at;
+	local_ravg.old = taskc->avg_util_ravg.old;
+	local_ravg.cur = taskc->avg_util_ravg.cur;
+
+	ravg_accumulate(&local_ravg, 0, now, 128 * NSEC_PER_MSEC);
+	u64 avg_util_fp = ravg_read(&local_ravg, now,
+				    128 * NSEC_PER_MSEC);
```
multics69: Better to define an enum constant for `128 * NSEC_PER_MSEC`.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1194,6 +1208,25 @@ void BPF_STRUCT_OPS(lavd_runnable, struct task_struct *p, u64 enq_flags)
 	WRITE_ONCE(p_taskc->acc_runtime_wall, 0);
 	WRITE_ONCE(p_taskc->acc_runtime_invr, 0);
 
+	/*
+	 * Mark the task as runnable in the duty-cycle ravg.
+	 * Accumulate LAVD_SCALE while runnable, 0 while sleeping.
```
multics69: Accumulating either 0 or 1024 essentially calculates utilization based on wall-clock time, right? If so, do we need to calculate **CPU capacity/frequency-invariant** utilization?

scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
```diff
@@ -358,6 +358,12 @@ static void collect_sys_stat(void)
 		cpuc->avg_steal_util_invr = calc_asym_avg(cpuc->avg_steal_util_invr,
 							   cpuc->cur_steal_util_invr);
 
+		ravg_accumulate(&cpuc->avg_irq_steal_ravg, cpuc->cur_steal_util_invr, c->now,
+					128 * NSEC_PER_MSEC);
+		u64 avg_irq_fp = ravg_read(&cpuc->avg_irq_steal_ravg, c->now, 128 * NSEC_PER_MSEC);
```
multics69: `128 * NSEC_PER_MSEC` ==> enum constant?

scheds/rust/scx_lavd/src/bpf/util.bpf.c
```diff
@@ -393,3 +393,31 @@ u64 get_target_dsq_id(struct task_struct *p, struct cpu_ctx *cpuc)
...
+	/*
+	 * Handle edge cases:
+	 * - If max_lat_cri is 0, return 0 (no tasks have run yet)
+	 * - If lat_cri >= max_lat_cri, return 1024 (maximum)
+	 */
+	if (max == 0)
+		return 0;
+
+	if (lat_cri >= max)
+		return 1024;
```
multics69, line 454: 1024 -> LAVD_SCALE for consistency?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1911,6 +1911,19 @@ static s32 init_cpdoms(u64 now)
 					      cpdomc->id, cpdomc->numa_id);
 				return err;
 			}
+
+			/*
+			 * Create a turbulent DSQ for tasks that
+			 * turbulent CPUs will primarily
+			 * consume from.
```
multics69, line 1928: I think it would be helpful to document the definitions of "turbulent CPU" and "turbulent DSQ" here.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
multics69: Oh, this is an interesting, pathological case that can happen; all CPUs in a domain are under an IRQ storm, so they are all turbulent; unfortunately, the dynamic threshold adjustment is behind, so some tasks are still classified and placed into a non-turbulent DSQ; since all CPUs are turbulent, nobody even tries to consume tasks in non-tubulenet DSQ.

With @bboymimi's idea, things are not perfect. The latency-critical tasks that are in the non-tubulent DSQ will be consumed later only when the turbulent DSQ becomes empty, causing latency-criticality inversion.

Can we use a queue length to avoid such a pathological case? What about this?

```c
dsqs[0] = (struct dsq_entry){ cpu_dsq_id,       U64_MAX, use_per_cpu_dsq() };
dsqs[1] = (struct dsq_entry){ cpdom_dsq_id,
                  U64_MAX,
                  use_cpdom_dsq() && (!turbulent  ||
                       (scx_bpf_dsq_nr_queued(cpdom_dsq_id) > 
                        scx_bpf_dsq_nr_queued(cpdom_turb_dsq_id)))};
dsqs[2] = (struct dsq_entry){ cpdom_turb_dsq_id, U64_MAX, use_cpdom_dsq() };
```

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -267,6 +271,15 @@ struct cpdom_ctx {
...
 	u32	dsq_consume_lat;		    /* latency to consume from dsq, shows how contended the dsq is */
 
+	/* per-cpdom preemption vulnerability threshold tracking */
+	u32	vuln_thresh;			    /* unified lat/util threshold step [0, LAVD_VULN_THRESH_MAX] */
+	u32	util_sum_high_cap;		    /* sum of util_est for high-cap CPUs in this cpdom */
+	u32	util_sum_low_cap;		    /* sum of util_est for low-cap CPUs in this cpdom */
+	u32	cap_sum_high_cap;		    /* sum of capacity for high-cap CPUs in this cpdom */
+	u32	cap_sum_low_cap;		    /* sum of capacity for low-cap CPUs in this cpdom */
+	u16	nr_high_cap_cpus;		    /* count of high-cap CPUs in this cpdom */
+	u16	nr_low_cap_cpus;		    /* count of low-cap CPUs in this cpdom */
```
multics69: Need to revise naming for consistency: high/low_cap vs. turbulant/non-turbulant CPU vs. turbulant /steady CPU.

scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
```diff
@@ -671,6 +696,41 @@ static void calc_sys_stat(void)
...
+		cpdomc = MEMBER_VPTR(cpdom_ctxs, [cpdom_id]);
+		if (!cpdomc)
+			continue;
+
+		if (cpdomc->nr_low_cap_cpus == 0 || cpdomc->cap_sum_low_cap == 0) {
+			cpdomc->vuln_thresh = 0;
+		} else if (cpdomc->cap_sum_high_cap > 0) {
+			u64 load_high = ((u64)cpdomc->util_sum_high_cap << LAVD_SHIFT) / cpdomc->cap_sum_high_cap;
+			u64 load_low = ((u64)cpdomc->util_sum_low_cap << LAVD_SHIFT) / cpdomc->cap_sum_low_cap;
+			u64 target = (load_high * lat_load_target_pct) / 100;
```
multics69, line 738: I think choosing the right `lat_load_target_pct` would be tricky because it would be workload-dependent (how much time is spent for IRQ/RT/DL). An alternative approach would aim to equalize the cur_util_wall between high_cap vs. low_cap CPUs. The goal of dynamically adjusting dyn_thres is to make  CPU utilization between high_cap and low_cap CPUs similar. 

I think we can try it now or later, after landing this PR and Gavin's PR (https://github.com/sched-ext/scx/pull/3479). What do you think, @daidavid ?

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
bboymimi: Good point! My original idea is flaw and Changwoo's queue-length comparison is a better. 

The other point to consider is that there is no high capability CPU when the number of stranded tasks in the main DSQ lower than the ones in the turb DSQ, such as 1 < 5 and still skipped. maybe something like this:

```c
(!turbulent ||
 scx_bpf_dsq_nr_queued(cpdom_dsq_id) > scx_bpf_dsq_nr_queued(cpdom_turb_dsq_id) ||
 cpdomc->nr_high_cap_cpus == 0)

scheds/rust/scx_lavd/src/bpf/sys_stat.bpf.c
daidavid: I think getting the util_wall equalize is the correct goal, I think we just need to improve on the EWMA function a bit. The current ravg util tracking for the CPU is broken because it didn't account for steal_irq time(and we can't simply add it because steal_irq time also seeps into the cpu's util). Let me see if we could just use RAVG to decay/accumulate the cur_util_wall instead to make it a bit more stable. I'll update this patch once I do more testing on it.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
daidavid: Right, I think the correct thing to do here is to use clock_pelt instead.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
daidavid: I couldn't refactor this without the verifier barfing at me. I'll clean this a bit later once ravg is better integrated in arena.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
daidavid: So I tried a few different things here:
- Using clock_pelt directly, I was thinking this would simplify my scaling problem. Oh boy did it open up another whole can of worms that I won't go into here.

- Using avg_perf_factor doesn't work well either:
1. since it suffers from the same decay problem that other calc_avg functions face.
2. since the next time when we call ravg_accumulate to flip to 0, that scaled value can be wildly different when we fold in the old value.

- I think the simplest thing to do here is add a scaling value into ravg_accumulate/read such that it scales the previously accumulated load and folds into the old value.

running:
ravg_accumulate(1024, scale)

stopping:
scale = get the most recent pelt_clock/wall_clock for the running duration.
ravg_accumulate(0, scale)

Since it's only important that we derive the invariant work given a period. I'll add a follow on improvement to ravg that adds this.

bboymimi (approved):

multics69 (approved):
Thanks @daidavid for addressing comments. Looks good to me.

PR #3483, opened 2026-03-30
mitosis: sync per-CPU vtime on cell reassignment in apply_cell_config

likewhatevs:
##  Summary

- When apply_cell_config moves a CPU from cell A to cell B, the old code only advanced cell B's llcs[llc_idx].vtime_now to match the CPU's cctx->vtime_now. The CPU's per-CPU vtime was never updated — it kept its old value from cell A.
- This created multi-billion-nanosecond vtime divergence between CPUs in the same cell. Tasks on the divergent CPU's per-CPU DSQ get clamped to a wrong basis_vtime, breaking vtime ordering within the cell's LLC domain.
- Fix: sync both directions — set both the cell's LLC vtime and the CPU's per-CPU vtime to the max of the two. Neither goes backward, and the reassigned CPU matches the cell's LLC vtime immediately rather than retaining a stale value from its previous cell.

## Test plan
 - Gauntlet run: 24/44 stall dumps had per-CPU vtime divergence >500M within a cell. After fix, divergence eliminated in retest (all CPUs within ~10M).

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -2282,12 +2282,13 @@ int apply_cell_config(void *ctx)
 				if (!cctx)
 					return -ENOENT;
 				/*
-				 * If the CPU is changing cells, advance the
-				 * new cell's vtime to at least match this
-				 * CPU's per-CPU vtime. Otherwise the per-CPU
-				 * DSQ and cell DSQ are in different vtime
-				 * domains and dispatch will starve the
-				 * per-CPU DSQ tasks.
+				 * If the CPU is changing cells, sync its
+				 * per-CPU vtime with the new cell's vtime.
+				 * Without this, the per-CPU DSQ and cell
+				 * DSQ are in different vtime domains —
+				 * tasks enqueued on the per-CPU DSQ get
+				 * clamped to a wrong basis_vtime, breaking
+				 * vtime ordering within the cell.
 				 */
 				if (cctx->cell != cell_id) {
 					cell = lookup_cell(cell_id);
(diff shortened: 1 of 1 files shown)


dschatzberg:
Hmm, the proposed code looks fine to me. I can't quite convince myself that the pre-existing code is wrong though. I guess the argument would be that per-cpu tasks just keep winning the vtime race against the cell tasks. Let me stew on it a bit

likewhatevs:
> Hmm, the proposed code looks fine to me. I can't quite convince myself that the pre-existing code is wrong though. I guess the argument would be that per-cpu tasks just keep winning the vtime race against the cell tasks. Let me stew on it a bit

Will generate a full test report w/ behavior w/ and w/o diff to see if that helps highlight why this change improved test pass rate.

likewhatevs:
This PR only matters w/ LLC awareness enabled, the following runs are w/o it:

```
    tests completed         999/1001        998/1001
    pass rate               84.3%          83.5%
    scheduler deaths        47             44
    unfair                  53             49
    stuck >2s               9              7
    timeout                 42             58

                            stuck (ms)                          spread (%)
                  p50   p95    p99    max     n        p50  p95  max   n
    main            221   911   3823   7512   676         29   60   76    53
    main+PR3483     223  1021   2870   8180   679         29   73   89    49

    h2h (959 common tests):
      main wins:      43
      PR3483 wins:    34
      both pass:      770
      both fail:      112
```

Lets skip this for now a la https://github.com/sched-ext/scx/pull/3487 . 

If LLC awareness is different enough whenever we go to enable it, we may not need this. 

If it's not, rebasing is easy enough and this is available.

likewhatevs:
Not opposed to this if we generally want it now anyhow, but the tests aren't arguing for it w/o LLC enabled.

dschatzberg:
Just documenting this for the future - but I think this logic will conflict with @dforsyth's change in https://github.com/sched-ext/scx/pull/3464 - since we might now change the cell vtime basis out from under a running task.

PR #3485, opened 2026-03-30
mitosis: check cell DSQ in ops.tick to preempt for queued cell work

likewhatevs:
## Summary

- ops.tick checked the per-CPU DSQ and (for borrowed tasks) the native cell DSQ, but never checked the running task's own cell DSQ.
- When work arrived on the cell DSQ while a task held the CPU, update_curr_scx (ext.c:957-958) decremented slice each tick. Until slice reached zero, task_tick_scx (ext.c:2777) did not call resched_curr, so balance_one (ext.c:2147) and ops.dispatch never ran. The cell DSQ work waited up to slice_ns (20ms, SCX_SLICE_DFL at ext.h:30).
- Fix: check the cell DSQ unconditionally in ops.tick. Zeroing slice when cell DSQ work is waiting triggers resched_curr (ext.c:2778) → schedule() → balance_one → ops.dispatch within the current tick. This also subsumes the old cross-cell check — for borrowed tasks, cctx->cell is the CPU's native cell, so the same check reclaims the CPU when native cell work arrives.

## Test plan

- io_borrowing on smt-2llc: 0/5 → 5/5 pass
- dynamic_remove, cpuset_change, borrowing_rebalancing: all failing → 3/3 pass each

## Note
- PR includes https://github.com/sched-ext/scx/pull/3469 -- conceptually this is a different issue (i.e. for a separate class of test failures), but it does depend upon that PR.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
+++ b/scheds/rust/scx_mitosis/src/bpf/dsq.bpf.h
@@ -212,13 +212,14 @@ static inline dsq_id_t get_cell_llc_dsq_id(u32 cell, u32 llc)
  */
 static inline u64 cpu_dsq_raw(u32 cpu)
 {
-	return ((dsq_id_t){ .cpu_dsq = { .cpu = cpu,
-					 .type = DSQ_TYPE_CPU } }).raw;
+	return ((dsq_id_t){ .cpu_dsq = { .cpu = cpu, .type = DSQ_TYPE_CPU } })
+		.raw;
 }
 
 static inline u64 cell_llc_dsq_raw(u32 cell, u32 llc)
 {
 	return ((dsq_id_t){ .cell_llc_dsq = { .llc  = llc,
 					      .cell = cell,
-					      .type = DSQ_TYPE_CELL_LLC } }).raw;
+					      .type = DSQ_TYPE_CELL_LLC } })
+		.raw;
 }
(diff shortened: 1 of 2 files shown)


tommy-u:
I think mitosis will want a preempt of some sort in tick() eventually. I've done a lot of experimentation on this path though and I'd like to hold off on this until I land a fix for dispatch which will change the math on getting tasks running. 

I think this will need to be flag gated.

likewhatevs:
> I think mitosis will want a preempt of some sort in tick() eventually. I've done a lot of experimentation on this path though and I'd like to hold off on this until I land a fix for dispatch which will change the math on getting tasks running.

Sure -- in the event this helps, I think I noticed that at least borrowing can't work w/o tick or something more involved.

tommy-u:
The summary is incredibly dense, please expand on it.

I like the direction of dynamic slice modification like in LAVD. I can see cutting the slice in half or to a quarter when there are pinned tasks waiting. Immediately relinquishing the task on either queue having an element seems overly aggressive. 

> ops.tick checked the per-CPU DSQ and (for borrowed tasks) the native cell DSQ, but never checked the running task's own cell DSQ.

It sounds like you're referring to a previous version of tick? We didn't have an ops.tick before this PR?

1. If tick is called every millisecond, why would we want CPUs to relinquish their tasks so frequently?
3. If there are 2 tasks pinned to the same CPU, would this make them timeshare the CPU at <= 1ms timeslices? This seems concerningly short.

My intuition here is that on large systems, there will almost always be something in the Cell queue.

tommy-u (changes requested):
See questions in comment above.

dschatzberg (changes requested):
This PR and https://github.com/sched-ext/scx/pull/3469 seem to have very similar logic. Can you distinguish them a bit more for me?

PR #3487, opened 2026-03-30
mitosis: fix LLC-related crashes and reduce stalls during cell reconfig

likewhatevs:
## Summary

- apply_cell_config Phase 3: use lookup_cgrp_ctx_fallible and skip cgroups without storage instead of fatally returning -ENOENT. Cgroups created after scheduler init may not have storage yet.
- update_task_llc_assignment: when the picked LLC's cpumask intersection is empty (LLC lost CPUs during concurrent reconfig), fall back to FAKE_FLAT_CELL_LLC instead of crashing via scx_bpf_error. The task gets a valid LLC on next enqueue via maybe_refresh_cell.
- ops.dispatch: after the main dispatch path (own-LLC DSQ + per-CPU DSQ), scan other LLC DSQs in the cell to rescue tasks stranded on DSQs whose LLC lost all CPUs. This runs on every dispatch call when LLC-aware — cost per empty DSQ is scx_bpf_dsq_move_to_local which includes an RCU-protected hash lookup, permission check, and list_empty.
- apply_cell_config Phase 6: kick all CPUs after reconfig so they re-enter dispatch and find stranded tasks via the cross-LLC scan.

## Test plan

- dynamic_churn crash (apply_cell_config -ENOENT): fixed, no longer crashes
- llc_cpuset_race crash (Empty cpumask after intersection): fixed, no longer crashes
- Cross-LLC dispatch stalls: reduced but not fully eliminated — no_ctrl_nested on 4-CPU topology remains intermittent

## Notes
 - This commit touched enough areas it depended upon many of the other fix commits to apply cleanly. This PR is only about HEAD.
 - This commit moved a number of tests from always fail consistently to consistently fail ~25% of the time. Something more invasive than a bugfix might be needed for better than this, I think. Also, these tests are pretty intense, so it's possible this is good enough for real-world use cases.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -963,20 +963,30 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
 		return;
 
 	/*
-	 * Cross-LLC fallback: check other LLC DSQs in this cell.
-	 * After cell reconfiguration, tasks can be stranded on a
-	 * cell-LLC DSQ whose LLC lost all CPUs. Each empty DSQ
-	 * check is a hash lookup + list_empty — negligible cost.
+	 * Cross-LLC orphan rescue: after cell reconfiguration, tasks
+	 * can be stranded on a cell-LLC DSQ whose LLC lost all CPUs
+	 * in this cell. Only scan LLCs with cpu_cnt == 0 — if an
+	 * LLC still has CPUs in the cell, those CPUs drain their own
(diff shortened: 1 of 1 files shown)


likewhatevs:
~~this one's just gonna be a nightmare btw -- that fixes crashes but wrecks fairness.~~ nvm maybe some scoping is enough to get the benefit.

dschatzberg:
Let's hold off on LLC awarness + cell_manager work for now - let's stabilize cell_manager first.

likewhatevs:
> Let's hold off on LLC awarness + cell_manager work for now - let's stabilize cell_manager first.

Yeah this one felt like it needed to be bigger than a one-off fix tbh.

PR #3489, opened 2026-03-30
scx_mitosis: revert dispatch peek to locked iterator approach

tommy-u:
scx_mitosis: add --use-iter-peek flag to work around lockless peek phantom tasks

Add a dsq_peek() helper that can optionally use the iterator-based DSQ peek path instead of the lockless scx_bpf_dsq_peek() kfunc, controlled by the new --use-iter-peek CLI flag. This works around a potential kernel bug where where the lockless peek can return stale tasks when the DSQ being peeked is empty.

Simply put, peek was always returning a task pointer for both DSQs regardless of whether the DSQ was empty or not. This lead to always checking task vtimes which almost always resulted in selecting the CPU DSQ (perhaps because CPU pinned tasks are less frequent and thus more stale thant cell tasks).

When the CPU DSQ is selected, dispatch tries to move the task to local, and almost always failing, it does not try the cell DSQ before going idle (even though the cell DSQ likely has a task in it).

In testing, this reduces task runnable stalls >100ms from thousands to 0 both baremetal and in virtualization.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -47,6 +47,7 @@ const volatile bool	     cpu_controller_disabled	     = false;
 const volatile bool	     reject_multicpu_pinning	     = false;
 const volatile bool	     userspace_managed_cell_mode     = false;
 const volatile bool	     enable_borrowing		     = false;
+const volatile bool	     use_lockless_peek		     = false;
 
 /*
  * Global arrays for LLC topology, populated by userspace before load.
@@ -839,6 +840,22 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
 		scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
 }
 
+/*
+ * Peek at the head of a DSQ. By default use a bpf_for_each iterator loop.
+ * When use_lockless_peek is set, use the lockless scx_bpf_dsq_peek() kfunc.
+ */
+static inline struct task_struct *dsq_peek(u64 dsq_id)
+{
+	struct task_struct *p;
+
+	if (use_lockless_peek)
+		return __COMPAT_scx_bpf_dsq_peek(dsq_id);
+
+	bpf_for_each(scx_dsq, p, dsq_id, 0)
+		return p;
+	return NULL;
+}
+
 void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
 {
 	struct cpu_ctx *cctx;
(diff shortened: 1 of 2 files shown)


likewhatevs:
IMO, this is a better version of https://github.com/sched-ext/scx/pull/3458

```
                    PR3458         PR3489
  pass rate       207/234 (88.5%) 209/232 (90.1%)
  unfair               17             14
  stuck >2s             6              8
  died                  4              2
  spread p50           26             27
  spread p95           58             47
  stuck p50           158            156
  stuck p95          2112           2229
  stuck max          6902           5344

```

This should not be flagged though -- peak is too broken for this to be flagged, will be death by flags if each bug needs a flag.

likewhatevs:
final results on both, my 2c is we go w/ whichever is more readily mergeable, they look identical:
```
                    PR3458         PR3489
  pass rate       756/827 (91.4%) 758/826 (91.8%)
  unfair               53             48
  stuck >2s            16             17
  died                  9              8
  stuck p50           159            165
  stuck p95          1932           2029
  stuck max          6902           6143
  spread p50           26             29
  h2h                  15             16

  PER TOPOLOGY:
    large-4llc     95% / 97%   medium-4llc    96% / 97%
    large-8llc     96% / 95%   medium-8llc   100% / 99%
    near-max-llc   97% / 97%   odd-3llc       86% / 86%
    odd-5llc       89% / 92%   odd-7llc       91% / 86%
    smt-2llc       85% / 86%   smt-3llc       88% / 85%
    tiny-1llc      88% / 92%   tiny-2llc      89% / 90%
```

dschatzberg:
> Simply put, peek was always returning a task pointer for both DSQs regardless of whether the DSQ was empty or not. This lead to always checking task vtimes which almost always resulted in selecting the CPU DSQ (perhaps because CPU pinned tasks are less frequent and thus more stale thant cell tasks).

Do we know why this is? It sounds like peek is just broken/buggy?

This change looks fine from my perspective and I saw from your testing that it clearly helps but I wonder if I was misusing peek or if there's a bug on the kernel side. Let's land and follow up

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -839,6 +840,25 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
...
+	struct task_struct     *p = NULL;
+	struct bpf_iter_scx_dsq it;
+
+	if (!use_iter_peek)
+		return __COMPAT_scx_bpf_dsq_peek(dsq_id);
+
+	/* Force iterator-based peek */
+	if (!bpf_iter_scx_dsq_new(&it, dsq_id, 0))
+		p = bpf_iter_scx_dsq_next(&it);
+	bpf_iter_scx_dsq_destroy(&it);
```
dschatzberg: It's a bit more idiomatic to use bpf_for_each rather than open coding the iterator. Something like this

```
        bpf_for_each(scx_dsq, p, dsq_id, 0)
                return p;
```

tommy-u:
Flipped the flag sense: the old --use-iter-peek (default off, opt into the safe iterator path) is now --use-lockless-peek (default off, opt into the lockless path). The iterator-based peek is now the default behavior — no flag needed to get the fix.

rrnewton:
> Do we know why this is? It sounds like peek is just broken/buggy?

Yes, if so let's fix!

PR #3491, opened 2026-03-30
scx_mitosis: Clean up Cell cpumasks handling

kkdwvd:
A couple of preparatory clean up commits, that refactor the BPF side to consolidate repeated patterns around cpumask handling. When we introduce support for more cpumasks per-cell (i.e., when splitting them into constituent subcells), we will need the same logic repeated over again, hence take the opportunity to clean things up while we're at it.

Details are in the commit log.


kkdwvd:
Note that the nested `struct cell_cpumask_pair` depends on changes made in kernel 6.11, nested structs with kptrs won't work before that, so please flag that in case it's a problem and I will rework it.

kkdwvd:
> This looks great to me - might be worthwhile also moving cell_cpumask_data into the header file as well?

This ended up being non-trivial, intf.h depends on the struct definition for cell_config, not sure that makes sense to move as well, and we depend on intf.h for MAX_CPUS. So I decided to leave it as is.

PR #3493, opened 2026-03-30
scheds/experimental: add scx_flow scheduler

galpt:
## Summary

Add `scx_flow`, a budget-based `sched_ext` scheduler focused on balancing interactive wakeup responsiveness with general-purpose throughput.

`scx_flow` uses per-task wakeup budget accounting, a reserved-vs-shared dispatch split, a bounded adaptive tuning loop, and targeted lifecycle handling through `runnable()`, `cpu_release()`, `enable()`, and `exit_task()`.

This PR places `scx_flow` under `scheds/experimental/scx_flow`.

## Included

- new `scheds/experimental/scx_flow` scheduler crate
- workspace integration in `Cargo.toml` and `Cargo.lock`
- concise scheduler README
- full build-version reporting in startup logs and `--version`
- local suppression of generated `vmlinux.h` warning noise during BPF builds

## Validation

- `cargo check -p scx_flow`
- targeted local benchmark and validation work during development
- external tester (@sirlucjan) confirmation that the current branch survives benchmark burn-in

Supplementary benchmark scripts and archived result bundles used during development are available here:
https://github.com/galpt/testing-scx_flow


sirlucjan:
kernel_version_comparison_All

Very good job!

scheds/rust/scx_flow/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,536 @@
...
+					 (single_cpu_task ||
+					  (tctx->last_refill_ns >= (s64)preempt_refill_min_ns &&
+					   tctx->budget_ns >= (s64)preempt_budget_min_ns)));
+
+				use_local_reserved = should_preempt ||
+					(tctx->wake_cpu_idle && is_wakeup);
+
+				if (should_preempt) {
+					enq_flags |= SCX_ENQ_PREEMPT;
+					scx_bpf_kick_cpu(target_cpu, SCX_KICK_PREEMPT);
```
htejun: `SCX_ENQ_PREEMPT` targeting a local DSQ already implies kicking and it looks like `should_preempt` always targets a local DSQ.

scheds/rust/scx_flow/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,536 @@
...
+					__sync_fetch_and_add(&wake_preempt_dispatches, 1);
+					if (rt_sensitive_wakeup)
+						__sync_fetch_and_add(&rt_sensitive_preempts, 1);
+				} else if (tctx->wake_cpu_idle && (is_wakeup || !scx_bpf_task_running(p))) {
+					scx_bpf_kick_cpu(target_cpu, SCX_KICK_IDLE);
```
htejun: Here, isn't `is_wakeup` and `!scx_bpf_task_running()` the same condition and here too `use_local_reserved` would be true and enqueueing into a local DSQ implies waking the CPU. No explicit kicking is necessary. I think the code would be easier to read if `use_local_reserved` is set in these `if (should_preempt) { ... } else if (tctx->wake_cpu_idle...) {}` blocks.

galpt:
@htejun I pushed this under `scheds/experimental` in [e1222ff1](https://github.com/galpt/scx/commit/e1222ff1) so the PR no longer adds `scx_flow` under `scheds/rust`.

scheds/rust/scx_flow/src/bpf/main.bpf.c
galpt: @htejun Thanks, good catch. I dropped the explicit `scx_bpf_kick_cpu(..., SCX_KICK_PREEMPT)` here and kept the preempt path on the local DSQ only, since `SCX_ENQ_PREEMPT` already covers the wakeup semantics in that case.

The fix is in [e1222ff1](https://github.com/galpt/scx/commit/e1222ff1).

scheds/rust/scx_flow/src/bpf/main.bpf.c
galpt: @htejun Yep, agreed. I rewrote this so `use_local_reserved` is set directly inside the `should_preempt` / `wake_cpu_idle` blocks and removed the redundant local kick at the same time.

That cleanup is also in [e1222ff1](https://github.com/galpt/scx/commit/e1222ff1).

scheds/experimental/scx_flow/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,533 @@
...
+{
+	struct task_ctx *tctx;
+	bool is_idle = false;
+	s32 cpu;
+	s32 this_cpu = bpf_get_smp_processor_id();
+	bool is_this_cpu_allowed = bpf_cpumask_test_cpu(this_cpu, p->cpus_ptr);
+
+	tctx = lookup_task_ctx(p);
+	if (tctx) {
+		update_budget_on_wakeup(p, tctx, bpf_ktime_get_ns());
```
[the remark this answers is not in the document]
galpt: Thanks. This was a real issue, so I fixed it in [`ea22de86`](https://github.com/galpt/scx/commit/ea22de86) by only calling `update_budget_on_wakeup()` from `flow_select_cpu()` when `sleep_started_at` is still set. That keeps `flow_runnable()` from having its refill observation cleared back to zero on the second wakeup path.

scheds/experimental/scx_flow/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,533 @@
...
+
+char _license[] SEC("license") = "GPL";
+
+UEI_DEFINE(uei);
+
+bool scx_bpf_dsq_move_to_local___v2(u64 dsq_id, u64 enq_flags) __ksym __weak;
+bool scx_bpf_dsq_move_to_local___v1(u64 dsq_id) __ksym __weak;
+bool scx_bpf_dsq_move_to_local___new(u64 dsq_id) __ksym __weak;
+bool scx_bpf_consume___old(u64 dsq_id) __ksym __weak;
+
```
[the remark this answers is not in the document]
galpt: Good catch. I removed the local compat plumbing in [`ea22de86`](https://github.com/galpt/scx/commit/ea22de86) and switched the dispatch path over to the shared `scx_bpf_dsq_move_to_local(..., 0)` wrapper from `compat.bpf.h` so the fallback logic lives in one place.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ members = [
     "scheds/rust/scx_rustland",
     "scheds/rust/scx_rusty",
     "scheds/rust/scx_tickless",
+    "scheds/experimental/scx_flow",
     "scheds/rust/scx_wd40",
     "tools/scxcash",
     "tools/scxtop",
```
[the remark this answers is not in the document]
galpt: This was addressed by moving the scheduler under `scheds/experimental` in [`e1222ff1`](https://github.com/galpt/scx/commit/e1222ff1), which matches the current workspace entry. The PR description should follow that location now.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ members = [
     "scheds/rust/scx_rustland",
     "scheds/rust/scx_rusty",
     "scheds/rust/scx_tickless",
+    "scheds/experimental/scx_flow",
```
htejun: Can you put this below scx_wd40?

galpt: Done in [`f76d7c2f`](https://github.com/galpt/scx/commit/f76d7c2f). I moved `scheds/experimental/scx_flow` below `scx_wd40` in the workspace member list.

htejun (approved):

PR #3500, opened 2026-03-31
scx_layered: Force minimum of 1 CPU as upper bound when using cpus_range_frac

etsal:
The cpus_range_frac layer sizing option currently clamps the minimum layer size, but not the maximum. This causes edge the edge case that a small enough maximum CPU bound gets rounded down to 0, while the minimum is clamped down to 1. Consistently enforce the minimum CPU size of 1 for both the minimum and the maximum.


htejun (approved):

etsal:
The merge failed due to an unrelated scx_flash failure. Requeueing.

PR #3502, opened 2026-04-02
common.bpf.h, scheds: Fix migration-disabled task dispatch bug on non-PREEMPT_RCU kernels

multics69:
This series fixes a bug where scx_lavd (and potentially other SCX
schedulers) could dispatch a migration-disabled task to a remote CPU's
local DSQ, triggering scx_error in task_can_run_on_remote_rq() and
terminating the scheduler. The bug manifests only on kernel v6.18 or
later without CONFIG_PREEMPT_RCU.

Background
----------

Since commit [8e4f0b1ebcf2](https://lore.kernel.org/lkml/20250821090609.42508-8-dongml2@chinatelecom.cn/) ("bpf: use rcu_read_lock_dont_migrate() for
trampoline.c") in v6.18, the BPF prolog (__bpf_prog_enter) calls
migrate_disable() only when CONFIG_PREEMPT_RCU is enabled. Before v6.18,
it called migrate_disable() unconditionally for all kernel configurations.

is_migration_disabled() in common.bpf.h has a special case: when
migration_disabled == 1 and p == current, it returns false, assuming the
count of 1 came from the BPF prolog rather than a real migrate_disable()
call. This assumption holds for pre-v6.18 kernels and v6.18+ with
CONFIG_PREEMPT_RCU, but is wrong for v6.18+ without CONFIG_PREEMPT_RCU,
where the prolog no longer touches migration_disabled. The resulting
false negative causes the scheduler to treat the task as freely
migratable, potentially dispatching it to a remote CPU and triggering
scx_error.

Patches
-------

Patch 1 (a4863764f): Fix is_migration_disabled() in common.bpf.h with
two fast paths for the common cases and a slow path that uses a
runtime-probed flag for the pre-v6.18 !CONFIG_PREEMPT_RCU case.

Patch 2 (95ea10940): Introduce scx_lib_init(), called from ops.init(),
which probes whether the BPF prolog actually called migrate_disable() by
checking migration_disabled of the current task at the moment BPF code
is already executing. This makes the detection robust against downstream
kernels that cherry-pick the v6.18 commit onto a pre-v6.18 base, where
a pure version check would give the wrong answer.

Patch 3 (88a4be5bb): Add the scx_lib_init() call to all SCX schedulers
that use is_migration_disabled().

A companion fix for the kernel-side is_bpf_migration_disabled() in
kernel/sched/ext_idle.c will be submitted separately.

### DIFF
--- a/tools/scxtop/src/bpf/intf.h
+++ b/tools/scxtop/src/bpf/intf.h
@@ -76,6 +76,7 @@ struct sched_switch_event {
 	u32  next_dsq_nr;
 	u64  next_dsq_vtime;
 	u64  next_slice_ns;
+	u64  next_wakeup_ts;
 	u32  next_pid;
 	u32  next_tgid;
 	int  next_prio;
--- a/tools/scxtop/src/bpf/main.bpf.c
+++ b/tools/scxtop/src/bpf/main.bpf.c
@@ -332,6 +332,36 @@ int BPF_KPROBE(scx_insert_vtime, struct task_struct *p, u64 dsq, u64 slice_ns,
 	return update_task_ctx(p, dsq, vtime, slice_ns);
 }
 
+/*
+ * __scx_bpf_dsq_insert_vtime takes a struct pointer as the second arg:
+ * struct scx_bpf_dsq_insert_vtime_args { u64 dsq_id, slice, vtime, enq_flags; }
+ * This is the current canonical vtime insert on 6.19+ kernels.
+ */
+SEC("kprobe/__scx_bpf_dsq_insert_vtime")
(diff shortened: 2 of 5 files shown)


multics69:
@htejun @arighi @hodgesds -- There are changes in scx_cosmos, scx_flash, scx_bpfland, scx_tickless, scx_beerland, scx_p2dq, and scx_layered. Please take a look.

multics69:
For record, the related kernel patch was sent: https://lore.kernel.org/sched-ext/20260402023150.660967-1-changwoo@igalia.com/T/#u

multics69:

hodgesds (approved):

PR #3503, opened 2026-04-02
scxtop: pass wakeup timestamp in BPF event and add v2 dsq insert probes

hodgesds:
Pass wakeup_ts directly in the sched_switch BPF event instead of correlating wakeup/switch events in userspace via pending_wakeups. Multi-ringbuffer delivery can reorder events across CPUs, causing stale entries and inflated latency measurements. By embedding the timestamp from the BPF task_ctx, latency calculation becomes self-contained per event.

Also add kprobe handlers for the v2 dsq insert API variants (__scx_bpf_dsq_insert_vtime and scx_bpf_dsq_insert___v2) used on 6.19+ kernels, and attach scx_dsq_move_vtime and scx_dispatch_vtime_from_dsq.

### DIFF
--- a/tools/scxtop/src/bpf/intf.h
+++ b/tools/scxtop/src/bpf/intf.h
@@ -76,6 +76,7 @@ struct sched_switch_event {
 	u32  next_dsq_nr;
 	u64  next_dsq_vtime;
 	u64  next_slice_ns;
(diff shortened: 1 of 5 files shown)


hodgesds:
TUI:
image
MCP server analysis:
image

PR #3506, opened 2026-04-03
mitosis: fix lsp config and warnings

likewhatevs:
fix lsp config for mitosis (scx repo has a config enabling lsp for bpf c).

address warnings in mitosis build by casting WRITE_ONCE return to void.

builds, runs and readable/navigable:

image


likewhatevs:
also setup ci clippy + address issues + update shared githook to deal with lints.

`git config core.hooksPath .githooks` sets them up so no more ci being annoying w/ lints.

PR #3507, opened 2026-04-03
arena/p2dq: address warnings

likewhatevs:
p2dq emits a bunch of warnings, fix them in arena lib.
  - scheds/include/bpf_arena_spin_lock.h — (void) cast on 6 unused WRITE_ONCE/smp_cond_load_acquire_label/arch_mcs_spin_lock_contended_label results
  - scheds/include/bpf_atomic.h — (void) cast on WRITE_ONCE inside smp_store_release macro
  - lib/sdt_alloc.bpf.c — save order to order_end before bpf_for() to fix self-comparison warning (order < order)

p2dq lint runs on ci and i'm updating shared git hook to match ci so no more commit -> push -> figure out lint command -> get lint cmd wrong -> figure out lint cmd for real -> lint -> amend -> push -f cycle.

once the shared git hook is updated, the following cmd will enable lint to just work always:
`git config core.hooksPath .githooks`

### DIFF
--- a/lib/sdt_alloc.bpf.c
+++ b/lib/sdt_alloc.bpf.c
@@ -1222,7 +1222,8 @@ u64 scx_buddy_chunk_alloc(scx_buddy_chunk_t *chunk, int order_req)
 	address = (u64)chunk_idx_to_mem(chunk, idx);
 
 	/* If we allocated from a larger-order chunk, split the buddies. */
-	bpf_for(order, order_req, order) {
+	int order_end = order;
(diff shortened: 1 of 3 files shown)


likewhatevs:
> Looks correct. The problem is that WRITE_ONCE should be an expression, so I assume the compiler is complaining about the expression being left unused. In that case this is the only sane change we can make, even if a bit wordy.

Yeah I was tempted to turn off the warning but at the same time, that warning looks like the sort that could help catch bugs, thanks for confirming!

PR #3511, opened 2026-04-03
mitosis: add watchdog timeout flag and dump trigger (stt compat)

likewhatevs:
This PR enables using stt with mitosis.

Components:
1) Add --watchdog-timeout-ms to set the kernel stall timeout. 
2) Add --trigger-dump with a SIGUSR2 handler that sets a BPF flag checked in ops.tick to trigger scx_bpf_error. 
3) Relax zero-queueing-decisions from bail to warn.

Any of these we don't want in mitosis proper can be carried as random patches but tbh, that's a bit of a pain and will limit usability of stt.

Not sure what stt stands for yet, first i thought scx test tools then i though scx tracing tools because that seems like the coolest part of it, idk.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -49,6 +49,9 @@ const volatile bool userspace_managed_cell_mode = false;
 const volatile bool enable_borrowing = false;
 const volatile bool use_lockless_peek = false;
 
+/* Set by userspace signal handler to trigger error exit with dump */
+bool trigger_dump;
+
 /*
  * Global arrays for LLC topology, populated by userspace before load.
  * Declared in llc_aware.bpf.h as extern.
@@ -2297,11 +2300,18 @@ int apply_cell_config(void *ctx)
 	return 0;
 }
 
+void BPF_STRUCT_OPS(mitosis_tick, struct task_struct *p)
+{
+	if (READ_ONCE(trigger_dump))
+		scx_bpf_error("dump requested via signal");
+}
+
 // clang-format off
 SCX_OPS_DEFINE(mitosis,
 	       .select_cpu		= (void *)mitosis_select_cpu,
 	       .enqueue			= (void *)mitosis_enqueue,
 	       .dispatch		= (void *)mitosis_dispatch,
+	       .tick			= (void *)mitosis_tick,
 	       .running			= (void *)mitosis_running,
 	       .stopping		= (void *)mitosis_stopping,
 	       .set_cpumask		= (void *)mitosis_set_cpumask,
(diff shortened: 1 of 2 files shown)


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -49,6 +49,9 @@ const volatile bool userspace_managed_cell_mode = false;
 const volatile bool enable_borrowing = false;
 const volatile bool use_lockless_peek = false;
 
+/* Set by userspace signal handler to trigger error exit with dump */
+bool trigger_dump;
```
tommy-u, line 53: const volatile?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -2297,11 +2300,18 @@ int apply_cell_config(void *ctx)
 	return 0;
 }
 
+void BPF_STRUCT_OPS(mitosis_tick, struct task_struct *p)
```
tommy-u, line 2303: I think this is the wrong design. On big busy machines, this will get polled >100k times a second.

I'm not convinced this even has to be called from BPF code. If you return from `run()` in main.rs, won't the sched get killed and invoke `mitosis_dump_task`?

If that truly can't work, maybe this pattern would? Def prefer exiting `run()` over this. 

```
BPF:
  SEC("syscall")
  int trigger_dump(void *ctx)
  {
      scx_bpf_error("userspace told me to dump");
      return 0;
  }

  Rust:
  self.skel.progs.trigger_dump
      .test_run(ProgramInput::default())?;
```

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -959,9 +996,10 @@ impl<'a> Scheduler<'a> {
             .flat_map(|cell| QUEUE_STATS_IDX.iter().map(|&idx| cell[idx as usize]))
             .sum();
 
-        // We don't want to divide by zero later, but this is never expected.
+        // Can fire transiently; not worth killing the scheduler over.
         if global_queue_decisions == 0 {
-            bail!("Error: No queueing decisions made globally");
+            warn!("No queueing decisions made globally, skipping stats");
```
tommy-u, line 1001: Thanks for doing this, I should have fixed it a while back.

tommy-u (changes requested):
I think we need a different mechanism than polling `tick()`. I propose two alternatives. Strongly prefer returning from `run()` on the Rust side if that gets us our dump.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
likewhatevs: that's a nice way to do that

PR #3513, opened 2026-04-06
scx_mitosis: Prep patches for subcell support

kkdwvd:
Final set of prep patches before actual subcell logic and config ingestion is introduced. See commit logs for details.


kkdwvd:
Merged in [mitosis-next](https://github.com/sched-ext/scx/commits/mitosis-next).

PR #3515, opened 2026-04-06
scx_layered: add util_compensation for unattributed CPU work

hodgesds:
## Summary

  On load balancers where softirq work concentrates on specific CPUs, layers on those CPUs see reduced effective capacity but their utilization metrics don't reflect the loss — because softirq/irq time isn't attributed to any layer. This
  causes layers to under-request CPUs.

  This PR adds a per-layer `util_compensation` config (0.0–1.0) that proportionally scales a layer's utilization to account for unattributed CPU work (softirq, irq, kernel threads, etc.).

  ## How it works

  - For each CPU, compare `/proc/stat` busy time against the sum of all layer-attributed utilization on that CPU.
  - The ratio (clamped to ≥ 1.0) yields a per-CPU scale factor capturing how much unattributed work is consuming capacity.
  - A layer's utilization is then scaled by the average of these per-CPU scale factors across its assigned CPUs, weighted by the `util_compensation` knob (0.0 = off, 1.0 = full).
  - The compensated utilization feeds directly into the existing `util_range` machinery — no new growth concept is introduced.

  Per-CPU granularity ensures that concentrated softirq load on specific CPUs is accurately captured rather than diluted by a system-wide average.

### DIFF
--- a/scheds/rust/scx_layered/README.md
+++ b/scheds/rust/scx_layered/README.md
@@ -67,5 +67,36 @@ per layer. This is useful if a layer has more latency sensitive tasks, where
 timeslices should be shorter. Conversely if a layer is largely CPU bound with
 less concerns of latency it may be useful to increase the `slice_us` parameter.
 
+### Utilization Compensation
+
+On systems where interrupt processing (irq, softirq) or stolen time
+concentrates on specific CPUs, those CPUs lose a significant fraction of
+their capacity to work that isn't attributed to any layer. Without
+compensation, layers running on these CPUs appear to use less CPU time
+than they actually need, causing them to under-request CPUs.
+
+When enabled, `scx_layered` computes a per-CPU scale factor from
+`/proc/stat` using:
+
+    s[cpu] = Δt / (Δt - irq - softirq - stolen)
+
+Each CPU's per-layer usage delta is multiplied by that CPU's scale
+factor before being summed into the layer total, so a layer doing most
+of its work on interrupt-heavy CPUs gets more compensation than one on
+cold CPUs even if both have the same raw utilization. The compensated
+utilization then drives `calc_target_nr_cpus` through the normal
+`util_range` mechanism. The scale is clamped to `[1.0, 20.0]`.
+
+The compensated utilization is shown alongside raw utilization in the
+per-layer stats output as ` comp_overhead=X.X%` when the layer is
+non-trivially active and the raw and compensated values differ
+meaningfully.
+
+Compensation is disabled by default and can be enabled with
+`--util-compensation` for cases where it is needed. On CPUs with no
+interrupt overhead the scale factor is 1.0 and compensated utilization
+equals raw utilization, so the flag has no effect on layers that only
+run on clean CPUs.
+
 `scx_layered` can provide performance wins, for certain workloads when
 sufficient tuning on the layer config.
--- a/scheds/rust/scx_layered/src/main.rs
+++ b/scheds/rust/scx_layered/src/main.rs
@@ -593,6 +593,12 @@ struct Opts {
     #[clap(long, default_value = "3")]
     gpu_kprobe_level: u64,
 
+    /// Enable utilization compensation for unattributed CPU work (irq, softirq, stolen). When
+    /// enabled, each CPU's layer usage is scaled by the inverse of available capacity to account
+    /// for time lost to interrupts.
+    #[clap(long, default_value = "false")]
+    util_compensation: bool,
+
     /// Enable netdev IRQ balancing. This is experimental and should be used with caution.
     #[clap(long, default_value = "false")]
     netdev_irq_balance: bool,
@@ -856,7 +862,6 @@ struct Stats {
 
     total_util: f64, // Running AVG of sum of layer_utils
     layer_utils: Vec<Vec<f64>>,
-    prev_layer_usages: Vec<Vec<u64>>,
     layer_node_pinned_utils: Vec<Vec<f64>>,
     prev_layer_node_pinned_usages: Vec<Vec<u64>>,
     layer_node_utils: Vec<Vec<f64>>,
@@ -870,6 +875,11 @@ struct Stats {
     cpu_busy: f64, // Read from /proc, maybe higher than total_util
     prev_total_cpu: fb_procfs::CpuStat,
     prev_pmu_resctrl_membw: (u64, u64), // (PMU-reported membw, resctrl-reported membw)
+
+    util_compensation: bool,
+    layer_utils_compensated: Vec<Vec<f64>>, // EWMA of per-CPU-scaled layer utils
+    prev_cpu_layer_usages: Vec<u64>,        // Per-CPU per-layer usages for computing deltas
+    prev_per_cpu_stats: BTreeMap<u32, fb_procfs::CpuStat>,
 
     system_cpu_util_ewma: f64,       // 10s EWMA of system CPU utilization
     layer_dsq_insert_ewma: Vec<f64>, // 10s EWMA of per-layer DSQ insertion ratio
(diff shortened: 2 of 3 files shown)


htejun:
I wonder whether this is too finicky to think about. Would it work to e.g. proportionately add the unattributed cpu consumption to the attributed ones?

hodgesds:
> I wonder whether this is too finicky to think about. Would it work to e.g. proportionately add the unattributed cpu consumption to the attributed ones?

Yeah, that might be easier to reason about. I'm going to first try refactoring some existing configs first to see if I can get them to work.

hodgesds:
The idea with this is we want to grow layers that are on IRQ heavy CPUs/LLCs a little more aggressively until they spill over into non IRQ heavy CPUs/LLCs.

htejun:
Wouldn't it be more accurate to scale the per-cpu usage individually before adding them up?

hodgesds:
Updated to use per CPU compensation adjustments.

htejun (commented):
Thanks Daniel — per-CPU weighting is the right shape for the softirq-concentration problem; a system-wide average would be too coarse. A few asks before I'd want to land this.

**1. Cleaner scale formula**

`s = B/A` mixes two different things:

- irq + softirq + stolen — hardware-imposed, genuinely unavailable to any layer.
- user/sys time from RT tasks, kthreads, non-SCX tasks — another legitimate user of the CPU, not stolen from the layer.

And because the denominator excludes idle time, mostly-idle CPUs can produce pathological scales: 1% softirq + 1% layer + 98% idle gives `s = 2.0` even though the CPU was basically free.

I think a cleaner base is:

```
s[c] = Δt / (Δt − irq[c] − softirq[c] − stolen[c])
```

Pure property of the CPU (no dependency on how many layers exist or how their time splits), direct meaning — "inverse of the fraction of capacity left for non-IRQ work" — and degenerates to 1.0 on idle CPUs. It also agrees with your formula in the specific README example, so the intuition carries over.

**2. Integrate into the main per-CPU aggregation**

Instead of a separate `if has_util_compensation { ... }` branch with parallel state (`prev_cpu_layer_usages`, `cpu_util_scales`, `layer_utils_compensated`), fold this into the main loop:

```
// One /proc/stat read, shared with the existing cpu_busy path.
for each cpu: s[c] = Δt / (Δt − irq_softirq_stolen[c])

// Single pass; builds both streams.
for each cpu, each layer, each usage:
    Δ = cur − prev_cpu[c][ℓ][u]
    raw[ℓ][u]  += Δ
    comp[ℓ][u] += Δ · s[c]
    prev_cpu[c][ℓ][u] = cur
```

Then normalize + `metric_decay` both. This collapses the three separate per-CPU readers (`read_layer_usages`, `read_cpu_total_attributed`, `read_cpu_layer_usages_flat`) into one and removes the double-EWMA (scale-smooth then result-smooth). Always computing both streams is ~one extra multiply per inner cell — negligible next to the BPF map reads you're already doing.

Nice side effect: `comp/raw` becomes a per-layer stat for free, useful both as a preview and as a sanity check. A few easy invariants fall out — `comp ≥ raw` always, `Σ raw` should match the existing `layer_utils` totals within rounding, `s[c]` bounded by some sane limit catches /proc/stat glitches.

**3. Drop the per-layer knob; make compensation always-on**

I don't think the per-layer `util_compensation` float belongs here. With the clean formula above, compensation is the physically correct interpretation of "fraction of available capacity the layer consumed" — which is what `util_range` wants as input. At low util, raw and compensated round to the same `low`/`high` anyway, so compensation only moves allocation when the layer was actually saturating the capacity it had — which is exactly when we want to add CPUs. If a layer ends up over-allocated because it wasn't really saturated, the extra CPU spreads the load, comp drops back into range, and the system settles.

A float from 0.0 to 1.0 also implies a tuning granularity that doesn't really exist — operators won't know whether 0.4 or 0.7 is right for their workload.

What I do think is worth having is a **scheduler-wide opt-out** — a single `--no-util-compensation` (or equivalent) so that if compensation ever causes trouble in production, you can disable it without redeploying layer configs. That's the one genuine use case I see for configurability.

**4. Small caveat worth a README line**

Compensation assumes a layer would have used the stolen capacity if available. That's true for saturated layers but inflates phantom demand for a layer running at its natural rate with idle slack. `util_range`'s discrete response hides small overshoots so this isn't a blocker, but worth documenting so users understand the assumption.

Thanks!

hodgesds:
> ```
> s[c] = Δt / (Δt − irq[c] − softirq[c] − stolen[c])
> ```
> 
> Pure property of the CPU (no dependency on how many layers exist or how their time splits), direct meaning — "inverse of the fraction of capacity left for non-IRQ work" — and degenerates to 1.0 on idle CPUs. It also agrees with your formula in the specific README example, so the intuition carries over.

Yeah, that's much cleaner.
 
> **2. Integrate into the main per-CPU aggregation**
> 
> Instead of a separate `if has_util_compensation { ... }` branch with parallel state (`prev_cpu_layer_usages`, `cpu_util_scales`, `layer_utils_compensated`), fold this into the main loop:
> 
> ```
> // One /proc/stat read, shared with the existing cpu_busy path.
> for each cpu: s[c] = Δt / (Δt − irq_softirq_stolen[c])
> 
> // Single pass; builds both streams.
> for each cpu, each layer, each usage:
>     Δ = cur − prev_cpu[c][ℓ][u]
>     raw[ℓ][u]  += Δ
>     comp[ℓ][u] += Δ · s[c]
>     prev_cpu[c][ℓ][u] = cur
> ```
> 
> Then normalize + `metric_decay` both. This collapses the three separate per-CPU readers (`read_layer_usages`, `read_cpu_total_attributed`, `read_cpu_layer_usages_flat`) into one and removes the double-EWMA (scale-smooth then result-smooth). Always computing both streams is ~one extra multiply per inner cell — negligible next to the BPF map reads you're already doing.

The separate path for compensation enabled or not was kind of a mess.

> Nice side effect: `comp/raw` becomes a per-layer stat for free, useful both as a preview and as a sanity check. A few easy invariants fall out — `comp ≥ raw` always, `Σ raw` should match the existing `layer_utils` totals within rounding, `s[c]` bounded by some sane limit catches /proc/stat glitches.

This is super useful when understanding layers that fall on heavy IRQ cores, here's some example metrics:
```
▶ normal ─ util/open/frac=1029.0/0.51/   34.3 comp_overhead=35.8% prot/prot_preempt=0.02/0.01 tasks=  6595
  sched   tot=102.0k dd_sel/enq=80.2/1.15 dsq/10s=11.7/9.57 wake/exp/re=17.8/0.51/0.18
  exec    keep/max/busy=0.20/0.00/0.00 yield/ign=0.20/    0  slc=5.0ms min_ex=0.00/0.0ms
  mig     mig=39.0 xnuma=0.01 xllc/skip=24.3/0.00 open_idle=0.00 affn_viol=0.39
  preempt preempt/first/xllc/xnuma/idle/fail=0.00/0.00/0.00/0.00/0.00/0.00
  xlayer  wake/re=0.22/0.04 llc_drain/try=0.01/0.01 skip_rnode=2110
  node    pin/ut/ld N0=  0.0/  0.7/  247.8 N1=  0.0/ 90.9/24008.7 N2=  0.0/937.3/51045.4
  node    pin/ut/ld N3=  0.0/  0.1/  385.6
  cpus= 24( 12c) [ 24, 26]
  N0 L00: ░░░░░░░░
  N1 L01: ░░██░██░
  N2 L02: ████████
  N3 L03: ░░░░░░░░
  LLC sched%/lat_ms
  [07]0.02/ 0.1 [10]1.28/ 0.1 [11]1.03/ 0.3 [12]0.02/ 4.9 [13]0.24/ 0.5 [14]0.12/ 1.3
  [15]0.03/ 1.4 [16]11.5/ 0.4 [17]13.1/ 0.3 [18]12.1/ 0.2 [19]14.0/ 0.5 [20]10.0/ 0.4
  [21]10.7/ 0.2 [22]12.7/ 0.2 [23]13.1/ 0.2 [24]0.01/ 1.2 [26]0.01/ 2.5 [28]0.01/ 3.9
  [31]0.01/ 0.8
  ```
  
  The `comp_overhead` shows 35.8% of the layer CPU time was unavailable.

> **3. Drop the per-layer knob; make compensation always-on**
> 
> I don't think the per-layer `util_compensation` float belongs here. With the clean formula above, compensation is the physically correct interpretation of "fraction of available capacity the layer consumed" — which is what `util_range` wants as input. At low util, raw and compensated round to the same `low`/`high` anyway, so compensation only moves allocation when the layer was actually saturating the capacity it had — which is exactly when we want to add CPUs. If a layer ends up over-allocated because it wasn't really saturated, the extra CPU spreads the load, comp drops back into range, and the system settles.

Fair point, I added the per-layer util_compensation float because I wasn't confident the compensation formula would be needed in all cases, and I didn't want to silently change util_range semantics without a way to opt out. With the correct formula I think it makes sense though.

> A float from 0.0 to 1.0 also implies a tuning granularity that doesn't really exist — operators won't know whether 0.4 or 0.7 is right for their workload.
> 
> What I do think is worth having is a **scheduler-wide opt-out** — a single `--no-util-compensation` (or equivalent) so that if compensation ever causes trouble in production, you can disable it without redeploying layer configs. That's the one genuine use case I see for configurability.

The only use case for per layer I can think of is dampening growth for lower-priority layers whose capacity is partially stolen. I do agree that it's much easier to reason about when the compensation is applied globally. I did have to tune some existing layered configs after making this change though.

htejun:
Yeah, enabling by default is likely too risky. Make it global opt-in for now and see whether we can gradually migrate configs?

hodgesds:
Yeah, that sounds good.

htejun:
Looks good. Two minor nits:

`read_per_cpu_layer_usages` returns a flat `Vec<u64>` with manual stride indexing — could you switch to nested `Vec<Vec<Vec<u64>>>` (or snapshot the per-CPU `layer_usages` arrays directly)? Matches the rest of the file and drops the index math.

Also a stray blank line in `calc_target_nr_cpus` after `let util = if util < 0.01 { 0.0 } else { util };`.

Thanks!

likewhatevs (approved):

PR #3528, opened 2026-04-13
scx_cake: fix build error with musl

CosmicToast:
see also: 4503c7080acbaba6a005009b74a16a5b29e5c70c


CosmicToast:
for the record: the bpfnext ci run is failing on scx_chaos and scx_p2dq, neither of which is affected by this patch

CosmicToast:
Would it be possible to estimate whether this will (eventually) make it in? This is the main thing blocking me from updating scx on void linux (I'm the maintainer there), since without this patch it will not build on half the platforms (void supports both glibc and musl, and I'm actively using musl), but void policy is to avoid patches that are unlikely to make it into the next release.

sirlucjan (approved):

PR #3529, opened 2026-04-14
slice shrinking: reduce runnable time for CPU-pinned tasks

tommy-u:
**TLDR:** Pinned tasks can get stuck behind long running task slices with no alternative placement opportunities leading to imbalance. When there are pinned waiters, we shrink the time slice of the running task. This has two intended goals: 1) to improve latency for short running tasks, and 2) to time share an overcommitted CPU at a finer granularity. This leads to better throughput on real workloads.

**Problem:** Mitosis slices are long (20ms). CPU-pinned tasks can get stuck waiting for a long time behind another task with nowhere else to run, even while other cpus idle. This can leave pinned tasks accumulating runnable time and has been measured to reduce throughput.

There are two main cases we have in mind here:
1. System tasks (kworkers, softirqd) that may only need microseconds to run and may wake other threads.
2. Workload threads pinned to a CPU subset. These could be latency sensitive wakers, or cpu bound tasks that need to timeshare the CPU.

**Solution Approach:** Dynamic slice shrinking (`--enable-slice-shrinking`). When a pinned waiter is detected on a CPU DSQ, shrink the running task's slice. We will shrink pinned or unpinned tasks that are ahead of a pinned waiter. We do this by shrinking the current runner's slice when the pinned task gets enqueued, or by shrinking the task about to run when it detects a pinned waiter in the DSQ. Slice times always will be between a min and max value, and scale proportionally to the task's historical runtime.

The minimum runtime is controlled by `--slice-shrink-min-us`, and the maximum by `--slice-shrink-max-us`, defaulting to [500us, 4ms]. The actual value is twice the task's historical runtime, which is calculated as an exponential weighted moving average (EWMA) (alpha=1/8):
- `task_EWMA_runtime * K`, where K is hardcoded to 2.
- Note on min times: the scx core only checks slice expiry on tick (~1ms at HZ=1000), so sub-1ms values can't be enforced precisely. They still matter because the shrink can happen at any point within a tick period (aliasing).

Empirically, tasks tend to have bimodal runtimes that get capped in practice, so we don't put much effort into tuning this parameter.

**What this doesn't do:**
- Vtime is still respected, so we don't "steal" any time from other tasks or "boost" pinned tasks.

**Possible limitations:**
- A pinned task with huge vtime could sit on the DSQ not running, but causing more context switches. We could check if a task is "competitive" before shrinking, but we don't implement this.
- This could over-eagerly run workers that otherwise might batch work (e.g. a softirqd worker). Empirically, these workers seem to have very short runnable times on the default scheduler, so we're not too concerned at this stage.

**Changes:**
- Add `slice_shrinking.bpf.h` with shrink logic and per-cell stats
- Add per-task EWMA runtime tracking in `stopping()` (always-on)
- Move `dsq_peek`/`cstat_add`/`cstat_inc` to `mitosis.bpf.h` for header visibility
- Wire 3 subcategory counters (max/proportional/min), total computed in Rust
- CLI: `--enable-slice-shrinking`, `--slice-shrink-max-us`, `--slice-shrink-min-us`


scheds/rust/scx_mitosis/src/stats.rs
```diff
@@ -36,6 +36,14 @@ pub struct CellMetrics {
     pub steal_pct: f64,
     #[stat(desc = "Pin reject skipped %")]
     pub pin_skip_pct: f64,
+    #[stat(desc = "Slice shrink events")]
```
likewhatevs, line 39: This feels like a lot of stats. I'd say reduce to maybe 2 total (one metric, one cell metric) *unless* these enable tuning params (which i don't think they do, but idk).

likewhatevs (approved):

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
```diff
@@ -86,12 +86,58 @@ struct task_ctx {
...
+/*
+ * Smoothed average of a task's per-wake runtime (EWMA, alpha=1/8).
+ * Updated in stopping() after each run. Starts at 0 and converges
+ * over ~8 runs. Used by features like slice shrinking to estimate
+ * how long a task typically runs.
+ */
+static inline void update_task_runtime_ewma(struct task_ctx *tctx, u64 used)
+{
+	tctx->avg_runtime_ns = (tctx->avg_runtime_ns * 7 + used) / 8;
+}
```
dschatzberg, line 110: I think we should seed this with the first measured value - e.g.

```
if (!tctx->avg_runtime_ns)
  tctx->avg_runtime_ns = used;
else
  tctx->avg_runtime_ns = (tctx->avg_runtimes_ns * 7 + used) / 8;
```

Otherwise you're really underestimating new tasks.

dschatzberg (approved):
I'm going to accept to unblock because I don't see anything buggy here and given the results, I rather we move quickly on this.

That being said, I don't have any intuition for why the slice of a current runner should be based in any way on the waiter's expected runtime. Additionally, as you flag - you're shortening slices anytime there is a waiter even if it's not vtime competitive.

Not having thought about this problem too much I would have wanted this behavior to be a function of the vtime discrepancy: Slice lengths should be a function of the delta between the running task's vtime (including accumulated since running()) and the next waiter. If vtime of the next waiter is << current running task - preempt immediately. If vtime of next waiter is >= current running task + slice length, then let the slice go full length. Then have some function between these two extremes where you e.g. shorten but don't fully preempt.

Again, no idea if that would work, perform well, etc. But I'm trying to reduce as many problems as we can to "just set vtime sensibly". It strikes to me that you're trying to solve some problem via the running time adjustment just for this slice shrinking when whatever the problem is, it probably also applies to vtime ordering.

scheds/rust/scx_mitosis/src/bpf/slice_shrinking.bpf.h
```diff
@@ -0,0 +1,176 @@
...
+ * EWMA runtime. Caller must check enable_slice_shrinking and curr.
+ */
+static inline void slice_shrink_on_enqueue(struct task_struct *curr,
+					   struct task_ctx *pinned_waiter_tctx, u32 cell,
+					   struct cpu_ctx *cctx)
+{
+	enum slice_shrink_result result;
+	u64 limit = slice_shrink_limit(pinned_waiter_tctx->avg_runtime_ns, &result);
+	slice_shrink_apply(curr, limit, result, cell, cctx);
+}
```
dschatzberg, line 137: I think as written you will "shrink" the slice length of the idle task - this probably is fine but messes up stats.

I think if you just short circuit on `curr->flags & PF_IDLE` it will be better

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
tommy-u: Converging from below was intentional, but this is almost certainly better if it matters.

scheds/rust/scx_mitosis/src/bpf/slice_shrinking.bpf.h
tommy-u: I did NOT think about this!

tommy-u: I guess no other sched class should read `p->scx.slice` tho.

PR #3531, opened 2026-04-16
scx_mitosis: fix cell 0 invalid CPU crash and starvation in auto-mode

likewhatevs:
## Summary

In auto-mode (no `--cell-parent-cgroup`), `update_timer_cb` publishes cell 0 with an empty cpumask when child cpusets cover all CPUs. Two failures result: an invalid CPU return from `select_cpu`, and a subsequent task starvation when cell 0's DSQ is never drained.

| Commit | What |
|--------|------|
| 1 | `--stall-timeout-ms` flag + downgrade `"No queueing decisions"` `bail!` to `warn!` |
| 2 | Reproducer (`mitosis_repros.rs`) |
| 3 | `bpf_cpumask_empty` guard in `update_task_cpumask` — invalid CPU fix |
| 4 | Reserve CPU 0 for cell 0 in `update_timer_cb` — starvation fix |

## Evidence: bail was blocking auto-repro

Before the first fix, the auto-repro VM bailed out before reproducing the actual bug:

```
Error: Error: No queueing decisions made globally
```

Cell 0 can transiently have 0 CPUs after cpuset reconfiguration, producing stats intervals with zero queueing decisions. Killing the scheduler blocked the probe infrastructure from replaying the scenario.

## Evidence: invalid CPU crash (no BPF fixes)

Reproducer dump:

```
Error: EXIT: runtime error (invalid CPU 64 from ops.select_cpu())

CELL[0] CPUS=00000000  vtime=3453731127 nr_queued=15
CELL[1] CPUS=00000003  nr_queued=0
CELL[2] CPUS=0000000c  nr_queued=0
CPU[0] cell=1  CPU[1] cell=1  CPU[2] cell=2  CPU[3] cell=2

Task[94] cell=0 dsq=20000000 all_cell_cpus_allowed=1

Backtrace:
  scx_exit+0x48/0x70
  ops_cpu_valid+0x48/0x60
  select_task_rq_scx+0xf1/0x260
```

Cell 0 has 0 CPUs assigned but 15 tasks queued on its DSQ, with `all_cell_cpus_allowed=1`. `bpf_cpumask_subset(empty, X)` returns `true` (vacuous truth: empty set is subset of everything), so the empty cell cpumask passes the subset check in `update_task_cpumask`. The unpinned path runs `bpf_cpumask_any_distribute(empty)` which returns `nr_cpu_ids` (64 on this kernel).

auto-repro (574 events, trigger fired) confirms the crash reaches `ops_cpu_valid`:

```
ops_cpu_valid                   kernel/sched/ext.c:1111
  msg  "from ops.select_cpu()"
```

## Evidence: task starvation (after invalid CPU fix)

After the empty-cpumask guard, the reproducer shows a task stall:

```
runnable task stall (init[98] failed to run for 41.595s)
```

The stuck tasks are in the host cgroup (no cpuset restriction, `CPUS=0000000f`). Dump shows them on cell 0's cell DSQ (`dsq_id=0x20000000`), unscheduled for ~21s:

```
R init[98] *root -20977ms
  dsq_id=0x20000000
  Task[98] cell=0 dsq=20000000 all_cell_cpus_allowed=1
R init[99] *root -20998ms
  dsq_id=0x20000000
  Task[99] cell=0 dsq=20000000 all_cell_cpus_allowed=1

CELL[0] CPUS=00000000  nr_queued=12
```

Cell state timeline:

```
t=6s:   CELL[0] CPUS=0000000c  (cell 0 has CPUs 2,3)
t=47s:  CELL[0] CPUS=00000000  (cell 0 drained)
```

These tasks were dispatched to cell 0's cell DSQ when cell 0 still had CPUs. When cell 0 loses all CPUs, no CPU is assigned to cell 0 — nobody drains its cell DSQ. The empty-cpumask guard (fix 3) only catches tasks going through `update_task_cpumask`; tasks already on the cell DSQ stay put.

## Test plan

- [x] `cargo ktstr test --kernel ../linux -p scx_mitosis --test mitosis_repros` passes after all 4 commits
- [x] auto-repro reproduces both failure modes at intermediate fix stages (574 events for invalid CPU crash, 430 events for task stall)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,6 +67,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
+[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


likewhatevs:
going different route w/ this one i think

PR #3537, opened 2026-04-17
scx_layered: fix per-node fallback for confined layers with skip_remote_node

hodgesds:
On multi-NUMA systems with virtual LLCs (e.g. 4 nodes, 32 LLCs, 64 CPUs), confined layers using skip_remote_node and the `Topo` growth algo may only have CPUs allocated on a subset of NUMA nodes. When tasks wake up on a node where their layer has no CPUs, they are enqueued to per-LLC DSQs on that node. The per-node fallback mechanism is designed to rescue these stranded tasks — each node has a designated fallback CPU that drains DSQs for layers with zero CPUs on that node.

However, two checks in `try_consume_layer()` prevented the fallback from actually consuming these tasks:

1. The confined layer gate `(LAYER_KIND_CONFINED && cpuc->layer_id != layer_id)` caused `try_consume_layer()` to return false immediately when the fallback CPU belonged to a different layer than the one being drained. Since the fallback CPU is allocated to whatever layer owns it, not to the empty layer it is rescuing, this blocked all confined layer fallback. With all layers configured as Confined, the entire fallback mechanism was effectively dead.

2. The `skip_remote_node` check skipped consumption from DSQs on remote NUMA nodes. When the fallback CPU on node N tried to drain a layer that had CPUs only on node M, it skipped node M's DSQs because they were "remote" — even though the stranded tasks were local to node N and the layer's CPUs were what was remote. This inverted the intent of skip_remote_node for the fallback case.

Together these caused runnable task starvation: tasks accumulated in per-LLC DSQs on nodes where their layer had no CPUs, no CPU could drain them, and the sched_ext watchdog triggered after 30 seconds with hundreds of tasks queued across starved DSQs.

Fix by adding a consume_remote parameter to `try_consume_layer()`. When set, it bypasses both the confined layer gate and the `skip_remote_node` check. The per-node fallback dispatch path passes `consume_remote=true`; all other call sites pass false, preserving existing behavior for normal scheduling.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -2518,7 +2518,8 @@ static bool try_drain_layer_node_llcs(struct layer *layer, struct cpu_ctx *cpuc)
 }
 
 static __always_inline bool try_consume_layer(u32 layer_id, struct cpu_ctx *cpuc,
-					      struct llc_ctx *llcc)
+					      struct llc_ctx *llcc,
+					      bool rescue_stranded)
 {
 	struct llc_prox_map *llc_pmap = &llcc->prox_map;
 	struct layer *layer;
@@ -2531,10 +2532,13 @@ static __always_inline bool try_consume_layer(u32 layer_id, struct cpu_ctx *cpuc
 		return false;
 
 	/*
-	 * If a layer is confined, and the CPU doesn't belong to it, we shouldn't
-	 * consume from it.
+	 * If a layer is confined, and the CPU doesn't belong to it, we
+	 * shouldn't consume from it — unless rescue_stranded is set,
+	 * which indicates fallback draining of stranded tasks on nodes
+	 * where the layer has no CPUs.
 	 */
-	if (layer->kind == LAYER_KIND_CONFINED && cpuc->layer_id != layer_id)
+	if (!rescue_stranded &&
+	    layer->kind == LAYER_KIND_CONFINED && cpuc->layer_id != layer_id)
 		return false;
 
 	skip_remote_node = layer->skip_remote_node;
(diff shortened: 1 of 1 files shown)


htejun:
Thanks for digging in. Two notes:

**1. Only the confined-gate relaxation is needed.**

`layered_enqueue()` uses the waking CPU's LLC (main.bpf.c:2047, 2156), so tasks stranded on node N land in DSQs on node N. In `try_consume_layer`, `skip_remote_node` only fires for `nid != remote_nid` — never for the local/same-node LLCs where stranded tasks live. Bypassing it just newly authorizes the fallback CPU to pull from node M's DSQs, violating the user's `skip_remote_node=true` setting for no gain.

The confined-gate check is what actually blocks rescue (fallback CPU belongs to another layer by construction). Relaxing just that should fix the starvation; the commit message's "skipped node M's DSQs" claim doesn't match enqueue behavior and should probably be dropped.

**2. Suffix.**

Minor, but `_cr` took me a moment to parse. Would something like `_rescue` (matching the commit message) read more clearly? `_drain` might be confusing since it's already used for the per-LLC drain mechanism. Param name like `rescue_stranded` could also help.

hodgesds:
> `layered_enqueue()` uses the waking CPU's LLC (main.bpf.c:2047, 2156), so tasks stranded on node N land in DSQs on node N. In `try_consume_layer`, `skip_remote_node` only fires for `nid != remote_nid` — never for the local/same-node LLCs where stranded tasks live. Bypassing it just newly authorizes the fallback CPU to pull from node M's DSQs, violating the user's `skip_remote_node=true` setting for no gain.

ahh good catch!

> **2. Suffix.**
> 
> Minor, but `_cr` took me a moment to parse. Would something like `_rescue` (matching the commit message) read more clearly? `_drain` might be confusing since it's already used for the per-LLC drain mechanism. Param name like `rescue_stranded` could also help.

yeah, that works

htejun:
Oh, if it's just adding a param to the function, maybe we don't need the _rescue suffix? Other than that, looks good to me.

htejun (approved):

PR #3538, opened 2026-04-21
scx_layered: fix clippy lints and crash on IO only NUMA nodes

hodgesds:
On systems with I/O-only NUMA nodes (no CPUs/CXL), `node.span` can be empty
causing `update_fallback_cpus` to panic on `unwrap()`. Skip nodes with empty
spans since they have no CPUs that need a fallback.

Address clippy warnings across scx_layered:

- Replace manual index loops with idiomatic iterators (iter_mut, enumerate, zip, flatten, keys)
- Remove redundant borrows and derefs
- Remove unnecessary imports (libc, scx_bpf_compat)
- Use #[derive(Default)] instead of manual Default impl
- Use div_ceil, is_multiple_of, <<=, and !is_empty() idioms
- Remove unnecessary return statements and as casts
- Use struct literal initialization for perf_event_attr
- Box large variants in StatsReq/StatsRes to reduce enum size
- Collapse else-if blocks and fix doc comment alignment
- Use &mut [usize] instead of &mut Vec<usize> in fn parameter

### DIFF
--- a/scheds/rust/scx_layered/src/alloc.rs
+++ b/scheds/rust/scx_layered/src/alloc.rs
@@ -354,6 +354,7 @@ pub fn unified_alloc(
 /// layer with per-node `pinned` counts and `unpinned_budget` set.
 /// Per-node `unpinned` distribution is NOT resolved here — that's done
 /// by `resolve_spread` and `place_unpinned`.
+#[allow(clippy::needless_range_loop)]
 fn allocate_budgets(
     total_units: usize,
     node_caps: &[usize],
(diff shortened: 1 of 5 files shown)


htejun (approved):
LGTM. Two nits, neither blocking:

1. `.take(N)` on iters already of length `N` is redundant (e.g. `gstats.iter_mut().enumerate().take(NR_GSTATS)` where `gstats.len() == NR_GSTATS`). Clippy-appeasement noise — bare `.iter_mut().enumerate()` reads cleaner. Several instances in `main.rs` (`BpfStats::read`, `read_layer_usages`, `read_layer_membw_agg`, etc.).

2. Replacing `for n in 0..nr_nodes { ... }` with `zip` in `alloc.rs` (`resolve_spread`, `place_unpinned`) silently shortens on length mismatch instead of panicking. Invariants hold today (all vecs sized to `nr_nodes`), but the failure mode quietly changes from loud to silent. Minor.

hodgesds:
Updated with some clippy overrides to keep some of the code more readable.

PR #3540, opened 2026-04-23
lib/topology: eliminate pointer chasing in topo_cpu_to_llc_id()

multics69:
topo_cpu_to_llc_id() can be called very frequently in the scheduler's critical path. For example, the cpu.max bandwidth controller calls it on every task execution via scx_cgroup_bw_consume() and cbw_put_aside() to determine which LLC domain the current CPU belongs to.

The original implementation traverses two hops of parent pointers to reach the LLC node:

    id = topo->parent->parent->id;

Each hop is a pointer dereference into arena memory that likely lands on a different cache line, resulting in two cache misses per call.

Replace struct topology's u64 id with s16 level_ids[TOPO_MAX_LEVEL]. Each entry holds the id of this node's ancestor (or self) at that topology level, with entries above the node's own level set to -EINVAL.

s16 is sufficient for CPU ids (max 32767) and chosen over u16 so that -EINVAL can be stored directly for uninitialised entries.

level and level_ids are placed adjacent in the struct to ensure they share the same cache line for fast-path access.

topo_cpu_to_llc_id() is then simplified to a single constant-index array lookup within the CPU node itself (one cache line):

    return topo->level_ids[TOPO_LLC];

### DIFF
--- a/lib/selftests/st_arena_topology_timer.bpf.c
+++ b/lib/selftests/st_arena_topology_timer.bpf.c
@@ -273,7 +273,7 @@ int scx_selftest_arena_topology_timer_timer_with_helpers(void)
 
 		child_data->magic_value = magic + i + 1;
 		child_data->access_count = i;
-		child_data->cpu_id = child->id;
+		child_data->cpu_id = child->level_ids[child->level];
 		child->data = (void __arena *)child_data;
 		i++;
 	}
@@ -411,10 +411,10 @@ int scx_selftest_arena_topology_timer_arena_data(void)
 
 		child_data->magic_value = magic + i + 2;
 		child_data->access_count = 0;
-		child_data->cpu_id = child->id;
+		child_data->cpu_id = child->level_ids[child->level];
 		child->data = (void __arena *)child_data;
 
-		ret = topo_update_data(child, magic + i + 2, child->id);
+		ret = topo_update_data(child, magic + i + 2, child->level_ids[child->level]);
 		if (ret) {
 			bpf_printk("TOPO ARENA DATA: failed to update child %d data", i);
 			return ret;
--- a/lib/topology.bpf.c
+++ b/lib/topology.bpf.c
@@ -38,18 +38,19 @@ int topo_subset(topo_ptr topo, scx_bitmap_t mask)
 }
 
 static
-topo_ptr topo_node(topo_ptr parent, scx_bitmap_t mask, u64 id)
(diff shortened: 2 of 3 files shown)


bboymimi (approved):
The cache locality enhancement is sound. On the other hand, I found that wd40 is still using the old approach: scheds/experimental/scx_wd40/src/bpf/main.bpf.c at line 129:

`id = topo->parent->parent->id;`

Link: [https://github.com/sched-ext/scx/blob/981d3bbf7c731690b5632aec93b14fa1964e611d/scheds/experimental/scx_wd40/src/bpf/main.bpf.c#L129](https://github.com/sched-ext/scx/blob/981d3bbf7c731690b5632aec93b14fa1964e611d/scheds/experimental/scx_wd40/src/bpf/main.bpf.c#L129)

Anyone still building scx_wd40 **manually** after this PR will get a compile break. However, scx_wd40 was already dropped from the default Cargo build by [commit 92356983883a on 2026-04-09](https://github.com/sched-ext/scx/commit/92356983883abaf7a552f3034ff6a067a71766ef), and it is not part of the Cargo default CI anymore. So, it should not be a blocker.

sirlucjan:
> The cache locality enhancement is sound. On the other hand, I found that wd40 is still using the old approach: scheds/experimental/scx_wd40/src/bpf/main.bpf.c at line 129:
> 
> `id = topo->parent->parent->id;`
> 
> Link: [https://github.com/sched-ext/scx/blob/981d3bbf7c731690b5632aec93b14fa1964e611d/scheds/experimental/scx_wd40/src/bpf/main.bpf.c#L129](https://github.com/sched-ext/scx/blob/981d3bbf7c731690b5632aec93b14fa1964e611d/scheds/experimental/scx_wd40/src/bpf/main.bpf.c#L129)
> 
> Anyone still building scx_wd40 **manually** after this PR will get a compile break. However, scx_wd40 was already dropped from the default Cargo build by [commit 92356983883a on 2026-04-09](https://github.com/sched-ext/scx/commit/92356983883abaf7a552f3034ff6a067a71766ef), and it is not part of the Cargo default CI anymore. So, it should not be a blocker.

wd40 hasn't compiled since the libbpf update. And it hasn't been working for six months.

etsal (approved):
Makes complete sense to me.

multics69:
The verifier error on scx_p2dq and scx_chaos are irrelevant to this change (not able to reproduce when tested locally with a real kernel (not veristat)). So, merge it.

PR #3542, opened 2026-04-23
scx_mitosis: reconcile cells by path and inode

dforsyth:
reconcile_cells() retained tracked cells by path existence but discovered new cells by reading the directory inode. If a cgroup directory was deleted and recreated at the same path between scans, the old cell could survive because the path still existed while a second cell was created for the new inode.

Fix this by snapshotting each child cgroup as (path, inode) during the scan and reconciling tracked cells against that identity. This drops replaced cgroups before creating cells for the new inode and avoids a second metadata lookup in the new-cell pass. Treat directories that disappear after readdir() as transient and retry on the next reconcile instead of failing the scan.

Add a regression test covering same-path replacement with a new inode.

Test Plan: cargo test -p scx_mitosis (45/45 passed)


likewhatevs (approved):
why use inode and not just path btw? if it's for inotify, makes sense.

PR #3544, opened 2026-04-25
lib/cgroup_bw + scx_lavd: arena migration, hot-path caching, correctness fixes

multics69:
This series bundles three loosely-related groups of changes that were
developed together while stress-testing CPU bandwidth throttling under
scx_lavd. 

Arena migration and per-task hot-path caching
---------------------------------------------

scx_cgroup_bw_throttled() runs on every ops.enqueue() and ops.stopping();
scx_cgroup_bw_consume() runs on every task stop. Both paths did two BPF
map lookups per call (cgroup ctx and per-LLC cgroup ctx), and the
throttle-path callers additionally went through bpf_cgroup_from_id() /
bpf_cgroup_release() to obtain a trusted struct cgroup * required by the
BPF_MAP_TYPE_CGRP_STORAGE map. This was pure overhead on the
scheduling critical paths.

The first two patches migrate scx_cgroup_llc_ctx and scx_cgroup_ctx out
of map values into the BPF arena, allocated via scx_static_alloc() and
recycled through a CAS-based lock-free list. cbw_cgrp_llc_map and
cbw_cgrp_map become thin lookup tables holding only an arena pointer per
entry.

The next three patches introduce scx_task_cgroup_bw as a cgroup-bw-aware
extension of scx_task_common (embedded as the first member, so existing
casts keep working), then populate cgx_raw and llcx_raw lazily on first
use and invalidate them on cgroup migration. The consume path gains an
infinite-quota fast path that skips even the LLC-id lookup. On the
cache-hit path both functions drop from two map lookups to zero.

The final patch switches cbw_cgrp_map from BPF_MAP_TYPE_CGRP_STORAGE to
a plain hash keyed by u64 cgrp_id, letting the public API take a u64
directly and removing every bpf_cgroup_from_id() / bpf_cgroup_release()
pair from the throttle/consume/put_aside paths.

- [ 1/13] 3c77f2645 lib/cgroup_bw: migrate scx_cgroup_llc_ctx to BPF arena
- [ 2/13] 94802e774 lib/cgroup_bw: migrate scx_cgroup_ctx to BPF arena
- [ 3/13] a0d68d53d lib/cgroup_bw: introduce scx_task_cgroup_bw for per-task caching
- [ 4/13] 2b8ba3367 lib/cgroup_bw: cache cgx per task to eliminate throttle-path map lookup
- [ 5/13] edb64f149 lib/cgroup_bw: cache llcx per task to eliminate consume-path map lookup
- [ 6/13] 6430d7798 lib/cgroup_bw: use cgrp_id instead of cgroup pointer in throttle/consume/put_aside

Correctness fixes in lib/cgroup_bw
----------------------------------

Four independent bugs uncovered while exercising the above.

The offline path drained the BTQ to the root cgroup without first
unthrottling the offline cgroup, so re-enqueues from the root bounced
straight back into the now-freed LLC ctx, triggering a UAF in
scx_cgroup_bw_enqueue_cb().

Three sites in cbw_free_llc_ctx() and scx_cgroup_bw_init() used the
literal 1 as the root cgroup id instead of the dynamically-detected
ROOT_CGID, silently breaking under cgroup namespaces.

Several while loops had can_loop after a side-effecting operand. The
critical case is cbw_drain_btq_batch(), where can_loop running after
scx_atq_pop() could orphan a task, since pop() clears taskc->atq
immediately. Moving can_loop to the first operand makes the
short-circuit evaluation safe.

When replenish_timerfn() found a cgroup mid-teardown (kernfs node
live but map entry already removed), it silently skipped the cgroup.
Any tasks still in its BTQ were never moved to cbw_throttled_cgroup_ids,
so has_throttled_tasks stayed false and the drain never ran, leaving
tasks permanently stuck in ops_state=QUEUED. Fixed by calling
cbw_cgroup_bw_offline() on the cgroup id so BTQ tasks are moved to
the root for immediate draining.

- [ 7/13] 9181f44a8 lib/cgroup_bw: unthrottle cgroup before draining BTQ on offline
- [ 8/13] cc432636e lib/cgroup_bw: replace hardcoded root cgroup id 1 with ROOT_CGID
- [ 9/13] 20cc620f2 lib/cgroup_bw: place can_loop before side-effecting loop conditions
- [10/13] e8c554684 lib/cgroup_bw: drain BTQ when cgroup ctx is missing in replenish timer

scx_lavd bug fixes
------------------

Three unrelated scx_lavd fixes found during the same test runs.

ops.init() reset cur_logical_clk to 0, but calc_when_to_run()
unconditionally subtracts LAVD_DL_COMPETE_WINDOW. The first enqueue
after init produced a dsq_vtime near UINT64_MAX, which
advance_cur_logical_clk() then latched into cur_logical_clk itself,
locking every subsequent task into near-UINT64_MAX vtimes and
breaking scheduling. Fixed by initialising cur_logical_clk to
LAVD_DL_COMPETE_WINDOW in ops.init(), matching the global initialiser.

lavd_select_cpu()'s can_direct_dispatch() path dispatched to
SCX_DSQ_LOCAL, bypassing ops.enqueue() and its throttle check, so
tasks in throttled cgroups could slip through, consume quota and
deepen debt. Fixed by checking scx_cgroup_bw_throttled() before the
direct dispatch so throttled tasks fall through to ops.enqueue() and
are put aside normally.

In init_task, suggested_cpu_id was left at zero in the parent-copy
branch and only set in the zero-init branch. Unconditionally
initialise it with scx_bpf_task_cpu(p) so the field always holds a
reasonable value from task creation.

- [11/13] 496004366 scx_lavd: fix cur_logical_clk underflow at scheduler init
- [12/13] dd7c64e6e scx_lavd: check cgroup throttle before direct dispatch in select_cpu
- [13/13] 9bdabeaa3 scx_lavd: initialize suggested_cpu_id with scx_bpf_task_cpu() in init_task

### DIFF
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -913,7 +913,7 @@ int cbw_free_llc_ctx(scx_cgroup_ctx_t *cgx, u64 cgrp_id)
 		 * Move all the throttled exiting tasks into the root cgroup.
 		 * Then, delete the LLC context and its associated BTQ.
 		 */
-		if (cgrp_id != 1) {
+		if (cgrp_id != ROOT_CGID) {
 			while ((taskc = scx_atq_pop(btq)) && can_loop) {
 				scx_task_cgroup_bw_t *t = (scx_task_cgroup_bw_t *)taskc;
 				/*
@@ -944,7 +944,7 @@ int cbw_free_llc_ctx(scx_cgroup_ctx_t *cgx, u64 cgrp_id)
 				 * cgroup, so it has to wait until the next
 				 * replenishment interval anyway.
 				 */
-				ret = cbw_put_aside(taskc, 0, 1);
+				ret = cbw_put_aside(taskc, 0, ROOT_CGID);
 				if (likely(!ret)) {
 					nr_moved++;
 				} else {
@@ -1118,7 +1118,7 @@ int scx_cgroup_bw_init(struct cgroup *cgrp __arg_trusted, struct scx_cgroup_init
 	 */
 	if ((cgrp->level > 0) &&
 	    (parent = bpf_cgroup_ancestor(cgrp, cgrp->level - 1))) {
-		if (cgroup_get_id(parent) != 1) {
+		if (cgroup_get_id(parent) != ROOT_CGID) {
 			parentx = cbw_get_cgroup_ctx(parent);
 			if (parentx && !cgroup_is_threaded(parent)) {
 				cbw_free_llc_ctx(parentx, parentx->id);
@@ -2268,7 +2268,7 @@ int replenish_timerfn(void *map, int *key, struct bpf_timer *timer)
 				continue;
 			}
 			WRITE_ONCE(ids[0], cur_cgx->id);
-			if (cur_cgx->id == 1)
+			if (cur_cgx->id == ROOT_CGID)
 				root_added = true;
 			nr_throttled++;
 		}
@@ -2281,7 +2281,7 @@ int replenish_timerfn(void *map, int *key, struct bpf_timer *timer)
 	if (nr_moved > 0 && !root_added) {
 		ids = MEMBER_VPTR(cbw_throttled_cgroup_ids, [nr_throttled]);
 		if (ids) {
-			WRITE_ONCE(ids[0], 1);
+			WRITE_ONCE(ids[0], ROOT_CGID);
 			nr_throttled++;
 		} else {
 			cbw_err("Failed to fetch a throttled cgroup table.");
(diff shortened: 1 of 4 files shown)


lib/cgroup_bw.bpf.c
daidavid: Update this to ROOT_CGID?

daidavid: ROOT_CGID?

daidavid: ROOT_CGID?

daidavid (approved):

lib/cgroup_bw.bpf.c
multics69: Thanks! Fixed!

multics69: Fixed!

multics69: Fixed!

PR #3545, opened 2026-04-27
scx_layered: add Overload layer kind

hodgesds:
Add a new `Overload` layer type that behaves like `Confined` by default but allows tasks to spill to idle unprotected CPUs when all of the layer's allocated CPUs are busy. This provides cache locality benefits under normal load while allowing overflow under pressure.

BPF side: `pick_idle_cpu()` searches layer CPUs first then falls back to unprotected CPUs. `try_consume_layer()` gates cross-layer consumption on `check_no_idle`. `keep_running()` and preemption use confined-style restrictions. Dispatch ordering counts Overload in the grouped bucket.

Rust side: Overload has the same config fields as Confined (`util_range`, `cpus_range`, `cpus_range_frac`, `membw_gb`, `protected`) and participates in CPU allocation identically.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -395,6 +395,7 @@ struct layer {
 	bool			has_cpuset;
 	bool			skip_remote_node;
 	bool			prev_over_idle_core;
+	bool			idle_confined;
 	int			growth_algo;
 
 	u64			nr_tasks;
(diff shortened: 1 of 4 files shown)


kkdwvd:
Couldn't you do this as an option for Grouped layers? Like `select_confined_until_saturation`? So it just skips trying idle selection on non-layer CPUs until the layer gets saturated (i.e., check_no_idle = true). Just like all other policy knobs on idle selection (protected, etc.). And for the dispatch path you probably want behavior like Grouped layers anyway.

hodgesds:
> Couldn't you do this as an option for Grouped layers? Like `select_confined_until_saturation`? So it just skips trying idle selection on non-layer CPUs until the layer gets saturated (i.e., check_no_idle = true). Just like all other policy knobs on idle selection (protected, etc.). And for the dispatch path you probably want behavior like Grouped layers anyway.

That works too, I don't have a strong preference either way. I initially went with a separate layer so that if it ever needed to behave differently to prevent making the Grouped layer overly complex in configuration.

kkdwvd:
LGTM.

kkdwvd (approved):

PR #3547, opened 2026-04-30
lib/cgroup_bw: close task-loss and use-after-free windows under arena_spin_lock contention

multics69:
This series closes a set of task-loss and use-after-free windows in
lib/cgroup_bw uncovered while stress-testing scx_lavd under heavy CPU
bandwidth throttling with concurrent cgroup churn. ATQ (BTQ)
operations use arena_spin_lock(), a bounded-retry spinlock that can
fail under contention. Existing call sites assumed those failures
were either impossible or benign; in practice they orphan tasks,
corrupt rb-trees, or trigger use-after-frees.

Existing-path fixes
-------------------

The deferred-destroy ring's 256-slot size is too small under heavy
cgroup churn: more than 256 BTQs can be queued before a reader (e.g.
cbw_drain_btq_batch() with a snapshotted llcx->btq) releases its
pointer, wrapping the ring and triggering a UAF. Resized to
CBW_NR_CGRP_LLC_MAX * 2.

scx_cgroup_bw_move()'s scx_cgroup_bw_cancel() can time out under
contention with the replenish-time drain; the -ETIMEDOUT was
propagated up to lavd_cgroup_move() and terminated the scheduler.
The failure is self-healing (the task stays in the old BTQ and the
drain re-checks throttle against the updated cgrp_id), so log and
return 0 instead.

scx_atq_cancel() failing under contention leaves taskc->atq stale, so
the next cbw_put_aside() could insert into the rb-tree while still
linked in the old ATQ. Cancel any stale membership first, gated on a
lock-free is_task_throttled() check.

- [1/5] 964cc555f lib/cgroup_bw: size deferred-BTQ ring to CBW_NR_CGRP_LLC_MAX * 2
- [2/5] f1b5a75b9 lib/cgroup_bw: treat BTQ cancel failure during cgroup move as non-fatal
- [3/5] 370a7718e lib/cgroup_bw: cancel stale ATQ membership in cbw_put_aside()

Deferred-retry rings for cgroup-exit drains
-------------------------------------------

cbw_free_llc_ctx() drains an exiting cgroup's BTQ to root before
destroying it. Two contention paths could silently lose tasks:
scx_atq_pop() returning NULL on lock failure (drain exits early,
remaining tasks dropped by scx_atq_destroy()), and cbw_put_aside() to
root returning -EBUSY (task already popped, now homeless). A
deferred-drain ring parks stranded BTQs and cbw_drain_deferred_atqs()
finishes them from rearm_out: each tick, pushing back into the old
BTQ on cbw_put_aside() failure so a popped task always has a home.

The push-back's scx_atq_insert_vtime() can itself fail under
contention. A task-level deferred-move ring parks such orphans and
cbw_move_deferred_tasks() retries cbw_put_aside() each tick. Together
the two rings give every throttled task during cgroup exit a
guaranteed home: old BTQ, root BTQ, deferred-drain ring, or
deferred-move ring.

- [4/5] 6323450cf lib/cgroup_bw: defer interrupted cgroup-exit drains to close a task-loss window
- [5/5] 383b673c5 lib/cgroup_bw: defer orphaned tasks when push-back also fails

### DIFF
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -61,6 +61,13 @@ enum scx_cgroup_consts {
 	 * CBW_NR_CGRP_LLC_MAX * 2 = 131072 slots, 1024 KB.
 	 */
 	CBW_DEFERRED_BTQ_SIZE		= (CBW_NR_CGRP_LLC_MAX * 2),
+	/*
+	 * Slots for BTQs whose drain in cbw_free_llc_ctx() was
+	 * interrupted by lock contention.  Much rarer than cgroup
+	 * destruction, so sized well below CBW_DEFERRED_BTQ_SIZE.
+	 * CBW_DEFERRED_BTQ_SIZE / 2 = 65536 slots, 512 KB.
+	 */
+	CBW_DEFERRED_DRAIN_SIZE		= (CBW_DEFERRED_BTQ_SIZE / 2),
 };
 
 /*
@@ -400,6 +407,23 @@ static u64		cbw_cgroup_ids[CBW_NR_CGRP_MAX];
  * An array of throttled cgroups that need to be reenqueued.
  */
 static u64		cbw_throttled_cgroup_ids[CBW_NR_CGRP_MAX];
+
+/*
+ * Ring buffer of BTQs whose cgroup-exit drain in cbw_free_llc_ctx() was
+ * interrupted by lock contention and therefore still hold tasks.
+ * cbw_drain_deferred_atqs() finishes the drain on each replenish tick
+ * and schedules destruction once empty.
+ *
+ * cbw_nr_deferred_drains counts currently parked BTQs.  The timer skips
+ * cbw_drain_deferred_atqs() entirely when the count is zero, avoiding
+ * a full ring scan in the (common) case of no deferred work.
+ */
+static u64		cbw_deferred_drain_slots[CBW_DEFERRED_DRAIN_SIZE]
+			__attribute__((aligned(SCX_CACHELINE_SIZE)));
+static u64		cbw_deferred_drain_tail
+			__attribute__((aligned(SCX_CACHELINE_SIZE)));
+static u64		cbw_nr_deferred_drains
+			__attribute__((aligned(SCX_CACHELINE_SIZE)));
 
 /*
  * Timer to replenish time budget for all cgroups periodically.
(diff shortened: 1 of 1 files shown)


kkdwvd:
How bad is this such that you're seeing failures? Are you sure it is contention? The timeout for the spin loop is 250ms, that seems like an awfully long time to spend spinning inside the kernel (in a BPF program) and fail, even with contention. Also, arena_spin_lock() is not written in a way to be recoverable after such failures, the guarantees on the lock and queue state are really weak, there will likely still be corruption and all guarantees with respect to data structure consistency are off.

In summary, even if you handle the return case, I don't see how it's working correctly, or that you have any guarantees with respect to correctness. It sounds like a deeper issue that needs fixing; the error return shouldn't really be happening, unless you hit other sources impeding progress (like deadlocks, etc.) that result in timeouts.

multics69:
@kkdwvd -- Thank you for the review. Yes, it is due to the contention; at least the spinlock returns -ETIMEOUT; I wasn't able to check if it happened in the real-world workload, but I confirmed it happens in the synthetic workloads. The synthetic workload that I tested is a bit extreme; 192 threads are contending for 0.5 CPUs of a cgroup (quota = 0.5 CPUs), so it constantly stresses the ATQ and its spinlock. It happens once every 2-3 hours of running in my testing.  The bad part is that once the ATQ operation fails, there is no way to recover silently, except for terminating the scheduler. My solution is essentially deferring the ATQ operations later. With these changes, I no longer see the ATQ failure or cascading failures.

kkdwvd:
I think you misunderstood what I meant. Right now, once the spin lock fails due to a timeout, you cannot assume anything about the validity of the data it protects. There is no way to recover. The scheduler should just do scx_bpf_error() at that point and quit. There is no attempt made at repairing the queue of the lock, so your program is basically in an undefined state.

It may appear that things are working fine with this change, but based on the above, I have no confidence that is the case.

The way to fix this is to either make the lock maintain the queue's validity in face of timeout (hard, but perhaps doable), or increase the timeout to a larger value, but realistically, if the program spends 250ms inside the kernel, something has definitely gone wrong.

Unless this is a realistic scenario, I would simply exit the scheduler once a timeout occurs. Otherwise, we need to seriously pursue a proper fix here (which is probably making changes to the lock algorithm to make it truly fallible, to allow the caller to recover), which these changes or not.

Another point I would stress is that once a may_goto / can_loop has timed out, all other can_loop invocations in the program are short circuited.

The budget of 250ms is across all loop calls. In that sense if you have contention multiple times that individually is not 250ms, but adds up to that in a loop, you can still fail the lock even if it doesn't spend 250ms.

I would do some analysis on the breakdown of time spent in the program and its various phases to dig deeper into the root cause here.

multics69:
> Also, arena_spin_lock() is not written in a way to be recoverable after such failures, the guarantees on the lock and queue state are really weak, there will likely still be corruption and all guarantees with respect to data structure consistency are off.

@kkdwvd  -- I dug into how arena spin lock is implemented. You are right. If the -ETIMEOUT happened, the lock is already dead so retrying is meaningless.

 I will close this PR for now and resubmit one considering the arena lock acquisition failure is irrecoverable.

Issue #3340, opened 2026-02-18
scx_lavd --performance leaks massive memory (~34GB) and pegs CPU on CachyOS 6.19.2 (Intel Core Ultra 9 275HX)


akh-droid:
### Summary

On my ASUS ROG Strix SCAR 18 (G835LX_G835LX, RTX 5090, Intel Core Ultra 9 275HX) running linux-cachyos 6.19.2-2 with scx-scheds 1.0.20-1, enabling scx_lavd in performance mode (Gaming) causes extreme CPU usage and memory growth.

When activated via scx-manager or scx_loader, scx_lavd --performance rapidly consumes high CPU, around 98 to 100 percent, and large amounts of memory. Systemd reported approximately 34 GB peak memory usage before the service had to be force killed.

Switching to scx_bpfland in Gaming mode under identical conditions does not reproduce the issue and remains stable.

The behavior appears specific to scx_lavd in performance mode on this hybrid Intel HX platform.

---

### System

- **Distro:** CachyOS  
- **Kernel:** `6.19.2-2-cachyos`  
- **Package:** `scx-scheds 1.0.20-1`  
- **Scheduler:** `scx_lavd` (started via `scx_loader` “Gaming” mode → `scx_lavd --performance`)  
- **CPU:** Intel Core Ultra 9 275HX (hybrid P/E core architecture)  
- **Laptop:** ASUS ROG Strix SCAR 18 G835LX_G835LX

---

### What happened

Starting `scx_lavd` in Gaming / performance mode caused:

- ~100% CPU usage (one core pegged continuously)
- Massive memory growth (systemd reports ~34 GB peak)
- Scheduler would not stop cleanly; systemd had to SIGKILL it

After switching to `scx_bpfland`, behavior is normal (low CPU usage, ~tens of MB RSS).

---

### Evidence (systemd)

```scx_loader.service: Consumed 1min 37.963s CPU time over 1min 39.093s wall clock time, 34.3G memory peak.```

---

### journalctl excerpt
```

[INFO]: switching Lavd with mode Gaming..
[INFO]: starting scx_lavd command
scx_lavd: Performance mode is enabled.
scx_lavd: Pinned task slice mode is enabled (5000 us). Pinned tasks will use per-CPU DSQs.
...
systemd[1]: scx_loader.service: State 'final-sigterm' timed out. Killing.
systemd[1]: scx_loader.service: Killing process (scx_lavd) with signal SIGKILL.
[...]

PR #3548, opened 2026-04-30
fix(scx_lavd): Fix combinatorial state explosion in energy model generation causing absurd memory usage

BHLuotianyi:
DISCLAIMER: This PR needs a **thorough review**, as I know nothing about the code AI writes! But it tests fine on my Laptop.

I know AI may write trash, but this can at least provide an insight of the problem.

Regarding issue #3340 
 ### Description:                                                                                                                                          
                                                                                                                                                                  
 #### Problem                                                                                                                                                     
                                                                                                                                                                  
 The original Energy Model (EM) initialization in scx_lavd used an exhaustive subset enumeration approach. On high-core-count systems, this triggered a $2^n$     
 complexity explosion, causing the scheduler to hang and consume excessive memory (RSS ballooning to several GiBs) during startup.                                
                                                                                                                                                                  
 #### Solution                                                                                                                                                    
                                                                                                                                                                  
 I have implemented a two-stage optimization to resolve this:                                                                                                     
                                                                                                                                                                  
 1. Algorithm Redesign (DP):                                                                                                                                      
 The initialization logic was refactored from subset enumeration to a Dynamic Programming (DP) approach. Instead of expanding all possible combinations, it now   
 directly considers CPU counts per performance domain and accumulates the lowest-power states for each performance level.                                         
 2. Aggressive State Pruning:                                                                                                                                     
 On symmetric systems (e.g., AMD Zen), many different CPU distributions yield the exact same (performance, power) metrics. Storing all these equivalent           
 permutations would still lead to a state explosion.                                                                                                              
   This fix implements strict pruning: for any given performance/power pair, the optimizer now retains only a single optimal representative. When multiple        
 combinations are equivalent, it prioritizes the one using the fewest performance domains (pd_id_set.len()). This ensures the state table remains small while     
 favoring configurations with better cache locality and reduced leakage power.                                                                                    
                                                                                                                                                                  
 #### Results                                                                                                                                                     
                                                                                                                                                                  
 - Performance: Initialization time reduced from a 15s+ hang to under 0.1s.                                                                                       
 - Memory: Eliminated the multi-GB RSS blowup; the process now starts with a stable, flat memory footprint (~80 MiB).                                             
 - Correctness: The Energy Model remains fully functional, and CPU preference ordering is correctly generated based on the system's energy profile.

### DIFF
--- a/scheds/rust/scx_lavd/src/cpu_order.rs
+++ b/scheds/rust/scx_lavd/src/cpu_order.rs
@@ -981,10 +981,19 @@ impl<'a> EnergyModelOptimizer<'a> {
     ) -> bool {
         match pdsi_map.get_mut(&new_pdsi.performance) {
             Some(v) => {
-                let pdsi = v.iter().next().unwrap();
-                if pdsi.power == new_pdsi.power {
-                    v.insert(new_pdsi)
-                } else if pdsi.power > new_pdsi.power {
+                let (pdsi_power, pdsi_pd_len) = {
+                    let pdsi = v.iter().next().unwrap();
+                    (pdsi.power, pdsi.pd_id_set.len())
+                };
+                if pdsi_power == new_pdsi.power {
+                    if new_pdsi.pd_id_set.len() < pdsi_pd_len {
+                        v.clear();
+                        v.insert(new_pdsi);
+                        true
+                    } else {
+                        false
+                    }
+                } else if pdsi_power > new_pdsi.power {
                     v.clear();
                     v.insert(new_pdsi);
                     true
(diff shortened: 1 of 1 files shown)


multics69:
@BHLuotianyi -- Thanks for trying LAVD. Could you share a tarball under `/sys/kernel/debug/energy_model`? What is the symptom you observed? I'd like to understand what the problem is first.

BHLuotianyi:
> @BHLuotianyi -- Thanks for trying LAVD. Could you share a tarball under `/sys/kernel/debug/energy_model`? What is the symptom you observed? I'd like to understand what the problem is first.

@multics69 

[energy_model.tar.gz](https://github.com/user-attachments/files/27382540/energy_model.tar.gz)

Symptom: scx_lavd causes absurd memory usage and one core 100% before the RAM usage hits its ceiling. The system stutters a lot. The symptom is not observed when using any other scxes.

According to #3340 , the more cores / threads the CPU has, the higher the mem usage is (observed ~30GB; ~5GB on my setup)

According to AI, an exponential growth of RAM usage happens due to energy model generation. The more cores the system has, the more RAM the scx_lavd takes.

图片
图片

multics69:
Thanks @BHLuotianyi for sharing the data. Could you share the processor model? If there is a machine that I can access, I will also try it on my side too.

BHLuotianyi:
> Thanks @BHLuotianyi for sharing the data. Could you share the processor model? If there is a machine that I can access, I will also try it on my side too.

@multics69 
It's my pleasure to help. For my machine it's `Intel(R) Core(TM) Ultra 7 255HX`, which is a laptop processor with a core setup of 8P12E. For the ~30GB mem usage case in the referenced issue, it's a `Intel Core Ultra 9 275HX` with 8P16E.

If needed, I can provide my machine for test via VNC connection.

multics69:
> > Thanks @BHLuotianyi for sharing the data. Could you share the processor model? If there is a machine that I can access, I will also try it on my side too.
> 
> @multics69 It's my pleasure to help. For my machine it's `Intel(R) Core(TM) Ultra 7 255HX`, which is a laptop processor with a core setup of 8P12E. For the ~30GB mem usage case in the referenced issue, it's a `Intel Core Ultra 9 275HX` with 8P16E.
> 
> If needed, I can provide my machine for test via VNC connection.



> > Thanks @BHLuotianyi for sharing the data. Could you share the processor model? If there is a machine that I can access, I will also try it on my side too.
> 
> @multics69 It's my pleasure to help. For my machine it's `Intel(R) Core(TM) Ultra 7 255HX`, which is a laptop processor with a core setup of 8P12E. For the ~30GB mem usage case in the referenced issue, it's a `Intel Core Ultra 9 275HX` with 8P16E.
> 
> If needed, I can provide my machine for test via VNC connection.

Thanks for the extra info. I will try to take a deeper look and come up with another solution (if necessary) this weekend.

xirreal:
Any updates on this?

multics69:
Sorry, @xirreal ! I didn't have time to work on this yet. Will find some time this week.

multics69:
Close as https://github.com/sched-ext/scx/pull/3724 fixed this.

@BHLuotianyi -- if you still hit the problem with PR 3724, please open an issue.

PR #3550, opened 2026-05-03
Multiple randomized proximity maps

newell-purdue:
We are looking for opportunities to make scx_layered more generic and get the best trade-offs of tail queueing latencies vs cache locality. This PR aims to provide a win in tail queueing latency without sacrificing cache locality. Noticed a consistent bias in DSQs of a layer where some queues had larger sizes consistently related to the initial proximity map ordering. Specifically, noticed a specific LLC being ordered later in each proximity map resulting in that LLC having higher queue size due to less frequently dispatched.

This diff produces 32 randomized proximity maps instead of just 1 per LLC, each with unique seed. Upon dispatch's attempt to consume a layer, a random map is selected instead of always using the same map.

Test Plan:

Ran a high utilization hhvm workload on a server with 2 modifications (1) extra logging logic to track average queue-size of DSQs every 100ms and (2) variable amount of maps to use ranging up to 64. I did a parameter sweep of 1 to 64 in powers of 2. Noticed consistently that not much extra balance (tightness of spread) after 32, so using 32 in this PR.

Chart has 1 line of average queue-size every minute per 8 active LLCs for the workload, looking for tighter spread.
Screenshot 2026-05-03 at 1 25 34 PM

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -45,6 +45,7 @@ enum consts {
 	DUTY_CYCLE_SHIFT	= 20,		/* duty_cycle 1.0 = 1 << 20 */
 	LAYER_LAT_DECAY_FACTOR	= 32,
 	CLEAR_PREEMPTING_AFTER	= 10000000,	/* 10ms */
+	NUM_PROXIMITY_MAPS = 32,
 
 	DSQ_ID_SPECIAL_MASK	= 0xc0000000,
 	HI_FB_DSQ_BASE		= 0x40000000,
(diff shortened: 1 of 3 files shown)


newell-purdue:
@htejun and @kkdwvd, this is change we had discussed last week. Just having multiple proximity maps fixes much of the spread in queue size. Also, let me know if I'm proposing this change incorrectly, first time creating PR here.

kkdwvd (approved):

newell-purdue:
Rebasing to check if lint issues go away. Need to investigate other CI failures (will also see if they reproduce)

newell-purdue:
Quite a lot of failures, seemingly due to dependency failures, trying another rebase to see if they're consistently failing

kkdwvd:
Landing, remaining errors are unrelated.

PR #3551, opened 2026-05-04
Fixing 'cargo fmt'

newell-purdue:
When creating https://github.com/sched-ext/scx/pull/3550, noticed CI complaining of unrelated "cargo fmt --check" issues. Putting up PR to fix that.

Simply ran "cargo fmt"


hodgesds (approved):
LGTM

PR #3552, opened 2026-05-05
lib/atq: treat arena_spin_lock -ETIMEDOUT as fatal

multics69:
arena_spin_lock() returns -ETIMEDOUT when one of the bounded spin loops inside arena_spin_lock_slowpath() exhausts its iterations (see scheds/include/bpf_arena_spin_lock.h).  The timed-out waiter just bails, leaving the MCS chain with stale ->next links and any waiters queued behind it stuck on their own bounded spins.  A later acquire races against an inconsistent queue; retrying is unsafe.

Route the four scx_atq_* call sites in lib/atq.bpf.c through the existing scx_atq_lock()/scx_atq_unlock() wrappers so there is one chokepoint, then have scx_atq_lock() call scx_bpf_error() on -ETIMEDOUT.  The error is still returned so callers cannot proceed on stale state; scx_bpf_error() triggers ops_exit so the scheduler tears down cleanly instead of stalling on a dead lock.

### DIFF
--- a/lib/atq.bpf.c
+++ b/lib/atq.bpf.c
@@ -97,13 +97,13 @@ int scx_atq_insert_vtime(scx_atq_t __arg_arena *atq, scx_task_common __arg_arena
 {
 	int ret;
 
-	ret = arena_spin_lock(&atq->lock);
+	ret = scx_atq_lock(atq);
 	if (ret)
 		return ret;
 
 	ret = scx_atq_insert_vtime_unlocked(atq, taskc, vtime);
 
-	arena_spin_unlock(&atq->lock);
+	scx_atq_unlock(atq);
 
 	return ret;
 }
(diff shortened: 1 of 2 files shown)


multics69:
This is based on the discussion about arena lock failure at https://github.com/sched-ext/scx/pull/3547

etsal:
If I understand correctly from #3547, this means we are spinning on a dead lock? If so, isn't this an active bug? Even if there is a race we consider benign, can't we add a state variable somewhere to mark we are a) exiting and b) "in flux"?

multics69:
> If I understand correctly from #3547, this means we are spinning on a dead lock? If so, isn't this an active bug? Even if there is a race we consider benign, can't we add a state variable somewhere to mark we are a) exiting and b) "in flux"?

From the cpu.max library point of view, when -ETIMEDOUT happened, there is no way to handle it properly since retrying it is not meaningful. So, I think better to stop the scheduler. I am not sure if the state variable helps. I think in this case, the error message should be more than sufficient.

Given the fact that arena spin lock can fail due to contention, I am working on avoiding contention at the first place.

etsal:
In that case, what about having a userspace thread monitor the BPF stderr and force a scheduler unload if it sees an arena timeout splat? The kernel emits a splat in the BPF stream when the lock times out.

This way we can also kill the scheduler on other errors, e.g., arena page fault.

multics69:
> In that case, what about having a userspace thread monitor the BPF stderr and force a scheduler unload if it sees an arena timeout splat? The kernel emits a splat in the BPF stream when the lock times out.

Reloading the scheduler is a good idea. What about simply using this?

```
scx_bpf_exit(SCX_ECODE_ACT_RESTART, "...");
```

With this, we can print out the messages and a userspace relaunching the scheduler.
What do you think?

kkdwvd (approved):

PR #3554, opened 2026-05-06
lib/cgroup_bw: cap task-stall latency under cpu.max

multics69:
This series caps the wall-clock latency a task can experience when its
cgroup runs out of cpu.max bandwidth.  It was driven by stalls
observed while stress-testing scx_lavd under heavy cpu.max throttling
with concurrent cgroup churn, SIGSTOP/SIGCONT cycles, and deep cgroup
hierarchies.  The first six commits are diagnostics, robustness, and
refactoring prep work; the last three carry the latency mitigation.

Bounding task-stall latency
---------------------------

Tasks already running when a cgroup throttles hold the CPU for the
full base time slice, blocking the scheduler from rechecking the
throttle for that long.  scx_cgroup_bw_pressure() exposes a 1024-
scale pressure hint that combines a hyperbolic budget term with a
linear BTQ-backlog term; scx_lavd consumes it in calc_time_slice()
to scale slices down by up to 16x under heavy pressure.

A throttled task's BTQ position is its scheduler vtime alone, so a
task with a large vtime can starve indefinitely behind a steady
stream of smaller-vtime arrivals.  Blend the upper 32 bits of
wall-clock nanoseconds with the lower 32 bits of vtime so any task
waits at most ~4.29 s before its epoch makes it the head of the
queue.

The kernel sometimes wakes a task specifically so it can observe a
pending kernel-mediated transition (group SIGSTOP, ptrace trap,
cgroup-v2 freeze, group exit).  Throttling such a task into the BTQ
delays the user-visible operation by the throttle window -- SIGSTOP
appears delayed for seconds, cgroup.freeze appears to hang.  Extend
the existing PF_EXITING bypass in scx_cgroup_bw_throttled() to cover
SIGNAL_GROUP_EXIT and JOBCTL_PENDING_MASK | JOBCTL_TRAP_FREEZE.

- [7/10] ec4a5167f lib/cgroup_bw, scx_lavd: add scx_cgroup_bw_pressure() API
- [8/10] d680de042 lib/cgroup_bw: blend wall-clock time into BTQ vtime to bound throttle delay
- [9/10] e27e6e59f lib/cgroup_bw: bypass throttling for transient kernel-mediated task states

Supporting changes
------------------

cpu_max_bench.py quantifies cpu.max overhead under varying
depth/quota/load and plots eevdf vs scx_lavd side-by-side.  The
verbose cbw_dbg* tracing flooded trace_pipe and slowed every hot
path; cbw_err / cbw_warn still cover the actionable cases.

- [1/10] 49364e459 scripts: add cpu_max_bench.py to measure cpu.max overhead
- [2/10] 905924833 lib/cgroup_bw: drop verbose cbw_dbg* tracing

The deferred-BTQ destruction ring at 256 slots could wrap under
heavy cgroup churn and trigger a UAF; resized to
CBW_NR_CGRP_LLC_MAX * 2.  CBW_CGRP_TREE_HEIGHT_MAX was 32 with a
misleading "matches the kernel" comment (kernel default is INT_MAX);
bumped to 64 and scx_cgroup_bw_init() now rejects deeper trees with
-ENOMEM instead of indexing tree_levels[] out of bounds.

- [3/10] 6ed19e84d lib/cgroup_bw: size deferred-BTQ ring to CBW_NR_CGRP_LLC_MAX * 2
- [4/10] 041da9c8b lib/cgroup_bw: raise cgroup tree height cap to 64 and reject deeper trees

scx_cgroup_bw_dump() was hard-coded to bpf_printk(); add a mode
parameter so ops.dump*() callbacks can route output to the SCX dump
buffer.  Factor the open-coded taskc-cached cgx/llcx lookup pattern
into helpers so the new pressure API picks up the caching for free.

- [5/10] ad0767f70 lib/cgroup_bw, scx_lavd: route scx_cgroup_bw_dump() to bpf_printk or scx_bpf_dump
- [6/10] e9baee052 lib/cgroup_bw: factor taskc-cached cgx/llcx accessors into helpers

arena_spin_lock() -ETIMEDOUT leaves the MCS chain with stale ->next
links; a running scheduler must tear down rather than retry against
an inconsistent queue, but a fresh load reinitialises the MCS state.
Surface the failure as scx_bpf_exit(SCX_ECODE_ACT_RESTART, ...) so
user-space orchestration can respawn the scheduler automatically.

- [10/10] a87edcf82 lib/atq: request scheduler restart on arena_spin_lock -ETIMEDOUT

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -285,8 +285,19 @@ static void advance_cur_logical_clk(struct task_struct *p)
 	}
 }
 
+static u32 get_cgroup_pressure(task_ctx *taskc)
+{
+	if (!enable_cpu_bw)
+		return LAVD_SCALE;
+
+	return scx_cgroup_bw_pressure(taskc->cgrp_id, (u64)taskc);
+}
+
 static u64 calc_time_slice(task_ctx *taskc, struct cpu_ctx *cpuc)
 {
+	u64 slice_wall;
+	u32 pressure;
+
 	/*
 	 * Calculate the time slice of @taskc to run on @cpuc.
 	 */
@@ -294,14 +305,21 @@ static u64 calc_time_slice(task_ctx *taskc, struct cpu_ctx *cpuc)
 		return LAVD_SLICE_MAX_NS_DFL;
 
 	/*
+	 * Get the cgroup pressure to scale down the time slice when the
+	 * cgroup is throttled. A pressure greater than LAVD_SCALE reduces
+	 * the slice proportionally, making tasks yield the CPU more often.
+	 */
+	pressure = get_cgroup_pressure(taskc);
+
+	/*
 	 * If pinned_slice_ns is enabled and there are pinned tasks waiting
 	 * to run on this CPU, unconditionally reduce the time slice for
 	 * all tasks to ensure pinned tasks can run promptly.
 	 */
 	if (pinned_slice_ns && cpuc->nr_pinned_tasks) {
-		taskc->slice_wall = min(pinned_slice_ns, sys_stat.slice_wall);
+		slice_wall = min(pinned_slice_ns, sys_stat.slice_wall);
 		reset_task_flag(taskc, LAVD_FLAG_SLICE_BOOST);
-		return taskc->slice_wall;
+		goto out;
 	}
 
 	/*
@@ -316,8 +334,12 @@ static u64 calc_time_slice(task_ctx *taskc, struct cpu_ctx *cpuc)
 	 * However, if there are pinned tasks waiting to run on this CPU,
 	 * we do not boost the task's time slice to avoid delaying the pinned
 	 * task that cannot be run on another CPU.
+	 *
+	 * Also, if a cgroup is under pressure (e.g., throttled), do not boost
+	 * time slice.
 	 */
-	if (!no_slice_boost && !cpuc->nr_pinned_tasks &&
+	if (!no_slice_boost && pressure <= LAVD_SCALE &&
+	    !cpuc->nr_pinned_tasks &&
 	    (taskc->avg_runtime_wall >= sys_stat.slice_wall)) {
 		/*
 		 * When the system is not heavily loaded, so it can serve all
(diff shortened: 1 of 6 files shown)


multics69:
PR 3554 includes the changes in PR 3552 (ATQ, -ETIMEDOUT PR) for ease of testing. Once PR 3552 lands, I will rebase it properly.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -353,22 +374,24 @@ static u64 calc_time_slice(task_ctx *taskc, struct cpu_ctx *cpuc)
...
 	 * If slice boost is either not possible, not necessary, or not
 	 * eligible, assign the regular time slice.
 	 */
-	taskc->slice_wall = sys_stat.slice_wall;
+	slice_wall = sys_stat.slice_wall;
 	reset_task_flag(taskc, LAVD_FLAG_SLICE_BOOST);
-	return taskc->slice_wall;
+
+out:
+	return taskc->slice_wall = min((slice_wall * LAVD_SCALE) / pressure, 1);
```
daidavid: Should this be max( ?

multics69: Oh, it should be max. Fixed. Thanks!

multics69:
Rebased to the master.

bboymimi (approved):
LGTM, thanks!

multics69:
Rebased to the main.

daidavid:
Thanks for the patches Changwoo, here are my two cents. While the patches themselves look ok,

- I'm not a fan of how this series is organized. The "supporting changes" aren't really relevant to the title of the "lib/cgroup_bw: cap task-stall latency under cpu.max" and really should be broken into their own series. Only patches 7 & 8 seems actually relevant to the problem at hand.

- The commit text and the series summary should focus on _why_ the patches are required and some description of the problem that they're trying to solve rather than just a description of what the patches are doing. For example, how does the problem manifest in patch 7? It seems like it would be only an issue with very long slice configurations, given that enforcement periods are intentionally loose (100ms). What use cases are actually using these long slice configurations and do the latencies actually matter?

- Patch 8 is based on the observation: "a task with a large vtime can starve indefinitely behind a steady stream of smaller-vtime arrivals" This seems a rather pathological case, under heavy load and very limited available bandwidth. Can you clarify and include in commit text under what condition/configuration this situation occurs?

multics69:
I will close this for now and will submit separate PRs when needed.

multics69:
Patches 6, 7, and 8 were addessed in https://github.com/sched-ext/scx/pull/3618

PR #3556, opened 2026-05-06
ci: disable scx_p2dq and scx_chaos veristat on bpf-next

hodgesds:
Mirrors the existing veristat-disable pattern already used by the gating veristat job for scx_layered/scx_lavd/scx_arena_selftests. The bpf-next veristat job (continue-on-error) was reporting verifier rejections for p2dq_select_cpu/enqueue/update_idle and the equivalent chaos programs even though the same BPF objects load and run cleanly under the same kernel via the actual scheduler binaries (verified with vng on bpf-next/master @ 2ca6723a5f7b).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,6 +136,7 @@ jobs:
           - name: scx__bpf__bpf-next
             url: https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git
             tag: master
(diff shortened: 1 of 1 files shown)


hodgesds:
This should clean up CI signals. It seems veristat isn't really reliable as the scheduler still loads fine on `bpf-next` built kernels.

likewhatevs (approved):
i have better replacements for how this works underway so i say do it

PR #3558, opened 2026-05-07
scx_pandemoniumv5.9.0:

wllclngn:
Affinity stranding closed. Topology-coverage ceiling raised. Slice-relative demotion. Tier-priority tick preempt.

The R_eff spill chain that v5.8.0 introduced for per-CPU DSQ reachability walked the affinity rank without consulting p->cpus_ptr. Under cgroup-cpuset churn (libvirt machine.slice, qemu iothreadpin, kthread_bind) the helper could land a task in a per-CPU DSQ that no allowed CPU consumes, reproducing the affinity-stranding class behind scx#728 and matching the 43.5s runnable-task stalls Roland reported on 32C (PANDEMONIUM#8: mon_iothread, rpc-libvirtd).

Alongside the affinity fix, the per-CPU R_eff peer table was hardcoded at MAX_AFFINITY_CANDIDATES = 16. At 12C this covered every peer; at 32C only half the topology was visible to the R_eff lookup. The compile-time bound is now derived from MAX_CPUS as `(MAX_CPUS >> 3)` (= 128 at the current MAX_CPUS=1024), tying the ceiling to the same knob that governs the rest of the BPF-side compile-time sizing. The verifier rejected the full MAX_CPUS - 1 expansion (1023-bound for-loops with multi-branch bodies blow up path-state); MAX_CPUS >> 3 stays well under the verifier limit while covering common datacenter systems through 128C fully. The runtime budget clamps now use min(nr_cpu_ids - 1, MAX_AFFINITY_CANDIDATES) so coverage tracks actual hardware up to the verifier-safe ceiling. PANDEMONIUM has computed the full N x N R_eff matrix at boot since v5.4.0; the truncation only happened at the BPF- emission boundary. populate_affinity_rank_map now emits the full sorted list with explicit u32::MAX sentinels for unused slots.

Two adjacent regressions cleaned up alongside: the v5.8.0 demotion gate could not fire for the tasks it was meant to catch (fixed 2.5ms threshold above a 1ms slice cap), and tick preempt was BATCH-only, leaving INTERACTIVE long-runners and fork-storm children unpreemptible even when a LAT_CRITICAL wakeup was pending.

src/bpf/main.bpf.c:

AFFINITY THREADING THROUGH R_EFF SPILL CHAIN
- find_idle_l2_sibling, find_idle_by_affinity, find_pcpu_with_room, and pick_pcpu_dsq_with_spill grow a `const struct cpumask *allowed` parameter. Each candidate CPU is gated by bpf_cpumask_test_cpu(cpu, allowed) before the existing idle / nr_queued predicate runs.
- Six call sites pass p->cpus_ptr: select_cpu's WAKE_SYNC R_eff search (find_idle_by_affinity + pick_pcpu_dsq_with_spill), select_cpu's WAKE_SYNC waker-DSQ fallback (pick_pcpu_dsq_with_spill), select_cpu's normal_path idle target (pick_pcpu_dsq_with_spill), enqueue tier 1's L2 sibling search (find_idle_l2_sibling), and enqueue tier 2's preempt-target spill (pick_pcpu_dsq_with_spill).
- pick_pcpu_dsq_with_spill's level-1 src_cpu check folds into a `src_ok` guard (allowed-aware); levels 2 and 3 unchanged in shape -- spill still ranks by R_eff, fallback still lands at node_dsq -- but neither can pick a CPU outside the task's cpumask. The node-DSQ escape valve is the by-construction safety net: any allowed CPU on the node drains it via STEP 3 of the dispatch waterfall.
- Closes the affinity-stranding class. cpus_ptr changes between the bpf_cpumask_test_cpu and the scx_bpf_dsq_insert_vtime are still racy in principle (Tejun, scx#728: kthread_bind path can publish a prev_cpu inconsistent with cpus_ptr), but the kernel auto-dequeues on cpumask change and re-enqueues fresh; the proactive filter collapses the racy window to the kernel's own dequeue/re-enqueue envelope.

TIER 2 FALLBACK ARMS interactive_waiting
- When pick_pcpu_dsq_with_spill returned the node-level DSQ (id >= nr_cpu_ids) from tier 2's preempt path, the SCX_KICK_PREEMPT to t2_cpu drained that CPU's per-CPU DSQ first; the victim sitting in node_dsq was only reached at STEP 3 of the dispatch waterfall. Under sustained wake load, STEP 0 always had work and STEP 3 was rarely reached.
- enqueue tier 2 now calls arm_interactive_waiting(tctx) when tier2_dsq >= nr_cpu_ids, so tick has a per-millisecond preempt pathway against BATCH residents on the kicked CPU and the dispatch waterfall reaches STEP 3.
- Pairs with the CPU-bound demotion below: demoted saturators are TIER_BATCH and now eligible for tick preemption within preempt_thresh_ns.

CPU-BOUND DEMOTION GATE: slice_cap-RELATIVE
- v5.8.0 demoted INTERACTIVE -> BATCH in stopping() at fixed avg_runtime >= 2.5ms. INTERACTIVE slice is capped at slice_ns (default 1ms; MWU compresses lower under load), so avg_runtime asymptotes to slice_ns and never crosses 2.5ms for the exact pure CPU-bound tasks the demotion targets. The gate fired only on rare measurement skew above slice_ns -- functionally inert for the saturator class.
- Replacement: avg_runtime * 4 >= slice_cap * 3 (75% of live slice_ns) AND ewma_age <= 4. Catches pure CPU-bound within ~6 stopping() cycles regardless of MWU's slice setting. The ewma_age guard spares legit interactive tasks that consume full slice but sleep frequently -- ewma_age increments on sleep->runnable in runnable(), so a true interactive worker grows ewma_age past 4 in short order; only the pure CPU-bound (which never sleep) stay below.
- slice_cap read once via get_knobs() at the top of the demote block; fallback 1ms preserves the previous behavior under knob unavailability.

TICK PREEMPT POLICY: TIER PRIORITY
- tick() preempted only when the running task was TIER_BATCH and on_cpu
  >= preempt_thresh_ns. INTERACTIVE long-runners (saturators not yet
  demoted, fork-storm children spawned at INTERACTIVE) held their CPU through any pending wakeup, including a TIER_LAT_CRITICAL victim's.
- New rule: preemptible = (tier == TIER_BATCH) || (latcrit_waiting && tier == TIER_INTERACTIVE). BATCH residents stay preemptible whenever interactive_waiting is set (any non-batch wake pending); INTERACTIVE residents become preemptible only when a TIER_LAT_CRITICAL wake is specifically pending (latcrit_waiting).
- Encodes the priority lattice: LAT_CRIT outranks INTERACTIVE outranks BATCH. INTERACTIVE keeps protection from ordinary BATCH-waiter contention. Closes the 10ms-cadence stall class observed in the synthetic reproducer (tests/issue8_repro.py) where a LAT_CRIT victim could not displace a non-yet-demoted INTERACTIVE saturator.

src/bpf/intf.h:

MAX_AFFINITY_CANDIDATES IS NOW (MAX_CPUS >> 3)
- The macro changes from the literal `16` to the MAX_CPUS-derived expression `(MAX_CPUS >> 3)`. At MAX_CPUS=1024 this evaluates to 128; if MAX_CPUS is ever changed, the ceiling moves with it. The verifier rejected the full MAX_CPUS - 1 expansion (1023-bound for-loops with multi-branch bodies blow up path-state); MAX_CPUS >> 3 stays well under the verifier limit while covering common datacenter systems through 128C with no truncation. Beyond that width the compile-time ceiling caps coverage, but the runtime budget still grows up to it.
- Map size grows from MAX_CPUS * 16 * 4 = 64KB to MAX_CPUS * (MAX_CPUS >> 3) * 4 = 512KB at MAX_CPUS=1024.

src/bpf/main.bpf.c:

CLAMPS USE nr_cpu_ids - 1 AS THE RUNTIME CEILING
- pcpu_spill_search_budget and affinity_search_online clamp to min(nr_cpu_ids - 1, MAX_AFFINITY_CANDIDATES). The runtime ceiling reflects actual topology; the compile-time ceiling is the verifier- safe upper bound. At 12C the clamp is 11; at 32C it's 31; at 128C it's 127; beyond 128C it caps at MAX_CPUS >> 3 (= 128 today).

src/bpf_intf.rs (new):

SINGLE SOURCE OF TRUTH FOR INTF.H CONSTANTS
- New module mirroring MAX_CPUS, MAX_AFFINITY_CANDIDATES, MAX_NODES. Replaces the literal `16` mirrors at topology.rs:402 and scheduler.rs:335. const_assert in scheduler.rs catches drift.

src/topology.rs::populate_affinity_rank_map:

EMITS FULL SORTED RANK LIST
- Drops `let max_candidates = self.nr_cpus.min(16);` truncation. Emits the full sorted list capped at MAX_AFFINITY_CANDIDATES, with explicit u32::MAX sentinels for slots beyond nr_cpu_ids - 1 so BPF early-exit fires correctly (zero-init aliases to "CPU 0" otherwise).

VALIDATION
- bench-scale clean sweep on 12C (BPF + ADAPTIVE, 2/4/8/12 cores): zero violations, zero stress_exit. Single v5.9.0 run vs the 5.8.0 24-run mean improves Burst P99 (392us -> 83us, -4.7x), Long-run P99 (489us -> 244us, -2.0x), Mixed P99 (724us -> 139us, -5.2x), IPC P99 (107us -> 36us, -3.0x), deadline miss (3.56% -> 0.1%, -35x). Schbench P99 holds (75us -> 77us). App-launch regressed (2983us -> 4692us, n=1; under-sample suspect).
- 32C affinity-stranding class is not reproducible on 12C without the cgroup-cpuset churn libvirt produces; awaiting Roland's bench-scale on his Threadripper as the next signal.


sirlucjan (approved):
I've tested few days. Everything was fine. 

Image

PR #3562, opened 2026-05-10
scx_cargo: Enforce pahole >= 1.26

arighi:
Schedulers built against a kernel produced by pahole < 1.26 fail at runtime with a confusing libbpf error:
```
  libbpf: extern (func ksym) 'scx_bpf_cpu_rq': func_proto [N] incompatible with vmlinux [M]
  libbpf: failed to load BPF skeleton 'bpf_bpf': -EINVAL
```
The root cause is that pahole < 1.26 does not emit DECL_TAG entries for __bpf_kfunc functions (it is missing the --btf_features=decl_tag_kfuncs option that scripts/Makefile.btf only enables for pahole >= 1.26).

Without those tags, resolve_btfids short-circuits in collect_kfuncs() and never splits a KF_IMPLICIT_ARGS kfunc into its public-facing prototype (without the implicit 'struct bpf_prog_aux *') and the matching _impl variant (with the full prototype). The visible BTF prototype keeps the implicit argument, so when libbpf compares it against the BPF program's 1-arg declaration the two are flagged as incompatible and the skeleton fails to load.

Many sched_ext kfuncs are affected (scx_bpf_cpu_rq, scx_bpf_create_dsq, scx_bpf_kick_cpu, ...). The most common environment hitting this is Ubuntu 24.04 LTS, which ships pahole 1.25. The kernel itself bumped its documented minimum to 1.26 in commit 9edd04c4189e ("docs: Raise minimum pahole version to 1.26 for KF_IMPLICIT_ARGS kfuncs").

Catch this at build time to get a clear error instead of chasing a quite obscure runtime BPF load failure. Add a pahole version check to BpfBuilder::new() that all scheds_ext schedulers go through:

  - pahole not installed: skip silently (cross-build is a valid case and the local pahole is irrelevant).
  - pahole >= 1.26: silent pass.
  - pahole < 1.26: bail with a clear message explaining the failure mode, the affected distros, and the kernel commit reference.
  - SCX_SKIP_PAHOLE_CHECK=1 in env: skip the check, for cases where the running kernel was built on a different machine with a newer pahole.

### DIFF
--- a/rust/scx_cargo/src/bpf_builder.rs
+++ b/rust/scx_cargo/src/bpf_builder.rs
@@ -249,6 +249,8 @@ impl BpfBuilder {
     /// configure and `build` method to compile and generate bindings. See
     /// the struct documentation for details.
     pub fn new() -> Result<Self> {
+        crate::pahole_info::check()?;
+
         let out_dir = PathBuf::from(env::var("OUT_DIR")?);
 
         let clang = ClangInfo::new()?;
--- a/rust/scx_cargo/src/lib.rs
+++ b/rust/scx_cargo/src/lib.rs
@@ -6,5 +6,7 @@
 mod clang_info;
 pub use clang_info::ClangInfo;
 
+mod pahole_info;
+
 mod bpf_builder;
 pub use bpf_builder::BpfBuilder;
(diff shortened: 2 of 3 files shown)


sirlucjan (approved):

likewhatevs (approved):
i vaguely recall this needing to be higher, but this is way better an error than before.

PR #3566, opened 2026-05-11
utils: move get_primary_cpus and Powermode enum to scx_utils

Abhinayckrs:
Resolves a long-standing TODO present in scx_bpfland, scx_flash,
scx_cosmos and scx_beerland: consolidate the duplicated Powermode
enum and get_primary_cpus() function into scx_utils.

Both are moved to scx_utils::topology, alongside CoreType which
they depend on. They are exported from scx_utils and all four
schedulers updated to use the shared implementation.

No functional changes.


hodgesds:
I think this looks good overall, need to appease the linters. Make sure to run `cargo fmt --check`

Abhinayckrs:
Thanks! Applied cargo fmt, should be clean now.

hodgesds (approved):

PR #3567, opened 2026-05-11
scx_mitosis: Introduce subcell support and profile config ingestion

kkdwvd:
Stacked on top of #3513.

Adds the meat of the logic to support demand-weighted subcells, parsing of configs generated from profiling workloads. Topological awareness is still missing, but the rest is there.


kkdwvd:
Merged in [mitosis-next](https://github.com/sched-ext/scx/commits/mitosis-next).

PR #3569, opened 2026-05-12
scx_mitosis: add ktstr stall test

likewhatevs:
add a couple of ktstr stall tests to mitosis. these are feature and dev-dependency gated.

these are both seen in the wild. 

one happens with kthreads, the other happens with imbalanced workloads.

```
cargo install ktstr # install cargo plugin/needed tooling
cargo ktstr test --kernel ../linux  --features ktstr-tests -E 'test(/ktstr/)'  # run tests/get dumps
```

there is a brief description of what happens wrt/ these tests here: https://likewhatevs.github.io/ktstr/guide/tutorial.html

infra is mostly (i.e. minus turn on button) in place such that, once we have these tests passing, we can add a ci job that more/less runs that command on arm/x86.

applying https://lore.kernel.org/sched-ext/20260512055632.1096713-1-patso@likewhatevs.io/T/#u to local kernel checkout enables getting chain-to-failure style outputs in the event that that helps (i.e. these are both stalls, but i can see how stall => error => chain-to-error could be helpful).

### DIFF
--- a/scheds/rust/scx_mitosis/tests/ktstr_mitosis_tests.rs
+++ b/scheds/rust/scx_mitosis/tests/ktstr_mitosis_tests.rs
@@ -1,0 +1,58 @@
+#![cfg(feature = "ktstr-tests")]
+
+use ktstr::prelude::*;
+
+#[derive(ktstr::Scheduler)]
+#[scheduler(
+    name = "mitosis",
+    binary = "scx_mitosis",
+    topology(1, 4, 4, 1),
+    cgroup_parent = "/ktstr",
+    sched_args = [
+        "--exit-dump-len", "1048576",
+        "--reject-multicpu-pinning",
+        "--cpu-controller-disabled",
+        "--enable-borrowing",
+        "--enable-rebalancing",
+        "--dynamic-affinity-cpu-selection",
+        "--enable-slice-shrinking",
+    ],
+)]
+#[allow(dead_code)]
+enum MitosisFlags {}
+
+// trigger stalls in cell 0 like happens w/ kthreads (i.e. khugepaged) for
+// certain workloads. run rebalance at a higher pace for timely test execution.
+#[ktstr_test(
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_mitosis/tests/ktstr_mitosis_tests.rs
```diff
@@ -0,0 +1,75 @@
...
+
+// kthread stall
+#[ktstr_test(
+    scheduler = MITOSIS_PAYLOAD,
+    duration_s = 30, watchdog_timeout_s = 5,
+    llcs = 1, cores = 16, threads = 2,
+    extra_sched_args = ["--rebalance-cooldown-s", "1", "--rebalance-threshold", "0.1"],
+    memory_mb = 96, workers_per_cgroup = 256,
+)]
+fn mitosis_root_cell_starvation(ctx: &Ctx) -> Result<AssertResult> {
```
dforsyth, line 33: Can we have some more explanation around what each of these are testing?

likewhatevs: dropped the other one (was for an older version of mitosis) + added a description.

dforsyth: Did this update? I still don't see much of an explanation -- it references a stall but it does not explain what was happening or how this tests it. I think if we're going to bring these tests on we should have very good descriptions/comments so that we can get rid of them as we eliminate vulnerable code paths.

likewhatevs: I need to make this draft (think syntax changed/improved w/ latest update), but more generally, I think we don't want to remove tests as vulnerable paths are removed because tests are representative of work patterns which are largely independent of scheduler.

That also means things need to be better documented.

likewhatevs:
reopening one w/ fix + reproducer/test

PR #3571, opened 2026-05-12
scx_mitosis: add fix and test for cell0 starvation

likewhatevs:
scx_mitosis: add fix and test for cell0 starvation + have this not break ci.

In a nutshell, the fix is that phase 2 exclusive cpuset assignments are capped at compute_targets so surplus CPUs flow back to under-allocated cells instead of starving cell 0.


tommy-u:
Think you'll have to separate the fix out from the 4k line changes

likewhatevs:
> Think you'll have to separate the fix out from the 4k line changes

Yeah this is 99% Cargo.lock update, opened https://github.com/sched-ext/scx/pull/3569 to get that over with.

Edit -- i counted it, 100 loc ignoring cargo.lock.

PR #3573, opened 2026-05-13
scx_lavd: Add BPF Streams support

etsal:


scheds/include/lib/alloc/bpf_helpers_local.h
```diff
@@ -10,17 +9,32 @@
...
+											\
+	___ret;										\
 })
 
-#endif /* __BPF_HELPERS_LOCAL__ */
+#define arena_stdout(fmt, ...) bpf_stream_printk(1, (fmt), ##__VA_ARGS__)
+#define arena_stderr(fmt, ...) bpf_stream_printk(2, (fmt), ##__VA_ARGS__)
+
+#define arena_stdout_linfo(fmt, ...) bpf_stream_printk(1, "%s:%d " fmt, __func__, __LINE__, ##__VA_ARGS__)
+#define arena_stderr_linfo(fmt, ...) bpf_stream_printk(2, "%s:%d " fmt, __func__, __LINE__, ##__VA_ARGS__)
```
htejun: Can we use more generic name? e.g. `stream_info()`, `stream_err()`, `stream_info_loc()`. or maybe just `scx_info()`, `scx_err()` and so on.

lib/sdt_alloc.bpf.c
```diff
@@ -241,12 +242,12 @@ static sdt_desc_t *scx_alloc_chunk(struct scx_alloc_stack __arena *stack)
 static int pool_set_size(struct sdt_pool *pool, __u64 data_size, __u64 nr_pages)
 {
 	if (unlikely(data_size % 8)) {
-		bpf_printk("%s: allocation size %llu not word aligned", __func__, data_size);
+		arena_stderr_linfo("%s: allocation size %llu not word aligned", __func__, data_size);
```
htejun: This wouldn't need explicit `__func__` anymore. Other places too.

scheds/include/lib/alloc/bpf_helpers_local.h
etsal: We can keep the arena_stderr (and rename to scx_stderr)  if we want to commit to always printing line/func numbers. I think it's reasonable, if anyone wants to emit text without the code location they can make a new macro.

lib/sdt_alloc.bpf.c
etsal: Oh that's wrong, `__func__` should be getting omitted with the _linfo() variants since they append it themselves.

kkdwvd (approved):
LGTM.

daidavid (approved):

PR #3574, opened 2026-05-13
scx_lavd: Gate dsq_peek_task_load behind no-fast-lb

daidavid:
Some production kernels are missing fixes[1][2] for peek operation where the task_struct pointers are stale and we can run into UAF kernel crashes when the check for the race window is large enough:

  bpf_task_storage_get+0x36/0x110
  bpf_prog_81e5a6181f79e459_scx_task_data+0x41/0x58
  bpf_prog_9bcd89b94a3f9fb2___get_task_ctx_slowpath+0x26/0x79
  bpf_prog_72d45f938176613a_dsq_peek_task_load+0x72/0x8b
  bpf_prog_0514a0e1f88fcbdc_consume_task+0x730/0x852
  bpf_prog_7509fba73b418dbd_lavd_dispatch+0x208/0x679
  bpf__sched_ext_ops_dispatch+0x47/0xa3
  balance_scx.llvm.5618296179579790706+0x1d2/0xf30
  __schedule+0x2c9/0xb70
  schedule+0x38/0x90
  schedule_hrtimeout_range+0x75/0xf0
  __x64_sys_epoll_wait+0x753/0x920
  do_syscall_64+0x6a/0x250
  entry_SYSCALL_64_after_hwframe+0x4b/0x53

In theory, the same race can occur for the common pattern of getting the vtime as well after checking for the p pointer, but the timing window is far short to observe in production.

While patching the kernels are in progress, gate the dsq_peek_task_load behind no-fast-lb to make forward progress and retest once kernel side fixes are widely available.

[1] https://lore.kernel.org/all/20251024220102.1556384-1-arighi@nvidia.com/
[2] https://lore.kernel.org/all/20260424204418.3809733-8-tj@kernel.org/

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -383,7 +383,9 @@ static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
 
 			/*
 			 * Peek at the head task to get its size for budget
-			 * accounting.
+			 * accounting. Skip the peek when no_fast_lb is set
+			 * since the budget path below is bypassed and the
+			 * value would be unused.
 			 *
 			 * TOCTOU: the task peeked here may not be the one
 			 * actually consumed by consume_dsq() below. To be more
@@ -394,7 +396,7 @@ static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
 			 * because the next LB round recomputes budgets from
 			 * scratch.
 			 */
-			u64 task_load = dsq_peek_task_load(dsq_id);
+			u64 task_load = no_fast_lb ? 0 : dsq_peek_task_load(dsq_id);
 
 			/*
 			 * On success, decrement both egress and ingress
@@ -468,9 +470,12 @@ static bool force_to_steal_task(struct cpdom_ctx *cpdomc)
 			dsq_id = pick_most_loaded_dsq(cpdomc_pick);
 
 			/*
-			 * Peek at the head task to get its size.
+			 * Peek at the head task to get its size. Skip the
+			 * peek when no_fast_lb is set since the budget
+			 * accounting below is bypassed and the value would
+			 * be unused.
 			 */
-			u64 task_load = dsq_peek_task_load(dsq_id);
+			u64 task_load = no_fast_lb ? 0 : dsq_peek_task_load(dsq_id);
 
 			/*
 			 * Force steal is unconditional for work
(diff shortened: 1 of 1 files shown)


etsal (approved):

multics69:
What about gating at the __COMPAT_scx_bpf_dsq_peek level? It would be easier since we can close the potential UAF window:

```
diff --git a/scheds/include/scx/compat.bpf.h b/scheds/include/scx/compat.bpf.h
index 2089f0124..b181cf332 100644
--- a/scheds/include/scx/compat.bpf.h
+++ b/scheds/include/scx/compat.bpf.h
@@ -100,7 +100,8 @@ static inline struct task_struct *__COMPAT_scx_bpf_dsq_peek(u64 dsq_id)
     struct task_struct *p = NULL;
     struct bpf_iter_scx_dsq it;
 
-    if (bpf_ksym_exists(scx_bpf_dsq_peek))
+    if (bpf_ksym_exists(scx_bpf_dsq_peek) &&
+        (LINUX_KERNEL_VERSION >= KERNEL_VERSION(7, 1, 0)))
         return scx_bpf_dsq_peek(dsq_id);
     if (!bpf_iter_scx_dsq_new(&it, dsq_id, 0))
         p = bpf_iter_scx_dsq_next(&it);
```

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -394,7 +396,7 @@ static bool try_to_steal_task(struct cpdom_ctx *cpdomc)
 			 * because the next LB round recomputes budgets from
 			 * scratch.
 			 */
-			u64 task_load = dsq_peek_task_load(dsq_id);
+			u64 task_load = no_fast_lb ? 0 : dsq_peek_task_load(dsq_id);
```
bboymimi, line 399: LGTM.

scheds/rust/scx_lavd/src/bpf/balance.bpf.c
```diff
@@ -468,18 +470,23 @@ static bool force_to_steal_task(struct cpdom_ctx *cpdomc)
...
 
 			/*
 			 * Force steal is unconditional for work
 			 * conservation. Decrement budgets to keep
 			 * the accounting consistent.
 			 */
 			if (consume_dsq(cpdomc_pick, dsq_id)) {
-				decrement_stealee_budget(cpdomc_pick, task_load);
-				decrement_stealer_budget(cpdomc, task_load);
+				if (!no_fast_lb) {
```
bboymimi, line 486: LGTM. Thanks for the cleanup.

bboymimi (approved):

daidavid:
Spoke to Changwoo offline with the following thoughts, even with peek using the locked case the follow race can occur:

1. We get a valid dsq entry with a struct_task pointer while we have the dsq_lock(but then dropped after returning)
2. We check p and it's valid
3. Another cpu consumes the task from the DSQ, and then the task dies.
4. We dereference the struct_task pointer and then we could potentially run into the same UAF

I need to ponder a bit more, but at worst this patch cleans up the extra code when we use --no-fast-lb.

daidavid:
CI is reproducing an unrelated(known issue), so bypassing.

PR #3576, opened 2026-05-14
scx_lavd: Wrap peek use cases for safe pointer access with RCU lock

bboymimi:
As the commit dsq_peek_task_load gating (ac863374ce4f) noted, the
task_struct can be stale. On the other hand, accessing the task_struct
pointer from __COMPAT_scx_bpf_dsq_peek is not safe and the pointer can
be freed nondeterministically. The race window exists for both
peek_dsq_vtime and dsq_peek_task_load. This patch fixes both.


etsal (approved):

bboymimi:
I'll close the PR as we discussed, there's already rcu protection around non-sleepable BPF progs—making the PR unnecessary. Thank you all for the help in the process.

PR #3577, opened 2026-05-14
common.bpf.h: probe BPF prolog migrate_disable() from non-sleepable context

multics69:
This series fixes a probe in common.bpf.h that mischaracterised the
BPF trampoline prolog.  scx_lib_init() probed migration_disabled from
ops.init() -- a sleepable struct_ops context whose prolog always
disables migration -- but consumers of is_migration_disabled() run
under the non-sleepable prolog, which on modern kernels does not.
The wrong probe value caused is_migration_disabled() to false-negative
for p == current, and schedulers with SCX_OPS_ENQ_MIGRATION_DISABLED
(notably scx_lavd) to dispatch migration-disabled tasks cross-CPU via
SCX_DSQ_LOCAL_ON, self-disabling on task_can_run_on_remote_rq().

Probe from a non-sleepable context
----------------------------------

Replace scx_lib_init() with scx_lib_init_probe, an fentry program on
bpf_scx_reg() -- the .reg callback in bpf_sched_ext_ops, whose vtable
address makes it non-inlinable.  The kernel calls bpf_scx_reg during
scheduler registration before ops.init(), entering it through the
non-sleepable prolog.  scx_ops_attach! already auto-attaches via
$skel.attach() before attach_struct_ops(), so no new userspace code
is needed.

- [1/2] c98c5ad22 common.bpf.h: probe BPF prolog migrate_disable() from non-sleepable context

Drop scx_lib_init() and its callers
-----------------------------------

The helper is no longer needed since the probe runs automatically.
Remove it and drop the call from each in-tree scheduler's ops.init().

- [2/2] 4f648f26e scx: drop scx_lib_init() and its callers from BPF

Fixes: a6101f6c277f ("compat.bpf.h: Add scx_lib_init() to reliably probe BPF prolog migrate_disable() behavior")

### DIFF
--- a/scheds/experimental/scx_flow/src/bpf/main.bpf.c
+++ b/scheds/experimental/scx_flow/src/bpf/main.bpf.c
@@ -1114,10 +1114,6 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(flow_init)
 	s32 cpu;
 	s32 ret;
 
-	ret = scx_lib_init();
-	if (ret)
-		return ret;
-
 	nr_cpu_ids = scx_bpf_nr_cpu_ids();
 
 	bpf_for(cpu, 0, nr_cpu_ids) {
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -547,16 +547,6 @@ int scx_lib_init_probe(void *ctx)
 }
 
 /*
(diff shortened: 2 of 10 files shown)


arighi (approved):

multics69:
Thanks @arighi for the review.

PR #3578, opened 2026-05-14
scx_lavd: harden placement for migrate-disabled and throttled tasks

multics69:
This series hardens scx_lavd's placement decisions for tasks whose
effective cpumask is narrower than p->nr_cpus_allowed suggests:
migrate_disable() narrows p->cpus_ptr without touching nr_cpus_allowed,
and sched_setaffinity() on a task parked in the cpu.max BTQ changes
the allowed set without notifying the scheduler.

Distinguish permanent vs effective pinning
------------------------------------------

is_pinned() tested p->nr_cpus_allowed == 1 and silently missed tasks
narrowed only via migrate_disable() (e.g., unbound kworkers in short
migrate_disable sections).  Rename it to is_permanently_pinned(), add
is_effectively_pinned(taskc) backed by a "cpus_ptr weight == 1" flag
maintained at every ops.set_cpumask(), and partition call sites by
intent.

- [1/6] c9dc38b51 scx_lavd: distinguish permanent vs effective task pinning

Clean up cpu_ctx scratch fields
-------------------------------

Prep for the next section.  Switch try_find_and_kick_victim_cpu() off
the buffer init_idle_ato_masks() uses for the idle/active intersection
onto a general-purpose scratch, and rename the tmp_*_mask fields to
match the pick_ctx alias names that already describe each role.

- [2/6] 4fe93241d scx_lavd: use a more general scratch in try_find_and_kick_victim_cpu
- [3/6] 0272ec5b0 scx_lavd: rename cpu_ctx tmp_*_mask scratch fields for clarity

Harden placement under cpu.max throttling
-----------------------------------------

sched_setaffinity() on a BTQ-parked task narrows p->cpus_ptr silently,
so taskc->suggested_cpu_id can go stale by the time enqueue_cb()
drains the task back and lands it on a forbidden CPU.  Fix with two
complementary mechanisms: enqueue_cb() validates suggested_cpu_id at
drain time and falls back to scx_bpf_task_cpu(p) if stale;
lavd_set_cpumask() runs a fresh pick_idle_cpu() for throttled tasks
so the cache is refreshed at the cpumask-change event itself.

- [4/6] 1d151a925 scx_lavd: refresh CPU choice in enqueue_cb on BTQ reenqueue
- [5/6] 59aa8b879 scx_lavd: refresh placement on cpumask change for throttled tasks

Consolidate throttle checks
---------------------------

Hoist the "skip the scheduler's own task" guard into
cgroup_throttled() and route lavd_select_cpu()'s direct-dispatch path
through it, collapsing all four sites to a uniform
"if (enable_cpu_bw && cgroup_throttled(...) == -EAGAIN)".

- [6/6] cc8c51f80 scx_lavd: consolidate throttle checks in cgroup_throttled()

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
@@ -680,10 +680,20 @@ s32 pick_idle_cpu(struct pick_ctx *ctx, bool *is_idle)
 	 */
 	if (!init_active_ovrflw_masks(ctx))
 		goto err_out;
-	if (is_pinned(ctx->p) || is_migration_disabled(ctx->p)) {
+	/*
+	 * Use effective pinning here so we cover both permanent pinning
+	 * (nr_cpus_allowed == 1) and transient migrate_disable narrowing
+	 * (cpus_ptr weight == 1, cached via ops.set_cpumask).
+	 */
+	if (is_effectively_pinned(ctx->taskc) || is_migration_disabled(ctx->p)) {
 		cpu = ctx->prev_cpu;
 		if (!bpf_cpumask_test_cpu(cpu, cast_mask(ctx->active))) {
-			if (is_pinned(ctx->p))
+			/*
+			 * Extend the overflow set only for permanent pinning;
+			 * migrate_disable is transient, so we don't want to
+			 * pollute the overflow set with short-lived restrictions.
+			 */
+			if (is_permanently_pinned(ctx->p))
 				bpf_cpumask_test_and_set_cpu(cpu, ctx->ovrflw);
 		}
 		*is_idle = scx_bpf_test_and_clear_cpu_idle(cpu);
--- a/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/idle.bpf.c
@@ -35,7 +35,7 @@ bool init_idle_i_mask(struct pick_ctx *ctx, const struct cpumask *idle_cpumask)
 	if (!test_task_flag(ctx->taskc, LAVD_FLAG_IS_AFFINITIZED))
 		ctx->i_mask = idle_cpumask;
 	else {
-		struct bpf_cpumask *_i_mask = ctx->cpuc_cur->tmp_i_mask;
+		struct bpf_cpumask *_i_mask = ctx->cpuc_cur->i_mask;
 		if (!_i_mask)
 			return false;
 		bpf_cpumask_and(_i_mask, ctx->p->cpus_ptr, idle_cpumask);
@@ -69,8 +69,8 @@ bool init_ao_masks(struct pick_ctx *ctx)
 		return true;
 	}
 
-	ctx->a_mask = ctx->cpuc_cur->tmp_a_mask;
-	ctx->o_mask = ctx->cpuc_cur->tmp_o_mask;
+	ctx->a_mask = ctx->cpuc_cur->a_mask;
+	ctx->o_mask = ctx->cpuc_cur->o_mask;
 	if (!ctx->a_mask || !ctx->o_mask)
 		return false;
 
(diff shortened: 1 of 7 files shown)


daidavid:
The clean up patches look fine, but some of the series is predicated on this observation:

"A throttled task's task->cpus_ptr can change while it sits in the BTQ:
sched_setaffinity() on a queued task narrows the allowed mask but
the kernel doesn't notify scx_lavd through ops.dequeue/ops.enqueue
since the task stays runnable from the kernel's view."

I thought sched_setaffinity will go through dequeue if a task is queued, but maybe there's a path I missed. Can you describe the path in kernel where this occurs?

multics69:
> The clean up patches look fine, but some of the series is predicated on this observation:
> 
> "A throttled task's task->cpus_ptr can change while it sits in the BTQ: sched_setaffinity() on a queued task narrows the allowed mask but the kernel doesn't notify scx_lavd through ops.dequeue/ops.enqueue since the task stays runnable from the kernel's view."
> 
> I thought sched_setaffinity will go through dequeue if a task is queued, but maybe there's a path I missed. Can you describe the path in kernel where this occurs?

Thank you for pointing this out. You are right. `sched_setaffinity` triggers (1) `ops.dequeue()`, (2) `ops.set_cpumask()`, and (3) `ops.enqueue()` in sequence while holding a rq lock. However, the unthrottling path with `enqueue_cb()`  makes things complicated. Essentially, the reason is that `enqueue_cb()` does not hold the rq lock so `enqueue_cb()` and the sequence of ops calls can run concurrently. Hence, if  `enqueue_cb()` is called somewhere after (1) and before (3), `p->cpus_ptr` is updated, but the task's `suggested_cpu_id` is not. I updated the description to be clear for the commit, [scx_lavd: validate cached CPU pick in enqueue_cb on BTQ reenqueue](https://github.com/sched-ext/scx/pull/3578/commits/3c756d36f5f98a0d44d243e912f96781ba037ded).

While investigating the kernel call path closely, I found that [scx_lavd: refresh placement on cpumask change for throttled tasks](https://github.com/sched-ext/scx/commit/59aa8b879030ce187562ea3e310217b00e96cc89) is not necessary because when `ops.set_cpumask()` is called, the task is already canceled from the BTQ.

daidavid:
So I think the scenario described is false/misleading:

In the following sequence:

(1) ops.dequeue()
(2) ops.set_cpumask()
(3) ops.enqueue()

"Hence, if enqueue_cb() is called somewhere after (1) and before (3), p->cpus_ptr is updated, but the task's suggested_cpu_id is not."

enqueue_cb() can't occur for a task that's already popped off the btq.

There's a race that occurs around scx_atq_pop where if enqueue_cb() occurs before ops.dequeue():
1. cbw_drain_btw_batch wins the scx_atq_lock.
2. ops.dequeue() becomes a no-op.
3. enqueue_cb reads an outdated cpumask.
4. ops.set_cpumask updates the new cpu_mask

However, this patch reduces the window but doesn't actually address the race, we can still read an outdated cpu_mask in enqueue_cb.

This is scenario a bit far fetched, but plausible. Do we have a reproducer for this? We should see a stalled task sitting in a queue some where if we have a stressor that constantly calls setaffinity on a throttled workload?

multics69:
> So I think the scenario described is false/misleading:
> 
> In the following sequence:
> 
> (1) ops.dequeue() (2) ops.set_cpumask() (3) ops.enqueue()
> 
> "Hence, if enqueue_cb() is called somewhere after (1) and before (3), p->cpus_ptr is updated, but the task's suggested_cpu_id is not."
> 
> enqueue_cb() can't occur for a task that's already popped off the btq.
> 
> There's a race that occurs around scx_atq_pop where if enqueue_cb() occurs before ops.dequeue():
> 
> 1. cbw_drain_btw_batch wins the scx_atq_lock.
> 2. ops.dequeue() becomes a no-op.
> 3. enqueue_cb reads an outdated cpumask.
> 4. ops.set_cpumask updates the new cpu_mask
> 
> However, this patch reduces the window but doesn't actually address the race, we can still read an outdated cpu_mask in enqueue_cb.

Your understanding is correct. Essentially, it checks if the chosen CPU is still valid and if not, chooses the one in the most conservative way.

> 
> This is scenario a bit far fetched, but plausible. Do we have a reproducer for this? We should see a stalled task sitting in a queue some where if we have a stressor that constantly calls setaffinity on a throttled workload?

I don't have a specific reproducer but I saw this once in my setup running stress test script and containter all together.

daidavid (approved):

multics69:
Thanks @daidavid for review.

PR #3580, opened 2026-05-15
compat.bpf.h: Gate scx_bpf_dsq_peek kfunc behind kernel version 7.1.0

bboymimi:
__COMPAT_scx_bpf_dsq_peek() selects the lockless kfunc whenever the
symbol resolves in the kernel BTF. However, its lockless implementation
could return a stale task_struct pointer. The stale pointer issues are
resolved only after v7.1 kernel with the following patches:

1). 2f2ea7709266 ("sched_ext: Use dsq->first_task instead of list_empty() in dispatch_enqueue() FIFO-tail")
2). 71d7847cad44 ("sched_ext: Fix scx_bpf_dsq_peek() with FIFO DSQs")

Require bpf_ksym_exists(scx_bpf_dsq_peek) AND LINUX_KERNEL_VERSION >=
KERNEL_VERSION(7, 1, 0) before calling the kfunc to mitigate the issue;
otherwise fall through to the existing bpf_iter_scx_dsq path instead.

See also the lavd patch, working around the bug by avoiding calling
the kfunc when unnecessary and having more context explanation:
ac863374ce4f ("scx_lavd: Gate dsq_peek_task_load behind no-fast-lb")

PR #3585, opened 2026-05-17
compat.bpf.h: Add override for scx_bpf_dsq_peek version gate

bboymimi:
a3f2784c432b ("compat.bpf.h: Gate scx_bpf_dsq_peek kfunc behind kernel
version 7.1.0") gates the lockless peek kfunc by LINUX_KERNEL_VERSION,
but the version check misses stable backports. A kernel that has both
upstream FIFO fixes

  71d7847cad44 ("sched_ext: Fix scx_bpf_dsq_peek() with FIFO DSQs")
  2f2ea7709266 ("sched_ext: Use dsq->first_task instead of list_empty()
                  in dispatch_enqueue() FIFO-tail")

cherry-picked but still reporting < 7.1.0 ends up on the iterator
fallback. CachyOS users on 6.18.31 reported a 4x FPS drop in Cyberpunk
2077.

Add __COMPAT_scx_allow_lockless_peek_unsafe as a __weak rodata bool.
When true, bpf_ksym_exists() alone is enough to take the fast path.
Expose it in scx_lavd as --allow-lockless-peek-unsafe. Default false,
so kernels without the fixes stay on the iterator path. Only enable
on kernels that have both fixes backported.

### DIFF
--- a/scheds/include/scx/compat.bpf.h
+++ b/scheds/include/scx/compat.bpf.h
@@ -90,6 +90,8 @@ int bpf_cpumask_populate(struct cpumask *dst, void *src, size_t src__sz) __ksym 
 	(bpf_ksym_exists(bpf_cpumask_populate) ?			\
 	 (bpf_cpumask_populate(cpumask, src, size__sz)) : -EOPNOTSUPP)
 
+__weak const volatile bool __COMPAT_scx_allow_lockless_peek_unsafe = false;
+
 /*
  * v6.19: Introduce lockless peek API for user DSQs.
  * v7.1:  Resolve the stale pointer issue of the lockless peek API.
@@ -98,6 +100,9 @@ int bpf_cpumask_populate(struct cpumask *dst, void *src, size_t src__sz) __ksym 
  * return stale task pointers. Require kernel version >= 7.1.0 before calling
  * it; otherwise fall through to the bpf_iter_scx_dsq fallback below.
  *
+ * The version gate can be bypassed at runtime by setting
+ * __COMPAT_scx_allow_lockless_peek_unsafe if the bug fixes have been backported
+ * to the LTS kernel.
  */
 static inline struct task_struct *__COMPAT_scx_bpf_dsq_peek(u64 dsq_id)
 {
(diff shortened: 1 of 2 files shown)


daidavid:
I'm not too convinced that using non-peek can cause this level of regression or else we'd have very poor performance for older kernels and/or older LAVD versions. Do we understand what causes the perf drop and can we replicate the regression in other microbenchmarks?

multics69:
Let's wait until we collect more evidence of the performance regression.

ms178:
Thanks a lot for the patch which I've just tested with the latest scx-scheds-git main branch on Kernel 6.18.32 in these two scenarios:

Scenario A: LAVD without `--allow-lockless-peek-unsafe` --> this yielded the previously reported massive performance regression and massive stuttering

Possibly related dmesg output:
```
[19. Mai 10:07] NOHZ tick-stop error: local softirq work is pending, handler #80!!!
[  +0,000000] NOHZ tick-stop error: local softirq work is pending, handler #200!!!
[  +0,000554] NOHZ tick-stop error: local softirq work is pending, handler #200!!!
```
Scenario B: LAVD with `--allow-lockless-peek-unsafe`  --> this brought back the expected performance level

multics69:
> Thanks a lot for the patch which I've just tested with the latest scx-scheds-git main branch on Kernel 6.18.32 in these two scenarios:
> 
> Scenario A: LAVD without `--allow-lockless-peek-unsafe` --> this yielded the previously reported massive performance regression and massive stuttering
> 
> Possibly related dmesg output:
> 
> ```
> [19. Mai 10:07] NOHZ tick-stop error: local softirq work is pending, handler #80!!!
> [  +0,000000] NOHZ tick-stop error: local softirq work is pending, handler #200!!!
> [  +0,000554] NOHZ tick-stop error: local softirq work is pending, handler #200!!!
> ```
> 
> Scenario B: LAVD with `--allow-lockless-peek-unsafe` --> this brought back the expected performance level

What's the hardware you are using? It sounds like a pretty beefy machine with many CPUs?

ms178:
> What's the hardware you are using? It sounds like a pretty beefy machine with many CPUs?

@multics69 As mentioned in https://github.com/sched-ext/scx/pull/3580#issuecomment-4485533542, I use an Intel 14700KF (Raptor Lake) with a Z690 motherboard, 64 GB DDR4-3600, AMD Vega 64. The CPU comes with 8 P-Cores and 12 E-Cores.

PR #3594, opened 2026-05-22
scx_layered: allow idle_confined spillover when system is fully allocated

hodgesds:
`idle_confined` gates open CPU search on `check_no_idle`, a per-layer flag that clears whenever ANY idle CPU is found within the layer's allocated CPUs.  This BPF-side check is too restrictive: on CPUs degraded by softirq or IRQ overhead (e.g. ~50% softirq from network processing), brief idle moments between bursts keep `check_no_idle` cleared even though the CPUs are effectively saturated.  The layer stays trapped on degraded CPUs with no way to spill to other layers' idle capacity.

When all CPUs are assigned to layers, `idle_confined` layers have nowhere to grow — the only relief valve is spillover.  Add a `fully_allocated` flag to the BPF layer struct, set by userspace when total allocated CPUs >= total system CPUs.  When set, `idle_confined` layers act as a regular `Grouped` layer for idle CPU selection.

When the system has unallocated CPUs, `idle_confined` continues to act like Confined for cache locality.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/intf.h
+++ b/scheds/rust/scx_layered/src/bpf/intf.h
@@ -397,6 +397,7 @@ struct layer {
 	bool			skip_remote_node;
 	bool			prev_over_idle_core;
 	bool			idle_confined;
+	bool			fully_allocated;
 	int			growth_algo;
 
 	u64			nr_tasks;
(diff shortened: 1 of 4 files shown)


likewhatevs:
no tests of the change?

likewhatevs (approved):

PR #3595, opened 2026-05-23
scx_lavd: fix delayed scheduling of turbulent-DSQ tasks

multics69:
Two cross-cpdom paths skip the per-cpdom turbulent DSQ
(cpdom_to_turb_dsq) when scanning for work.  Tasks routed there
become invisible to load-balancing and dispatch, raising their
scheduling latency.

Stealer path
------------

pick_most_loaded_dsq() doesn't include the turbulent cpdom DSQ
in its depth comparison.  A cpdom whose backlog sits in its
turbulent DSQ looks empty to stealers, so its turbulent tasks
are not migrated to an under-loaded neighbor cpdom.  Fix:
compare both cpdom DSQ depths and pick the deeper; tie ->
non-turbulent first, matching the existing consume priority.

- [1/2] scx_lavd: include turbulent cpdom DSQ in pick_most_loaded_dsq()

Dispatch overflow extension
---------------------------

When lavd_dispatch() runs on a CPU outside the active+overflow
set, it scans only the non-turbulent cpdom DSQ for a
pinned/affinitized task that only this CPU can run.  A
non-vulnerable pinned task routed to the turbulent DSQ (per
get_target_dsq_id()) is missed when this CPU is its only legal
target, delaying its dispatch.  Fix: extract the body into
scan_dsq_for_ovflw_ext() and call it once per cpdom DSQ flavor.

- [2/2] scx_lavd: scan turbulent cpdom DSQ in dispatch overflow extension

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/balance.bpf.c
@@ -512,21 +512,29 @@ static bool force_to_steal_task(struct cpdom_ctx *cpdomc)
 }
 
 __hidden
-bool consume_task(u64 cpu_dsq_id, u64 cpdom_dsq_id)
+bool consume_task(u64 cpdom_id)
 {
 	struct cpdom_ctx *cpdomc;
 	struct cpu_ctx *cpuc;
-	u64 cpdom_turb_dsq_id;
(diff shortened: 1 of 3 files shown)


multics69:
This changes conflict with https://github.com/sched-ext/scx/pull/3610. 
Will reopen once https://github.com/sched-ext/scx/pull/3610 is merged.

PR #3597, opened 2026-05-23
scx_layered: fix compute_target_llcs for asymmetric topologies

Abhinayckrs:
compute_target_llcs() assumed all LLCs have equal numbers of cores
and all cores have equal numbers of threads, using system-wide averages.
This gives wrong LLC and core counts on asymmetric hardware such as
AMD Ryzen (variable cores per LLC) and Intel hybrid (P-cores with SMT,
E-cores without).

Fix by walking the actual per-node LLC topology instead of using
integer division over system-wide totals. Update all three call sites
in recompute_layer_core_order() to pass the per-node LLC map.

Also fixes the avail computation in Phase 3 to use the actual LLC
core count rather than the precomputed uniform cpus_per_llc.

Resolves the TODO(kkd) comments at the original call site.

Note: a heterogeneous topology test helper in scx_utils::testutils
would be needed to explicitly cover this case. That is left as a
follow-up.

Tested: scx_layered loads and runs cleanly under virtme-ng (v6.12).

### DIFF
--- a/rust/scx_utils/src/topology.rs
+++ b/rust/scx_utils/src/topology.rs
@@ -1312,6 +1312,58 @@ pub mod testutils {
         (Topology::instantiate(span, nodes).unwrap(), total_cpus)
     }
 
+    pub fn make_het_test_topo(
+        cores_per_llc: &[&[usize]],
+        hts_per_core: usize,
+    ) -> (Topology, usize) {
+        let nr_nodes = cores_per_llc.len();
+
+        let total_cpus = cores_per_llc
+            .iter()
+            .map(|node| node.iter().sum::<usize>())
+            .sum::<usize>()
+            * hts_per_core;
+
+        set_cpumask_test_width(total_cpus);
+
+        let mut cpu_id = 0usize;
+        let mut core_id = 0usize;
+        let mut llc_id = 0usize;
+        let mut nodes = BTreeMap::new();
+
+        for (node_id, llcs_in_node) in cores_per_llc.iter().enumerate() {
+            let mut llcs = BTreeMap::new();
+
+            for &cores_in_llc in *llcs_in_node {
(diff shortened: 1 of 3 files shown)


hodgesds:
Might be worth adding some tests as well?

Abhinayckrs:
Hi, thanks for the review!
I've added make_het_test_topo to scx_utils::testutils and a heterogeneous test test_compute_target_llcs_het_single_node. However, updating the function signature to accept &BTreeMap<usize, Arc<Llc>> instead of &Topology broke all existing callers in lib.rs — about 25 call sites across tests and simulation code.
Before I fix all of them, I wanted to check: does the overall approach look correct to you? Specifically:

Walking actual per-LLC CPU counts via llc.span.weight() instead of uniform division
Passing per-node LLCs instead of the whole topology for NUMA correctness
Walking cores in the partial LLC to get exact extra core count

If the approach is sound I'll update all the lib.rs callers and existing tests. If you'd prefer a narrower fix I can keep the &Topology signature and use topo.all_llcs internally — that fixes the uniform assumption while minimizing caller changes. Or am I missing something?

hodgesds:
I think in general it looks ok. I refactored a bit of the the topology/allocator code so you may need to rebase.

Abhinayckrs:
Added tests

Introduced `make_het_test_topo` in `scx_utils::testutils` to support both single-node and multi-node heterogeneous topologies, and added two new tests: `test_compute_target_llcs_het` and `test_compute_target_llcs_het_multi_node` covering asymmetric LLC sizes.

Also updated all callers in `lib.rs` to match the new `compute_target_llcs()` signature.

Abhinayckrs:
Friendly ping — happy to make any further changes

hodgesds (approved):

PR #3583, opened 2026-05-15
scx_cosmos: Fix lib/pmu.bpf.c path

sirlucjan:
The following bug has been reported: https://github.com/sched-ext/scx/issues/3582 and it appears that https://github.com/sched-ext/scx/commit/df21501c4bf98d26ad0497e5f74010351991e377 introduced a bug (I checked, and the bug occurs with `cargo install scx_cosmos` in versions 1.1.0 and 1.1.1). scx_layered has the following solution: https://github.com/sched-ext/scx/blob/main/scheds/rust/scx_layered/build.rs#L13 so it’s worth using something similar in scx_cosmos. 

```
lucjan at cachyos ~/Patchownia/other/scx 23:44:10 605c7d2c for-next    
❯ cargo build --release -p scx_cosmos
warning: /home/lucjan/Patchownia/other/scx/rust/scx_rustland_core/Cargo.toml: unused manifest key: lib.include
   Compiling proc-macro2 v1.0.106
   Compiling quote v1.0.45
   Compiling unicode-ident v1.0.24
   Compiling libc v0.2.186
   Compiling serde_core v1.0.228
   Compiling serde v1.0.228
   Compiling crossbeam-utils v0.8.21
   Compiling autocfg v1.5.0
   Compiling cfg-if v1.0.4
   Compiling bitflags v2.11.1
   Compiling cfg_aliases v0.2.1
   Compiling shlex v1.3.0
   Compiling find-msvc-tools v0.1.9
   Compiling zmij v1.0.21
   Compiling memchr v2.8.0
   Compiling nix v0.31.3
   Compiling strsim v0.11.1
   Compiling serde_json v1.0.149
   Compiling cc v1.2.62
   Compiling pin-project-lite v0.2.17
   Compiling getrandom v0.4.2
   Compiling pkg-config v0.3.33
   Compiling rustix v1.1.4
   Compiling glob v0.3.3
   Compiling anyhow v1.0.102
   Compiling prettyplease v0.2.37
   Compiling once_cell v1.21.4
   Compiling thiserror v2.0.18
   Compiling linux-raw-sys v0.12.1
   Compiling camino v1.2.2
   Compiling parking v2.2.1
   Compiling clang-sys v1.8.1
   Compiling num-traits v0.2.19
   Compiling winnow v1.0.3
   Compiling log v0.4.29
   Compiling regex-syntax v0.8.10
   Compiling concurrent-queue v2.5.0
   Compiling syn v2.0.117
   Compiling fastrand v2.4.1
[...]

PR #3604, opened 2026-05-27
scx_cake: bump version + scx_cargo to fix crates.io install

rrnewton:
## Summary

Follow-up to #3583 / commits 97cf25fe + 11cb5f25 that landed the symlink-based path fix for `scx_cake` (the same pattern `scx_cosmos` adopted in commit 605c7d2c). That fix resolved the original

```
clang: error: no such file or directory: '../../../lib/arena.bpf.c'
```

blocking `cargo install scx_cake` from crates.io for 1.1.1. However, **`scx_cake` 1.1.2 still cannot install from crates.io** for a closely-related but distinct reason — surfaced only by `cargo package -p scx_cake` (verify step) end-to-end against the published deps:

```
src/bpf/lib/sdt_alloc.bpf.c:45:2: error: call to undeclared function 'scx_err_loc'
```

### Why

Unlike `scx_cosmos` (which only bundles `pmu.bpf.c`), `scx_cake` bundles `sdt_alloc.bpf.c` and `sdt_task.bpf.c` — both of which now use the `scx_err_loc` / `scx_err` / `scx_out_loc` macros introduced in commit 73b45430 (`lib: move allocators to BPF streams reporting`, 2026-05-13). Those macros live in `scheds/include/lib/alloc/bpf_helpers_local.h`, which ships inside the `scx_cargo` crate's `bpf_h` tarball.

But the `scx_cargo` on crates.io is still 1.1.1 (published 2026-05-13, before the streams refactor). Its bundled `bpf_helpers_local.h` predates the macros, so any `cake` 1.1.2 install from crates.io against the published `scx_cargo` 1.1.1 fails clang at sdt_alloc.

### Fix

Re-publish `scx_cargo` with the current (post-streams) `bpf_h`, and bump `scx_cake` to consume it:

- `rust/scx_cargo`: `1.1.1` → `1.1.2` (no code change — `bpf_h` is a symlink to `../../scheds/include`, resolved at `cargo package` time, so the bump alone carries the updated `bpf_helpers_local.h` to crates.io)
- `scheds/rust/scx_cake`: `1.1.2` → `1.1.3`, and bump build-dep `scx_cargo` `"1.1.1"` → `"1.1.2"`
- `Cargo.lock` updated to match

### Release ordering

`scx_cargo` 1.1.2 must be published to crates.io **before** `scx_cake` 1.1.3 so cargo can resolve the new build-dep version.

## Test plan

Local pre-publish simulation (the published `scx_cargo` 1.1.2 doesn't exist yet, so the standard `cargo package --verify` chicken-and-eggs — the workaround mirrors what `cargo install` will do post-publish):

- [x] `cargo build -p scx_cake` clean (workspace path deps)
- [x] `cargo package -p scx_cargo -p scx_cake --no-verify` produces both tarballs
- [x] Extract both tarballs side-by-side under `/tmp/cake-install-sim/`:
  - `scx_cargo-1.1.2/bpf_h/lib/alloc/bpf_helpers_local.h` contains `scx_err_loc` ✓
  - `scx_cake-1.1.3/src/bpf/lib/` contains all 8 lib `.bpf.c` files referenced by `build.rs` ✓
- [x] In extracted `scx_cake-1.1.3/`, inject

  ```toml
  [workspace]

  [patch.crates-io]
  scx_cargo = { path = "../scx_cargo-1.1.2" }
  ```

  then `cargo build` — succeeds end-to-end. `scx_err_loc` resolves via the new `bpf_h` tar that `scx_cargo` 1.1.2 ships, and all `add_source("src/bpf/lib/*.bpf.c")` paths resolve inside the flat extracted tarball. This is the most faithful local simulation of `cargo install scx_cake` post-publish: the contents inside the extracted tarballs are exactly what the registry would serve.
- [ ] Reviewer / release engineer: publish `scx_cargo` 1.1.2 first, then `scx_cake` 1.1.3, then `cargo install scx_cake` from a clean toolchain to confirm crates.io install path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3635,7 +3635,7 @@ dependencies = [
 
 [[package]]
 name = "scx_cake"
-version = "1.1.2"
+version = "1.1.3"
 dependencies = [
  "anyhow",
  "arboard",
@@ -3660,7 +3660,7 @@ dependencies = [
 
 [[package]]
 name = "scx_cargo"
-version = "1.1.1"
+version = "1.1.2"
 dependencies = [
  "anyhow",
  "bindgen 0.72.1",
--- a/rust/scx_cargo/Cargo.toml
+++ b/rust/scx_cargo/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scx_cargo"
-version = "1.1.1"
+version = "1.1.2"
 edition = "2021"
 authors = ["Tejun Heo <tj@kernel.org>"]
 license = "GPL-2.0-only"
--- a/scheds/rust/scx_cake/Cargo.toml
+++ b/scheds/rust/scx_cake/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scx_cake"
-version = "1.1.2"
+version = "1.1.3"
 authors = ["RitzDaCat"]
 edition = "2021"
 description = "A sched_ext scheduler applying CAKE bufferbloat concepts to CPU scheduling"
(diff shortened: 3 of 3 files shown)


hodgesds (approved):

sirlucjan (approved):
Thanks Ryan

PR #3605, opened 2026-05-27
mitosis: rescue cell 0 when child cpusets cover every CPU

likewhatevs:
mitosis bails when a cell ends up with zero CPUs. Cell 0 — the
catch-all that receives any CPUs no child cgroup's cpuset.cpus
claims — is squeezed to empty once child cpusets collectively cover
every host CPU, and the scheduler exits mid-recompute:

```
  running scheduler main loop
  Caused by:
    0: checking cpuset changes
    1: recomputing cell configuration after cpuset change
    2: computing demand-weighted CPU assignments
    3: Cell 0 has no CPUs assigned (nr_cpus=N, num_cells=M)
```

Add `--cell0-min-cpus <N>` (default 0): before the cpuset-based
assignment runs, reserve up to N CPUs as a holdout for cell 0 —
preferring CPUs no child cpuset claims and, when those run out,
stealing from the cells round-robin (one CPU per cell in turn,
largest cell first) but never a CPU that would leave a child cell
with none — including when cpusets overlap, since a shared CPU is
assigned to only one of its claimants. So the reservation is spread
across cells and can't starve a child into the same bail. (Within a
donor the pick prefers an exclusive CPU, then the LLC split across
the most cells, then the lowest CPU.) The held-out CPUs are skipped
by the assignment and pre-seeded onto cell 0, so a non-zero value
keeps cell 0 from being starved to zero; if N exceeds what can be
reserved without emptying a child, cell 0 simply gets fewer than N.
Default 0 disables the holdout; the existing bail is unchanged.

First commit adds a ktstr e2e reproducer for the bail; the second
adds the flag and the holdout.

Test plan
- [x] `--cell0-min-cpus 1`: ktstr repro
  mitosis_cell0_starved_by_full_coverage_cpusets passes — cell 0
  rescued, scenario runs to a clean shutdown, no "Cell 0 has no CPUs"
  bail and no stall
- [x] Default (`--cell0-min-cpus 0`): the bail is preserved —
  test_cpu_assignments_cpusets_cover_all_cpus
- [x] Holdout spreads the steal across cells (round-robin, largest
  first) — test_cpu_assignments_holdout_takes_from_largest_cell,
  test_cpu_assignments_holdout_steals_evenly_across_cells
- [x] An over-large `--cell0-min-cpus` never empties a child cell, with
  disjoint or overlapping cpusets —
  test_cpu_assignments_holdout_never_starves_a_child,
  test_cpu_assignments_holdout_overlapping_cpusets_no_starvation
- [x] 53 mitosis unit tests pass; cargo fmt --check clean

### DIFF
--- a/scheds/rust/scx_mitosis/src/cell_manager.rs
+++ b/scheds/rust/scx_mitosis/src/cell_manager.rs
@@ -242,7 +242,7 @@ impl CellManager {
         if !path.exists() {
             bail!("Cell parent cgroup path does not exist: {}", path.display());
         }
-        Self::new_with_path(
+        Self::new_with_path_opts(
             path,
             max_cells,
             all_cpus,
@@ -252,7 +252,20 @@ impl CellManager {
         )
     }
 
+    /// Test-only 4-argument constructor with the holdout defaulted off
+    /// (`cell0_min_cpus = 0`, empty `cpu_to_llc`). Tests that exercise the
+    /// holdout call [`Self::new_with_path_opts`] directly with those arguments.
+    #[cfg(test)]
     fn new_with_path(
+        path: PathBuf,
+        max_cells: u32,
+        all_cpus: Cpumask,
+        exclude: HashSet<String>,
+    ) -> Result<Self> {
+        Self::new_with_path_opts(path, max_cells, all_cpus, exclude, 0, HashMap::new())
+    }
+
+    fn new_with_path_opts(
         path: PathBuf,
         max_cells: u32,
         all_cpus: Cpumask,
@@ -605,12 +618,12 @@ impl CellManager {
             }
         }
 
-        // Holdout: reserve up to `cell0_min_cpus` CPUs for cell 0 before the
-        // cpuset-based assignment below. Prefer CPUs no child cpuset claims;
-        // when those run out, steal from cells — largest cell first, but never a
-        // cell's last CPU — so no child is starved (cell 0 may end up with fewer
-        // than requested). Reserved CPUs are skipped by the phases below and
-        // pre-seeded onto cell 0.
+        // Holdout: try to reserve `cell0_min_cpus` CPUs for cell 0 before the
+        // cpuset-based assignment below, but reserve fewer if reserving more
+        // would mean taking a workload cell's last CPU. Prefer CPUs no child
+        // cpuset claims; when those run out, steal from cells — largest cell
+        // first, but never a cell's last CPU — so no child is starved. Reserved
+        // CPUs are skipped by the phases below and pre-seeded onto cell 0.
         let mut holdout = Cpumask::new();
         if self.cell0_min_cpus > 0 {
             let mut unclaimed: Vec<usize> = self
@@ -1074,8 +1087,6 @@ mod tests {
             256,
             cpumask_for_range(16),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
 
@@ -1095,8 +1106,6 @@ mod tests {
             256,
             cpumask_for_range(16),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
 
@@ -1119,8 +1128,6 @@ mod tests {
             256,
             cpumask_for_range(16),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
         assert_eq!(mgr.cell_count(), 1);
@@ -1147,8 +1154,6 @@ mod tests {
             256,
             cpumask_for_range(16),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
         assert_eq!(mgr.cell_count(), 2);
@@ -1176,8 +1181,6 @@ mod tests {
             256,
             cpumask_for_range(16),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
         assert_eq!(mgr.cell_count(), 1);
@@ -1216,8 +1219,6 @@ mod tests {
             256,
             cpumask_for_range(16),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
 
@@ -1247,8 +1248,6 @@ mod tests {
             256,
             cpumask_for_range(16),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
 
@@ -1270,8 +1269,6 @@ mod tests {
             256,
             cpumask_for_range(16),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
         let assignments = mgr.compute_cpu_assignments(false).unwrap();
@@ -1297,8 +1294,6 @@ mod tests {
             256,
             cpumask_for_range(10),
             HashSet::new(),
-            0,
-            HashMap::new(),
         )
         .unwrap();
         let assignments = mgr.compute_cpu_assignments(false).unwrap();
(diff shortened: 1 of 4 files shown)


etsal (approved):

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -410,13 +418,20 @@ impl<'a> Scheduler<'a> {
...
         let cell_manager = CellManager::new(
             parent_cgroup,
             MAX_CELLS as u32,
             topology.span.clone(),
             exclude,
+            opts.cell0_min_cpus,
         )
-        .with_context(|| format!("initializing cell manager for cgroup {}", parent_cgroup))?;
+        .with_context(|| format!("initializing cell manager for cgroup {}", parent_cgroup))?
+        .with_cpu_to_llc(cpu_to_llc);
```
dforsyth: This seems weird, why not pass it in like everything else?

likewhatevs: that was just to minimize delta, changed to this instead

scheds/rust/scx_mitosis/src/main.rs
```diff
@@ -155,6 +155,15 @@ struct Opts {
     #[clap(long)]
     cell_exclude: Vec<String>,
 
+    /// Reserve up to this many CPUs for cell 0 (the root/catch-all cell)
+    /// before child cpusets are applied (the holdout). A non-zero value keeps
+    /// cell 0 from being starved to zero when child cgroups' cpusets cover
+    /// every CPU. The reservation never takes a child cell's last CPU, so cell 0
```
tommy-u, line 161: Is it strictly true that the cousets need to cover every cpu? I thought you could repro this when there were held out cpus.

likewhatevs: Where I've seen this exact error in the wild, yes. I have not seen this exact issue (the error in the PR) w/ holdout present.

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -228,19 +235,30 @@ impl CellManager {
...
         }
-        Self::new_with_path(path, max_cells, all_cpus, exclude)
+        Self::new_with_path(
+            path,
+            max_cells,
+            all_cpus,
+            exclude,
+            cell0_min_cpus,
+            cpu_to_llc,
+        )
```
tommy-u: Let's change this to 
```
        Self::new_with_path_opts(
            path,
            max_cells,
            all_cpus,
            exclude,
            cell0_min_cpus,
            cpu_to_llc,
        )
```

Which your 4 tests call directly, and all the others are unchanged because they call a wrapper

They call
```
    fn new_with_path(
        path: PathBuf,
        max_cells: u32,
        all_cpus: Cpumask,
        exclude: HashSet<String>,
    ) -> Result<Self> {
        Self::new_with_path_opts(path, max_cells, all_cpus, exclude, 0, HashMap::new())
    }
```

Then decorate with
```
    #[cfg(test)]
    fn new_with_path(
```

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -585,15 +605,137 @@ impl CellManager {
             }
         }
 
+        // Holdout: reserve up to `cell0_min_cpus` CPUs for cell 0 before the
```
likewhatevs: > Also, I think by "up to" you mean the number we reserve in this step can be lower than cell0_min_cpus because there may be unclaimed cpus to start with.

No -- by "up to", I mean that if there are 4 non-root cells (i.e. excluding cell 0 from those 4), 5 CPUs, and 2 CPUs are specified for holdout, holdout being 2 would trigger this error by leaving another cell without CPUs if we respected it ultimately. ATM, this PR would give only 1 CPU to cell 0 in this scenario (i.e. the "up to").

I think in that situation "up to" is less problematic than "must", but that's really a judgement call. Do "must" instead of "up to"?

tommy-u: Holdout tries to reserve `cell0_min_cpus` many cpus, but it may reserve fewer if that would mean taking a workload cell's last cpu.

scheds/rust/scx_mitosis/src/cell_manager.rs
likewhatevs: Good idea, thx!

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -585,15 +618,137 @@ impl CellManager {
...
+                    .and_then(|llc| llc_cells.get(llc))
+                    .map_or(0, |cells| cells.len())
+            };
+
+            // Take unclaimed CPUs first — they cost no cell anything.
+            let mut taken = 0usize;
+            for cpu in unclaimed.into_iter().take(self.cell0_min_cpus) {
+                holdout.set_cpu(cpu).ok();
+                taken += 1;
+            }
```
tommy-u, line 668: A future concern is picking hyper thread pairs.

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -585,15 +618,137 @@ impl CellManager {
...
+            // CPUs yielded to the holdout per cell, for round-robin selection.
+            let mut taken_from: HashMap<u32, usize> =
+                cell_claimed.keys().map(|&id| (id, 0usize)).collect();
+            while taken < self.cell0_min_cpus {
+                // Unreserved EXCLUSIVE CPUs per cell, recomputed as the holdout
+                // grows. Only exclusive CPUs count toward survival: a contested CPU
+                // may be awarded to a co-claimant in Phase 3, so it cannot be what
+                // keeps a cell alive -- counting it would let the steal take a
+                // cell's last exclusive CPU and zero the cell (the Phase-5 bail).
```
tommy-u, line 694: If we got into this loop, we are now taking claimed cpus from someone. Can you please add a "enforced_holdout" boolean that gets initialized to False when the scheduler starts up and gets or'ed to true here if we get into the body of this loop?

When we hear of some hosts getting wonky perf, I'd like to be able to query this to know if we've put our hand on the scales.

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -585,15 +618,137 @@ impl CellManager {
...
+                let Some(donor) = donor else {
+                    break; // no CPU can be reserved without starving a child cell
+                };
+                let cpu = cell_claimed[&donor]
+                    .iter()
+                    .copied()
+                    .filter(|&cpu| !holdout.test_cpu(cpu) && reservable(cpu))
+                    .min_by(|&a, &b| {
+                        let contested =
+                            |cpu: usize| contention.get(&cpu).map_or(0, |v| v.len()) > 1;
```
tommy-u, line 737: Instead of binary "pick uncontested before contested" you could pick in ascending number of contesting cgrpus. Just an idea.

tommy-u (changes requested):
The only change I need is to keep a flag in the stats that says "I actually enforced holdout during this run." Just a sticky boolean that you set if, at any time, you took a cpu away from a workload cell. I think it will help us with debugging. Otherwise good to squash and merge.

scheds/rust/scx_mitosis/src/cell_manager.rs
likewhatevs: This is subtle enough that i'm going to leave it as is, but IIUC, this would boil down to the same result in every scenario while looking like it was doing something smarter.

tommy-u (approved):
Looks good, great work.

PR #3607, opened 2026-05-28
scx_cosmos: Server tuning

arighi:
scx_cosmos has recently shifted its focus toward more CPU-intensive, server-oriented workloads. This set of changes tunes the scheduler defaults to better match large server systems out of the box. At the same time, the scheduler continues to prioritize latency and responsiveness, so little to no behavioral changes are expected on desktops or smaller systems.

A follow-up change will also update the default settings in scx-loader accordingly:
https://github.com/sched-ext/scx-loader


multics69 (approved):
LGTM. Go, scx_cosmos, go!

PR #3608, opened 2026-05-28
scx_flow: cpufreq performance hint for latency-lane dispatches

galpt:
scx_flow's dispatch path for latency-sensitive tasks (LATENCY_DSQ and URGENT_LATENCY_DSQ) does not set a cpufreq performance hint. The kernel tracks PELT utilization for sched_ext tasks, but the PELT time constant (~32ms half-life) can be slower to respond than an explicit hint on each scheduling event. On a latency-lane dispatch the scheduler already knows the task is time-sensitive, so it can request a frequency ramp immediately rather than waiting for the utilization signal to build up.

Fix by calling scx_bpf_cpuperf_set(SCX_CPUPERF_ONE) at two sites:

  - note_urgent_latency_dispatch(), the existing helper that tracks urgent-latency burst accounting
  - the LATENCY_DSQ consume path in flow_dispatch()

Only these two lanes are targeted. Tasks reach them only after accumulating latency allowance or latency pressure through the scheduler's own classification, so the hint is backed by proven sensitivity. Reserved, contained, and shared dispatches are left to the kernel's default governor.

Signed-off-by: Galih Tama <galpt@v.recipes>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3743,7 +3743,7 @@ dependencies = [
 
 [[package]]
 name = "scx_flow"
-version = "2.2.5"
+version = "2.2.6"
 dependencies = [
  "anyhow",
  "clap",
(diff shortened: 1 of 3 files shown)


hodgesds (approved):
LGTM

PR #3609, opened 2026-05-28
scx_cake: Downgrade sysinfo dependency

arighi:
scx_cake currently rewquires sysinfo >= 0.39, which in turn requires rustc >= 1.95. This breaks builds on Ubuntu 24.04, where the newest rustc version currently available in the distro is 1.91:

 error: rustc 1.91.1 is not supported by the following package:
  sysinfo@0.39.1 requires rustc 1.95

scx_cake does not rely on any specific features introduced in sysinfo 0.39, so downgrade the dependency to 0.38 to restore compatibility with Ubuntu.

### DIFF
--- a/scheds/rust/scx_cake/Cargo.toml
+++ b/scheds/rust/scx_cake/Cargo.toml
@@ -20,7 +20,7 @@ arboard = "3"
 
 scx_utils = { path = "../../../rust/scx_utils", version = "1.1.1" }
 scx_arena = { path = "../../../rust/scx_arena/scx_arena", version = "1.1.1" }
-sysinfo = "0.39"
+sysinfo = "0.38"
 core_affinity = "0.8"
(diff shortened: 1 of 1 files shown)


sirlucjan (approved):

arighi:
Hm... sysinfo 0.39 is used also by scx_layered and scxtop, I'll close this one and send a new PR.

arighi:
The problem is bigger than I thought. It's basically impossible now to support Ubuntu 24.04, because we need a kernel built with pahole >= 1.26 and Ubuntu 24.04 is shipping pahole 1.25, which makes KF_IMPLICIT_ARGS kfuncs incompatible.

Basically all schedulers are now broken in Ubuntu 24.04. I'll check with Canonical if we can upgrade pahole, but it's going to take some time (assuming it's even possible).

PR #3610, opened 2026-05-31
scx_lavd: Per-CPU DSQ queued-load tracking

bboymimi:
# scx_lavd: Per-CPU DSQ queued-load tracking

This series extends scx_lavd's RAVG-weighted `qload_invr` accumulator —
introduced for the per-domain load balancer in 7a4266eab0f5 and consumed
by the cross-cpdom fast-LB picker in e63dc3b64f3e — onto the per-CPU DSQ
path. When `--per-cpu-dsq` is enabled, `pick_most_loaded_dsq` now reads
the same task-size-weighted metric the cross-cpdom LB has used instead
of falling back to raw `scx_bpf_dsq_nr_queued()` counts.

The change is intended as the bedrock for a future inter-CPU load
balancer.

## Patches

1. **scx_lavd: Track per-CPU queued load using RAVG util_est** — Lays
   the bedrock for per-CPU load balancing by mirroring the per-domain
   `qload_invr` accumulator onto `cpu_ctx`. Future load-tracking and
   enqueue-decision heuristics get an accurate, per-CPU load signal.

2. **scx_lavd: Use per-CPU queued load in pick_most_loaded_dsq** —
   Switches the comparison metric from raw DSQ queue lengths
   to the RAVG-weighted `qload_invr`. 

3. **scx_lavd: Fix nr_queued double-count for SMT in sys_stat
   aggregation** — Pre-existing aggregation bug uncovered while
   reviewing the area: `cpu_to_dsq()` collapses SMT siblings to the
   primary (since fbd147521961), so the `sys_stat` loop counted each
   per-CPU DSQ twice on SMT hosts when `--per-cpu-dsq` was set.

4. **scx_lavd: Respect --no-fast-lb in pick_most_loaded_dsq** —
   Mirrors the dual-metric pattern `plan_x_cpdom_migration` has had
   since c2678ad99827: `--no-fast-lb` falls back to the pre-feature
   count metric, giving users who set that knob the same end-to-end
   picker behavior as before this series.

## Testing

Workload: memtier_benchmark against memcached, **Zipfian access**
(`--key-pattern=Z:Z --key-zipf-exp=0.99`) over 1 M keys, OLAP ratio
(`--ratio=1:10`, ~91 % read), 16 threads × 50 conns = 800 client
connections, 128 B values. Per scenario: 1 × 10 s warm-up + **10 × 30 s**
measurement runs, **trimmed mean** (drop min+max, average middle 8).

Three schedulers compared:

- `eevdf` — kernel default
- `lavd_pcd` — `scx_lavd --performance --slice-min-us 3000 --slice-max-us 10000 --pinned-slice-us 3000 --per-cpu-dsq` (current behavior)
- `lavd_pcd_ql` — same flags, **this branch** with queue-load tracking

Six scenarios: unpinned 100 % / 75 % / 50 % / 25 % (rate-limited),
`pin16` (cpu 0-15, 2 CCDs), `pin8` (cpu 0-7, single CCD / single L3).
Loopback only, no NIC. 800 client connections × per-conn rate =
total target RPS.

### pin8 (cpu 0-7, single CCD / single L3)

The previous behavior had a catastrophic worst-case tail when
constrained to one CCD. The new queue-load tracking eliminates it:

| metric | eevdf | lavd_pcd (old) | **lavd_pcd_ql (new)** | new vs old |
|---|---:|---:|---:|---:|
| ops/sec | 1,091,087 | 1,371,969 | **1,403,858** | +2.3 % |
| p50 (ms) | 0.725 | **0.060** | 0.192 | +220 % (still 3.8× better than EEVDF) |
| p90 | **0.964** | 1.013 | 1.661 | +64 % |
| p99 | **1.171** | 7.323 | **2.477** | **−66.2 %** |
| **p99.9** | 2.033 | **40.823** | **4.139** | **−89.9 %** |

### Unpinned (192 CPUs)

ops/sec — the new branch wins back the throughput that old
`--per-cpu-dsq` lost at 192-CPU saturation:

| load | eevdf | lavd_pcd (old) | **lavd_pcd_ql (new)** | new vs old |
|---|---:|---:|---:|---:|
| 100 % | 1,088,371 | 915,424 | **1,148,894** | **+25.5 %** |
| 75 % | 830,292 | 789,724 | 844,659 | +7.0 % |
| 50 % | 599,687 | 595,078 | 597,678 | +0.4 % |
| 25 % | 294,408 | 294,367 | 294,330 | −0.0 % |

p99 latency:

| load | eevdf | lavd_pcd (old) | **lavd_pcd_ql (new)** | new vs old |
|---|---:|---:|---:|---:|
| 100 % | **0.826** | 1.825 | 0.919 | −49.6 % |
| 75 % | **0.842** | 1.627 | 1.052 | −35.3 % |
| 50 % | **0.824** | 1.433 | 1.102 | −23.1 % |
| 25 % | **1.599** | 2.274 | 2.575 | +13.2 % |

p99.9 latency — **most metrics improved, one regression at 75 % load**:

| load | eevdf | lavd_pcd (old) | **lavd_pcd_ql (new)** | new vs old |
|---|---:|---:|---:|---:|
| 100 % | **1.314** | 2.697 | 1.611 | −40.3 % |
| 75 % | **1.449** | 5.315 | **9.439** | **+77.6 %** ⚠️ |
| 50 % | **1.554** | 8.091 | 7.663 | −5.3 % |
| 25 % | **2.467** | 5.391 | 5.443 | +1.0 % |

### Pinned 16 CPUs (2 CCDs on socket 0)

| metric | eevdf | lavd_pcd (old) | **lavd_pcd_ql (new)** | new vs old |
|---|---:|---:|---:|---:|
| ops/sec | 1,308,808 | **1,555,573** | 1,510,863 | −2.9 % |
| p50 | 0.521 | 0.353 | **0.326** | −7.6 % |
| p90 | **1.080** | 1.183 | 1.254 | +6.0 % |
| p99 | **1.840** | 2.026 | 2.076 | +2.5 % |
| p99.9 | 10.375 | **3.185** | 4.261 | +33.8 % |

## Notes

- **Cache-line placement of `cpu_ctx::qload_invr` is still open.**
  The current patch force-aligns the new field to its own cacheline
  to avoid false-sharing. It's not obvious whether there's a
  better slot inside the existing layout that would avoid introducing
  a new cacheline without disturbing the established access
  patterns Changwoo carefully laid out — would appreciate Changwoo's
  take on whether this placement is acceptable as-is or whether the
  field should be tucked into existing slack elsewhere. Verified the
  current layout via `pahole`.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
+++ b/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
@@ -228,12 +228,15 @@ struct task_ctx {
 	u32	cpu_id;			/* where a task is running now */
 	u32	prev_cpu_id;		/* where a task ran last time */
 	u8	queued_in_cpdom_id;	/* cpdom this task's load is counted in; LAVD_CPDOM_MAX_NR = not queued */
-	u32	queued_load_snapshot;	/* task_load_metric() value snapshotted at enqueue time */
+	s16	queued_on_cpu_id;	/* primary CPU id this task's load is counted on; -1 = not queued */
+	u32	queued_load_snapshot;	/* task_load_metric() value snapshotted at enqueue time for the per-cpdom counter */
+	u32	queued_load_snapshot_cpu; /* task_load_metric() value snapshotted at enqueue time for the per-CPU counter */
 	pid_t	pid;			/* pid for this task */
 	pid_t	waker_pid;		/* last waker's PID */
 
-	/* --- cacheline 4 boundary (256 bytes) --- */
-	u32	util_est;		/* Estimated task util using ravg duty cycle */
+	/* --- cacheline 5 boundary (320 bytes): ravg/util read-mostly group --- */
+	u32	util_est __attribute__((aligned(CACHELINE_SIZE)));
+					/* Estimated task util using ravg duty cycle */
 	struct ravg_data avg_util_ravg;	/* Running average of task utilization using ravg */
 	char	waker_comm[TASK_COMM_LEN + 1]; /* last waker's comm */
 } __attribute__((aligned(CACHELINE_SIZE)));
@@ -529,6 +532,14 @@ struct cpu_ctx {
 	struct ravg_data avg_irq_steal_ravg;	/* Running average of IRQ steal utilization using ravg */
 	struct ravg_data avg_util_ravg;	/* Running average of CPU utilization using ravg */
 	volatile u32	util_est;	/* Estimated CPU utilization from ravg tracking */
+
+	/* --- cacheline 7 boundary (448 bytes): cross-CPU write-heavy accumulator --- */
+	/*
+	 * qload_invr is incremented from any CPU on enqueue (remote write
+	 * via bpf_map_lookup_percpu_elem + __sync_fetch_and_add) and
+	 * decremented likewise on dispatch/dequeue/exit.
+	 */
+	u64	qload_invr __attribute__((aligned(CACHELINE_SIZE)));
 } __attribute__((aligned(CACHELINE_SIZE)));
 
 extern const volatile u64	nr_llcs;	/* number of LLC domains */
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -706,6 +706,41 @@ static __always_inline void unaccount_queued_load(task_ctx *taskc)
 	WRITE_ONCE(taskc->queued_in_cpdom_id, LAVD_CPDOM_MAX_NR);
 }
 
+static __always_inline void account_queued_load_pcpu(task_ctx *taskc,
+						     s32 primary_cpu)
+{
+	struct cpu_ctx *cpuc;
+	u32 load;
+
+	if (primary_cpu < 0 || primary_cpu >= LAVD_CPU_ID_MAX)
+		return;
+
+	if (READ_ONCE(taskc->queued_on_cpu_id) >= 0)
+		return;
+
+	load = task_load_metric(taskc);
+	cpuc = get_cpu_ctx_id(primary_cpu);
+	if (cpuc)
+		__sync_fetch_and_add(&cpuc->qload_invr, load);
+	taskc->queued_load_snapshot_cpu = load;
+	WRITE_ONCE(taskc->queued_on_cpu_id, (s16)primary_cpu);
+}
+
+static __always_inline void unaccount_queued_load_pcpu(task_ctx *taskc)
+{
+	struct cpu_ctx *cpuc;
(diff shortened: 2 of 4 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -666,6 +666,40 @@ static __always_inline void unaccount_queued_load(task_ctx *taskc)
 	WRITE_ONCE(taskc->queued_in_cpdom_id, LAVD_CPDOM_MAX_NR);
 }
 
+static __always_inline void account_queued_load_pcpu(task_ctx *taskc, s32 cpu)
```
multics69: It is worth adding comments saying that `cpu` here is the primary CPU when SMT is enabled. Or rename `cpu` to `primary_cpu` for fool-proofing.

multics69 (commented):
Overall, LGTM. I left one minor comment.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
bboymimi: Thanks for the suggestion. It's a good suggestion to keep a good awareness of the primary cpu being used for the queue load accumulation when SMT is enabled.

multics69 (approved):

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -228,12 +228,22 @@ struct task_ctx {
...
 
-	/* --- cacheline 4 boundary (256 bytes) --- */
-	u32	util_est;		/* Estimated task util using ravg duty cycle */
+	/* --- cacheline 5 boundary (320 bytes): ravg/util read-mostly group --- */
+	/*
+	 * util_est is force-aligned to a cacheline so the ravg state below
+	 * (read on every ravg_accumulate / consulted on every dispatch)
+	 * sits in its own line, isolated from the write-heavy accounting
+	 * fields above. Without the alignment, avg_util_ravg straddles
+	 * CL4/CL5 and ravg_accumulate pulls two L1 lines per call.
```
daidavid: nit: IMO, comments like this don't add much value and just bloats readability.

daidavid (approved):
Thanks Gavin, appreciate the extra testing and data!

Pending outcome of the jemalloc investigation we may opt to undo the per-core dsq designation and go to per logical cpu dsqs. Userspace memory allocators may make assumptions at the logical cpu level.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
bboymimi: Thanks! I've removed the comment.

PR #3611, opened 2026-06-01
scx_cosmos: Skip CPU-less NUMA nodes when creating per-node DSQ

nvax-r:
## Summary
`cosmost_init()` iterates over every NUMA node and creates a per-node DSQ plus a node context. On systems with CPU-less NUMA nodes, e.g. GPU-memory node or CXL memory node, this will create DSQs with node contexts for nodes that have no CPU at all.

Implement `node_has_cpus()` helper and skip CPU-less nodes in the init loop.

On a 40-node system with 38 CPU-less nodes this drops per-node DSQ creation from 40 to 2, and shortens the init loop by ~27%.

## Result Before/After
  | Metric                     | Baseline           | Patched            | Δ                        |
  |----------------------------|--------------------|--------------------|--------------------------|
  | **Per-node DSQs created**  | **40** (all runs)  | **2** (all runs)   | −38 dead DSQs            |
  | Init loop, median          | 240.5 µs           | 174.4 µs           | **−66 µs (~27% faster)** |
  | Init loop, min             | 211.6 µs           | 162.4 µs           | −49 µs (~23%)            |

## Note
The same "dead DSQ for CPU-less NUMA node" scenario pattern likely exists in other scx schedulers as well, if it is the case, I'll be happy for assist on fixing those cases, and I think "node_has_cpus()" helper will be put in utils file , etc rather than inside scx_cosmos only.

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -1485,14 +1485,29 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(cosmos_init)
 		int node;
 
 		bpf_for(node, 0, nr_node_ids) {
-			err = scx_bpf_create_dsq(node, node);
-			if (err) {
-				scx_bpf_error("failed to create node DSQ %d: %d", node, err);
-				return err;
-			}
+			struct node_ctx *nctx;
+
 			err = init_node(node);
 			if (err) {
 				scx_bpf_error("failed to initialize NUMA node %d: %d", node, err);
+				return err;
+			}
+
+			/*
+			 * Skip per-node DSQ creation for CPU-less NUMA nodes
+			 * (e.g. GPU-memory or CXL-memory nodes): with no CPU
+			 * to consume from it, the DSQ would never be used.
+			 * init_node() already populated nctx->cpumask, so an
+			 * empty mask means the node has no CPU.
+			 */
+			nctx = try_lookup_node_ctx(node);
+			if (!nctx || !nctx->cpumask ||
+			    bpf_cpumask_empty(cast_mask(nctx->cpumask)))
+				continue;
+
+			err = scx_bpf_create_dsq(node, node);
+			if (err) {
+				scx_bpf_error("failed to create node DSQ %d: %d", node, err);
 				return err;
 			}
 		}


nvax-r:
Testing script if you want to verify on your machine , note this will take no effect if your system has no CPU-less NUMA node
```shell
  set -euo pipefail

  FILE="scheds/rust/scx_cosmos/src/bpf/main.bpf.c"
  FIX_REF="${FIX_REF:-HEAD}"        # commit containing the fix
  BASE_REF="${BASE_REF:-HEAD~1}"    # commit immediately before the fix
  ITERS="${ITERS:-7}"               # scheduler loads per build
  SETTLE="${SETTLE:-4}"             # seconds to wait for attach before reading trace
  export PATH="$HOME/.cargo/bin:$PATH"

  TR=/sys/kernel/tracing
  [ -d "$TR" ] || TR=/sys/kernel/debug/tracing

  command -v cargo >/dev/null || { echo "cargo not in PATH"; exit 1; }
  git diff --quiet -- "$FILE" || { echo "ERROR: uncommitted changes to $FILE; commit or stash first"; exit 1; }

  # --- topology (informational): which nodes have CPUs --------------------------
  echo "== NUMA topology (empty cpulist == CPU-less node) =="
  for d in /sys/devices/system/node/node*; do
      echo "$(basename "$d"): [$(cat "$d/cpulist" 2>/dev/null)]"
  done
  echo

  # --- temporary measurement instrumentation -----------------------------------
  # Applied to whichever version of $FILE is currently checked out. Anchors are
  # common to both the baseline and the fixed loop and do not touch the fix's guard.
  instrument() {
  python3 - "$1" <<'PY'
  import sys
  p = sys.argv[1]
  s = open(p).read()

  decl = "\t\tint node;\n\n\t\tbpf_for(node, 0, nr_node_ids) {"
  assert s.count(decl) == 1, ("decl anchor", s.count(decl))
  s = s.replace(decl,
      "\t\tint node;\n\t\tu64 __m_t0 = bpf_ktime_get_ns();\n\t\tu32 __m_ndsq = 0;\n\n\t\tbpf_for(node, 0,
  nr_node_ids) {", 1)

  tail = ('\t\t\terr = init_node(node);\n\t\t\tif (err) {\n'
          '\t\t\t\tscx_bpf_error("failed to initialize NUMA node %d: %d", node, err);\n'
          '\t\t\t\treturn err;\n\t\t\t}\n\t\t}\n\t} else {')
  assert s.count(tail) == 1, ("loop-tail anchor", s.count(tail))
  s = s.replace(tail,
      '\t\t\terr = init_node(node);\n\t\t\tif (err) {\n'
      '\t\t\t\tscx_bpf_error("failed to initialize NUMA node %d: %d", node, err);\n'
      '\t\t\t\treturn err;\n\t\t\t}\n\t\t\t__m_ndsq++;\n\t\t}\n'
      '\t\tbpf_printk("COSMOS_MEASURE dsqs=%u loop_ns=%llu", __m_ndsq, bpf_ktime_get_ns() - __m_t0);\n\t} else
  {', 1)

  open(p, "w").write(s)
  PY
  }

  build_variant() {  # $1 = git ref, $2 = output binary path
      git checkout -q "$1" -- "$FILE"
      instrument "$FILE"
      cargo build --release -p scx_cosmos >/tmp/cosmos_build.log 2>&1 \
          || { echo "build failed for $1"; tail -20 /tmp/cosmos_build.log; git checkout -q -- "$FILE"; exit 1; }
      cp target/release/scx_cosmos "$2"
      git checkout -q -- "$FILE"      # discard instrumentation, restore tree
  }

  echo "Building baseline ($BASE_REF) and fix ($FIX_REF) with instrumentation..."
  build_variant "$BASE_REF" /tmp/scx_baseline
  build_variant "$FIX_REF"  /tmp/scx_fix
```

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -1470,6 +1469,28 @@ static int init_node(int node)
...
+ * CPU-less nodes (e.g. memory-only or GPU memory nodes) have no CPU to
+ * consume from a per-node DSQ, so there is no point creating a DSQ or a
+ * node context for them.
+ */
+static bool node_has_cpus(int node)
+{
+	u32 cpu;
+	bool found = false;
+
+	bpf_for(cpu, 0, nr_cpu_ids) {
```
arighi: Instead of adding another O(N) loop on the CPUs, can we invert the order of operation in `cosmos_init()` and call `init_node()` before creating the per-node DSQ?

Then we can check `bpf_cpumask_empty(cast_mask(nctx->cpumask)))` to detect if a node doesn't have any CPU assigned and avoid creating the DSQ for that node. We may still want to initialize `node_ctx` also for empty nodes, so that part should be fine.

arighi (changes requested):
Thanks for looking at this, I left a small comment. Let me know what you think.

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
nvax-r: Agree, thanks for the feedback, I've made the change.
Though I wonder why do we want to keep `node_ctx` for CPU-less node ? `node_ctx` only has cpumask for now, guessing you're keeping it for future usage such as memory-aware policies ?

arighi: Right, we don't need empty nodes right now, but it might be useful to keep them for potential future use. They can carry locality information for device-attached memory (e.g., GPU HBM), which could eventually be used to influence task placement. For example, a scheduler could track the memory node associated with a task and use that information when making placement decisions.

arighi (approved):

PR #3615, opened 2026-06-01
lib/cgroup_bw: fix orphan task from cbw_put_aside() vs LLC-free race

multics69:
cbw_put_aside() parks a runnable task in a per-LLC BTQ. If the cgroup's LLC contexts are freed in the middle of that operation, the task lands in a detached BTQ that nobody drains, and stalls until the SCX watchdog fires.

There should be no race: cbw_put_aside() runs against a live cgroup, while the LLC-free path runs only when the cgroup is exiting. But the replenish timer was tearing down LLC contexts on a live cgroup because bpf_cgroup_from_id() returned NULL and that NULL was treated as "this cgroup is exiting." On kernels older than v6.18, bpf_cgroup_from_id() filters against \`current\`'s cgroup namespace; inside the BPF timer \`current\` is whichever task happened to be on the CPU -- often a containerised service -- so the lookup fails on cgroups outside that namespace even though the cgroups are perfectly alive. The root cause is that bpf_cgroup_from_id() is not a reliable signal on <v6.18 kernels.

Replace every bpf_cgroup_from_id() call site with a direct cgroup-id lookup against cbw_cgrp_map, following the same idea cbw_get_root_cgrp() already uses for the root. To support cbw_update_nquota_ub(), which previously needed a cgroup pointer only for ancestor walks, cache the parent cgroup id and tree level in scx_cgroup_ctx at init. The only remaining bpf_cgroup_from_id() is in the non-root branch of scx_cgroup_bw_dump(), a cosmetic debug path.

With the spurious offline trigger gone, cbw_cgroup_bw_offline()'s early-drain rescue no longer fires on live cgroups, and on a genuine offline the same drain runs inside scx_cgroup_bw_exit() one BPF instruction after the map entry is deleted -- so the helper is redundant. Remove it along with the offline_cgroup label and the nr_moved/root_added bookkeeping it relied on. As defence in depth, cbw_put_aside() re-reads llcx->btq under the atq lock and warns if the BTQ pointer changed under it; with this patch that warn should never fire.

### DIFF
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -99,7 +99,13 @@ struct scx_cgroup_ctx {
 
 		/* cgroup id */
 		u64		id;
-	
+
+		/* parent cgroup id (0 for root); set once at init */
+		u64		parent_id;
+
+		/* cgroup tree depth (root = 0); set once at init */
+		u32		level;
+
 		/*
 		 * Given @quota, @period, and @burst in nanoseconds.
 		 */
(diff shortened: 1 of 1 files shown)


daidavid (approved):

multics69:
Thanks @daidavid for the review.

PR #3616, opened 2026-06-02
scx_cosmos: Change default time slice to 1ms

arighi:
Commit e589435e86aa ("scx_cosmos: Tune default time slice for server workloads") increased the default time slice to 20ms to better target CPU-intensive workloads.

However, a 20ms time slice appears overly aggressive and can lead to performance regressions in multiple scenarios, including for CPU-bound workloads.

Scale down the default time slice to 1ms, which provides a more balanced trade-off and serves as a more reasonable default for general-purpose use.


hodgesds (approved):

ColinIanKing:
is is possible to scale the time slice according to the load type?

PR #3618, opened 2026-06-02
scx_lavd, lib/cgroup_bw: cap task-stall latency under heavy cpu.max throttling

multics69:
When a cgroup's cpu.max quota is tight relative to the work its tasks
want to do, the cgroup spends most of its time throttled.  Tasks then
wait a long time for the small amount of CPU the quota allows, and
that wait is not bounded: it can grow until a task trips the
30-second SCX runnable-task-stall watchdog and brings the scheduler
down.  Worse, the wait is not even fairly distributed -- some tasks
can be starved indefinitely while others make progress.

A single throttled cgroup should never be able to stall the whole
scheduler.  This series makes that guarantee by attacking the two
ways an unbounded wait arises.

The first is slice length.  Under a tight quota, the scheduler can
only admit a few tasks per period, and if each runs a full slice,
the rest wait proportionally longer -- the more contended the cgroup,
the worse it gets.  The scheduler needs to know how contended a
cgroup is so it can react.  [2/3] gives it that: scx_cgroup_bw_pressure()
reports a cgroup's bandwidth pressure, and scx_lavd shortens slices as
pressure rises, so more tasks get a turn within the watchdog window
instead of a few monopolizing each period.

The second is queue ordering.  A throttled task's place in line is
its scheduler vtime, which is fine for fairness but offers no
liveness guarantee: a compute-heavy task accumulates large vtime and
can sit behind an endless stream of short, small-vtime tasks, never
reaching the front.  Fairness alone does not bound how long it waits.
[3/3] adds that bound -- a task is guaranteed to reach the head of
the queue within a few seconds regardless of vtime -- while keeping
the scheduler's vtime ordering intact for tasks queued close
together.

[1/3] is preparatory refactoring with no functional change: it
factors the per-task context-cache accessors into helpers so [2/3]
can reuse them.

- [1/3] lib/cgroup_bw: factor taskc-cached cgx/llcx accessors into helpers
- [2/3] lib/cgroup_bw, scx_lavd: add scx_cgroup_bw_pressure() API
- [3/3] lib/cgroup_bw: blend wall-clock time into BTQ vtime to bound throttle delay

### DIFF
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -80,7 +80,17 @@ enum scx_cgroup_consts {
 	CBW_PRESSURE_BUDGET_KNEE	= 256,
 	CBW_PRESSURE_BACKLOG_STEP	= 128,
 	CBW_PRESSURE_BACKLOG_CAP	= 128,
+	/*
+	 * Shift used to split the 64-bit BTQ key evenly into a wall-clock
+	 * epoch (upper 32 bits) and a scheduler vtime (lower 32 bits).
+	 * Tasks queued in the same ~4-second epoch (2^32 ns ~= 4.29 s)
+	 * compete by their vtime; tasks from an earlier epoch are always
+	 * dispatched first, bounding the maximum BTQ wait to ~4 seconds.
+	 */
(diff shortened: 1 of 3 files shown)


multics69:
Rebased to the main branch.

PR #3619, opened 2026-06-02
[scx_layered] Add opt-in peak-hold shrink resistance

sravi-in:
Add `util_peak_half_life_ms` as an opt-in per-layer config knob for `Confined` and `Grouped` layers. When set, `scx_layered` keeps a decaying peak of the sizing utilization and uses `max(util, peak_util)` only for the upper target bound, so recent bursts can delay shrink while growth remains based on current utilization. The default `0` preserves existing behavior, and the knob is rejected on layers without `util_range`.


kkdwvd:
Looks great, will give it a spin tonight and then land.

kkdwvd (approved):
Could you also run cargo fmt and squash them into the commit?

kkdwvd:
Closing in favor of #3632.

PR #3621, opened 2026-06-03
scx_cosmos: Introduce sleeper virtual lag

arighi:
The current deadline is determined in function of the exec_runtime (CPU time since last wakeup). Since exec_runtime is reset whenever a task sleeps, regardless of its fairness position, it fails to distinguish between a task that went to sleep with its vruntime far ahead of vtime_now (i.e., a CPU-intensive task that briefly blocked) and a task that slept while lagging behind. In both cases, the task wakes up with exec_runtime == 0 and competes based solely on its vruntime, allowing a CPU-bound task to effectively discard its accumulated penalty by briefly sleeping.

Improve this, by replacing per-task exec_runtime with sleep_vlag: a snapshot of the task's vruntime distance from the current global vruntime (vtime_now) at the moment it voluntarily sleeps.

A positive sleep_vlag indicates that the task slept behind the virtual clock (credit, typical sleeper), while a negative sleep_vlag indicates that it slept ahead of the virtual clock (debt, having consumed CPU aggressively before sleeping).

On wakeup, debt is re-applied before DSQ insertion by bumping task's vruntime past vtime_now (re-aligning debt to the current global vruntime); credit is preserved in the task's vruntime and bounded to a limit, so long sleepers cannot claim an unbounded head start. Both paths go through task_dl(); tasks dispatched directly to local DSQs bypass it (no dsq_vtime update), so ops.running() applies any outstanding debt as a fallback.

The credit/debt bound is scaled with wakeup frequency (frequent sleepers earn a wider window) and is expressed in vtime units, so the real-time budget is normalized across task weights.

While at it, also switch to the new APIs for updating the slice and DSQ virtual time: scx_bpf_task_set_dsq_vtime() and scx_bpf_dsq_set_slice().

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -35,6 +35,15 @@ char _license[] SEC("license") = "GPL";
 #define SHARED_DSQ		0
 
 /*
+ * Maximum rate of task wakeups/sec (tasks with a higher rate are capped to this
+ * value).
+ *
+ * Note that the wakeup rate is evaluated over a period of 100ms, so this number
+ * must be multiplied by 10 to determine the actual limit in wakeups/sec.
+ */
+#define MAX_WAKEUP_FREQ                1000ULL
+
+/*
  * Thresholds for applying hysteresis to CPU performance scaling:
  *  - CPUFREQ_LOW_THRESH: below this level, reduce performance to minimum
  *  - CPUFREQ_HIGH_THRESH: above this level, raise performance to maximum
@@ -200,7 +209,7 @@ static u64 vtime_now;
 struct task_ctx {
 	struct bpf_cpumask __kptr *cpumask;
 	u64 last_run_at;
-	u64 exec_runtime;
+	s64 sleep_vlag;
 	u64 wakeup_freq;
 	u64 last_woke_at;
 	u64 perf_events;
@@ -393,10 +402,14 @@ static u64 calc_avg(u64 old_val, u64 new_val)
  */
 static u64 update_freq(u64 freq, u64 interval)
 {
-        u64 new_freq;
+	u64 new_freq;
 
-        new_freq = (100 * NSEC_PER_MSEC) / interval;
-        return calc_avg(freq, new_freq);
+	if (!interval)
+		interval = 1;
+	new_freq = (100 * NSEC_PER_MSEC) / interval;
+	new_freq = MIN(new_freq, MAX_WAKEUP_FREQ);
+
+	return calc_avg(freq, new_freq);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


hodgesds:
>  tasks dispatched directly to local DSQs bypass it (no dsq_vtime update), so ops.running() applies any outstanding debt as a fallback.

Does this need to clamp vtime as well if a task sleeps for a long time and then is direct dispatched? Maybe something like this in `task_apply_sleep_debt` since it is being used to clamp the sleep time, should there be a clamp on the vtime for it as well?
```
u64 vtime_min = vtime_now - vsleep_max;

if (!vime_before(p->scx.dsq_vtime, vtime_min)) 
```

hodgesds (approved):
Seems like an improvement and closer to EEVDF handling of lag.

arighi:
> > tasks dispatched directly to local DSQs bypass it (no dsq_vtime update), so ops.running() applies any outstanding debt as a fallback.
> 
> Does this need to clamp vtime as well if a task sleeps for a long time and then is direct dispatched? Maybe something like this in `task_apply_sleep_debt` since it is being used to clamp the sleep time, should there be a clamp on the vtime for it as well?
> 
> ```
> u64 vtime_min = vtime_now - vsleep_max;
> 
> if (!vime_before(p->scx.dsq_vtime, vtime_min)) 
> ```

The condition in your snippet should be inverted, right? (clamp if `p->scx.dsq_vtime < vtime_min`)

I think we don't need this this. It's true that after a long sleep `dsq_vtime` can be arbitrarily far behind, but on direct dispatch this doesn't matter because a task goes to a local FIFO DSQ (`dsq_vtime` is ignored) and as soon as the task is inserted to a shared DSQ then `task_dl()` floors the deadline at `vtime_min`.

The only side effect, if we have a bunch of tasks with a `dsq_vtime` far behind, is that `vtime_now` can be essentially "frozen" until tasks catch up with it (because it advances only when `p->scx.dsq_vtime > vtime_now`). But this is also not a problem: as soon as these tasks join a shared DSQ, `task_dl()` bumps their `dsq_vtime` up to `vtime_min`, so they quickly catch up and the global clock resumes advancing.

We only need to charge the debt in `ops.running()` via `task_apply_sleep_debt()`, because we snapshot the debt in `ops.quiescent()`, then if a task sleeps briefly and it's directly dispatched on the next wakeup, we re-apply its debt as the task starts running. That way `dsq_vtime` already carries the debt if the task lands on a shared DSQ and the debt is not lost just because the task did a short sleep (in the previous `exec_runtime`-based implementation this debt was always reset to 0 on sleep and lost).

arighi:
> Seems like an improvement and closer to EEVDF handling of lag.

Exactly, this is similar to the `PLACE_LAG` logic in EEVDF, with a small difference: the debt part is essentially EEVDF-style lag preservation, but the credit is not restored to the snapshot like EEVDF, it's more similar to the classic CFS "sleeper bonus" logic, bounded by a limit.

In practice, this change makes scx_cosmos a hybrid between EEVDF and CFS: it preserves lag in an EEVDF-like manner on the debt side, while applying bounded CFS-style sleeper credit on the credit side.

This should improve responsiveness compared to EEVDF. Consider a CPU-intensive task that briefly sleeps (i.e., a game's render/sim thread that runs hard then briefly block on vsync, GPU fence, frame queue, etc.). Under EEVDF, if the task goes to sleep with debt (or near-neutral lag), that same lag is effectively preserved and restored when the task wakes up.

With scx_cosmos, when a task sleeps with debt, other tasks continue to run and `vtime_now` advances, gradually reducing the task's accumulated debt while it remains asleep. By the time it wakes up, the gap between `vtime_now` and its frozen `dsq_vtime` may have narrowed significantly or even become positive, effectively turning debt into credit if the system was busy enough during the sleep interval.

Although a short sleep only reduces debt incrementally on each cycle, the effect accumulates over time (up to the `vtime_min` cap). For workloads that repeatedly alternate between bursts of execution and brief sleeps (e.g., a game) this approach should be less punitive than EEVDF, which snapshots the full debt at sleep time and restores it unchanged on wakeup.

And on the debt side, re-applying the debt to `dsq_vtime` on wakeup prevents abusing such sleep prioritization (see `stress-ng --sleep N` tasks monopolizing CPUs with the previous implementation, because their debt - time while they use CPU - was completely reset at every sleep, which is wrong because these stressors are actually using a lot of CPU).

And on the debt part, re-applying the debt to `dsq_vtime` on wakeup prevents abusing such sleep prioritization. In the previous `exec_runtime`-based implementation, any task could clear its accumulated CPU penalty by briefly sleeping, since `exec_runtime` was unconditionally reset to 0 on every wakeup (from `ops.runnable()`). This allowed CPU-intensive tasks which alternate brief sleeps (see `stress-ng --sleep N`) to repeatedly erase their CPU usage history and monopolize CPUs, starving everyone else.

arighi:
Force pushed, updating the description of `--slice-lag-us` that now has a slightly different meaning with this change.

hodgesds:
>  But this is also not a problem: as soon as these tasks join a shared DSQ, task_dl() bumps their dsq_vtime up to vtime_min, so they quickly catch up and the global clock resumes advancing.

Right, it only prioritizes once the task runs and the vtime catches up.

> In the previous exec_runtime-based implementation, any task could clear its accumulated CPU penalty by briefly sleeping, since exec_runtime was unconditionally reset to 0 on every wakeup (from ops.runnable()). 

I see that now, thanks for the explanation!

PR #3622, opened 2026-06-03
scx_pandemoniumv5.12.0

wllclngn:
Migration-potential Phi -- a cache move priced by its graph resistance --
with the sojourn potential warp and L3/CCX/CCD topology in the BPF;
Phi-coherent eviction, a Phi-aware scatter pathway, and a low-core slice
discipline in ADAPTIVE.

The warp is now a bounded per-task POTENTIAL, not a flat per-tier constant
-- the decisive interactivity of the vtime era returns EARNED rather than
assigned, with no virtual-time clock resurrected and no unbounding into
starvation; an L3/CCX/CCD topology tier stops communicating pairs from
crossing the Infinity-Fabric boundary at the placement and steal sites;
the migration potential Phi -- R_eff, the graph resistance of a move,
weighed against the queueing relief it buys -- prices every cross-domain
steal, so it crosses a cache boundary only when the backlog justifies the
cost; and the adaptive control plane is driven into
coherence with the sharpened BPF -- Phi-priced eviction keyed to the same
equilibrium the kernel rescues at, a cross-CCX scatter pathway that lets
MWU see the migration storm it can induce, and a core-count slice
discipline that ends the low-core bimodal tail -- so the Rust layer
reinforces the kernel's placement instead of contending with it. Sojourn
stays the engine; the ordering key is unchanged (now - warp, oldest-first).
On the real-world cachyos suite this build takes the lowest total time of
any scheduler measured, EEVDF and scx_cake included.

SOJOURN POTENTIAL WARP (task_deadline)
- warp was a flat per-tier constant: lag_cap_ns for LAT_CRITICAL,
  lag_cap_ns/2 for INTERACTIVE, 0 for BATCH, maturity-gated. Under
  heterogeneous load a swarm of background threads that had merely aged
  past the constant out-sorted a genuine interactive task -- the flat
  warp could not be made decisive without unbounding it, and unbounding
  reintroduces starvation. The decisive interactivity of the vtime era
  was lost in the trade for bounded fairness.
- warp is now a bounded per-task potential:
    warp = (lag_cap_ns * task_potentiality_q16(tctx, knobs)) >> 16
  lag_cap_ns remains the ceiling (the starvation bound); the Q16
  potential in [0, 65536] continuously positions the back-date from the
  task's own behavior. The sort key is unchanged -- now - warp,
  oldest-first sojourn. No dispatch-waterfall change, no new step.
- task_potentiality_q16: BATCH -> 0 (never warps); LAT_CRITICAL -> 65536
  (RT floor absolute, full ceiling, untouched); INTERACTIVE -> earned
  continuously as a capped sum of two monotonic Q16 axes:
    * SERVICE DEFICIT -- a short runtime per activation (a task that
      yields before its quantum) earns warp; a full-slice hog earns
      none. (slice_ns - avg_runtime) / slice_ns, capped at 32768.
    * SHAPE -- low runtime variance relative to mean (a periodic,
      frame-paced workload) earns warp; chaotic burstiness earns none.
      Inverse coefficient-of-variation from runtime_dev, capped at 32768.
- A genuine interactive task (sleeps, periodic, short bursts) earns a
  large warp and wins decisively; a CPU-hog -- or a chaotic, high-
  frequency I/O kthread that the classifier happens to tag INTERACTIVE
  -- earns ~0. The bound (<= lag_cap_ns) keeps the engine starvation-
  free; the maturity gate still returns 0 for an unmatured fork, so a
  fork storm cannot leapfrog established work. This rebuilds the vtime-
  era "under-served wins, then self-corrects" property entirely inside
  the sojourn key: no second fairness engine, no waterfall surgery.

L3 / CCX / CCD TOPOLOGY TIER (src/topology.rs)
- build_laplacian gains a conductance rung between L2 and socket:
  same-L3/CCX = 3.0, cross-CCX-same-socket = 1.0 (the prior socket
  weight, unchanged), cross-socket = 0.3. The L3 domain is parsed from
  cache/index3 and the rung is gated on ccx_active -- true only when a
  socket holds more than one L3 group (an AMD chiplet part).
- Raising the same-CCX rung (rather than lowering the cross-CCX cut)
  ranks same-CCX peers ahead of cross-CCX ones in affinity_rank WITHOUT
  moving lambda_2: the cross-CCX cut stays at 1.0, so tau and codel_eq
  are unchanged. Verified on the 12C dual-CCX reference -- lambda_2 = 12,
  tau = 13.3ms, codel_eq identical before and after.
- Monolithic-L3 parts (ccx_active false) collapse to socket granularity:
  an exact no-op. Fixes the cross-CCD migration of communicating pairs
  (the dual-CCD stutter class, Issue https://github.com/wllclngn/PANDEMONIUM/issues/10) at the placement and steal
  level; the fix flows through the existing affinity_rank map, so the
  BPF dispatch path is unchanged and no new DSQ is added.

QUEUE CLEANUP (src/bpf/main.bpf.c, src/bpf/intf.h)
- Replacing virtual time with the sojourn key left vtime-era vestiges in
  the dispatch queue; this release clears them without touching the
  waterfall or adding a step. task_deadline sheds two dead parameters
  (p, dsq_id); the unused K_VTIME_CEILING constant is deleted; the stale
  comments are corrected to the sojourn engine (DSQ_VTIME / AWAKE_VTIME
  headers -> SOJOURN, vtime_floor / lag_scale / awake_cap -> warp
  ceiling, "STEP 6" -> "STEP 3").
- The per-task awake_vtime (u64) narrows to a ran_since_wake flag (u8
  folded into existing padding) -- eight bytes per task reclaimed, the
  just-woke-up signal that drives kick selection preserved.
- Six open-coded overflow drain-and-clear sequences collapse into one
  overflow_drain_clear helper. nr_reenqueue is now incremented in
  cpu_release from the scx_bpf_reenqueue_local return -- a statistic that
  was exposed but never counted.

IPC WAKE_SYNC KICK RESTORE (pandemonium_select_cpu)
- Both WAKE_SYNC fast-path arms had lost their scx_bpf_kick_cpu calls in
  an in-flight 12C refactor -- the arms inserted the wakee into a per-CPU
  DSQ and returned without kicking, so the target sat until its next tick
  (~1ms at HZ=1000), silently reintroducing the IPC round-trip tail closed
  at commit 5224a8d5e. Restored both: idle-target -> SCX_KICK_IDLE,
  no-idle/saturation -> SCX_KICK_PREEMPT (under SYNC the waker blocks
  imminently). select_cpu's return value alone only auto-wakes an idle CPU
  for SCX_DSQ_LOCAL, not a custom per-CPU DSQ, so the explicit kick is
  required. Both sites now carry a load-bearing marker against a fourth
  accidental removal. Bench gate: IPC round-trip median should return to
  ~29us -- the missing-kick failure mode is silent, so build-clean is
  necessary but not sufficient.

IPC / FORK-THREAD WARM-ANCHOR PLACEMENT (pandemonium_enqueue, dispatch STEP 1)
- Companion to the kick restore above: the kick only helps if the wakee lands
  on its warm core. Through v5.11.0 enqueue inserted woken tasks into the
  shared node_dsq, drained by whichever core dispatched next (STEP 3), so the
  wakee almost never returned to its prev/last_cpu -- last_cpu was read only to
  pick which CPU to kick, never to choose the landing queue. That scatter is
  the IPC round-trip tail and the cache-miss source on wakeup-heavy work.
- Enqueue now anchors the wakee to its own warm cache. TIER 1 (idle wakee):
  find_idle_l2_sibling biases toward the wakee's last_cpu L2 group, then seats
  it on that cpu's own per-CPU DSQ ((u64)cpu), not node_dsq -- the found-idle
  core is shallow, so no spill search is needed. TIER 2 (wakeup preemption /
  LAT_CRITICAL re-enqueue): pick_pcpu_dsq_with_spill seats toward last_cpu --
  the warm per-CPU DSQ if it has room, else the nearest R_eff sibling, else
  node_dsq as the last-resort escape valve. Placement is symmetric with
  select_cpu.
- Dispatch STEP 1 pulls by cache distance, not queue depth. The single
  affinity_rank loop (slot 0 = L2 sibling, slots 1+ = R_eff-ranked cross-L2
  peers) steals from the first peer with nr_queued > 1 -- the closest surplus,
  not the deepest queue the v5.11.0 scan dragged work in from. The > 1 guard
  leaves a peer's lone (warm) task unyanked under light imbalance, so a wakee
  that landed on a sibling rather than its exact last_cpu is still recovered
  near its warm cache.
- Net: communicating pairs stop migrating across cores on every wakeup, so
  cpu-migrations collapse and the IPC round-trip rides the already-tuned
  sojourn key cleanly. The fork-storm throughput gap is the remaining open
  axis -- the warm anchor holds a pair, but a dense storm of short-lived
  communicating threads scatters faster than placement re-anchors. Closing it
  needs self-adaptive in-kernel detection (a pipe pair and a fork storm want
  opposite placement, so no single static threshold separates them) and a
  throughput lift that does not spend the scheduler's latency -- delivered
  below by FLOW SIGNATURE shape routing and the MIGRATION POTENTIAL Phi.

MIGRATION POTENTIAL PHI (src/bpf/main.bpf.c, src/topology.rs, src/bpf/intf.h)
- The fork-storm cache axis the warm anchor left open. R_eff was a placement
  ranking only -- it ordered candidate CPUs by distance but never priced a
  migration against the queueing relief it bought, so a cross-CCX steal was as
  cheap as an SMT-sibling one once a head aged (the measured ~47% cross-CCX
  migration scatter). The migration potential Phi = R_eff - beta*sojourn is now
  operational on the dispatch STEP 1 steal: a peer's head must wait past
  codel_target PLUS a distance penalty before this CPU pulls it. An SMT sibling
  (R_eff ~ 0) stays freely relievable at the flat target; a cross-CCX pull needs
  ~tau of sustained backlog before the steal crosses the fabric.
- R_eff cost oracle (new reff_value BPF map): sized and indexed exactly like
  affinity_rank, paired 1:1, (u32)-1 sentinels past the topology end. The rank
  says which peer is next-nearest; the oracle says how far. Crucially the
  distance penalty is PRE-FOLDED by Rust at topology-detect:
  reff_value[cpu*MAX_AFFINITY_CANDIDATES+slot] = (R_eff(cpu,peer) *
  phi_dist_scale_q16) >> 16, stored directly in nanoseconds. The dispatch steal
  reads it as dist_extra and adds it to codel_target -- ONE indexed array read,
  no runtime multiply, no per-tick scale global on the hot path. This is the
  free-compute discipline the rest of the scheduler follows: the expensive
  derivation happens once at init where R_eff is already computed, not per-peer
  per-dispatch (the earlier per-tick-global form did the multiply on the hot
  path and was the dominant per-dispatch cost in the cache profile).
- phi_dist_scale_q16 is topology-owned (intf.h knob, consumed only by the Rust
  fold at topology-detect). All-zero reff_value -- monolithic / single-CCX /
  ccx_active false / --phi-scale 0 -- collapses dist_extra to 0 and the
  threshold to the flat codel_target: an exact no-op against prior behavior.
- topology.rs FIX A: SMT siblings share L2, so a move between them costs ~0
  cache; the L2 conductance rung is made very stiff (R_eff(SMT-sib) ~ 0), which
  lands the Phi migration barrier exactly at the physical-core / L2 boundary (a
  real cold-L2 refill) instead of penalizing free intra-core moves. lambda_2 is
  the cross-CCX Fiedler cut, independent of L2 stiffness, so tau is unchanged
  and codel_eq stays clamped at its ceiling -- oscillator timescales invariant.
  FIX B: phi_dist_scale_q16 is calibrated so the most distant pair (max R_eff)
  maps to ~tau of required steal-wait, an SMT sibling to ~0.

STEAL HOT-PATH FREE-COMPUTE (src/bpf/main.bpf.c)
- The dispatch STEP 1 peer scan is the dominant per-dispatch cache cost under
  wake-heavy load; three changes drop its cost to already-maintained reads:
  * The peer's backlog age is read from pcpu_enqueue_ns[peer] -- the stamp the
    scheduler already maintains -- instead of peeking the remote DSQ head and
    dereferencing that task's BPF storage per peer (a remote cacheline touch
    per candidate, the prior form).
  * pcpu_enqueue_ns is now an array of 64-byte-aligned pcpu_stamp structs, one
    cacheline per CPU, ending the false-sharing ping when STEP 1 CASes a peer's
    slot while that peer reads its own.
  * The scan is rate-limited to once per codel_target via a per-CPU
    last_spill_scan timestamp: a peer cannot accrue stealable backlog (aged past
    codel_target + dist_extra) faster than codel_target, so scanning more often
    is pure waste. Between scans dispatch falls through to overflow/idle and the
    tick remote scan still kicks any aged per-CPU DSQ, so no work strands.
  * An idle peer (stamp 0) skips the remote nr_queued touch entirely -- the
    common wake-heavy case pays nothing.

SHAPE ROUTING AND WARM-STAY PLACEMENT (src/bpf/main.bpf.c)
- FLOW SIGNATURE is the self-adaptive in-kernel detection the warm-anchor note
  called for -- a pipe pair and a fork storm want opposite placement, and no
  static threshold separates them, so each task earns its own classification.
  Every wakeup sets the waker CPU's bit in a per-task bitmap; the popcount is
  the task's distinct-partner cardinality, a topology-free read of the live
  communication graph's conductance. At maturity (ewma_age >= EWMA_AGE_MATURE)
  the shape is classified once and frozen: <= SHAPE_TIGHT_MAX (2) partners is a
  TIGHT pair/loop; a partner set spanning at least half of nr_cpu_ids is a STORM
  mesh; everything between defaults to TIGHT (latency-safe). Portable -- the
  STORM threshold scales with the machine, no hardcoded core geometry.
- Shape drives PLACEMENT, never the steal (a shape-gated steal once starved a
  STORM-classed audio thread on a busy core). Shape-routed warm placement (B):
  before scx_bpf_select_cpu_dfl's topology-blind any-idle pick -- which can seat
  a wakee on a cross-CCX idle core (cold L3, the storm's residual cache-miss
  source) -- the wakee is anchored on its last core and searched R_eff-near
  (same L2/CCX first). A TIGHT pair grabs its exact warm core when free; a STORM
  spreads. B v2: when no near sibling has room, the wakee stays over-depth on
  its OWN warm core rather than spilling to the shared node_dsq (which any core
  drains, scattering it cross-CCX cold); the deep per-CPU queue is bounded by
  the CoDel sojourn steal.
- WARM-STAY (warm_stay_anchor + enqueue TIER 0) is the placement dual of the
  steal threshold. A wakee whose last_cpu is UNCONGESTED -- its per-CPU sojourn
  (now - the maintained pcpu_enqueue_ns stamp, the same signal the STEP 1 steal
  reads) under codel_target -- is held on that warm core rather than fanned out
  to a cold idle sibling. Over codel_target, the anchor is genuinely backed up
  and the wakee fans out as before. Same yardstick both directions: a wakee
  stays on its warm core right up to the sojourn at which the steal machinery
  would relocate it.
- The busy placement lives in enqueue, NOT select_cpu's idle fast path. The
  hard-won rule from a prior place-on-busy attempt stands: a KICK_IDLE on a busy
  core is a no-op and strands the wakee (~90% deadline miss under saturation),
  so the idle fast path still NEVER places on busy. Instead select_cpu DEFERS --
  warm_stay_anchor returns the anchor and select_cpu returns it without
  dispatching, so the task flows to enqueue TIER 0, which seats it on last_cpu's
  per-CPU DSQ with KICK_PREEMPT (the kick a busy core actually honors).
  LAT_CRITICAL and kthreads are exempt (flee for immediacy); a full / congested
  anchor falls through to the existing idle-seek tiers unchanged.
- Result: deadline-miss out of the ~90% crater to sub-1% across 2-12C, latency
  P99 down 10-40x, no bimodality, audio clean through a fork storm. On the
  real-world cachyos suite PANDEMONIUM ADAPTIVE takes the lowest total time of
  any scheduler measured (EEVDF and scx_cake included) and lands top-cluster in
  the full scx field; the storm cache axis is the one
  quantified line item Phi's cache work targets.

ADAPTIVE PHI-COHERENCE (src/tuning.rs, src/adaptive.rs, src/scheduler.rs)
- The adaptive control plane was calibrated against the pre-Phi BPF and had
  begun to contend with the sharpened one: it evicted on a deadline the kernel
  no longer used, ran a slice the topology no longer wanted at low core count,
  and could not see the cross-CCX scatter its own knob choices induced. Five
  edits bring it back into coherence -- none rips out a subsystem; each
  re-points an existing knob at the value the kernel already maintains.
- SOJOURN FLOOR RE-ANCHORED TO THE PHI EQUILIBRIUM. The MWU sojourn floor was a
  dead 4ms constant -- a guard from when the BPF overflow rescue fired at a
  hand-tuned ~10ms. The rescue is now overflow_sojourn_rescue_ns =
  codel_target_equilibrium_ns (R_eff-derived, sub-ms to ~2ms), so the 4ms floor
  pinned the per-CPU eviction deadline three to nine times above the threshold
  warm-stay and the STEP 1 steal actually release at -- the adaptive layer held
  tasks long past the point the Phi gates assumed relief had arrived, the
  low-core LONG-RUN / LAT P99 tail. The floor now anchors to b.codel_eq_ns
  (guarded >= 200us; the +blended_slice dispatch-service-window guard retained),
  so userspace eviction and the kernel Phi release agree on one number.
  adaptive.rs seeds the MWU baseline with the live codel_eq at tick 0 --
  otherwise the floor falls back to the dead constant on any run whose regime
  never changes.
- LOW-CORE SLICE DISCIPLINE (scaled_regime_knobs). tau is LARGEST at low core
  count -- lambda_2 shrinks as cores drop -- so the tau slice cap ran loosest
  exactly where a wide batch slice does the most damage. At 2-4 cores the HEAVY
  profile's 4ms slice passed through untouched, and the regime detector latched
  HEAVY there because idle_pct reads low when warm-stay and the per-CPU landing
  deliberately bypass the idle fast path -- the kernel placement working
  correctly was misread as saturation, and the 4ms slice denied a
  latency-sensitive probe across many consecutive slices. That is the 2C/4C
  bimodal blowout: LONG-RUN / MIXED P99 spiking to tens of milliseconds on a
  different workload each run. The slice is now capped to the MIXED value at
  nr_cpus <= 4, where a wide slice buys no throughput (the long-run WORK numbers
  confirm it earned none); 8C/12C are untouched, where the wider slice earns
  throughput and the adaptive layer already wins.
- CROSS-CCX SCATTER PATHWAY (PATHWAY 6). MWU balanced six experts blind to
  migration -- the stats it reads carried dispatches, kicks, rescue, and L2
  hit/miss but no cross-CCX signal, so the controller picking slice and affinity
  could not see the migration storm those choices induced. nr_xccx, the
  per-placement-path cross-CCX landing counter, is promoted from diagnostic to a
  permanent control input and feeds a sixth loss pathway. adaptive.rs computes
  the placement-side scatter fraction -- XCCX_SEL_* + XCCX_ENQ_T1, EXCLUDING the
  Phi-correct steal and step5 work-conservation paths (deliberate moves MWU must
  not punish) -- as a percentage of dispatches. PATHWAY 6 is rising-edge gated
  past 20% (EEVDF's measured baseline is ~14%): a climbing scatter penalizes the
  THROUGHPUT, SATURATED, and FORK_STORM experts -- the ones favoring wide slices
  and weak/off affinity -- so the blend re-weights toward LATENCY / BALANCED and
  affinity STRONG. A stable high-scatter regime (8C+ throughput, where the trade
  wins) plateaus and is never penalized; only a climbing storm is.
- OSCILLATOR DEFER GATE MADE PHI-AWARE. MWU defers its rescue-driven pathways
  when the BPF oscillator has already moved, but it read only the bare codel
  window -- and the kernel releases warm-stay and the STEP 1 steal at
  codel_target + dist_extra, the nearest-peer Phi hold. With a large hold the
  bare window can read "loose" while the effective release is still hard, and
  MWU would stand down on genuine rescue pressure. OscillatorState gains
  home_dist_extra_ns (reff_value slot 0, the cheapest-peer hold warm_stay_anchor
  itself reads), and the defer decision qualifies position() against the
  EFFECTIVE release (codel_target + dist_extra) versus codel_eq. It degrades
  exactly: dist_extra 0 reduces to the prior bare-window behavior, codel_eq 0
  disables the absolute qualifier.
- FORK-STORM AFFINITY OFF -> WEAK (DV_AFFINITY). The BPF reads affinity_mode as
  a binary gate on the L2-sibling pre-filter; STRONG and WEAK are identical to
  it, only OFF differs. The FORK_STORM expert voted affinity OFF -- disabling
  the L2 pre-filter during exactly the storm warm placement exists to contain,
  seating wakees topology-blind and scattering them cross-CCX. It now votes
  WEAK, keeping the pre-filter on.
- The scatter signal is surfaced through the existing bench suite, not a
  throwaway probe: the per-path cross-CCX counts and the placement-side scatter
  percentage ride the [KNOBS] shutdown line in both BPF and ADAPTIVE modes, and
  pandemonium-tests.py emits pandemonium_bench_xccx_scatter_pct and
  pandemonium_bench_xccx_path{path=...}, so every run measures which path
  scatters and whether ADAPTIVE scatters less than the raw BPF.
- Result: the 4ms low-core slice that enabled the 2C/4C bimodal blowout is gone,
  and the post-fix run lands low-core MIXED and LONG-RUN P99 at the BPF column or
  below (2C LONG-RUN leads BPF outright) with no tens-of-millisecond tail. On the
  cachyos suite ADAPTIVE takes the lowest total time of any scheduler measured
  and wins four of eight workloads outright (argon2, xz, x265,
  ffmpeg-compilation); on the storm messaging bench its cache-misses fall below
  the raw BPF. The adaptive layer is now a net gain over the kernel it drives,
  not a tax on it.

HYGIENE PASS
- src/chaos.rs: removed four dead helpers (min, max, hvg_entropy,
  hvg_mean_degree) -- all #[allow(dead_code)], zero callers; hvg_stats is
  the live amortized path that derives both lambda and entropy from
  hvg_degrees in one O(N^2) pass. -79 lines.
- Cargo.toml: dropped the redundant libc from [dev-dependencies]; libc is
  a real [dependencies] entry (event.rs / log.rs / cli/probe.rs), the dev
  copy a leftover from the inline #[cfg(test)] tests removed in v5.11.0.
- main.bpf.c: two residual vtime fossils the QUEUE CLEANUP pass left --
  "VTIME LAG CAP" (it is the sojourn-warp clamp now) and a "VTIME-CEILING
  STATICS" reference (vtime_ceiling is gone). Two further audit-flagged
  comments were checked against the code and left unchanged because the
  claims were wrong (STEP 2 genuinely uses overflow_sojourn_rescue_ns; the
  SAFETY NET is an intentional unnumbered backstop).
- main.bpf.c: cut the write-only pcpu_idle_start_ns / pcpu_idle_total_ns
  arrays, their pandemonium_update_idle accounting, and the .update_idle ops
  registration -- measure-only scaffolding for a DNB / Kim-Jo regime consumer
  that never landed. That consumer needs windowed per-CPU + cross-CPU
  (slow-loop) processing the existing chaos + MWU regime detector already
  covers, so the arrays were dead weight; the idea is retained in research
  notes, not as write-only code.
- Added PANDEMONIUM-HYGIENE-AUDIT.md: the full four-axis audit with
  per-finding verification status and the deferred worklist
  (MAX_AFFINITY_CANDIDATES single-sourcing).

MAX_CPUS MASKING FOOTGUN (main.bpf.c)
- The four & (MAX_CPUS-1) index masks are gone. They silently aliased an
  out-of-range CPU to a valid slot instead of failing; behavior-neutral in
  practice (indices are always < nr_cpu_ids <= MAX_CPUS) but a latent
  silent-wrong. Two were redundant masks inside existing < MAX_CPUS guards
  (dropped); the two tick-scan sites became explicit
  if (scan_cpu >= MAX_CPUS) continue; skips. No computed value changes.
- This de-foot-guns the existing MAX_CPUS usage; it does NOT remove
  MAX_CPUS (still sizes the per-CPU arrays + MAX_AFFINITY_CANDIDATES).
  Full removal (per-CPU arrays -> load-time maps) remains deferred and
  bench-gated. Verifier-clean; bench confirms no regression.

REFERENCES (README.md)
- The reference list is brought to comprehensive coverage of the utilized
  foundations -- every implemented mechanism now carries its primary citation.
  Added Fiedler (algebraic connectivity / lambda_2), Blumofe-Leiserson (work
  stealing), Bandt-Pompe (permutation entropy), Luque-Lacasa (horizontal
  visibility graph), Marwan et al. (recurrence quantification), and Butterworth
  (maximally-flat filter damping). Topology, steal, the chaos primitives, and
  the adaptive damping were previously uncited; each addition verified against
  the primary source.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3885,7 +3885,7 @@ dependencies = [
 
 [[package]]
 name = "scx_pandemonium"
-version = "5.9.1"
+version = "5.12.0"
 dependencies = [
  "anyhow",
  "clap",
--- a/scheds/rust/scx_pandemonium/Cargo.toml
+++ b/scheds/rust/scx_pandemonium/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scx_pandemonium"
-version = "5.9.1"
+version = "5.12.0"
 edition = "2021"
 description = "A behavioral, adaptive sched_ext scheduler with three-tier classification, L2 affinity, and process learning"
 license = "GPL-2.0-only"
@@ -14,9 +14,6 @@ clap = { version = "4", features = ["derive"] }
 ctrlc = { version = "3", features = ["termination"] }
 flate2 = "1"
 regex = "1"
-
-[dev-dependencies]
-libc = "0.2"
 
 [build-dependencies]
 scx_cargo = { path = "../../../rust/scx_cargo", version = "1.1.1" }
--- a/scheds/rust/scx_pandemonium/README.md
+++ b/scheds/rust/scx_pandemonium/README.md
@@ -1,178 +1,149 @@
 # PANDEMONIUM
 
-A Linux kernel scheduler for sched_ext, built in Rust and C23. PANDEMONIUM classifies every task by behavior — wakeup frequency, context switch rate, runtime, sleep patterns — and adapts scheduling decisions in real time. A critically-damped harmonic oscillator drives CoDel-inspired stall detection with the literal RFC 8289 sojourn metric and an R_eff-derived equilibrium reference. Resistance affinity (effective resistance from the Laplacian pseudoinverse of the CPU topology graph) provides topology-aware task placement for pipe/IPC storms. Multiplicative Weight Updates (MWU) balance six competing expert profiles across four loss pathways.
+A Linux kernel scheduler for sched_ext, built in Rust and C23. PANDEMONIUM classifies every task by behavior — wakeup frequency, context switch rate, runtime, sleep patterns — and adapts scheduling decisions in real time. A damped harmonic oscillator drives CoDel-inspired stall detection with the literal RFC 8289 sojourn metric and an R_eff-derived equilibrium reference. Resistance affinity (effective resistance from the Laplacian pseudoinverse of the CPU topology graph) provides topology-aware task placement for pipe/IPC storms. A migration potential Φ — R_eff priced against the queueing relief a move buys — gates cross-domain work stealing, so a steal crosses a cache boundary only when the backlog it relieves outweighs the cache cost. Multiplicative Weight Updates (MWU) balance six competing expert profiles across six loss pathways — one of them a cross-CCX scatter signal that holds the adaptive layer's knob choices in line with the Φ placement it sits above.
 
-Three-tier behavioral dispatch, overflow sojourn rescue, longrun detection, tier-aware preempt scaling, sleep-informed batch tuning, classification-gated DSQ routing, workload regime detection, vtime ceiling with new-task lag penalty, hard starvation rescue, and a persistent process database that learns task classifications across lifetimes.
+Three-tier behavioral dispatch, overflow sojourn rescue, longrun detection, tier-aware preempt scaling, sleep-informed batch tuning, classification-gated DSQ routing, workload regime detection, flow-signature shape routing (per-task partner-cardinality classification into TIGHT loops vs STORM meshes), a migration-potential-gated R_eff work steal, a Φ-priced warm-stay home anchor, a sojourn selector with bounded maturity-gated tier warp, a per-CPU tau-derived preempt, an RT-policy latency floor, hard starvation rescue, and a persistent process database that learns task classifications across lifetimes.
 
-PANDEMONIUM is included in the [sched-ext/scx](https://github.com/sched-ext/scx) project alongside scx_rusty, scx_lavd, scx_cosmos, scx_cake and the rest of the sched_ext family. Thank you to Piotr Gorski and the sched-ext team. PANDEMONIUM is made possible by contributions from the sched_ext, CachyOS, Gentoo, OpenSUSE, Arch and Ubuntu communities within the Linux ecosystem.
+PANDEMONIUM is included in the [sched-ext/scx](https://github.com/sched-ext/scx) project alongside scx_rusty, scx_lavd, scx_cosmos and the rest of the sched_ext family. Thank you to Piotr Gorski and the sched-ext team. PANDEMONIUM is made possible by contributions from the sched_ext, CachyOS, Gentoo, OpenSUSE, Arch and Ubuntu communities within the Linux ecosystem.
 
 ## Performance
 
-12 AMD Zen CPUs, kernel 6.18+, clang 21. The tables below report v5.8.0 PANDEMONIUM numbers from a single-iteration bench-scale run across 2C/4C/8C/12C; the v5.9.0 deltas at 12C are summarized in the next subsection. EEVDF and scx_bpfland baselines are best-3-of-N historical (28 and 26 complete runs across 75 bench-scale sessions) — those baselines don't drift across PANDEMONIUM versions. v5.8.0 closed the 2C structural starvation class via the rescue-fairness fix and the missing CPU-bound demotion in stopping() that was leaving long-runners stuck at TIER_INTERACTIVE (longrun WORST 19s -> 13ms at 2C, mixed WORST 29s -> 16ms at 2C). Bench-fork-thread lands within 1% of EEVDF and ~40% ahead of scx_bpfland.
+12 AMD Zen CPUs, kernel 7.0.10, clang 21. Numbers below are from a single fresh-boot 3-iteration bench-scale run (2026-06-01, v5.12.0) — all five schedulers (EEVDF, scx_bpfland, scx_cake, PANDEMONIUM BPF + ADAPTIVE) measured in the same run.
 
-### v5.9.0 Deltas (12C, single iteration vs v5.8.0 24-run mean)
+### v5.12.0
 
-| Metric            | v5.8.0 mean | v5.9.0     | Delta              |
-|-------------------|-------------|------------|--------------------|
-| Burst P99         | 392us       | **83us**   | -4.7x              |
-| Longrun P99       | 489us       | **244us**  | -2.0x              |
-| Mixed P99         | 724us       | **139us**  | -5.2x              |
-| IPC P99           | 107us       | **36us**   | -3.0x              |
-| Deadline miss     | 3.56%       | **0.1%**   | -35x               |
-| Schbench P99      | 75us        | 77us       | hold               |
-| App launch P99    | 2,983us     | 4,692us    | regression (n=1)   |
-
-Architectural changes in v5.9.0 (affinity threading through the R_eff spill chain, `MAX_AFFINITY_CANDIDATES = MAX_CPUS >> 3` topology coverage, slice-relative INTERACTIVE→BATCH demotion, tier-priority tick preempt) address structural classes that did not show on the v5.8.0 12C single-iteration cells. The 32C affinity-stranding class closure was confirmed by a 3.5+ hour real-workload run on Threadripper (game + KVM + libvirt, no watchdog kills, no stalls); multi-iteration sweeps across 2C/4C/8C/12C are pending. The app-launch regression is single-iteration and likely under-sampled — to be retested under multi-run aggregation.
+v5.12.0 drives the adaptive control plane into Φ-coherence with the BPF placement it sits above: eviction keyed to the same equilibrium the kernel rescues at, a cross-CCX scatter loss pathway that lets MWU see migration, Φ-aware oscillator gating, and a core-count slice discipline that holds the low-core regime to a tight slice where a wide one buys no throughput. The Rust layer now reinforces the kernel's placement instead of contending with it — ADAPTIVE tracks BPF across the board (2C longrun 182µs vs 181µs, steady across three iterations), and PANDEMONIUM leads the full scheduler field on every latency, burst, longrun, mixed, and deadline-miss table below. The two corners it still trails — IPC round-trip and app launch — are called out plainly under their tables.
 
 ### P99 Wakeup Latency (interactive probe under CPU saturation)
 
-| Cores | EEVDF    | scx_bpfland | PANDEMONIUM (BPF) | PANDEMONIUM (ADAPTIVE) |
-|-------|----------|-------------|-------------------|------------------------|
-| 2     | 2,058us  | 2,004us     | **87us**          | 93us                   |
-| 4     | 1,246us  | 2,003us     | **65us**          | 68us                   |
-| 8     | 425us    | 2,003us     | **66us**          | 68us                   |
-| 12    | 344us    | 2,002us     | **68us**          | 70us                   |
-
-Sub-95us in both modes at every core count.
+| Cores | EEVDF    | scx_bpfland | scx_cake  | PANDEMONIUM (BPF) | PANDEMONIUM (ADAPTIVE) |
+|-------|----------|-------------|-----------|-------------------|------------------------|
+| 2     | 2,500us  | 2,873us     | 1,029us   | **74us**          | 173us                  |
+| 4     | 3,865us  | 1,318us     | 1,916us   | **61us**          | 62us                   |
+| 8     | 988us    | 1,943us     | 62us      | **61us**          | 64us                   |
+| 12    | 107us    | 1,956us     | **56us**  | 63us              | 63us                   |
 
 ### Burst P99 (fork/exec storm under CPU saturation)
 
-| Cores | EEVDF    | scx_bpfland | PANDEMONIUM (BPF) | PANDEMONIUM (ADAPTIVE) |
-|-------|----------|-------------|-------------------|------------------------|
-| 2     | 2,262us  | 2,006us     | 189us             | **69us**               |
-| 4     | 3,223us  | 2,003us     | 687us             | **207us**              |
-| 8     | 2,331us  | 2,004us     | 611us             | **93us**               |
-| 12    | 1,891us  | 2,001us     | **79us**          | 196us                  |
-
-Both modes beat EEVDF and scx_bpfland by a wide margin. Burst-storm cost is workload-sensitive across iterations.
+| Cores | EEVDF    | scx_bpfland | scx_cake   | PANDEMONIUM (BPF) | PANDEMONIUM (ADAPTIVE) |
+|-------|----------|-------------|------------|-------------------|------------------------|
+| 2     | 2,565us  | 1,303us     | 14,189us   | **64us**          | 150us                  |
+| 4     | 2,090us  | 1,052us     | 24,764us   | 63us              | **61us**               |
+| 8     | 2,865us  | 1,827us     | 27,869us   | **63us**          | 2,048us                |
+| 12    | 1,871us  | 1,965us     | 27,948us   | **71us**          | 118us                  |
 
 ### Longrun P99 (interactive latency with sustained CPU-bound long-runners)
 
-| Cores | EEVDF    | scx_bpfland | PANDEMONIUM (BPF) | PANDEMONIUM (ADAPTIVE) |
-|-------|----------|-------------|-------------------|------------------------|
-| 2     | 2,293us  | 2,000us     | 2,327us           | **559us**              |
-| 4     | 1,421us  | 2,003us     | 870us             | **68us**               |
-| 8     | **60us** | 2,003us     | **439us**         | 655us                  |
-| 12    | 126us    | 2,002us     | 674us             | **80us**               |
-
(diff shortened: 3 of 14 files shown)


wllclngn:
PR was pulled back due to suspected issues. However, after more extensive Tests said issues were found to be false. Reopening PR.

sirlucjan (approved):

PR #3631, opened 2026-06-08
scx_beerland: Sleep virtual-lag and per-CPU load balancing

arighi:
Rework the core scheduling logic around a sleep virtual-lag model and an event-driven, per-CPU load balancer, dropping the previous heuristics based on wakeup frequency and global CPU utilization.

Sleep virtual lag:

 - Replace the wakeup-frequency / awake-vtime deadline heuristic with a bounded sleep virtual lag: on voluntary sleep save the task's lag relative to vtime_now, clamped by weight and slice_lag, so sleepers carry a limited amount of credit or debt. Re-align the vruntime on wakeup (and in ops.running() for directly dispatched tasks).
 - Add wakeup preemption: an eligible waking sleeper can preempt the current task on its CPU (SCX_ENQ_PREEMPT) when its deadline is earlier, comparing against the current task's up-to-date vruntime.

Per-CPU busy filter and load balancing:

 - Remove the user-space /proc/stat polling that maintained a global cpu_util value (and the -p/--polling_ms option). Replace the global "system busy" notion with a lightweight, more fine-grained event-driven per-CPU busy map: a CPU is marked busy for one slice in ops.stopping() when the task consumed enough CPU time (utime+stime) relative to the busy_threshold.
 - The load balancer (in ops.dispatch()) steals the earliest-vruntime task from CPUs currently marked busy; cpu_busy_thresh now expresses an accounted-time fraction and defaults to 0 (disabled): with the filter off, the load balancer can steal from any CPU, prioritizing latency and responsiveness through fast, unconditional load balancing.

NUMA awareness:

 - Add NUMA optimizations: per-node idle cpumask lookups and SCX_OPS_BUILTIN_IDLE_PER_NODE, automatically disabled on non-NUMA systems and overridable via -n/--disable_numa.

Cleanups:

 - Drop preferred idle CPU scanning entirely (preferred_idle_scan, preferred_cpus[], pick_idle_cpu_scan/pick_idle_cpu_pref_smt, -P).
 - Simplify the dispatch paths to use a flat slice_ns slice (removing task_slice() and do_direct_dispatch()) and add SCX_OPS_ALWAYS_ENQ_IMMED.
 - Refactor keep_running() to account for per-CPU tasks, SMT contention and the primary domain via the new is_primary_cpu() helper.
 - Lower the default slice_us_lag from 40ms to 20ms.


arighi:
This PR is similar to https://github.com/sched-ext/scx/pull/362 in scx_cosmos, later reverted by https://github.com/sched-ext/scx/pull/3630, due to stall regressions reported by CI.

Let's see if this approach is more successful in scx_beerland (the per-CPU DSQ approach should help).

scheds/rust/scx_beerland/src/bpf/main.bpf.c
```diff
@@ -691,76 +716,78 @@ void BPF_STRUCT_OPS(beerland_enqueue, struct task_struct *p, u64 enq_flags)
...
+
+	now = bpf_ktime_get_ns();
 
-	/*
-	 * Pick the task with the lowest vruntime within the same LLC.
-	 *
-	 * Restricting rebalancing to the LLC improves cache locality and
-	 * also reduces lock contention on CPU runqueues.
-	 */
 	bpf_for(cpu, 0, nr_cpu_ids) {
```
hodgesds: Will this lead to CPU 0 trying to steal the most aggressively? Might work ok on a single LLC, may want to offset by current CPU.

hodgesds (approved):

scheds/rust/scx_beerland/src/bpf/main.bpf.c
arighi: Yeah, if a lot of CPUs end up having the exact same min_vtime we prioritize the one with a lower CPU ID. I'll add the current CPU offset, looks like a good idea, thanks!

PR #3635, opened 2026-06-09
scx: add lib/bpf_cpumask.h and dedupe the cpumask-alloc helpers

jkkm:
Three commits:
1. `scx: add lib/bpf_cpumask.h for bpf_cpumask kptr lifecycle helpers`
2. `scx_lavd,bpfland,flash,tickless: use create_save_bpfmask() from lib`
3. `scx: add idempotent init_bpfmask() and adopt in five schedulers`
## Problem
Schedulers hand-roll two near-identical cpumask-allocation helpers:
- the bare `bpf_cpumask_create()`/`bpf_kptr_xchg()`/`bpf_cpumask_release()`
  sequence that installs a fresh mask into a kptr slot; and
- an idempotent wrapper around it ("allocate only if the slot is empty")
  used where a mask is populated incrementally.
`lib/percpu.h` already had the bare helper (`create_save_bpfmask`) but
bundled with the arena `scx_bitmap` library and the per-CPU storage map,
so it could not be reused without pulling those in.
## Change
- Add `lib/bpf_cpumask.h` (depends only on `<scx/common.bpf.h>`):
  - `create_save_bpfmask()` — allocate + install, releasing any prior mask.
  - `init_bpfmask()` — idempotent: allocate only if the slot is empty.
  `lib/percpu.h` now includes it (dropping its two debug `bpf_printk()`s).
- `create_save_bpfmask()`: adopt in `scx_lavd`, `scx_bpfland`,
  `scx_flash`, `scx_tickless` (their local copy was byte-identical).
- `init_bpfmask()`: adopt in `scx_bpfland`, `scx_flash`, `scx_tickless`
  (had a local `init_cpumask` wrapper) and `scx_cosmos`, `scx_beerland`
  (had it inlined). The idempotency is load-bearing — their
  `enable_primary_cpu`/`enable_sibling_cpu` syscall progs build a mask one
  CPU at a time, so the guard must be preserved (it is, in the shared
  helper).
This header manages the kernel `struct bpf_cpumask` kptr object and is
deliberately distinct from `lib/cpumask.h`, the arena-resident
`scx_bitmap` reimplementation.
## Left as-is (deliberate per-scheduler behavior)
- `scx_rusty` / `scx_layered`: `scx_bpf_error()` (fatal) on failure.
- `scx_p2dq`: treats an already-populated slot as an error.
- `scx_mitosis`: inline create+xchg (no named helper).
These encode intentional error policy and aren't behavior-identical, so
they're not converted here.
## Testing
`cargo build` + `veristat` on an NR_CPUS=1024 kernel: all converted
schedulers (`scx_lavd`, `scx_bpfland`, `scx_flash`, `scx_tickless`,
`scx_cosmos`, `scx_beerland`) and `scx_p2dq` (a `lib/percpu.h` consumer)
build with 0 verifier failures. Net ~130 lines of duplication removed.
No functional change (other than the dropped debug prints in percpu.h).


scheds/rust/scx_tickless/src/bpf/main.bpf.c
```diff
@@ -612,47 +613,10 @@ s32 BPF_STRUCT_OPS(tickless_init_task, struct task_struct *p,
...
-	cpumask = bpf_kptr_xchg(p_cpumask, cpumask);
-	if (cpumask)
-		bpf_cpumask_release(cpumask);
-
-	return 0;
-}
 
 /*
  * Initialize a cpumask (if not already initialized).
  */
```
hodgesds, line 648: I think these comments aren't needed?

arighi (commented):
This looks good to me, we can save some code duplication.

How about renaming `init_bpfmask()` -> `init_bpf_cpumask()` and `create_save_bpfmask()` to `create_bpf_cpumask()`? I think it'd be more clear to express that we're dealing with a cpumask.

htejun:
Yeah, looks good to me too. @jkkm, can you incorporate the suggestions?

multics69:
@jkkm -- It looks good to me. As long as @arighi's comments are addressed, I am okay to merge it.

PR #3636, opened 2026-06-09
scx_cosmos: Avoid verifier failure when skipping empty nodes

arighi:
Commit 8dc8ff4c85bd ("scx_cosmos: Skip CPU-less NUMA nodes when creating per-node DSQ") introduced a regression on NUMA systems by reloading nctx->cpumask in cosmos_init() and passing it to bpf_cpumask_empty().

This triggers the following verifier error:

  ; bpf_cpumask_empty(cast_mask(nctx->cpumask))) @ main.bpf.c:1506
  110: (85) call bpf_cpumask_empty#105356
  R1 must be a rcu pointer

Avoid the second cpumask kptr read by making init_node() detect CPU-less nodes while it already has the RCU-proven cpumask pointer; cosmos_init() now skips DSQ creation for any node that init_node() cannot initialize, preserving the CPU-less node optimization without triggering the verifier failure.

Fixes: 8dc8ff4c85bd ("scx_cosmos: Skip CPU-less NUMA nodes when creating per-node DSQ")

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -1439,11 +1439,15 @@ void BPF_STRUCT_OPS(cosmos_exit_task, struct task_struct *p,
 
 /*
  * Initialize a NUMA node context.
+ *
+ * Return 0 if @node contains at least one CPU, -ENODEV if it is CPU-less, or
+ * another negative errno on failure.
  */
 static int init_node(int node)
 {
 	struct bpf_cpumask *cpumask;
 	struct node_ctx *nctx;
+	bool has_cpus = false;
 	u32 cpu;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_cosmos/src/bpf/main.bpf.c
```diff
@@ -1486,24 +1495,14 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(cosmos_init)
...
+			 * initialize, including CPU-less NUMA nodes (e.g.,
+			 * GPU-memory or CXL-memory nodes): with no CPU to
+			 * consume from it, the DSQ would never be used.
 			 */
-			nctx = try_lookup_node_ctx(node);
-			if (!nctx || !nctx->cpumask ||
-			    bpf_cpumask_empty(cast_mask(nctx->cpumask)))
+			err = init_node(node);
+			if (err)
 				continue;
```
nvax-r, line 1506: If the failure happens on CPU-bearing node, it silently fall through and leave it with no DSQ.
Maybe we should make it abort at runtime instead of just going through
Something like that following
```
err = init_node(node);
if (err == -ENODEV)
    continue;
if (err) {
    scx_bpf_error("failed to initialize NUMA node %d: %d", node, err);
    return err;
}
```

nvax-r (commented):
Thanks for this, just a small question

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
arighi: @nvax-r I considered doing that, but `init_node()` can only fail if the node has no CPUs or if the node ID doesn't exist (e.g., sparse node numbering). In both cases, I think the right behavior is to gracefully skip the node rather than treat it as a failure.

hodgesds (approved):

scheds/rust/scx_cosmos/src/bpf/main.bpf.c
nvax-r: Makes sense, then I have no problem, thanks for this !

nvax-r:
LGTM

PR #3637, opened 2026-06-10
scx_lavd: avoid a stray CPU pick in pick_idle_cpu() error paths

multics69:
If an error occurs in pick_idle_cpu(), we rely on pick_random_cpu() to choose an arbitrary CPU from ctx->{a_mask, o_mask}, which is already filtered by p->cpus_ptr. However, if the error occurs before ctx->{a_mask, o_mask} are initialized, we end up choosing a CPU from garbage {a_mask, o_mask}, which may not be in p->cpus_ptr.

We harden this case by choosing prev_cpu, which is already in p->cpus_ptr. Note that such an error cannot happen in reality -- the per-CPU scratch masks always exist and the failure handling is only there to satisfy the BPF verifier -- so this is for completeness.


etsal (approved):

bboymimi (approved):
LGTM.

PR #3638, opened 2026-06-10
lib/atq: Fix error paths, size accounting, insert races

etsal:
Address three possible sources of bugs in ATQs:

- Size accounting was incomplete for direct ATQ removals, like in the case of dequeues. Properly decrement the queue size when we remove an item as opposed to popping.
- Atomically set the owner ATQ of a task. The ->atq field of a taskc is not actually protected by the ATQ lock, because different threads may try to enqueue the same task to different ATQs at the same time. Atomically set the owner ATQ so those races have a single winner.
- Do not unset the atq field of a task if the removal fails.


multics69:
@etsal -- The first commit seems irrelevant.

etsal:
Sorry about that, I included it in a rebase and it got desynced from the main branch. I will re-push

multics69 (approved):

PR #3639, opened 2026-06-10
lib/cgroup_bw: read ->atq with READ_ONCE and treat -EALREADY as benign

multics69:
A task's scx_task_common.atq is the pointer that identifies which BTQ currently owns the task, if any. cbw_put_aside() can run concurrently for the same task when an initial .enqueue() races a .dequeue()/ .enqueue() pair, so the same task may be inserted into or deleted from two different BTQs in different LLCs. The ATQ spinlock does not help because the update race on taskc->atq happens across two different ATQs.

Commit 2295cd4cc341 ("lib/atq: Handle insert races to different ATQs") resolves that race by claiming ->atq with an atomic cmpxchg in scx_atq_insert_vtime_unlocked(), so a losing inserter is rejected with -EALREADY rather than corrupting the rbtree.

Update the cgroup_bw consumers of ->atq to match the writer:

- Treat -EALREADY from scx_atq_insert_vtime_unlocked() as the benign "already in BTQ" outcome rather than a hard error, and return 0 so the caller does not also enqueue the task to a DSQ.

- Read ->atq with READ_ONCE() in cbw_put_aside()'s fast-path check and in scx_cgroup_bw_is_task_throttled(). Both are lockless reads of a field now mutated by an atomic writer; READ_ONCE() pairs with it.

### DIFF
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -1845,14 +1845,14 @@ int cbw_put_aside(u64 ctx, u64 vtime, u64 cgrp_id)
 		return -ESRCH;
 	}
 
-	if (taskc->atq != NULL) {
-		/*
-		 * Not really a bug: The initial .enqueue() may race with
-		 * a pair of .dequeue()/.enqueue() calls, and cause two
-		 * instances of this function to happen simultaneously
-		 * for the task. This should be rare, but possible.
-		 * The spinlock turns the race into a benign one.
-		 */
+	/*
+	 * A task can be claimed by only one BTQ at a time. The atomic cmpxchg
+	 * of ->atq inside scx_atq_insert_vtime_unlocked() elects a single
+	 * winner; the loser sees the task already queued. That benign case is
+	 * detected either here on the fast path or as EALREADY returned by the
+	 * insert below.
+	 */
+	if (READ_ONCE(taskc->atq) != NULL) {
 		cbw_dbg("Possible double enqueue detected.");
 		scx_atq_unlock(btq);
 		cbw_warn("put_aside skipped: already in BTQ; cgid=%llu", cgrp_id);
@@ -1860,10 +1860,14 @@ int cbw_put_aside(u64 ctx, u64 vtime, u64 cgrp_id)
 	}
 
 	ret = scx_atq_insert_vtime_unlocked(btq, taskc, vtime);
-	if (ret)
-		cbw_err("Failed to insert a task to BTQ: %d", ret);
-
 	scx_atq_unlock(btq);
+
+	if (unlikely(ret == -EALREADY)) {
+		cbw_warn("put_aside skipped: already in BTQ; cgid=%llu", cgrp_id);
+		return 0;
+	} else if (unlikely(ret)) {
+		cbw_err("Failed to insert a task to BTQ: %d", ret);
+	}
 
 	return ret;
 }
@@ -2638,7 +2642,7 @@ __hidden
 int scx_cgroup_bw_is_task_throttled(u64 taskc)
 {
 	scx_task_common *ctx = (scx_task_common *)taskc;
-	return ctx && (ctx->atq != NULL);
+	return ctx && (READ_ONCE(ctx->atq) != NULL);
 }
 
 /**


etsal:
My only nit is, is READ_ONCE() really necessary? The read cannot be optimized out or moved since it's in the control path.

lib/cgroup_bw.bpf.c
```diff
@@ -1845,26 +1845,33 @@ int cbw_put_aside(u64 ctx, u64 vtime, u64 cgrp_id)
...
-		/*
-		 * Not really a bug: The initial .enqueue() may race with
-		 * a pair of .dequeue()/.enqueue() calls, and cause two
-		 * instances of this function to happen simultaneously
-		 * for the task. This should be rare, but possible.
-		 * The spinlock turns the race into a benign one.
-		 */
+	/*
+	 * A task can be claimed by only one BTQ at a time. The initial
+	 * .enqueue() may race with a pair of .dequeue()/.enqueue() calls,
```
daidavid: Let's be eagle eyed about how these races happen in the kernel and be very succinct when describing how these races happen.

AFAICT:
Almost all the paths that uses enqueue involves holding on the the pi_lock that would prevent the multi-enqueue problem on the same task that's being described with the following exceptions:

1. ttwu_queue_wakelist can race, but we don't allow SCX_OPS_ALLOW_QUEUED_WAKEUP in LAVD, so that's ruled out.

2. smp_cond_load_acquire(&p->on_cpu, !VAL) protects against the race between ttwu and the migrate_disable_switch path.

Wouldn't it be much more likely that we race on the other paths involving cbw_put_aside instead? such as cbw_cgroup_bw_offline?

etsal: After giving it some thought, I am looking into the following:

- We are calling cgroup_init(), which causes the parent cgroup to become a non-leaf and frees it. We go through the ATQ, pop it, and push the tasks to the root cgroup (which I assume is what we want).
- The task currently being moved to a different (throttled) cgroup. This causes it to be dequeue/reenqueue.
- The task gets reenqueued both to the root cgroup and the new cgroup.

This is slightly racy already because we are emptying out the now non-leaf parent's ATQ while initializing the new cgroup. I am not sure whether this means the tasks have already moved cgroups or not, but in whatever case there is some asynchrony between when the tasks move between cgroups and when we move the BTQ contexts between them.

IIRC the .dequeue()/.enqueue() comment was from back when .dequeue() did not work, but ATQ worked around that fact so we had to explicitly state that fact. I think we should remove it.

multics69:
> My only nit is, is READ_ONCE() really necessary? The read cannot be optimized out or moved since it's in the control path.

I agree. However, I think it will be better to add READ_ONCE() to match the atomic update of taskc->atc for maintainability without making any assumption compiler optimizations.

lib/cgroup_bw.bpf.c
multics69: > Wouldn't it be much more likely that we race on the other paths involving cbw_put_aside instead? such as cbw_cgroup_bw_offline?

After https://github.com/sched-ext/scx/pull/3615, the race between cbw_put_aside and cbw_cgroup_bw_offline should be gone. If there is, that's a bug. ^^;;;

multics69: > After giving it some thought, I am looking into the following:
> 
> * We are calling cgroup_init(), which causes the parent cgroup to become a non-leaf and frees it. We go through the ATQ, pop it, and push the tasks to the root cgroup (which I assume is what we want).
> * The task currently being moved to a different (throttled) cgroup. This causes it to be dequeue/reenqueue.
> * The task gets reenqueued both to the root cgroup and the new cgroup.
> 
> This is slightly racy already because we are emptying out the now non-leaf parent's ATQ while initializing the new cgroup. I am not sure whether this means the tasks have already moved cgroups or not, but in whatever case there is some asynchrony between when the tasks move between cgroups and when we move the BTQ contexts between them.

I think this can happen. 

> 
> IIRC the .dequeue()/.enqueue() comment was from back when .dequeue() did not work, but ATQ worked around that fact so we had to explicitly state that fact. I think we should remove it.

I didn't think much about those comments. I'd rather drop those comments ("The initial .enqueue() may race with a pair of .dequeue()/.enqueue() calls, so two instances of this function can run for the same task concurrently, possibly targeting BTQs in different LLCs.") to avoid the confusion.

etsal:
> > My only nit is, is READ_ONCE() really necessary? The read cannot be optimized out or moved since it's in the control path.
> 
> I agree. However, I think it will be better to add READ_ONCE() to match the atomic update of taskc->atc for maintainability without making any assumption compiler optimizations.

Fair enough, makes sense. I found only one instance for ATQs where a compiler reorder/merge of a read could cause a problem, but the generated code looked fine.

etsal (approved):

PR #3642, opened 2026-06-10
scx_lavd: fix an SCX_DSQ_LOCAL_ON runtime error in lavd_enqueue

likewhatevs:
`lavd_misplaced_local_on` (a new ktstr test, gated behind the `ktstr-tests`
feature) drives scx_lavd into a runtime error. Against unmodified scx_lavd it
fails consistently:

    SCX_DSQ_LOCAL[_ON] target CPU N not allowed for <task>
    scx_exit <- task_can_run_on_remote_rq <- dispatch_to_local_dsq

Two changes to `lavd_enqueue`, described by their effect on that test:

- Fall back to the task's current CPU when the picked CPU is invalid or not in
  the task's allowed mask. This takes the test from a consistent failure to a
  sporadic one (same error).
- Direct-dispatch `SCX_DSQ_LOCAL_ON` only when the picked CPU is the task's
  current CPU; otherwise enqueue to the domain DSQ. The remaining sporadic
  failure surfaced as a runnable-task stall (`runnable task stall (<task>
  failed to run for N s)`); after this the test passes.

The test is feature-gated, so a normal scx_lavd build is unaffected — the
harness is pulled in only with `--features ktstr-tests`.

## Test plan
- [x] `lavd_misplaced_local_on` passes with the fix (`cargo ktstr test`, `--features ktstr-tests`).
- [x] Without the fix, the test reproduces the runtime error above.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
- "gimli",
+ "gimli 0.32.3",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


htejun:
I don't think this is supposed to happen. Not in the enqueue path with rq lock held. This needs root-causing.

multics69:
@likewhatevs -- I think https://github.com/sched-ext/scx/pull/3692 should fix the problem.

PR #3643, opened 2026-06-10
scx_cake: align README with shipped behavior

kurtislin:
The scx_cake README documents several mechanisms that don't exist in the code, and in most cases the code comments already say so. This aligns the README with what actually ships (the in-tree scx_cake is unchanged since v1.1.1).

The bigger items:

- The "VPROT: Preemption Protection" section (O(1) victim finding, per-class protection thresholds, SCX_KICK_PREEMPT kicks) is not implemented anywhere in executable BPF code. cake_select_cpu's own comment says "DSQ ordering handles all priority — no preemption or kicks needed", so the section now describes the DSQ-ordering-only design that actually exists. The nr_vprot_suppressed counter has no BPF writer, so it's gone from the TUI docs too.
- "Kfunc Tunneling" describes caching scx_bpf_now() in cached_now. That staging was removed; the comment at the removal site says it "had ZERO READERS".
- The class table contradicts tier_base[] = {8192, 0, 49152, 32768}: the real dispatch order is GAME < NORMAL < BG < HOG, while the README listed HOG before BG with different values.
- The DRR++ deficit section claims per-bout drain, bonus removal on exhaustion and a GAME exemption. In the shipped code the new-flow flag is set once in init_task and never cleared, and deficit_u16 is initialized but never drained or consulted, so the section now says that.
- Sleep-lag credit and the waker-boost chain were documented in five places, but the code says "sleep_lag REMOVED" and CAKE_FLOW_WAKER_BOOST has no setter.
- "TUI still works" in release mode is wrong: the loader force-disables --verbose and --testing in release builds (and warns about it). --starvation is parsed but discarded at its only call site, so it's now marked as unused.
- The data-structure overview listed five struct fields that don't exist (nice_shift, sleep_lag, cached_cpumask, waker_boost, cached_perf), and the schbench link pointed at a repo that doesn't exist (it's masoncl/schbench).

For what it's worth, the author's out-of-tree nightly branch has independently deleted the remnants of most of these mechanisms, describing them as dead code — so this documents shipped behavior, it isn't a divergence from where the scheduler is headed. Whether the half-wired deficit/new-flow logic should be completed or removed instead is a behavioral question, intentionally not answered here; I'll file that separately. Doc-only change.

### DIFF
--- a/scheds/rust/scx_cake/README.md
+++ b/scheds/rust/scx_cake/README.md
@@ -10,7 +10,7 @@
 > - **Zero Global Atomics** — Per-CPU BSS arrays with MESI-guarded writes eliminate bus locking
 > - **3-Gate select_cpu** — prev_cpu idle → performance-ordered scan → kernel fallback → tunnel
 > - **Per-LLC DSQ Sharding** — Eliminates cross-CCD lock contention on multi-chiplet CPUs
-> - **EEVDF-Inspired Weighting** — Virtual runtime with sleep lag credit, nice scaling, and tiered DSQ ordering
+> - **EEVDF-Inspired Weighting** — Virtual runtime with nice scaling and tiered DSQ ordering
 
 ---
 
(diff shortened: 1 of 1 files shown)


RitzDaCat:
Valid! I will state that scx_cake is going through quite a lot of code mutations on my nightly branch so I will need to do a pretty heavy review and fix up the readme to match the current scope and design.

PR #3644, opened 2026-06-11
scx_lavd: make --no-use-em reach the userspace CPU-order builder

kurtislin:
## Problem

`--no-use-em` is documented as "Do not use the energy model in making CPU
preference order decisions", but the flag is only written to BPF rodata and
never reaches `CpuOrder::new()`. The userspace `EnergyModelOptimizer` always
runs when the kernel exposes an energy model, regardless of the flag.

This matters for #3340: the EM table generation enumerates 2^n
performance-domain subsets (one PD per CPU on the affected Intel HX hybrids,
e.g. ~34GB peak memory on a 24-CPU Core Ultra 9 275HX), and users in that
thread report `--no-use-em` does not avoid the blow-up
("Tested scx_lavd --performance --no-use-em ... It still reproduces").

## Fix

Plumb the (already normalized) flag into `CpuOrderCtx` and skip
`EnergyModel` construction when set, so every downstream consumer (optimizer
table, `pd_adx` assignment, `has_energy_model`) takes the exact same path as
on a machine without an energy model. That is also the only PCO table layout
the BPF side expects when `no_use_em` is set: its branches assume the
two-state fallback table, while the loader previously kept feeding it the
EM-derived one.

Also print the "Energy model won't be used" message whenever `no_use_em` is
in effect, so users can confirm the flag took effect.

## Scope

This restores the documented bypass and gives #3340 users an immediate
mitigation. It is not the root fix for the 2^n enumeration itself — that is
addressed by the algorithmic rewrite in #3548.


multics69 (approved):
Thank you, @kurtislin , for the patch! Looks good to me.

PR #3645, opened 2026-06-11
scx_cosmos: Bump up version

arighi:
Since scx_cosmos is in a pretty good state now, let's cut a new version so that we can use this as a baseline for future changes.


sirlucjan (approved):
Should we adjust scx_loader?

arighi:
I think scx_loader is already aligned with the latest scx_cosmos

hodgesds (approved):

PR #3647, opened 2026-06-12
scx_cosmos: Handle offline CPUs when reading per-CPU stats

arighi:
read_per_cpu_cpu_times() assumes that `/proc/stat` contains a contiguous sequence of CPU entries starting from `cpu0`. However, when a CPU is offlined, its corresponding `cpuN` line is removed from `/proc/stat`, creating gaps in the sequence.

The parser currently aborts on the first gap (`cpu_id != result.len()`), returning fewer entries than `nr_cpus` and causing the scheduler to fail during initialization with:
```
  thread 'main' panicked at scheds/rust/scx_cosmos/src/main.rs:1254:31:
  Failed to read initial per-CPU stats
```
This issue can be reproduced by offlining a subset of CPUs before starting the scheduler:
```
  for c in $(seq 2 11); do echo 0 > /sys/devices/system/cpu/cpu$c/online; done
```
Fix this by storing statistics in a vector indexed by CPU ID, using `None` for offline CPUs and tolerating non-contiguous `cpuN` entries.

Reported-by: Will C. <wuz352@hotmail.com>

### DIFF
--- a/scheds/rust/scx_cosmos/src/main.rs
+++ b/scheds/rust/scx_cosmos/src/main.rs
@@ -1194,12 +1194,13 @@ impl<'a> Scheduler<'a> {
         }
     }
 
-    /// Read per-CPU times from /proc/stat (lines "cpu0", "cpu1", ...).
-    /// Returns one CpuTimes per CPU, in order.
-    fn read_per_cpu_cpu_times(nr_cpus: usize) -> Option<Vec<CpuTimes>> {
-        let file = File::open("/proc/stat").ok()?;
-        let reader = BufReader::new(file);
-        let mut result = Vec::with_capacity(nr_cpus);
+    /// Parse per-CPU times from /proc/stat (lines "cpu0", "cpu1", ...).
+    /// Returns entries indexed by CPU id. Offline CPUs may be absent.
+    fn parse_per_cpu_cpu_times<R: BufRead>(
+        reader: R,
+        nr_cpus: usize,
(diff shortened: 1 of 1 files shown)


sirlucjan (approved):

scheds/rust/scx_cosmos/src/main.rs
```diff
@@ -1228,19 +1229,22 @@ impl<'a> Scheduler<'a> {
...
                 .sum();
-            result.push(CpuTimes { user, nice, total });
-            if result.len() >= nr_cpus {
-                break;
-            }
+            result[cpu_id] = Some(CpuTimes { user, nice, total });
         }
 
-        if result.len() == nr_cpus {
+        if result.iter().any(Option::is_some) {
```
hodgesds: might be able to simplify this?
```
Some(result).filter(|v| v.iter().any(Option::is_some))
```

hodgesds (approved):

scheds/rust/scx_cosmos/src/main.rs
arighi: @hodgesds true, but IMHO it's not that great in terms of readability. How about:
```
result.iter().any(Option::is_some).then_some(result)
```

I think this reads more clearly as: "if any CPU exists, return Some(result)". This seems more concise and idiomatic. What do you think?

hodgesds: Yeah, that seems better.

PR #3648, opened 2026-06-12
scx_mitosis: Fixes for LLC awareness

kkdwvd:
Removes the work stealing flag since separation makes little semantic sense. Fix a task stall when tasks end up in DSQs of LLCs for cells with no CPUs in said LLC, and a stale task cpumask causing intersection mismatch with tagged/chosen LLC. Final two patches rework to remove extra state in the scheduler around handing task to LLC association. The orphan DSQ handling needs to be replaced by proper interlocked draining logic, but is WIP. Much of these changes are a prerequisite for changes sitting in mitosis-next branch.


kkdwvd:
Added the interlocked orphan LLC draining fix as well. PTAL.

likewhatevs (approved):

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
tommy-u, line 1: This is good, we don't intend to run anywhere with LLC and without stealing.
Nit: I don't think it's a correctness problem to run with LLC but without stealing, but it's certainly a performance issue.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -253,15 +264,7 @@ static inline s32 try_stealing_work(u32 cell, s32 local_llc)
...
-		/*
-		 * Skip if the cell doesn't have CPUs in this LLC.
-		 * Note: rechecking cell_ptr for verifier.
-		 * This is racy with try_stealing_this_task, but we don't care -
-		 * if the LLC actually doesn't have CPUs come steal time,
-		 * we will fail the steal and continue to the next LLC.
-		 */
-		if (cell_ptr && READ_ONCE(cell_ptr->llcs[candidate_llc].cpu_cnt) == 0)
-			continue;
+		u32 candidate_cpu_cnt = READ_ONCE(cell_ptr->llcs[candidate_llc].cpu_cnt);
```
tommy-u: We're going to drop cpu_cnt later in this stack. Could do that first, but this isn't a priority to me

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1739,9 +1790,21 @@ static int init_task_impl(struct task_struct *p, struct cgroup *cgrp)
 	}
 
 	/* Initialize LLC assignment fields */
-	if (enable_llc_awareness)
+	if (enable_llc_awareness) {
 		init_task_llc(tctx);
 
+		llc_cpumask = bpf_cpumask_create();
+		if (!llc_cpumask)
```
tommy-u: This should scx_bpf_error, right?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1422,7 +1468,11 @@ void BPF_STRUCT_OPS(mitosis_stopping, struct task_struct *p, bool runnable)
 	 * domain.
 	 */
 	if (!tctx->borrowed && tctx->vtime_charge_cell == cidx) {
-		advance_cell_llc_vtime(cell, tctx, p->scx.dsq_vtime);
+		u32 llc_idx = enable_llc_awareness && llc_is_valid(cctx->llc) &&
```
tommy-u: Why use FAKE_FLAT_CELL_LLC if llc is not valid? Isn't this an error case?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -290,49 +273,76 @@ static inline s32 try_stealing_work(u32 cell, s32 local_llc)
...
 		return -ENOENT;
 	}
 
-	const struct cpumask *llc_mask = NULL;
-
-	// Let's get a new LLC for this task
-	s32 new_llc = pick_llc_for_task(tctx->cell);
-	if (new_llc < 0)
+	if (!llc_is_valid(new_llc) || new_llc >= nr_llc || new_llc >= MAX_LLCS)
 		return -EINVAL;
```
tommy-u, line 470: The convention we're following for error handling is to scx_bpf_error once as early as we can detect it, and then to return an error code all the way up the stack. This should help us see errors right where they happen instead of on a caller stack frame. That would imply we should error here and then return.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1342,14 +1384,18 @@ void BPF_STRUCT_OPS(mitosis_running, struct task_struct *p)
...
+	struct cell *cell;
 
 	if (!(cctx = lookup_cpu_ctx(-1)) || !(tctx = lookup_task_ctx(p)))
 		return;
 
-	/* Handle cross-LLC task retag (LLC-aware mode only) */
-	if (enable_llc_awareness) {
-		if (maybe_retag_stolen_task(p, tctx, cctx) < 0)
+	if (enable_llc_awareness && tctx->all_cell_cpus_allowed) {
+		if (maybe_update_task_llc(p, tctx, scx_bpf_task_cpu(p)) < 0)
```
tommy-u, line 1405: What is this value for a task that's never run before? I think we're good if it's < 0.

I'm just concerned about initializing the schedulers with massively overloaded queues if it looks like a valid cpu.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
```diff
@@ -62,6 +62,8 @@ struct task_ctx {
 	/* cpumask is the set of valid cpus this task can schedule on */
 	/* (task's cpumask and-ed with its cell cpumask) */
 	struct bpf_cpumask __kptr *cpumask;
+	/* Cached cpumask & LLC mask */
```
dschatzberg: Can you just make this clearer? I don't know what "Cached cpumask" means. My understanding is that `llc_cpumask` is `cpumask and-ed with its LLC's cpumask`. If that's correct, just make that explicit.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
```diff
@@ -86,11 +88,10 @@ struct task_ctx {
 	u64 cgid;
 	/* Which LLC this task is assigned to */
 	s32 llc;
+	/* LLC id for llc_cpumask */
```
dschatzberg: Also here - this is actually a cache key for `llc_cpumask`? Basically telling us we need to recompute `llc_cpumask` if this doesn't == `llc`?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -386,8 +393,16 @@ static inline int update_task_cpumask(struct task_struct *p, struct task_ctx *tc
 	 */
 
 	/* Per-CPU pinned path */
-	if (!tctx->all_cell_cpus_allowed) {
+	if (!all_cell_cpus_allowed) {
+		if (enable_llc_awareness) {
+			tctx->llc = LLC_INVALID;
+			invalidate_task_llc_cpumask(tctx);
```
dschatzberg: You already invalidate earlier in this function

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -396,13 +411,18 @@ static inline int update_task_cpumask(struct task_struct *p, struct task_ctx *tc
 			return -EINVAL;
 
 		p->scx.dsq_vtime = READ_ONCE(cpu_ctx->vtime_now);
+		/* Publish false after installing the CPU DSQ. */
```
dschatzberg: This comment makes it sound like the ordering is important, but it's not, right? This is all per-task data and therefore this logic doesn't run concurrently with anything else touching the per-task data?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -396,13 +411,18 @@ static inline int update_task_cpumask(struct task_struct *p, struct task_ctx *tc
...
+		tctx->all_cell_cpus_allowed = false;
 		return 0;
 	}
 
-	/* Cell-wide path */
-	/* LLC aware version */
+	/* Publish true after assigning a cell DSQ. */
```
dschatzberg: Same here

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
dschatzberg, line 1405: I think I understand it, but it's probably worth a comment explaining why this should happen here in running()

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -36,20 +35,99 @@ static inline bool llc_is_valid(u32 llc_id)
...
 
+static inline void invalidate_task_llc_cpumask(struct task_ctx *tctx)
+{
+	tctx->llc_cpumask_id = LLC_INVALID;
+}
+
+static inline s32 llc_from_cpu(s32 cpu)
+{
+	if (cpu < 0 || cpu >= nr_possible_cpus || cpu >= MAX_CPUS)
+		return LLC_INVALID;
```
dschatzberg, line 62: There are a bunch of places where we do this kind of bounds checking/verifier-required checking but no subsequent scx_bpf_error in the failure case. Can you make sure those are covered? Otherwise I worry we will fail silently

dschatzberg (commented):
Working through this commit-by-commit. I'll post per-commit reviews as I go

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -920,6 +920,9 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
 
 	scx_bpf_dsq_insert_vtime(p, tctx->dsq.raw, slice_ns, vtime, enq_flags);
 
+	if (enable_llc_awareness && tctx->all_cell_cpus_allowed && llc_is_valid(tctx->llc))
+		maybe_enable_cell_llc_drain(tctx->cell, (u32)tctx->llc);
```
dschatzberg: A comment here explaining the ordering is worthwhile - you're doing this after enqueueing in case you raced with reconfiguration which left this LLC orphaned.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -140,42 +146,163 @@ static void zero_cell_vtimes(struct cell *cell)
...
+		}
+	}
+	WRITE_ONCE(cell->llcs_with_cpus, llcs_with_cpus);
+
+	/*
+	 * Pair the post-publication DSQ scan with enqueue's post-insert cached
+	 * no-CPU check. Either this path sees already queued work, or a racing
+	 * enqueue sees the CPU-less LLC and enables draining itself.
+	 */
+	__sync_fetch_and_add(&cell->llcs_to_drain, 0);
```
dschatzberg: This is just a memory barrier? I've not seen this pattern before - does LLVM not handle `__sync_synchronize()` correctly?

Also, is it not possible to use atomic writes/reads on `cell->llcs_with_cpus` - that's really what you're trying to synchronize on, yeah?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -184,25 +311,86 @@ static inline s32 try_stealing_work(u32 cell, s32 local_llc)
...
+		bit = 1LLU << candidate_llc;
+		if (!(drain_mask & bit))
 			continue;
 
-		const struct cpumask *candidate_mask = lookup_llc_cpumask(candidate_llc);
-		if (!candidate_mask)
-			return -EINVAL;
+		dsq_id_t candidate_dsq = get_cell_llc_dsq_id(cell_id, candidate_llc);
+
+		if (cell_llc_has_cpus(cell, candidate_llc))
```
dschatzberg: It's not obvious why you need this check - I think it's because of the later comment that we can spuriously enable the drain bit? Maybe just another comment here would be useful

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -184,25 +311,86 @@ static inline s32 try_stealing_work(u32 cell, s32 local_llc)
...
+		 * or the recheck below sees the new depth and re-enables it.
 		 */
-		bool orphaned_llc = !bpf_cpumask_intersects(cell_mask, candidate_mask);
+		nr_queued = scx_bpf_dsq_nr_queued(candidate_dsq.raw);
+		if (nr_queued <= 1) {
+			cell_llc_drain_disable(cell, candidate_llc);
+			disabled = true;
+		}
+
+		consumed = scx_bpf_dsq_move_to_local(candidate_dsq.raw, 0);
```
dschatzberg, line 389: I'm really worried about a case where the local CPU is not eligible to run the candidate task. I think this can happen any time we reconfigure a cell between disjoint sets of CPUs and a task is cpu-pinned. E.g. If we have the following:

```
LLCs: {0,1}, {2,3}
Cell: {0,1}
Task cpumask {0,1}
```

`all_cpus_allowed == true` so it goes on the LLC cpumask. Then reconfiguration happens and now `Cell: {2,3}`. We've orphaned the other cell LLC, but we'll never be able to grab this task off the queue (in fact it would cause head-of-line blocking too, so we couldn't grab other tasks off the queue).

I wonder if we need to identify this case and re-enqueue the task on some per-CPU queue?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
dschatzberg: Also, can you disable the drain bit for the candidate_llc here to prevent further checks?

dschatzberg (commented):
Mostly smaller comments but one bigger topic on the actual drain mechanics and pinned tasks.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
kkdwvd: Yeah, I meant Cached 'cpumask & LLC mask'. Will reword.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
kkdwvd: Will reword.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: Ack.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: Yeah, it's less about ordering, more about the branch, but I see why it's confusing, will drop.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: Ack.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: Will do.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: Will audit and cover these cases.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: I will add some comments on the ordering.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: I don't think full barriers have corresponding instructions in BPF, so we have to use a fetch_and_add substitute. That said, this particular variant might lead to more cache line bouncing than necessary, I will switch to a local stack variable.

dschatzberg (changes requested):
High-level this all looks good, but let's resolve the drain mechanics discussion before landing

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: We cannot disable it if cell_llc_has_cpus() is true. In parallel, between the check and disabling, we can have enqueue race to queue a task on the said LLC, refresh coming in and making it lose CPUs and enabling drain, and then us disabling it here. Given it's a very short tight race, but it will happen given a high rate of churn.

Disabling needs to always be coupled with checks on whether tasks are queued on the DSQ or not. Either we see tasks queued, other side trying to do that sees disable.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: Yes, this is real, along with the case where a cell loses all CPUs so no one drains anything. I agree we should probably handle this, though for the pinned cases I don't _think_ it's a problem right away. I can work on a follow up commit on top of the stack.

It might make sense to do a local dispatch to an eligible CPU from this path for such a task. We might not want to do it repeatedly lest we run out of the dispatch buffer list. One task consumption or move per iteration.

The case of losing complete CPUs is harder, in that case we likely should move in-flight tasks once the new cpumask has been published, synchronize against end of all-inflight enqueue callbacks somehow (maybe per-cpu seq across all CPUs changing), then drain all stuck tasks out to CPUs eligible to run them.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: Just to add on this, continuing when cell_llc_has_cpus() means the drain bit got disabled in parallel and the cell gained CPUs. This is racy indeed, but if it loses CPUs once again, we will see that drain bit and reenter draining for this LLC.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
dschatzberg: I'm not so worried about losing complete CPUs right now as I don't think that can happen in our setup. But you're right to flag that. I'm ok with followup PR on top to address this.

dschatzberg (approved):
Approving given the latest discussion

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -1348,6 +1358,10 @@ static int update_timer_cb(void *map, int *key, struct bpf_timer *timer)
 		scx_bpf_error("failed to publish root cpumask");
 		return 0;
 	}
+	if (refresh_cell_llc_draining(0)) {
```
tommy-u: Use ROOT_CELL_ID

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -162,6 +163,16 @@ static inline void cell_llc_drain_disable(struct cell *cell, u32 llc)
 	__sync_and_and_fetch(&cell->llcs_to_drain, ~(1LLU << llc));
 }
 
+static inline void cell_llc_nr_queued_inc(struct cell *cell, u32 llc)
```
tommy-u, line 173: From the commit message: scx_mitosis: Track LLC DSQ queue depth for draining

This is specifically working around a kernel bug, right? Please explicitly name it if so. You wouldn't have to do this tracking if it weren't for the bug, right?

tommy-u (approved):
Your work testing this convinces me this is very solid. Thanks for the awesome work, you've improved a ton of my janky code and left Mitosis in a much better place.

A few overall notes:
- The commit messages are hard for me to understand because they are so dense. 
- scx_bpf_errors need to happen as early as possible with error codes propagated up the stack.
- These commits seem to come more in the order that you did the work than in a refactored and flattened set of orthogonal components. E.g. steal_work_result being introduced then removed 

Let's sort out the errors and land, this is great work we should be building on.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: Ack.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: Ack.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: Ack.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
kkdwvd: Ack.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: I will let it be to reduce churn.

kkdwvd:
Landing. If there are more fixes to make (in this stack or otherwise), will open follow up PRs.

PR #3656, opened 2026-06-16
scx_mitosis: Move vtime update near clamp check

kkdwvd:
We may update the task vtime when we update its designated LLC in maybe_update_task_llc(). However, we use a stale vtime recorded before it to do time_before() and time_after() checks to ensure it's not nonsensical. Move the assignment right before the checks to ensure we don't hit errors in time_after() due to arbitrary and weird results because of the stale vtime.

Remove extra assignment to vtime since we can read it right before the
actual checks for the up-to-date value.


tommy-u (approved):
Good catch

PR #3668, opened 2026-06-23
scx_mitosis: Drain ineligible tasks due to affinity restrictions

kkdwvd:
We have cases where we queued tasks in cell DSQs with specific affinity masks that matched cpumask of the cell such that all_cell_cpus_allowed was true. Later, these cells may undergo rebalancing and lose CPUs on LLCs where such tasks are queued. Attempts to drain such tasks will continually fail due to affinity restrictions wrt CPUs trying to drain such tasks.

In such cases, when a scx_bpf_dsq_move_to_local() consumption fails, we dispatch the task to an eligible per-CPU DSQ. In case the task undergoes affinity mask changes, it will undergo dequeue()/enqueue() cycle again.

When such a task starts running, set all_cell_cpus_allowed=false for it to avoid updating its LLC, and invalidate the current selection. The call to maybe_refresh_cell() in the enqueue path will eventually fix any drift from the intended state of the task.

The script reproduces the failure in absence of the fix, and verifies non-zero drain_affn_cnt in presence of it.


scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -371,6 +377,65 @@ static inline s32 try_draining_work(u32 cell_id, s32 local_llc)
...
+		 * candidate DSQ is not eligible to move to the local DSQ of
+		 * this CPU due to affinity restrictions. This can happen when
+		 * a cell loses CPUs on an LLC where it previously ended up
+		 * queuing tasks into the LLC DSQs since all cell CPUs were
+		 * allowed for it.
+		 */
+		if (unlikely(!consumed && pending)) {
+			struct task_struct *p;
+
+			bpf_for_each(scx_dsq, p, candidate_dsq.raw, 0) {
```
tommy-u, line 407: Ok, I like the shape of this but I want a bit more info. Is your thinking here:

1) We know for certain that if we're here it's because of an affinity violation where the task has it's affinity restricted (after getting placed in the cell-llc DSQ) so this CPU (and perhaps all others in the cell) will never be able to drain the task, so we need to remap it to a CPU pinned DSQ now.

2) We can't be certain why we're here, but an affinity violation is a good guess and a pretty safe way to un-wedge this task and it will get corrected later on a select/enqueue if that was a false positive.

3) Probably something else

I think we can't know 1) for certain because each piece of info we're acting on (consumed, pending, the task we're iterating) is racy and unreliable.

My problem with 2) is that if have a false positive and demote the task from all_cell_cpus_allowed to not, it might be a very long time before we have a reconfiguration that heals the restricted placement via select_cpu() / enqueue() -> maybe_refresh_cell() -> refresh_task_cell() -> update_task_cpumask()

I'd treat it more agnostically:
- If the head can run on the dispatching CPU, do nothing, some other CPU will get it on dispatch.
- Only if the head cannot run there, move it to an allowed CPU’s per-CPU DSQ.
-- Prefer a CPU in cell_cpumask & p->cpus_ptr; fall back to p->cpus_ptr.
- Mark placement for one-shot refresh so any optimistic demotion is recomputed on the next select/enqueue, conditionally calling update_task_cpumask

Hope some part of that makes sense. I feel like we could benefit from some degree of locking so we could reason more simply about invariants.

tommy-u (changes requested):
Just one point I'm stuck on. I bet you can explain what I'm missing.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
tommy-u: Oh I reread and I'm pretty sure you're thinking 2). I think this won't get fixed for second(s) where false positive tasks will contend with true pinned cpus, so maybe we should force an immediate update_task_cpumask. (immediate meaning on the next select / enqueue)

Please make an obvious comment in the code that spells this out for simple people --like reviewers

kkdwvd: It is 1, not 2. All information is not racy, only the pending items is. It would make sense to sample pending after we set consumed to make it more accurate, I'll push that change, since we otherwise enter this branch when we have last racing LLC DSQ dequeue.

It is likely this happens when some CPU can drain an affinitized task but not another in the same cell, and they race. But this will happen upon reconfiguration such that those tasks are no longer eligible for LLC DSQs. It should be a rare occurrence.

> My problem with 2) is that if have a false positive and demote the task from all_cell_cpus_allowed to not, it might be a very long time before we have a reconfiguration that heals the restricted placement via select_cpu() / enqueue() -> maybe_refresh_cell() -> refresh_task_cell() -> update_task_cpumask().

Why would it be a problem? After this dispatch, the task will get to run (or be placed in per-CPU DSQ waiting). Next time it gets enqueued, we will have a correct picture of its all_cell_cpus_allowed state and act accordingly. That may happen much later, but that doesn't have an implication with respect to correctness.

tommy-u: The problem I see with with 2) is that all_cell_cpus_allowed will not be fixed on the next select/enqueue in general. It's only when the configuration sequence changes, which happens O(seconds), which concerns me. During this time, the task will be pinned to a single cpu and any real pinned tasks will have to fight with it for runtime while other cell cpus potentially sit idle. We have real tasks doing single cpu pinning in the wild unfortunately. Setting a flag would allow us to fix this up ~100x faster assuming it wakes up quickly (on the order of a slice + wakeup).

I'll get back to 1) in a bit :)

kkdwvd: I think my assumption is that for tasks in the queue being drained, the applied configuration sequence should already be out of sync. We can add a flag to close the theoretical concern, but I don't see how we will end up with a task that won't go through refresh. If drain is enabled, the enqueue was racing while it operated on a stale cpumask and ended up having to enable drain to compensate for the CPU loss in the LLC. In that case it missed a pending refresh.

What can end up happening is that this ineligible task was taken by some other CPU, and the task we're operating on is really eligible. In that case we might end up marking it as all_cell_cpus_allowed=false, but given the above, it should still cycle through a refresh.

dschatzberg: This discussion confuses me a bit. The check can fire for reasons beyond just a task having out-of-date configuration as far as I understand. For example, if we race on dequeueing the DSQ on multiple cores, we could see !consumed because this core lost the race. Then the subsequent check on nr_queued might also race and be non-zero.

In that case, we might have a correctly configured task which we assign to a per-cpu DSQ which as Tommy points out, might not be fixed for quite a while

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -371,6 +377,65 @@ static inline s32 try_draining_work(u32 cell_id, s32 local_llc)
...
+				/*
+				 * Obviate any LLC updates during running(),
+				 * next cell refresh on enqueue() will recompute
+				 * these based on the current cell state.
+				 */
+				tctx->all_cell_cpus_allowed = false;
+				invalidate_task_llc(tctx);
+
+				scx_bpf_dsq_move_set_vtime(BPF_FOR_EACH_ITER, basis_vtime);
+				consumed = scx_bpf_dsq_move_vtime(BPF_FOR_EACH_ITER, p, cpu_dsq.raw,
```
dschatzberg, line 435: This can fail if some other CPU which could dispatch the task consumed it before we hit here, but we've already mutated tctx. Is there some way we can move this off the dsq it's on first, and then do the rest of the work here to make this non-racy?

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -371,6 +377,65 @@ static inline s32 try_draining_work(u32 cell_id, s32 local_llc)
...
+								  0);
+				if (consumed) {
+					cstat_inc(CSTAT_DRAIN_AFFN_CNT, cell_id, target_cctx);
+					scx_bpf_kick_cpu(cpu, SCX_KICK_IDLE);
+				}
+				break;
```
dschatzberg, line 456: What are the semantics of the return value of this function? Should it be a positive non-zero if we dispatched to local or if we drained an llc? I think this distinction didn't matter until this commit but we will just idle the CPU right away now even if we didn't dispatch to local.

1) Can we document the return value expectations of this function
2) I *think* this will idle a CPU incorrectly as currently implemented - we probably want to fix that.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -371,6 +377,65 @@ static inline s32 try_draining_work(u32 cell_id, s32 local_llc)
...
+				 * these based on the current cell state.
+				 */
+				tctx->all_cell_cpus_allowed = false;
+				invalidate_task_llc(tctx);
+
+				scx_bpf_dsq_move_set_vtime(BPF_FOR_EACH_ITER, basis_vtime);
+				consumed = scx_bpf_dsq_move_vtime(BPF_FOR_EACH_ITER, p, cpu_dsq.raw,
+								  0);
+				if (consumed) {
+					cstat_inc(CSTAT_DRAIN_AFFN_CNT, cell_id, target_cctx);
```
dschatzberg: Hmm, I don't think we ever do this cstat_inc on a remote cpu context. These counters aren't atomic so this is data-racy. If you really want to track this you probably want to mark the task ctx and then bump the counter on the remote cpu's dispatch()

dschatzberg (changes requested):

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: The first part is true, but I think one point of confusion is that we don't have new tasks entering the DSQ anymore where their configuration seq is up to date. Existing queued ones are tasks that came in before the configuration was bumped already, and before the DSQ marked eligible for draining. As such, everything in this queue should be at least one configuration seq behind the current seq. In case the DSQ gains CPUs again, we will have new tasks entering, but at that point also disable drain, and, still, for older queued tasks, be behind in terms of configuration seq (now by > 1 seq).

kkdwvd: That said, just for peace of mind, I will add an extra flag to force a refresh for tasks. It doesn't cost much in terms of overhead but it allows us to avoid cases where the reasoning supplied above is wrong (I don't think so, but this is all hard to reason about in general) and thus we end up invalidating something that won't go through refresh until we drift again in the configuration sequence.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: I will take a look.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: Ack.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: Ack, this was sloppy on my part, I will address it properly. In the test script I cared more about it being non-zero, but we should do this properly.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
dschatzberg: > but I think one point of confusion is that we don't have new tasks entering the DSQ anymore where their configuration seq is up to date

Ah, I had not made this connection but I think it holds true. In this case I'm happy if you can explain in a comment and keep the logic as is.

tommy-u: Noting #3678 showed this path can get hit when we are in-between configuration updates. The concern wasn't theoretical.

kkdwvd: I think it was a month ago, so I lost most of the context by now.


If memory serves right, my theory (which I encouraged you to investigate) was that the check itself could also race, given the actual bump of `applied_configuration_seq` happens after a fair amount logic executes between it (Phase 5) and the Phase 2 refreshing the drain bit. Thus it is probable your current approach surfaces such false positives.

If we do prove such a case exists, we should apply the suggestion above, where we add a new flag to force a refresh. If you don't have the bandwidth to investigate it but still feel we should close this, I am fine with adding a flag to close the concern.

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
```diff
@@ -371,6 +387,76 @@ static inline s32 try_draining_work(u32 cell_id, s32 local_llc)
...
+				dsq_id_t cpu_dsq;
+				u64 basis_vtime;
+				u32 cpu;
+
+				tctx = lookup_task_ctx(p);
+				if (!tctx) {
+					scx_bpf_error(
+						"lookup_task_ctx() failed in try_draining_work()");
+					break;
+				}
```
tommy-u, line 419: Two minor issues here.
1) lookup_task_ctx already internally errors
2) don't break, return error up the chain.

tommy-u: 2) Is what Dan and I agreed on to stop execution ASAP after an error. 

AI suggests to also consider adding these immediate-return sites:

  - scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h:415: remove the second error and return -EINVAL.
  - scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h:422: report invalid CPU, then return -EINVAL.
  - scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h:426: return -EINVAL.
  - scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h:430: return -EINVAL.
  - scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c:920: return from mitosis_dispatch() for fatal negative results, while
    allowing -ENOENT and CONTINUE_DISPATCH to continue.

tommy-u (commented):
This looks great, you addressed my remaining concern about returning consumed when the task was really enqueued on another cpu dsq.

Please make sure errors are reported once and errors are returned immediately.

Squash and merge when ready.

tommy-u (approved):
Oops I meant to approve before. Just address those error returns and squash :)

scheds/rust/scx_mitosis/src/bpf/llc_aware.bpf.h
kkdwvd: I will fix 1, let me follow up for 2 separately once this lands.

dschatzberg (approved):
As we discussed, the tctx mutation is now racy on the other side (can be mutated while the remote cpu is dispatching it). The proper fix is probably to have a per-cpu private DSQ where tasks can be moved temporarily to mutate. This is fine for now though

PR #3672, opened 2026-06-26
common.bpf.h: Make scx_bpf_cpu_rq() weak

arighi:
Recent kernels (for-7.3) removed the deprecated scx_bpf_cpu_rq() kfunc. The helper added by the compatibility layer still references it as the fallback for kernels without scx_bpf_cpu_curr().

Because the declaration was a strong ksym extern, libbpf required the symbol to exist even when bpf_ksym_exists(scx_bpf_cpu_curr) would make the fallback dead on new kernels. This made veristat fail before the verifier ran with:

  libbpf: extern (func ksym) 'scx_bpf_cpu_rq': not found

Mark scx_bpf_cpu_rq() weak so kernels that removed it can load the objects and resolve the fallback to NULL, while older kernels that still export the kfunc can continue using the fallback.


arighi:
This should fix the veristat errors in #3671

sirlucjan (approved):

likewhatevs (approved):

PR #3673, opened 2026-06-26
lib, scx_lavd: fix runnable-task stalls under cpu.max

multics69:
This series fixes runnable-task-stall watchdog timeouts in scx_lavd's
cgroup bandwidth (cpu.max) path, seen under heavy cgroup churn plus a
fork/exit storm. Two independent root causes were found: an allocator
double-allocation (patch 1), which is what stops the reproducer, and a
set of non-atomic backlog-queue relocations (patches 2-7) that corrupt
the queue during cgroup teardown and task migration.

Allocator double-allocation
---------------------------

scx_alloc_internal() reserves a distinct radix-tree index under
alloc_lock, then drops the lock before carving the task's data element
from a shared slab cursor (pool->idx).  ops.init_task is sleepable and
runs concurrently on many CPUs, so under a fork storm two allocations
race on that cursor: both read the same pool->idx and compute the same
element pointer.  The two distinct indices end up bound to one element
-- two tasks share a single task context.  One task's exit then zeroes
the context the other still uses (its ->atq, cgrp_id, and lavd fields
read back zero), orphaning the live task.

The fix carves the element under alloc_lock, dropping the lock only for
the sleepable bpf_arena_alloc_pages() refill and re-checking the cursor
before installing the new slab -- the same lock-drop-allocate-recheck
pattern scx_alloc_stack() already uses for the page stack.

- [1/8] lib/sdt_alloc: carve pool elements under alloc_lock to prevent double allocation

Atomic BTQ relocation
---------------------

A throttled task's backlog-queue (BTQ) membership is the single pointer
taskc->atq, and the BTQ links the rbnode embedded in the task context.
The invariant is that the node must be unlinked from every BTQ before
scx_task_free() zeroes the context it lives in.

Relocating a task with cancel-then-insert briefly leaves ->atq == NULL.
A lockless reader -- a concurrent task exit, or a cgroup teardown
draining the BTQ -- observes that NULL, concludes the task is in no
queue, and skips its unlink; a later free then zeroes a still-linked
node and corrupts the destination rbtree, stranding every task parked
there.

Two move primitives close the window by holding both ATQ locks across
the move, so ->atq transitions directly from source to destination and
is never observed NULL: scx_atq_move_vtime() moves the front task (used
when a dying cgroup's BTQ is drained into root's), and
scx_atq_move_task() moves a specific task (used when a throttled task
migrates between cgroups).  scx_atq_cancel() is made authoritative --
it retries against wherever a concurrent move relocated the task -- and
lavd_exit_task() cancels before freeing, so the node is always unlinked
first.

- [2/8] lib/atq: add scx_atq_move_vtime() to atomically move a task between ATQs
- [3/8] lib/cgroup_bw: move dying-cgroup tasks to the root BTQ atomically
- [4/8] lib/atq: retry scx_atq_cancel() until the task is unlinked
- [5/8] scx_lavd: unlink a throttled task from its BTQ before freeing it
- [6/8] lib/atq: add scx_atq_move_task() to atomically move a specific task between ATQs
- [7/8] lib/cgroup_bw: relocate a migrating throttled task atomically

Benign teardown log
-------------------

The reenqueue path walks a cgroup's per-LLC contexts off a snapshot of
throttled cgroup ids.  A lookup can miss because the cgroup is being
torn down concurrently after its id was recorded -- expected under
churn, not an error.  Downgrade the message from error to debug,
matching the sibling cgroup-context lookup-failure path.

- [8/8] lib/cgroup_bw: downgrade a benign LLC-lookup failure to debug

### DIFF
--- a/lib/atq.bpf.c
+++ b/lib/atq.bpf.c
@@ -205,6 +205,142 @@ u64 scx_atq_pop(scx_atq_t *atq)
 	return taskc_ptr;
 }
 
+/*
+ * Lock two distinct ATQs, lowest address first, to avoid ABBA deadlock with
+ * a concurrent move in the opposite direction. @a and @b must differ;
+ * scx_atq_move_vtime() rejects @src == @dst before calling this. Keeping the
+ * lock count unconditional (always two) also keeps the verifier's
+ * preempt-disable/enable balance provable.
+ */
+static __always_inline
+int scx_atq_lock2(scx_atq_t *a, scx_atq_t *b)
+{
+	scx_atq_t *first, *second;
+	int ret;
+
+	if ((u64)a < (u64)b) {
+		first = a;
+		second = b;
+	} else {
+		first = b;
+		second = a;
+	}
+
+	if ((ret = scx_atq_lock(first)))
+		return ret;
+	if ((ret = scx_atq_lock(second))) {
+		scx_atq_unlock(first);
+		return ret;
+	}
+
+	return 0;
+}
+
+static __always_inline
+void scx_atq_unlock2(scx_atq_t *a, scx_atq_t *b)
+{
+	scx_atq_t *first, *second;
+
+	if ((u64)a < (u64)b) {
+		first = a;
+		second = b;
+	} else {
+		first = b;
+		second = a;
+	}
+
+	scx_atq_unlock(second);
+	scx_atq_unlock(first);
+}
+
+/*
+ * Atomically move the front (min-key) task of @src into @dst, re-keyed by
+ * @vtime. Both ATQ locks are held across the move and the task's ->atq is
+ * updated directly from @src to @dst, so a lockless reader of ->atq never
+ * observes the transient NULL that a separate scx_atq_pop()+insert() would
+ * expose. Locks are taken lowest-address-first (see scx_atq_lock2()).
+ *
+ * On insertion failure the task is left in @src; it is never lost.
+ *
+ * Return the moved task (scx_task_common *) as u64, or (u64)NULL if @src is
+ * empty or the move failed.
+ */
+__hidden
+u64 scx_atq_move_vtime(scx_atq_t *src, scx_atq_t *dst, u64 vtime)
+{
+	scx_task_common *taskc;
+	rbnode_t *node;
+	u64 src_key, taskc_ptr;
+	int ret;
+
+	/*
+	 * Relocation is between distinct ATQs; @src == @dst would self-deadlock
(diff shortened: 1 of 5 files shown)


etsal:
Hi Changwoo, thank you for looking into this. I will take a second look tomorrow, to sum up for now:

a) Patch 1 looks valid to me.
b) The atomic locking patches add significant amounts of complexity. The root problem does not sounds like atomic atq operations, but that we cannot distinguish between a task being moved and a task not in any ATQ. What if we instead used some sentinel value in the ATQ to differentiate between them? We mark the task as being moved, and only complete the move if a concurrent exit has not invalidated it.

multics69:
@etsal -- Thanks for taking a look. Could you elabore a bit more on the sential approach?

etsal:
Picking this up from our offlist chat (the TL;DR being that we need some way of moving tasks between ATQs that doesn't temporarily make it look like the task is not being throttled, and that a state variable would require a busy wait section).

I have been working on seeing how explicit reference tracking for throttle state would look like, and I am starting to believe we should more explicitly model throttling state in cgroup_bw because that's the right level of abstraction for this. The races are coming from two interrelated API issues:

1) We use ATQ membership as a proxy for throtttling state in cgroup_bw. This seemed the most elegant solution at the time, but it turns out it's not a good proxy since we do moves between ATQs all the time.
2) We use the scx_cgroup_bw_cancel() call in two different contexts, unthrottling the task and removing it from an ATQ to move it to another one.

Both these mismatches are due to using ATQ layer state to represent throttling state. The thing is, we _do_ have a way of directly representing cgroup_bw state, the scx_task_common struct. Imo we should explicitly add state there (reference counting should be enough). I am working on that and will post the result here. If we do this right I think we can also avoid the benign race we have in cgroup_bw.bpf.c, and minimize the ATQs' API surfaces by removing the _unlocked() variants that seem extra difficult to use right and expose internal ATQ state.

I think this also ties well with what @multics69 mentioned on our offlist discussion: ATQs and BTQs seem to somehow diverge in terms of requirements. I'd rephrase slightly but I think I agree: BTQs have atomicity requirements that ATQs don't have. In my opinion, however it's not that we need a more powerful queue data structure but that we need more explicit bookkeeping on top of them.

multics69:
As discussed offline, let me close this PR, then open a separate PR for the allocator fix (https://github.com/sched-ext/scx/pull/3679). We will discuss more about how to handle the atomic task move between BTQs correctly.

PR #3675, opened 2026-06-27
scx_pandemoniumv5.15.0:

wllclngn:
v5.15.0 is rebuilt from the v5.12.0 base -- the last clean floor -- and what makes it v5.15.0 is that the latent bug in that floor was finally found and resolved. v5.12.0's per-CPU kthread and interactive verdict was scattered across six uncoordinated sites that overlapped on the easy on-home wakeup and left the cross-CPU strand with no net, so the base stranded pinned kthreads (ksoftirqd/N, kworker/N) worse than EEVDF. Collapsing all six into ONE judgement -- set once in runnable(), read everywhere else -- clears it: montauk kstrand at 2C, the worst case, reports the strands gone. Everything else is re-added onto the cleaned base one gated, measured piece at a time so each earns its place against the storm and latency numbers before the next lands. ONE-JUDGEMENT PER-CPU KTHREAD UNIFICATION below carries the fix; v5.15.0-CHECKLIST.md carries the full re-add ledger and the original release notes; this records what has landed and validated.

RESET TO THE v5.12.0 BASE -- src/
- The scheduler crate (BPF, the Rust loader and intf.h) resets to the v5.12.0 commit, the clean placer the rest stands on: per-CPU DSQs, one flattened cache_domain, the node_dsq spill, no CCX/Phi machinery yet. The harness under tests/ is kept current so its fixes survive the reset.

IDLE QUIESCENCE ENVELOPE -- src/bpf/main.bpf.c, src/bpf/intf.h, src/adaptive.rs, src/scheduler.rs
- The v5.13.0 envelope re-adds onto the base's existing CoDel oscillator: a decayed disp^2+v^2 energy reservoir parks the oscillator at codel equilibrium when it settles -- pin the target, freeze the velocity integrator, stop the per-tick recompute -- until a rescue event or an equilibrium retune disturbs it. A Schmitt trigger on the energy (park below env_park, release at 2x) with a heartbeat valve. nr_osc_park lands as a stat with adaptive park: telemetry.

INTERACTIVE-WAITING PREEMPT -- src/bpf/main.bpf.c
- The per-CPU tick preempt gated only on pcpu_enqueue_ns -- the per-CPU DSQ waiter -- and was blind to interactive_enqueue_ns, the shared node_dsq waiter. An interactive task spilled to node_dsq under load forced no resident off its core, dispatch never ran, and the task starved behind a batch hog up to the 167ms net: the 116-195ms interactive-stall tail under mixed and long-run load. The v5.11->v5.12 boundary split one waiter signal in two and wired only the per-CPU half.
- The fix restores v5.11.0's global interactive-waiting path keyed on the existing interactive_enqueue_ns, placed before the per-CPU early-return so it is reached: a BATCH resident held past preempt_thresh while a node_dsq interactive waiter is pending is preempted -- dispatch runs and STEP 2 drains the waiter (sojourn_gate_pass already keys on the same signal). BATCH residents only; interactive residents keep their protection.
- Measured: long-run worst 153ms -> 2.6ms on the BPF arm at every width (2/4/8/16C), mixed worst 167ms -> 4ms, no preemption storm at 16 threads.

V7.1 REENQUEUE GUARD REMOVED -- src/bpf/main.bpf.c
- The SCX_ENQ_REENQ short-circuit added for the 7.1 cpu_release rework routed a reenqueued task straight to node_dsq with no kick, on the theory it broke a 7.1.1 reenqueue<->kick livelock. It moved nothing measurable on 7.0.13 and is removed: pandemonium_enqueue falls from the wakeup classify straight into Tier 1 as on the v5.12.0 base. cpu_release reenqueues unconditionally.

SOJOURN UNIFICATION -- src/bpf/main.bpf.c, src/bpf/intf.h, src/tuning.rs, src/adaptive.rs
- codel_thresh_ns renames sojourn_thresh_ns (the batch-DSQ rescue deadline, a Rust-set knob); the pcpu_sojourn_thresh local alias collapses into it.
- codel_starve_ns renames starvation_rescue_ns (the ~100ms last-resort net, KEPT distinct from the ~6ms target band).
- codel_seed_ns renames codel_target_equilibrium_ns (the R_eff x tau equilibrium the oscillator parks at; codel_eq_ns stays the feeding knob, the clamp preserved).
- sojourn_stamp_overflow[] as struct { u64 inter, batch; } folds interactive_enqueue_ns + batch_enqueue_ns onto one cacheline per domain (per-tier x per-domain preserved, zero cells lost); sojourn_stamp_pcpu[] renames pcpu_enqueue_ns (stays separate, 64-byte padded).
- DELETED overflow_sojourn_rescue_ns. The storm fix had the CPU-0 tick set it = codel_target_ns every beat while apply_tau_scaling still seeded it = the equilibrium -- a two-writer conflict that disagreed whenever the oscillator drove the target off equilibrium. Resolved to the LIVE target (every other sojourn deadline in dispatch keys off codel_target_ns), dropped the vestigial equilibrium writer; the reads read codel_target_ns directly.
- Effect: behavior-preserving. montauk sojourn-pressure A/B -- 16C p99 1018us against the v5.14.0 baseline ~2994us, held and under, 4/4 widths survived. The redundancy was real and free.

ONE-JUDGEMENT PER-CPU KTHREAD UNIFICATION -- src/bpf/main.bpf.c
- The per-CPU kthread (ksoftirqd/N, kworker/N) latency verdict was scattered across six uncoordinated sites -- the procdb tier vote, the runnable behavioral classify, a three-branch override cascade, an enqueue fast path, the enqueue affinity skip and the tick preempt -- each added to patch a different strand, overlapping on the on-home wakeup and leaving the cross-CPU wakeup with no net. The result stranded pinned kthreads worse than EEVDF: pcpu-burst 5 strands to 58ms, sojourn-pressure 2 to 56ms, fork-thread 26 to 50ms where EEVDF stranded none.
- The collapse: runnable() is the sole decider. The kthread flags (PF_KTHREAD, nr_cpus_allowed, static_prio) are tested once and folded into tctx->pinned_service alongside the tier; enqueue() and tick() become pure reads of that bit. No struct growth -- the bit reuses existing pad.
- enqueue() routes a pinned-service task straight to its home CPU's per-CPU DSQ (drained first by dispatch STEP 0, cache-hot) and PREEMPT-kicks it. A pinned task has no placement choice, so this covers BOTH the on-home and the cross-CPU wakeup -- the old on-home-only fast path left the cross-CPU strand falling to the ~167ms net.
- tick() restores the pinned-waiter backstop keyed on the per-CPU sojourn signal, placed before the LAT_CRITICAL tight-band exemption so a welded ksoftirqd/kworker behind a LAT_CRITICAL resident is rescued -- steal cannot move a pinned task. The separate pcpu_pinned_strand_ns[] stamp array is gone; the one bit carries it.
- Measured: montauk kstrand at 2C, the worst case. ipc, longrun, mixed-ADAPTIVE and longrun-ADAPTIVE report no strands over threshold; mixed-BPF shows one 8.1ms kworker/1:1 held 99% by montauk itself -- the tracer, not the scheduler -- far under the net. dispatch-stall attributes the residual floored wakes to python3 saturation at 2 CPUs, 0% priority inversion and 0% FIFO violation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6358,7 +6358,7 @@ dependencies = [
 
 [[package]]
 name = "scx_pandemonium"
-version = "5.14.0"
+version = "5.15.0"
 dependencies = [
  "anyhow",
  "clap",
--- a/scheds/rust/scx_pandemonium/Cargo.toml
+++ b/scheds/rust/scx_pandemonium/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "scx_pandemonium"
-version = "5.14.0"
+version = "5.15.0"
 edition = "2021"
 description = "A behavioral, adaptive sched_ext scheduler with three-tier classification, L2 affinity, and process learning"
 license = "GPL-2.0-only"
--- a/scheds/rust/scx_pandemonium/README.md
+++ b/scheds/rust/scx_pandemonium/README.md
@@ -10,136 +10,131 @@ PANDEMONIUM is included in the [sched-ext/scx](https://github.com/sched-ext/scx)
 
 ## Performance
 
-12 AMD Zen CPUs (Ryzen 5 3600), kernel 7.0.12.arch1, clang 21. v5.14.0, EEVDF, scx_cosmos, scx_lavd are from a fresh-boot full-field bench-scale (2026-06-17, 3 iterations). All five schedulers completed every cell; no crash or hotplug insta-exit contamination.
+12 AMD Zen CPUs (Ryzen 5 3600), kernel 7.0.13-arch1-1, clang 21. v5.15.0 numbers are from a prism-scale run (2026-06-27, 3 iterations): EEVDF baseline vs PANDEMONIUM (BPF and ADAPTIVE). External schedulers are omitted from this comparison.
 
 ### P99 Wakeup Latency (interactive probe under CPU saturation)
 
-| Cores | EEVDF    | scx_cosmos | scx_lavd  | PANDEMONIUM (BPF) | PANDEMONIUM (ADAPTIVE) |
-|-------|----------|------------|-----------|-------------------|------------------------|
-| 2     | 72,740us | 11,847us   | 11,561us  | 612us             | **463us**              |
-| 4     | 6,020us  | 2,274us    | 5,957us   | 108us             | **66us**               |
-| 8     | 1,305us  | 1,705us    | 3,291us   | **67us**          | **67us**               |
-| 12    | 694us    | 1,902us    | 1,789us   | 69us              | **68us**               |
+| Cores | EEVDF   | PANDEMONIUM (BPF) | PANDEMONIUM (ADAPTIVE) |
+|-------|---------|-------------------|------------------------|
+| 2     | 3,009us | **105us**         | 136us                  |
+| 4     | 1,996us | 1,059us           | **77us**               |
+| 8     | 728us   | **66us**          | 67us                   |
+| 12    | 313us   | **68us**          | **68us**               |
 
 ### Burst P99 (fork/exec storm under CPU saturation)
 
-| Cores | EEVDF    | scx_cosmos | scx_lavd  | PANDEMONIUM (BPF) | PANDEMONIUM (ADAPTIVE) |
-|-------|----------|------------|-----------|-------------------|------------------------|
-| 2     | 5,809us  | 10,336us   | 13,934us  | **266us**         | 924us                  |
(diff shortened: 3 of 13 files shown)


wllclngn:
Will amend veristat issues in a few hours.

sirlucjan:
I noticed some regressions. Scheduler didn't crash but I had some freezes. Also benchmark noticed regressions

kernel_version_comparison_All

PR #3678, opened 2026-07-01
Mitosis llc drain error repo

tommy-u:
Just sharing a draft.

It was an open question if the maybe_refresh_cell following an affinity restricting dsq drain would imply an immediate task cpumask update (and affinity correction).

I added a task specific flag `check_affinity` that I set when the affinity restricted draining task was consumed. Then I checked if the configuration sequence did not detect a change (called by enqueue or select). If it detected no change, that would mean the task affinity would not be immediately corrected. 

The fact that we hit this error provides a counterexample. This is not guaranteed by the current state of our system.

```
24:  scx_bpf_error (src/bpf/mitosis.bpf.c:480: check_affinity task reached maybe_refresh_cell without config refresh: pid=2347406 cell=1 dsq=1000006c)
684:    EXIT: scx_bpf_error (src/bpf/mitosis.bpf.c:480: check_affinity task reached maybe_refresh_cell without config refresh: pid=2347406 cell=1 dsq=1000006c)
```


kkdwvd:
Closing this, feel free to reopen if the comments on false positives turns out to be untrue.

PR #3680, opened 2026-07-02
scx_mitosis: refactor cell manager CPU assignment

tommy-u:
The goal of this functionality preserving refactor is to make the process of assigning CPUs to cells easier to understand, modify, unit test, and maintain. It also validates invariants after all assignment is finished.

This splits the monolithic flow into smaller mostly-pure helpers with specific responsibilities and explicit inputs and outputs.

The new final validation turns implicit assumptions into explicit checks: assignments must cover every CPU exactly once, every cell must receive CPUs, cpuset constraints must be honored, and borrowable masks must be valid.

### Final Assignment Set

- Every known cell has exactly one final assignment.
- Every cell's `assignment.primary` mask is non-empty.

### Primary CPU Ownership

- Every CPU in `all_cpus` appears in exactly one `assignment.primary` mask.
- Every primary CPU is within `all_cpus`.

### Cpuset And Claim Ownership

- Pinned cells' primary CPUs stay within their cpuset.
- Claimed CPUs are assigned only to cell `0` or to one of their claiming cells.

### Borrowable Masks

- Borrowable masks are present iff borrowing is enabled.
- Borrowable CPUs are within `all_cpus`.
- Borrowable CPUs do not overlap that cell's primary CPUs.
- Pinned cells' borrowable CPUs stay within their cpuset.

### DIFF
--- a/scheds/rust/scx_mitosis/src/cell_manager.rs
+++ b/scheds/rust/scx_mitosis/src/cell_manager.rs
@@ -213,6 +213,574 @@ pub struct CpuAssignment {
     pub borrowable: Option<Cpumask>,
 }
 
+#[derive(Debug, Clone)]
+struct CellSpec {
+    cell_id: u32,
+    cpuset: Option<Cpumask>,
+}
+
+#[derive(Debug)]
+struct HoldoutResult {
+    mask: Cpumask,
+    enforced: bool,
+}
+
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


kkdwvd (approved):
This looks ok to me, looks like NFC apart from extra validation (which is probably ok in cost relative to all the other work done).

PR #3683, opened 2026-07-02
[mitosis] Enrich trace output for cells

tommy-u:
Improve cell TRACE output. 

Cells used to print in unstable order, this sorts them by cell number.

Adds CPU count plus current/smoothed utilization, making rebalance/debug runs easier to inspect.

Before:
`CELL[0]: fffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff`

After:
`... CELL[0]: fffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff,ffffffff (316 CPUs, util=2.7%, smoothed=1.8%)`

### DIFF
--- a/scheds/rust/scx_mitosis/src/main.rs
+++ b/scheds/rust/scx_mitosis/src/main.rs
@@ -1160,8 +1160,20 @@ impl<'a> Scheduler<'a> {
         self.collect_demand_metrics(&cpu_ctxs)
             .context("collecting demand metrics")?;
 
-        for (cell_id, cell) in &self.cells {
-            trace!("CELL[{}]: {}", cell_id, cell.cpus);
+        let mut sorted_cells: Vec<_> = self.cells.iter().collect();
+        sorted_cells.sort_by_key(|(cell_id, _)| *cell_id);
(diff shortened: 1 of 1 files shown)


kkdwvd (approved):

likewhatevs:
Adding in a cell.cpus.to_cpulist() to print in there would be nice (for hex challenged folks).

likewhatevs (approved):

scheds/rust/scx_mitosis/src/main.rs
dschatzberg, line 1243: Here we skip updating util_pct for an idle cell. This seems incorrect since it should effect util which is going to reflect in your trace output.

dschatzberg: It also short-circuits from updating smoothed_util

dschatzberg (approved):
This looks good but I think we have a bug in collect_demand_metrics - see my inline comment.

PR #3685, opened 2026-07-04
Initial pull Lunar v1.0

WhitePeace36:
Hi, 

I created a new cpu scheduler. Which i have been working on for quite some while now.

I think it works great and wanted to give other people also the opportunity to use it, so i am creating this PR.

How the scheduler works is in the README below.

To be honest i first created the loader in c++ and this is what i mostly used for development because i am not that into rust and i first tried to search if there would be any possibility that this repo would accept c++ loaders. But i don't think that would be a good idea, so i asked claude to translate the loader from c++ to rust and it works fine. So there looks to be no problem.
Only so that you are aware.


arighi (changes requested):
Please write meaningful commit messages following the style of other commits. I typically stop reviewing as soon as I see commit message that don't say anything. Moreover, there's a binary (scx_lunar_intree.zip), you should remove that. Thanks.

WhitePeace36:
> Please write meaningful commit messages following the style of other commits. I typically stop reviewing as soon as I see commit message that don't say anything. Moreover, there's a binary (scx_lunar_intree.zip), you should remove that. Thanks.

ahh. Sry for that.  and thx. I didn't see that i got a .zip in there. I overlooked it.

WhitePeace36:
i will close it for the time being.

PR #3695, opened 2026-07-08
lib/cgroup_bw: Pack cgroup LLC map key

### LINKED ISSUE #3693: scx_lavd: LLC map lookups fail because of key padding
## Summary

When `scx_lavd` runs with `--enable-cpu-bw`, entries in `cbw_cgrp_llc_map` are not reclaimed as cgroups exit. The map
eventually fills and a subsequent cgroup creation disables the scheduler with:

```text
scx_bpf_error (src/bpf/main.bpf.c:2533: Failed to init a cgroup: -12)
```

This was observed with scx v1.1.2 at commit `b68cb3c829312b43785225bcfd445a3ecefef5c2`.

## Cause

The map key has four bytes of implicit tail padding:

```c
struct cgroup_llc_id {
        u64 cgrp_id;  /* bytes 0..7 */
        int llc_id;   /* bytes 8..11 */
                      /* bytes 12..15 are padding */
};
```

`cbw_cgrp_llc_map` therefore has a 16-byte key. The update, lookup, and delete paths construct separate stack instances where the padding contents is whatever is lying around on the stack. When the cleanup lookup uses different padding it returns `NULL`, and `cbw_free_llc_ctx()` skips the map delete.

## Proposed fix

Pack the map key:

```c
struct cgroup_llc_id {
        u64 cgrp_id;
        int llc_id;
} __attribute__((packed));
```

This changes the map key from 16 bytes to 12 bytes, so every byte belongs to an initialized field.

mfleming:
Fixes #3693.

`struct cgroup_llc_id` has four bytes of implicit tail padding. BPF hash maps compare every byte of the key, so separately allocated update and lookup keys can differ even when `cgrp_id` and `llc_id` match.

The failed lookups prevent cgroup-exit cleanup from deleting per-LLC contexts. The map eventually fills and a later cgroup initialization fails.

Pack the key so only its declared fields participate in map operations.

### DIFF
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -289,7 +289,7 @@ struct {
 struct cgroup_llc_id {
 	u64		cgrp_id;
 	int		llc_id;
-};
+} __attribute__((packed));
 
 struct cbw_llc_entry {
 	u64	llcx;


multics69 (approved):
Thanks @mfleming ! Good catch!

PR #3696, opened 2026-07-08
scx_lavd: Size cgroup LLC map from topology

### LINKED ISSUE #3694: scx_lavd: LLC context map limits cgroups on large topologies
## Summary

`cbw_cgrp_llc_map` is sized for 65,536 entries using a fixed assumption of at most 32 LLC domains:

```c
CBW_NR_CGRP_MAX     = 2048,
CBW_NR_CGRP_LLC_MAX = CBW_NR_CGRP_MAX * 32,
```

Each leaf cgroup requires one context per LLC domain. Systems with more than 32 LLCs therefore support fewer than the advertised 2,048 cgroups. An Ampere Altra Max M128-30 has 128 private L2 caches and no shared L3, so scx detects 128 LLC domains and the effective limit is only 512 cgroups.

Once the limit is exceeded, cgroup initialization fails and sched_ext disables `scx_lavd`:

```text
Error: EXIT: scx_bpf_error (src/bpf/main.bpf.c:2533: Failed to init a cgroup: -12)
```

This was reproduced with scx v1.1.2 at commit `b68cb3c829312b43785225bcfd445a3ecefef5c2`.

## QEMU reproduction

QEMU 10.2.2 can expose a separate L3 cache for each virtual socket:

```text
-machine q35,smp-cache.0.cache=l3,smp-cache.0.topology=socket
-smp 64,sockets=64,cores=1,threads=1
```

Linux reports 64 CPUs and 64 independent L3/LLC domains with this configuration. In the guest, enable the CPU controller and create persistent leaf cgroups before starting lavd:

```sh
echo +cpu > /sys/fs/cgroup/cgroup.subtree_control

nr_llcs=64
max_entries=65536
max_cgroups=$((max_entries / nr_llcs))

for i in $(seq 1 "$max_cgroups"); do
        mkdir "/sys/fs/cgroup/scx-capacity-$i"
done

scx_lavd --partial --enable-cpu-bw
```

## Proposed fix

Since `scx_lavd` already detects the LLC topology before loading the BPF object we can set the map size at that point using the existing cgroup map capacity:

```rust
let nr_llcs = u32::try_from(order.nr_llcs)?;
let max_entries = skel
    .maps
    .cbw_cgrp_map
    .max_entries()
    .checked_mul(nr_llcs)?;

skel.maps
    .cbw_cgrp_llc_map
    .set_max_entries(max_entries)?;
```

mfleming:
Fixes #3694.

`cbw_cgrp_llc_map` has a fixed 65,536-entry capacity, which supports 2,048 cgroups only on systems with at most 32 LLC domains. On an Ampere Altra Max M128-30, scx detects 128 LLC domains and the effective capacity drops to 512 cgroups.

Set `max_entries` before BPF load using the existing cgroup map capacity multiplied by the detected LLC count.

### DIFF
--- a/scheds/rust/scx_lavd/src/main.rs
+++ b/scheds/rust/scx_lavd/src/main.rs
@@ -486,6 +486,16 @@ impl<'a> Scheduler<'a> {
 
         // Initialize CPU topology with CLI arguments
         let order = CpuOrder::new(opts.topology.as_ref(), opts.no_use_em).unwrap();
+        if opts.enable_cpu_bw {
+            let nr_llcs = u32::try_from(order.nr_llcs).context("Too many LLC domains")?;
+            let max_entries = skel
+                .maps
+                .cbw_cgrp_map
+                .max_entries()
+                .checked_mul(nr_llcs)
+                .context("LLC context map is too large")?;
+            skel.maps.cbw_cgrp_llc_map.set_max_entries(max_entries)?;
+        }
         Self::init_cpus(&mut skel, &order);
         Self::init_cpdoms(&mut skel, &order);
 


multics69:
@mfleming -- Thanks for suggesting the changes. The approach makes sense to me. However, I think the initialization part needs to be improved; the cpu.max library should be self-contained as a library, so I think creating a dependency with the lavd's initialization logic is not idea. I tried another attempt to add a map resize function as a common function. Please take a look: https://github.com/sched-ext/scx/pull/3699

mfleming:
> @mfleming -- Thanks for suggesting the changes. The approach makes sense to me. However, I think the initialization part needs to be improved; the cpu.max library should be self-contained as a library, so I think creating a dependency with the lavd's initialization logic is not idea. I tried another attempt to add a map resize function as a common function. Please take a look: #3699

Yep, looks good to me :+1:

multics69:
> > @mfleming -- Thanks for suggesting the changes. The approach makes sense to me. However, I think the initialization part needs to be improved; the cpu.max library should be self-contained as a library, so I think creating a dependency with the lavd's initialization logic is not idea. I tried another attempt to add a map resize function as a common function. Please take a look: #3699
> 
> Yep, looks good to me 👍

Thanks for checking the PR. Let me close this.

PR #3698, opened 2026-07-09
lib/cgroup_bw: correctness hardening and cleanup

multics69:
This series collects small, self-contained robustness fixes and cleanups
for the cgroup bandwidth (cpu.max) support in lib/cgroup_bw. It is the
outcome of auditing the object lifetimes and teardown races in the
throttle/put-aside/reenqueue paths.

  [1/5] downgrade a benign LLC-lookup failure to debug
        A lookup miss in cbw_reenqueue_cgroup() during concurrent cgroup
        teardown is expected churn, not an error; cbw_err -> cbw_dbg.

  [2/5] drop verbose development debug prints
        Remove bring-up-era cbw_dbg()/cbw_dbg_cgrp() traces. No
        functional change.

  [3/5] remove cgroup ctx map entry before freeing it
        cbw_del_cgroup_ctx() freed the context before deleting its map
        entry, leaving a window where a concurrent lookup could resolve
        a freed/recycled (ABA) context. Reorder to delete-then-free,
        matching cbw_free_llc_ctx().

  [4/5] guard llcx->btq read in cbw_has_backlogged_tasks()
        A plain load of llcx->btq racing teardown's CAS-null could hand
        scx_atq_nr_queued() a NULL or torn pointer. Snapshot with
        READ_ONCE() and skip the LLC on NULL.

  [5/5] harden cached cgx/llcx use with an id check
        Validate that a cached taskc->cgx_raw/llcx_raw still belongs to
        cgrp_id before trusting it. Currently safe (caches are cleared on
        move and teardown); this guards against a future stale-cache
        regression turning into a silent wrong-cgroup access.

### DIFF
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -2581,7 +2581,13 @@ int cbw_reenqueue_cgroup(scx_cgroup_ctx_t *cgx, u64 cgrp_id, u64 nuance)
 		idx = (nuance + i) % TOPO_NR(LLC);
 		llcx = cbw_get_llc_ctx_with_id(cgrp_id, idx);
 		if (!llcx) {
-			cbw_err("Failed to lookup an LLC context: cgid%llu", cgrp_id);
+			/*
+			 * After putting the cgroup ID to cbw_throttled_cgroup_ids[],
+			 * the cgroup is exiting, freeing its LLC context
+			 * (cbw_free_llc_ctx). This is expected under cgroup
+			 * churn, not an error; skip it.
+			 */
+			cbw_dbg("LLC context gone (cgroup exiting): cgid%llu", cgrp_id);
 			continue;
 		}
 
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -1965,6 +1965,7 @@ static
 bool cbw_has_backlogged_tasks(scx_cgroup_ctx_t *cgx)
 {
 	scx_cgroup_llc_ctx_t *llcx;
+	scx_atq_t *btq;
 	int i;
 
 	if (!cgx || !cgx->has_llcx)
@@ -1975,7 +1976,7 @@ bool cbw_has_backlogged_tasks(scx_cgroup_ctx_t *cgx)
 		if (!llcx)
 			continue;
 
-		if (scx_atq_nr_queued(llcx->btq))
+		if ((btq = READ_ONCE(llcx->btq)) && scx_atq_nr_queued(btq))
 			return true;
 	}
 
--- a/lib/cgroup_bw.bpf.c
+++ b/lib/cgroup_bw.bpf.c
@@ -755,11 +755,20 @@ scx_cgroup_ctx_t *cbw_get_cgroup_ctx(struct cgroup *cgrp)
 
 long cbw_del_cgroup_ctx(u64 cgrp_id)
 {
+	/*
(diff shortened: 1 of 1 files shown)


etsal:
[This](https://github.com/sched-ext/scx/pull/3698/changes/4669b433aaef912d4134acdc9a48ab55ae75cf78) is the commit that requires some kind of delayed reclamation, correct?

multics69:
> [This](https://github.com/sched-ext/scx/pull/3698/changes/4669b433aaef912d4134acdc9a48ab55ae75cf78) is the commit that requires some kind of delayed reclamation, correct?

Yes

lib/cgroup_bw.bpf.c
```diff
@@ -2583,7 +2504,13 @@ int cbw_reenqueue_cgroup(scx_cgroup_ctx_t *cgx, u64 cgrp_id, u64 nuance)
...
 		llcx = cbw_get_llc_ctx_with_id(cgrp_id, idx);
 		if (!llcx) {
-			cbw_err("Failed to lookup an LLC context: cgid%llu", cgrp_id);
+			/*
+			 * After putting the cgroup ID to cbw_throttled_cgroup_ids[],
+			 * the cgroup is exiting, freeing its LLC context
+			 * (cbw_free_llc_ctx). This is expected under cgroup
+			 * churn, not an error; skip it.
+			 */
+			cbw_dbg("LLC context gone (cgroup exiting): cgid%llu", cgrp_id);
```
bboymimi, line 2581: I was thinking about what would happen if this were a real use-after-free case. If we just assume this is correct, what happens if the real bug occurs and we ignore it?

lib/cgroup_bw.bpf.c
```diff
@@ -1627,19 +1627,23 @@ int cbw_cgroup_bw_throttled(u64 cgrp_id, u64 taskc_raw)
...
-	if (taskc && taskc->cgx_raw) {
-		cgx_raw = taskc->cgx_raw;
-	} else {
+	/*
+	 * Validate that the cached context still belongs to this cgroup; a
+	 * recycled cgx_raw could otherwise resolve another cgroup's throttle
+	 * state (ABA). Re-resolve on a cache miss or an id mismatch.
+	 */
+	cgx = taskc ? (scx_cgroup_ctx_t *)taskc->cgx_raw : NULL;
+	if (!cgx || cgx->id != cgrp_id) {
```
bboymimi, line 1702: I'm not sure whether the cgx->id will be changed right after the checking. If it happens, it seems cgx->is_throttled will read the wrong value. Or we need to check the cgx->id != cgrp_id again. If a race condition happens, we drop the cgx->is_throttled and goto the previous "if" check to retrieve the cgx again. Also, may set a limit to limit the number of "goto" loop.

lib/cgroup_bw.bpf.c
multics69: Even if the UAF happens and we  access already freed llcx and btq, it does no harm. That’s because llcx and btq aren't physically freed; they're reused. In particular, the destruction of the BTQ will be deferred like RCU and drained before reuse. The following commit in PR 3754 further hardens this path.

https://github.com/sched-ext/scx/pull/3754/changes/eff82c526c25be05b25f84935c31d11163cadebd

lib/cgroup_bw.bpf.c
multics69: Such a thing can happen only when taskc->cgx_raw points to an already freed ctx. That’s a separate bug if there is.

PR #3701, opened 2026-07-10
lib/cgroup_bw: harden cgroup init and the static limits

multics69:
Two cgroup-init robustness fixes for cpu.max in `lib/cgroup_bw` (scx_lavd).

1. **publish cgroup ctx only after full init** — `scx_cgroup_bw_init()`
   inserted the ctx into `cbw_cgrp_map` before creating its LLC contexts, so
   a concurrent throttle could park a task in a cgroup whose `has_llcx` was
   still false. Such a task is invisible to the drain rotation and stalls.
   Publish only after `cbw_init_llc_ctx()` succeeds.

2. **abort cgroups over the static limits** — a cgroup exceeding
   `CBW_CGRP_TREE_HEIGHT_MAX` or `CBW_NR_CGRP_MAX` can't be represented.
   Abort instead -- the kernel falls back to CFS, which enforces cpu.max
   natively, so enforcement is handed back, not lost.


bboymimi (approved):

lib/cgroup_bw.bpf.c
```diff
@@ -1112,6 +1121,33 @@ int scx_cgroup_bw_init(struct cgroup *cgrp __arg_trusted, struct scx_cgroup_init
 
 	cgrp_id = cgroup_get_id(cgrp);
 
+	/*
+	 * Give up bandwidth control for cgroups that exceed the static limits
```
daidavid: Skipping CPU Bandwidth control is not acceptable behavior and can cause service degradations/outages. I rather us fail to load and fall back to default SCHED_OTHER.

multics69: That makes sense. I will change it.

multics69: @daidavid -- I changed to abort when going beyond static limit. Please take a look.

daidavid (approved):

PR #3703, opened 2026-07-13
scx_cosmos: Decouple GPU sync from CPU polling

nvax-r:
## Summary

GPU PID sync was nested inside the CPU utilization polling block. With default --polling-ms 0, that block never runs, leaving the GPU PID map empty or stale even when GPU support is enabled.

Fix this control-flow bugy by moving GPU sync block outside the CPU polling condition while preserving its one-second throttle and error handling. CPU polling remains disabled at 0, while GPU PIDs are now refreshed about once per second when GPU support is active.

## Test
```
$ ~/scx/target/release/scx_cosmos --gpu --polling-ms 0
```

Before the fix, CPU polling was disabled, and GPU PID can't be updated even though we turn the config up.


arighi (approved):
Looks good to me, thanks Richard.

PR #3704, opened 2026-07-14
scx_metrics: add standalone scheduler metrics exporter

newell-purdue:
# scx_metrics with work conservation as first metric

Trying to fill a need to track key metrics for scx schedulers over long periods of time with very little overhead. 

Next steps listed in comment of: (1) handling tasks with affinity, (2) grouping metrics by task groups at cgroup boundary, and (3) including new low-overhead metrics critical to tracking scx scheduler behavior.

## Testing

Ran on a 48 core machine where I ran `stress -c 20`. Then, I toggled on an scx scheduler that happens to do poorly in terms of scheduling `cargo run -p scx_layered -- f:scheds/rust/scx_layered/examples/avgruntime.json`, effectively only limited cores to be used due to protected mode.

Ran the following for scx_metrics, allowing prometheus to export the counters:
```
cargo run -p scx_metrics -- --listen 127.0.0.1:9345
```

Screenshot 2026-07-13 at 10 31 45 PM
Screenshot 2026-07-13 at 10 34 03 PM


newell-purdue:
@kkdwvd to review

kkdwvd:
It might make sense to keep this under tools/, but otherwise looks great.

sirlucjan:
> It might make sense to keep this under tools/, but otherwise looks great.

Good point, in my opinion, it definitely belongs in /tools. 

What do you think @htejun?

newell-purdue:
Apologies on delay, had some high priority work come up. Can be more responsive now and get this in while working on some follow-ups. Did a rebase onto latest, moved to tools, and retested with grafana.

sirlucjan (changes requested):
Could you rebase on top main? 

1. rust edition update: https://github.com/sched-ext/scx/pull/3798
2. libbpf update: https://github.com/sched-ext/scx/pull/3810

newell-purdue:
Rebased onto e9ddd720e3a166660e596a1a0b871cf463bbbad9. Re-tested, still able to measure work conservation matching original test plan

sirlucjan (changes requested):
You didn't include rust edition changes (edition 2024, rust version 1.91) and libbpf update (0.27.1 version)

newell-purdue:
Apologies, missed that, updated those now

kkdwvd:
Before we move ahead with this, could we also figure out how to make this work in an optimal way (overhead vs fidelity) for affinitized tasks? In practice we will have many such cases.

newell-purdue:
**Fix from affinity testing**. Made fix based off testing with affinities. Was seeing that idle cpus was reporting 0 on server with 8 cpus when I ran "taskset -c 0 stress -c 8". Instead as I'm iterating cpus anyways, counting when active running pid is idle to determine if its busy. This correctly shows that "taskset -c 0 stress -c 8" has the invalid work conservation of 1/8 (when it should be 1/1).

**New commit handling affinity**. There seems to be no way to walk run-queues and learn affinities, so need to add tracking as tasks enter/exit queues. Trying to be minimal, that means for tasks with affinities we store those affinities and those without affinities just bump/decrement a count. Then, when sampling, track which cpus are busy, greedily try to associate affinitized tasks unclaimed cpus, and finally include non-affinity.

Agree handling affinity from the start could be important, otherwise will always under-report work conservation. Prefer accurate and more costly, and then strive to optimize later.

**Testing**. Ran `taskset -c 1,2 stress -c 8` without this new commit and then with the new commit. Without new commit, this reported 2/8 work conservation each sampling period. With the new commit, reports 2/2 work conservation each sampling period.

PR #3710, opened 2026-07-16
scx_layered: Fix some issues with protected configuration for a layer

ByteLab-David:
A protected layer's CPUs are supposed to serve only the owner layer. layered_dispatch() currently refuses to consume any other layer's DSQ on them, but that gate only covers dispatch-time consumption. Three other paths let a foreign task occupy a protected CPU:

- In keep_running(), both success branches refresh p->scx.slice to the task's full layer slice without ever re-checking whose CPU it is. A foreign occupant (could be e.g. pre-existing from a cpumask re-fit, fallback-DSQ execution, or an open preempt) chains full slices on a CPU which dispatch would never have granted.

  I saw this when reviewing a perfetto trace of a 4-vCPU node, where background spinner tasks in a foreign open layer were saturating CPUs owned by a protected nginx layer. The spinner held a protected, nginx-owned CPU for 51.6ms (its layer's full 50ms slice) continuously while an nginx task sat in runnable state the whole time.

- At the end of dispatch, if a protected layer has nothing else to run on a CPU it owns, it will continue running the previously-running task. This task, however, might be from another layer. For example, it could have been consumed from the fallback DSQ. In this case, we're giving that task a full slice refresh. This isn't necessarily _wrong_, but it's risky in the sense that if the layer gets any additional work, it could cause tail latencies.

- try_preempt_cpu(): inserts into the candidate's local DSQ via SCX_DSQ_LOCAL_ON, bypassing dispatch and its protection gate entirely, so an open preempt layer could seize a protected CPU outright.

This diff closes all three issues. For foreign tasks that must run on protected CPUs (e.g. per-CPU kthreads via hi_fb, lo_fb forced-share, antistall), this change should only cost them a round trip through their own DSQs.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -1704,6 +1704,15 @@ s32 BPF_STRUCT_OPS(layered_select_cpu, struct task_struct *p, s32 prev_cpu, u64 
 	return prev_cpu;
 }
 
+/*
+ * Is @taskc a "guest" on a CPU (@cpuc) owned by a protected layer?
+ */
+static __always_inline bool is_protected_layer_guest(struct cpu_ctx *cpuc,
+						     struct task_ctx *taskc)
+{
+	return cpuc->is_protected && taskc->layer_id != READ_ONCE(cpuc->layer_id);
+}
+
 enum preempt_flags {
 	PREEMPT_FIRST		= 1LLU << 0,
 	PREEMPT_IGNORE_EXCL	= 1LLU << 1,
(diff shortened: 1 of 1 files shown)


htejun:
A few minor suggestions:

- `is_protected_layer_guest()` can use the existing `cpuc->is_protected` instead of re-deriving through `layers[cpuc->layer_id]`: `return cpuc->is_protected && taskc->layer_id != READ_ONCE(cpuc->layer_id);` — equivalent (open layers are never protected) and keeps one source of truth.

- The end-of-dispatch hunk mixes the pre-existing space indentation with tabs; might as well re-indent the whole block to tabs.

ByteLab-David:
TFTR, applied both suggestions.

htejun (approved):

PR #3711, opened 2026-07-17
scx_lavd: Scale the deadline compete window by the oversubscription ratio

### LINKED ISSUE #3303: [scx_lavd] Huge performance regression in hackbench with scx v1.0.20
With v1.0.20, I see a really big performance regression using scx_lavd in hackbench:

- v1.0.19: 94 seconds
- v1.0.20: 135 seconds

Tests were run on kernel 6.19-rc7 for scx v1.0.19, and kernel 6.19-final for scx v1.0.20, ceteris paribus.

Other benchmarks I've run showed only very little or no difference.

There have been quite a lot of commits between .19 and .20, and unfortunately I don't have the time for a proper bisect. In both runs, the `--performance`  parameter was used. Hardware is a poor old dual-core (2c/4t) Skylake.

Maybe someone else can confirm this problem.

ByteLab-David:
In calc_when_to_run(), lavd computes a task's vtime as (logical_clock - LAVD_DL_COMPETE_WINDOW + delta). The idea here is to give a boost to tasks that have bursty workloads by giving them a boost when they're first enqueued (thus, tasks that run for short windows enqueue more frequently, and get more frequent boosts).

In a8a25fcb ("Increase greedy factor in deadline calculation"), this window was extended as part of an effort to prevent fork bombs from starving latency-critical tasks. Unfortunately, this also introduced a regression in lavd:

When a task's deadline delta is smaller than the aforementioned window, then its vtime lands behind cur_logical_clk. cur_logical_clk is monotonically increasing, and advance_cur_logical_clk() therefore correctly checks that the vtime of a task is not behind the current clock value. However, with a much larger window, it will become much less likely that a given task will actually be able to increment the logical clock due to its vtime being a function of subtracting the window. For workloads such as hackbench which have a ton of bursty tasks, this can cause problems because the only fairness mechanism is now dl_delta.

To fix this, we can scale the window by how oversubscribed the system is. We can grant the full window while the runnable backlog fits the active CPUs, and shrink it by nr_active/nr_queued beyond that so aging resumes as the backlog grows. This uses the same normalization calc_sys_time_slice() and advance_cur_logical_clk() already apply, and the same nr_queued <= nr_active boundary as can_boost_slice().

On my AMD 7950X, this resulted in the following improvements on hackbench -g8 -l10000:

Before:
25.54 +/- 0.37x

After:
17.43 +/- 0.04s

Fixes: #3303

Note: This fix was found + initial draft created by the Crucible project: https://github.com/Byte-Lab/crucible

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/lat_cri.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/lat_cri.bpf.c
@@ -346,18 +346,33 @@ static u64 calc_virtual_deadline_delta(struct task_struct *p,
 	return deadline >> LAVD_SHIFT;
 }
 
+static u64 calc_compete_window(void)
+{
+	u64 nr_q = sys_stat.nr_queued_task;
+	u64 nr_a = max(sys_stat.nr_active, 1);
+
+	/*
+	 * The compete window is meant to boost bursty tasks that wake and
+	 * block frequently, as they'll benefit from the enqueue-time boost
+	 * more frequently than longer-running tasks. However, if the window is
+	 * too large, then the global clock will be incremented very
+	 * infrequently, which can cause fairness issues.
+	 *
+	 * We therefore shrink the window proportionally to the
+	 * oversubscription ratio so that the window scales inversely with the
+	 * number of enqueued tasks.
+	 */
+	if (nr_q > nr_a)
+		return (LAVD_DL_COMPETE_WINDOW * nr_a) / nr_q;
+	return LAVD_DL_COMPETE_WINDOW;
+}
+
 __hidden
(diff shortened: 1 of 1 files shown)


multics69 (approved):
Thanks @ByteLab-David for the patch. This makes a lot of sense. 

@daidavid -- If you don't see regression on your side, let's merge it.

daidavid:
Nice find, things look fine from limited sanity testing.

>Note: This fix was found + initial draft created by the Crucible project: https://github.com/Byte-Lab/crucible

Out of curiosity, @ByteLab-David. I'm only asking because of this note. Is this patch fully drafted by AI? If so, the commit text and comments are very readable and I'd like to steal some of that.

daidavid (approved):

PR #3714, opened 2026-07-23
scx_lavd: Modulate CPU Stickness using Task Warmth

daidavid:
LAVD, amongst other sched_ext schedulers, aggressively migrate tasks by always favoring completely idle cores. On symmetric SMT systems, this generally means finding the CPU with the highest available compute capacity(CPU with an idle sibling). This type of behavior is useful for truly latency sensitive and compute intensive workloads by maximizing work conservation.

For workloads such as caching services that are both compute and memory throughput bound, we've observed that over migrating tasks can cause TLB thrashing and reduced IPC. In particular, most of the increased TLB misses due to migrations are from accesses to the application heap memory. In userspace managed memory schemes such jemalloc where per-thread caches are employed and are typically hot, migrating tasks and thrashing TLBs can adversely affect performance (~3% decrease in IPC at nominal loads when comparing to EEVDF).

Introduce a knob to modulate CPU stickiness by prefering previous logical CPU with an estimated waiting period based on a task's historical time spent on previous CPU.


scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -691,6 +717,33 @@ u32 preemption_vulnerability(u16 normalized_lat_cri, u32 util_est)
...
+	if (!cpuc)
+		return false;
+
+	heat = task_cpu_warmth(taskc, cpu, now);
+	budget = (warm_cpu_ns * (LAVD_WARM_HEAT_MAX + heat)) >> LAVD_SHIFT;
+	est = READ_ONCE(cpuc->est_stopping_clk);
+	wait = (est > now) ? (est - now) : 0;
+
+	wait += (u64)scx_bpf_dsq_nr_queued(cpu_to_dsq(cpu)) *
+		taskc->avg_runtime_wall;
```
bboymimi: I was curious about the difference between `sys_stat.slice_wall` and `taskc->avg_runtime_wall`. It seems that `sys_stat.slice_wall` represents the system-wide average and better reflects the waiting time than `taskc->avg_runtime_wall` does. How do you think?

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -240,6 +254,11 @@ struct task_ctx {
 					/* Estimated task util using ravg duty cycle */
 	struct ravg_data avg_util_ravg;	/* Running average of task utilization using ravg */
 	char	waker_comm[TASK_COMM_LEN + 1]; /* last waker's comm */
+
+	/* --- per-CPU warmth clock (cache/TLB/branch-predictor state) --- */
+	u64	cpu_heat_updated_clk;	/* last time cpu_heat was integrated or decayed */
+	u16	cpu_heat;		/* residence-integrated heat [0, LAVD_WARM_HEAT_MAX] */
+	u16	warm_cpu_id;		/* CPU the heat belongs to; LAVD_CPU_ID_NONE if none */
```
multics69: Instead of ading a new one, we can reuse `prev_cpu_id` which is updated at `update_stat_for_running()`. So it should be a drop-in replacement.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -240,6 +254,11 @@ struct task_ctx {
 					/* Estimated task util using ravg duty cycle */
 	struct ravg_data avg_util_ravg;	/* Running average of task utilization using ravg */
 	char	waker_comm[TASK_COMM_LEN + 1]; /* last waker's comm */
+
+	/* --- per-CPU warmth clock (cache/TLB/branch-predictor state) --- */
+	u64	cpu_heat_updated_clk;	/* last time cpu_heat was integrated or decayed */
```
multics69: I think `last_stopping_clk` would be a better name since it follows the naming style of other clock fields.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -125,6 +125,19 @@ enum consts_internal {
...
 };
 
+/*
+ * Per-CPU warmth clock: a proxy for the microarchitectural state a task warms
+ * up on a CPU (L1/L2 cache, TLB entries, branch-predictor history). Heat builds
+ * up with on-CPU residence and decays once the task leaves; a warmer previous
+ * CPU earns a longer wait before the task migrates.
+ */
+enum consts_cpu_warmth {
+	LAVD_WARM_HEAT_MAX		= 1024,	  /* full-heat scale */
```
multics69: Instead of defining a new one, let's uer `LAVD_SCALE`.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -691,6 +717,33 @@ u32 preemption_vulnerability(u16 normalized_lat_cri, u32 util_est)
...
+	struct cpu_ctx *cpuc = get_cpu_ctx_id(cpu);
+	u64 heat, budget, est, wait;
+
+	if (!cpuc)
+		return false;
+
+	heat = task_cpu_warmth(taskc, cpu, now);
+	budget = (warm_cpu_ns * (LAVD_WARM_HEAT_MAX + heat)) >> LAVD_SHIFT;
+	est = READ_ONCE(cpuc->est_stopping_clk);
+	wait = (est > now) ? (est - now) : 0;
```
multics69: We can use `time_delta` here: `wait = time_delta(est, now);`

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -956,6 +967,14 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
 		reset_task_flag(taskc, LAVD_FLAG_IDLE_CPU_PICKED);
 	}
 
+	/*
+	 * pick_idle_cpu() sets LAVD_FLAG_WARM_CPU when it chose to wait on
+	 * the previous CPU. Consume it here so the dispatch below honors that
+	 * one decision instead of recomputing it against a clock that has moved.
+	 */
+	warm_wait = test_task_flag(taskc, LAVD_FLAG_WARM_CPU);
+	reset_task_flag(taskc, LAVD_FLAG_WARM_CPU);
```
multics69: I think these two lines can be merged to the if condition at line 1033:

```
} else if (test_task_flag(taskc, LAVD_FLAG_WARM_CPU)) {
	reset_task_flag(taskc, LAVD_FLAG_WARM_CPU);
    ...
}
```

scheds/rust/scx_lavd/src/bpf/idle.bpf.c
```diff
@@ -715,6 +716,32 @@ s32 pick_idle_cpu(struct pick_ctx *ctx, bool *is_idle)
...
 
+	/*
+	 * Warm-CPU preference: Prefer the previous CPU while its cache and TLB
+	 * are still warm. Take it if idle or, if it is busy but predicted to
+	 * free up within the warmth-extended budget.
+	 */
+	if (warm_cpu_ns && ctx->prev_cpu >= 0 &&
+	    bpf_cpumask_test_cpu(ctx->prev_cpu, cast_mask(ctx->active)) &&
+	    bpf_cpumask_test_cpu(ctx->prev_cpu, ctx->p->cpus_ptr)) {
+		if (scx_bpf_test_and_clear_cpu_idle(ctx->prev_cpu)) {
```
multics69, line 731: The condition needs to be refined to consider asym CPUs (big/LITTLE). If the task type and prev_cpu's core type match, take this if branch. Otherwise, don't take this if branch to give a chance for cross-CPU cluster migration. I will take over this after this PR lands.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
multics69: This is a rough estimation of when the enqueued task will be served. It is rough because it makes two assumptions: 1) the task will be served last in the queue and 2) the runtime duration of all the previously waiting tasks will be the same. The first one is hard to be more accurate but we can make the second one more accurate by maintaining `qload_invr` per-core as we did per-domain.

I am okay with the current approximation for now, but let's add comments about our assumptions for future improvement.

Let's improve this once this PR lands. I have another patch (not yet submitted) which estimates the completion of a domain. That patch can be extended for better estimation of the completion time of a core.

multics69 (changes requested):
Overall, I like the PR. I left some comments for improvements and future TODOs.

multics69 (approved):
Thanks for addresing the comments!

PR #3720, opened 2026-07-28
scx_lavd: fix combinatorial explosion in gen_pds_combinations

### LINKED ISSUE #3340: scx_lavd --performance leaks massive memory (~34GB) and pegs CPU on CachyOS 6.19.2 (Intel Core Ultra 9 275HX)

carvsdriver:
## Problem

`gen_pds_combinations()` in `scx_lavd` built every possible k-subset (for `k = 1..n-1`, plus the full set) of the fully-expanded performance-domain vector, using the `combinations` crate. On hybrid Intel CPUs (Alder Lake, Arrow Lake, ...) the kernel registers one performance domain per CPU thread, even though domains of the same core type (P-core / E-core / LP-E-core) are functionally identical. For a 28-thread hybrid CPU this meant evaluating `2^28 - 2` (~268 million) combinations, each allocating a `Vec` and a `PDSetInfo` (`BTreeSet`/`BTreeMap`) — observed as ~20-34GB of RSS growth and a pegged CPU core during scheduler init.

Closes #3340. Reproduced and confirmed on two different hybrid CPUs (Core Ultra 9 275HX, 28 threads; Core Ultra 9 285K, 24 threads).

## Fix

Group performance domains by identical `perf_table` before enumerating. Domains within a group are interchangeable for scheduling purposes, so "which specific domains are active" collapses into "how many CPUs from this group are active" — turning a per-domain search (`2^n_domains`) into a per-group one (`prod(weight_i + 1)`). For the 28-thread case, that's `(8+1)*(16+1)*(4+1) - 1 = 764` combinations instead of ~268 million.

Two things added on top of the core fix:

- **A hard cap (`MAX_PDS_COMBINATIONS`)** on the search space. Grouping only helps when domains actually collapse into a handful of groups; a topology with many genuinely distinct performance tables (e.g. per-core binning) would reproduce the same unbounded blowup with group count == domain count. If the cap would be exceeded, this falls back to the one combination the original code always included unconditionally regardless of `k`: every CPU at its own top performance state.
- **A bounded "backward-fill" variant** alongside the default "forward-fill" when filling a partially-selected group. `insert_pds_combinations()` keeps every `PDSetInfo` tied for the lowest power at a given performance level so `dist()` can later pick whichever tied candidate best overlaps the previously-active domain set (cache locality / smooth transitions). Since grouping produces only one representative combination per count by construction, `dist()` would otherwise always be choosing between exactly one candidate — this gives it at least two. This is a **partial mitigation, not a full restoration** of the original search's diversity; happy to discuss whether a fuller fix is worth the added complexity.

## Testing

- `cargo build --release -p scx_lavd`: clean, no warnings.
- `cargo test --release -p scx_lavd`: 3 new tests, all passing — a synthetic reproduction of the reported 28-domain/3-group hybrid topology (exact combination count assertion), a pathological 24-distinct-domain topology asserting the safety-cap fallback engages, and a single-uniform-group case.
- `cargo fmt -p scx_lavd -- --check`: clean.
- Manually verified on real hardware (Core Ultra 9 285K, 8P+16E): standalone (idle and under real game load) and through the normal `scx_loader`/`scx-manager` path, RSS stayed flat at ~78MB with <1% CPU over multiple 2-3 minute windows, vs. the pre-fix binary reaching ~20-25GB RSS and ~98-100% CPU within 60-90 seconds under identical conditions.

## Attribution

An initial version of this fix, identifying the same root cause and using the same grouping strategy, was posted by @Cloo77 in [a comment on #3340](https://github.com/sched-ext/scx/issues/3340#issuecomment-4753488018). That version wasn't reviewed upstream and had no safety cap, diversity handling, doc comments, or tests; this PR reworks it with those additions after independently reproducing and verifying the leak and the fix. Credited as co-author on the commit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -953,12 +953,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "combinations"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "728b7ab3119b5167cc60a4aad16b6086b762278f5571cc650e7eed2e89a23a15"
-
-[[package]]
 name = "combine"
 version = "4.6.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6241,7 +6235,6 @@ dependencies = [
  "clap",
  "clap-num",
  "clap_main",
- "combinations",
  "crossbeam",
  "ctrlc",
  "gpoint",
--- a/scheds/rust/scx_lavd/Cargo.toml
+++ b/scheds/rust/scx_lavd/Cargo.toml
@@ -34,7 +34,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter", "fmt", "parkin
 static_assertions = "1"
 plain = "0.2"
 gpoint = "0.3"
-combinations = "0.1"
 rlimit = "0.11"
 nix = "0.31"
 
--- a/scheds/rust/scx_lavd/src/cpu_order.rs
+++ b/scheds/rust/scx_lavd/src/cpu_order.rs
@@ -8,7 +8,6 @@
 
 use anyhow::anyhow;
 use anyhow::Result;
-use combinations::Combinations;
 use itertools::iproduct;
 use scx_utils::CoreType;
 use scx_utils::Cpumask;
@@ -24,6 +23,7 @@ use std::collections::BTreeSet;
 use std::collections::HashSet;
 use std::fmt;
 use std::hash::{Hash, Hasher};
+use std::sync::Arc;
 use tracing::debug;
 
 #[derive(Debug, Clone)]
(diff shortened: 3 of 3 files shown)


multics69:
Thanks @carvsdriver and @Cloo77 for the patches. I like the core idea of the changes- merging performance domains with the same performance table to avoid combinational explosion. However, the code is entangled and hard to understand. I will try my version based on your core ideas in a separate PR.

carvsdriver:
> Thanks @carvsdriver and @Cloo77 for the patches. I like the core idea of the changes- merging performance domains with the same performance table to avoid combinational explosion. However, the code is entangled and hard to understand. I will try my version based on your core ideas in a separate PR.

No worries mate.  Thanks for taking the time to look at the submission.  I honestly just took a crack at it based on the original approach posted in the bug report.  I assumed there'd probably be a cleaner way to do it.

multics69:
Close this PR as a refined version https://github.com/sched-ext/scx/pull/3724 is posted.

PR #3721, opened 2026-07-28
scx_rusty: release retained task cpumask before reuse

ltrager:
Linux 7.2 changed BPF map update and delete semantics in commit a3a81d247651 (bpf: Cancel special fields on map value recycle). Referenced kptrs are no longer eagerly destroyed when a map value is recycled because their destructors are not safe in every BPF execution context.

task_masks is a preallocated hash map whose value owns a referenced bpf_cpumask kptr. A recycled entry can therefore retain the old mask. Rusty then exchanges in a new mask, sees the retained pointer, and reports kptr already had cpumask, causing scheduler startup to fail.

Detach and release any retained cpumask before initializing the entry. Older kernels already leave the slot empty, so the exchange returns NULL and this is a no-op there.

Validated on Linux 7.2.0-rc4+ and Linux 6.18.40.

### DIFF
--- a/scheds/rust/scx_rusty/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_rusty/src/bpf/main.bpf.c
@@ -123,6 +123,21 @@ static s32 create_save_cpumask(struct bpf_cpumask **kptr)
 	return 0;
 }
 
+/*
+ * Hash map value recycling may retain referenced kptrs instead of eagerly
+ * releasing them. Explicitly detach and release any retained cpumask before
+ * initializing a recycled map value. On older kernels the exchange normally
+ * returns NULL, making this a no-op.
+ */
+static void release_retained_cpumask(struct bpf_cpumask **kptr)
+{
+	struct bpf_cpumask *cpumask;
+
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_rusty/src/bpf/main.bpf.c
```diff
@@ -1688,6 +1703,7 @@ s32 BPF_STRUCT_OPS_SLEEPABLE(rusty_init_task, struct task_struct *p,
 		return -EINVAL;
 	}
 
+	release_retained_cpumask(&mask_map_value->instance);
```
arighi, line 1706: I'm wondering if we should release the cpumask in rusty_exit_task(). We could keep the cleanup here as well, as a defensive backup, but it shouldn't be the primary ownership cleanup.

ltrager: I think this does what you asked. Verified the docker test still works with this.

arighi (approved):
LGTM, thanks!

PR #3724, opened 2026-07-31
scx_lavd: fix the combinatorial explosion in the CPU set search

multics69:
`gen_pds_combinations()` searches for the cheapest set of CPUs reaching a target performance by enumerating how many CPUs to take from each performance domain, which is `prod(nr_cpus_i + 1) - 1` combinations. That is a handful when a performance domain covers several CPUs. But Intel hybrid processors register one performance domain per CPU, so every factor becomes `1 + 1` and the product becomes `2^nr_cpus`: ~268 million combinations on a 28-thread part, each allocating a `PDSetInfo`. The RSS grows to tens of gigabytes and a CPU stays pegged while the scheduler starts up, which is [sched-ext/scx#3340](https://github.com/sched-ext/scx/issues/3340).

The performance domains of such a processor are not all different, though. All the P-cores share one performance table, and so do all the E-cores, so the domains within a group are interchangeable: taking any 3 of the 8 P-cores gives
the same performance at the same power. This series groups them into *equivalence performance domains* and enumerates over those, which turns `2^28 - 1` into `(8 + 1) * (16 + 1) * (4 + 1) - 1 = 764` on that processor, without dropping any distinct combination.

### [1/3] scx_utils: add equivalence performance domains to the energy model

Groups the performance domains sharing an identical performance table into `EqPerfDomain`. The relation is a property of the energy model — exact performance table equality — so it lives next to `PerfDomain` and is reusable by
any scheduler doing an energy-model CPU search.

### [2/3] scx_lavd: build the CPU sets from equivalence performance domains

Enumerates the CPU counts over the equivalence performance domains as a cartesian product, once at initialization since the counts do not depend on the CPU utilization, and expands a combination into performance domains and states only when it is evaluated.

Grouping only helps when the domains do collapse, which per-core binning could prevent, so the enumeration is capped; above the cap only the runs of equivalence performance domains are considered, at most `3 * nr_cpus` of them.

Drops the `combinations` dependency in favour of `itertools`.

### [3/3] scx_lavd: take the preferred CPUs of an equivalence performance domain first

Not part of the fix. Members of an equivalence performance domain were in performance domain id order, so a partial fill took the lowest ids rather than the preferred CPUs. Orders them, and the equivalence performance domains themselves, by CPU preference.

## Testing

The RSS and CPU figures above are from [#3340](https://github.com/sched-ext/scx/issues/3340) and [#3720](https://github.com/sched-ext/scx/pull/3720), measured on a different implementation of the same idea, not on this series.

Verified here on a 20-CPU Intel hybrid processor by diffing the resulting CPU order table against the pre-series binary:

- the ten capacity bounds are identical,
- every level takes the same number of CPUs from each equivalence performance   domain, so no combination was lost,
- [3/3] changes only which CPUs are used, and touches fewer L2 domains at five of the ten levels and never more.

Plus unit tests: 4 in `scx_utils` for the grouping, 5 in `scx_lavd` for the enumeration and the ordering, including the capped fallback and the exact combination counts. `cargo build --release`, `cargo fmt --check`, and `cargo clippy` are clean.

Note that the capped fallback is not reachable on any processor available today — it needs enough distinct performance tables to exceed 100,000 combinations — so it is covered by unit tests only.

## Credit

The bug was reported in [sched-ext/scx#3340](https://github.com/sched-ext/scx/issues/3340).

[Cloo77](https://github.com/Cloo77) identified the root cause and the grouping strategy, and posted an initial fix as [a comment on that issue](https://github.com/sched-ext/scx/issues/3340#issuecomment-4753488018).

[carvsdriver](https://github.com/carvsdriver) independently confirmed and measured the leak on a Core Ultra 9 275HX and a Core Ultra 9 285K, and proposed a reworked version of that fix in [sched-ext/scx#3720](https://github.com/sched-ext/scx/pull/3720).

This series supersedes [#3720](https://github.com/sched-ext/scx/pull/3720). It keeps their grouping idea and reworks it around an explicit equivalence performance domain in the energy model.


multics69:
@carvsdriver and @Cloo77 -- Please take a look at this PR, which adopted your idea and refined it a bit. I tested it on my Intel machine, but if you can test it on your side, it would be great.

carvsdriver:
## Test results (Intel Core Ultra 9 285K, 24 CPUs, 8P+16E, Arrow Lake - 32G DDR5)

### Build & automated tests

```
cargo build --release -p scx_lavd   -- clean, no warnings
cargo test --release -p scx_lavd    -- 5 passed, 0 failed
  test_eq_pds_in_cpu_preference_order
  test_members_in_cpu_preference_order
  test_uniform_combinations
  test_uncollapsed_combinations_fall_back_to_runs
  test_hybrid_combinations

cargo test --release -p scx_utils   -- 28 passed, 0 failed
  (includes test_group_hybrid_perf_doms, test_group_distinct_perf_doms,
   test_group_uniform_perf_doms for the new EqPerfDomain grouping logic)
```

### Standalone, idle (3 min, sampled every 10s)

| Metric | Result |
|---|---|
| sched_ext attach | lavd_1.1.2_x86_64_unknown_linux_gnu, confirmed via /sys/kernel/sched_ext/root/ops |
| RSS | flat at 79,140-79,200 KB the entire time |
| CPU | 6.4% down to 0.2% (settles quickly) |
| System memory/swap | unaffected |

### Standalone, under real load - Cyberpunk 2077 load + benchmark (5 min, sampled every 10s)

| Metric | Result |
|---|---|
| scx_lavd RSS | flat at 79,140 KB, never moved |
| scx_lavd CPU | 3.0% down to 0.1% |
| System load average | 0.79, peaked at 7.10 during the benchmark, back down to 1.80 after |
| System memory | 8.2G, peaked at 14G (the game itself, not the scheduler), back to 8.4G after closing |

### Through the normal deployment path (scxctl / scx-manager)

- Installed at /usr/bin/scx_lavd
- `scxctl start -s scx_lavd -m gaming` -- "running Lavd in Gaming mode"
- /sys/kernel/sched_ext/root/ops shows lavd_1.1.2_x86_64_unknown_linux_gnu attached
- Currently running for ongoing real-world testing

No leak, no runaway CPU, and no swap pressure in any of the idle, benchmark, or normal-path scenarios above.

Cloo77:
Tested this PR series on my Asus ROG G835LX with the same **Intel Core Ultra 9 275HX** (24 CPUs, 8P + 16E - Arrow Lake architecture).
### Automated Tests & Quality Checks
- `cargo build --release -p scx_lavd`: Clean build.
- `cargo test --release -p scx_utils`: 28 passed, 0 failed (all `EqPerfDomain` tests pass).
- `cargo test --release -p scx_lavd`: 5 passed, 0 failed (all combination enumeration and preference ordering tests pass).
- `cargo fmt --check` and `cargo clippy`: Clean.
### Runtime Metrics & Memory Footprint (Standalone & Production Deployment)
- **Initialization duration**: Combination search and capacity bound calculation completed in **~74 ms** at startup (compared to pegging a CPU core indefinitely and growing memory to tens of GBs prior to this fix).
- **RSS Memory**: Flat at **~80 MB** (`80,184 KB`) continuously, zero memory growth or leak.
- **Scheduler Attachment**: Confirmed active via `/sys/kernel/sched_ext/root/ops` (`lavd_1.1.2_x86_64_unknown_linux_gnu`).
- **Deployment**: Installed to `/usr/bin/scx_lavd` and currently running seamlessly under `scx_loader.service`.
### Personal tests and feels
I've run on my computer and played about an hour Forza Horizon 6 without any trouble, scx_lavd seems to be working as intended, when in performance mode major tasks seems to go to P-Core while E-Cores handle back-end tasks.
Still following mode selector and balance P-Cores & E-Cores accordingly.
### Summary
The equivalence performance domain grouping completely eliminates the combinatorial explosion ($2^{28} \rightarrow 764$ combinations on this processor) while preserving optimal CPU selection and order preference. 
Huge thanks to @multics69 for structuring this cleanly into `scx_utils`!

multics69:
> Are there specific tests and log output you'd like to see?

If possible, please check the PR on your hardware and check the RSS size.

multics69:
@carvsdriver and @Cloo77 -- I will close the other two PRs.

jlucaso1:
Tested this on a Core Ultra 9 275HX (same CPU as #3340) and it fixes the hang here.

Setup: Arrow Lake-HX, 8 P + 16 E, 24 threads, no SMT. CachyOS, `linux-cachyos` 7.1.5-1. Baseline is the distro's `scx-scheds` 1.1.2-1; patched build is this branch off the same 1.1.2. `intel_pstate` with HWP, no `schedutil` and no EAS (`/proc/sys/kernel/sched_energy_aware` doesn't exist on this box).

**Before:** `scx_lavd` never attaches. It just spins at 97% CPU on one core, with `enable_seq=0` and `nr_rejected=0` the entire time, so the struct_ops was never registered, not even rejected. Running with `-v` gives zero libbpf output; the last thing printed is the `Opts { ... }` dump and then nothing, which puts the hang before any BPF work happens. It also ignores SIGTERM: `timeout 8` left it running for 85s, and `scxctl stop` returned `stopped` with exit 0 while the process stayed alive as an orphan still burning the core. Only SIGKILL gets rid of it. That last part bit me before I understood what was going on, and I suspect it's why the original reporter had to force kill via systemd. Memory grows the whole time too: I didn't have RSS instrumented on that first run, but I caught it at roughly 14 GB in btop before killing the process, so it's the same runaway allocation as the ~34 GB in #3340, just caught earlier.

I initially assumed a struct_ops ABI mismatch against this kernel, so I pulled the embedded BPF object out of `/usr/bin/scx_lavd` and diffed `struct sched_ext_ops` against the kernel BTF: 45 members, `size=472`, identical offsets on both sides. So, not ABI, which lines up with this PR putting the blowup in the userspace CPU-set search. `scx_bpfland` and `scx_cosmos` both attach fine on the same kernel (~18-19 MB RSS, <2% CPU), so sched_ext itself is healthy here.

**After:** attaches normally, and from the scheduler's own log timestamps the CPU-set search finishes **63.2 ms** in, emitting **11** CPU sets (all within 31 µs of each other). RSS sits flat at 78 MB for the whole 15s run and it unregisters cleanly on SIGTERM.

```
before:  state=disabled enable_seq=2
t= 5.5s  cpu=10.6  rss=78MB  state=enabled  enable_seq=3
t=14.5s  cpu=4.0   rss=78MB  state=enabled  enable_seq=3
EXIT: unregistered from user space
after:   state=disabled enable_seq=3
```

For context, on 24 threads the old `prod(nr_cpus_i + 1) - 1` enumeration works out to 2^24 = 16.7M combinations, which matches "never finishes" pretty well. The 63.2 ms is in the same ballpark as the ~74 ms @Cloo77 measured on the same 275HX.

One caveat on that `cpu=` column: it's `ps`'s cumulative average, not an instantaneous sample, so the decay from 10.6% to 4.0% is just the ~0.6s of startup being amortized. Steady state is effectively idle.

Happy to rerun with other flags or share the test script if it's useful.

rust/scx_utils/src/energy_model.rs
```diff
@@ -68,8 +93,48 @@ impl EnergyModel {
...
+        perf_doms: &BTreeMap<usize, Arc<PerfDomain>>,
+    ) -> BTreeMap<usize, Arc<EqPerfDomain>> {
+        let mut eq_perf_doms: Vec<EqPerfDomain> = vec![];
+
+        for pd in perf_doms.values() {
+            match eq_perf_doms
+                .iter_mut()
+                .find(|eq_pd| eq_pd.perf_table == pd.perf_table)
+            {
+                Some(eq_pd) => {
```
bboymimi, line 119: This makes total sense to group the performance domains together with the same performance table and the cpumask "span" to record CPUs within this equivalence performance domain.

scheds/rust/scx_lavd/src/cpu_order.rs
```diff
@@ -951,23 +981,161 @@ impl<'a> EnergyModelOptimizer<'a> {
...
+    fn gen_run_nr_cpus(max_nr_cpus: &[usize]) -> Vec<Vec<usize>> {
+        let nr_eq_pds = max_nr_cpus.len();
+        let mut combinations = vec![];
+
+        // Take 1 to all the CPUs of the @i-th equivalence performance domain,
+        // which is the last one of a forward run, the last one of a backward
+        // run, and the only one of a single run.
+        for i in 0..nr_eq_pds {
+            let mut forward = vec![0; nr_eq_pds];
+            forward[..i].copy_from_slice(&max_nr_cpus[..i]);
```
bboymimi, line 1077: The design of the algorithm has three types of combinations:
1) left to right (contiguous), and the rightmost PD can be partial.
2) right to left (contiguous), and the leftmost PD can be partial.
3) single domain, and the domain CPU selection can be partial.

Since the design of choosing the CPU is contiguous along the PDs, there could be some situations where the optimized CPU selection jumps across the domains. For example, in the contiguous layout like: 

P | E  | LP-E

One possible optimized selected CPUs are in P and LP-E PDs. However, in this design, this combination doesn't exist.

On the other hand, this algorithm's time complexity is O(n). So, nothing to complain about. We can optimize this in the future once it's required.

bboymimi (approved):

multics69:
Thanks, Gavin, for the review.

scheds/rust/scx_lavd/src/cpu_order.rs
multics69: Yes, moreover, this is a fallback path when the # of combinations is still too large, which is unlikely to happen in practice.

PR #3727, opened 2026-08-02
scx_mitosis: Make cell management ordering predictable

dforsyth:
Tests like `test_cpu_assignments_holdout_overlapping_cpusets_no_starvation` are flaky because they expect the cell manager to have predictable behavior:

```
---- cell_manager::tests::test_cpu_assignments_holdout_overlapping_cpusets_no_starvation stdout ----

thread 'cell_manager::tests::test_cpu_assignments_holdout_overlapping_cpusets_no_starvation' (2291865) panicked at scheds/rust/scx_mitosis/src/cell_manager.rs:2487:14:
holdout must not starve a cell that shares a contested CPU: Cell 2 has no CPUs assigned (nr_cpus=5, num_cells=3)


failures:
    cell_manager::tests::test_cpu_assignments_holdout_overlapping_cpusets_no_starvation

test result: FAILED. 78 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s

error: test failed, to rerun pass `-p scx_mitosis --bin scx_mitosis`
```

I don't think there is any downside to having deterministic ordering in cell management.

This test failure exposes an issue where the cell manager may not find survivable assignments in extreme cases because of how we do allocation (https://github.com/dforsyth/scx/commit/e1b57294fc17122fa643afbd159138e2d0406854) and this change (I think) makes it consistently reproducible.

### DIFF
--- a/scheds/rust/scx_mitosis/src/cell_manager.rs
+++ b/scheds/rust/scx_mitosis/src/cell_manager.rs
@@ -9,7 +9,8 @@
 //! for direct child cgroups of a specified parent. Uses inotify to watch for
 //! cgroup creation/destruction and manages cell ID allocation.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
+use std::fs::DirEntry;
 use std::os::unix::fs::MetadataExt;
 use std::os::unix::io::{AsFd, BorrowedFd};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -781,6 +782,22 @@ fn validate_cpu_assignments(
...
 
+fn read_dir_sorted(path: &Path) -> Result<Vec<DirEntry>> {
+    let readdir = std::fs::read_dir(path)
+        .with_context(|| format!("Failed to read cell parent directory: {}", path.display()))?;
+
+    let mut entries: Vec<_> = readdir
+        .map(|entry| {
+            entry
+                .with_context(|| format!("Failed to read directory entry in: {}", path.display()))
+                .unwrap()
```
likewhatevs: maybe something like this:
```
fn read_dir_sorted(path: &Path) -> Result<Vec<DirEntry>> {
    let mut entries = std::fs::read_dir(path)
        .with_context(|| format!("Failed to read cell parent directory: {}", path.display()))?
        .collect::<std::io::Result<Vec<_>>>()
        .with_context(|| format!("Failed to read directory entry in: {}", path.display()))?;
    entries.sort_by_key(|entry| entry.path());
    Ok(entries)
}
```

or just replacing the unwrap w/ ? and an extra w/ context?

likewhatevs (approved):
LGTM minus that unwrap

kkdwvd (approved):
Fixing Patrick's concern makes sense, otherwise LGTM.

PR #3729, opened 2026-08-03
scx_lavd: reduce resident memory by 23MB

multics69:
Two independent reductions to scx_lavd's memory footprint, measured on a 16-CPU
machine with no monitoring client attached — the default way it runs.

| | RssAnon |
|---|---|
| before | 27,308 kB |
| avoid the 17.5MB sample channel unless monitoring | 9,284 kB |
| return init-time heap to the OS before running | **3,856 kB** |

VmRSS: 78,196 kB -> ~54,700 kB.

**avoid the 17.5MB sample channel unless monitoring** — `init()` unconditionally
built a `bounded(65536)` channel for scheduling samples. Each slot is a 272 byte
`SchedSample` plus crossbeam's 8 byte stamp, written up front, so all 17.5MB was
resident even though nothing reads it unless a stats client asks for samples. The
channel and ring buffer are now built per request and dropped afterwards, and the
channel is unbounded since how many samples arrive depends on load, not on the
request.

**return init-time heap to the OS before running** — libbpf's copy of the BPF
object and its BTF, the CPU topology and the energy model are all dead once the
scheduler starts, but glibc keeps the pages in its arenas. `malloc_trim(0)` after
startup releases 5.3MB. glibc only; musl has no `malloc_trim()` and needs none.

### DIFF
--- a/scheds/rust/scx_lavd/src/main.rs
+++ b/scheds/rust/scx_lavd/src/main.rs
@@ -33,7 +33,6 @@ use clap_num::number_range;
 use cpu_order::CpuOrder;
 use cpu_order::PerfCpuOrder;
 use crossbeam::channel;
-use crossbeam::channel::Receiver;
 use crossbeam::channel::RecvTimeoutError;
 use crossbeam::channel::Sender;
 use crossbeam::channel::TrySendError;
@@ -448,9 +447,7 @@ impl introspec {
 struct Scheduler<'a> {
     skel: BpfSkel<'a>,
     struct_ops: Option<libbpf_rs::Link>,
-    rb_mgr: libbpf_rs::RingBuffer<'static>,
     intrspc: introspec,
-    intrspc_rx: Receiver<SchedSample>,
     monitor_tid: Option<ThreadId>,
     stats_server: StatsServer<StatsReq, StatsRes>,
     mseq_id: u64,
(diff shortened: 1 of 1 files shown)


multics69:
Note that the additional memory-saving patch (in bpf_arena_spin_lock.h) was proposed. See https://lore.kernel.org/sched-ext/20260803001807.646357-1-changwoo@igalia.com/T/#u

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -831,10 +815,36 @@ impl<'a> Scheduler<'a> {
         }
     }
 
+    /// Collect the scheduling samples the BPF side has queued, waiting up to
+    /// @timeout for them, or draining without waiting when it is None.
+    ///
+    /// Built per request rather than kept: crossbeam initializes every slot up
+    /// front, so a standing bounded(65536) channel cost 17.5MB resident. Unbounded
+    /// because the count depends on load, not on the request; intrspc.arg caps it.
+    fn drain_sched_samples(&mut self, timeout: Option<Duration>) -> Result<Vec<SchedSample>> {
```
bboymimi, line 824: Building the channel and the ring buffer on-demand makes sense to me. It saves space compared to building it statically in init stage.

scheds/rust/scx_lavd/src/main.rs
```diff
@@ -1019,6 +1024,21 @@ impl Drop for Scheduler<'_> {
...
+/// libbpf drops its copy of the BPF object's sections and its BTF once loading
+/// finishes, and the CPU topology and energy model are dropped once the BPF side
+/// has been initialized from them. glibc holds those pages until asked.
+///
+/// glibc only: musl has no malloc_trim(), and needs none, since it returns freed
+/// memory to the kernel rather than parking it in per-arena free lists.
+fn trim_heap_after_init() {
+    #[cfg(target_env = "gnu")]
+    unsafe {
+        libc::malloc_trim(0);
```
bboymimi, line 1038: Make sense to trim the heap after initialization.

bboymimi (approved):

multics69:
Thanks @bboymimi for the review.

PR #3730, opened 2026-08-03
scx_lavd: replace deprecated ops.cpu_acquire/release with a sched_switch hook

multics69:
ops.cpu_acquire and ops.cpu_release are being deprecated. scx_lavd uses these hooks to coordinate with higher-priority scheduling classes -- in particular, cpu_release drains local-DSQ tasks stranded by such preemption.

Drop ops.cpu_acquire unconditionally and replace the cpu_release drain with a tp_btf/sched_switch hook that calls
scx_bpf_reenqueue_local() on fair -> RT/DL transitions. v1 of scx_bpf_reenqueue_local() can only be called from ops.cpu_release; v2, added in kernel 6.19, is callable from any context, including tracepoints. This gives two tiers, selected at init:

  - kernel >= 6.19: tp_btf/sched_switch hook calls
                  scx_bpf_reenqueue_local() (v2) on fair -> RT/DL
                  transitions; ops.cpu_release is dropped.
  - kernel < 6.19:  keep ops.cpu_release as the legacy drain;
                  v1 cannot be called from a tracepoint, so
                  sched_switch is not an option here.

Detected via ksym_exists("scx_bpf_reenqueue_local___v2"). The sched_switch hot path is one struct-field load + one signed compare against MAX_RT_PRIO.

Reenqueued tasks (SCX_ENQ_REENQ) take a fast path in lavd_enqueue() that reuses the cached CPU pick, recomputes the task's deadline, and reinserts into the previously chosen cpdom DSQ. The deadline is recomputed because the logical clock may have advanced while the task waited behind the higher class; reusing the stale, smaller deadline would over-prioritize it.


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1995,53 +2025,50 @@ void BPF_STRUCT_OPS(lavd_set_cpumask, struct task_struct *p,
...
+		/*
+		 * Reenqueue any SCX tasks stranded on the local DSQ.
+		 * Use the v2 (call-from-anywhere) kfunc directly rather
+		 * than the generic compat wrapper: the wrapper inlines a
+		 * v1 fallback that is not callable from a tracepoint
+		 * context, and veristat (which doesn't honor autoload)
+		 * would reject it on kernels without v2.
+		 */
+		if (bpf_ksym_exists(scx_bpf_reenqueue_local___v2___compat))
+			scx_bpf_reenqueue_local___v2___compat();
```
arighi, line 2083: Not necessarily for this PR, but I was wondering if we should introduce a compat helper that references only v2, like scx_bpf_reenqueue_local_from_anywhere() or similar. It'd probably make the code more readable.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -909,22 +909,51 @@ void BPF_STRUCT_OPS(lavd_enqueue, struct task_struct *p, u64 enq_flags)
...
+		 * while the task was bounced, so reusing the stale (smaller)
+		 * deadline would over-prioritize it. Re-anchor to the current
+		 * clock.
+		 */
+		p->scx.dsq_vtime = calc_when_to_run(p, taskc);
+
+		dsq_id = get_target_dsq_id(p, cpuc, taskc);
+		scx_bpf_dsq_insert_vtime(p, dsq_id, p->scx.slice,
+					 p->scx.dsq_vtime, enq_flags);
+		goto kick_cpu_out;
```
arighi, line 959: This fast path skips the cgroup_throttled() check below; lavd_stopping() accounts the runtime consumed before an RT/DL preemption, so that accounting may have just exhausted the task’s cpu.max quota. In that case, the reenqueued task should be moved to the throttle queue, but this branch inserts it directly into a DSQ. Is this intentional or should we run the bandwidth check before re-inserting?

Moreover, are we also breaking the queued-load accounting? The previously running task was unaccounted by lavd_running(), so after it is preempted and reenqueued this path never adds its domain load back.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -1995,53 +2025,50 @@ void BPF_STRUCT_OPS(lavd_set_cpumask, struct task_struct *p,
...
+		 * Use the v2 (call-from-anywhere) kfunc directly rather
+		 * than the generic compat wrapper: the wrapper inlines a
+		 * v1 fallback that is not callable from a tracepoint
+		 * context, and veristat (which doesn't honor autoload)
+		 * would reject it on kernels without v2.
+		 */
+		if (bpf_ksym_exists(scx_bpf_reenqueue_local___v2___compat))
+			scx_bpf_reenqueue_local___v2___compat();
+	}
+	return 0;
```
arighi, line 2085: The old cpu_release() path also marked this CPU as unavailable for LAVD preemption by clearing flags, setting lat_cri = SCX_SLICE_INF, and est_stopping_clk = 0. This hook only drains the local DSQ, so try_find_and_kick_victim_cpu() can still select an RT/DL-owned CPU using stale state. Should we restore that release-state reset here before reenqueuing?

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: That's a good idea. I will submit another PR after landing this one.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Not checking cgroup_throttled() in this fast path was intentional because it was checked in the previous enqueue. But, as you point out, the cgroup could be throttled in between. Adding cgroup_throttled() checking will improve the accuracy of bandwidth control. Will add it.

Yes, queued-load accounting needs to be added.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: Actually, I implemented the same thing differently by directly checking whether a victim CPU is running an RT/DL task in the following PR. But, it will be better to include that part here to make this PR self-contained. Thanks!

multics69:
Thanks, @arighi, for the review. I addressed all your comments and force-pushed.

arighi (approved):

PR #3731, opened 2026-08-04
scx: add scx_bpf_reenqueue_local_from_anywhere() compat helper

multics69:
scx_bpf_reenqueue_local()'s generic compat wrapper inlines a v1 fallback that is only callable from ops.cpu_release, and veristat rejects it on kernels without v2. Callers draining a local DSQ from an arbitrary context (e.g. a tracepoint) must gate on the v2 kfunc directly.

Add scx_bpf_reenqueue_local_from_anywhere() to compat.bpf.h to encapsulate that: call the v2 kfunc when present and return 0, else return -ENOTSUP so the caller knows the drain did not run.

Update lavd_sched_switch() to use the helper in place of its open-coded ksym check.

No functional change intended.

### DIFF
--- a/scheds/include/scx/compat.bpf.h
+++ b/scheds/include/scx/compat.bpf.h
@@ -6,6 +6,7 @@
  */
 #ifndef __SCX_COMPAT_BPF_H
 #define __SCX_COMPAT_BPF_H
+#include <errno.h>
 
 #define __COMPAT_ENUM_OR_ZERO(__type, __ent)					\
 ({										\
@@ -391,6 +392,15 @@ static inline void scx_bpf_reenqueue_local(void)
 		scx_bpf_reenqueue_local___v2___compat();
 	else
 		scx_bpf_reenqueue_local___v1();
+}
+
+static inline int scx_bpf_reenqueue_local_from_anywhere(void)
+{
+	if (__COMPAT_scx_bpf_reenqueue_local_from_anywhere()) {
+		scx_bpf_reenqueue_local___v2___compat();
+		return 0;
+	}
+	return -ENOTSUP;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


arighi (approved):
LGTM, thanks.

bboymimi (approved):

multics69:
Thanks @arighi and @bboymimi !

htejun:
A belated thought: this hard-wires the drain to the v2 kfunc, which kernel-side is now just a wrapper around scx_bpf_dsq_reenq() and may be removed down the line. Both this helper and scx_bpf_reenqueue_local() would then silently regress. How about reshaping both to prefer the generic kfunc?

```c
static inline void scx_bpf_reenqueue_local(void)
{
	if (__COMPAT_has_generic_reenq())
		scx_bpf_dsq_reenq___compat(SCX_DSQ_LOCAL, 0, NULL);
	else if (__COMPAT_scx_bpf_reenqueue_local_from_anywhere())
		scx_bpf_reenqueue_local___v2___compat();
	else
		scx_bpf_reenqueue_local___v1();
}

static inline int scx_bpf_reenqueue_local_from_anywhere(void)
{
	if (__COMPAT_has_generic_reenq() ||
	    __COMPAT_scx_bpf_reenqueue_local_from_anywhere()) {
		scx_bpf_reenqueue_local();
		return 0;
	}
	return -ENOTSUP;
}
```

veristat stays happy: when neither newer kfunc exists, the gate is compile-time false and the whole block including the inlined v1 arm is dead-code-eliminated.

Also, compat.bpf.h is kept in sync with the kernel's tools/sched_ext/include/scx/compat.bpf.h. Can you route this through the kernel tree and then sync the scx copy to it?

PR #3732, opened 2026-08-07
scx_mitosis: Simplify topology preparation

dforsyth:
There is a lot of unused code in `mitosis_topology_utils`, and it is also hard to test.

This change adds a top level `MitosisTopology` type that we can use for all of our init, with a testable apply mechanism that will write the topology into bss. The way things are put together is mostly the same, except this uses the Cpumask that is already on the topo LLC struct instead of building one from scratch.

https://github.com/dforsyth/scx/commit/407afc96771da6a2a608f1c80b80c808ef9f4403 has a test to verify old vs new behavior.


kkdwvd (approved):
I would also just rename the file to topology.rs, the current name is overtly verbose, but I don't have strong feelings.

PR #3733, opened 2026-08-07
scx_mlfq 1.0.0

galpt:
## About

A three-queue MLFQ scheduler with per-queue EEVDF virtual time. An EMA
interactivity gauge classifies tasks into interactive, unclassified and
CPU-bound queues, with wakeup preemption across queues, per-CPU
vtime-ordered dispatch queues with cross-CPU stealing, and aging and
demotion following the MLFQ rules. Handles SCHED_NORMAL, SCHED_BATCH
and SCHED_IDLE tasks; real-time tasks remain with the kernel classes.

## Summary

The first commit adds the scheduler: the three-queue MLFQ structure
with the EMA interactivity gauge, vtime-ordered per-CPU dispatch
queues, wakeup preemption across queues, and cross-CPU stealing.

The second commit reworks placement and dispatch around a lock-free
per-queue virtual clock, removing the per-queue spinlock from the hot
path, and fixes the multi-node idle path and the watchdog timeout
under real-time load.

The third commit makes the CPU selection boost-aware, so interactive
wakeups get the primary-core preference before their promotion, and
derives the cpuperf target for the other queues from the CPU's recent
activity so the frequency follows the load.

## Changes

- [`b5c87de7`](https://github.com/galpt/scx/commit/b5c87de7) **scx_mlfq: add a multilevel feedback queue scheduler**
- [`c88bbd3a`](https://github.com/galpt/scx/commit/c88bbd3a) **scx_mlfq: replace the locked aggregate with a lock-free virtual clock**
- [`1d21bfe`](https://github.com/galpt/scx/commit/1d21bfe) **scx_mlfq: make the CPU selection boost-aware and follow the load with cpuperf**


scheds/experimental/scx_mlfq/src/topology.rs
```diff
@@ -0,0 +1,454 @@
...
+    } else {
+        let mut cpus = primary_cpus.to_vec();
+        cpus.sort_unstable();
+        cpus.dedup();
+        CapacityPlan {
+            primary_all: false,
+            primary_cpus: cpus,
+        }
+    }
+}
```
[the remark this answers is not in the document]
galpt: The duplicate scenario cannot occur. `get_primary_cpus()` walks the topology map once per CPU, so each online CPU appears at most once, and `nr_online` comes from the same map. With duplicates impossible, `len() >= nr_online` means every online CPU is primary, which is exactly the uniform-capacity case this branch handles. The sort and dedup remain as defensive hygiene.

scheds/experimental/scx_mlfq/src/bpf/enqueue.bpf.c
```diff
@@ -0,0 +1,392 @@
...
+		    prev_state->running_pid != p->pid &&
+		    prev_state->running_queue > 0 &&
+		    (s32)qid < prev_state->running_queue &&
+		    bpf_cpumask_test_cpu((u32)prev_cpu, p->cpus_ptr)) {
+			scx_bpf_dsq_insert(p, SCX_DSQ_LOCAL_ON | (u64)prev_cpu,
+					   slice, enq_flags | SCX_ENQ_PREEMPT);
+			__sync_fetch_and_add(&mlfq_stats.preemption_kicks, 1);
+			tctx->wake_cpu_state = 0;
+			goto done;
+		}
```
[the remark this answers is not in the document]
galpt: Placement is a pure function of the queue aggregate, the task virtual runtime and weight, recomputed at every enqueue, and the preemption path never accounts the task into the aggregate. The task deadline is only written between placements, so nothing stale is read. The only difference from the placement-taking fast paths arises when the lag is pinned at the clamp, within the `DELAY_ZERO`-style approximation documented in the README. Skipping the placement keeps the highest-frequency latency path lock-free, and the next enqueue recomputes the same state up to the clamp.

scheds/experimental/scx_mlfq/src/bpf/enqueue.bpf.c
```diff
@@ -0,0 +1,353 @@
...
+	 * previous stay because the task just ran a full slice. A queue
+	 * change also clears the consecutive slice-exhaustion counter (band
+	 * semantics). Only Q2/Q3 stays carry a non-zero queued_at.
+	 */
+	if (tctx->queue != old_queue) {
+		tctx->queued_at = tctx->queue >= 2 ? now : 0;
+		tctx->reenq_cnt = 0;
+	} else if (wakeup || runout) {
+		tctx->queued_at = tctx->queue >= 2 ? now : 0;
+	}
```
arighi, line 162: A task waiting in Q2/Q3 DSQ has no callback while it waits, so it can never be boosted after one second. Maybe we need an aging check during dispatch, before the task is selected?

scheds/experimental/scx_mlfq/src/bpf/select_cpu.bpf.c
```diff
@@ -0,0 +1,235 @@
...
+	 * sticks to prev when it is a primary core: settling an interactive
+	 * wakeup on an efficiency core would trade cache warmth for
+	 * sustained capacity. The idle mark is cleared only after the
+	 * primary check passes, so an idle efficiency core is never lost
+	 * from the idle pool for a Q1 wakeup.
+	 */
+	if ((tctx->queue != 1 || mlfq_is_primary(primary_bm, prev_cpu)) &&
+	    scx_bpf_test_and_clear_cpu_idle(prev_cpu)) {
+		cpu_id = prev_cpu;
+		goto direct;
```
arighi, line 207: Short-sleep promotion to Q1 happens later in ops.enqueue(), so a prior-Q3 task that wakes interactively can claim its previous idle E-core, then be promoted and inserted directly there. Does this contradicts the stated Q1 "never stick to an efficiency core" policy?

scheds/experimental/scx_mlfq/src/bpf/lifecycle.bpf.c
```diff
@@ -0,0 +1,183 @@
...
+	 * stores the target per CPU and it persists until overwritten, so
+	 * setting it on every ops.running() makes the last value always
+	 * reflect the interactive task now on the CPU, and schedutil then
+	 * picks the frequency. The other queues leave the target untouched
+	 * so the kernel's governor drives the frequency without
+	 * throttling. The counter reports Q1 placements as the interactive
+	 * boost signal.
+	 */
+	if (tctx->queue == 1) {
+		scx_bpf_cpuperf_set(scx_bpf_task_cpu(p), MLFQ_CPUPERF_Q1);
```
arighi, line 99: IIUC the CPU boost is never cleared, so once a CPU sets MLFQ_CPUPERF_Q1 the cpufreq boost is never cleared (Q2/Q3 do nothing). Meaning a CPU that once ran an interactive task always remains at max performance, even if it's running background work?

arighi (approved):
I've done a quick review, left few comments, but overall looks good to me. I think we can also fix/address my comments in tree later.

scheds/experimental/scx_mlfq/src/bpf/enqueue.bpf.c
galpt: Thanks for the note. No change needed here, the aging fires at the next re-enqueue within one slice of the one second mark, and the Q3 share of every dispatch batch is the actual starvation bound. Let me know if that still looks off.

scheds/experimental/scx_mlfq/src/bpf/select_cpu.bpf.c
galpt: Thanks! Addressed in commit 1d21bfe, the selection now applies the primary-core preference when the short-sleep boost is about to fire. Let me know if I still miss something.

scheds/experimental/scx_mlfq/src/bpf/lifecycle.bpf.c
galpt: Thanks! Addressed in commit 1d21bfe, the other queues now set the target from the CPU's recent activity so the frequency follows the load. Let me know if I still miss something.

sirlucjan (approved):
Good job!

PR #3735, opened 2026-08-10
scx_lavd: restrict direct dispatch to usable, uncongested CPUs

multics69:
`can_direct_dispatch()`'s low-utilization bypass decides purely on average
utilization, so it can drop a task into the (non-stealable) local DSQ of
a CPU where it will not run promptly. Refuse the bypass in two such cases.

  - a CPU running an RT/DL task, unusable to scx until the higher class yields;
  - a congested CPU, where already-queued work would be jumped ahead of.

Also carries a no-functional-change prep: pass cpdom_id to
try_find_and_kick_victim_cpu() directly instead of re-deriving it.

Tested the change with the memtier benchmark on a server with two AMD
EPYC 9R45 (96-core x2). Overall, this PR lowers the tail latency while
providing a similar level of throughput, especially when system
utilization is low.

| Utilization | Metric  | baseline (main) | this PR | delta       |
| ----------- | ------- | --------------: | ------: | ----------- |
| 50%         | ops/sec |         509,056 | 509,588 | 0.1% higher |
| 50%         | avg_lat |            0.45 |    0.46 | 2.4% higher |
| 50%         | p99.9   |            7.08 |    6.93 | 2.2% lower  |
| 25%         | ops/sec |         251,971 | 251,995 | 0.0% higher |
| 25%         | avg_lat |            0.44 |    0.43 | 0.3% lower  |
| 25%         | p99.9   |            5.21 |    4.29 | 17.7% lower |

- [1/3] scx_lavd: pass cpdom_id to try_find_and_kick_victim_cpu()
- [2/3] scx_lavd: avoid CPUs taken by RT/DL tasks for direct dispatch
- [3/3] scx_lavd: do not direct-dispatch to a congested CPU

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -1114,16 +1114,11 @@ kick_cpu_out:
 	}
 
 	/*
-	 * If there is no idle CPU for an eligible task, try to preempt a task.
-	 * Try to find and kick a victim CPU, which runs a less urgent task,
-	 * from dsq_id. The kick will be done asynchronously.
-	 *
-	 * In the case of the forced enqueue mode, we don't try preemption
-	 * since it is a batch of bulk enqueues.
+	 * If there is no idle CPU, try to preempt a task. Find and kick a
+	 * victim CPU, which runs a less urgent task.
 	 */
 	if (!no_preemption) {
-		try_find_and_kick_victim_cpu(p, taskc, cpu,
-					     cpdom_to_dsq(cpuc->cpdom_id));
+		try_find_and_kick_victim_cpu(p, taskc, cpu, cpuc->cpdom_id);
 	}
 }
 
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -677,7 +677,8 @@ static bool can_direct_dispatch(struct cpu_ctx *cpuc, bool is_cpu_idle)
 {
 	return (is_cpu_idle && !queued_on_cpu(cpuc)) ||
 	       (lb_local_dsq_util_wall > 0 &&
-		cpuc->avg_util_wall < lb_local_dsq_util_wall);
+		cpuc->avg_util_wall < lb_local_dsq_util_wall &&
+		!is_rt_or_dl_task_running(cpuc->cpu_id));
 }
 
 /*
--- a/scheds/rust/scx_lavd/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_lavd/src/bpf/main.bpf.c
@@ -675,10 +675,27 @@ static void update_stat_for_refill(struct task_struct *p,
 
 static bool can_direct_dispatch(struct cpu_ctx *cpuc, bool is_cpu_idle)
 {
-	return (is_cpu_idle && !queued_on_cpu(cpuc)) ||
-	       (lb_local_dsq_util_wall > 0 &&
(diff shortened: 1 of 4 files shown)


scheds/rust/scx_lavd/src/bpf/main.bpf.c
```diff
@@ -675,9 +675,27 @@ static void update_stat_for_refill(struct task_struct *p,
...
+	 * across its DSQs, and inserting into the local DSQ would let the
+	 * new task jump ahead of them) nor into a CPU an RT/DL task has
+	 * taken (the task would be stranded in a non-stealable local DSQ
+	 * until the higher class yields). Both walk/peek remote state, so
+	 * evaluate them last, only after the cheap utilization checks pass.
+	 */
+	return lb_local_dsq_util_wall > 0 &&
+	       cpuc->avg_util_wall < lb_local_dsq_util_wall &&
+	       !is_cpu_congested(cpuc) &&
+	       !is_rt_or_dl_task_running(cpuc->cpu_id);
```
bboymimi, line 698: I was thinking another case could be worth a revisit—the lock holder condition. It seems that we currently don't have any restriction on the lock holder, which cannot be preempted. If the lock holder runs for a long time, it could impact the tasks waiting in the DSQs. And if it's a real bug (meaning that the lock holder runs indefinitely), it could trigger a soft lockup or other errors. I think I was overthinking the problem. Overall, it looks good to me.

scheds/rust/scx_lavd/src/bpf/util.bpf.c
```diff
@@ -411,6 +411,34 @@ bool queued_on_cpu(struct cpu_ctx *cpuc)
 	return false;
 }
 
+__hidden
+bool is_cpu_congested(struct cpu_ctx *cpuc)
```
bboymimi, line 415: I was gazing at the function for a while and comparing it with the queued_on_cpu. When LAVD_CPU_CONGESTED_THRES == 1, it appears to me that the logic is almost the same except for the order of the if condition. I think the function is paving the road for future enhancements for possible longer congestions with different scenarios. There could be a case where a higher-priority task (or any kind of urgency) is chiming in and jumping the queue based on the congestion length.

bboymimi (approved):

daidavid:
I'm curious what the benchmark numbers are if we only allow dispatch on idle. I think using util is an incorrect metric to determine when to direct dispatch. Util doesn't provide any info on the duty_cycle or wake up timings, so low util tasks that match up wake up timings will result in contention anyways. I'm okay with just disabling the util bypass unless there's some benefit there.

scheds/rust/scx_lavd/src/bpf/main.bpf.c
multics69: I think avoiding a CPU that holds a lock for direct dispatch makes sense. I will experiment it.

multics69:
> I'm curious what the benchmark numbers are if we only allow dispatch on idle. I think using util is an incorrect metric to determine when to direct dispatch. Util doesn't provide any info on the duty_cycle or wake up timings, so low util tasks that match up wake up timings will result in contention anyways. I'm okay with just disabling the util bypass unless there's some benefit there.

I agree that the util is not quite the right metric to decide the direct dispatch. In my experience, the direct dispatch is a double-edged sword; it helps improve the throughput by reducing the DSQ overhead, but when it is done at the wrong moment, it significantly increases the latency. So far, the util metrics have been used to opportunistically increase the chance of direct dispatch. I will revisit this once again.

multics69:
I will revist the utilization based direct dispatch and potentially replace it to the completion time based approach after merging this one.

PR #3736, opened 2026-08-13
scx_cosmos: key GPU hints by TGID

nvax-r:
## Summary

NVML reports GPU users as process IDs, but `scx_cosmos` currently expands each process into every TID and stores every thread in `gpu_pid_map`

The commit basically do the following things
* Stores one TGID-to-NUMA-node entry per GPU process.
* Looks up GPU locality using `task_struct::tgid`
* Removes `/proc/<tgid>/task` enumeration.
* Automatically covers threads created after synchronization.
* Preserves the existing process-scoped GPU policy.

## Performance comparison
I compared the before/after performance with a CUDA process with controlled parked threads.

| Process tasks | Map Entries before | Map Entries after | Updates per sync before | Updates per sync after | Reduction |
| -------- | -------- | -------- | -------- | -------- | -------- |
| 4 | 4 | 1 | 4 | 1 | 75% |
| 64 | 64 | 1 | 64 | 1 | 98.44% |
| 256 | 256 | 1 | 256 | 1 | 99.61% |
| 512 | 512 | 1 | 512 | 1 | 99.80% |
| 1024 | 1024 | 1 | 1024 | 1 | 99.90% |

For a 492-thread CUDA process, 8 balanced paired trials produced:
- scx_cosmos CPU time: 1.493 → 0.619 ms/s
- Mean absolute saving: 0.874 ms/s
- Mean relative reduction: 58.3%
- Candidate improved all 8/8 pairs

An independent `perf task-clock` measurement showed a similar 59.0% reduction.

Note that these numbers measure the `scx_cosmos` userspace control-plane overhead, not application runtime.

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -268,31 +268,31 @@ struct {
 } gpu_node_map SEC(".maps");
 
 /*
- * PID -> NUMA node mapping for GPU tasks (updated from userspace via NVML).
- * Key is the task's pid (thread id). Entries are removed when the task
- * is no longer using a GPU.
+ * Process TGID -> NUMA node mapping for GPU tasks (updated from userspace via
+ * NVML). Entries are removed when the process is no longer using a GPU.
  */
 #define MAX_GPU_PIDS	8192
 
 struct {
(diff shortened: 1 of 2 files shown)


arighi (approved):
This looks like a sane change. Thanks!

PR #3740, opened 2026-08-16
scx_mitosis: Prepare for subcell support

kkdwvd:
NFC changes which add all basic infrastructure to prepare for splitting cells into subcells. See commit logs for details.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
@@ -101,30 +101,31 @@ static inline struct subcell_llc *lookup_subcell_llc(struct subcell *subcell, u3
  */
 struct task_ctx {
 	/* cpumask is the set of valid cpus this task can schedule on */
-	/* (task's cpumask and-ed with its cell cpumask) */
+	/* (task's cpumask and-ed with its subcell cpumask) */
 	struct bpf_cpumask __kptr *cpumask;
 	/* Cached intersection of cpumask and the selected LLC's cpumask */
 	struct bpf_cpumask __kptr *llc_cpumask;
 	/* started_running_at for recording runtime */
 	u64 started_running_at;
-	/* Cell whose vtime domain should be charged for this task */
-	u32 vtime_charge_cell;
+	/* Packed subcell whose vtime should be charged for this task. */
+	u32 vtime_charge_subcell;
 	u64 basis_vtime;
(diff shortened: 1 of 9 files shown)


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
```diff
@@ -55,6 +55,47 @@ static inline struct cell *lookup_cell(int idx)
...
+	struct subcell *subcell;
+
+	if (subcell_idx >= MAX_SUBCELLS_PER_CELL) {
+		scx_bpf_error("invalid subcell %u for cell %u", subcell_idx, cell_idx);
+		return NULL;
+	}
+
+	cell = lookup_cell(cell_idx);
+	if (!cell)
+		return NULL;
```
dforsyth, line 70: scx_bpf_error? I'm not sure we'd ever expect this lookup to fail?

scheds/rust/scx_mitosis/src/cell_manager.rs
```diff
@@ -39,113 +39,114 @@ pub struct CellInfo {
...
-        // Standard proportional distribution
-        let mut assigned = 0usize;
-        let mut raw: Vec<(u32, f64, usize, bool)> = recipients
+    /// Compute the global target CPU count for each recipient.
+    ///
+    /// Each recipient gets a floor of 1 CPU, with the remainder distributed
+    /// proportionally by weight. Returns only counts, not actual CPU assignments.
+    fn compute_targets(
+        total_cpus: usize,
+        recipients: &[(u32, f64)],
```
dforsyth, line 125: I realize that a lot of this is copy/paste churn, but it would be nice if we didn't pass around/return tuples like this. This change is already making a lot of noise, so I wouldn't be opposed to it happening here.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.h
kkdwvd: lookup_cell() already has scx_bpf_error() internally.

scheds/rust/scx_mitosis/src/cell_manager.rs
kkdwvd: Yeah, I think that sort of cleanup is better done as a follow up. The diff would be too confusing otherwise (it's too big already).

dforsyth (approved):

kkdwvd:
Landing. Will fix forward as necessary when rubber hits the road.

PR #3741, opened 2026-08-17
scx_layered: Make init_task sleepable

hodgesds:
In certain cases with heavy thread/process allocations init_task can fail when allocating the task_ctx. Make init_task sleepable so that allocations go through GFP_KERNEL compared to GFP_ATOMIC.

### DIFF
--- a/scheds/rust/scx_layered/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_layered/src/bpf/main.bpf.c
@@ -3890,10 +3890,15 @@ static int init_cached_cpus(struct cached_cpus *ccpus)
 static int maybe_init_task_unprotected_mask(struct task_struct *p, struct task_ctx *taskc)
 {
 	struct bpf_cpumask *cpumask;
+	bool full;
 	int ret;
 
+	bpf_rcu_read_lock();
+	full = bpf_cpumask_full(p->cpus_ptr);
+	bpf_rcu_read_unlock();
+
 	/* We don't need our own mask, we have no placement restrictions. */
-	if (bpf_cpumask_full(p->cpus_ptr))
+	if (full)
 		return 0;
 
 	/* Already initialized. */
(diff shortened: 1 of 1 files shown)


htejun (approved):

scheds/rust/scx_layered/src/bpf/main.bpf.c
```diff
@@ -3890,10 +3890,15 @@ static int init_cached_cpus(struct cached_cpus *ccpus)
 static int maybe_init_task_unprotected_mask(struct task_struct *p, struct task_ctx *taskc)
 {
 	struct bpf_cpumask *cpumask;
+	bool full;
 	int ret;
 
+	bpf_rcu_read_lock();
+	full = bpf_cpumask_full(p->cpus_ptr);
+	bpf_rcu_read_unlock();
```
htejun, line 3898: Hmmm... `bpf_cpumask_full()` is `KF_RCU` which means that the argument can either be trusted or rcu protected. I wonder why the explicit rcu locking is necessary. Anyways, whatever works but this probably needs some probing.

hodgesds: Yeah, that was just to get around the verifier.

kkdwvd: Sleepable programs do not have RCU read protection (instead, they have RCU tasks trace protection, which allows sleeping in its critical section), so an explicit read section inside the program is necessary.

htejun: But doesn't `KF_RCU` mean that the input parameter is trusted or rcu protected? The kfunc is not marked `KF_RCU_PROTECTED`, just `KF_RCU`.

kkdwvd: That part is right, but I think `cpus_ptr` in `task_struct` is annotated as `BTF_TYPE_SAFE_RCU` (in verifier.c), which means on dereference to load the pointer, the RCU read section should be open, and remain open while that pointer is passed to anything expecting RCU protected pointers.

kkdwvd: Otherwise the load does not obtain an RCU protected pointer, but an untrusted pointer instead.

htejun: So, it's `p->cpus_ptr` that needs RCU protection, not `bpf_cpumask_full()`. Yeah, that makes sense. Probably worth a comment.

PR #3746, opened 2026-08-18
scx_pandemoniumv5.18.0:

wllclngn:
"The lighting effects beggared description. The whole country was lighted by a searing light with the intensity many times that of the midday sun... Thirty seconds after the explosion came first, the air blast pressing hard against the people and things, to be followed almost immediately by the strong, sustained, awesome roar which warned of doomsday."
— Brigadier General Thomas F. Farrell, quoted in Gen. Leslie R. Groves,
  "Memorandum for the Secretary of War: The Test," 18 July 1945

This release opens on a field report. A user ran PRISM on hardware this project does not own, a 20-core Intel Core Ultra 7 265 on 7.1.5-1-cachyos, and sent back both a report and two complaints: sublimation had trouble installing, and the uninstall option did not uninstall. Both turned out to be one defect wearing two faces, and neither could have been found here, because every machine in this project already has the binaries installed permanently. That is the whole lesson: An install path is only tested by a system that does not already have what it installs.

SCHEDULER, THE WARP IS BOUNDED BY THE MEASUREMENT IT CLAIMS TO USE: src/bpf/main.bpf.c
- The dispatch key called itself a sojourn selector and was neither bounded by sojourn nor by CoDel. warp = lag_cap_ns * potentiality, and lag_cap_ns is one commute time of the topology graph: 40ms at the 12C reference, against a live CoDel band of 700us to 2ms. The ordering bound was twenty to fifty-seven times the entire band it was supposed to modulate, at every topology, by construction. Inside the band the key was a static tier priority and sojourn ordered nothing beneath it.
- Worse, the ordering bound and the starvation bound were the same variable. A task could back-date itself by exactly as long as the system tolerates anyone waiting, so a warp could consume the whole starvation budget. They are now separate numbers: the warp is bounded by codel_target_ns, and lag_cap_ns keeps its real job as the age at which sweep_bound_preempt forces a head off its CPU.
- The warp is now the share of one live CoDel target the task left unconsumed on its last run: codel_target_ns - last_run_ns, floored at zero. A task that blocked immediately earns a full target, one that held the CPU for a target or longer earns nothing, and everything between is continuous. One measured quantity, priced in the one unit the gate already maintains.
- task_potentiality_q16 is gone entirely, and with it both runtime EWMAs, the tier read, the maturity gate and the fresh-fork prime. Ordering takes no classifier input at all now, which matters more than it sounds: whatever the tier classifier turns out to be, it can no longer reach the sort key.
- THE BASE WAS BEING RESTAMPED ON EVERY REQUEUE, which is the older half of the same defect. The key read now at each insert, so a task passed over N times was re-stamped to the back N times and its aging was erased on every pass. A vtime DSQ sorts by the key at insert and never re-sorts, so a stationary base under a moving clock is sojourn ordering at no recompute cost: wait_since is stamped once, preserved across requeues, and released in running(). quiescent() clears it too -- a task queued, dequeued unrun and then slept would otherwise return holding a seconds-old claim and sort ahead of the machine.
- Measured on the IPC bench at 2/4/8/12C, both arms, two samples: tick-floored wakes fell from 4.4-7.6% to 1.0-3.5%, wake2run p99 from 1006-1637us to 870-1358us, and the floored-wake count by 55-80%. Eight of eight BPF cells moved the right way. Order-starvation composition did not change, which is the expected shape: this is the bound, not the term.

SCHEDULER, EVERY KNOB IS NOW DERIVED AND NONE ARE LEARNED: src/adaptive.rs, src/tuning.rs, src/scheduler.rs, src/bpf/intf.h, src/bpf/main.bpf.c
- The adaptive layer could not see the machine it was steering. read_stats() called lookup_percpu, which hands back one stats struct per CPU, and summed every field into a single scalar set inside the loop that first touched it -- roughly 600 numbers collapsed to 30 on a 20-core box, with the spatial dimension destroyed at the boundary. What reached the controller each tick was about nine scalars, and the entire chaos layer ran over 16 samples of ONE system-wide integer percentage while BPF saw every dispatch decision per-CPU at nanosecond resolution. The loop was competing with the layer feeding it, on strictly less information, one second late.
- The array now survives. It costs nothing: the syscall, the copy and the cache traffic were always paid, and userspace was discarding the structure on arrival. read_stats() is unchanged for every existing caller, folded from the array so the aggregate is derived rather than computed beside it.
- BPF exports per-CPU runnable depth, sampled at tick rate, sum and samples in the shape wake_lat_sum/wake_lat_samples already used. Deliberately not gated to CPU 0: the spatial dimension is the point. The knob map became per-CPU with max_entries still 1, so each CPU reads its own slot and all thirteen BPF read sites are untouched. Six fields that must not diverge -- tau, codel_eq, spill_temp, affinity and the two classifier thresholds -- are broadcast from slot 0 by construction rather than by convention.
- THE LIVE LOAD GRAPH is what the userspace loop now exists to build, because it is the half of the boundary BPF structurally cannot cross: relating CPU i to CPU j means reading another CPU's state, which per-CPU maps exist to avoid, and the arithmetic wants f64 and unbounded loops. Nodes are CPUs weighted by mean depth, traffic shape, critical slowing and persistence; edges are CPU pairs weighted by coupling. The chip's electrical graph is a constant and R_eff already prices against it -- this is the graph the WORKLOAD forms, which moves every second and which nothing measured.
- EVERY KNOB IS NOW A FUNCTION OF A MEASUREMENT. Depth sets the per-CPU slice, a deep queue slicing shorter so it drains and a shallow one longer so it stops paying context-switch cost for contention that is not there. Critical slowing tightens the preempt window AHEAD of the burst, which is the only thing in this loop that acts on a prediction rather than a completed loss. Traffic shape sets the batch and burst ceilings. Persistence sets the rescue threshold, since a queue that is deep on a persistent CPU will still be deep. Coupling is measured but does not actuate -- see the withdrawal below.
- THE MWU ORCHESTRATOR IS GONE, 728 lines. It searched an eleven-dimensional space with one loss signal because it had no model; every dimension it searched now has one. The asymmetry is decisive rather than aesthetic: a derived knob is correct on the tick it is computed, where a learned one is correct several convergence windows later and only if the regime holds still that long. The archive agreed in advance -- across 18,710 adaptive-vs-BPF comparisons the layer was a coin flip, 49.1% of non-negligible effects, and 10-18% at the only well-powered cells.
- THE REGIME ENUM IS GONE with it, and the audit that killed it is worth recording: across Light, Mixed and Heavy, lat_cri_thresh_high, lat_cri_ thresh_low, codel_thresh_ns, burst_slice_ns and spill_temp_q16 were IDENTICAL. They were never regime-dependent; they were defaults wearing a selector. Of what genuinely varied, slice, preempt and batch are now measured, leaving affinity_mode as the enum's entire surviving contribution -- 1.58 bits resolving to a single binary. Three profiles collapse to one base the derivations move from, and affinity holds its base value there.
- The safety property changed and the test says so rather than papering over it. With only the depth derivation an unstructured machine came out exactly where it went in. With four sensors that is no longer true and should not be: a machine with no SPATIAL structure can still have TEMPORAL structure, and every CPU slowing critically at once is a real state worth acting on. What must still hold is that identical CPUs derive identical knobs, since divergence without structure would mean a sensor is reading noise. That is the assertion.
- ONE DERIVATION WAS WITHDRAWN THE SAME DAY IT SHIPPED, AND THE REASON IS THE POINT. Coupling briefly set affinity strength, on the argument that runqueues moving together mean tasks sharing data. That was reasoning, not measurement. The first full PRISM pass on this tree put a number against it: baseline_gate reported every IPC primitive regressed on every arm, worst pipe p50 20 -> 336us at drift 1.04x, which is 16x with the environment ruled out. IPC is precisely a two-task ping-pong with tightly coupled queue depths, so the derivation fired hardest exactly where the damage landed. Affinity keeps its base value until something measures the direction; the coupling reading stays as telemetry, because the reading was never the problem. Every other derivation on this list has a measured mechanism behind it and this one had an argument, which is the distinction worth keeping.
- The double-correction hazard retired itself. MWU read the BPF oscillator's state to avoid two controllers both adapting on global_rescue_count. There is now one controller on that signal; the graph derives from depth, shape and persistence instead.

SCHEDULER, A PREDICTIVE SENSOR AND A SHORT-WINDOW-HONEST ONE: src/chaos.rs
- Lag-1 autocorrelation is the first primitive in this module that LEADS. HVG, Bandt-Pompe and RQA all describe the window that already closed; critical slowing down rises before the transition it precedes, because a system nearing saturation recovers from perturbations more slowly and that shows up as successive samples resembling each other. On a runqueue-length series it is the difference between watching a burst form and recording that one happened.
- The biased estimator is used deliberately. It is the lower-variance of the two, and on windows this short variance is the entire problem -- an unbiased estimate that swings run to run is not usable as a lead signal.
- A flat window returns nothing rather than 1.0. A perfectly steady series has no perturbation to recover from and therefore carries no critical-slowing information at all, and an idle CPU reading as maximally autocorrelated, which is to say about to tip, would be the most common false positive available.
- Kim-Jo finite-size-corrected burstiness replaces a statistic that could not be used here. The classical Goh-Barabasi parameter drifts toward -1 as the sample count falls, so on a short window it reads REGULAR regardless of what the window contains -- disqualifying when every window is short by construction. The corrected form holds -1 / 0 / +1 for regular, Poisson and maximally bursty at every n, which is what makes two window sizes comparable.
- That correction is asserted directly rather than assumed: The same traffic is measured at n=32 and n=8 and required to agree. Beside it, the three shapes the tier classifier and the MWU pathways keep conflating are required to separate in order -- deadline-paced negative, longrun near zero, burst-starvation positive.
- A non-positive mean returns nothing. Burstiness is defined on a positive interval series, and a caller handing over a centered signal has handed over something the statistic does not describe.
- NEITHER IS WIRED TO A DECISION, on the same rule as the two estimators below.

SCHEDULER, TWO CHAOS ESTIMATORS PORTED AND MEASURED, NOT WIRED: src/chaos.rs
- Veitch-Abry Hurst, a wavelet-variance estimator over a Haar multiresolution cascade. For fractional Gaussian noise the detail variance at octave j scales as 2^(j(2H-1)), so a least-squares fit of log2 E[d_j^2] against j has slope 2H-1. What it buys this scheduler is a read it has never had: H > 0.5 says the workload is persistent, which is precisely the statement that the same task is likely to return and therefore that cache amortization will pay.
- Validated against signals whose answer is known by construction rather than against a golden number, since a short window cannot pin an exponent tightly and a test asserting otherwise would be asserting noise. Differenced noise reads 0.04, white noise 0.47 against its theoretical 0.5, a random walk and a ramp both 1.0. The ordering across those classes is the property under test.
- A perfect alternation returns nothing, and that is the correct answer rather than a gap. All of its energy lands in the first octave, every coarser approximation is exactly zero, and one point is not a scaling law. The noisy version of the same shape reads as maximally anti-persistent, which is the case that occurs in practice. Reporting an exponent fitted from a single octave would not be an estimate.
- Pecora-Carroll coupling asks whether neighbors in one series stay neighbors in another: If x_i and x_j are close, are y_i and y_j close. When y is driven by x the cross-prediction error collapses; when the two are independent it rises to the series' own mean pairwise spread, which is the baseline it is scored against. Both series are standardized first, so this measures shared dynamics and not shared units -- two runqueues couple or do not regardless of which carries more work.
- A flat series returns nothing rather than perfect synchrony. Idle is the common case, and two idle CPUs reading as fully slaved would be the most frequent possible false positive.
- NEITHER IS WIRED TO A DECISION. Both are ported and measured first, on the standing rule that a sensor earns a consumer only after its behavior on this scheduler's own signals is known. They are natural fusion inputs once the chaos layer fuses rather than conjoins.


wllclngn:
Allowing CI/CD to finish. Will fix veristat momentarily.

htejun (approved):

sirlucjan (approved):

PR #3748, opened 2026-08-19
scx_lavd: validate cached CPU pick against cpus_ptr on REENQ path

1343503048:
### Problem
When running schbench with scx_lavd, worker threads can experience severe starvation (~21 seconds) before the stall watchdog fires SCX_EXIT_ERROR_STALL.Stall dump showing the symptom:
```
  R k3s-server[8417] *root -21185ms
      scx_state/flags=4/0x3 dsq_flags=0x1 ops_state/qseq=0/0
      sticky/holding_cpu=-1/-1 dsq_id=0x5b
      dsq_vtime=45425777259 slice=4981350 weight=100
      cpus=00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000020,00000000,00000000,00000000 no_mig=1
      \_ cpdom_id: 11   scpu: 91   cgroup: unknown[1303]
```

Key observations from the dump:
no_mig=1: the task is migration-disabled, cpus_ptr is narrowed to a single CPU
scpu: 91: suggested_cpu_id is 91, which is not in cpus_ptr
The task was routed to CPU 91's per-CPU DSQ, but CPU 91 cannot run it → starvation

### Root Cause
The REENQ path in lavd_enqueue() reuses taskc->suggested_cpu_id to determine DSQ routing without validating it against the task's current p->cpus_ptr.When a task's CPU affinity changes after suggested_cpu_id was cached (e.g., migrate_disable() narrows cpus_ptr, or sched_setaffinity() changes the mask), the cached CPU may no longer be valid. The framework's select_task_rq() skips calling lavd_select_cpu() when is_migration_disabled(p) is true, so the cache is never refreshed. The task gets routed to a per-CPU DSQ whose owner CPU cannot run it, and since per-CPU DSQ tasks are not stealable, the task starves.

### Fix
Validate suggested_cpu_id against p->cpus_ptr in the REENQ path. If the cached CPU is out of range or not in the allowed mask, fall back to bpf_cpumask_first(p->cpus_ptr) and update the cache.This mirrors the existing validation in enqueue_cb() (the BTQ drain path), which was added by commit 3efc31c5 ("scx_lavd: validate cached CPU pick in enqueue_cb on BTQ reenqueue").


1343503048:
@multics69 please review this pr, thx.

multics69 (changes requested):
Thanks @1343503048 for the fix.

Could you elaborate possible scenarios of how taskc->suggested_cpu can be stale at REENQ path? Please add the scenarios to the comemnts and commit message. Thanks again!

1343503048:
> Thanks @1343503048 for the fix.
> 
> Could you elaborate possible scenarios of how taskc->suggested_cpu can be stale at REENQ path? Please add the scenarios to the comemnts and commit message. Thanks again!

Thanks for the review! Updated both the code comment and commit message with the detailed scenarios.

multics69 (approved):
Thanks!

PR #3753, opened 2026-08-20
scx_mitosis: Mark init_task callback sleepable

kkdwvd:
Make a fix similar to commit 3eb68fe2d010 ("scx_layered: Make init_task sleepable") such that the BPF verifier can make allocations in the function fallback to non-GFP_ATOMIC case. Passing args->cgroup directly is problematic since it is not annotated as a trusted pointer. When this happens, non-sleepable callbacks mark it as plain PTR_TO_BTF_ID, which is not accepted by most kfuncs taking trusted args but taken by bpf_cgrp_storage_get() for backward compatibility reasons. When the callback is marked sleepable, this pointer becomes untrusted is not even accepted by bpf_cgrp_storage_get(). Thus obtain a new referenced pointer through bpf_cgroup_from_id(). We cannot use bpf_cgroup_acquire() in either sleepable or non-sleepable case since it needs a trusted pointer. This extra refcount bump can be dropped once the minimum kernel version is bumped to v7.4, per the patch in [0].

  [0]: https://lore.kernel.org/bpf/95d7ccc17681aa3a4a2eeb1b073f00f7@kernel.org

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -323,13 +323,21 @@ static inline int update_task_cell(struct task_struct *p, struct task_ctx *tctx,
 		 * and the task is exiting.
 		 */
 		if (exiting_task_workaround_enabled && (p->flags & PF_EXITING)) {
-			struct cgroup *rootcg = READ_ONCE(root_cgrp);
-			if (!rootcg) {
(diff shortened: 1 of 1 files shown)


dforsyth (approved):

htejun (approved):

htejun:
BTW, I think we should make init_task() trigger scx_bpf_error() on failures. Failures from init_task() propagate as fork failures which leads to subtle and surprising failure modes from userspace POV. Might as well just boot the scheduler.

PR #3754, opened 2026-08-20
lib/cgroup_bw: manage only cpu.max-limited cgroups

multics69:
`lib/cgroup_bw` reserved a per-cgroup context set for every cpu-controller
cgroup at `ops.cgroup_init()`, against a static cap. On real hosts that both
overflows the cap and wastes memory -- e.g. 2500 cpu-controller cgroups with
only 4 actually `cpu.max`-limited fails to load with `-ENOMEM`, and even within
the cap most reserved sets are never used.

This series manages only the cgroups that define a finite `cpu.max` limit, plus
the root. Tasks in unlimited cgroups bill to their nearest managed ancestor,
and the roll-up and throttle propagation treat unmanaged cgroups as
pass-through gaps, so the footprint scales with the number of limits, not the
number of cgroups.

### Organization

The series is organized in four steps:

- **Groundwork** prepares the machinery the redesign needs: **1)** the cgroup-count
  and nesting-depth caps become configurable and load-time sized; **2)** a context
  is unpublished from the lookup map before it is recycled and **3)** BTQs are
  recycled instead of destroyed, making context churn race-safe; and **4)** the
  runtime roll-up is made gap-safe so it tolerates unmanaged cgroups in the
  hierarchy.

- **The core change** **5)** manages only cgroups with a finite `cpu.max` limit and
  **6)** bills tasks in unlimited cgroups to their nearest managed ancestor.

- **Runtime limit changes** cover a cgroup that gains a `cpu.max` limit after
  init through `ops.cgroup_set_bandwidth()`: **7)** spare context sets are reserved
  up front and **8)** a billing-cache generation id is added, then **9)** such a cgroup
  is managed on the fly. The generation id is needed because managing a new
  cgroup changes which managed ancestor its descendants bill to; bumping it
  lazily invalidates the now-stale per-task billing caches so they re-resolve,
  instead of walking every task.

- **Task-based billing, a cleanup, and on-demand allocation**: **10)** billing is
  resolved from the task instead of a cgroup id, which works around a known
  pre-v6.18 kernel bug -- `bpf_cgroup_from_id()` filters by the caller's cgroup
  namespace, so a cross-namespace wakeup silently unenforced `cpu.max` -- by
  changing the entry-point arguments to take the task; **11)** with managed cgroups
  now keeping their per-LLC contexts for life, the `has_llcx` flag is redundant
  and is removed; and **12)** when the kernel allows `ops.cgroup_set_bandwidth()` to
  sleep (commit `5f01293930d1`), contexts are allocated on demand instead of
  pre-reserved, with `setup_cgroup_bw()` picking the path at load and falling
  back to pre-reservation on older kernels.

### Testing

Tested on patched (sleepable) and unpatched kernels, and under a cgroup
mass-create/mass-death stress test.

### Commits

1. `lib/cgroup_bw`: make the cgroup count and nesting-depth caps configurable
2. `lib/cgroup_bw`: unpublish cgroup ctx before recycling it
3. `lib/cgroup_bw`: recycle BTQs instead of destroying them
4. `lib/cgroup_bw`: make the runtime_total_sloppy roll-up gap-safe
5. `lib/cgroup_bw`: manage only cgroups that define a limit
6. `lib/cgroup_bw`: bill tasks to their nearest managed ancestor
7. `lib/cgroup_bw`: reserve spare context sets for runtime cpu.max limits (groundwork)
8. `lib/cgroup_bw`: add a billing-cache generation id (groundwork)
9. `lib/cgroup_bw`: manage a cgroup that gains a cpu.max limit at runtime
10. `lib/cgroup_bw`: resolve a task's billing cgroup from the task, not its id
11. `lib/cgroup_bw`: drop the now-redundant has_llcx flag
12. `lib/cgroup_bw`: allocate cpu.max contexts on demand when the callback may sleep


multics69:
The runtime error on 6.13 kernel was fixed by the last commit ([lib/cgroup_bw: resolve the billing cgroup without scx_bpf_task_cgroup()](https://github.com/sched-ext/scx/pull/3754/commits/a1bbb6119556433d59d28c20396aa3465489d083)).

daidavid (approved):

PR #3757, opened 2026-08-21
scx_bpfland: Use SCX_ENQ_IMMED instead of cpu_release()

EricccTaiwan:
ops.cpu_release() is deprecated in favor of tracking CPU preemption from a sched_switch tracepoint. scx_bpfland only uses it to drain a CPU's local DSQ via scx_bpf_reenqueue_local() when TIMELY mode is enabled and a higher-priority class takes the CPU over.

Mirrors the same conversion already done in scx_lavd and, in the kernel tree, tools/sched_ext/scx_pair.bpf.c (2abee44c7db9).

### DIFF
--- a/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_bpfland/src/bpf/main.bpf.c
@@ -133,7 +133,7 @@ volatile u64 nr_kthread_dispatches, nr_direct_dispatches, nr_shared_dispatches,
 	nr_idle_global_misses, nr_waker_cpu_biases, nr_keep_running_reuses,
 	nr_keep_running_queue_empty, nr_keep_running_smt_blocked,
 	nr_keep_running_queued_work, nr_dispatch_cpu_dsq_consumes,
-	nr_dispatch_node_dsq_consumes, nr_cpu_release_reenqueue;
+	nr_dispatch_node_dsq_consumes;
 
 /*
  * Amount of currently running tasks.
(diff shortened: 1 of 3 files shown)


scheds/rust/scx_bpfland/src/bpf/main.bpf.c
```diff
@@ -1455,14 +1455,47 @@ void BPF_STRUCT_OPS(bpfland_exit, struct scx_exit_info *ei)
...
 		__sync_fetch_and_add(&nr_cpu_release_reenqueue, 1);
 	}
 }
 
+/*
+ * ops.cpu_acquire/release() are deprecated in favor of tracking CPU
+ * preemption from a sched_switch tracepoint. Autoloaded only on kernels
+ * that support scx_bpf_reenqueue_local_from_anywhere(); see main.rs.
+ */
+SEC("?tp_btf/sched_switch")
```
arighi: Honestly I'm not a big fan of adding a sched_switch tracepoint, this runs on every context switch and on every CPU, the BPF program entry/exit overhead might be relevant. It'd be interesting to measure it...

Moreover, if I'm not wrong, the condition also fires for idle -> RT/DL transitions because idle has prio >= 100.

IMHO, we should just remove ops.cpu_release() completely. We have the ext deadline server now, so tasks interrupted by higher sched classes will have a chance to run on the CPU they're dispatched to.

Another possibility could be to add an explicit cmdline option to enable/disable this behavior (having it disabled by default).

arighi (commented):
Thanks for looking at this @EricccTaiwan , I left a comment, let me know what you think.

scheds/rust/scx_bpfland/src/bpf/main.bpf.c
htejun: Or use `SCX_ENQ_IMMED`.

EricccTaiwan: > Or use `SCX_ENQ_IMMED`.

souds good!

arighi (approved):
Looks good, thanks!

PR #3758, opened 2026-08-21
scx_utils: Require CONFIG_FUNCTION_TRACER for scx_lib_init_probe()

### LINKED ISSUE #3745: scx_bpfland >=1.1.2 incompatibility with previous kernel config
scx_bpfland included in 1.1.2 and beyond appears to have an incompatibility with previous kernel configs if CONFIG_FUNCTION_TRACER (and possibly other options) set =n (originally thought/claimed this was a problem with v6.18 LTS kernels). scx_loader fails to load it with the following repeated log:
Aug 13 18:26:47 hostname scx_loader[1307]: Error: Failed to attach non-struct_ops BPF programs
Aug 13 18:26:47 hostname scx_loader[1307]: Caused by:
Aug 13 18:26:47 hostname scx_loader[1307]:     Device or resource busy (os error 16)
Aug 13 18:26:47 hostname scx_loader[1153]: [ERROR]: Failed to start scheduler (attempt 5/5)

Also attaching a verbose log from a recent main build which also has the issue.
[scx_bpfland_v1.1.2_v6.18_kernel.txt](https://github.com/user-attachments/files/31180968/scx_bpfland_v1.1.2_v6.18_kernel.txt)

LLM (Codex) concise analysis:
• scx_bpfland 1.1.2 fails on kernel 6.18.44 because its new scx_lib_init_probe fentry probe cannot attach to bpf_scx_reg; the
  kernel returns EBUSY. No other sched_ext scheduler is active, and all other BPF components load successfully. This appears
  to be a 1.1.2/kernel compatibility regression; 1.1.1 did not include this probe.

EricccTaiwan:
scx_lib_init_probe() is a fentry probe auto-attached by every scheduler via scx_ops_attach!(). Without CONFIG_FUNCTION_TRACER it can't attach, and skel.attach()'s all-or-nothing semantics takes the whole scheduler down.

Skip its autoload when `/proc/sys/kernel/ftrace_enabled` is absent. Reproduced and fixed in vng with a CONFIG_FUNCTION_TRACER=n kernel.

Closes: #3745

### DIFF
--- a/rust/scx_utils/src/compat.rs
+++ b/rust/scx_utils/src/compat.rs
@@ -433,6 +433,14 @@ pub fn tracepoint_exists(tracepoint: &str) -> Result<bool> {
     Ok(false)
 }
 
+/// Whether the kernel supports function tracing (CONFIG_FUNCTION_TRACER),
+/// needed for fentry/fexit BPF programs to load or attach.
+/// `/proc/sys/kernel/ftrace_enabled` only exists when it's compiled in,
+/// so its presence is a cheap proxy.
+pub fn function_tracer_available() -> bool {
+    std::path::Path::new("/proc/sys/kernel/ftrace_enabled").exists()
+}
+
 pub fn cond_kprobe_enable<T>(sym: &str, prog_ptr: &OpenProgramImpl<T>) -> Result<bool> {
     if in_kallsyms(sym)? {
         unsafe {
@@ -744,6 +752,10 @@ macro_rules! __scx_ops_load {
                         ops.rescue_quantum_us = 0;
                     }
                 }
+            }
+
+            if !scx_utils::compat::function_tracer_available() {
+                break 'block Err(anyhow::anyhow!("kernel is missing CONFIG_FUNCTION_TRACER"));
             }
 
             scx_utils::uei_set_size!($skel, $ops, $uei);


arighi:
Hi @EricccTaiwan, I just noticed this after sending #3759... Maybe we can apply this first to minimize the potential behavioral regression on pre-v6.18 kernels, while we evaluate whether #3759 can be an acceptable compromise for removing the ftrace dependency entirely.

So this one LGTM for now.

arighi (approved):

htejun:
Shouldn't we just require `CONFIG_FUNCTION_TRACER`?

arighi:
Probably yes, it's easier to just require CONFIG_FUNCTION_TRACER for now and trigger a clear error when it's not available. At some point, when v6.18 becomes too old we can probably remove this requirement. I'll close #3759 we can continue in this PR.

EricccTaiwan:
> Probably yes, it's easier to just require CONFIG_FUNCTION_TRACER for now and trigger a clear error when it's not available. At some point, when v6.18 becomes too old we can probably remove this requirement. I'll close #3759 we can continue in this PR.

Makes sense, throwing an error is intuitive, thx!

htejun (approved):

arighi (approved):

PR #3759, opened 2026-08-21
[RFC] Prevent CONFIG_FUNCTION_TRACER dependency in is_migration_disabled()

arighi:
This series allows schedulers using the common scx infrastructure to load on kernels built without ftrace support.

The immediate problem is that common.bpf.h uses an `fentry/bpf_scx_reg` program to determine whether the BPF trampoline calls `migrate_disable()`. Attaching this program requires ftrace, so kernels built with `CONFIG_FUNCTION_TRACER=n` reject the attachment with `-EBUSY`. As a result, every scheduler including common.bpf.h fails to load.

Similarly, `UEI_RECORD()` currently uses tracing-only helpers even when `CONFIG_BPF_EVENTS=n`. The first patch avoids those helpers in that configuration, retaining the numeric exit information while omitting the unavailable strings.

The second patch removes the fentry-based migration-disabled probe. Since kernel commit `8e4f0b1ebcf2 ("bpf: use rcu_read_lock_dont_migrate() for trampoline.c")`, included in v6.18, the non-sleepable trampoline calls `migrate_disable()` only when `CONFIG_PREEMPT_RCU` is enabled. We can therefore use `CONFIG_PREEMPT_RCU` directly on v6.18 and newer kernels:
```
   Kernel configuration     Behavior
   PREEMPT_RCU              Discount the prolog increment for current
   !PREEMPT_RCU             Read migration_disabled directly
```
This preserves the existing classification on v6.18 and newer kernels, without introducing false positives or additional runtime/load-time overhead.

There is a deliberate compatibility tradeoff for kernels older than v6.18 with `CONFIG_PREEMPT_RCU=n`. On those kernels, the older trampoline may still increment `current->migration_disabled`. Without the runtime fentry probe, a normal task passed as current may therefore be conservatively classified as migration-disabled, potentially re-introducing a false-positive migration-disabled detection.

However, this possible regression is limited to the current task on pre-v6.18 kernels with `CONFIG_PREEMPT_RCU=n`. The field remains accurate for non-current tasks and genuinely migration-disabled tasks. The conservative result may keep current local unnecessarily, but it avoids the unsafe opposite result of migrating a genuinely migration-disabled task.

I also considered alternative runtime detection mechanisms, such as using `bpf_dummy_ops` with `BPF_PROG_TEST_RUN`. However, `bpf_dummy_ops` requires `CONFIG_NET=y`, so that approach would introduce another configuration dependency. The proposed compromise appears to be the simplest way to support v6.18 and newer kernels reliably across all valid sched_ext kernel configurations, while removing the dependency on ftrace. Feedback on alternative approaches is welcome.

This fixes issue #3745.

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -530,65 +530,18 @@ static __always_inline const struct cpumask *cast_mask(struct bpf_cpumask *mask)
 }
 
 /*
- * True if the non-sleepable BPF trampoline prolog (__bpf_prog_enter) calls
- * migrate_disable() for the current task. Recorded once by
- * scx_lib_init_probe, an fentry program on bpf_scx_reg() that fires during
- * the natural scheduler-attach call chain (auto-attached by scx_ops_attach!).
- *
- * Defaults to true (conservative). Over-reporting in is_migration_disabled()
- * causes local-only dispatch, which is safe. Under-reporting can crash the
- * scheduler, so we err high if the probe somehow fails to run.
- */
-bool __scx_prolog_disables_migration __weak = true;
-
-/*
- * scx_lib_init_probe - non-sleepable prolog probe.
- *
- * Attached to bpf_scx_reg(), the .reg callback in bpf_sched_ext_ops
- * (kernel/sched/ext.c). The kernel's struct_ops machinery invokes
- * bpf_scx_reg when userspace creates the scheduler link, before
- * ops.init() fires. Its address is taken in the vtable, so the symbol
- * is non-inlinable and has been stable since introduction.
- *
- * Entering via fentry runs us through __bpf_prog_enter -- the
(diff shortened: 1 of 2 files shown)


htejun:
Shouldn't we just require `CONFIG_FUNCTION_TRACER`?

arighi:
> Shouldn't we just require `CONFIG_FUNCTION_TRACER`?

But CONFIG_FUNCTION_TRACER is only required pre-v6.18. Maybe we can require CONFIG_FUNCTION_TRACER for now to preserve the exact behavior for pre-v6.18 kernels (and possible backports) and, at some point, remove `scx_lib_init_probe()` and drop the requirement, since CONFIG_PREEMPT_RCU describes the trampoline behavior directly with v6.18+.

arighi:
Let's continue with a simpler approach in #3758 for now.

PR #3761, opened 2026-08-22
lib: Multi-instance sdt allocator and arena-backed per-cgroup data

htejun:
Three prep patches for arena-backed per-cgroup data:

- Split the allocator declarations out of sdt_task.h into sdt_alloc.h so
  that non-task users don't have to pull in the per-task API.
- Prepare scx_alloc for multiple allocator instances: initialize the shared
  pools only once and fail loudly on an uninitialized allocator, whose NULL
  root otherwise aliases arena page 0 and silently corrupts the static
  allocator.
- Add arena-backed per-cgroup data mirroring the per-task machinery, with
  BPF cgroup local storage holding the reference so that lookups go through
  the cgroup pointer with no separate index. Unlike task data, cgroup data
  is created on demand from multiple racing paths and freed from hooks
  which fire for every cgroup in the system, so alloc publishes with
  cmpxchg and free claims with xchg, making both safe against repeated and
  concurrent calls.

sdt_cgroup.bpf.c has no in-tree consumer yet so CI does not compile it. It
is build- and runtime-verified by an upcoming scheduler which uses it for
cgroup contexts freed from an fentry on bpf_cgrp_storage_free(): alloc/free
cycled through cgroup mkdir/rmdir churn with the free hook firing exactly
once per removed cgroup, plus concurrent creation from racing init paths.


multics69 (approved):
LGTM

PR #3765, opened 2026-08-23
lib: Add scx_task/cgrp_free_rcu(), poor man's RCU deferred freeing

htejun:
Consumers reaching sdt task and cgroup data through side channels need RCU
semantics: lookups that fail once the owner is gone and pointers that stay
valid until the end of the RCU section. BPF is yet to grow bpf_call_rcu(),
so add scx_task_free_rcu() and scx_cgrp_free_rcu() backed by scx_urcu, a
poor man's substitute with the grace period driven from userspace,
membarrier(MEMBARRIER_CMD_GLOBAL) being synchronize_rcu(). The names
promise only the semantics: once BPF can defer frees natively, the
implementation swaps out under the same callers.

Freed nodes accumulate on the active side of a two-sided list and a
doorbell ringbuf wakes the reclaim daemon, which waits a grace period and
returns the draining side to the allocator. The driver programs ride in
the lib objects and ArenaLib::setup() discovers them and spawns the daemon
automatically, so the whole scheduler-side surface is calling the free_rcu
variants from free path hooks. The first patch is a small prep splitting
out __scx_task_data() for quiet lookups.

The membarrier() trick was suggested by Kumar Kartikeya Dwivedi.

Based on #3761; retargets to main once it merges.

Testing: fork storms plus burst/quiescence patterns with every urcu event
traced alongside membarrier syscalls in a single ftrace timeline, verifying
per-node lifecycle alternation, that every reclaimed node was pushed before
the grace period opening its drain cycle, flip -> grace period -> walk
ordering on every cycle, no stalls or stranded frees, and that reclaim is
fully demand-gated with no grace periods while idle.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5724,6 +5724,7 @@ dependencies = [
  "anyhow",
  "libbpf-rs",
  "libbpf-sys",
+ "libc",
  "scx_cargo",
  "scx_utils",
  "simplelog",
--- a/lib/sdt_alloc.bpf.c
+++ b/lib/sdt_alloc.bpf.c
@@ -1449,3 +1449,116 @@ int scx_userspace_arena_free_pages(struct scx_userspace_arena_free_pages_args *c
 	bpf_arena_free_pages(&arena, ctx->addr, pages);
 	return 0;
 }
+
+/*
+ * Poor man's userspace-driven RCU, standing in until BPF grows bpf_call_rcu().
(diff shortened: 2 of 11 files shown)


kkdwvd (approved):
LGTM.

multics69 (approved):
LGTM. I like the membarrier trick.

etsal (approved):

PR #3766, opened 2026-08-24
common.bpf.h: default __scx_prolog_disables_migration to false

multics69:
__scx_prolog_disables_migration is a flag that is_migration_disabled() uses on its slow path (pre-v6.18, !CONFIG_PREEMPT_RCU, migration_disabled == 1). It defaulted to true, which under-reports: for the current task the slow path then returns

	bpf_get_current_task_btf() != p

i.e. false, reporting migration as not disabled. Under-reporting can crash the scheduler while over-reporting only forces local-only dispatch, so the fallback is meant to err high -- and erring high means defaulting to false. So, default to false and fix the comment to match.

Fixes: c98c5ad22f7e ("common.bpf.h: probe BPF prolog migrate_disable() from non-sleepable context")
Reported-by: Sashiko <sashiko-bot@kernel.org>
Link: https://lore.kernel.org/sched-ext/20260818210732.AB27B1F000E9@smtp.kernel.org/
Suggested-by: Tejun Heo <tj@kernel.org>

### DIFF
--- a/scheds/include/scx/common.bpf.h
+++ b/scheds/include/scx/common.bpf.h
@@ -535,11 +535,11 @@ static __always_inline const struct cpumask *cast_mask(struct bpf_cpumask *mask)
  * scx_lib_init_probe, an fentry program on bpf_scx_reg() that fires during
  * the natural scheduler-attach call chain (auto-attached by scx_ops_attach!).
  *
- * Defaults to true (conservative). Over-reporting in is_migration_disabled()
+ * Defaults to false (conservative). Over-reporting in is_migration_disabled()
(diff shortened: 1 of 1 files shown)


arighi (changes requested):
Hi @multics69, the code looks correct, but we should probably remove the merge commit and rebase onto main. Thanks!

multics69:
> Hi @multics69, the code looks correct, but we should probably remove the merge commit and rebase onto main. Thanks!

Thanks @arighi! Done.

htejun (approved):

arighi (approved):

PR #3767, opened 2026-08-25
scx_cake: select_cpu redesign — self-park mailboxes become default (2…

RitzDaCat:
Rebuild CPU selection so the wake path reads a precomputed decision instead of making one. Idle CPUs evaluate their own placement gates at idle-entry (local reads, free time) and park themselves into per-physical-core mailboxes; a waking task takes prev-CPU-if-idle or its own mailbox entry with plain loads and one direct dispatch — no gate chain, no ranked LLC scan, no consuming claim on the common path. Census fallback and a zero-skip cover the miss and saturated cases.

Measured improvements (bpf_stats per-callback timing, mirrored ABBA rotations, appsim hd2-mission fitted load unless noted):

  select_cpu            185 -> 83 ns/call   (27.4k -> 12.5k us/s, -57%)
  dispatch              219 -> 116 ns/call  (halved)
  enqueue path          48k -> ~1k calls/s  (wakes place directly)
  total scheduler cost  ~80 -> ~48 ms/s    (-40%)
  context switches      -5% vs old default; cake runs 143k/s where EEVDF
                        ran 303k/s on the same workload
  per-wake total        ~81 ns vs EEVDF's ~186 (select 66 + enqueue 120)

Saturated regime (schbench -m4 -t16): census zero-skip cuts select_cpu 213 -> 63-73 ns/call (-66%), sealed across mirrored slots.

Game screen (Helldivers 2 menu, MangoHud frame capture, mirrored ABBA, ~12k frames/slot, new design vs previous default):

  severe frames (>2x median)  0.062% -> 0.037%
  0.1% low                    171 -> 286 fps
  p99.9 - median              tie (1.13 vs 1.14 ms)
  average fps                 tie (617 vs 615)
  worst single frame jump     9.3 -> 4.0 ms

Workload guards vs native EEVDF (exact-pair, 2 blocks, new default):

  mutex-handoff    +39.7% [+38.4, +41.0]  (previous seal was +10.3%;
                   migrations -162k -- the mailbox amplifies the handoff
                   shape; 8-block seal pending)
  blender-render   -0.21% [-0.29, -0.14]  (practical tie; closes the
                   BLENDER-RED regression measured at -1.6..-2.3% on the
                   old base)
  earlier HD2 game ab, old default vs native: cake 616/527/388 avg/1%/0.1%
  fps against native 599/455/366

Hardwired as default (toggles removed): §G45 idle census, §G50-R census-empty scan skip, §G53 census first-fit fallback, §G54 self-park mailboxes, §G43 going-idle hint.

Removed (measured null, falsified, or closed -- history in STATE.md): §G44 qmask wake routing (null; emptiness question moves to the M4 consolidation program), §G47 ISR-successor (falsified at the KovaaKs frame gate: severe frames doubled, wake benefit exists only at deep idle), §G48 hint-first placement (verified 74% idle claim rate but no regime where the claim costs less than the scan it skips), §G55 insert probe.

Design law recorded (§R.29, bought with data): identity claims only work from event-pushed hints; pulled aggregate state answers count questions ("someone is idle") but never identity ("take this one") under concurrent wakers -- census-as-gate works, census-as-identity loses the race and pays both paths.

New instrumentation: kernel bpf_stats per-callback cost windows (cake-bpfstats scoped helper), core event counter readout at detach, per-function spill gate held throughout (select_cpu 3 spills, 486 insns).

Remaining toggles, each with a reason: g46 slice cache (separated memcpy win awaiting its pre-registered ccm-memcpy + mutex seal; costs +2.6k us/s on the new base), g51 idle-depth model (inert without a cpuidle driver; BIOS Global C-State Control decision pending), g52 CPPC preferred-core rank (consumer scheduled with the parking-policy work), m6/m7 direct kernel reads (null on the new base; their query sites left the hot path).

Registered, not yet built: K1 chronic-sink accounting in BPF, K2 fused place kfunc (the path from 12.5k to the 8k us/s target), M2 frame-clock demotion, EEVDF head-to-head HD2 capture (needs a human-present launch).

### DIFF
--- a/scheds/rust/scx_cake/.gitignore
+++ b/scheds/rust/scx_cake/.gitignore
@@ -15,6 +15,8 @@ docs/*
 !docs/SIGNALS_AUDIT_2026-08-17.md
 !docs/SWEEP_ARITH_2026-08-03.md
 !docs/THEORY_CONSTRUCTS_ABC_2026-08-17.md
+!docs/AUDIT_COMPONENT_COST_2026-08-23.md
+!docs/KERNEL_TOOL_INVENTORY_2026-08-23.md
 
 # Benchmark binaries and local captures
 bench/*
--- a/scheds/rust/scx_cake/STATE.md
+++ b/scheds/rust/scx_cake/STATE.md
@@ -29,6 +29,22 @@ live WoW confirm); all hashes cited below resolve via those local backup branche
 
 ## RESUME HERE
 
+**PICKUP 2026-08-21 EOD — machine left on EEVDF, nothing attached, all
+background jobs stopped.** Code tip `60b238311` (§G38.1-repaired + §G40; §G39-B
+aborted), tip receipt `20260821T155506Z_head-5e0e8244b21a` built + attach-tested.
+Three staged actions, in graph-cut order:
+1. **R1 rotation** (needs KovaaKs being played, ~3.5 min):
+   `bash runs/rotation_r1_staged/r1_rotation.sh` — 3-arm native/G38/tip; draws
+   the tip's chain lane, scores §G38.1+§G40 (GameThread p99 128 → toward 6?).
+2. **E4a pinned chain3 split** (no game needed, ~3 min): rerun
+   `runs/causeb_chain3_20260821/ab.sh` (bug fixed) with stages pinned —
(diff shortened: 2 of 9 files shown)


sirlucjan (approved):
Good job Paul!

PR #3769, opened 2026-08-26
scx_mitosis: Always use tracepoints for cgroup tracking

dforsyth:
Make the existing --cpu-controller-disabled path unconditional.

This change ensures we have valid cell management whether or not the cpu controller is enabled, and makes --cpu-controller-disabled a deprecated no-op.


kkdwvd (approved):
Needs rebase, but LGTM fwiw.

PR #3770, opened 2026-08-26
lib: cmask distribute helpers and caller-chosen allocator element alignment

htejun:
Prep patches for an upcoming cid conversion, useful on their own. Updated
per review: the allocator alignment work now sits on top of a set of lib
preps addressing both review comments.

- cid.bpf.h gains cmask_any_distribute() and cmask_any_and_distribute(),
  cmask counterparts of the kernel's cpumask distribute helpers sharing a
  per-CPU rotor, so pick sites can spread the picks instead of always
  taking the first set bit.

- New lib/const-defs.h carries SCX_CACHELINE_SIZE with per-arch values
  mirroring the kernel's L1_CACHE_SHIFT (64 default, 128 on powerpc64, 256
  on s390), replacing the ad hoc always-64 definitions (@arighi's
  arch-dependent cacheline comment).

- The userspace half of lib/selftests builds again (the lib headers'
  userspace mode had rotted unnoticed; CI only builds
  lib/selftests/compat). Used below for validation.

- The allocation metadata (sdt_data) moves behind the payload: elements
  are payload-first, so scx_alloc() returns pointers that genuinely carry
  the requested alignment (@arighi's returned-object alignment comment).
  New scx_free() frees by pointer with the index fetched from the trailing
  metadata, letting ATQs, rbtrees, rbnodes and bitmaps drop their embedded
  allocation ids.

- scx_alloc_init(), scx_task_init(), scx_cgrp_init() and ArenaLib::init()
  gain the element alignment parameter (0 = word). ATQs are allocated
  cacheline-aligned; existing schedulers keep word alignment and identical
  layouts.

Testing: cargo build and fmt clean (scx_mitosis, scx_lavd, scx_p2dq,
scx_chaos); lib selftests (atq, btree, lvqueue, minheap, rbtree) pass in a
VM on sched_ext/for-7.3; traced stress runs verified the allocator
alloc/free/RCU-free lifecycle invariants (no premature frees, exact
payload zeroing, no reclaim stalls) and payload alignment (task ctx
pointers cacheline-aligned); exercised further under a WIP scheduler with
fork-storm and cell-reconfig churn.

### DIFF
--- a/lib/sdt_alloc.bpf.c
+++ b/lib/sdt_alloc.bpf.c
@@ -291,7 +291,7 @@ static int pool_set_size(struct sdt_pool *pool, __u64 data_size, __u64 nr_pages)
 
 /* initialize the whole thing, maybe misnomer */
 __hidden int
-scx_alloc_init(struct scx_allocator *alloc, __u64 data_size)
+scx_alloc_init(struct scx_allocator *alloc, __u64 data_size, __u64 align)
 {
 	size_t min_chunk_size;
 	int ret;
@@ -321,9 +321,15 @@ scx_alloc_init(struct scx_allocator *alloc, __u64 data_size)
 			return -ENOMEM;
 	}
 
-	/* Wrap data into a descriptor and word align. */
+	if (!align)
+		align = 8;
+	if (unlikely(align < 8 || (align & (align - 1)))) {
+		scx_err_loc("invalid alignment %llu", align);
+		return -EINVAL;
+	}
+
 	data_size += sizeof(struct sdt_data);
-	data_size = round_up(data_size, 8);
+	data_size = round_up(data_size, align);
 
 	/*
 	 * Ensure we allocate large enough chunks from the arena to avoid excessive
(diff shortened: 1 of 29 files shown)


rust/scx_arena/scx_arena/src/lib.rs
```diff
@@ -24,6 +24,9 @@ use anyhow::Result;
 use libbpf_rs::libbpf_sys;
 use libbpf_rs::MapCore as _;
 
+/// Cacheline size assumed by the arena allocator's alignment parameter.
+pub const CACHELINE_SIZE: usize = 64;
```
arighi, line 34: Not a blocker for this PR, but I'm wondering if we should have an arch-dependent cache-line size. IIRC on s390 L1_CACHE_BYTES is like 256 bytes, 128 bytes on PPC, etc.

lib/sdt_alloc.bpf.c
```diff
@@ -321,9 +321,15 @@ scx_alloc_init(struct scx_allocator *alloc, __u64 data_size)
...
+	if (!align)
+		align = 8;
+	if (unlikely(align < 8 || (align & (align - 1)))) {
+		scx_err_loc("invalid alignment %llu", align);
+		return -EINVAL;
+	}
+
 	data_size += sizeof(struct sdt_data);
-	data_size = round_up(data_size, 8);
+	data_size = round_up(data_size, align);
```
arighi, line 332: This aligns the allocation stride, but does it also guarantee the requested alignment for the object returned to the caller?

arighi (approved):

PR #3771, opened 2026-08-26
lib: Guard the arena NULL pages and watch BPF streams

htejun:
Make stray NULL arena dereferences detected and fatal instead of silent
corruption. Three patches.

- bpf_stream_printk() gated on and called bpf_stream_vprintk_impl(), the
  kfunc spelling that v7.0 retired for the implicit-prog-aux
  bpf_stream_vprintk(), so on v7.0+ every stream print in the lib (scx_err()
  and friends included) has been a silent no-op. Point it at the current
  kfunc; prints on v6.18/v6.19 revert to the designed no-op fallback.

- ArenaLib::init() is folded into setup(), which now returns an owned
  ArenaLib that the scheduler keeps alive for the scheduler instance's
  lifetime (p2dq per restart iteration, lavd and chaos as a scheduler
  struct field). It owns the lib's background threads, the urcu reclaim
  daemon and the stream watcher below, each paired with a stop eventfd,
  and dropping it signals them, lets them run a final drain and joins
  them. In-process scheduler restarts no longer accumulate detached
  threads whose cloned fds pin the retired object's maps and arena
  pages (@arighi's watcher lifetime comment). Program fds are cloned
  via a raw fd probe so that programs that did not load (feature-gated)
  are skipped instead of constructing a BorrowedFd from an invalid fd
  (@arighi's autoload comment).

- A NULL arena pointer aliases the arena's first page: once an allocation
  claims that page, a stray NULL dereference in BPF reads or corrupts live
  data with no fault and no report (verified by experiment). Keep the front
  32MB of the arena unallocated with bpf_arena_reserve_pages() so NULL
  dereferences stay in the faulting class: the access faults, is recovered
  (loads read zero, stores are dropped) and is reported with a stack trace
  on the program's BPF stderr stream. The guard is skipped on kernels
  without the kfunc, and on kernels without full-range LDIMM64 offset
  support where libbpf places the __arena globals at the bottom of the
  arena, occupying the guard region (v6.17 and earlier, including the CI
  runner kernels).

  Nothing read those streams in production, so ArenaLib::setup() now spawns
  a watcher thread that polls every program's stdout/stderr streams,
  forwards them to the scheduler's stdout/stderr, and aborts the scheduler
  on the kernel's "ERROR: " reports (recovered arena faults, rqspinlock
  deadlocks, may_goto timeouts). Lines prefixed "IGN: " are dropped, for
  prints with side effects that nobody needs to see. Stream content is
  handled line-wise: the lib print wrappers newline-terminate every print
  and the watcher reassembles complete lines across reads (unit-tested
  splitter), so a report cannot be missed by gluing onto an unterminated
  print or splitting at a read boundary.

Testing on sched_ext/for-7.3 in a VM with scx_p2dq: clean runs load with the
guard reserved and the watcher silent, clean exit. An injected NULL arena
read produced the kernel's "ERROR: Arena READ access at unmapped address
..." report on the BPF stderr stream, forwarded intact right next to an
adjacent same-tick debug print, and the scheduler aborted with exit 1. Debug
dumping verified: BPF stdout prints land on stdout and stderr prints on
stderr, an injected >64KB stderr backlog (850 lines) forwarded with every
line intact across the read-size boundary, an unterminated raw print flushed
as its own line, IGN lines dropped. Lib selftest suites (atq, btree,
lvqueue, minheap, rbtree) pass in the VM (with the pre-existing broken
topology selftest bypassed); cargo build (scx_lavd, scx_p2dq, scx_chaos,
scx_arena), cargo test and fmt clean. Also verified on v6.17 (arena
globals at the bottom): the guard is skipped and both scx_p2dq and
scx_lavd load and run cleanly, with programs that fail to load on older
kernels skipped by the stream watcher instead of failing setup.

### DIFF
--- a/rust/scx_arena/scx_arena/src/lib.rs
+++ b/rust/scx_arena/scx_arena/src/lib.rs
@@ -260,3 +260,216 @@ pub(crate) fn urcu_spawn(obj: &libbpf_rs::Object) -> Result<Option<Daemon>> {
         thread: Some(thread),
     }))
 }
+
+const BPF_STDOUT: u32 = 1;
+const BPF_STDERR: u32 = 2;
+const BPF_STREAMS: [(u32, &str); 2] = [(BPF_STDOUT, "stdout"), (BPF_STDERR, "stderr")];
+const STREAM_POLL_INTERVAL: Duration = Duration::from_secs(1);
+
+fn stream_read(fd: &OwnedFd, stream_id: u32, buf: &mut [u8]) -> i32 {
+    unsafe {
+        libbpf_sys::bpf_prog_stream_read(
+            fd.as_raw_fd(),
+            stream_id,
+            buf.as_mut_ptr() as *mut _,
+            buf.len() as u32,
+            std::ptr::null_mut(),
+        )
+    }
+}
+
+/// The kernel prefixes the errors it reports on a program's BPF stderr
+/// stream with "ERROR: ".
+fn stream_is_fatal(lines: &[String]) -> bool {
+    lines.iter().any(|l| l.starts_with("ERROR: "))
+}
+
+/// Split the complete lines out of @carry, keeping a trailing partial for the
+/// next read. Stream elements are concatenated without separators and reads
+/// split lines arbitrarily, so prefixes may only be matched on complete lines.
+/// @new_data says whether this poll read anything: a partial with no
+/// continuation after a full poll interval is flushed as a line of its own.
+fn stream_extract_lines(carry: &mut String, new_data: bool) -> Vec<String> {
+    let mut lines: Vec<String> = Vec::new();
+
+    if let Some(pos) = carry.rfind('\n') {
+        lines = carry[..pos].split('\n').map(str::to_string).collect();
+        carry.drain(..=pos);
+    }
+    if !new_data && !carry.is_empty() {
+        lines.push(std::mem::take(carry));
+    }
+
+    lines
+}
+
+/// Forward the BPF stdout/stderr streams of every program in the object to
+/// the scheduler's stdout and stderr respectively, and abort when the
+/// kernel reports an error on a stderr stream. Nothing reads these streams
+/// otherwise, so messages would be silently dropped. Lines prefixed "IGN: "
+/// are dropped instead of forwarded, for prints with side effects that
+/// nobody needs to see, the arena association print in
+/// scx_arena_subprog_init() for example. Called from ArenaLib::setup(), the
+/// watcher is owned by the returned ArenaLib.
+pub(crate) fn stream_watcher_spawn(obj: &libbpf_rs::Object) -> Result<Daemon> {
+    let mut progs = Vec::new();
+
+    for prog in obj.progs() {
+        let Some(name) = prog.name().to_str() else {
+            continue;
+        };
+        /* feature-gated programs may not be loaded and then have no fd */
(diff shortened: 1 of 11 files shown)


rust/scx_arena/scx_arena/src/lib.rs
```diff
@@ -128,6 +128,131 @@ fn urcu_daemon(doorbell: libbpf_rs::MapHandle, pairs: Vec<(OwnedFd, OwnedFd)>) -
...
+/// runs detached like the urcu daemon.
+pub(crate) fn stream_watcher_spawn(obj: &libbpf_rs::Object) -> Result<()> {
+    let mut progs = Vec::new();
+
+    for prog in obj.progs() {
+        let Some(name) = prog.name().to_str() else {
+            continue;
+        };
+        /* feature-gated programs may not be loaded and then have no fd */
+        let Ok(fd) = prog.as_fd().try_clone_to_owned() else {
```
arighi: obj.progs() includes programs with autoload=false, whose libbpf fd remains -1. Calling prog.as_fd() constructs a BorrowedFd from that value and panics before try_clone_to_owned() can return an error. So I guess schedulers that have SEC("?…") programs should fail?

Can we call intead bpf_program__fd() directly, skip negative fds, and only then construct/duplicate the fd? And clone failures for loaded programs should probably be propagated rather than silently disabling monitoring.

rust/scx_arena/scx_arena/src/lib.rs
```diff
@@ -128,6 +128,131 @@ fn urcu_daemon(doorbell: libbpf_rs::MapHandle, pairs: Vec<(OwnedFd, OwnedFd)>) -
...
+            continue;
+        };
+        progs.push((name.to_string(), fd));
+    }
+
+    std::thread::Builder::new()
```
arighi: This detached thread never terminates and owns cloned fds for every program. If  we restart a scheduler in-process, each restart leaves another watcher polling the retired programs. The cloned program fds also keep their referenced arena maps/pages alive. Maybe the watcher could return a handle stored with the scheduler, with Drop signalling and joining the thread?

rust/scx_arena/scx_arena/src/lib.rs
htejun: Reworked as suggested: the fd is probed with bpf_program__fd() first, fd < 0 (not loaded) is skipped, a BorrowedFd is only constructed from a valid fd, and a clone failure for a loaded program now propagates instead of silently dropping the program from monitoring.

FWIW the -1 case didn't panic in practice: dup(-1) returned EBADF, which is how this first surfaced on the CI kernels via scx_lavd. But you're right that BorrowedFd from -1 is UB by contract, so it's gone.

rust/scx_arena/scx_arena/src/lib.rs
htejun: Restructured along that line, one step further: ArenaLib::init() is folded into setup(), which now returns an owned ArenaLib that holds both the stream watcher and the urcu reclaim daemon, each as a {stop eventfd, thread handle} pair. Dropping the handle signals the eventfds, both threads run a final drain/scan pass and exit, and the drop joins them.

Schedulers keep the handle alive for the scheduler instance's lifetime: p2dq per restart iteration, lavd and chaos as a scheduler struct field. A restart now tears down the old daemons and their cloned fds before spawning new ones against the new object.

arighi (approved):

PR #3772, opened 2026-08-27
scx_cosmos: Expand GPU locality hints to cgroup workloads

nvax-r:
## Summary

\NVML reports processes that own GPU contexts, but it does not describe the complete workload. For multiprocess services such as vLLM, NVML reports EngineCore while the API server and helper process remain invisible. Moving only EngineCore toward the GPU can therefore split the request path across NUMA nodes and regress latency.

Use an NVML-reported process as the seed for discovering its exact cgroup workload. Read cgroup.procs and extend the GPU-local NUMA hint to peer process in the same domain cgroup.

A six-pair AB/BA Qwen3-8B vLLM test used fresh services for every arm, 512 requests per arm, 512 input and 128 output tokens, 10 requests/s, and concurrency 32. All 6144 requests completed without failure.

Compared with the same scheduler without GPU awareness:

mean TTFT:  27.704 -> 22.708 ms  (-18.0%)
p95 TTFT:   32.723 -> 26.285 ms  (-19.7%)
p99 TTFT:   35.245 -> 27.920 ms  (-20.8%)
mean TPOT:   8.020 ->  7.332 ms  (-8.6%)
mean E2E:  1046.191 -> 953.897 ms (-8.8%)
throughput:  9.7968 -> 9.8156 req/s (+0.19%)

All six pairs improved. Every GPU-aware map snapshot contained the API, helper, and EngineCore TGIDs on node 0, while NVML directly reported only EngineCore. Earlier process-only testing regressed mean and p99 TTFT by 3.6% and 34.1%, respectively.

### Conservative choices
The initial implementation deliberately avoids guessing:
* Expand only exact cgroup-v2 membership. Child cgroups are not included.
* A direct NVML TGID receives a hint only when all its observed GPUs resolve to exactly one NUMA node.
* If a peer is observed in cgroups associated with conflicting nodes, omit its hint instead of selecting a node based on iteration order.
* Perform cgroup expansion only when the NVML snapshot completed successfully
* Keep --gpu-util-threshold > 0 process-only. Expanding a filtered process set could reintroduce peers intentionally excluded by the threshold.

### capacity policy
gpu_pid_map has 8,192 entries. The implementation does not partially fill the remaining capacity from HashMap iteration:

1. If the complete expanded set fits, use it.
2. If expansion does not fit but the complete direct-NVML set fits, use only the direct set.
3. If even the direct set does not fit, use an empty set.

## Testing
Comparing before/after the change , with `--gpu` enabled

| Metric | before | after | before v.s. after |
|---|---:|---:|---:|
| Mean TTFT | 29.197 ms | 22.708 ms | -22.2% |
| Average per-run p95 TTFT | 44.171 ms | 26.285 ms | -40.5% |
| Average per-run p99 TTFT | 50.329 ms | 27.920 ms | -44.5% |
| Mean TPOT | 7.511 ms | 7.332 ms | -2.4% |
| Mean E2E | 983.047 ms | 953.897 ms | -3.0% |
| Average per-run p99 E2E | 1047.374 ms | 988.946 ms | -5.6% |
| Throughput | 9.8140 req/s | 9.8156 req/s | +0.02% |

### DIFF
--- a/scheds/rust/scx_cosmos/src/main.rs
+++ b/scheds/rust/scx_cosmos/src/main.rs
@@ -10,7 +10,11 @@ pub use bpf_skel::*;
 pub mod bpf_intf;
 pub use bpf_intf::*;
 
+mod cgroup;
+mod gpu;
 mod stats;
+use cgroup::CgroupReader;
+
 use std::collections::{HashMap, HashSet};
 use std::ffi::{c_int, c_ulong};
 use std::fs::File;
@@ -527,10 +531,12 @@ struct Scheduler<'a> {
     stats_server: StatsServer<(), Metrics>,
     /// GPU device index -> NUMA node (for NVML PID sync). Only set when --gpu and NUMA enabled.
     gpu_index_to_node: Option<HashMap<u32, u32>>,
-    /// Previous (pid, node) set so we can remove PIDs that stopped using the GPU.
+    /// Previous (TGID, node) set so we can remove processes that stopped using the GPU.
     previous_gpu_pids: Option<HashMap<u32, u32>>,
     /// Reused NVML handle to avoid re-initializing on every sync (expensive).
     nvml: Option<Nvml>,
+    /// Host cgroup v2 reader used to discover peer processes of NVML GPU processes.
+    gpu_cgroup_reader: Option<CgroupReader>,
     /// Dynamic threshold state for perf event migrations (when --perf-threshold is 0/dynamic).
     perf_threshold_state: Option<DynamicThresholdState>,
     /// Dynamic threshold state for sticky perf events (when --perf-sticky-threshold is 0/dynamic).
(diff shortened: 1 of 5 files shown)


arighi:
cc: @ltrager

scheds/rust/scx_cosmos/src/main.rs
```diff
@@ -836,17 +866,32 @@ impl<'a> Scheduler<'a> {
...
-        // First collect pid -> set of nodes (GPUs) per process.
+        // First collect TGID -> set of nodes (GPUs) per process.
         let mut pid_to_nodes: HashMap<u32, HashSet<u32>> = HashMap::new();
+        let mut snapshot_complete = true;
 
-        let count = nvml.device_count().context("NVML device count")?;
+        let count = match nvml.device_count() {
+            Ok(count) => count,
+            Err(error) => {
+                self.reconcile_gpu_pid_hints(&HashMap::new())?;
```
arighi: Not a blocker, because it's probably a corner case, but should we retain the last successfully completed snapshot here instead of clearing the map? Before this change a device_count() or device_by_index() failure just returned, so existing GPU hints remained intact.

scheds/rust/scx_cosmos/src/main.rs
```diff
@@ -836,17 +866,32 @@ impl<'a> Scheduler<'a> {
...
-                None => continue,
+                None => {
+                    snapshot_complete = false;
+                    continue;
+                }
+            };
+            let device = match nvml.device_by_index(i) {
+                Ok(device) => device,
+                Err(error) => {
+                    snapshot_complete = false;
```
arighi: Can we log this error as we already do for process-query failures below? Something like "NVML device {i} lookup failed: {error:#}")" or similar.

arighi (approved):
The cgroup awareness looks really interesting. I left a couple of comments (not blockers, minor issues), but I like this, thanks!

scheds/rust/scx_cosmos/src/main.rs
nvax-r: It makes sense, thanks.
I agree `device_count()` / `device_by_index()` failures are observation miss, not evidence that GPU processes are gone.

I changed the behavior so on those kinds of failures we return without reconciling.

scheds/rust/scx_cosmos/src/main.rs
nvax-r: Changed, thanks.

arighi (approved):
Looks good to me, thanks!

PR #3778, opened 2026-08-31
common.bpf.h: probe the BPF prolog on the cid-form attach path too

devnexen:
scx_lib_init_probe hooks bpf_scx_reg(), but a cid-form scheduler enters through bpf_scx_reg_cid(), so the probe never fires and __scx_prolog_disables_migration keeps its default, leaving is_migration_disabled() to over-report on its slow path. Give that path its own probe. bpf_scx_reg_cid() only exists from v7.2, so the new program is "?" and the cid-form open paths switch it on.


multics69:
The changes make sense to me. And this should also be routed to the kernel tree. 
@htejun -- Could you take a look?

htejun (approved):

multics69:
Thanks a lot @devnexen !

PR #3782, opened 2026-09-01
scx_cosmos: Avoid redundant shared DSQ kicks

arighi:
Commit 59bb398f7af9 made every shared DSQ insertion kick the previous CPU. While this prevents constrained tasks from waiting indefinitely for an eligible CPU to inspect the shared queue, it also adds a kick to the hot enqueue path when ops.select_cpu() already supplied the wakeup side effect. This regresses 7-Zip (and other CPU-intensive benchmarks) on the DGX Spark.

Kick after shared DSQ insertion only for affinity-constrained or migration-disabled tasks, and when ops.select_cpu() was skipped. This retains the wakeup needed by constrained tasks without penalizing regular unbound wakeups.

Fixes: 59bb398f7af9 ("scx_cosmos: Avoid redundant kicks after local dispatch")

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -1034,6 +1034,20 @@ static bool task_should_migrate(struct task_struct *p, u64 enq_flags)
 }
 
 /*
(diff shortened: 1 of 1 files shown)


hodgesds (approved):
LGTM

sirlucjan (approved):

PR #3784, opened 2026-09-02
scheds/experimental: Introduce scx_cidland

arighi:
Add scx_cidland: a topology-aware scheduler written as a cid-form scheduler.

A cid (topological CPU ID) is an alternative way to address a CPU: each CPU gets an id from a dense space ordered by topology, so that the CPUs of a core, of an LLC and of a NUMA node always occupy contiguous ranges of that space.

The motivation for this scheduler is to explore what that property buys us in practice: a topology domain becomes a plain (base, len) slice of the cid space, so questions that normally need cpumask allocations and per-CPU topology lookups turn into range scans over a bitmap:

 - "is this core fully idle?" -> are all the bits in [core_base, core_base + core_nr) set?
 - "is there an idle CPU in my LLC?" -> scan [llc_base, llc_base + llc_nr)

Task affinities are handled in cid space as well: the cid form reports them through ops.set_cmask() as a cmask, the cid-space counterpart of a cpumask, which is copied into a per-task bitmap, so checking whether a task can run on a candidate cid is a bit test with no cid to CPU translation in the pick loop.

Design:

 - the per-cid core and LLC ranges are built once in ops.init() from scx_bpf_cid_topo(), walking the cid space backwards (since the cids of a domain are contiguous, seeing the highest cid of a domain first is enough to derive the length of the whole range),

 - idle state is tracked in a private bitmap of idle cids updated from ops.update_idle(),

 - on wakeup an idle cid is searched in this order: the previous cid if its whole core is idle, any fully idle core in the previous LLC, the previous cid, any idle cid in the previous LLC, any idle cid in the system (--primary-domain can restrict the preferred set of CPUs),

 - tasks that don't get an idle cid are queued to a single shared DSQ, consumed by the first cid that runs out of work, and ordered by the virtual deadline vruntime + burst_vruntime (same logic used by scx_cosmos), which keeps fairness while prioritizing tasks that sleep often and run in short bursts.


scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,902 @@
...
+/*
+ * Per-task context.
+ */
+struct task_ctx {
+	u64 last_run_at;		/* when the task last started running */
+	u64 last_woke_at;		/* when the task last woke up */
+	u64 burst_runtime;		/* runtime accumulated since the last sleep */
+	u64 wakeup_freq;		/* average wakeup frequency */
+	u64 allowed[MAX_CID_WORDS];	/* cids the task is allowed to run on */
+};
```
htejun, line 182: You can easily put this on arena too with `sdt_task`.

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,902 @@
...
+ * space and are guaranteed to be contiguous.
+ */
+struct cid_ctx {
+	u32 core_base;		/* first cid of the core this cid belongs to */
+	u32 core_nr;		/* number of cids (SMT siblings) in the core */
+	u32 llc_base;		/* first cid of the LLC this cid belongs to */
+	u32 llc_nr;		/* number of cids in the LLC */
+};
+
+static struct cid_ctx cid_ctxs[MAX_CIDS];
```
htejun: This and other globals too as `__arena_global`. It unburdens the verifier from needing to track the derefs so the actual logic code becomes a lot nicer.

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,902 @@
...
+{
+	u32 idx = (u32)cid;
+
+	if (cid < 0)
+		return NULL;
+
+	/* Keep the compiler from hoisting the address computation. */
+	barrier_var(idx);
+	if (idx >= MAX_CIDS)
+		return NULL;
```
htejun: Then, you wouldn't need the above. The cid ctx lookup can just be direct `cid_ctxs[cid]` when you know cid is not an error value.

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,902 @@
...
+
+	if (cid < 0)
+		return NULL;
+
+	/* Keep the compiler from hoisting the address computation. */
+	barrier_var(idx);
+	if (idx >= MAX_CID_WORDS)
+		return NULL;
+
+	return &mask[idx];
```
htejun: Ditto here. You would be able to drop all the unnecessary range checks.

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,902 @@
...
+	u64 mask = 1ULL << (cid & 63);
+	u64 *word = idle_word(cid);
+
+	if (!word)
+		return;
+
+	if (idle)
+		__sync_fetch_and_or(word, mask);
+	else
+		__sync_fetch_and_and(word, ~mask);
```
htejun: The downside is that it'd have to use xchg loop for the above atomic ops.

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,902 @@
...
+ * It's optimistic: if the claimed cid ends up with nothing to run, its idle bit
+ * is re-armed in ops.dispatch(), see cidland_dispatch().
+ */
+static bool cid_claim_idle(s32 cid)
+{
+	u64 mask = 1ULL << (cid & 63);
+	u64 *word = idle_word(cid);
+
+	if (!word)
+		return false;
```
htejun: and the cleanup propagates upward and you wouldn't need this either.

htejun (approved):

sirlucjan (approved):
Image

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
arighi: Yeah, but I was trying to reduce the amount of dependencies for the first version. And I guess this brings also the are allocator with lots of other stuff, right?

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
arighi: Let me try to go "full arena" :) I'm curious to see how this turns on, especially on arm64.

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,1075 @@
...
+}
+
+/*
+ * Task contexts are allocated from the arena so that @allowed is an arena
+ * pointer like @all_cids and @primary_cids, and the pick loop can take either
+ * without caring which one it got.
+ */
+static struct task_ctx __arena *try_lookup_task_ctx(const struct task_struct *p)
+{
+	return scx_task_data((struct task_struct *)p);
```
htejun, line 208: Minor but the only time `scx_task_data()` wouldn't return NULL from all the ops paths where the task is known to be between init and exit, so there's no need to worry about NULL task_ctx. And if by mistake NULL ctx gets accessed (e.g. from a TP w/o NULL test), the arena lib will auto-abort the scheduler with backtrace and error message.

htejun (approved):

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
htejun: Oh, this requires `ArenaLib::setup()`.

arighi:
@htejun I pushed a few more changes on top and it should be "full arena" now!

The atomic idle management turned out to be a bit more tricky than expected. I asked claude to fix it. In short, cid_claim_idle() needs the xchg loop, but so does cid_set_idle() even though its result is dead. LLVM still emits atomic64_fetch_or/_and there and writing the non-fetching instruction out in inline asm makes the verifier lose the arena type pointer. So both paths use compare-and-swap loops now. That requires LSE on arm64, so the scheduler won't even load on pre-LSE systems, but I guess it's not a big problem for now.

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
arighi: Good point, updated!

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
multics69: > The downside is that it'd have to use xchg loop for the above atomic ops.

If I understand correctly, we anyway need an xchg loop to support the concurrent call of cid_set_idle. So that's a BPF restriction, not the cid restriction, right? @htejun

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,902 @@
...
+ */
+static bool cid_claim_idle(s32 cid)
+{
+	u64 mask = 1ULL << (cid & 63);
+	u64 *word = idle_word(cid);
+
+	if (!word)
+		return false;
+
+	return __sync_fetch_and_and(word, ~mask) & mask;
```
multics69: And, this also needs an xchg loop, right?

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
```diff
@@ -0,0 +1,902 @@
...
+
+/*
+ * Return true if @cid is set in @mask.
+ */
+static bool cid_isset(u64 *mask, s32 cid)
+{
+	const u64 *word = cid_word(mask, cid);
+
+	return word && (*word & (1ULL << (cid & 63)));
+}
```
multics69: Better to move this upwards to collocate the cid-related utility functions?

multics69 (approved):
Thanks @arighi for sharing this!

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
arighi: Good point, moved up, close to the other cid-related functions. Thanks for taking a look!

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
arighi: Yes, both cid_set_idle() and cid_claim_idle() need a loop. That's the only thing that looks a bit ugly and I wish there was a "more atomic" way to express that in BPF. Tejun mentioned the BPF people may have an idea how to remove the atomic op restrictions, so we'll see.

And, thinking more about it, it might be worth moving all the cid-bitmap functions to a separate generic cid-bitmap.h header, so other schedulers can reuse the high-level abstractions without re-implementing the low-level cid-bitmap accesses.

arighi:
@multics69 I introduced some generic cid bitmap abstractions (scheds/include/scx/cid_bitmap.bpf.h), so other scheds can reuse them. I think it looks nicer now.

scheds/experimental/scx_cidland/src/bpf/main.bpf.c
htejun: Yeah, that's a limitation from the current arena implementation. BPF folks are saying that it can be lifted, so hopefully it will go away soon.

scheds/include/scx/cid_bitmap.bpf.h
```diff
@@ -0,0 +1,187 @@
...
+ * @nr: number of cids in the range
+ *
+ * The range must fit in the caller-provided storage. An empty range is full.
+ * Words are read independently, so the range as a whole is not observed
+ * atomically.
+ *
+ * Return: true if every cid in [@start, @start + @nr) is set, false otherwise.
+ */
+static __always_inline bool
+cid_bitmap_test_range_all_set(const u64 __arena *bits, u32 start, u32 nr)
```
htejun: Isn't this equivalent to `cmask_full_range()`? Why not just extend and use `cmask`?

arighi: The reason is that I overlooked the existing cmask helpers when implementing this. There's no fundamental reason for the separate bitmap abstraction. I'll switch to the cmask API.

arighi:
@htejun I updated the scheduler to use the cmask API and dropped the custom cid bitmap  stuff.

One caveat is that I had to add `scheds/include: Add an unchecked cmask test`. Using the checked `cmask_test()` in the hot scan path causes pretty severe performance regressions, because it repeats the active-range check for every candidate, even though the scan has already bounded the cid. Can you check if this makes sense to you?

With this version, I'm getting pretty much the same performance as scx_cosmos, which is really cool!

htejun:
Yeah, that's fine although I'm surprised that those checks can cause severe perf issues. Can you share a bit more on the test scenario and behavior?

arighi:
> Yeah, that's fine although I'm surprised that those checks can cause severe perf issues. Can you share a bit more on the test scenario and behavior?

I've done these tests on a Vera 176 cores system with SMT (352 CPUs), 2 NUMA nodes. I've used an internal GEMM benchmark, but I was getting similar results also with https://github.com/OpenMathLib/OpenBLAS.git, running something like this:
```
  make -C benchmark sgemm.goto \
      TARGET=ARMV8SVE \
      USE_OPENMP=1 \
      NUM_THREADS=176 \
      NOFORTRAN=1

  env OMP_NUM_THREADS=88 \
      OMP_DYNAMIC=false \
      OPENBLAS_LOOPS=10 \
      OPENBLAS_PARAM_M=16384 \
      OPENBLAS_PARAM_N=16384 \
      OPENBLAS_PARAM_K=16384 \
      numactl -C 0-87,176-263 --membind=0 \
      ./benchmark/sgemm.goto 1 1 1
```

I'm forcing the workload to use only node 0 and one task per SMT core. Adding the cmask checks was introducing a 10-16% performance drop. I'll re-run the tests to confirm.

arighi:
@htejun I repeated my tests and now I can't reproduce the performance drop anymore... maybe it was some background noise in the system. So, ignore what I said about performance.

That said, I'd still keep the __cmask_test() change, it's still removing some unnecessary BPF instructions and it shouldn't cost too much to provide also this unchecked API.

PR #3789, opened 2026-09-03
scx_cosmos: Contain tasks in the primary domain unless it is overloaded

arighi:
When a primary domain is defined, pick_idle_cpu() currently falls back to any idle CPU in the system when no CPU in the primary domain is idle.

This immediate spill is harmful for bursty CPU-intensive workloads. For example, transient saturation caused by barrier wakeups in a parallel job can bounce tasks to non-primary CPUs, only for them to be pulled back into the primary domain almost immediately when the primary domain is not overloaded.

Strict containment has the opposite problem: when the primary domain is genuinely overloaded, non-primary CPUs are neither picked nor kicked. Excess tasks remain queued in the shared DSQs even when idle CPUs are available outside the primary domain.

Find a middle ground by containing tasks in the primary domain and allowing them to spill to non-primary CPUs only when the primary domain is persistently overloaded. This borrows the idea behind the EAS overutilized flag in the fair scheduler class, where preferred placement is abandoned once the system crosses an overload threshold. Here, queueing delay is used as the overload signal:

 - ops.dispatch() timestamps, per NUMA node, the last time the node's shared DSQ was observed empty,

 - when a task is queued to a shared DSQ that has been continuously backlogged for longer than a threshold (option --primary-overload-us, default 1ms), the node is marked overloaded for a short decay window (renewed by further backlogged enqueues) and an idle non-primary CPU is kicked to start draining the backlog,

 - while a node is overloaded, pick_idle_cpu() is allowed to pick idle non-primary CPUs again.

Queueing delay is preferred over CPU utilization as the overload signal: full CPU utilization without queueing is not overload and should not immediately trigger spilling. Once tasks start accumulating in the shared DSQs, however, waiting can cost more than running on a less-preferred CPU. The decay window provides the hysteresis needed to avoid ping-pong around the overload threshold. Tasks that cannot run on any primary CPU are exempt from containment.

### DIFF
--- a/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
+++ b/scheds/rust/scx_cosmos/src/bpf/main.bpf.c
@@ -50,6 +50,13 @@ char _license[] SEC("license") = "GPL";
 private(COSMOS) struct bpf_cpumask __kptr *primary_cpumask;
 
 /*
+ * Complement of @primary_cpumask (only initialized when a primary domain
+ * is defined): CPUs that can be used only when the primary domain is
+ * overloaded.
+ */
+private(COSMOS) struct bpf_cpumask __kptr *nonprimary_cpumask;
+
+/*
  * Set to true when @primary_cpumask is empty (primary domain includes all
  * the CPU).
  */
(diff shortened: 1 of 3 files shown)


RitzDaCat:
Doing a quick spotcheck now. Was able to pull down and build correctly and test the new flag. 
image

RitzDaCat (approved):
Was able to pull down and test. Did a quick audit against Kovaaks and my usual benchmarks and had no issues or concerns.  Looks good to me.

sirlucjan (approved):

PR #3792, opened 2026-09-06
scx_lavd: make the duty-cycle ravgs capacity- and frequency-invariant

multics69:
taskc->avg_util_ravg and cpuc->avg_util_ravg accumulate against scx_bpf_now(), so a duty cycle measures time spent rather than work done. The same work looks like more utilization on a little core at a low frequency than on a big core at full tilt, leaving util_est incomparable across CPUs.

We solve this by scaling the clock using rq->clock_pelt, which advances in proportion to the CPU's capacity and frequency. util_est is now a fraction of maximum capacity, so readings on a little core shrink by the capacity ratio.

However, switching to clock_pelt is challenging because clock_pelt is per-rq and not comparable across CPUs. Hence, we create ravg_data_invr which pairs a ravg_data with the CPU its timestamps belong to. An entity arriving from another CPU is rebased -- only the timestamp moves, as in attach_entity_load_avg().

The task average keeps counting runnable time, as it always has: like the kernel's runnable_avg, that is what lets the queued-load sum see oversubscription, which running time alone cannot since it saturates at capacity. But it was refreshed only at ops.quiescent(), so a task that never slept never updated its util_est. Move the update to ops.stopping(), marking the task off only when it is no longer runnable; every deschedule now refreshes util_est.

ops.running() and ops.stopping() can fire off-rq, where clock_pelt is unreadable, so those paths drop the anchor and let the next local update rebase.

A CPU going offline drops its average. A forked task drops the copy inherited from its parent: past CPU consumption is not a heritable property like lat_cri, and the kernel likewise does not seed a new task's util from its parent.

All four util_est consumers take the new units unchanged: the two preemption_vulnerability() sites (DSQ routing and idle CPU selection), the queued-load sums behind task_load_metric(), and the vuln_thresh control loop -- which already divides util_sum by cap_sum, so it was expecting invariant units all along.

Measured on a Snapdragon chipset (little cap 216, big 803-1024) with one task doing identical work per period, pinned to each cluster in turn. Before, it was routed to the steady DSQ on the little cluster and to the turbulent one on the big -- same task, opposite placement. After, the vulnerability score is identical on both clusters and so is the decision. On a 192-CPU homogeneous x86 machine, where both terms are identity, before and after are indistinguishable, as expected.


scheds/rust/scx_lavd/src/bpf/util.bpf.h
```diff
@@ -54,6 +54,139 @@ static __always_inline bool is_rt_or_dl_task_running(s32 cpu)
...
+ * carries up to half of what ravg_read() returns, so clearing it would erase
+ * the recent history on every migration.
+ *
+ * Operates on a stack copy; the callers below own the arena bounce.
+ */
+static __always_inline void
+ravg_accumulate_anchored(struct ravg_data *rd, u64 val, u64 pelt_now,
+			 bool anchored)
+{
+	if (unlikely(!anchored || !rd->val_at)) {
```
daidavid, line 99: How are you testing that the util reported matches the task's actual workload? I don't see how not decaying previously accumulated load wouldn't lead to large inaccuracies when tasks migrate.

Given that clock_pelt can be unavailable or hasn't advanced yet, we can use the wall_clock and decay at 1024, then accumulate at some ratio between clock_pelt/clock_task gotten from active runtimes. The rq_clock can also drift small amounts, but it should be an improvement over not decaying the load when tasks migrate.

Accumulating <1024 also has some tradeoffs. For example, the math works out such that for a task running at 100% on a smaller capacity CPU means it will never converge past the capacity of the CPU. Accumulating clock_pelt will eventually allow tasks to reach maximum load, the practical effect of is that there's a smaller frequency response when migrating to a larger CPU, or it takes longer to ramp up to its "true" size. 

This is a classic DVFS problem, since there simply isn't enough capacity/time available on the CPU to help measure the size of the task. Overshooting is better for perf, but not so great for battery life.

PR #3793, opened 2026-09-06
scx_lavd: fix and rate-limit the CPU performance target for schedutil governor

multics69:
lavd drives cpufreq through scx_bpf_cpuperf_set(). The value it passed was
recomputed redundantly, and sent unfiltered. What started this was
schedutil's per-policy kthread (sugov:N) consuming a noticeable amount of
CPU time (3-5% of a single CPU) on an asymmetric arm64 chipset. This affects
only systems running the schedutil governor.

1/4 fixes the scale. scx_bpf_cpuperf_set() takes a target on the
system-wide capacity scale, where SCX_CPUPERF_ONE is the most capable CPU,
but lavd passed the CPU's own wall-clock utilization ratio -- the same
thing only when that CPU's capacity is SCX_CPUPERF_ONE. A little core
therefore over-asks and pins at its top frequency well below full
utilization.

2/4 stops double-counting. schedutil treats the value as fair-class
utilization and adds RT, DL and IRQ on top, but lavd's utilization is all
non-idle time and already included them, so the request was inflated by
their share.

3/4 removes redundant work. The target was recomputed on every
ops.running() from inputs the sys_stat timer only refreshes once per
interval. Computation moves to that timer; ops.running() just commits,
which it must, since schedutil ignores an update made from outside the
policy.

4/4 filters. Any change of the computed target, down to a single unit,
reached the governor, so utilization jitter around a steady load did too --
and on a slow-switching driver every update wakes a SCHED_DEADLINE kthread
with an IPI.

Measured on a Snapdragon chipset, at the same load on the little cluster:
requested frequency 2089 -> 1250 MHz, and scx_bpf_cpuperf_set() calls
551.7/s -> 225.1/s.

### DIFF
--- a/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
+++ b/scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
@@ -104,6 +104,8 @@ enum consts_internal {
 
 	LAVD_CPU_UTIL_MAX_FOR_CPUPERF	= p2s(85), /* 85.0% */
 	LAVD_CPU_UTIL_THR_FOR_MAX_FREQ	= p2s(80), /* cpu utilization threshold to update max freq */
+	LAVD_CPUPERF_UP_SHIFT		= 6, /* raise in steps of capacity / 64 */
+	LAVD_CPUPERF_DOWN_SHIFT		= 5, /* lower only past a capacity / 32 deadband */
 
 	LAVD_CC_REQ_CAPACITY_HEADROOM	= p2s(25), /* 25%: inflate required capacity by 25% to handle sudden spikes */
 	LAVD_CC_PER_CPU_UTIL		= p2s(50), /* 50%: maximum per-CPU utilization */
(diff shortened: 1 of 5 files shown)


daidavid:
The fixes to capacity scaling and util LGTM, how we calculate util for IRQ/steal will differ from how kernel calculates it. Not sure how much of that will actually matter in practice, in theory we can configure ravg to accumulate/decay the load in a similar cadence(32ms half-life).

As for the perf update rate, that doesn't seem like it should be a scheduler policy decision. Why isn't "rate_limit_us" offered by schedutil sufficient? Is there some other overhead that doesn't involve cpufreq updates?

multics69:
> As for the perf update rate, that doesn't seem like it should be a scheduler policy decision. Why isn't "rate_limit_us" offered by schedutil sufficient? Is there some other overhead that doesn't involve cpufreq updates?

I found that issuing a frequency change from the scx side can cause a measurable overhead in some CPU frequency driver implementations (e.g., Qualcomm) involving locking, IPI, and waking a DL task. I am also looking for driver-side optimization opportunities. However, there could be plenty of non-optimal driver implementations out there. :-)

daidavid (approved):

PR #3795, opened 2026-09-06
scx_mitosis: Assorted fixes for fairness and vtime errors

kkdwvd:
Various fixes that address a set of bugs affecting vtime calculation and task fairness. See commit logs for details.

### DIFF
--- a/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
+++ b/scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
@@ -274,11 +274,15 @@ static inline int update_task_cpumask(struct task_struct *p, struct task_ctx *tc
 		if (!(cpu_ctx = lookup_cpu_ctx(cpu)))
 			return -ENOENT;
 
-		tctx->dsq = get_cpu_dsq_id(cpu);
-		if (dsq_is_invalid(tctx->dsq))
+		dsq_id_t cpu_dsq = get_cpu_dsq_id(cpu);
+
+		if (dsq_is_invalid(cpu_dsq))
 			return -EINVAL;
 
-		scx_bpf_task_set_dsq_vtime(p, READ_ONCE(cpu_ctx->vtime_now));
+		/* Only re-base when the task actually changes domain. */
+		if (cpu_dsq.raw != tctx->dsq.raw || tctx->all_cell_cpus_allowed)
+			scx_bpf_task_set_dsq_vtime(p, READ_ONCE(cpu_ctx->vtime_now));
+		tctx->dsq = cpu_dsq;
 		tctx->all_cell_cpus_allowed = false;
 		return 0;
 	}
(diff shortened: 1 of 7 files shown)


scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -308,6 +308,24 @@ static inline int update_task_cpumask(struct task_struct *p, struct task_ctx *tc
...
 }
 
+/*
+ * A pinned task is placed on a per-CPU DSQ and takes its basis from that CPU's
+ * clock, which belongs to whichever subcell owns the CPU. Charge that subcell
+ * rather than the task's own, so stopping() advances the clocks the task is
+ * compared against. Otherwise a task pinned outside its subcell never advances
+ * them and its vtime ratchets ahead until it loses every comparison.
+ */
+static __always_inline int set_vtime_charge_cpu(struct task_ctx *tctx, struct cpu_ctx *cpu_ctx)
```
dforsyth: If this is meant to only be used for pinned tasks, can we either make sure it's only ever called that way with hard validation or at least just update the name of the function to make it more clear.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -849,11 +851,14 @@ void BPF_STRUCT_OPS(mitosis_enqueue, struct task_struct *p, u64 enq_flags)
...
-			      p->scx.dsq_vtime, basis_vtime, p->scx.dsq_vtime - basis_vtime, tctx->cell, tctx->subcell);
-		return;
-	}
+	/*
+	 * A task can legitimately owe at most one weighted slice over the clock
+	 * it competes against; anything beyond that is an accounting gap (a
+	 * clock that did not follow the task) and would only starve the task.
+	 * Cap the debt instead of carrying or aborting on it.
+	 */
+	if (time_after(vtime, basis_vtime + VTIME_DEBT_CAP_SLICES * slice_ns))
```
dforsyth: Can we just wrap this calc in a function or macro?

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
```diff
@@ -942,12 +946,23 @@ void BPF_STRUCT_OPS(mitosis_dispatch, s32 cpu, struct task_struct *prev)
...
+	 * of the queue, so also bound how long the per-CPU DSQ goes unserved.
+	 */
 	p = dsq_peek(cpu_dsq.raw);
-	if (p && (!found || time_before(p->scx.dsq_vtime, min_vtime))) {
-		min_vtime = p->scx.dsq_vtime;
-		min_vtime_dsq = cpu_dsq;
-		found = true;
+	if (p) {
+		u64 since = READ_ONCE(cctx->pinned_waiting_since);
+		bool overdue = since && time_delta(scx_bpf_now(), since) > PINNED_MAX_WAIT_SLICES * slice_ns;
```
dforsyth: same re: put calcs into functions or macros

dforsyth (approved):
LFTM, just style nits. I'll let you decide how much of it you want to address.

Also, I don't mind the formatting changes, but let's try to put then in their own PRs.

scheds/rust/scx_mitosis/src/bpf/mitosis.bpf.c
htejun: I'm not sure it's an improvement to encapsulate an one liner logic in a macro or function when the comment above is already providing enough context. What would it add? You now have to come up with a name which is specific enough, determine where to place that function and jump to a different spot to read the actual logic. Sure, it's used twice but it's in succession and obvious.

In general, unless you can provide strong enough rationale, or something is clearly out of line with the surrounding code, I generally think it's more productive to leave minor style differences to whoever is driving the bulk of development. It *really* doesn't matter one way or another.

PR #3800, opened 2026-09-09
scx_cidland: Port fair/EEVDF to sched_ext

arighi:
## What this is

A port of the kernel's EEVDF/`fair.c` to `sched_ext`, built on top of the existing `scx_cidland`.

The placement and the fairness rules are `fair.c`'s: tasks are ordered by the virtual deadline `vd_i = ve_i + r_i / w_i` against a per-runqueue reference `V`, lag is captured at dequeue and spent at placement, runtime is charged through `calc_delta_fair()`, and the wakeup preemption, the idle search, the utilization tracking and the cpufreq hand-off follow the functions they are named after.

It is **not** a complete 1:1 port:

- The EEVDF core is faithful, minus `DELAY_DEQUEUE`, the hrtick (`sched_ext` has none, so a slice that has been granted is only revisited at the tick) and the eligibility filter among tasks already queued: a running task that has had its share is dropped from the wakeup pick the way `pick_eevdf()` drops it, but a DSQ is ordered by its key and dispatch takes the head, so an over-served task already in the queue is not skipped.
- Load balancing goes by queue depth, not by measured load: there is no `load_avg` equivalent yet, no group classification, no active balancing.
- `wake_affine_weight()`, `sched_idle_cpu()`, uclamp on the placement side are not there.
- `cpu.max`, `cpu.idle`, CPU hotplug, NUMA balancing, EAS and core scheduling are not implemented. `cpu.weight` is honored, though as a per-task weight rather than a share of the machine divided among a cgroup's members.

The README has the full list under "What is missing", so that anyone picking one up knows what they are signing for.

## Why

The goal is to have a baseline playground: a scheduler close enough to the upstream one that changes can be applied to it incrementally and measured against the default kernel scheduler.

## Results

The goal here is to stay as close as possible to fair/EEVDF.

Measured against the kernel's own EEVDF on an i7-13800H, 20 CPUs, 1 node, 1 LLC, HZ=1000, Linux v7.2.2.

In the Delta column, `^` is better than EEVDF, `v` worse, `~` within run-to-run noise (the two schedulers' ranges overlap).

**Latency** (default cpufreq: intel_pstate active, powersave; median of 5)

```
+------------------------------------------------+---------+------------+-------------+------------+
| Metric                                         | Unit    | EEVDF/fair | scx_cidland | Delta      |
+================================================+=========+============+=============+============+
| schbench, light  (2 msg x 10 wrk, 200 rps)                                                       |
|   wakeup latency p50                           | us      |         76 |          76 |    +0.0% ~ |
|   wakeup latency p90                           | us      |        135 |         169 |   +25.2% ~ |
|   request latency p50                          | us      |       4328 |        4552 |    +5.2% v |
|   request latency p99                          | us      |      14352 |       17824 |   +24.2% v |
| schbench, saturated (4 msg x 20 wrk, 1000 rps)                                                   |
|   wakeup latency p50                           | us      |        112 |          70 |   -37.5% ~ |
|   wakeup latency p90                           | us      |      33216 |        3924 |   -88.2% ^ |
|   request latency p50                          | us      |      15632 |       31392 |  +100.8% v |
|   request latency p99                          | us      |     196352 |      203008 |    +3.4% v |
+------------------------------------------------+---------+------------+-------------+------------+
```

**Throughput** (median of 5)

```
+------------------------------------------------+---------+------------+-------------+------------+
| Metric                                         | Unit    | EEVDF/fair | scx_cidland | Delta      |
+================================================+=========+============+=============+============+
| perf bench sched messaging -t -g 4 -l 2000     | s       |      0.255 |       0.249 |    -2.4% ~ |
| perf bench sched pipe -l 1000000               | ops/s   |     573712 |      626305 |    +9.2% ~ |
| stress-ng --cpu 20, 15 s                       | bogo/s  |      28195 |       27798 |    -1.4% ~ |
| 20 spinners, aggregate CPU utilization         | %       |      99.95 |       99.95 |    +0.0% ~ |
+------------------------------------------------+---------+------------+-------------+------------+
```

**stress-ng** (median of 5; the loaded cyclic rows median of 3)

```
+------------------------------------------------+---------+------------+-------------+------------+
| Metric                                         | Unit    | EEVDF/fair | scx_cidland | Delta      |
+================================================+=========+============+=============+============+
| stress-ng --cpu 20, 15 s                       | bogo/s  |      28195 |       27798 |    -1.4% ~ |
| stress-ng --sock 8, 8 s                        | bogo/s  |      16384 |       12646 |   -22.8% v |
| stress-ng --msg 8, 8 s                         | bogo/s  |   15214034 |    12689846 |   -16.6% v |
| stress-ng --sem 8, 8 s                         | bogo/s  |   46894116 |    39030748 |   -16.8% v |
| stress-ng --yield 8, 8 s                       | bogo/s  |   20776179 |    15664288 |   -24.6% v |
+------------------------------------------------+---------+------------+-------------+------------+
```

```
+------------------------------------------------+---------+------------+-------------+------------+
| Metric                                         | Unit    | EEVDF/fair | scx_cidland | Delta      |
+================================================+=========+============+=============+============+
| stress-ng --cyclic, idle machine                                                                 |
|   scheduling latency p50                       | ns      |      57529 |       57159 |    -0.6% ~ |
|   scheduling latency p90                       | ns      |      66564 |       58949 |   -11.4% ^ |
|   scheduling latency p99                       | ns      |      70478 |       60145 |   -14.7% ^ |
|   scheduling latency p99.9                     | ns      |     133822 |       69722 |   -47.9% ^ |
| stress-ng --cyclic against 20 CPU hogs on 20 CPUs                                                |
|   scheduling latency p50                       | ns      |      52783 |       55025 |    +4.2% v |
|   scheduling latency p90                       | ns      |      53349 |       56479 |    +5.9% v |
+------------------------------------------------+---------+------------+-------------+------------+
```

**Power and frequency** (schedutil, intel_pstate passive; median of 3)

```
+------------------------------------------------+---------+------------+-------------+------------+
| Metric                                         | Unit    | EEVDF/fair | scx_cidland | Delta      |
+================================================+=========+============+=============+============+
| idle, 10 s                                     | PkgWatt |       1.19 |        1.23 |    +3.4% ~ |
| wakeup storm (schbench -L, 50k rps)            | PkgWatt |       3.16 |        3.10 |    -1.9% ~ |
| full load (stress-ng -c 0)                     | PkgWatt |      67.77 |       71.81 |    +6.0% v |
| idle, 10 s                                     | Avg_MHz |          4 |           5 |   +25.0%   |
| wakeup storm                                   | Avg_MHz |         42 |          42 |    +0.0%   |
| full load                                      | Avg_MHz |       3361 |        3345 |    -0.5%   |
+------------------------------------------------+---------+------------+-------------+------------+
```

## Reviewing this PR

66 commits, ~4k insertions. It is realistically not reviewable end to end, and most of it does not need to be. Most of commits touch nothing outside `scheds/experimental/scx_cidland/` and one more only adds a `libc` line for it to the root `Cargo.lock`.

Four commits touch shared code. Two are worth another pair of eyes:

- **`lib: Introduce bpf_arena_for()`** — @htejun. A `may_goto` loop whose counter the verifier tracks as a precise constant never converges; keeping the counters in arena memory instead makes every read an unknown scalar.  `bpf_for()` is the right answer wherever the cost does not matter, but it calls `bpf_iter_num_next()` on every iteration, which is too much in the loops that run on every wakeup. New header, `scheds/include/lib/arena_loop.h`, in `lib/` rather than `scx/` so only the schedulers that want it get the global.

- **`lib/ravg: Copy to and from the arena field by field`** — @etsal @multics69. Required to build the scheduler on Ubuntu: up to clang 19 the struct assignment in `ravg_to_arena()`/`ravg_from_arena()` becomes a memcpy that is expanded after the arena pass, losing the address space cast, and the verifier rejects the load. Copying the four fields one at a time keeps the cast on every compiler. I do not expect a regression — clang was already expanding it to the same four load/store pairs, and the disassembly of the whole scheduler is identical before and after — but it is a shared header, so it is worth a second look.

The other two shared commits are small and self-explanatory:

- `lib/ravg: Drop the arena headers it does not use` — removes two unused includes from `lib/ravg.bpf.c` that leave an unresolved `.addr_space.1` extern behind and break skeleton generation for any scheduler linking it.
- `scx: Add word access helpers to the BPF cmask API` — @htejun. Three inline helpers in `scheds/include/scx/cid.bpf.h` for scanning a cmask a word at a time, which `scx_cosmos` currently open codes.

Everything else is self-contained in `scx_cidland`.


sirlucjan (approved):
Tested, no issues. Good job!

htejun:
@arighi, could we use a writable volatile zero for `bpf_arena_for()`?

```c
volatile u32 loop_zero = 0;

/*
 * The body can make the counter precise, so an unknown initializer alone does
 * not stop its bounds from shifting on each iteration. Reload loop_zero in the
 * increment to break that progression.
 */
#define bpf_arena_for(var, start, end) \
	for (var = (start) + loop_zero; var < (end) && can_loop; var += 1 + loop_zero)
```

`loop_zero` must be non-const so the verifier cannot resolve its value.
Using it only in the initializer still hits the jump-sequence limit in
`cidland_dispatch()`.

With this variant, all 21 scx_cidland BPF programs passed load verification
with veristat on my 7.2-rc6 test kernel, using clang 22.1.8.

arighi:
> @arighi, could we use a writable volatile zero for `bpf_arena_for()`?
> 
> ```c
> volatile u32 loop_zero = 0;
> 
> /*
>  * The body can make the counter precise, so an unknown initializer alone does
>  * not stop its bounds from shifting on each iteration. Reload loop_zero in the
>  * increment to break that progression.
>  */
> #define bpf_arena_for(var, start, end) \
> 	for (var = (start) + loop_zero; var < (end) && can_loop; var += 1 + loop_zero)
> ```
> 
> `loop_zero` must be non-const so the verifier cannot resolve its value. Using it only in the initializer still hits the jump-sequence limit in `cidland_dispatch()`.
> 
> With this variant, all 21 scx_cidland BPF programs passed load verification with veristat on my 7.2-rc6 test kernel, using clang 22.1.8.

That is so much nicer! I pushed a new version that uses a u32 volatile and it seems to work (also on Ubuntu - with clang-19). Can you take a look now? Thanks!

scheds/include/lib/arena_loop.h
```diff
@@ -0,0 +1,38 @@
...
+	({								\
+		_Static_assert(sizeof(var) <= sizeof(u32),		\
+			       "bpf_arena_for() index must fit in u32");	\
+		(start) + bpf_arena_loop_zero;				\
+	})
+
+#define bpf_arena_for(var, start, end)					\
+	for (var = __bpf_arena_loop_start(var, start);			\
+	     var < (end) && can_loop;					\
+	     var += 1 + bpf_arena_loop_zero)
```
htejun: Andrii says `var += one` with `volatile one = 1;` works too and likely reads more natural.

htejun (approved):

scheds/include/lib/arena_loop.h
arighi: @htejun nice, I like that! And it even produces less code, tested both with clang-22 (CachyOS) and clang-19 (Ubuntu):
```
               insns         insns
clang 22       16987         16947        .bss -> .data   90 -> 63
clang 19       17591         17529        .bss -> .data   91 -> 63
```

Say thanks to Andrii. :)

PR #3803, opened 2026-09-11
lavd, lib:replace legacy scx-agnostic files with libarena

etsal:
The existing lib/ directory has two kinds of files: Generic data structures and functionality (e.g., a slab allocator), and scx-specific files (e.g., cgroup_bw, atq). The generic data structures have been superseded by more actively managed versions hosted in the libarena project, itself part of the Linux kernel repo. These more centralized libarena versions are used by multiple BPF-based projects and as such are more actively maintained and exercised.

Directly use the libarena version of the generic BPF arena code. The code is linked as a submodule that points to a git repo with an exported version of the code. The repo includes two versions of the libarena code, the kernel version that only maintains compatibility with the most up-to-date kernel, and a compat version modified for backwards compatibility. This allows us to keep track of code patterns that are only for compatibility reasons and remove them over time. For scx, we use the compat version of the repo.

The main change to the existing scx codebase is the use of a generic allocator. Out of all the allocators currently in the repo, only the slab allocator is in use. The slab allocator requires defining a different allocator for each data structure, along with embedding special id fields on each data structure. The slab allocators also have to be individually initialized before starting the scheduler. All the allocators get replaced with a single generic allocator that provides a conventional malloc/calloc/free interface.

Apart from that, most other changes are related to replacing the generic data structures in this repo with the libarena versions. These are just the bitmap and rbtree data structures.


htejun:
There are several problems with the current PR:

- CI doesn't initialize the new submodule. The [unit-test job](https://github.com/sched-ext/scx/actions/runs/34558426594/job/103135959327) fails building `scx_arena` because `../../../libarena/src/bitmap.bpf.c` is missing.

- The generated `scx_arena` crate archive contains none of the libarena sources or headers, but its `build.rs` still references `../../../libarena/`. Initializing the submodule fixes the checkout build, but doesn't make the published crate self-contained.

- With the submodule initialized, `scx_arena_selftests --test atq topology` fails during BPF loading on a 7.2-rc6 VM, before any selected test runs. `rb_integrity_check()` passes the address of a stack local to `rb_print_pop_up()`, whose first argument has `__arg_arena`. The verifier rejects it with `R1 is not a pointer to arena or scalar.`

- ATQ creates its tree with `RB_NOALLOC`, but `scx_atq_pop()` calls the imported `rb_pop()`, which returns `-EINVAL` for a nonempty tree unless it uses `RB_ALLOC`. This is from code inspection. The selftest load failure prevented runtime confirmation of the pop failure.

I'd suggest restructuring this as a staged migration with libarena exposed as a Cargo dependency:

1. Introduce `libarena-rs`, following the `libbpf-rs` model, for generic Rust arena support. Have it package the upstream libarena BPF sources and headers and provide build support for consumers to use them. Published crates should obtain these files through their declared dependencies.

2. Fork `scx_arena` as, say, `scx_libarena`, built on `libarena-rs`. Keep the scx-specific Rust and BPF code in this crate, including its own BPF sources and headers. Using a distinct package name lets both crates remain members of the same Cargo workspace. Different major versions alone don't allow duplicate package names among workspace members.

3. Switch schedulers one by one, converting each scheduler's Rust dependency, BPF sources and includes, and arena initialization together. Preserve the existing `scx_arena` and arena files in `lib/` and `scheds/include/lib/` so changes to the new implementation don't affect unmigrated schedulers. Remove the legacy implementation after the last scheduler migrates.

`scx_cargo` would remain the common BPF build infrastructure: shared sched_ext and vmlinux headers, compilation and linking, and Rust binding and skeleton generation. The arena crates would supply their library-specific sources, headers, and build requirements. `scx_cargo` would retain the legacy arena headers until migration finishes.

etsal:
Thank you for the review. TL;DR let's go with the staggered approach.

> There are several problems with the current PR:
> 
> * CI doesn't initialize the new submodule. The [unit-test job](https://github.com/sched-ext/scx/actions/runs/34558426594/job/103135959327) fails building `scx_arena` because `../../../libarena/src/bitmap.bpf.c` is missing.
> * The generated `scx_arena` crate archive contains none of the libarena sources or headers, but its `build.rs` still references `../../../libarena/`. Initializing the submodule fixes the checkout build, but doesn't make the published crate self-contained.

Not sure why the submodule doesn't get initialized, and the ../../../ can be symlinked the same way we include lib/ nowadays. That's neither here nor there if we split into crates.

> * With the submodule initialized, `scx_arena_selftests --test atq topology` fails during BPF loading on a 7.2-rc6 VM, before any selected test runs. `rb_integrity_check()` passes the address of a stack local to `rb_print_pop_up()`, whose first argument has `__arg_arena`. The verifier rejects it with `R1 is not a pointer to arena or scalar.`
> * ATQ creates its tree with `RB_NOALLOC`, but `scx_atq_pop()` calls the imported `rb_pop()`, which returns `-EINVAL` for a nonempty tree unless it uses `RB_ALLOC`. This is from code inspection. The selftest load failure prevented runtime confirmation of the pop failure.
> 

This is fallout from the bulk refactor that weren't triggered by my selftests/smoke tests, will fix in the final version.

> I'd suggest restructuring this as a staged migration with libarena exposed as a Cargo dependency:
> 
> 1. Introduce `libarena-rs`, following the `libbpf-rs` model, for generic Rust arena support. Have it package the upstream libarena BPF sources and headers and provide build support for consumers to use them. Published crates should obtain these files through their declared dependencies.
> 2. Fork `scx_arena` as, say, `scx_libarena`, built on `libarena-rs`. Keep the scx-specific Rust and BPF code in this crate, including its own BPF sources and headers. Using a distinct package name lets both crates remain members of the same Cargo workspace. Different major versions alone don't allow duplicate package names among workspace members.

This seems tractable, provided there is a way to a) expose headers from the Cargo dependencies and include them in the repo here, and b) we can do the same with the BPF *.c files and still compile them here. Both seem possible but I haven't tried them yet.

> 3. Switch schedulers one by one, converting each scheduler's Rust dependency, BPF sources and includes, and arena initialization together. Preserve the existing `scx_arena` and arena files in `lib/` and `scheds/include/lib/` so changes to the new implementation don't affect unmigrated schedulers. Remove the legacy implementation after the last scheduler migrates.
> 
> `scx_cargo` would remain the common BPF build infrastructure: shared sched_ext and vmlinux headers, compilation and linking, and Rust binding and skeleton generation. The arena crates would supply their library-specific sources, headers, and build requirements. `scx_cargo` would retain the legacy arena headers until migration finishes.

htejun:
We can follow what `scx_cargo` already does for headers: embed an archive with `include_bytes!()`, extract it into the scheduler's build output directory, and add that directory to the compiler's include paths. The libbpf tooling uses the same idea: `libbpf-sys` exposes bundled headers through `API_HEADERS`, and `libbpf-cargo` writes them out for compilation.

For `libarena-rs` and `scx_libarena`, the bundles would contain both headers and BPF `.c` files. Their build helpers would extract the files and pass the include directories and source paths to `scx_cargo`, which already compiles and links supplied sources. The extracted files can stay in the build directory.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
```diff
@@ -578,14 +584,14 @@ extern const volatile u8	verbose;
 #define debugln(fmt, ...)						\
 ({									\
 	if (verbose > 0)						\
-		bpf_printk("[%s:%d] " fmt, __func__, __LINE__,		\
+		arena_stderr("[%s:%d] " fmt, __func__, __LINE__,		\
```
arighi: Is arena_stderr() a no-op? Are we shutting down LAVD's --verbose logging with this?

lib/sdt_task.bpf.c
```diff
@@ -41,37 +39,42 @@ void __arena *scx_task_alloc(struct task_struct *p)
...
 int scx_task_init(__u64 data_size, __u64 align)
 {
-	return scx_alloc_init(&scx_task_allocator, data_size, align);
+	/*
+	 * @align is satisfied implicitly: libarena's buddy allocator rounds a
+	 * request up to a power of two and returns blocks aligned to their own
+	 * size, so an allocation of @data_size bytes is at least @align aligned
+	 * for any @align no larger than it.
+	 */
+	task_ctx_size = data_size;
```
arighi: `align` is still accepted but ignored, the previous implementation validated and honored it. Can we validate it and size the allocation accordingly (or reject in case of unsupported alignment)?

arighi (commented):
Left few comments. I'm also not a big fan of submodules, but this has been already discussed and agreed on a Cargo-based approach, which looks nicer. Thanks.

scheds/rust/scx_lavd/src/bpf/lavd.bpf.h
etsal: That is fair, this is a combination of a) moving to arena-based reporting for lavd and b) 6.13 not supporting arenas. Let's keep the reporting as printk's for now while we still support < 6.16 kernels.

lib/sdt_task.bpf.c
etsal: Good point, this is a regression. I'll update this to honor alignment.

etsal:
Added a v2 with the proposed structure. If it looks good I will split it into a base series that is mergeable by itself then send out the diffs for the individual schedulers (then do the final removal once those come through). Submodules are gone and the libarena-rs include happens through Cargo.

multics69:
@etsal -- libarena-rs v0.3.0 hasn't been published yet. 

```
$ cargo build --profile release                                                              ─╯
warning: /home/changwoo/ws-multics69/dev/scx-upstream/rust/scx_rustland_core/Cargo.toml: unused manifest key: lib.include
    Updating crates.io index
error: failed to select a version for the requirement `libarena-rs = "^0.3.0"`
candidate versions found which didn't match: 0.2.0, 0.1.0
location searched: crates.io index
required by package `scx_arena v1.1.3 (/home/changwoo/ws-multics69/dev/scx-upstream/rust/scx_arena/scx_arena)`
```

etsal:
> @etsal -- libarena-rs v0.3.0 hasn't been published yet.
> 
> ```
> $ cargo build --profile release                                                              ─╯
> warning: /home/changwoo/ws-multics69/dev/scx-upstream/rust/scx_rustland_core/Cargo.toml: unused manifest key: lib.include
>     Updating crates.io index
> error: failed to select a version for the requirement `libarena-rs = "^0.3.0"`
> candidate versions found which didn't match: 0.2.0, 0.1.0
> location searched: crates.io index
> required by package `scx_arena v1.1.3 (/home/changwoo/ws-multics69/dev/scx-upstream/rust/scx_arena/scx_arena)`
> ```

I'll publish it soon, in the meantime can you clone locally and amend the Cargo.toml to use the local git tree? (Repo is https://github.com/prozak/libarena-rs)

multics69:
@etsal -- I fixed a few small problems to make it run on v6.13 kernel without verifier errors. Please check the following PRs. 

https://github.com/libbpf/libarena/pull/5
https://github.com/sched-ext/scx/pull/3824

PR #3805, opened 2026-09-12
scx_lavd: scale the compete window by online CPUs, not active ones

multics69:
3ff16ec62 ("scx_lavd: Scale the deadline compete window by the oversubscription ratio") shrinks the window by nr_active / nr_queued. Core compaction changes nr_active in steps at every sys_stat interval, so tasks enqueued around a step get windows of different size and vtimes that no longer compare. It also sizes the active set for about half utilization, so nr_queued exceeds nr_active at light load and the burst boost shrinks on a machine that is not oversubscribed.

Use the number of online CPUs, which changes only on hotplug, so the window moves only with the averaged queue length. This eliminates the micro-stuttering under mild system load with no regression in hackbench.


arighi (approved):

multics69:
Thanks @arighi

PR #3809, opened 2026-09-15
scx_arena: Fall back to map-in-map grace periods where membarrier cannot wait

htejun:
scx_urcu reclaims arena memory after a userspace thread waits for an RCU grace period with membarrier(MEMBARRIER_CMD_GLOBAL). That command fails with EINVAL when nohz_full is active, which made the daemon exit and silently stopped reclaim for the rest of the scheduler's life. It also returns without waiting when only one CPU is online, so on a preemptible kernel a reader preempted inside its read-side section could see its object reclaimed.

Fall back to updating a private map-in-map slot in both cases. The kernel documents that update as waiting for the running non-sleepable programs. It is an expedited grace period and IPIs every CPU that is in the kernel, so the daemon paces those rounds 100ms apart instead of 1ms and warns once. The online count is checked before and after membarrier, and the sched_ext hotplug counter is compared across the call to catch a hotplug racing it. That counter moves at the active transitions, not when the online mask changes, so a CPU leaving the mask and another entering inside one call can go unnoticed when the daemon is descheduled between them. A moved counter with CPUs left retries membarrier a few times before the fallback takes over.

Probe membarrier once at setup so an unexpected failure is reported before attaching, and terminate the scheduler if the daemon fails later instead of losing reclaim silently.

Fixes: 05b6d696ada5 ("scx: lib - Add scx_task/cgrp_free_rcu(), poor man's RCU deferred freeing")

Testing: built scx_arena and the schedulers that include the arena library, cargo fmt clean. In a 16-CPU VM on a debug kernel, a scheduler using the deferred-free path ran under fork churn with CPUs 1-15 offlined and separately on a nohz_full=8-15 boot. Each run printed the single warning with the expected reason, kept reclaiming, and exited cleanly with no kernel diagnostics. The retry path needs a hotplug racing the call and was not exercised.


rust/scx_arena/scx_arena/src/lib.rs
```diff
@@ -117,10 +122,169 @@ fn urcu_run_prog(fd: &OwnedFd) -> Result<u32> {
...
+                if online_cpus()? <= 1 {
+                    break;
+                }
+                let seq = hotplug_seq()?;
+
+                membarrier_global().context("membarrier(GLOBAL) failed")?;
+                if online_cpus()? <= 1 {
+                    break;
+                }
+                if hotplug_seq()? == seq {
```
arighi, line 260: For schedulers that restart on hotplug, this is mostly mitigated because all readers disappear once shutdown completes. However, scx_exit() queues asynchronous disable work, so the reclaim daemon can still run in the interval between the restart request and completed detachment. And hotplug-aware schedulers do not restart at all.

Should we document this as best-effort, since it reduces the exposure but does not fully guarantee a grace period across hotplug?

rust/scx_arena/scx_arena/src/lib.rs
```diff
@@ -157,12 +322,13 @@ fn urcu_daemon(
         /* run one final drain below before honoring a stop request */
         let stopping = fds[1].revents != 0;
 
-        /* drain until nothing is awaiting reclaim */
+        /* drain until nothing is awaiting reclaim, unpaced once stopping */
+        let pace = if stopping { Duration::ZERO } else { interval };
```
arighi, line 326: `pace` is captured once before the inner drain loop, but interval is updated inside that loop below. On nohz_full, the first fallback sets interval to 100 ms, while pace remains at its initial 1 ms for the rest of this drain cycle.

With continuous frees, the inner loop can keep running and issue expedited map-in-map grace periods every 1 ms instead of every 100 ms, causing up to 100x the intended system-wide IPI rate. Could we calculate pace inside the inner loop so each iteration observes the latest interval?

arighi (approved):
Thanks for looking at this!

I made a couple of comments, the `pace` one would nice to fix before merging, if possible, while the other one can be just documented and potentially improved later (not a blocker).

multics69 (approved):
Abuing map-in-map update is great!

rust/scx_arena/scx_arena/src/lib.rs
htejun: Good catch, thanks. Moved the pace computation into the inner loop so each iteration uses the current interval.

rust/scx_arena/scx_arena/src/lib.rs
htejun: Agreed, it is best effort. Added a sentence to the comment: the hole stays open for a scheduler that runs through hotplug, and for a restarting one until the asynchronous disable completes.

PR #3814, opened 2026-09-17
lib, scheds/include: cid and arena library prep for cid-form schedulers

htejun:
Library and header changes from an upcoming arena-native cid conversion. Each one stands on its own:

- scheds/include: sync the scx_bpf_online_cmask() declaration from the kernel.
- lib/cgroup_bw: the replenish timer wakes dispatch through the built-in idle tracking, which a scheduler with its own idle tracking cannot use. A load-time knob, bw_kick_builtin_idle, selects the built-in kick; a scheduler that leaves it off overrides the weak scx_cgroup_bw_kick_idle_cb(). lavd sets the knob.
- scheds/include: cmask_for_each() is bounded with can_loop, and cmask_weight() converges under the verifier when a caller compares its sum.
- scheds/include: the cmask distribute rotor moves from a per-CPU array map into the arena, one cacheline slot per cid, allocated by the shared arena init.
- lib: __scx_task_data() reads the task-storage association once, and sdt_task.h documents the borrowed-pointer contract.
- lib: the unused per-CPU cpumask and bitmap scratch and the scx_bitmap to cpumask bridge helpers are dropped.

Testing: scx_lavd, scx_p2dq, scx_chaos, scx_cidland, scx_arena and the lib selftests build and cargo fmt is clean. scx_lavd ran in a 16-CPU VM on a 7.2-rc6 sched_ext kernel with cpu.max enabled, which exercises the built-in replenish kick, and a WIP cid-form scheduler built on these changes ran its smoke and cpu.max phases with the callback override, the arena rotor and the bounded traversals.


scheds/include/scx/cid.bpf.h
```diff
@@ -970,28 +971,63 @@ static __always_inline u32 cmask_next_and_set_wrap(const struct scx_cmask __aren
...
+{
+	s32 cid = scx_bpf_this_cid();
+
+	if (unlikely(!cmask_distribute_rotors)) {
+		scx_bpf_error("cmask distribute rotor not allocated");
+		return NULL;
+	}
+	if (cid < 0)
+		return NULL;
+	return &cmask_distribute_rotors[cid].cid;
```
arighi, line 1024: Could we retain nr_slots from cmask_distribute_init() and check cid < nr_slots here? The shared arena_init() path currently allocates enough slots, but a scheduler using its own arena could accidentally undersize the allocation and turn this into an out-of-bounds arena access. Returning NULL (and perhaps reporting scx_bpf_error()) would preserve the fallback behavior without risking arena corruption and obscure errors.

scheds/include/scx/cid.bpf.h
```diff
@@ -970,28 +971,63 @@ static __always_inline u32 cmask_next_and_set_wrap(const struct scx_cmask __aren
...
+ * The shared arena init calls this once. A scheduler with its own arena calls
+ * it before the first distribute pick. Return 0 on success, -ENOMEM if the
+ * allocation fails.
+ */
+static __always_inline int cmask_distribute_init(void *map, u32 nr_slots)
+{
+	u64 size = nr_slots * sizeof(*cmask_distribute_rotors);
+	u32 pages = (size + PAGE_SIZE - 1) / PAGE_SIZE;
+
+	cmask_distribute_rotors = bpf_arena_alloc_pages(map, NULL, pages, NUMA_NO_NODE, 0);
```
arighi, line 1010: Non-blocking for this PR (something to keep in mind as a potential future optimization): this replaces NUMA-local per-CPU map storage with one bulk arena allocation using NUMA_NO_NODE, so the rotor pages may all be allocated near the CPU running arena_init().

It'd be interesting to measure if there's any benefit, on NUMA systems with expensive cross-node access, to allocate per-node rotor banks (or individual rotor pages) with an explicit node_id.

arighi (approved):
Left a couple of comments, both non blocking. Everything else looks good to me.

multics69 (approved):

scheds/include/scx/cid.bpf.h
```diff
@@ -971,28 +971,63 @@ static __always_inline u32 cmask_next_and_set_wrap(const struct scx_cmask __aren
...
```
multics69, line 1010: NUMA-based page placement will be a nice optimization. It is okay for now.

scheds/include/scx/cid.bpf.h
htejun: Dropped the argument instead: cmask_distribute_init() now sizes the array from scx_bpf_nr_cids() itself, and allocates nothing on kernels without the cid kfuncs since nothing can pick there. No caller can undersize it, and the accessor keeps its single compare. Force-pushed, rebased on current main.

scheds/include/scx/cid.bpf.h
htejun: bpf_arena_alloc_pages() honors node_id, so per-node banks are doable, but even with each node's cids contiguous the accessor would need a cid-to-node lookup and a per-node base on every pick, an indirection the flat array avoids. A slot is rewritten by its one CPU, so the line should stay hot in that CPU's cache anyway.

Longer term the right tool is an in-arena per-cid allocator along the lines of the kernel's percpu allocator: every per-cid object gets node-local pages, and the cacheline padding that keeps CPUs apart in one block goes away. Leaving node awareness to that and noting it in the comment. @etsal FYI, since that would live in the library.

PR #3815, opened 2026-09-17
scx_lavd: Fix the autopilot flag binding, torn task_ctx cache fills and futex return reads

htejun:
Three lavd fixes found while porting it to the cid interface. They stand alone, but the task_ctx fix pairs with the library prep PR, which makes the underlying task-data lookup read its association once.

- Autopilot flag binding: sys_stat.bpf.c carries a weak duplicate of is_autopilot_on next to the strong definition in util.bpf.c. Both survive the link at different .rodata offsets, the Rust skeleton writes one byte and the BPF side reads another that stays zero, so in the default autopilot mode do_autopilot() never runs and lavd stays in its initial power mode. Verified on a fresh main build; the weak duplicate is removed.
- Torn task_ctx cache fills and the racy parent lookup: the per-CPU lookup cache was filled from the preemptible futex and execve hooks and from the sleepable init_task(), which can tear an entry or write another CPU's, and the parent lookup in init_task() raced with the parent's exit. Own-task lookups keep the cache, every other lookup goes through the new uncached find_task_ctx(), the parent copy sits in an RCU read-side section and exit_task() frees through scx_task_free_rcu().
- Futex return reads: the fexit hooks read the return value through bpf_get_func_ret() instead of a positional argument, so a kernel that adds a futex argument no longer hands them that argument as the result.

Testing: scx_lavd builds, cargo fmt is clean, and scx_lavd ran in a 16-CPU VM on a 7.2-rc6 sched_ext kernel with cpu.max enabled.


arighi (approved):

multics69 (approved):
Thanks @htejun ! LGTM

PR #3819, opened 2026-09-17
scx_arena: Retry arena page allocations and fail the allocation, not …

arighi:
…the scheduler

A burst of parallel task creation takes the scheduler down from the task context allocator:
```
  sched_ext: BPF scheduler "cidland" disabled (scx_bpf_error)
  sched_ext: src/bpf/lib/sdt_alloc.bpf.c:621: failed to allocate data from pool
     scx_bpf_error_bstr
     bpf_prog_..._scx_alloc_internal
     bpf_prog_..._cidland_init_task
     __scx_init_task / scx_fork / copy_process / kernel_clone
```
bpf_arena_alloc_pages() returned NULL although the machine had 1.9 GB of free memory and 26 GB of page cache. It fails that easily because of how the kernel allocates arena pages: it holds the arena's raw spinlock with interrupts disabled, and in that context __bpf_alloc_page() cannot use a normal allocation and falls back to alloc_pages_nolock(). That allocator only tries to lock the per-CPU and zone free lists, never reclaims, and returns NULL when another CPU holds those locks or free memory is at the zone's minimum watermark; the kernel documents it as "best effort and to be expected to fail easily".  Passing the kfunc's sleepable flag makes no difference, it only selects how pages are freed. The library then treated that first NULL as fatal.

Retry the allocation a bounded number of times: contention on the free lists is over in microseconds. When the kernel still refuses, which happens once a fork storm outruns reclaim and free memory sits at the watermark, count it in alloc_stats.alloc_nomem and return NULL, so the caller fails that one fork with -ENOMEM instead of taking every task with it. scx_task_alloc() and scx_cidland already return -ENOMEM on a NULL context.

Reproducer, on a 20-CPU desktop with 32 GB running 7.2.3, three rounds under the scheduler:
```
  python3 storm.py 20 1500 2

  import os, sys, threading, time
  procs, per, hold = int(sys.argv[1]), int(sys.argv[2]), float(sys.argv[3])
  pids = []
  for _ in range(procs):
      pid = os.fork()
      if pid == 0:
          ev = threading.Event()
          ts = [threading.Thread(target=ev.wait) for _ in range(per)]
          for t in ts: t.start()
          time.sleep(hold); ev.set()
          for t in ts: t.join()
          os._exit(0)
      pids.append(pid)
  for p in pids: os.waitpid(p, 0)
```
Results with scx_cidland:

 - Before: the scheduler dies two seconds after attach, in the first round, with the error above. Every round of the storm reproduces it.

 - Attempts needed, traced with a bpf_printk() in the retry loop: at 30000 threads one allocation succeeded on its second attempt. At 60000 threads, where the storm also runs into the session's pids limit, three allocations retried, two twice and one seven times, and one exhausted all 64 attempts. That last case is the one now reported as -ENOMEM to the caller instead of aborting.

 - After: 30000 and 50000 threads, three rounds each, the scheduler survives with zero failed allocations. The arena settles at 24.3 MiB and 32.3 MiB, as the pools keep their slabs.


htejun:
Is your init_task sleepable? Sleepable allocations shouldn't fail.

arighi:
> Is your init_task sleepable? Sleepable allocations shouldn't fail.

Yes, init_task is sleepable and the verifier picks the sleepable kfunc, but since b8467290edab ("bpf: arena: make arena kfuncs any context safe") arena_alloc_pages() calls bpf_map_alloc_pages() under raw_res_spin_lock_irqsave(), so __bpf_alloc_page() sees irqs disabled and goes through alloc_pages_nolock() for sleepable callers too...

lib/sdt_alloc.bpf.c
```diff
@@ -24,6 +24,30 @@ struct scx_alloc_stack __arena *prealloc_stack;
...
 static __u64 zero = 0;
 
+/*
+ * Arena pages are handed out by the kernel under the arena's raw spinlock with
+ * interrupts disabled, so they come from alloc_pages_nolock(): a trylock on the
+ * per-CPU and zone free lists, no reclaim, and NULL as soon as another CPU
+ * holds them. That failure is transient, twenty CPUs creating tasks at once hit
+ * it within seconds, so retry a bounded number of times before treating it as
+ * out of memory.
+ */
```
htejun, line 38: Let's clearly note that this is a workaround for a kernel problem. This shouldn't be happening.

htejun (approved):

PR #3822, opened 2026-09-18
clippy: Address workspace clippy and rustdoc warnings

lespea:
## Description

This PR addresses compiler, Clippy, and Rustdoc warnings across the workspace to keep CI clean and ensure adherence to default lint levels.

No intentional functional or behavioral changes are introduced.

### Summary of Changes

The fixes cover the following categories across 58 files:

- **Documentation & Rustdoc**:
    - Wrapped bare URLs in angle brackets (`rustdoc::bare_urls`).
    - Fixed markdown list indentation, code blocks, and doc comment escaping (`clippy::doc_markdown`, `clippy::doc_lazy_continuation`, `rustdoc::broken_intra_doc_links`, `clippy::needless_doctest_main`).
- **Function Signatures & Complex Types**:
    - Allowed `clippy::too_many_arguments` on test helpers and internal complex functions.
    - Simplified complex signatures by factoring aliases (`OpenOpsEntry`, `CpuTierTuple`) and allowed `clippy::type_complexity` where factoring is impractical.
    - Accepted slices (`&[T]` and `&str`) instead of `&Vec<T>` and `&String` (`clippy::ptr_arg`).
- **Control Flow & Expressions**:
    - Collapsed nested `if` / `if let` blocks into let-chains (`clippy::collapsible_if`).
    - Simplified `expr != true` boolean checks (`clippy::bool_comparison`).
    - Used `.is_none()` and `.contains()` / `!contains_key()` instead of `.find().is_some()` or `.get().is_none()` (`clippy::search_is_some`, `clippy::manual_contains`).
    - Inlined string literal in `writeln!` formatting (`clippy::print_literal`).
- **Loops & Collections**:
    - Replaced index-based range loops with direct iterators, `.iter().take(...)`, or `.enumerate()` (`clippy::needless_range_loop`).
    - Replaced heap `vec![]` with fixed-size arrays where appropriate (`clippy::useless_vec`).
    - Used slice operations (`copy_from_slice`, `.fill(true)`).
- **Struct Updates & Idioms**:
    - Adopted struct update syntax `..Default::default()` and dropped redundant `..self.clone()` in `Metrics::sub`.
    - Hoisted `Regex` compilation outside of test loops (`clippy::regex_creation_in_loops`) and used `into_owned()` on `Cow`.
    - Used standard constants like `std::f64::consts::SQRT_2`, lowercase hex literals, and `(min..=max).contains()`.
- **Cargo & Manifest**:
    - Moved `include` from `[lib]` to `[package]` in `scx_rustland_core/Cargo.toml`.
- **Allow Attributes for False Positives**:
    - Allowed `arc_with_non_send_sync`, `deref_addrof`, `unwrap_used` (in tests), `erasing_op` (in tests), and `clippy::all` on generated `perf_bindings.rs`.

---

### Note on Commit History

The branch currently contains **17 individual commits grouped by lint/fix type** to facilitate granular review. If the maintainers prefer a single monolithic commit, I am happy to squash them before merging.

---

### Attribution

Pair-programmed and authored in collaboration with **Antigravity** (Google DeepMind).

Co-authored-by: Antigravity <antigravity@google.com>


htejun (approved):

htejun (changes requested):
Sorry, was too quick. Will respond further.

htejun (changes requested):
Thanks for the sweep, and sorry to re-scope after the fact. Going through this made me realize I don't want to carry clippy's idiom lints as a standard for this tree. The bar I want is simple: if a change clearly reads better or is less error prone, it goes in; if it's a rule for the rule's sake, it doesn't, and 21 `#[allow]`s just to quiet a tool isn't something I want in the code either. So the ask is: keep the changes that pass that bar, drop the rest, and silence the lints we're opting out of once, in a `[workspace.lints]` table, instead of per site. This review is AI assisted; I've skimmed the details rather than gone over every line.

By commit:

- Keep as is: 5 (slices), 7 (`!= true`), 8 (`contains`/`is_none`), 10 (struct update, dead `..self.clone()`), 11 (`fill`, `copy_from_slice`, `vec![]`), 12 (casts), 13 (regex hoist, `into_owned`), 16 (manifest, see below).
- Keep 1 and 2 (docs) except: the doc lines on clap fields, which are the `--help` text and now print angle brackets and backticks (`--log-level` in several schedulers, `scx_mlfq --no-webui`, `scx_forge_agent --crate-dir`/`--spec`, `scxtop --default-profiling-event`); the `\[0,4\]` escape in the C header `scx_mlfq/src/bpf/intf.h` (`[0, 4]` with a space is what the rest of the header does and doesn't warn); and the tab swap in `scx_cidland/src/main.rs:725`, which still doesn't render as a code block.
- Keep 14 minus the range-contains hunk, and 15 minus the `arc_with_non_send_sync` and `deref_addrof` allows. The `unwrap_used` allows on test modules follow existing precedent in those crates, and `clippy::all` on the generated bindings and the test-local `erasing_op` are fine.
- In 9, keep the hunks where the index disappears (`for x in &arr`, slice ranges, `enumerate()` replacing a hand-rolled counter). Drop the ones that turn an index range into an iterator chain, and the function-level allows. `for (cell, cell_stats) in cell_stats_delta.iter().enumerate().take(MAX_CELLS)` replacing `for cell in 0..MAX_CELLS` over a `[_; MAX_CELLS]` array is the lint making the code worse: it's longer, the `.take()` of the array's own length does nothing, and the index is still there. That is exactly the kind of change I don't want. The `.take(4)` in `flow_tests_group.rs:1246` also drops the length check that test had.
- Drop 3, 4, 6 and 17.
- Add a first commit with `[workspace.lints.clippy]` allowing `too_many_arguments`, `type_complexity`, `needless_range_loop`, `collapsible_if`, `manual_range_contains`, `arc_with_non_send_sync`, `deref_addrof` and `write_literal`, plus `[workspace.lints.rustdoc] bare_urls = "allow"`, with `[lints] workspace = true` in the crates. Crates that already have a `[lints.clippy]` table can move their allows into the workspace table; the two that deny `unwrap_used` keep a local table and repeat the allows there, since a crate can't both inherit and add.

On 16: the `[lib] include` was inert (that is the cargo warning), so the published crate shipped every tracked file. The `[package] include` is now a real whitelist and the tarball loses `LICENSE`, `bindings.h` and the `bpf_h` symlink. The latter two are unreferenced and can go, but please add `LICENSE` back and call the package-content change out in the commit message.

Smaller things: the `ver_re`/`sha1_re` hoist landed in commit 8 while 13 removes the inline regexes, so 8 through 12 build with unused-variable warnings; the `chunks_exact_to_as_chunks` allow was an unknown lint name (`unknown_lints`), not an unused allow; and "keep CI clean" overstates it, since CI runs clippy only on the `use_clippy` crates, which cover 11 of these 58 files.

lespea:
Thanks for the thoughtful review and feedback! I completely agree with keeping the bar focused on readability and maintainability rather than chasing lints for their own sake.

With assistance from Antigravity, I’ve restructured and rebased the branch to address all the review directives:

### 1. Workspace-Level Lint Configuration (New First Commit)

• Added [workspace.lints.clippy] to the root Cargo.toml opting out of idiom lints globally: too_many_arguments, type_complexity, needless_range_loop, collapsible_if, manual_range_contains, arc_with_non_send_sync, deref_addrof, write_literal, and tabs_in_doc_comments.
• Added [workspace.lints.rustdoc] with bare_urls = "allow".
• Moved existing crate-level allows (not_unsafe_ptr_arg_deref and non_canonical_partial_ord_impl) into the workspace table and enabled [lints] workspace = true across the crates.
• scx_mitosis and scx_nitosis (which deny unwrap_used) maintain a local table repeating the workspace allows.

### 2. Dropped Commits

• Dropped commit 3 (too_many_arguments), commit 4 (type_complexity), commit 6 (collapsible_if), and commit 17 (write_literal).

### 3. Kept Commits

• Kept commits 5 (slice references), 7 (!= true), 10 (struct update syntax & dead ..self.clone()), 11 (arrays & slice operations), and 12 (type casts) as is.

### 4. Docs Refinements (Commits 1 & 2)

• Reverted all doc changes on clap fields (--log-level in several schedulers, scx_mlfq --no-webui, scx_forge_agent --crate-dir/--spec, and scxtop --default-profiling-event) so CLI --help output remains clean without brackets or backticks.
• Replaced \[0,4\] in scx_mlfq/src/bpf/intf.h with [0, 4] (space after comma), aligning with the rest of the header.
• Formatted the C snippet in scx_cidland/src/main.rs as a fenced ```c block so it renders as an actual code block in rustdoc.

### 5. needless_range_loop Cleanup (Commit 9)

• Kept only hunks where indexing disappears: slice ranges (scx_flow, scxtop), direct slice loops (scx_pandemonium/src/adaptive.rs), and .enumerate() replacing hand-rolled counters (scx_pandemonium/src/chaos.rs).
• Dropped iterator chains (take(), etc.) in scx_flow, scx_nitosis, scx_mitosis, scx_layered, and scx_pandemonium, and dropped all function-level allow attributes.

### 6. Regex Hoist & Intermediate Warnings (Commits 8 & 13)

• Cleaned up commit 8 so it only touches contains / is_none.
• Consolidated the ver_re and sha1_re compilation hoist, .into_owned(), and inline regex removal together in commit 11, eliminating unused-variable warnings in intermediate commits.

### 7. Range Contains & Domain Allows (Commits 14 & 15)

• Dropped the manual_range_contains hunk in commit 14 (now silenced via workspace lints), keeping 0xeb9f and SQRT_2.
• Dropped site-level arc_with_non_send_sync and deref_addrof allows in commit 15; kept unwrap_used on test modules, clippy::all on generated bindings, erasing_op, and noted the removal of unknown lint allow chunks_exact_to_as_chunks.

### 8. scx_rustland_core Manifest Whitelist (Commit 16)

• Added LICENSE back into the include array in rust/scx_rustland_core/Cargo.toml.
• Updated the commit message to explain that moving include from [lib] to [package] activates the manifest whitelist (shipping src, BPF assets, Cargo.toml, and LICENSE while omitting unreferenced bindings.h and the bpf_h symlink).

### Validation

• cargo clippy --workspace --all-targets -- -D warnings: Clean (0 warnings, 0 errors).
• CI clippy check (use_clippy packages): Passed with 0 warnings.
• cargo test --workspace --doc: All doc tests pass.

htejun (changes requested):
Thanks, this is close. Three more things:

- Drop `unwrap_used = "deny"` from `scx_mitosis` and its experimental sibling and let them inherit the workspace table like every other crate. A panic's backtrace is enough; the deny buys little for the churn it causes. With it gone, the `#[allow(clippy::unwrap_used)]` attributes in those two crates are dead, the two new ones and the pre-existing ones on the `cell_manager` tests and `bpf_skel`, so drop those too.
- In the `needless_range_loop` commit, drop the `&t[..4]` / `&t[4..8]` / `&t[8..16]` loops in `scx_flow/src/flow_tests_group.rs` and the slice-range loops in the two scxtop tests. The previous code is shorter and easier to read. The `enumerate()` replacements for the hand-rolled counters in `chaos.rs` and the `delta_hist` loop in `adaptive.rs` are fine.
- In the regex commit, delete `.to_owned()` instead of changing it to `.into_owned()`. The `Cow` already worked for every use, so the call just goes.

lespea:
Okay hopefully this should cover everything

likewhatevs:
> Drop unwrap_used = "deny" from scx_mitosis and its experimental sibling and let them inherit the workspace table like every other crate. A panic's backtrace is enough; the deny buys little for the churn it causes. With it gone, the #[allow(clippy::unwrap_used)] attributes in those two crates are dead, the two new ones and the pre-existing ones on the cell_manager tests and bpf_skel, so drop those too.

> A panic's backtrace is enough; the deny buys little for the churn it causes.

This has not been the case which, iirc, is why that is there. @dforsyth 

Edit: IIRC, that lint catches places where with_context (or something like it) isn't used w/ unwrap, which lead to hard-to-impossible to debug backtraces.

dforsyth:
> > A panic's backtrace is enough; the deny buys little for the churn it causes.

What's the churn here? Code or in our traces? New code already has to respect this.

htejun:
Some sites needing explicit explanation doesn't mean that all usages of `.unwrap()` is wrong. I actively dislike this sort of rules imposed unconditionally which leads to unnecessary and sometimes non-sensical annotations on something which truly shouldn't fail (e.g. you know logically the condition can't be reached). These make code tedious to write and read. This is a hard nack from me. If this is a binary decision, I'd rather just ditch clippy entirely.

PR #3827, opened 2026-09-20
scx_nitosis: preserve arena pointer type in CID topology init

liulangrenaaa:
mitosis_init() uses a loop to snapshot the CID topology into an arena-resident array. The combination of bpf_for() and an aggregate assignment causes the generated arena destination to fail verification with -EACCES on current compiler/verifier combinations.

Use bpf_arena_for() and populate the destination through an explicit __arena pointer field by field. This avoids the aggregate copy and keeps the arena pointer type visible to the verifier.

Tested on Linux 7.3.0-rc3-g520aa469e2ee with clang 20. All scx_nitosis BPF programs load successfully and the scheduler reaches the sched_ext enabled state in QEMU.

Fixes: 8da3f08a55b5 ("scx_nitosis: Convert to the cid form of sched_ext ops")


htejun:
Thanks. Reproduced with clang 20.1.8 on 7.3-rc3: mitosis_init fails verification on main and passes with this change. clang 22.1.8 verifies main as is.

The description doesn't match what's going on though. It isn't the combination of bpf_for() and the aggregate assignment. Building the two halves separately with clang 20:

- field-by-field copy only, bpf_for() kept: passes
- bpf_arena_for() only, struct copy kept: fails at the same instruction
- __builtin_memcpy() into the arena slot: fails at the same instruction

LLVM 20 derives the destination of the struct copy from the allocation result before the addr_space_cast, so the copy's stores go through a value the verifier still types as a scalar:

```
137: (79) r1 = *(u64 *)(r10 -96)      ; R1=scalar(id=414)
138: (0f) r1 += r8
140: (63) *(u32 *)(r1 +24) = r2
R1 invalid mem access 'scalar'
```

Every other access to the topo struct in the function is cast correctly, and clang 22 casts the copy destination too.

The bpf_arena_for() conversion is fine to keep, we're moving everything over anyway. Please rewrite the description around the actual cause: the compiler version, the struct copy into arena memory losing the address space cast, and the verifier message. "Current compiler/verifier combinations" overstates it.

htejun:
Can you also add a comment explaining why the copy needs to be open coded?

htejun (approved):

